PR #3392, opened 2022-07-18
Adding sed as a post install requirement in the vim spec

aadhar-agarwal:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

- [Bug 40363041](https://microsoft.visualstudio.com/OS/_workitems/edit/40363041?src=WorkItemMention&src-action=artifact_link) indicates that when installing vim into a Mariner 2.0 container, the post-install script fails because the sed tool is not there by default.
- While working on this issue, I noticed that the sed package exists in the Mariner 2.0 container, but I have added `Requires(post): sed` in `SPECS/vim/vim.spec` file to ensure that the sed tool is present when the post-install script runs.

###### Change Log  

- Added a `Requires(post): sed` in the `SPECS/vim/vim.spec` file

**NO**

###### Test Methodology

- Local build

- Using rpm to manually install the updated vim package (fails gracefully indicating that sed is required)
- Using tdnf to install vim successfully
vim_testing

### DIFF
--- a/SPECS/vim/vim.spec
+++ b/SPECS/vim/vim.spec
@@ -2,7 +2,7 @@
 Summary:        Text editor
 Name:           vim
 Version:        9.0.0050
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        Vim
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -13,6 +13,8 @@ BuildRequires:  ncurses-devel
 BuildRequires:  python3-devel
 Provides:       vi = %{version}-%{release}
 Provides:       %{name}-minimal = %{version}-%{release}
+
+Requires:       sed
 
 %description
 The Vim package contains a powerful text editor.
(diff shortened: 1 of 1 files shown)


SPECS/vim/vim.spec
```diff
@@ -30,7 +31,7 @@ The vim extra package contains a extra files for powerful text editor.
 %prep
 %autosetup -p1
 echo '#define SYS_VIMRC_FILE "%{_sysconfdir}/vimrc"' >> src/feature.h
-%py3_shebang_fix runtime/tools/demoserver.py
+%{py3_shebang_fix} runtime/tools/demoserver.py
```
PawelWMS, line 34: I think this will make the macro run without any arguments and fail. Please double-check, though.
```suggestion
%py3_shebang_fix runtime/tools/demoserver.py
```

PawelWMS (changes requested):
One comment in the spec.

Please also update the PR's title to reflect what changes, instead of describing the issue. This makes summarizing our releases easier.

SPECS/vim/vim.spec
```diff
@@ -11,6 +11,7 @@ URL:            https://www.vim.org
 Source0:        https://github.com/%{name}/%{name}/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
 BuildRequires:  ncurses-devel
 BuildRequires:  python3-devel
+Requires:       sed
```
PawelWMS, line 14: If it's required by the post-install script, then please user `Requires(post)`. Otherwise, `sed` might be installed after `vim` and the script will still fail.

PawelWMS (changes requested):
Please also fill out the PR's checklist and non-applicable items should be left checked.

oliviacrain (approved):
Looks great- congrats on your first Mariner PR!

aadhar-agarwal:
> Looks great- congrats on your first Mariner PR!

Thank you! :)

PR #3395, opened 2022-07-19
Update `e2fsprogs`: running tests in a single thread to make them more consistent.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Some of the `e2fsprogs` ptests are failing inconsistently and running them in a single thread seems to make them more stable. They still fail every now and then, but it doesn't look like a regression. I've added a few lines to print more logs for failing tests to make it easier for us to debug in the future.

###### Change Log  

- Running `e2fsprogs` tests in a single thread.
- Printing logs for failing tests to ease debugging.

Yes.

###### Test Methodology

- Pipeline build 215120.

### DIFF
--- a/SPECS/e2fsprogs/e2fsprogs.spec
+++ b/SPECS/e2fsprogs/e2fsprogs.spec
@@ -1,7 +1,7 @@
 Summary:        Contains the utilities for the ext2 file system
 Name:           e2fsprogs
 Version:        1.46.5
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv2 AND LGPLv2 AND BSD AND MIT
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -62,7 +62,8 @@ rm -rf %{buildroot}%{_infodir}
 %find_lang %{name}
 
 %check
-make %{?_smp_mflags} check
+# Mult-threaded runs are flaky.
+make -j1 check
 
 %post -p /sbin/ldconfig
 %postun -p /sbin/ldconfig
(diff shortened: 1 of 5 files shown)


SPECS/e2fsprogs/e2fsprogs.spec
```diff
@@ -62,7 +62,17 @@ rm -rf %{buildroot}%{_infodir}
 %find_lang %{name}
 
 %check
-make %{?_smp_mflags} check
+# Mult-threaded runs are flaky.
+make -j1 check
```
oliviacrain, line 66: As far as I know, our package test scripts use the last exit code output in the `%check` section to determine whether tests failed or not. Could we save the exit code of the call to `make`, and use that to ensure the proper code is returned? My concern is that having the log output here will mask the actual exit code.

PawelWMS: Ah, right, good catch. Let me fix that.

jslobodzian (approved):

SPECS/e2fsprogs/e2fsprogs.spec
```diff
@@ -62,7 +62,22 @@ rm -rf %{buildroot}%{_infodir}
 %find_lang %{name}
 
 %check
-make %{?_smp_mflags} check
+# Multi-threaded runs are flaky.
+make -j1 check
+test_status=$?
+
+for failed_test in $(find tests -name "*.failed")
```
mfrw, line 69: 🎉
I think this is pretty great :)

mfrw (approved):

PR #3401, opened 2022-07-20
Add ability to build quickly by packing a subset of RPMs

oliviacrain:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 
One of the larger time/network/computational bottlenecks when building locally is (a) packing/expanding/reading SRPMs and (b) resolving the package graph nodes of each package generated by such SRPMs. This bottleneck doesn't go away even when we specify a subset of packages to build using `PACKAGE_BUILD_LIST` and `PACKAGE_REBUILD_LIST`- we always build the full set of SRPMs.

One of the ways people get around that bottleneck is by copying the desired spec folders to build to a separate folder, and setting `SPECS_DIR` to that separate folder. Doing so ensures that only the specs in the separate folder are built as SRPMs. This has the consequence of thinning out the package graph to only include what is needed/generated by the smaller set of SRPMs.

This is a great method to cut down local build times during development, but as with any manual process, it's error prone. This PR adds the ability to build quickly by only packing a subset of RPMs under `SPECS_DIR`. This achieves the same goals as the previous method, but without the manual copying and variable setting.

###### Change Log  

- Add `SRPM_PACK_LIST` make flag, which triggers the SRPM packer to only pack the specs/srpms in that list.

**NO**

###### Test Methodology

- Manual validation of spec build workflow, ensuring only a subset of SRPMs are built and that 
- A simple formula for testing yourself:
```bash
# From a fresh checkout of 2.0-stable
cd toolkit
sudo make toolchain REBUILD_TOOLCHAIN=n
sudo make build-packages SRPM_PACK_LIST="vim" REBUILD_TOOLS=y REPO_LIST="<repo_root>/toolkit/repos/mariner-official-base.repo"
```


toolkit/scripts/srpm_pack.mk
```diff
@@ -14,12 +14,26 @@
...
+toolchain_spec_list = $(toolchain_build_dir)/toolchain_specs.txt
+fast_build_pack_list = $(BUILD_SRPMS_DIR)/fast_build_pack_list.txt
+
+ifeq ($(FAST_BUILD),y)
+spec_names_build_or_rebuild = $(filter-out $(PACKAGE_IGNORE_LIST),$(strip $(PACKAGE_BUILD_LIST) $(PACKAGE_REBUILD_LIST)))
+ifeq ($(spec_names_build_or_rebuild),)
+$(error Must specify packages to build through PACKAGE_BUILD_LIST and PACKAGE_REBUILD_LIST when FAST_BUILD=y)
+endif
+local_specs = $(wildcard $(addprefix $(SPECS_DIR)/*/,$(addsuffix .spec,$(spec_names_build_or_rebuild))))
+$(fast_build_pack_list): $(depend_PACKAGE_BUILD_LIST) $(depend_PACKAGE_REBUILD_LIST) $(depend(PACKAGE_IGNORE_LIST)
```
PawelWMS, line 28: Typo?
```suggestion
local_specs = $(wildcard $(addprefix $(SPECS_DIR)/*/,$(addsuffix .spec,$(spec_names_build_or_rebuild))))
$(fast_build_pack_list): $(depend_PACKAGE_BUILD_LIST) $(depend_PACKAGE_REBUILD_LIST) $(depend_PACKAGE_IGNORE_LIST)
```

oliviacrain: Yep, thanks!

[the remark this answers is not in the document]
oliviacrain: Ah, very valid feedback. In that case, will explore either (a) fixing the other issue or (b) making this functionality use a new flag.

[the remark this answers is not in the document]
oliviacrain: After looking at the scheduler code concerning such a change- will go with adding a new flag. Best to not overload semantics, anyways.

dmcilvaney:
Something you may run into is a bug with the Goal node functions. Currently if you pass a list of packages to build, its meant to add a goal node to each of them. If `strict` mode is set for the function it should error out if it can't find any. Currently it will only error our if the LAST package is missed. Just a heads up if you start scratching your head on why its behaving oddly. Its been on my todo list to fix for a while.

toolkit/docs/building/building.md
```diff
@@ -516,26 +519,36 @@ sudo make hydrate-rpms PACKAGE_ARCHIVE=./rpms.tar.gz
...
 
 ##### `DELTA_BUILD=`**`n`** *(default)*
 
 > Normal build.
 
+#### `FAST_BUILD=...`
+
+##### `FAST_BUILD=`**`y`**
+
+> Only pack a subset of spec files under `$(SPECS_DIR)`. The subset to pack is the intersection of `$(PACKAGE_BUILD_LIST)` and `$(PACKAGE_REBUILD_LIST)`, with specs listed in `$(PACKAGE_IGNORE_LIST)` ignored. This drastically speeds up the build by (a) skipping packing/expanding most SRPMs and (b) not considering as large of a package graph during the package build phase. This type of build is not a substitute for a full build for most changes, but is appropriate for fast development iteration.
```
dmcilvaney, line 546: Should this be 'union' rather than 'intersection'?
```suggestion
> Only pack a subset of spec files under `$(SPECS_DIR)`. The subset to pack is the union of `$(PACKAGE_BUILD_LIST)` and `$(PACKAGE_REBUILD_LIST)`, with specs listed in `$(PACKAGE_IGNORE_LIST)` ignored. This drastically speeds up the build by (a) skipping packing/expanding most SRPMs and (b) not considering as large of a package graph during the package build phase. This type of build is not a substitute for a full build for most changes, but is appropriate for fast development iteration.
```

oliviacrain: You're absolutely correct, thanks!

toolkit/docs/building/building.md
```diff
@@ -636,6 +635,7 @@ To reproduce an ISO build, run the same make invocation as before, but set:
 | UNATTENDED_INSTALLER          |                                                                                                        | Create unattended ISO installer if set. Overrides all other installer options.
 | PACKAGE_BUILD_LIST            |                                                                                                        | Additional packages to build. The package will be skipped if the build system thinks it is already up-to-date.
 | PACKAGE_REBUILD_LIST          |                                                                                                        | Always rebuild this package, even if it is up-to-date. Base package name, will match all virtual packages produced as well.
+| SRPM_PACK_LIST                |                                                                                                        | List of spec basenames to build into SRPMs. If empty, all specs under `$(SPECS_DIR)` will be packed.
```
jslobodzian, line 638: Looks good...just want to double check on how well this interplays with the DELTA_BUILD or HYDRATED_BUILD option.  (Ok to say they are mutually exclusive, just make sure to add some kind of check and fail if someone tries to combine them)

toolkit/scripts/srpm_pack.mk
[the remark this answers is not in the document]
oliviacrain: Changed to use a new make variable `$SRPM_PACK_LIST`.

toolkit/scripts/srpm_pack.mk
```diff
@@ -14,12 +14,22 @@
 SRPM_FILE_SIGNATURE_HANDLING ?= enforce
 
 SRPM_BUILD_CHROOT_DIR = $(BUILD_DIR)/SRPM_packaging
+SRPM_BUILD_LOGS_DIR = $(LOGS_DIR)/pkggen/srpms
 
+toolchain_spec_list = $(toolchain_build_dir)/toolchain_specs.txt
+srpm_pack_list_file = $(BUILD_SRPMS_DIR)/pack_list.txt
+
+ifeq ($(strip $(SRPM_PACK_LIST)),)
```
PawelWMS, line 22: Shouldn't this be `ifneq`? It looks like we'd hit this branch when `SRPM_PACK_LIST` is empty. Alternatively, we can switch the contents of each branch.

toolkit/scripts/srpm_pack.mk
```diff
@@ -14,12 +14,22 @@
 SRPM_FILE_SIGNATURE_HANDLING ?= enforce
 
 SRPM_BUILD_CHROOT_DIR = $(BUILD_DIR)/SRPM_packaging
+SRPM_BUILD_LOGS_DIR = $(LOGS_DIR)/pkggen/srpms
 
+toolchain_spec_list = $(toolchain_build_dir)/toolchain_specs.txt
```
PawelWMS, line 19: Is this still used?
```suggestion
```

toolkit/docs/building/building.md
oliviacrain: **Quick summary of how SRPM_PACK_LIST will interplay technically:**

If running with `DELTA_BUILD=y`, packages will use upstream repos as a cache to resolve nodes in the package graph when possible. By using `SRPM_PACK_LIST` to limit which SRPMs are built, the set of RPMs that need to be downloaded/marked as cached by DELTA_BUILD=y is restricted to what is required by that subset of SRPMs.

If running with `HYDRATED_BUILD=y`, each package is built in its own little microcosm using cached packages, replacing all Run nodes with Prebuilt nodes. By using `SRPM_PACK_LIST`, we reduce the set of SRPMs that need to be processed, and thus we also reduce the number of Run nodes that need to be marked as cached.

**Quick summary of how SRPM_PACK_LIST will interplay semantically with how we use them in the pipelines:**

I don't think `DELTA_BUILD=y` and `SRPM_PACK_LIST` make sense to use together. The point of a delta build is to detect packages that have changed and are unavailable upstream, then rebuild those packages. If we already know what has changed, then why are we using `DELTA_BUILD=y`? It's a bit redundant in purpose.

I also don't think that `HYDRATED_BUILD=y` makes sense to use with `SRPM_PACK_LIST`. If we want to specify that we're only running certain tests, we already have the option to use `PACKAGE_(RE)BUILD_LIST` to do so.

**Now what?**

So, technically, we're fine to use SRPM_PACK_LIST in conjunction with those two flags. In terms of using the delta/hydrated builds for their intended purpose, however, it doesn't make sense to use the flags in this way. Despite this, I don't think we should have these flags be mutually exclusive with the packing list. 

I think it's important to remember the use case for `SRPM_PACK_LIST`. First and foremost, this is a way to get faster local validation builds for our devs by (a) cutting down the number of SRPMs to build and (b) cutting down the size of the package graph that's considered. We don't really have much use for `SRPM_PACK_LIST` in our official pipelines, since we have blazing fast connections to PMC and our source blobstore there. If a dev has a valid reason to do a delta/hydrated build on their local machine, and they also want to save time by using `SRPM_PACK_LIST`, then I don't think we should prevent that. I don't foresee harm/confusion arising from these flags being used together.

toolkit/scripts/srpm_pack.mk
oliviacrain: Yes, it's used when packing the toolchain specs outside of a chroot environment- see `build_toolchain_srpms.flag` target when `DOWNLOAD_SRPMS=n`. I just moved it around to be with the other pack list.

toolkit/scripts/srpm_pack.mk
oliviacrain: Yes, thanks, fixed

PawelWMS (approved):

PR #3575, opened 2022-08-19
Update cert-manager to v1.7.3.

cwize1:
- [x] Ready to merge

###### Summary

Update cert-manager to v1.7.3.

###### Change Log

cert-manager:
- Update cert-manager to v1.7.3.
- Split cert-manager binaries into separate packages.
- Remove cert-manager build dependency on Bazel and just build the
binaries directly using `go build`. This makes building easier. Also,
the latest upstream version of cert-manager does this.
- Use the Go "vendor" directory for Go dependencies instead of dumping
files in the global Go cache.

No

###### Test Methodology

Local build


SPECS/cert-manager/cert-manager.spec
```diff
@@ -1,96 +1,117 @@
...
 # We're using pre-populated GO dependencies from this tarball, since network is disabled during build time.
 #   1. wget https://github.com/jetstack/%{name}/archive/refs/tags/v%{version}.tar.gz -o %%{name}-%%{version}.tar.gz
 #   2. tar -xf %%{name}-%%{version}.tar.gz
 #   3. cd %%{name}-%%{version}
-#   4. go mod download
-#   5. cd $HOME
+#   4. go mod vendor
 #   5. tar  --sort=name \
-#           --mtime="2021-04-26 00:00Z" \
+#           --mtime="2022-08-10 00:00Z" \
```
PawelWMS, line 18: Let's not change the `mtime`. This is just some constant date we use everywhere else for the sake of getting a reproducible tarball. Keeping the date always in line with the date the tarball was introduced is probably not worth the hustle and the potential issues we'd hit, if at some point we'd like to get a tarball for an earlier version of the package (who knows what date one might have used at that point?).

SPECS/cert-manager/cert-manager.spec
```diff
@@ -1,96 +1,117 @@
...
+Summary: cert-manager's webhook binary
+
+%description webhook
+Webhook component providing API validation, mutation and conversion functionality for cert-manager.
+
 %prep
 %autosetup -p1
 %setup -q -T -D -a 1
 
-%build
```
PawelWMS, line 53: Let's keep building inside the `%build` section instead of moving things to `%prep`. We don't really expect too much to happen there.

cwize1: Good catch.

SPECS/cert-manager/cert-manager.spec
cwize1: Fair.

SPECS/cert-manager/cert-manager.spec
```diff
@@ -1,96 +1,117 @@
...
+
+%files cainjector
+%license LICENSE
+%license LICENSES
+%doc README.md
+%{_bindir}/cainjector
+
+%files controller
 %license LICENSE
+%license LICENSES
```
PawelWMS, line 94: We tend to use just one `%license` macro:

```suggestion
%license LICENSE LICENSES
```

SPECS/cert-manager/cert-manager.spec
```diff
@@ -1,96 +1,117 @@
...
+%files webhook
+%license LICENSE
+%license LICENSES
+%doc README.md
+%{_bindir}/webhook
 
 %changelog
+* Fri Aug 05 2022 Chris Gunn <chrisgun@microsoft.com> - 1.7.3-1
+- Update to v1.7.3
+- Split binaries into separate packages.
```
PawelWMS, line 113: That's great but for backward compatibility, let's keep the default package installing all of the binaries. This can be done with an empty `%files` section and a bunch of `Requires: %{name}-[subpackage_name]` under the `BuildRequires`.

PawelWMS (changes requested):
Looks good, just a few small comments.

Also, I think you accidentally removed `golang-1.17`.

cwize1:
> Looks good, just a few small comments.
> 
> Also, I think you accidentally removed `golang-1.17`.

RE: golang-1.17, as best as I could tell that only was being kept around for the sake of cert-manager. Should it be kept regardless?

SPECS/cert-manager/cert-manager.spec
cwize1: Fixed.

SPECS/cert-manager/cert-manager.spec
cwize1: Fixed.

SPECS/cert-manager/cert-manager.spec
cwize1: Fixed.

PawelWMS (approved):
Changes look good.

Fair point regarding the older version of Golang, I haven't noticed it being only required by that one package. I think it's safe to remove but I'm also fine, if you merge the PR as it is now.

SPECS/cert-manager/cert-manager.spec
```diff
@@ -1,96 +1,120 @@
...
-git add .
-GIT_AUTHOR_DATE=2000-01-01T01:01:01 GIT_COMMITTER_DATE=2000-01-01T01:01:01 \
-git commit -m "Dummy commit just to satisfy bazel" &> /dev/null
-
-for cmd in cmd/* ; do
-  if [ "$cmd" != cmd/util ]; then
-    bazel --batch build --repository_cache=BAZEL_CACHE //$cmd
-  fi
-done
+go build -o bin/acmesolver cmd/acmesolver/main.go
```
mfrw, line 72: Love the clean-up done here :)

mfrw (approved):

cwize1:
@PawelWMS Ready to merge.

oliviacrain (approved):

PR #3606, opened 2022-08-24
fix cloud-init dependency issue

liunan-ms:
###### Merge Checklist  

- [X] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [ ] Ready to merge

---

###### Summary 

This PR is to fix the cloud-init issue which makes Penguinator 2.0 smoke tests fail. This issue is caused by cloud-init using the default dependencies that leads to a dependency cycle, so this PR set `DefaultDependencied=no` and enabled custom dependencies for cloud-init in mariner 2.0 VMs.

https://microsoft.visualstudio.com/OS/_workitems/edit/40987945?src=WorkItemMention&src-action=artifact_link

###### Change Log  

- cloud-init

**NO**

###### Test Methodology

- Local build

### DIFF
--- a/SPECS/cloud-init/add-mariner-distro-support.patch
+++ b/SPECS/cloud-init/add-mariner-distro-support.patch
@@ -359,9 +359,33 @@ index 6951a0e3..411065f9 100644
     # Other config here will be given to the distro class and/or path classes
     paths:
        cloud_dir: /var/lib/cloud/
+diff -ruN a/systemd/cloud-init-local.service.tmpl b/systemd/cloud-init-local.service.tmpl
+--- a/systemd/cloud-init-local.service.tmpl     2022-05-18 08:23:24.000000000 -0700
++++ b/systemd/cloud-init-local.service.tmpl     2022-08-23 16:23:57.854218334 -0700
+@@ -1,9 +1,7 @@
+ ## template:jinja
+ [Unit]
+ Description=Initial cloud-init job (pre-networking)
+-{% if variant in ["ubuntu", "unknown", "debian", "rhel" ] %}
+ DefaultDependencies=no
+-{% endif %}
+ Wants=network-pre.target
+ After=hv_kvp_daemon.service
+ After=systemd-remount-fs.service
+@@ -21,10 +19,8 @@
+ Before=firewalld.target
+ Conflicts=shutdown.target
+ {% endif %}
+-{% if variant in ["ubuntu", "unknown", "debian"] %}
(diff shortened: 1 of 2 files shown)


nicogbg (approved):

anphel31 (approved):

SPECS/cloud-init/add-mariner-distro-support.patch
```diff
@@ -359,9 +359,33 @@ index 6951a0e3..411065f9 100644
...
     paths:
        cloud_dir: /var/lib/cloud/
+diff -ruN a/systemd/cloud-init-local.service.tmpl b/systemd/cloud-init-local.service.tmpl
+--- a/systemd/cloud-init-local.service.tmpl     2022-05-18 08:23:24.000000000 -0700
++++ b/systemd/cloud-init-local.service.tmpl     2022-08-23 16:23:57.854218334 -0700
+@@ -1,9 +1,7 @@
+ ## template:jinja
+ [Unit]
+ Description=Initial cloud-init job (pre-networking)
+-{% if variant in ["ubuntu", "unknown", "debian", "rhel" ] %}
```
PawelWMS, line 369: Which variant are we? The "unknown" one? I think we might not even need any of these checks, since we're building only for Mariner anyway.

liunan-ms: Yes, we are not any one of these variants, and we tried to add "mariner" in the list but it is not recognized so that this if condition is always skipped. So we removed this variant check here.

cjp256:
Looking at the latest published mariner2, this should already be the case?

cpatterson@test-mariner2-x1 [ ~ ]$ cat /lib/systemd/system/cloud-init.service
[Unit]
Description=Initial cloud-init job (metadata service crawler)
DefaultDependencies=no
...

The "mariner" variant should work fine, it is quite extensively used in this patch.

liunan-ms:
> 
`DefaultDependencies=no` is within this if catch : https://github.com/microsoft/CBL-Mariner/blob/50e2132e5a911b049c1f3dd83786c8f3d2a58887/SPECS/cloud-init/add-mariner-distro-support.patch#L370
which means the variant is not one of ["mariner", "photon", "rhel"]. We also tried to add "mariner" into other if checks, and it ends up that if conditions are not satisfied when running Mariner.

PR #3615, opened 2022-08-25
Add Multiple Disk Support for Image Build

josephknierman:
This PR adds support for creating multiple disks for a specific image config.  The PR adds some optional configs for adding IDs to Disks and allows setting an optional primary disk in the SystemConfiguration, which is used for determining  where to install the bootloader. The PR assumes that all disks should be using the same target type. This PR is intended to have all single disk configurations work without any additional changes. 

###### Merge Checklist  

- [ ] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -321,79 +321,81 @@ func DetachLoopbackDevice(diskDevPath string) (err error) {
 }
 
 // CreatePartitions creates partitions on the specified disk according to the disk config
-func CreatePartitions(diskDevPath string, disk configuration.Disk, rootEncryption configuration.RootEncryption, readOnlyRootConfig configuration.ReadOnlyVerityRoot) (partDevPathMap map[string]string, partIDToFsTypeMap map[string]string, encryptedRoot EncryptedRootDevice, readOnlyRoot VerityDevice, err error) {
+func CreatePartitions(diskDevPaths []string, disks []configuration.Disk, rootEncryption configuration.RootEncryption, readOnlyRootConfig configuration.ReadOnlyVerityRoot) (partDevPathMap map[string]string, partIDToFsTypeMap map[string]string, encryptedRoot EncryptedRootDevice, readOnlyRoot VerityDevice, err error) {
 	const timeoutInSeconds = "5"
 	partDevPathMap = make(map[string]string)
 	partIDToFsTypeMap = make(map[string]string)
 
-	// Clear any old partition table info to prevent errors during partition creation
-	_, stderr, err := shell.Execute("sfdisk", "--delete", diskDevPath)
-	if err != nil {
-		logger.Log.Warnf("Failed to clear partition table. Expected if the disk is blank: %v", stderr)
-	}
-
-	// Create new partition table
-	partitionTableType := disk.PartitionTableType
-	logger.Log.Debugf("Converting partition table type (%v) to parted argument", partitionTableType)
-	partedArgument, err := partitionTableType.ConvertToPartedArgument()
-	if err != nil {
-		logger.Log.Errorf("Unable to convert partition table type (%v) to parted argument", partitionTableType)
-		return
-	}
-	_, stderr, err = shell.Execute("flock", "--timeout", timeoutInSeconds, diskDevPath, "parted", diskDevPath, "--script", "mklabel", partedArgument)
-	if err != nil {
-		logger.Log.Warnf("Failed to set partition table type using parted: %v", stderr)
-		return
-	}
-
-	usingExtendedPartition := (len(disk.Partitions) > maxPrimaryPartitionsForMBR) && (partitionTableType == configuration.PartitionTableTypeMbr)
-
-	// Partitions assumed to be defined in sorted order
-	for idx, partition := range disk.Partitions {
-		partType, partitionNumber := obtainPartitionDetail(idx, usingExtendedPartition)
-		// Insert an extended partition
-		if partType == extendedPartitionType {
-			err = createExtendedPartition(diskDevPath, partitionTableType.String(), disk.Partitions, partIDToFsTypeMap, partDevPathMap)
-			if err != nil {
-				return
-			}
-
-			// Update partType and partitionNumber
-			partType = logicalPartitionType
-			partitionNumber = partitionNumber + 1
+	for i, diskDevPath := range diskDevPaths {
+		// Clear any old partition table info to prevent errors during partition creation
+		_, stderr, err := shell.Execute("sfdisk", "--delete", diskDevPath)
+		if err != nil {
+			logger.Log.Warnf("Failed to clear partition table. Expected if the disk is blank: %v", stderr)
 		}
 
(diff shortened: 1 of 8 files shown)


toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -321,79 +321,81 @@ func DetachLoopbackDevice(diskDevPath string) (err error) {
...
+				logger.Log.Warnf("Failed to format partition")
 				return partDevPathMap, partIDToFsTypeMap, encryptedRoot, readOnlyRoot, err
 			}
-			partDevPathMap[partition.ID] = readOnlyRoot.MappedDevice
-		} else {
-			partDevPathMap[partition.ID] = partDevPath
-		}
 
-		partIDToFsTypeMap[partition.ID] = partFsType
+			if rootEncryption.Enable && partition.HasFlag(configuration.PartitionFlagDeviceMapperRoot) {
```
dmcilvaney, line 379: Could you double check in the validator function `checkDeviceMapperFlags()` that only one disk sets this for a given systemconfig? (I'm fairly sure it works as is)

toolkit/tools/imager/imager.go
```diff
@@ -328,50 +335,81 @@ func setupRootFS(outputDir, installRoot string) (extraMountPoints []*safechroot.
...
+			return
+		}
+
+	}
+
+	if DiskType == realDiskType {
 		if liveInstallFlag {
-			diskDevPath = diskConfig.TargetDisk.Value
-			partIDToDevPathMap, partIDToFsTypeMap, encryptedRoot, readOnlyRoot, err = setupRealDisk(diskDevPath, diskConfig, rootEncryption, readOnlyRootConfig)
+			partIDToDevPathMap, partIDToFsTypeMap, encryptedRoot, readOnlyRoot, err = setupRealDisks(diskNames, disks, rootEncryption, readOnlyRootConfig)
```
dmcilvaney, line 362: Does `diskNames` contain the actual device path here (i.e. `/dev/sda`)? Might be missing the translation step. If not, a comment/rename of variable might be in order.

toolkit/tools/imager/imager.go
```diff
@@ -328,50 +335,81 @@ func setupRootFS(outputDir, installRoot string) (extraMountPoints []*safechroot.
...
+func setupLoopDeviceDisks(outputDir string, diskNames []string, diskConfigs []configuration.Disk, rootEncryption configuration.RootEncryption, readOnlyRootConfig configuration.ReadOnlyVerityRoot) (diskDevPaths []string, partIDToDevPathMap, partIDToFsTypeMap map[string]string, encryptedRoot diskutils.EncryptedRootDevice, readOnlyRoot diskutils.VerityDevice, err error) {
 	defer func() {
 		// Detach the loopback device on failure
-		if err != nil && diskDevPath != "" {
-			detachErr := diskutils.DetachLoopbackDevice(diskDevPath)
-			if detachErr != nil {
-				logger.Log.Errorf("Failed to detach loopback device on failed initialization. Error: %s", detachErr)
+		if err != nil && len(diskDevPaths) != 0 {
+			for _, diskDevPath := range diskDevPaths {
+				detachErr := diskutils.DetachLoopbackDevice(diskDevPath)
```
dmcilvaney, line 379: Any chance we could have a list of paths that are `""` here? Might be worth still checking that inside the loop.

toolkit/tools/imager/imager.go
```diff
@@ -540,8 +593,13 @@ func buildImage(mountPointMap, mountPointToFsTypeMap, mountPointToMountArgsMap,
 	}
 
 	// Only configure the bootloader or read only partitions for actual disks, a rootfs does not need these
+	// TODO add extrat config to specify bootloader to use
```
dmcilvaney, line 596: Is this still needed, or does the primary disk conver it?

toolkit/tools/imagegen/diskutils/diskutils.go
josephknierman: after looking at checkDeviceMapperFla() it seems like only one disk would be selected for a given systemconfig. Based off the functions the disk that gets selected would be the first in the list that has the root partition "/"

toolkit/tools/imager/imager.go
josephknierman: need to removed that comment. This would be resolved by the PrimaryDisk specified in systemconfig which tells us which disk to install the legacy bootloader on.

toolkit/tools/imagegen/diskutils/diskutils.go
dmcilvaney: Ok, that was my reading as well.

toolkit/tools/imager/imager.go
josephknierman: I dont think it is possible for the diskdevpaths to be "", they will all be the value "{some_path}/disk#.raw" which is derived from the DefaultTempDiskNames which will always be defined. I am not super familiar with golang so would diskdevpaths be modified before running the defer functioned? If so then i should make the change since some other function at some point could add "" to the list. I think If diskdevpaths contained a "" then that would be caught when calling SetupLoopbackDevice()

toolkit/tools/imager/imager.go
josephknierman: diskNames contains a list of names for disks that we need to create "disk#.raw", these are initally placed in defaultTempDiskNames. Would something like rawFileNames be better here?

dmcilvaney: I believe this path is used only for ISO installers etc., where it's meant to target an actual physical disk (ie diskDevPath, eg /dev/sda).

dmcilvaney: Storing them in a lookup could work, although isn't the path part of the disk object...? Might be better to refactor and just pull from there unless I'm missing something (`diskDevPath = diskConfig.TargetDisk.Value`)

toolkit/tools/imager/imager.go
dmcilvaney: Yes, parameters to a defer call are passed at the time the call is executed (mutable). It seems likely that "" won't get added, so probably fine.

toolkit/tools/imager/imager.go
josephknierman: Ah i see what you are saying. i deleted the translation step there and did not add back an equivalent. in the next commit I will have a refactor that gets the correct value for diskdevpaths for real disks not using loopback

PawelWMS (changes requested):
A couple of things:
- The change needs to be rebased to a development branch. At this point, I'd consider migrating this change to `3.0-dev`
- Since this is an image generation change, I'd consider moving it into the `imagecustomizer` tool, which is a more up-to-date tool for customizing images.

PR #3650, opened 2022-08-31
set crash kernel memory based on host memory

Redent0r:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Set crashkernel param in kernel.spec based on host memory. This makes kernel crash recovery work consistently. Using less memory than needed causes kdump to take longer to create the vmcore file (memory dump)

Based on a few sources ([1](https://www.suse.com/support/kb/doc/?id=000016171), [2](https://www.golinuxhub.com/2018/08/how-to-configure-and-install-kdump-rhel7-crashkernel/)), the memory allocated to crash kernel ought to be based on the total memory available to the host. There's a crashkernel=auto option that tries to do this and I think ideally, we would like to use this. However, on a 2gb ram hyper vm machine, the auto option will still assign 128mb to the crash kernel, and this will cause kdump issues. So, we'd want to do something similar to the auto option, but with higher memory allocated to crash kernel. The range I use is based on a [recommended config for RHEL6.0 and RHEL6.1](https://access.redhat.com/solutions/59432) where I slightly modify the lower range of the values (2gb-6gb -> 1gb-6gb) to be able to get consistent kernel recoveries on my hyper vm instance(128mb wasn't enough for 1.8gb of ram available) Using this range, I was able to recover consistently with 6gb and 9gb of ram as well, showing this range works for higher ram values. 

Follow up: 
Why does mariner crashkernel requires more than 128mb ([default recommended by canonical](https://ubuntu.com/server/docs/kernel-crash-dump) and assigned by auto option)for kernel crashes to recover consistently

Do higher ram values really require a bigger crashkernel, maybe we can use crashkernel=256mb(or lower) for higher values as well. 
A: Further testing on hyper v suggest we could stick to 256m for everything. I was able to crash and recover logs fast and consecutively using 256m on a vm with 10gb of ram

###### Change Log  

- Updated crashkernel param

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/40870902?src=WorkItemMention&src-action=artifact_link

###### Test Methodology

- tested on hypervm
- used `echo c > /proc/sysrq-trigger` to trigger a kernel crash and noticed inconsistent results. (Sometimes it would recover after a few minutes, sometimes it would hang and not recover)
- modified crashkernel param to allocate more memory
- Verified that kernel crash recovery happens fast and consistently now (about 15 seconds to get back to login prompt) and vmcore file gets created every time
-  Full local image build succeeded
-  AMD image build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=231743&view=results

### DIFF
--- a/SPECS-SIGNED/kernel-signed/kernel-signed.spec
+++ b/SPECS-SIGNED/kernel-signed/kernel-signed.spec
@@ -10,7 +10,7 @@
 Summary:        Signed Linux Kernel for %{buildarch} systems
 Name:           kernel-signed-%{buildarch}
 Version:        5.15.57.1
-Release:        3%{?dist}
+Release:        4%{?dist}
 License:        GPLv2
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -153,6 +153,9 @@ ln -sf linux-%{uname_r}.cfg /boot/mariner.cfg
 %exclude /module_info.ld
 
 %changelog
+* Wed Aug 31 2022 Saul Paredes <saulparedes@microsoft.com> - 5.15.57.1-4
+- Bump release number to match kernel release
+
 * Tue Aug 02 2022 Rachel Menge <rachelmenge@microsoft.com> - 5.15.57.1-3
 - Bump release number to match kernel release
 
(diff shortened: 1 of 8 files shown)


Redent0r:
moved to https://github.com/microsoft/CBL-Mariner/pull/3705

PR #3705, opened 2022-09-07
adjust crashkernel param based on available ram

Redent0r:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Set crashkernel param in kernel.spec based on host memory. This makes kernel crash recovery work consistently. Using less memory than needed causes kdump to take longer to create the vmcore file (memory dump)

Based on a few sources ([1](https://www.suse.com/support/kb/doc/?id=000016171), [2](https://www.golinuxhub.com/2018/08/how-to-configure-and-install-kdump-rhel7-crashkernel/)), the memory allocated to crash kernel ought to be based on the total memory available to the host. There's a crashkernel=auto option that tries to do this and I think ideally, we would like to use this. However, on a 2gb ram hyper vm machine, the auto option will still assign 128mb to the crash kernel, and this will cause kdump issues. So, we'd want to do something similar to the auto option, but with higher memory allocated to crash kernel. The range I use is based on a [recommended config for RHEL6.0 and RHEL6.1](https://access.redhat.com/solutions/59432) where I slightly modify the lower range of the values (2gb-6gb -> 1gb-6gb) to be able to get consistent kernel recoveries on my hyper vm instance(128mb wasn't enough for 1.8gb of ram available) Using this range, I was able to recover consistently with 6gb and 9gb of ram as well, showing this range works for higher ram values. 

Follow up: 
Why does mariner crashkernel requires more than 128mb ([default recommended by canonical](https://ubuntu.com/server/docs/kernel-crash-dump) and assigned by auto option)for kernel crashes to recover consistently

Do higher ram values really require a bigger crashkernel, maybe we can use crashkernel=256mb(or lower) for higher values as well. 
A: Further testing on hyper v suggest we could stick to 256m for everything. I was able to crash and recover logs fast and consecutively using 256m on a vm with 10gb of ram

###### Change Log  

- Updated crashkernel param

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/40870902?src=WorkItemMention&src-action=artifact_link

###### Links to CVEs  

###### Test Methodology

- tested on hypervm
- used `echo c > /proc/sysrq-trigger` to trigger a kernel crash and noticed inconsistent results. (Sometimes it would recover after a few minutes, sometimes it would hang and not recover)
- modified crashkernel param to allocate more memory
- Verified that kernel crash recovery happens fast and consistently now (about 15 seconds to get back to login prompt) and vmcore file gets created every time
-  Full local image build succeeded
-  AMD toolchain build: in progress https://dev.azure.com/mariner-org/mariner/_build/results?buildId=233685&view=logs&j=f3215316-5135-57cb-291b-0903c8ad86f7
- ARM toolchain build: in progress https://dev.azure.com/mariner-org/mariner/_build/results?buildId=233686&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e


Redent0r:
This is my first pr, so **any** type of feedback is welcomed :)

neha170:
Hi, was this tested for both x86 and ARM images? The bug mentions only x86 image, so I wasn't sure if you tested for both. Thanks

Redent0r:
> Hi, was this tested for both x86 and ARM images? The bug mentions only x86 image, so I wasn't sure if you tested for both. Thanks

I only tested on a x86 image on hyper v. I'll do some testing in ARM as well and add results to the pr. Thanks!

neha170:
Thanks, the ARM experiment result will be nice to have, but if it doesn't work out, there might be more fixes required in ARM unrelated to your bug. I would also suggest trying this on an x86 Azure VM. Thanks

Redent0r:
Results contrasting arm and x86 so far. 256M seems to fix crashkernel issue on x86. In arm, it seems to make it better, but not fully functional: vmcore file gets marked as incomplete and kdumpctl is locked by another process after recovering, unlike in x86. Testing with bigger crashkernel params yielded same result
Screenshot 2022-09-12 114134
Screenshot 2022-09-12 114809

Redent0r:
updated test matrix. Based on this, I would suggest we allocate 256mb to crashkernel, regardless of the amount of ram available. Reason being that 256 seems to work well across differet architectures and sizes of machines; no need to allocate more memory than we need to
Screenshot 2022-09-13 114839

toolkit/resources/manifests/package/pkggen_core_aarch64.txt
```diff
@@ -1,5 +1,5 @@
 filesystem-1.1-11.cm2.aarch64.rpm
-kernel-headers-5.15.63.1-2.cm2.noarch.rpm
```
dmcilvaney, line 2: Also make these updates to the full toolchain manifests (the pkggen core is a subset of the full toolchain list)

dmcilvaney (approved):

toolkit/resources/manifests/package/pkggen_core_aarch64.txt
Redent0r: Added. Thank you

PR #3733, opened 2022-09-10
kernel: enable CONFIG_VFAT_FS 

aditjha-msft:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

- Setting vfat module in kernel config to Y
- It was noticed that the kernel config was inherited from photon and other distros like ubuntu have vfat compiled in, and our aarch64 config does as well. Vfat is pretty widely used and will hopefully this change benefit more customers on top of the one that is already asking for it.

###### Change Log  

- Setting vfat module in kernel config to Y

**YES**

###### Test Methodology

- Pipeline build id: 234931
- On VHD image of Mariner 2.0, `modprobe --first-time vfat` now shows an error because vfat is built into the kernel, prior to fix it would be able to load vfat module as it was available but not loaded.


dmcilvaney:
Could you expand on the rationale behind the change so we can understand why it was done in the future? ie, if we go to turn this back to a module in a year or two, it will help us remember why we decided to set it to a built-in instead of a module.

SPECS/kernel/config
```diff
@@ -6517,9 +6517,9 @@ CONFIG_UDF_FS=y
 #
 # DOS/FAT/EXFAT/NT Filesystems
 #
-CONFIG_FAT_FS=m
+CONFIG_FAT_FS=y
 CONFIG_MSDOS_FS=m
-CONFIG_VFAT_FS=m
+CONFIG_VFAT_FS=y
```
aadhar-agarwal, line 6522: What is the difference between `m` and `y` and what does CONFIG_FAT_FS mean exactly?

aditjha-msft: The difference between `m` and `y` is that the first makes sure the module is _available_ to be installed/uninstalled into the kernel, but the latter makes sure the module is _always_ installed into the kernel from boot up.

aditjha-msft:
> Could you expand on the rationale behind the change so we can understand why it was done in the future? ie, if we go to turn this back to a module in a year or two, it will help us remember why we decided to set it to a built-in instead of a module.

It was noticed that the config was inherited from photon and other linux distros like ubuntu have it compiled in, as well as our aarch64 config. Also vfat is pretty widely used and will hopefully benefit more customers on top of the one that is already asking for it.

dmcilvaney:
> 

Could you put something along those lines in the PR description? That is generally where people will look when trying to understand what was done.

christopherco (approved):

MaxBrodeurUrbas (approved):
🥇

PR #2983, opened 2022-05-11
[RFC] tools: refactor all binary packages as library packages

mfrw:
Request for Comments

Closes: #2982

###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [ ] Ready to merge

---

###### Summary 

What does the PR accomplish, why was it needed?
In the current form, all the code in the binary packages is not re-useable to enable building on top of the great work we have.
This PR abstracts all the binary packages as library packages which can be imported in other packages as well to enable code-reuse.
The changes in this PR are 100% backward compatible and it does not break anything. To achieve this, the path of least resistance seemed
to expose a `Config` structure from the library package and pass all the relevant details that are passed as `flags`.

###### Change Log  

- tools: srpmpacker: refactor as a lib instead of a main package
- tools: specreader: refactor as a lib instead of a main package
- tools: pkgworker: refactor as a lib instead of a main package
- tools: graphpkgfetcher: refactor as a lib instead of main package
- tools: grapher: refactor as a lib instead of main package
- tools: graphpreprocessor: refactor as a lib instead of main package
- tools: mv imagegen -> pkg/imagegen
- tools: imageconfigvalidator: refactor as a lib instead of main package
- tools: imagepkgfetcher: refactor as a lib instead of main package
- tools: imager: refactor as a lib instead of main package
[...]

PR #3734, opened 2022-09-10
toolkit: added RPMs snapshots.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Extending the toolkit archive to contain a snapshot of all RPMs, which could have been built at the time the archive was created. This is part of a larger change to enable building specs with an explicitly defined set of RPMs instead of the latest ones available on PMC.

###### Change Log  

- Extended toolkit archive to contain an RPMs snapshot.

No.

###### Test Methodology

- Local toolkit build.
- Pipeline toolkit build: 236165


toolkit/scripts/toolkit.mk
```diff
@@ -30,7 +32,7 @@ clean-package-toolkit:
 	rm -f $(toolkit_remove_archive)
 	rm -rf $(toolkit_build_dir)
 
-package-toolkit: go-tools
+package-toolkit: go-tools $(toolkit_rpms_snapshot)
```
dmcilvaney, line 35: Not sure we want to be super pedantic here with the dependencies, but take a look at `$(local_specs)` in srpm_pack.mk. It adds a dependency on every spec file itself. The directory dependency will track adding/removing folders but not changes to files inside those folders.

toolkit/scripts/rpms_snapshot_chroot.sh
```diff
@@ -0,0 +1,73 @@
...
+    if rpmspec -q -D "dist $DIST_TAG" -D "with_check 1" --queryformat="[%{EXCLUSIVEARCH} ]\n" --srpm "$spec" 2>/dev/null | grep -qP "\b$CURRENT_ARCH\b"
+    then
+        echo "$spec" >> "$TEMP_FILE"
+    fi
+done
+
+rm -f "$OUTPUT"
+while read -r spec
+do
+    rpmspec -q -D "_sourcedir $(dirname "$spec")" -D "dist $DIST_TAG" -D "with_check 1" --builtrpms "$spec" >> "$OUTPUT" 2>/dev/null
```
oliviacrain, line 70: We won't catch any auto-generated subpackages this way. Today this is mostly just applicable to `debuginfo` subpackages, but this will probably expand to other language ecosystems as their tooling becomes increasingly more automated. This PR looks good to me, as long as we're fine with this tradeoff for now.

oliviacrain: A comment about this tradeoff somewhere would be nice, though!

toolkit/scripts/toolkit.mk
PawelWMS: Oh, we (I) want to be as pedantic as humanly possible, so thank you for pointing this out. I've updated the targets quite a bit since this comment to make take into account all possible dependencies. I've only skipped `DIST_TAG` since this one shouldn't be manually modified.

toolkit/scripts/toolkit.mk
```diff
@@ -12,36 +12,90 @@ toolkit_component_extra_files = \
 	$(PROJECT_ROOT)/LICENSES-AND-NOTICES/LICENSE.md \
 	$(toolkit_root)/.gitignore
 
-mariner_repos_dir = $(SPECS_DIR)/mariner-repos
+mariner_repos_dir = $(PROJECT_ROOT)/SPECS/mariner-repos
```
dmcilvaney, line 15: Never really gave it any thought, but should toolkit packaging be expected to work for non-standard layouts? (ie `SPECS_DIR=/path/to/my/SPECS`)

toolkit/scripts/toolkit.mk
```diff
@@ -12,36 +12,90 @@ toolkit_component_extra_files = \
...
-	echo "$(toolkit_version)" > $(toolkit_release_file) && \
-	rm -rf $(toolkit_build_dir)/out && \
-	tar -I $(ARCHIVE_TOOL) -cvp -f $(toolkit_archive) -C $(toolkit_build_dir)/.. $(notdir $(toolkit_build_dir))
+	rm -rf $(toolkit_prep_dir)/out && \
+	tar -cvp -f $(toolkit_archive) -C $(dir $(toolkit_prep_dir)) $(notdir $(toolkit_prep_dir))
+
+rpms-snapshot: $(rpms_snapshot)
+	@echo "RPMs snapshot generated under '$(rpms_snapshot)'."
+
+$(rpms_snapshot): $(rpms_snapshot_per_specs) $(depend_SPECS_DIR)
```
dmcilvaney, line 86: I think this `$(rpms_snapshot):` middle step can be omitted, we could just have `$(rpms_snapshot_per_specs)` be `$(OUT_DIR)/$(rpms_snapshot_name)` from the get-go, right?

toolkit/scripts/toolkit.mk
```diff
@@ -12,36 +12,90 @@ toolkit_component_extra_files = \
...
+	cp $(rpms_snapshot_per_specs) $(rpms_snapshot)
+
+$(rpms_snapshot_per_specs): $(go-rpmssnapshot) $(chroot_worker) $(LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(SPECS_DIR)
```
dmcilvaney, line 89: Adding the chroot worker will slow down the generation of the toolkit. Not sure I can really see a way around it admittedly.

I've wondered more than once if it would be possible to create a super lightweight parsing only worker. (For the future)

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
```diff
@@ -0,0 +1,255 @@
+// Copyright (c) Microsoft Corporation.
```
dmcilvaney, line 1: This shares a lot of functionality with specreader, which is in dire need of an update. Might be a good future todo item to take the stuff here and replace the old specreader code with a common version.

dmcilvaney (approved):

dmcilvaney:
No major issues that I see, I think this should work as intended.

toolkit/scripts/toolkit.mk
PawelWMS: The value of `SPECS_DIR` should have had no relevance to building the toolkit before the changes from this PR. It only needed the repo files and they are always in the same place, so this was just a corner case bug we never hit, because there was no point in setting `SPECS_DIR` to build the toolkit tarball.

With the current changes the snapshot relies on the value of `SPECS_DIR` but we still want to look for the repo files inside the core specs.

toolkit/scripts/toolkit.mk
PawelWMS: I wanted to achieve two things:

1. Rebuilding existing snapshots on `SPECS_DIR` changes and
2. still have only one snapshot file for the current `SPECS_DIR` inside the output folder.
 
It's a minor thing, but to keep things clean, I wanted to make sure that you can always rely on one set output file to contain your results after running the `rpms-snapshot` target. It makes no difference inside out toolkit, since you can always have other targets ask for `rpms_snapshot_per_specs`, but any external dependence on the snapshot's location would have a harder time finding it.

toolkit/scripts/toolkit.mk
PawelWMS: You're right, there are certainly some improvements we could do here. However, building the toolkit tarball end-to-end took ~70 seconds for me, so I figured there's bigger fish to fry.

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
PawelWMS: Right again. I even tried to re-use some of the specreader's code but its chroot setup was slightly different than the one I needed here, so refactoring would have been too robust for this PR.

What's more important, though, is that I'm waiting for #2983 to make it into `main`. I see it as sort of more thorough groundwork enabling us to re-use code more easily. I didn't want to make too many changes to existing code, which would make updating that large rework even more difficult.

PR #3748, opened 2022-09-13
Police `glibc-static` versions

bossmc:
###### Merge Checklist  

- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.  ❤️ 
- [ ] Ready to merge

---

###### Summary 
As per the 4th bullet above it's important that when glibc-static (and other paclkages) are released that other packages that depend on those static artefacts are also rebuilt and released at the same time.

This PR attempts to police this for `glibc-static` in two stages:

* Splitting `glibc-static` out as a real package (rahter than an alias to `glibc-devel`) that contains the `.a` and static runtime files - this means that packages that wish to build statically against glibc will have to name `glibc-static` in their `BuildRequires`.
* Adding a script that checks that every instance of `glibc-static` in a `BuildRequires` has a version bound of `>=` the current version of the `glibc` packages - this means that they can't miss any `glibc` releases.

I'm raising this PR now to kick of discussions around the approach, along with any discussions about how we want to run this script (Github action?  On PR branches or just main/release branches?  Official build pipeline?  Other?).  Obviously if the pattern is deemed a good idea we can also apply it to other similar packages (e.g. `golang` as a `BuildRequire`, or `dietlibc`/`ulibc`) too.

If a specfile fails the test you get output similar to:

```
glibc-static version detected: 2.35-2

Specfile tini (at SPECS/tini/tini.spec):
  * Does not specify the latest `glibc-static` version in its BuildRequires
  Suggestion: Use `glibc-static >= 2.35-2`

====================== Spec verification FAILED ======================

Please update the spec files listed above.
```

###### Change Log
- Rebuild statically linked packages for the `glibc 2.35-2` release

YES - though exactly how is still up for debate

###### Links to CVEs
- https://nvd.nist.gov/vuln/detail/CVE-2022-23218
- https://nvd.nist.gov/vuln/detail/CVE-2022-23219

###### Test Methodology
Local builds

### DIFF
--- a/SPECS-EXTENDED/buildah/buildah.spec
+++ b/SPECS-EXTENDED/buildah/buildah.spec
@@ -40,7 +40,7 @@ BuildRequires: device-mapper-devel
 BuildRequires: golang
 BuildRequires: git
 BuildRequires: glib2-devel
-BuildRequires: glibc-static
+BuildRequires: glibc-static >= 2.35-2
 BuildRequires: go-md2man
 BuildRequires: go-rpm-macros
 BuildRequires: gpgme-devel
--- a/SPECS-EXTENDED/buildah/buildah.spec
+++ b/SPECS-EXTENDED/buildah/buildah.spec
@@ -31,7 +31,7 @@ Distribution:   Mariner
 
 Name: %{repo}
 Version: 1.18.0
-Release: 4%{?dist}
+Release: 5%{?dist}
 Summary: A command line tool used for creating OCI Images
 License: ASL 2.0
 URL: https://%{name}.io
@@ -146,6 +146,9 @@ cp imgtype %{buildroot}/%{_bindir}/%{name}-imgtype
 %{_datadir}/%{name}/test
 
 %changelog
+* Tue Sep 13 2022 Andy Caldwell <andycaldwell@microsoft.com> - 1.18.0-5
+- Rebuilt for glibc-static 2.35-2
+
 * Mon Aug 22 2022 Olivia Crain <oliviacrain@microsoft.com> - 1.18.0-4
 - Bump release to rebuild against Go 1.18.5
 
--- a/SPECS-EXTENDED/catatonit/catatonit.spec
+++ b/SPECS-EXTENDED/catatonit/catatonit.spec
@@ -13,7 +13,7 @@ BuildRequires: automake
 BuildRequires: file
 BuildRequires: gcc
 BuildRequires: git
-BuildRequires: glibc-static
+BuildRequires: glibc-static >= 2.35-2
 BuildRequires: libtool
 BuildRequires: make
 
--- a/SPECS-EXTENDED/catatonit/catatonit.spec
+++ b/SPECS-EXTENDED/catatonit/catatonit.spec
@@ -3,7 +3,7 @@ Distribution:   Mariner
 
 Name: catatonit
 Version: 0.1.7
-Release: 5%{?dist}
+Release: 6%{?dist}
 Summary: A signal-forwarding process manager for containers
 License: GPLv3+
 URL: https://github.com/openSUSE/catatonit
@@ -61,6 +61,9 @@ ln -s %{_libexecdir}/%{name}/%{name} %{buildroot}%{_libexecdir}/podman/%{name}
 %{_libexecdir}/podman/%{name}
 
 %changelog
+* Tue Sep 13 2022 Andy Caldwell <andycaldwell@microsoft.com> - 0.1.7-6
+- Rebuilt for glibc-static 2.35-2
+
 * Fri Mar 18 2022 Pawel Winogrodzki <pawelwi@microsoft.com> - 0.1.7-1
 - Updating to 0.1.7 using Fedora 35 spec (license: MIT) for guidance.
 - License verified.
--- a/SPECS/glibc/glibc.spec
+++ b/SPECS/glibc/glibc.spec
@@ -49,11 +49,18 @@ Summary:        Header files for glibc
 Group:          Applications/System
 Requires:       %{name} = %{version}-%{release}
 Provides:       %{name}-headers = %{version}-%{release}
-Provides:       %{name}-static = %{version}-%{release}
-Provides:       %{name}-static%{?_isa} = %{version}-%{release}
 
 %description devel
 These are the header files of glibc.
+
+%package static
+Summary:        Static glibc library and runtimes
+Group:          Applications/System
+Requires:       %{name}-devel = %{version}-%{release}
+Provides:       %{name}-static%{?_isa} = %{version}-%{release}
+
+%description static
+These are the static artefacts for glibc.
 
 %package lang
 Summary:        Additional language files for glibc
@@ -288,9 +295,15 @@ grep "^FAIL: nptl/tst-eintr1" tests.sum >/dev/null && n=$((n+1)) ||:
 %defattr(-,root,root)
 # TODO: Excluding for now to remove dependency on PERL
 # /usr/bin/mtrace
-%{_lib64dir}/*.a
-%{_lib64dir}/*.o
+%{_lib64dir}/{g,M,S}crt1.o
+%{_lib64dir}/crti.o
+%{_lib64dir}/crtn.o
 %{_includedir}/*
+
+%files static
+%defattr(-,root,root)
+%{_lib64dir}/{,r,gr}crt1.o
+%{_lib64dir}/*.a
 
 %files -f %{name}.lang lang
 %defattr(-,root,root)
--- a/toolkit/scripts/check_static_glibc.py
+++ b/toolkit/scripts/check_static_glibc.py
@@ -1,0 +1,67 @@
+#!/usr/bin/python3
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from pyrpm.spec import Spec
+from os.path import dirname, realpath
+
+import argparse
+import re
+import sys
+
+def get_glibc_version():
+    glibc_spec = Spec.from_file("SPECS/glibc/glibc.spec")
+    epoch_prefix = "{}:".format(glibc_spec.epoch) if glibc_spec.epoch else ""
+    release_without_dist = glibc_spec.release.replace("%{?dist}", "")
+    release_suffix = "-{}".format(release_without_dist)
+    glibc_version = epoch_prefix + glibc_spec.version + release_suffix
+    return glibc_version
+
+def check_spec(path, glibc_version):
+    spec = Spec.from_file(path)
+    for br in spec.build_requires:
+        if br.name == "glibc-static":
+            issues = []
+            if not br.version:
+                issues.append(f"  * Does not specify a minimum version for its `glibc-static` BuildRequires")
+            if br.version and br.version != glibc_version:
+                issues.append(f"  * Does not specify the latest `glibc-static` version in its BuildRequires")
+            if br.operator and br.operator != ">=":
+                issues.append(f"  * Uses bad BuildRequire `glibc-static` comparison operator `{br.operator}`")
+
+            if issues:
+                print(f"Specfile {spec.name} (at {path}):")
+                for issue in issues:
+                    print(issue)
+                print(f"  Suggestion: Use `glibc-static >= {glibc_version}`")
+                print()
+                return False
+    return True
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description="Tool for checking if an RPM spec file follows CBL-Mariner's guidelines.")
+    parser.add_argument('specs',
+                        metavar='spec_path',
+                        nargs='+',
+                        help='path to an RPM spec file')
+    args = parser.parse_args()
+
+    glibc_version = get_glibc_version()
+
+    print(f"glibc-static version detected: {glibc_version}")
+    print()
+
+    specs_correct = True
+    for spec in args.specs:
+        if not check_spec(spec, glibc_version):
+            specs_correct = False
+
+    if specs_correct:
+        print("====================== Spec verification PASSED ======================")
+    else:
+        print("""====================== Spec verification FAILED ======================
+
+Please update the spec files listed above.
+""")
+        sys.exit(1)
--- a/toolkit/scripts/check_static_glibc.py
+++ b/toolkit/scripts/check_static_glibc.py
@@ -40,7 +40,7 @@ def check_spec(path, glibc_version):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(
-        description="Tool for checking if an RPM spec file follows CBL-Mariner's guidelines.")
+        description="Tool for checking if an RPM spec file correctly handles dependencies for statically linked glibc.")
     parser.add_argument('specs',
                         metavar='spec_path',
                         nargs='+',
--- a/SPECS-EXTENDED/dyninst/dyninst.spec
+++ b/SPECS-EXTENDED/dyninst/dyninst.spec
@@ -1,7 +1,7 @@
 Summary: An API for Run-time Code Generation
 License: LGPLv2+
 Name: dyninst
-Release: 7%{?dist}
+Release: 8%{?dist}
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 URL: http://www.dyninst.org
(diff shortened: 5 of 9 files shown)


jslobodzian (approved):

SPECS-EXTENDED/buildah/buildah.spec
```diff
@@ -40,7 +40,7 @@ BuildRequires: device-mapper-devel
 BuildRequires: golang
 BuildRequires: git
 BuildRequires: glib2-devel
-BuildRequires: glibc-static
+BuildRequires: glibc-static >= 2.35-3
```
PawelWMS, line 43: The release component (at least in case of Mariner) always contain the `.cm[:digit:]` part as well, so it's needed, I'd suggest something like this:

```suggestion
BuildRequires: glibc-static >= 2.35-3.cm2
```

Ideally, I'd even add a macro with the required version and release at the top of the spec to make it easier to spot, in case any updates are necessary.

Having said the above, we usually ever need to set a requirement for a specific version, because release tag bumps are not expected to introduce any significant changes. If that's the case for `buildah` as well, then I'd stick to this:

```suggestion
BuildRequires: glibc-static >= 2.35
```

SPECS-EXTENDED/catatonit/catatonit.spec
```diff
@@ -13,7 +13,7 @@ BuildRequires: automake
 BuildRequires: file
 BuildRequires: gcc
 BuildRequires: git
-BuildRequires: glibc-static
+BuildRequires: glibc-static >= 2.35-3
```
PawelWMS, line 16: Similar comments here and other specs as in `buildah.spec`.

SPECS/glibc/glibc.spec
```diff
@@ -288,14 +295,23 @@ grep "^FAIL: nptl/tst-eintr1" tests.sum >/dev/null && n=$((n+1)) ||:
 %defattr(-,root,root)
 # TODO: Excluding for now to remove dependency on PERL
 # /usr/bin/mtrace
-%{_lib64dir}/*.a
-%{_lib64dir}/*.o
+%{_lib64dir}/{g,M,S}crt1.o
+%{_lib64dir}/crti.o
+%{_lib64dir}/crtn.o
```
PawelWMS, line 300: Why are we leaving some `.o` files here and some are moved to the new subpackage?

bossmc: I've kept the CRT files needed for linking dynamic executables (and PIEs) and shared objects in `glibc-devel` and moved the ones for `static` and `static-pie` binaries to the `glibc-static` package.

The goal is to force packages that statically link to `glibc` to explicitly depend on `glibc-static` (today they can depend on `glibc-devel` and link statically - or even don't depend on `glibc` at all and still statically link to it because it's in `pkggen`)

SPECS-EXTENDED/buildah/buildah.spec
bossmc: We discussed and settled on `2.35-3%{?dist}` (in case we were somehow mixing different releases of Mariner)

toolkit/scripts/check_static_glibc.py
```diff
@@ -0,0 +1,67 @@
...
+
+import argparse
+import re
+import sys
+
+def get_glibc_version():
+    glibc_spec = Spec.from_file("SPECS/glibc/glibc.spec")
+    epoch_prefix = "{}:".format(glibc_spec.epoch) if glibc_spec.epoch else ""
+    release_without_dist = glibc_spec.release.replace("%{?dist}", "")
+    release_suffix = "-{}".format(release_without_dist)
```
PawelWMS, line 16: For consistency (and to keep things short), could we use the f-strings everywhere:

```suggestion
    glibc_spec = Spec.from_file("SPECS/glibc/glibc.spec")
    epoch_prefix = f"{glibc_spec.epoch}:" if glibc_spec.epoch else ""
    release_without_dist = glibc_spec.release.replace("%{?dist}", "")
    release_suffix = f"-{release_without_dist}"
```

toolkit/scripts/check_static_glibc.py
```diff
@@ -0,0 +1,67 @@
...
+
+def check_spec(path, glibc_version):
+    spec = Spec.from_file(path)
+    for br in spec.build_requires:
+        if br.name == "glibc-static":
+            issues = []
+            if not br.version:
+                issues.append(f"  * Does not specify a minimum version for its `glibc-static` BuildRequires")
+            if br.version and br.version != glibc_version:
+                issues.append(f"  * Does not specify the latest `glibc-static` version in its BuildRequires")
```
PawelWMS, line 28: Nit-pick: I'd avoid f-strings when we use plain strings, so that it's clear there's nothing special happening here.

```suggestion
                issues.append("  * Does not specify a minimum version for its `glibc-static` BuildRequires")
            if br.version and br.version != glibc_version:
                issues.append("  * Does not specify the latest `glibc-static` version in its BuildRequires")
```

toolkit/scripts/check_static_glibc.py
```diff
@@ -0,0 +1,67 @@
...
+from os.path import dirname, realpath
+
+import argparse
+import re
+import sys
+
+def get_glibc_version():
+    glibc_spec = Spec.from_file("SPECS/glibc/glibc.spec")
+    epoch_prefix = "{}:".format(glibc_spec.epoch) if glibc_spec.epoch else ""
+    release_without_dist = glibc_spec.release.replace("%{?dist}", "")
```
PawelWMS, line 15: I think we still might need to remove this one. Otherwise that `br.version and br.version != glibc_version` below will probably always fail.

toolkit/scripts/check_static_glibc.py
```diff
@@ -0,0 +1,67 @@
...
+            if br.version and br.version != glibc_version:
+                issues.append(f"  * Does not specify the latest `glibc-static` version in its BuildRequires")
+            if br.operator and br.operator != ">=":
+                issues.append(f"  * Uses bad BuildRequire `glibc-static` comparison operator `{br.operator}`")
+
+            if issues:
+                print(f"Specfile {spec.name} (at {path}):")
+                for issue in issues:
+                    print(issue)
+                print(f"  Suggestion: Use `glibc-static >= {glibc_version}%{{?dist}}`")
```
PawelWMS, line 36: At this point it's probably no longer a suggestion.;)

```suggestion
                print(f"  Use: `glibc-static >= {glibc_version}%{{?dist}}`")
```

toolkit/scripts/check_static_glibc.py
bossmc: Turns out that `rpm.spec` "expands" macros (even though it doesn't know what any of them are?) so it parses the version in the `BuildRequires` as simply `2.35-3`...

Hence all I've done is changed the suggested string, but we don't actually check that the user has written the `%{?dist}`.

toolkit/scripts/check_static_glibc.py
bossmc: "Strongly suggest" 😝

bossmc:
I've realised that this is waay to agressive about moving `.a` files into `glibc-static` sadly there are a few (`libc_nonshared, dl, rt, g, util, mcheck`) that are linked statically into dynamically linked executables/libraries (most of those are making e.g. `-lpthread` not break now that `libpthread` is baked into `libc`).

I'm putting together a fuller change to the `glibc` spec that filters out the files we should be keeping in `glibc-devel` and which can be sequestered into `glibc-static`.

toolkit/scripts/check_static_glibc.py
```diff
@@ -0,0 +1,67 @@
...
+    epoch_prefix = "{}:".format(glibc_spec.epoch) if glibc_spec.epoch else ""
+    release_without_dist = glibc_spec.release.replace("%{?dist}", "")
+    release_suffix = "-{}".format(release_without_dist)
+    glibc_version = epoch_prefix + glibc_spec.version + release_suffix
+    return glibc_version
+
+def check_spec(path, glibc_version):
+    spec = Spec.from_file(path)
+    for br in spec.build_requires:
+        if br.name == "glibc-static":
```
PawelWMS, line 23: It might be worth to return an error in case there's no `BuildRequires: glibc-static...` in the spec. Or at least a warning, so that it's clear we're using this script wrong.

bossmc: The normal use case is to run this over _all_ spec files, we'd get ~2000 warnings each run so I thought it better to be quiet (channelling that "if nothing's wrong be silent" adage that many linux tools use)

PawelWMS: Well, as long as it's still quick.

bossmc: It takes ~1min to run over the whole spec set (`SPECS*/**/*.spec`)

SPECS/glibc/glibc.spec
```diff
@@ -209,6 +209,14 @@ popd
 sed -i 's@#! /bin/bash@#! /bin/sh@' %{buildroot}%{_bindir}/ldd
 sed -i 's@#!/bin/bash@#!/bin/sh@' %{buildroot}%{_bindir}/tzselect
 
+# Determine which static libs are needed in `glibc-devel` - the rest will be put
+# into `glibc-static`.  We need to keep the static shims for function that's now
+# in `libc.so` (since 2.34 - see https://developers.redhat.com/articles/2021/12/17/why-glibc-234-removed-libpthread)
+# and the "statically linked bit" of `libc.so` (called `libc_nonshared.a`)
+static_libs_in_devel_pattern="lib\(c_nonshared\|pthread\|dl\|rt\|g\|util\|mcheck\).a"
+ls -1 %{buildroot}%{_lib64dir}/*.a | grep -e "$static_libs_in_devel_pattern" | sed "s:^%{buildroot}::g" > devel.filelist
+ls -1 %{buildroot}%{_lib64dir}/*.a | grep -v -e "$static_libs_in_devel_pattern" | sed "s:^%{buildroot}::g" > static.filelist
```
PawelWMS, line 218: Idea to remove most of the `\` to make the pattern a bit easier to read:

```suggestion
static_libs_in_devel_pattern="lib(c_nonshared|pthread|dl|rt|g|util|mcheck)\.a"
ls -1 %{buildroot}%{_lib64dir}/*.a | grep -P "$static_libs_in_devel_pattern" | sed "s:^%{buildroot}::g" > devel.filelist
ls -1 %{buildroot}%{_lib64dir}/*.a | grep -v -P "$static_libs_in_devel_pattern" | sed "s:^%{buildroot}::g" > static.filelist
```

Another one to drop `sed`:

```suggestion
static_libs_in_devel_pattern="(?<=%{buildroot}).*lib(c_nonshared|pthread|dl|rt|g|util|mcheck)\.a"
ls -1 %{buildroot}%{_lib64dir}/*.a | grep -oP "$static_libs_in_devel_pattern" > devel.filelist
ls -1 %{buildroot}%{_lib64dir}/*.a | grep -v -oP "$static_libs_in_devel_pattern" > static.filelist
```

Feel free to completely ignore both, though. ;)

PawelWMS (approved):
Looks good, added one more small suggestion, but nothing important.

SPECS/libacvp/libacvp.spec
```diff
@@ -1,15 +1,16 @@
...
 Version:        1.3.0
-Release:        2%{?dist}
+Release:        3%{?dist}
 License:        ASL 2.0
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 Group:          Development/Libraries
 URL:            https://github.com/cisco/libacvp
-# Source0:      https://github.com/cisco/%%{name}/archive/v%%{version}.tar.gz
+# Source0:        https://github.com/cisco/%%{name}/archive/v%%{version}.tar.gz
```
PawelWMS, line 10: Since we're already here, could we remove the commented-out `Source0` and replace the line below with:

```suggestion
Source0:        https://github.com/cisco/%{name}/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
```

SPECS/kernel/config
```diff
@@ -12,7 +12,6 @@ CONFIG_LD_IS_BFD=y
 CONFIG_LD_VERSION=23700
 CONFIG_LLD_VERSION=0
 CONFIG_CC_CAN_LINK=y
-CONFIG_CC_CAN_LINK_STATIC=y
```
PawelWMS, line 15: Modifying the kernel config will take more changes than that, I'm afraid. For example, take a look at #3826.

I'm not sure, if we want to bundle them up into this PR. We usually try to keep kernel changes in a separate PR.

SPECS/libacvp/libacvp.spec
```diff
@@ -35,7 +36,7 @@ and the library itself.
 ./configure \
     --prefix=%{_prefix} \
     --enable-offline \
-    CFLAGS="-pthread -DACVP_NO_RUNTIME -DOPENSSL_KWP -DOPENSSL_KDF_SUPPORT -O0 -g -fcommon" \
+    CFLAGS="-pthread -DACVP_NO_RUNTIME -DOPENSSL_KWP -DOPENSSL_KDF_SUPPORT -O2 -g -fcommon" \
```
PawelWMS, line 39: Curious: why was this necessary?

PawelWMS (changes requested):
Blocking until we figure out the kernel changes and dropping `libhugetlbfs-tests`.

SPECS/kernel/config
jslobodzian: This would need to be a separate change.  Why is this needed?

SPECS/libhugetlbfs/libhugetlbfs.spec
```diff
@@ -62,36 +62,24 @@ segment remapping behavior. hugectl sets environment variables for using huge
 pages and then execs the target program. hugeadm gives easy access to huge page
 pool size control. pagesize lists page sizes available on the machine.
 
-%package tests
```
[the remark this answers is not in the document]
jslobodzian: A %check section would be better, but I agree we shouldn't drop this.

SPECS/kernel/config
christopherco: I had a chat with @bossmc on this. The glibc change alters this `CONFIG_CC_CAN_LINK_STATIC` Kconfig, which is dynamically detected during kernel build. The config is NOT meant to be user-settable. The only spots this config is used is when `CONFIG_BPFILTER_UMH` or `CONFIG_BPF_PRELOAD_UMD` is enabled, both of which are not active/enabled in our kernels. Therefore there is no resulting change in the underlying kernel builds.

So IMO we can consider this change as a no-patch situation and forego the dash-roll

SPECS/libacvp/libacvp.spec
bossmc: `_FORTIFY_SOURCE` has no effect with `-O0` (also why build a non-optimized library in a package)

SPECS/libhugetlbfs/libhugetlbfs.spec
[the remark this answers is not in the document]
bossmc: I can (fairly easily) move the tests into a `%check` but those tests are dependent on the CPU and kernel in the build system supporting hugetables, and being configured to have enough of them (>=512).  I'm not sure that's an assumption I can make safely (or reasonably) - in particular it makes building locally much more annoying (for example WSL has no configured hugetables).

I do hear the "don't remove packages" adage though... an alternative would be to remove the statically linked tests from the package (or give up and make this package depend on `glibc-static` which feels like a cop-out).

bossmc:
That's everything except `libhugetlbfs` - I'm going to look at disabling the statically linked tests and re-instating the `-tests` package.

SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -8,13 +8,11 @@ Vendor:         Microsoft Corporation
 Distribution:   Mariner
 Group:          System/Libraries
 URL:            https://github.com/microsoft/SymCrypt
-#Source0:       https://github.com/microsoft/SymCrypt/archive/v%{version}.tar.gz
-Source0:        %{name}-%{version}.tar.gz
-#Source1        https://github.com/smuellerDD/jitterentropy-library/archive/v3.3.1.tar.gz
-Source1:        jitterentropy-library-3.3.1.tar.gz
+Source0:        https://github.com/microsoft/SymCrypt/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
+Source1:        https://github.com/smuellerDD/jitterentropy-library/archive/v3.3.1.tar.gz#/jitterentropy-library-3.3.1.tar.gz
```
PawelWMS, line 12: Nice, thank you for this clean-up.

PawelWMS (approved):

bossmc:
🚀

PR #3753, opened 2022-09-14
klipper-lb: add package

guptavinayak11:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

New spec file created for klipper-lb. This package is used for AksIoT and a container image with the package must be built by Microsoft.

###### Change Log  

- New spec file for klipper-lb

**NO**

###### Test Methodology

- Built locally with the Mariner Toolchain.

BuddyBuild:
- AMD64 - [236923](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=236923&view=results)
- ARM64 - [236924](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=236924&view=results)

Closes : #3729


SPECS/klipper-lb/klipper-lb.spec
```diff
@@ -0,0 +1,38 @@
...
+
+# Dockerfile in the package folder runs CMD ["entry"], so putting this bash script file in appropriate location.
+# So that it can be executed without any problem.
+
+%install
+install -d %{buildroot}%{_bindir}
+install entry %{buildroot}%{_bindir}/entry
+
+%files
+%{_bindir}/entry
```
PawelWMS, line 33: Considering that we're only installing a short script with this package, the `Summary` and `%description` fields seem not to match. Are you sure this is all this package should contain?

SPECS/klipper-lb/klipper-lb.spec
```diff
@@ -0,0 +1,38 @@
...
+Name:           klipper-lb
+Version:        0.3.5
+Release:        1%{?dist}
+License:        ASL 2.0
+URL:            https://github.com/k3s-io/klipper-lb
+Group:          Applications/Text
+Vendor:         Microsoft
+Distribution:   Mariner
+Source0:        https://github.com/k3s-io/%{name}/archive/refs/tags/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
+#Note that the source file should be renamed to the format {name}-%{version}.tar.gz
```
PawelWMS, line 11: Nit-pick: usually we add comments regarding a tag above it and use double `%` characters in the comments, so that the spec parsing tooling won't print warnings:

```suggestion
# Note that the source file should be renamed to the format '%%{name}-%%{version}.tar.gz'
Source0:        https://github.com/k3s-io/%{name}/archive/refs/tags/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
```

SPECS/klipper-lb/klipper-lb.spec
guptavinayak11: The thing here is that the [entry script](https://github.com/k3s-io/klipper-lb/blob/master/entry) that I am installing has already been built in the [repo](https://github.com/k3s-io/klipper-lb). 
Also, this script does the simple thing of setting up iptables, as this is a load balancer. You can see [dockerfile](https://github.com/k3s-io/klipper-lb/blob/master/package/Dockerfile) for more information.

I am not sure by if I understand your question about Summary and %description fields seem not to match.
Let me know if you have any more questions. :-)

PawelWMS: Sure thing. What I had in mind about the summary and description is that they claim this package contains the "klipper runtime image", which is not true. It only install one simple script, no images in sight. Also, the script seems to have heavy expectations regarding existence of environment variables like `DEST_IPS` or `DEST_PROTO`. So there's two things:

- The package summary/description is incorrect.
- I'm not sure how the script is supposed to even work.

Considering these two points, I think I'd need more details about what this package is expected to achieve.

guptavinayak11: Sure, in the summary thing I will update it to "Provides a way to use an integrated service load balancer for kubernetes users". 
For the existence of environment variables like `DEST_IPS` or `DEST_PROTO`, these are provided when we deploy them using yaml files.
So, for example here is a small part of deployment yaml of how I can provide them:

```
spec:
      serviceAccountName: klipper-lb
      containers:
      - env:
        - name: SRC_PORT
          value: "80"
        - name: DEST_PROTO
          value: TCP
        - name: DEST_PORT
          value: "80"
```

These environment variables will then get passed to the pods or containers, when they get deployed. Hence the [entry script](https://github.com/k3s-io/klipper-lb/blob/master/entry) will fetch these from here.

Let me know if you have any further questions. I tried my best to explain it :-)

lilustga: @PawelWMS, the end goal of spec file is to produce a golden container image for klipper-lb. In the [onboarding instructions](https://dev.azure.com/mariner-org/ECF/_wiki/wikis/ECF.wiki/923/%F0%9F%93%A6-Golden-Containers) creating specs/RPMs is listed as the first step. The idea being that the container image will be then constructed based on the dockerfile.

mfrw: @lilustga If this is the case, are we doing the right thing by including this in the core SPECS ?
If the end goal is a container image, should we not use the `Golden-Container` pipelines for this ?

I think, I maybe missing something trivial :(

guptavinayak11: **Update:** I discussed with @mfrw about this. So, in [Golden Containers process](https://dev.azure.com/mariner-org/ECF/_wiki/wikis/ECF.wiki/923/%F0%9F%93%A6-Golden-Containers) the first step is to create the spec file to build the RPMs in mariner, then to use these RPMs in a dockerfile as @lilustga pointed in the earlier comment. That is why we are including this is the SPECS folder of the [CBL-Mariner repo](https://github.com/microsoft/CBL-Mariner/).

Now, if we don't want to include spec files then is there any other way to provide CBL-Mariner pipelines the RPMs which we need in the dockerfiles used for provisioning them? 

@mfrw has suggested me to get views from @PawelWMS @mandeepsplaha 

Also, I am completely new to packaging these repos, but this is what understanding I have from reading the documentation.

PawelWMS (changes requested):
It seems the PR might have been abandoned. Blocking until we resolve the comments or close the PR. Please ping me in case of updates.

guptavinayak11:
@PawelWMS, yes, I am closing this PR for now. Thanks for reminding me.

PR #3767, opened 2022-09-14
Adding grubenv file by default. 

MaxBrodeurUrbas:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Background: The GRUB environment block is a preallocated 1024 byte file that is used to save a small amount of information from one boot to the next. This file is typically located in /boot/grub/grubenv in most distributions and is edited using the grub-editenv or grubby commands. Any variables added to the block are persistently stored and are read during the next boot.  

Problem: Mariner configures it's “grubenv” block differently from other popular distributions causing standard methods of grubenv modification to be unsupported. This makes it more difficult to support Mariner in other projects and standard tooling won’t work out of the box.  

Mariner has a /boot/grub2/grub.cfg file which acts as a grubenv equivalent. This file is modified with sed commands rather than using the standard tools. Mariner’s dependency on sed commands causes friction when adding Mariner support in other projects and some hesistancy from others as this approach is slightly risky. 

Solution: I've added an empty grub env block to mariner by default and referenced the contents of this block in our current /boot/grub2/grub.cfg file. This way any attempt to modify the grubenv using standard tools like grubby will be supported. No functionality will regress since Mariner’s sed insertion approach would still be supported.  

[grubenv overview](https://www.gnu.org/software/grub/manual/grub/html_node/Environment-block.html)  

###### Change Log  

- Adding grubenv file by default and reference

**NO**

###### Associated issues  

- [work item](https://microsoft.visualstudio.com/OS/_workitems/edit/40767682)

###### Test Methodology

- Pipeline image gen and manual testing.


toolkit/resources/assets/grub2/grub.cfg
```diff
@@ -8,11 +8,14 @@ if [ -f  $bootprefix/systemd.cfg ]; then
...
 fi
+if [ -f $bootprefix/grub2/grubenv ]; then
+	load_env -f $bootprefix/grub2/grubenv
+fi
 
 set rootdevice={{.RootPartition}}
 
 menuentry "CBL-Mariner" {
-	linux $bootprefix/$mariner_linux {{.LuksUUID}} {{.LVM}} {{.IMAPolicy}} {{.ReadOnlyVerityRoot}} {{.SELinux}} rd.auto=1 root=$rootdevice $mariner_cmdline lockdown=integrity sysctl.kernel.unprivileged_bpf_disabled=1 $systemd_cmdline {{.ExtraCommandLine}}
+	linux $bootprefix/$mariner_linux {{.LuksUUID}} {{.LVM}} {{.IMAPolicy}} {{.ReadOnlyVerityRoot}} {{.SELinux}} rd.auto=1 root=$rootdevice $mariner_cmdline lockdown=integrity sysctl.kernel.unprivileged_bpf_disabled=1 $systemd_cmdline $kernelopts {{.ExtraCommandLine}}
```
dmcilvaney, line 18: I'm wondering if the grubenv's $kernelopts should be last, that way editing the grubenv would override the entries passed at image build time (ExtraCommandLine).

MaxBrodeurUrbas: oh interesting, would you recommend moving it to the front so it has the lowest priority?

dmcilvaney: It should probably have the highest priority, since this is effectively trying to overwrite the existing configs

dmcilvaney (approved):

PR #3815, opened 2022-09-22
Add initial support for finalizeImage

neha170:
###### Merge Checklist  

- [X] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [X] Ready to merge

---

###### Summary 

Add option to provide a final script that modifies the chroot before creating an image

###### Change Log  
- Add option to provide a final script that modifies the chroot before creating an image

**NO**

###### Associated issues  

- #[41439909](https://microsoft.visualstudio.com/OS/_workitems/edit/41439909)

###### Test Methodology

- Functional testing done: tested path with different number of scripts (no script, 1 script, multiple script paths), with and without args

### DIFF
--- a/toolkit/tools/imager/imager.go
+++ b/toolkit/tools/imager/imager.go
@@ -469,7 +469,7 @@ func fixupExtraFilesIntoChroot(installChroot *safechroot.Chroot, config *configu
 }
 
 func cleanupExtraFiles() (err error) {
-	dirsToRemove := []string{additionalFilesTempDirectory, postInstallScriptTempDirectory, sshPubKeysTempDirectory}
+	dirsToRemove := []string{additionalFilesTempDirectory, postInstallScriptTempDirectory, finalizeImageScriptTempDirectory, sshPubKeysTempDirectory}
 
 	for _, dir := range dirsToRemove {
 		logger.Log.Infof("Cleaning up directory %s", dir)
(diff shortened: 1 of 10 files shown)


dmcilvaney:
We should have a section in the docs in `formats/imageconfig.md` for this.

toolkit/tools/imager/imager.go
```diff
@@ -577,6 +593,13 @@ func buildImage(mountPointMap, mountPointToFsTypeMap, mountPointToMountArgsMap,
 		}
 	}
 
+	// Run finalize image scripts from within the installroot chroot
```
dmcilvaney, line 596: Curious if we should limit this to non-rootfs type images... I vaguely recall having misgivings about allowing this for non-image/disk based install scenarios but I don't recall what they were.

Any thoughts?

christopherco: Only issue I could think of is possible confusion with the post-install scripts field for rootfs only.

Maybe something to do with the bespoke delta update path, which would fall under the category of non-image-based install scenarios?

christopherco (approved):

toolkit/tools/imager/imager.go
christopherco: Spoke with @dmcilvaney and we both can't see any issues with the non-image path

PR #3864, opened 2022-09-28
systemd: gpt-auto fixes for backing device detection

avylove:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

gpt-auto-generator returns 1 on Sphere because the backing block device or `/` and `/usr` can't be determined. In systemd >= 251  this error is better handled. This applies the fix from later versions to 250.3.

###### Change Log  

- systemd and systemd-bootstrap specs updated to include patch for https://github.com/systemd/systemd/pull/22506

**YES**

###### Associated issues  

- https://github.com/Azure/azure-sphere-v2-are-fs/issues/114

###### Test Methodology

- Local build
- [Sphere pipeline](https://dev.azure.com/msazuresphere/Public/_build/results?buildId=298232&view=results)

### DIFF
--- a/SPECS/systemd/systemd-bootstrap.signatures.json
+++ b/SPECS/systemd/systemd-bootstrap.signatures.json
@@ -1,7 +1,5 @@
 {
  "Signatures": {
-  "add-fsync-sysusers-passwd.patch": "c86c20c9bf0d79faeb99c454e8fb79df1b81d9243f45a209f7c8f58924ec7dba",
-  "gpt-auto-devno-not-determined.patch": "b75ee0a9b0e4f978d039cfbe9e5c9fd9314f99c3cd5e29f0d79752f9fc628c",
   "50-security-hardening.conf": "c9fc6624a3178cc0980ec3495ac2d5f02a8a1f3a4cdadb8770628d7394b92e2c",
   "99-dhcp-en.network": "158c0003826cef9d780799c6edb2814800d4518f85f884710550f4c46d14be67",
   "systemd-250.3.tar.gz": "87b0eee7b6e5aaab2ab56d158f9536daa6bfd5de011f2a5fc6ccdd81ee1e7a24",
--- a/SPECS/systemd/systemd-bootstrap.signatures.json
+++ b/SPECS/systemd/systemd-bootstrap.signatures.json
@@ -1,6 +1,7 @@
 {
  "Signatures": {
   "add-fsync-sysusers-passwd.patch": "c86c20c9bf0d79faeb99c454e8fb79df1b81d9243f45a209f7c8f58924ec7dba",
+  "gpt-auto-devno-not-determined.patch": "b75ee0a9b0e4f978d039cfbe9e5c9fd9314f99c3cd5e29f0d79752f9fc628c",
   "50-security-hardening.conf": "c9fc6624a3178cc0980ec3495ac2d5f02a8a1f3a4cdadb8770628d7394b92e2c",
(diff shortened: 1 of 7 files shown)


SPECS/systemd/systemd-bootstrap.signatures.json
```diff
@@ -1,6 +1,7 @@
 {
  "Signatures": {
   "add-fsync-sysusers-passwd.patch": "c86c20c9bf0d79faeb99c454e8fb79df1b81d9243f45a209f7c8f58924ec7dba",
+  "gpt-auto-devno-not-determined.patch": "b75ee0a9b0e4f978d039cfbe9e5c9fd9314f99c3cd5e29f0d79752f9fc628c",
```
PawelWMS, line 4: We don't use and need signatures for patches. It seems like the previous one was erroneously added - we can remove both here.

SPECS/systemd/systemd-bootstrap.spec
```diff
@@ -12,9 +12,12 @@ Source1:        50-security-hardening.conf
 Source2:        systemd.cfg
 Source3:        99-dhcp-en.network
 Patch0:         fix-journald-audit-logging.patch
-# Can be removed once we update systemd to a version containing the following commit:
+# Patch1 can be removed once we update systemd to a version containing the following commit:
 # https://github.com/systemd/systemd/commit/19193b489841a7bcccda7122ac0849cf6efe59fd
 Patch1:         add-fsync-sysusers-passwd.patch
+# Patch2 can be removed once we update systemd to a version containing the following commit:
+# https://github.com/systemd/systemd/commit/d74da762a3b1edbb7935c3b3a229db601f734125
```
PawelWMS, line 19: It seems the commit also modified `src/libsystemd/sd-device/sd-device.c`. Is this change unapplicable for our version of `systemd`?

SPECS/systemd/systemd.signatures.json
```diff
@@ -1,5 +1,7 @@
 {
  "Signatures": {
+  "add-fsync-sysusers-passwd.patch": "c86c20c9bf0d79faeb99c454e8fb79df1b81d9243f45a209f7c8f58924ec7dba",
+  "gpt-auto-devno-not-determined.patch": "b75ee0a9b0e4f978d039cfbe9e5c9fd9314f99c3cd5e29f0d79752f9fc628c0d",
```
PawelWMS, line 4: Same comment here - no need to add the signatures.

avylove: Removed

SPECS/systemd/systemd-bootstrap.spec
avylove: I had the merge commit in the comment instead of the commit for the patch. There are two commits in that PR that were merged. They are related but not dependent on each other. I could update the patch for both commits in the PR, but I'm not sure it's needed. What do you think?

PawelWMS: It depends but the main point, if I understand your change correctly, is to backport a bug fix to our version of `systemd`. If the whole PR is aimed at addressing fixing the bug, then we should take the whole PR. If the PR contains changes unrelated to the bug, then I'd suggest taking only the commits explicitly aimed at fixing the bug. I'm going to trust your judgement here, since I'm not familiar with the issue.

avylove: I think the included commit is sufficient. It solves the issue where gpt-auto was returning 1. The other commit only improves the error message for a corner case we haven't encountered.

PawelWMS (approved):

PR #3952, opened 2022-10-07
add libyang 2 to mariner SPECs

tofay:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add libyang to Mariner, for use in an Azure for Operators service.

###### Change Log  

- Change Add libyang2 to Mariner

**NO**

###### Test Methodology

- Built locally and verified against existing project that uses libyang.


SPECS/libyang/libyang.spec
```diff
@@ -0,0 +1,103 @@
...
+%ifarch %{arm} ppc64le
+%global run_valgrind_tests OFF
+%else
+%global run_valgrind_tests ON
+%endif
+Summary:        YANG data modeling language library
+Name:           libyang
+Version:        2.0.231
+Release:        1%{?dist}
+License:        BSD-3-Clause
```
PawelWMS, line 12:
```suggestion
License:        BSD
```

SPECS/libyang/libyang.spec
```diff
@@ -0,0 +1,103 @@
...
+%install
+pushd build
+%make_install
+popd
+mkdir -m0755 -p %{buildroot}/%{_docdir}/libyang
+cp -a doc/html %{buildroot}/%{_docdir}/libyang/html
+
+%files
+%license LICENSE
+%{_libdir}/libyang.so.2
```
PawelWMS, line 78: If this file's a symbolic link, then it should probably go to the `devel` subpackage.

tofay: Why? This will be needed at runtime for binaries that need libyang.so.2. (This is just standard SO versioning, right? I don't think there's anything unusual here)

SPECS/libyang/libyang.spec
tofay: Sure, but OOI why? I see a bunch of specs in mariner with BSD-3-Clause as the license, and assumed it was better to have a more specific declared license.

SPECS/libyang/libyang.spec
```diff
@@ -0,0 +1,103 @@
+# valgrind finds invalid writes in libcmocka on arm and power
+# see bug #1699304 for more information
+%ifarch %{arm} ppc64le
+%global run_valgrind_tests OFF
+%else
+%global run_valgrind_tests ON
+%endif
```
oliviacrain, line 7: We don't support 32-bit ARM and PowerPC architectures, so we can remove this option.

tofay: Ah I hadn't realised `arm` was 32-bit arm, thanks!

SPECS/libyang/libyang.spec
PawelWMS: Fair question. We try to stick to [the "Short names" Fedora is using for its specs](https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#Good_Licenses). There are some leftovers from before we were more strick here and sometimes things slip through the cracks, but in general we try to follow that convention.

SPECS/libyang/libyang.spec
PawelWMS: You're right, I got all mixed up here. Sorry for the trouble.

PawelWMS (approved):

PR #3958, opened 2022-10-10
mariner-release: add %{dist} macro to persistent macro file.

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When using a Mariner image to invoke `rpmbuild`, it is expected that `%{dist}` is defined (".cm2" etc). In our build tooling we explicitly pass that definition in on the command line, but this is not standard. This macro is defined in `/usr/lib/rpm/macros.d/macros.dist` on Fedora systems.

###### Change Log  

- Add `/usr/lib/rpm/macros.d/macros.dist` to `mariner-release` which contains a single macro `%%{dist} %{dist}` which matches the distro tag at the time of the build.

**NO**

###### Test Methodology

- Local build
- Full pipeline in progress


SPECS/mariner-release/mariner-release.spec
```diff
@@ -52,6 +52,11 @@ cat > %{buildroot}%{_sysconfdir}/issue.net <<- EOF
 Welcome to CBL-Mariner %{mariner_release_version} (%{_arch})
 EOF
 
+mkdir -p %{buildroot}%{_rpmconfigdir}/macros.d
+cat > %{buildroot}%{_rpmconfigdir}/macros.d/macros.dist <<- EOF
+%%dist %{dist}
+EOF
+
```
PawelWMS, line 59: 2 ideas:

1. Move it to one of the `macros*` files installed with `mariner-rpm-macros` (I'm thinking either `macros` or some new one). This way we keep macro definitions in one place, so they are still easy to analyze/discover when searching through our repo. Also, if anyone is trying to build on Mariner but not with out tooling, it's a reasonable requirement to have `mariner-rpm-macros` installed.
2. I'm guessing we'd still like to avoid having to update the macro for each release of Mariner, so we could do something along the lines of `%dist @DIST_TAG@` in the file and then `sed` during the build of the package.

What do you think?

PawelWMS: Ad 1. I think I found 2 core specs where the macros file is created inside a spec (I searched for `\s>?>.*macro` through all spec files). Not great but I think we can/should fix that, so they don't serve as precedence, like you mentioned. There's 3 more such specs in `SPECS-EXTENDED` I believe.

Ad 2. You're right, that `%{dist}` should be resolved during build time but if we use spec sources instead of dynamically building the macro files in the spec, then we need to update them with this resolved value. Take a look at what `multilib-rpm-config.spec` is doing with its `SOURCE1` inside the `%build` section, to see what I had in mind.

dmcilvaney: For 1), I ran `tdnf provides /usr/lib/rpm/macros.d/*` to check what packages are claiming to provide macro files.

Bit of cleanup gives:
```
cmake
color-filesystem
cups-devel
fdupes
firewalld-filesystem
fontpackages-devel
httpd-devel
javapackages-filesystem
javapackages-local-bootstrap
javapackages-tools
kf5-rpm-macros
lua-rpm-macros
lua-srpm-macros
mariner-check-macros
mariner-rpm-macros
meson
multilib-rpm-config
ninja-build
perl-macros
pkgconf
pyproject-rpm-macros
qt5-qtbase-common
qt5-rpm-macros
ruby
rubygems-devel
systemd-bootstrap-rpm-macros
systemd-rpm-macros
```

For 2, sounds good. I'll shift over to that flow.

SPECS/mariner-rpm-macros/mariner-rpm-macros.spec
```diff
@@ -87,6 +88,9 @@ mkdir -p %{buildroot}%{rcluadir}/{rpm,srpm}
 install -p -m 644 -t %{buildroot}%{rcluadir} common.lua
 install -p -m 644 -t %{buildroot}%{rcluadir}/srpm forge.lua
 
+sed -e 's|@DIST@|%{dist}|g' %{SOURCE28} > macros.dist
+install -m 644 -p macros.dist %{buildroot}/%{macrosdir}
```
PawelWMS, line 92: I'd move `sed` before the `mkdir -p %{buildroot}%{_rpmconfigdir}/macros.d` line and remove the new `install`, since all `macros.*` files are already installed above.

dmcilvaney: Done

dmcilvaney:
I've rerun a local build with the new changes and the %dist macro expands the same way as before.

SPECS/mariner-rpm-macros/macros.dist
```diff
@@ -0,0 +1 @@
+%dist @DIST@
```
PawelWMS, line 1: I see the PR check is now failing. I think we might just need a small tweak inside `rpmops.sh` where we replace

```bash
rpmspec "${DEFINES[@]}" -D "_sourcedir $sourcedir" "${MACROS[@]}" "$@"
```

with

```bash
rpmspec "${MACROS[@]}" "${DEFINES[@]}" -D "_sourcedir $sourcedir" "$@"
```

PawelWMS (approved):

dmcilvaney:
Rebased to current main branch

jiasli:
Interestingly, `%{dist}` can be resolved on Fedora, Red Hat, but not Mariner:

```
> docker run -it fedora rpm --eval '%{dist}'
.fc37

> docker run -it registry.access.redhat.com/ubi8/ubi:8.4 rpm --eval '%{dist}'
.el8

> docker run -it mcr.microsoft.com/cbl-mariner/base/core:2.0 rpm --eval '%{dist}'
%{dist}
```

dmcilvaney:
I had a dig around the upstream repos when looking at this change. This macro is something that is set for most (if not all) rpm distros.

If a non-standard tag is desired, this can be overridden by passing `rpmbuild -D 'dist <custom_tag>' ... `, or if you want to explicitly clear it you can use `rpmbuild --undefine 'dist' ...`

PR #3970, opened 2022-10-11
Remove DOWNLOAD_SRPMS flag and functionality

rlmenge:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This flag is broken in 2.0. When we query for the spec name and version, the file can't be parsed due to new macros within 2.0 specs. These macros are resolved in the package mariner_macros but since we are building outside of a chroot, this package can't be loaded.

Remove this flag for now as there is not a strong need for it. Future fixes will require a chroot so that mariner_macros is present or loaded.

https://microsoft.visualstudio.com/OS/_workitems/edit/40942962

###### Change Log  

- Remove DOWNLOAD_SRPMS and associated flags

**NO**

###### Associated issues  

https://microsoft.visualstudio.com/OS/_workitems/edit/40942962

###### Test Methodology

- [Pipeline build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=249043&view=results)

### DIFF
--- a/toolkit/scripts/srpm_pack.mk
+++ b/toolkit/scripts/srpm_pack.mk
@@ -50,30 +50,6 @@ $(BUILD_SRPMS_DIR): $(STATUS_FLAGS_DIR)/build_srpms.flag
 	@touch $@
 	@echo Finished updating $@
 
-ifeq ($(DOWNLOAD_SRPMS),y)
-$(STATUS_FLAGS_DIR)/build_srpms.flag: $(LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(SPECS_DIR)
-	for spec in $(LOCAL_SPECS); do \
-		spec_file=$${spec} && \
-		srpm_file=$$(rpmspec -q $${spec_file} --srpm --define='with_check 1' --define='dist $(DIST_TAG)' --queryformat %{NAME}-%{VERSION}-%{RELEASE}.src.rpm) && \
-		for url in $(SRPM_URL_LIST); do \
-			wget $${url}/$${srpm_file} \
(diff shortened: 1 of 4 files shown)


toolkit/scripts/srpm_pack.mk
```diff
@@ -50,30 +50,6 @@ $(BUILD_SRPMS_DIR): $(STATUS_FLAGS_DIR)/build_srpms.flag
 	@touch $@
 	@echo Finished updating $@
 
-ifeq ($(DOWNLOAD_SRPMS),y)
-$(STATUS_FLAGS_DIR)/build_srpms.flag: $(LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(SPECS_DIR)
```
PawelWMS, line 54: Is the idea here to remove the `DOWNLOAD_SRPMS` flag only until the [new srpmdownloader tool](https://github.com/rlmenge/CBL-Mariner/pull/23) is added? I'm asking because in general removing published public APIs/interfaces/features within one release of any project is usually discouraged. I doubt anyone except for us is using this flag, so it's probably not a big deal here, but to be on the safe side, I'd prefer to leave the flag in place and instead consider adding a warning, that this feature might be malfunctioning, but a fix is on the way. After all this still has a chance to work with derivative specs and `SRPM_URL_LIST` values. A minor benefit of doing it this way, is that we minimize chances of missing some documentation or functionality when we try to restore this feature later on.

I have nothing against removing `ALLOW_SRPM_DOWNLOAD_FAIL`. I think it doesn't do anything anyway.

rlmenge:
Looking to enable this functionality in future versions

PR #4019, opened 2022-10-13
add Microsoft GPG keys to installer env

Xiaohong-Deng:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add Microsoft GPG keys to the installer. The installer does not enable the feature to add Microsoft gpg key during the installation process which ensures pulling secure packages from secure repo servers.

###### Change Log  

- add GPG keys to `/toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json`

**NO**

###### Test Methodology

- Local build


toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
```diff
@@ -51,6 +51,7 @@
         "openssh-server",
         "openssl",
         "mariner-release",
+        "mariner-repos",
```
PawelWMS, line 54: Why are we doing this? What's the scenario/requirement/project? So far the ISO didn't use any external repositories by design, as it was expected to have all of the packages required by the final system available locally.

PawelWMS (changes requested):

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
henryli001, line 54: I believe mariner-repos-shared is the package that you would like to install here instead of the mariner-repos package since this will also pull in the official repo file that's not needed.

henryli001 (changes requested):

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
henryli001: This is for the PXE boot scenario where users won't pull packages from the local RPM repo built into the ISO but from external repos. Thus, these keys would need to be reside within the installer for that operation to succeed.

toolkit/resources/imageconfigs/iso_initrd.json
```diff
@@ -46,4 +46,4 @@
             ]
         }
     ]
-}
\ No newline at end of file
+}
```
henryli001, line 49: This change is not necessary.

henryli001 (approved):

PawelWMS (approved):

PR #4024, opened 2022-10-14
Support Mariner 1.0 build on Mariner 2.0 host

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Support Mariner 1.0 build on Mariner 2.0 host
As of today, Mariner 1.0 package build environment does not work on Mariner 2.0 hosts. This is due to the rpm version (4.17+) in Mariner 2.0 which defaults to an sqlite database backend and does not support bdb (Berkeley DB) database. Mariner 1.0 has rpm version 4.14 which only supports bdb. When the package worker chroot is created, the host machine's rpm tool is used to install the packages and create the database. I've modified the create_worker_chroot.sh script re-created the database using the rpm version in the chroot (which creates a valid bdb database).

Various symptoms seen when the sqlite database is created when building Mariner 1.0 on a host with incompatible rpm version:
```
Error(1022) : distroverpkg config entry is set to a package that is not installed. Check /etc/tdnf/tdnf.conf

INFO[0016] Initializing repository configurations       
ERRO[0016] Failed to clone kernel(,):<ID:13 Type:Remote State:Unresolved> from '<NO_SRPM_PATH>' in '<NO_REPO>' from RPM repo. Error: exit status 254 
ERRO[0016] Failed to resolve all nodes in the graph while resolving 'kernel(,):<ID:13 Type:Remote State:Unresolved> from '<NO_SRPM_PATH>' in '<NO_REPO>''

error: Failed to initialize NSS library
```

Note that on Mariner 2.0, this script (create_worker_chroot.sh) has a similar problem and similar solution. In that case, the typical host machine is Ubuntu 18.04 (with rpm 4.14 creating a bdb database) and Mariner 2.0 in the chroot expects an sqlite database.

Current Host OS/rpm/db versions (checked by running `rpm -E "%{_db_backend}"` in the latest containers)
```
Mariner 1.0: RPM version 4.14.2 - bdb
Mariner 2.0: RPM version 4.18.0 - sqlite
Ubuntu 18.04: RPM version 4.14.1 - bdb
Ubuntu 20.04: RPM version 4.14.2.1 - bdb
Ubuntu 22.04: RPM version 4.17.0 - sqlite
```

For relevant information see: https://rpm.org/user_doc/db_recovery.html and https://fedoraproject.org/wiki/Changes/RPM-4.16

###### Change Log  

- Change create_worker_chroot.sh to rebuild the rpm database inside of the chroot after the host machine installs the packages

**NO**

###### Test Methodology

- Pipeline build id: 250896 (rebuild 1.0-dev AMD64 packages)
- Pipeline build id: 250895 (trigger end-2-end AMD64 1.0-dev build)
- Pipeline build id: 250982 (rebuild 1.0-dev AMD64 images)
- Pipeline build id: 250987 (rebuild 1.0-dev Arm64 images)

### DIFF
--- a/toolkit/tools/pkggen/worker/create_worker_chroot.sh
+++ b/toolkit/tools/pkggen/worker/create_worker_chroot.sh
@@ -54,6 +54,31 @@ while read -r package || [ -n "$package" ]; do
     install_one_toolchain_rpm "$package"
 done < "$packages"
 
+# If the host machine rpm version is >= 4.16 (such as Mariner 2.0), it will create an "sqlite" rpm database backend incompatible with Mariner 1.0 (which uses "bdb")
+# To resolve this, enter the 1.0 chroot after the packages are installed, and use the older rpm tool in the chroot to re-create the database in "bdb" format.
+TEMP_DB_PATH=/temp_db
+echo "Setting up a clean RPM bdb (Berkeley DB) database in the chroot under '$TEMP_DB_PATH'."  | tee -a "$chroot_log"
+
+# These nodes are required in the chroot for certain tools (most importantly, the NSS library initialization required by rpm)
+sudo mkdir -pv $chroot_builder_folder/{dev,proc,sys,run}
+sudo mknod -m 600 $chroot_builder_folder/dev/console c 5 1
+sudo mknod -m 666 $chroot_builder_folder/dev/null c 1 3
+sudo mknod -m 444 $chroot_builder_folder/dev/urandom c 1 9
+
+chroot "$chroot_builder_folder" mkdir -pv "$TEMP_DB_PATH"
+chroot "$chroot_builder_folder" rpm --initdb --dbpath="$TEMP_DB_PATH"
+# Populating the bdb database with package info.
+while read -r package || [ -n "$package" ]; do
+    full_rpm_path=$(find "$rpm_path" -name "$package" -type f 2>>"$chroot_log")
+    cp $full_rpm_path $chroot_builder_folder/$package
+    echo "Adding RPM DB entry to worker chroot: $package."  | tee -a "$chroot_log"
+    chroot "$chroot_builder_folder" rpm -i -v --nodeps --noorder --force --dbpath="$TEMP_DB_PATH" --justdb "$package" &>> "$chroot_log"
+    chroot "$chroot_builder_folder" rm $package
+done < "$packages"
+echo "Overwriting old RPM database with the results of the conversion." | tee -a "$chroot_log"
+chroot "$chroot_builder_folder" rm -rf /var/lib/rpm
+chroot "$chroot_builder_folder" mv "$TEMP_DB_PATH" /var/lib/rpm
+
 HOME=$ORIGINAL_HOME
 
 # In case of Docker based build do not add the below folders into chroot tarball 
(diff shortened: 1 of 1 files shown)


toolkit/tools/pkggen/worker/create_worker_chroot.sh
```diff
@@ -54,6 +54,31 @@ while read -r package || [ -n "$package" ]; do
     install_one_toolchain_rpm "$package"
 done < "$packages"
 
+# If the host machine rpm version is >= 4.16 (such as Mariner 2.0), it will create an "sqlite" rpm database backend incompatible with Mariner 1.0 (which uses "bdb")
```
jslobodzian, line 57: Anything we can do to skip this step if rpm is older (or even better if the host rpm was built with a compatible version of rpm?)

jslobodzian (approved):

toolkit/tools/pkggen/worker/create_worker_chroot.sh
anphel31: We can probably query the rpm db backend version of the host, and skip this step. Let me take a look.

toolkit/tools/pkggen/worker/create_worker_chroot.sh
```diff
@@ -54,6 +54,31 @@ while read -r package || [ -n "$package" ]; do
...
+# To resolve this, enter the 1.0 chroot after the packages are installed, and use the older rpm tool in the chroot to re-create the database in "bdb" format.
+TEMP_DB_PATH=/temp_db
+echo "Setting up a clean RPM bdb (Berkeley DB) database in the chroot under '$TEMP_DB_PATH'."  | tee -a "$chroot_log"
+
+# These nodes are required in the chroot for certain tools (most importantly, the NSS library initialization required by rpm)
+sudo mkdir -pv $chroot_builder_folder/{dev,proc,sys,run}
```
PawelWMS, line 63: Do we need `proc`, `sys`, and `run`? And if so, shouldn't they be created with `mknod` as well? Or is RPM content with empty directories but will fail without them?

anphel31: I don't think these are necessary for the commands we run here in the chroot. Thanks for pointing it out - I'll remove these.

toolkit/tools/pkggen/worker/create_worker_chroot.sh
PawelWMS: I guess we could run `rpm -E "%{_db_backend}"` on the host and inside the chroot, skip if they match, and perform the DB conversion if they don't or the chroot run crashes (that's my guess about what would happen, if the chroot's RPM didn't support the DB created by the host's).

PawelWMS (approved):
Just one comment and +1 to Jon's question, but in I'm fine with the changes are they are right now as well.

toolkit/tools/pkggen/worker/create_worker_chroot.sh
PawelWMS: If that's going to work, could we also migrate it to 2.0?:)

anphel31: Yes, I've implemented this optimization in the latest commit. I can port this to 2.0 as well.

anphel31: For 2.0 changes see https://github.com/microsoft/CBL-Mariner/pull/4025

PR #4046, opened 2022-10-19
Fix chroot cleanup scripts

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

In 2.0 we made the package fetchers use a chroot, we have a separate clean target for the pkg fetcher (`clean-cache`) which did not have an enforced ordering with the more general `clean-workplan`. If the package fetcher was killed halfway through it could leave behind a chroot which hasn't been cleaned up. Running `make clean` will generally run `clean-workplan` before running `clean-cache` which will try (and fail) to remove the chroot's `/proc/*` etc., and possibly put the host system into a bad state. This adds an explicit dependency so the chroot is cleaned up first.

We were also missing cleanups for the spec parsing chroot and snapshot chroots.

###### Change Log  

- Add a dependency on `clean-cache` to `clean-workplan`
- Add safeunmount.sh call for `$(rpms_snapshot_build_dir)`
- Add safeunmount.sh call for (new) `$(parse_working_dir)`

**NO**

###### Test Methodology

- Local test

### DIFF
--- a/toolkit/scripts/pkggen.mk
+++ b/toolkit/scripts/pkggen.mk
@@ -43,7 +43,7 @@ $(call create_folder,$(rpmbuilding_logs_dir))
 graph-cache: $(cached_file)
 workplan: $(graph_file)
 clean: clean-workplan clean-cache
-clean-workplan:
+clean-workplan: clean-cache
 	rm -rf $(PKGBUILD_DIR)
 	rm -rf $(LOGS_DIR)/pkggen/workplan
 clean-cache:
--- a/toolkit/scripts/pkggen.mk
+++ b/toolkit/scripts/pkggen.mk
@@ -23,6 +23,7 @@ pkggen_rpms     = $(shell find $(RPMS_DIR)/*  2>/dev/null )
 
 # Pkggen workspace
 cache_working_dir      = $(PKGBUILD_DIR)/tdnf_cache_worker
+parse_working_dir      = $(BUILD_DIR)/spec_parsing
 rpmbuilding_logs_dir   = $(LOGS_DIR)/pkggen/rpmbuilding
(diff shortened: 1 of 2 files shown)


oliviacrain (approved):

oliviacrain (approved):

toolkit/scripts/pkggen.mk
```diff
@@ -39,11 +40,11 @@ logging_command = --log-file=$(LOGS_DIR)/pkggen/workplan/$(notdir $@).log --log-
 $(call create_folder,$(LOGS_DIR)/pkggen/workplan)
 $(call create_folder,$(rpmbuilding_logs_dir))
 
-.PHONY: clean-workplan clean-cache graph-cache analyze-built-graph workplan
+.PHONY: clean-workplan clean-cache clean-spec-parse graph-cache analyze-built-graph workplan
 graph-cache: $(cached_file)
 workplan: $(graph_file)
-clean: clean-workplan clean-cache
-clean-workplan:
+clean: clean-workplan clean-cache clean-spec-parse
```
PawelWMS, line 46: Do we need this change, if `clean-workplan` already depends on the other two now?

PawelWMS (approved):
One small comment, looks good.

anphel31 (approved):
Thanks for the fix.

toolkit/scripts/pkggen.mk
dmcilvaney: Technically no, but I'm leaning towards explicitly encoding all the dependencies I can to future proof against changes (ie we change `clean-workplan` to no longer require the others to run first, I'd still like them to trigger when we run `clean`).

PR #4056, opened 2022-10-20
Addressing Numpy bugs (IDs: 37701221, 37701269)

osamaesmailmsft:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Addresses the bugs listed above, which indicate issues with numpy's %check and numpy's own tests.
Also needed to update our version of Cython to be >= 0.29.30 since that's what numpy requires.

###### Change Log  

- Updated numpy.spec 's %check section to install `hypothesis`
- Moved openblas from SPECS-EXTENDED to SPECS
- Upgraded Cython from 0.29.26 to 0.29.32

Yes. Cython is updated from 0.29.26 to 0.29.32

###### Test Methodology

Local AMD64 build, running buddy builds at the moment (AMD64: 265121, ARM64: 265122)


SPECS/numpy/generate_source_tarball.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
+set -e
+
+SRC_TARBALL="numpy-1.23.4.tar.gz"
```
nicogbg, line 5: why not using $PKG_VERSION ?

SPECS/numpy/generate_source_tarball.sh
```diff
@@ -0,0 +1,102 @@
...
+OUT_FOLDER="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+PKG_VERSION="1.23.4"
```
PawelWMS, line 7: I'd remove the hard-coded version, so the `-z` checks below make sense. This will make it clearer to the caller how to use this script and we limit the risk of setting inconsistent arguments (for instance setting the source tarball but forgetting to set the package version).

SPECS/numpy/numpy.spec
```diff
@@ -1,131 +1,856 @@
...
+# Everything is BSD except for class SafeEval in numpy/lib/utils.py which is Python
+License:        BSD AND Python AND ASL 2.0
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
-Group:          Development/Languages/Python
-URL:            https://numpy.org/
-Source0:        https://github.com/numpy/numpy/releases/download/v%{version}/%{name}-%{version}.tar.gz
+URL:            http://www.numpy.org/
+Source0:        https://github.com/%{name}/%{name}/releases/download/v%{version}/%{name}-%{version}.tar.gz
+Source1:        https://numpy.org/doc/1.23/numpy-html.zip
```
nicogbg, line 17: I'd recommend to add 1.23 in zip file name so blob store and have several versions of this guy (this will solve issue that can happen if we move to version 1.24.x later, e.g.: 1.24.x for Mariner 3.0 and still on 1.23.x for Mariner 2.0)
You might also want to automate downloading of that zip file in generate_source_tarball.sh

nicogbg: e.g.: "Source1:  https://numpy.org/doc/1.23/numpy-html.zip#/numpy-html-1.23.zip"
fyi, RPM build only care about what's after the last '/'

nicogbg (changes requested):

SPECS/numpy/generate_source_tarball.sh
osamaesmailmsft: I was just using the script as given, but am cleaning up the parameters as noted.

SPECS/Cython/Cython.spec
```diff
@@ -1,11 +1,13 @@
...
+Cython is an optimising static compiler for both the Python programming language and the extended Cython programming language (baded on Pyrex). It makes writing C extensions for Python as easy as Python itself.}
+Summary:        Language for writing Python extension modules
 Name:           Cython
-Version:        0.29.26
+Version:        0.29.32
 Release:        1%{?dist}
-Summary:        Language for writing Python extension modules
-Vendor:         Microsoft
-Distribution:   Mariner
 License:        ASL 2.0
```
oliviacrain, line 8: License reference: https://spdx.org/licenses/Apache-2.0.html

Since we're moving to SPDX short identifiers for licenses, mind making the following change?

```suggestion
License:        Apache-2.0
```

toolkit/Makefile
```diff
@@ -120,6 +122,21 @@ ifeq ($(USE_PREVIEW_REPO),y)
...
+      override REPO_LIST += $(EXTENDED_REPO)
+   else
+      $(warning )
+      $(warning ######################### WARNING #########################)
+      $(warning 'USE_EXTENDED_REPO=y' set but '$(EXTENDED_REPO)' is missing. Regenerate toolkit's 'repos' directory. Remove 'USE_EXTENDED_REPO' for core builds.)
+      $(warning ######################### WARNING #########################)
+      $(warning )
+   endif
+endif
+
```
oliviacrain, line 139: These changes look unrelated to this PR. Also, GitHub is saying the PR "wants to merge 197 commits into microsoft:main". Can you rebase your changes on top of the latest main branch revision?

oliviacrain (dismissed):

toolkit/Makefile
osamaesmailmsft: They're not directly related to this PR, but building openblas takes a few hours, so I used this addition to the Makefile to skip that. I left it in this PR because I thought it might be useful for anyone else using dependencies from extended that take a long time to build. Working on the rebasing.

oliviacrain: I really like the change! We should separate it out from the packaging changes. One half of this PR is packaging bugfixes, the other half is a toolkit change to improve developer experience. Separate PRs allow us to have a more nuanced and focused conversation about each topic.

osamaesmailmsft: Gotcha! I'll get on that then :)

nicogbg (approved):

nicogbg (approved):

SPECS/numpy/numpy.spec
```diff
@@ -1,131 +1,857 @@
-Summary:        Array processing for numbers, strings, records, and objects
+#uncomment next line for a release candidate or a beta
+#%%global relc rc1
```
PawelWMS, line 2: We usually don't want to use these releases:
```suggestion
```

SPECS/numpy/numpy.spec
```diff
@@ -1,131 +1,857 @@
...
+BuildRequires:  python3-Cython
 BuildRequires:  python3-devel
-BuildRequires:  python3-setuptools
-BuildRequires:  python3-Cython >= 0.29.24
+BuildRequires:  python3-setuptools < 60
+Provides:       libnpymath-static = %{version}-%{release}
+Provides:       libnpymath-static%{?_isa} = %{version}-%{release}
+Provides:       numpy = %{version}-%{release}
+Provides:       numpy%{?_isa} = %{version}-%{release}
+Obsoletes:      numpy < 1.10.1-3
```
PawelWMS, line 51: Not needed - Mariner never provided this package.
```suggestion
```

SPECS/numpy/numpy.spec
```diff
@@ -1,131 +1,857 @@
...
 Summary:        f2py for numpy
+License:        BSD AND Python AND ASL 2.0
 Requires:       python3-devel
-Requires:       python3-numpy = %{version}-%{release}
+Requires:       python3-numpy%{?_isa} = %{version}-%{release}
 Provides:       python3-f2py = %{version}-%{release}
+Obsoletes:      python3-f2py <= 2.45.241_1927
+Provides:       f2py = %{version}-%{release}
+Provides:       numpy-f2py = %{version}-%{release}
+Obsoletes:      numpy-f2py < 1:1.10.1-3
```
PawelWMS, line 78: Also not needed `Obsoletes` - we never had these packages anyway.

A bit of Mariner trivia: what's probably more important, we've "reset" the epoch numbers on almost all of our packages, so `numpy-f2py < 1:1.10.1-3` would be even problematic, if we ever introduced it as a separate package.

```suggestion
Provides:       f2py = %{version}-%{release}
Provides:       numpy-f2py = %{version}-%{release}
```

SPECS/numpy/numpy.spec
```diff
@@ -1,131 +1,857 @@
...
+
+#%%{__python3} setup.py install -O1 --skip-build --root %%{buildroot}
+# skip-build currently broken, this works around it for now
+env OPENBLAS=%{_libdir} \
+    FFTW=%{_libdir} BLAS=%{_libdir} \
+    LAPACK=%{_libdir} CFLAGS="%{optflags}" \
+    %{__python3} setup.py install --root %{buildroot} --prefix=%{_prefix}
+pushd %{buildroot}%{_bindir} &> /dev/null
+ln -s f2py3 f2py.numpy
+popd &> /dev/null
```
PawelWMS, line 128: Nit-pick: this pushd-popd seems a bit excessive for what we're doing.
```suggestion
ln -s f2py3 %{buildroot}%{_bindir}/f2py.numpy
```

SPECS/numpy/numpy.spec
```diff
@@ -1,131 +1,857 @@
...
+pip install pytest hypothesis typing-extensions
+# test_ppc64_ibm_double_double128 is unnecessary now that ppc64le has switched long doubles to IEEE format.
+# https://github.com/numpy/numpy/issues/21094
+%ifarch %{ix86}
+# Weird RuntimeWarnings on i686, similar to https://github.com/numpy/numpy/issues/13173
+# Some tests also overflow on 32bit
+%global ix86_k and not test_vector_matrix_values and not test_matrix_vector_values and not test_identityless_reduction_huge_array and not (TestKind and test_all)
+%endif
+python3 runtests.py --no-build -- -ra -k 'not test_ppc64_ibm_double_double128 %{?ix86_k}'
+%endif
```
PawelWMS, line 154: We skip the `%check` section anyway, if `with_check` is set to 0, so we can simplify:
```suggestion
export PYTHONPATH=%{buildroot}%{python3_sitearch}
pip install pytest hypothesis typing-extensions
# test_ppc64_ibm_double_double128 is unnecessary now that ppc64le has switched long doubles to IEEE format.
# https://github.com/numpy/numpy/issues/21094
%ifarch %{ix86}
# Weird RuntimeWarnings on i686, similar to https://github.com/numpy/numpy/issues/13173
# Some tests also overflow on 32bit
%global ix86_k and not test_vector_matrix_values and not test_matrix_vector_values and not test_identityless_reduction_huge_array and not (TestKind and test_all)
%endif
python3 runtests.py --no-build -- -ra -k 'not test_ppc64_ibm_double_double128 %{?ix86_k}'
```

SPECS/numpy/numpy.spec
```diff
@@ -1,131 +1,857 @@
...
+
 
 %check
-pip3 install nose pytest
-mkdir -pv test
-cd test
-PYTHONPATH=%{buildroot}%{python3_sitelib} PATH=$PATH:%{buildroot}%{_bindir} %python3 -c "import numpy; numpy.test()"
+%if %{with_check}
+export PYTHONPATH=%{buildroot}%{python3_sitearch}
+pip install pytest hypothesis typing-extensions
```
PawelWMS, line 145: We've had some issue with `pytest` breaking our builds, if allowed to install the latest version like this. Please take a look at the version currently downloaded for these tests and "freeze" it here like so:
```suggestion
pip install pytest==[some_version_number] hypothesis typing-extensions
```

SPECS/numpy/numpy.spec
```diff
@@ -1,131 +1,857 @@
...
+# test_ppc64_ibm_double_double128 is unnecessary now that ppc64le has switched long doubles to IEEE format.
+# https://github.com/numpy/numpy/issues/21094
+%ifarch %{ix86}
+# Weird RuntimeWarnings on i686, similar to https://github.com/numpy/numpy/issues/13173
+# Some tests also overflow on 32bit
+%global ix86_k and not test_vector_matrix_values and not test_matrix_vector_values and not test_identityless_reduction_huge_array and not (TestKind and test_all)
+%endif
```
PawelWMS, line 152: Can be dropped - we don't build for these architectures and I don't think we will.
```suggestion
```

SPECS/numpy/numpy.spec
```diff
@@ -1,131 +1,857 @@
...
 
 %changelog
-* Thu Sep 15 2022 Riken Maharjan <rmaharjan@microsoft.com> - 1.22.3-2
-- Replaced py3_install macro to include numpy header files. --skip-build arg is removed essentially with the py3_install macro removal.
+* Tue Nov 15 2022 Osama Esmail <osamaesmail@microsoft.com> - 1.23.4-2
+- Initial CBL-Mariner import from Fedora 37 (license: MIT)
+- License verified.
+- Created patch to address test issues in numpy PR-22487
+- Replaced %%{with test} with %%{with_check}
+- Using pip install for tests
```
PawelWMS, line 203: We could also mention that we reset the epoch to indicate it was intentional.

SPECS/openblas/openblas.spec
```diff
@@ -15,7 +15,7 @@
 
 Name:           openblas
 Version:        0.3.21
-Release:        1%{?dist}
+Release:        2%{?dist}
```
PawelWMS, line 18: For extended -> core moves we must also clean-up the specs a bit, because the bar is lower for the extended specs. Please lint the spec and remove references to distro-specific macros (for instance `rhel` or `fedora`) or things that do not matter for Mariner (like unsupported architectures).

PawelWMS (changes requested):
A few clean-up requests before we merge.

SPECS/openblas/openblas.spec
```diff
@@ -1,7 +1,31 @@
...
+# LAPACK
+%if %{with system_lapack}
+# Do we have LAPACKE? (Needs at least lapack 3.4.0)
+%global lapacke 1
+BuildRequires:  lapack-static
+%else
+# Use bundled LAPACK
+%global lapacke 1
+Provides:       bundled(lapack) = %{lapackver}
+%endif
```
PawelWMS, line 27: Please move this conditional `BuildRequires` and `Provides` below, next to the other ones.

SPECS/openblas/openblas.spec
```diff
@@ -1,7 +1,31 @@
...
+Computational Science, ISCAS. http://www.rdcps.ac.cn
+%bcond_with system_lapack
+# LAPACK
+%if %{with system_lapack}
+# Do we have LAPACKE? (Needs at least lapack 3.4.0)
+%global lapacke 1
+BuildRequires:  lapack-static
+%else
+# Use bundled LAPACK
+%global lapacke 1
```
PawelWMS, line 25: It looks like we're setting `lapacke` to 1 in both branches of the `if`, so I'd double-check if that's correct and if so, just remove the macro and assume it's always set to 1 to clean-up its uses in the spec.

SPECS/openblas/openblas.spec
```diff
@@ -254,7 +230,7 @@ rm -rf lapack-netlib
 cd ..
 cp -ar OpenBLAS-%{version} openmp
 cp -ar OpenBLAS-%{version} threaded
-%if %build64
+%if %{build64}
```
PawelWMS, line 233: To be honest, we can also assume that we're always building for a 64-bit architecture and remove the `build64` macro completely. Not a blocker, completely ignorable for this PR, though.

SPECS/openblas/openblas.spec
```diff
@@ -422,17 +384,16 @@ make -C openmp64_   $TARGET USE_THREAD=1 USE_OPENMP=1 FC=gfortran CC=gcc COMMON_
...
+make -C serial USE_THREAD=0 PREFIX=%{buildroot} OPENBLAS_LIBRARY_DIR=%{buildroot}%{_libdir} OPENBLAS_INCLUDE_DIR=%{buildroot}%{_includedir}/%{name} OPENBLAS_BINARY_DIR=%{buildroot}%{_bindir} OPENBLAS_CMAKE_DIR=%{buildroot}%{_libdir}/cmake install
 
 # Copy lapacke include files
 %if %{with system_lapack} && %{lapacke}
 cp -a %{_includedir}/lapacke %{buildroot}%{_includedir}/%{name}
 %endif
 
 # Fix i686-x86_64 multilib difference
-%multilib_fix_c_header --file %{_includedir}/openblas/openblas_config.h
+%{multilib_fix_c_header} --file %{_includedir}/openblas/openblas_config.h
```
PawelWMS, line 396: This line change should be reverted. It's one of the remaining issues with out linter: it doesn't differentiate between macros, that are just simple values/strings, and ones that are functions accepting arguments. Surrounding the latter in curly brackets makes them being run with no arguments and thing end up breaking or misconfigured. Having said that, I don't see this macro even being defined anywhere, so it might be worth taking a look at the build logs, but for this review I'd just need to remove the brackets.

```suggestion
%multilib_fix_c_header --file %{_includedir}/openblas/openblas_config.h
```

PawelWMS (approved):
Left a few short comments but pre-approving to not block the PR once they are addressed.

PR #4124, opened 2022-10-28
Add RPM package dependencies to CDI components

ameyapu:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR uses the Requires directive of RPM specification to enforce dependent RPM packages that will form part of the final components from which golden containers are built.  See [bug 3349](https://dev.azure.com/mariner-org/ECF/_workitems/edit/3349) for more details.

###### Change Log  

- Added dependent RPMs using the Requires directive
- Included build of binaries from tools directory and into their corresponding sub-components
- Updated release version and change log

**NO**

###### Associated issues  

- Link to successful buddy build is [here](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=257151&view=results).

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: [Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=264257&view=results)

### DIFF
--- a/SPECS/containerized-data-importer/containerized-data-importer.spec
+++ b/SPECS/containerized-data-importer/containerized-data-importer.spec
@@ -19,7 +19,7 @@ Summary:        Container native virtualization
 Name:           containerized-data-importer
 Version:        1.51.0
 Release:        4%{?dist}
-License:        ASL 2.0
+License:        Apache-2.0
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 Group:          System/Packages
@@ -32,22 +32,6 @@ BuildRequires:  pkgconfig
 BuildRequires:  rsync
 BuildRequires:  sed
 ExclusiveArch:  x86_64 aarch64
-Requires:       cpio
-Requires:       curl
-Requires:       diffutils
-Requires:       findutils
-Requires:       jq
-Requires:       libnbd
-Requires:       nbdkit
-Requires:       patch
-Requires:       rsync
-Requires:       librsync
-Requires:       qemu-tools
-Requires:       qemu-img
-Requires:       shadow-utils
-Requires:       tar
-Requires:       unzip
-Requires:       util-linux
 Provides:       cdi = %{version}-%{release}
 
 %description
@@ -56,6 +40,22 @@ Containerized-Data-Importer (CDI) is a persistent storage management add-on for 
 %package        api
 Summary:        CDI API server
 Group:          System/Packages
+Requires:       cpio
+Requires:       curl
+Requires:       diffutils
+Requires:       findutils
+Requires:       jq
+Requires:       libnbd
+Requires:       librsync
+Requires:       nbdkit
+Requires:       patch
+Requires:       rsync
+Requires:       qemu-tools
+Requires:       qemu-img
+Requires:       shadow-utils
+Requires:       tar
+Requires:       unzip
+Requires:       util-linux
 Provides:       cdi-apiserver = %{version}-%{release}
 
 %description    api
@@ -64,6 +64,22 @@ The containerized-data-importer-api package provides the kubernetes API extensio
 %package        cloner
 Summary:        Cloner for host assisted cloning
 Group:          System/Packages
+Requires:       cpio
+Requires:       curl
+Requires:       diffutils
+Requires:       findutils
+Requires:       jq
+Requires:       libnbd
+Requires:       librsync
+Requires:       nbdkit
+Requires:       patch
+Requires:       rsync
+Requires:       qemu-tools
+Requires:       qemu-img
+Requires:       shadow-utils
+Requires:       tar
+Requires:       unzip
+Requires:       util-linux
 
 %description    cloner
 Source and Target cloner image for host assisted cloning
@@ -71,6 +87,22 @@ Source and Target cloner image for host assisted cloning
 %package        controller
 Summary:        Controller for the data fetching service
 Group:          System/Packages
+Requires:       cpio
+Requires:       curl
+Requires:       diffutils
+Requires:       findutils
+Requires:       jq
+Requires:       libnbd
+Requires:       librsync
+Requires:       nbdkit
+Requires:       patch
+Requires:       rsync
+Requires:       qemu-tools
+Requires:       qemu-img
+Requires:       shadow-utils
+Requires:       tar
+Requires:       unzip
+Requires:       util-linux
 
 %description    controller
 Controller for the data fetching service for VM container images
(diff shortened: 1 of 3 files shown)


SPECS/containerized-data-importer/containerized-data-importer.spec
```diff
@@ -32,6 +32,22 @@ BuildRequires:  pkgconfig
...
+Requires:       jq
+Requires:       libnbd
+Requires:       nbdkit
+Requires:       patch
+Requires:       rsync
+Requires:       librsync
+Requires:       qemu-tools
+Requires:       qemu-img
+Requires:       shadow-utils
+Requires:       tar
```
romoh, line 48: Are we sure those are required dependencies? I highly doubt it.
If you attempt running the binary without them, does it not work?

SPECS/containerized-data-importer/containerized-data-importer.spec
```diff
@@ -172,11 +197,15 @@ install -m 0644 _out/manifests/release/cdi-cr.yaml %{buildroot}%{_datadir}/cdi/m
 %license LICENSE
 %doc README.md
 %{_bindir}/virt-cdi-importer
+%{_bindir}/tools/cdi-containerimage-server
+%{_bindir}/tools/cdi-image-size-detection
+%{_bindir}/tools/cdi-source-update-poller
```
oliviacrain, line 202: Is there a reason we're putting these in a unique `tools` folder, and not putting them in `/usr/bin` directly? Do we not want them to be in the default `$PATH`? As far as I know, we don't currently ship subdirectories under `/usr/bin`, and `/usr/bin/tools` doesn't really mean anything, semantically.

For what it's worth- [openSUSE just puts these under `/usr/bin`.](https://build.opensuse.org/package/view_file/openSUSE:Factory/containerized-data-importer/containerized-data-importer.spec?expand=1)

SPECS/containerized-data-importer/containerized-data-importer.spec
```diff
@@ -18,7 +18,7 @@
 Summary:        Container native virtualization
 Name:           containerized-data-importer
 Version:        1.51.0
-Release:        3%{?dist}
+Release:        4%{?dist}
 License:        ASL 2.0
```
oliviacrain, line 22: Small nit- recently, we decided to start using SPDX short license identifiers rather than Fedora's short names. Can we update this to be [`Apache-2.0`](https://spdx.org/licenses/Apache-2.0.html), instead?

```suggestion
License:        Apache-2.0
```

SPECS/containerized-data-importer/containerized-data-importer.spec
oliviacrain, line 48: There's no `%files` section for the base package, which means we don't actually produce an RPM for the `containerized-data-importer` package.  [You can confirm this by looking at PMC](https://packages.microsoft.com/cbl-mariner/2.0/prod/base/x86_64/).

These requirements, then, are being attached to a non-existent package. Effectively, there's no change here. These requirements should go under the subpackages that actually require them.

oliviacrain (changes requested):
GitHub is making me put something in this comment field- see above comments, please!

SPECS/containerized-data-importer/containerized-data-importer.spec
ameyapu: This is the [RPM tree](https://github.com/kubevirt/containerized-data-importer/blob/main/rpm/BUILD.bazel) from CDI sources.  Containers, such as cdi-importer and cdi-uploadserver, call sed, tar or qemu-img commands internally in some code paths.  During the initial unit testing of golden containers for platform helm charts, these dependencies got detected during experimentation and included in the docker file.

There are other executables, such as readline, tzdata, gawk, which are seen in the RPM tree hierarchy link above.  We have not hit those errors as yet.  Should we include each and every package in the tree under the Requires directive?

SPECS/containerized-data-importer/containerized-data-importer.spec
ameyapu: If you see [this file](https://github.com/kubevirt/containerized-data-importer/blob/main/cmd/cdi-importer/BUILD.bazel), the cdi-containerimage-server is under tools directory.  But [this source file](https://github.com/kubevirt/containerized-data-importer/blob/46c6aa994aec9267532e56d84dfa4deca5d0aefe/pkg/controller/import-controller.go) expects it in the /usr/bin/ directory.  I followed the former and included in the tools.  But you could be right here - it is prudent to include under /usr/bin.  I'll change that.

SPECS/containerized-data-importer/containerized-data-importer.spec
ameyapu: Ok, will be done!

SPECS/containerized-data-importer/containerized-data-importer.spec
ameyapu: Let me check on this and revert.

ameyapu:
@oliviacrain - Few changes have been done per review feedback.  We will revisit this PR after merging [nbdkit PR](https://github.com/microsoft/CBL-Mariner/pull/4163).

SPECS/containerized-data-importer/containerized-data-importer.spec
ameyapu: @romoh - With the nbdkit packages moved to core, it is time to review this PR.  Please help address above queries.

ameyapu:
> @oliviacrain / @romoh - Some changes have been done per review feedback. We will revisit this PR after merging [nbdkit PR](https://github.com/microsoft/CBL-Mariner/pull/4163).

@oliviacrain - With PR #4163 merged, it is time to revisit.  Request to review the changes.

SPECS/containerized-data-importer/containerized-data-importer.spec
oliviacrain: Another question- Is the RPM tree hierarchy reflecting just the tools needed, or does it also include the full closure of dependencies required for the necessary tools? If it's the second option, then we can trim down these lists to just the tools needed and let our RPM requirements handle the dependencies of those tools.

ameyapu: > Another question- Is the RPM tree hierarchy reflecting just the tools needed, or does it also include the full closure of dependencies required for the necessary tools? If it's the second option, then we can trim down these lists to just the tools needed and let our RPM requirements handle the dependencies of those tools.

It is the latter.  The RPM tree lists all the libraries, tools and other CLIs.  In the initial check-in for CDI golden containers, I manually sifted through the test failures to figure out reasons for failures.  As there was no data volume import test, we slipped on the nbdkit-server dependency.

oliviacrain: I think I found where they keep the list of the tools that are actually required, which they use to generate the RPM tree:

https://github.com/kubevirt/containerized-data-importer/blob/main/hack/build/rpm-deps.sh

Mind trimming down these lists to just what's required in this document?

ameyapu: This requires two more packages - crypto-policies and codespell - which are absent in Mariner's set of packages.  These two packages also form the base for all the CDI binaries.

ameyapu: And there are further sub-dependencies from crypto-policies namely - python3-flake8, java-1.8.0-openjdk-devel and python3-pylint.  So, this is beginning to explode in scope as I gather more and more packages here.

PR #4181, opened 2022-11-08
Fix missing dependency tracking for spec sources, refactor Makefile to skip shell commands during dry-run

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

The `input-srpms` build target has not been tracking changes to miscellaneous source files in the `./SPECS/pkg/*` directories. In Mariner 1.0 we tracked these files with https://github.com/microsoft/CBL-Mariner/blob/e72e08fdf005802a4dec67c5fc9906df954b5e36/toolkit/scripts/srpm_pack.mk#L18 which was used as a dependency of srpm packing; however, this was lost in the refactor for 2.0.

Unfortunately adding this dependency tracking back has the unintended side effect of breaking Make's `--dry-run` behavior. This is used when tab completing on the command line among other things. The call to `$(shell find $(local_spec_dirs) -type f -name '*')` takes a long time to complete and is re-run every time the tab-completion runs. 

To avoid this a new function `shell_real_build_only` was added which will only call through to `$(shell ...)` when Make is not running with `--dry-run`. 

```make
# Consider replacing this:
local_specs = $(shell find $(SPECS_DIR)/ -type f -name '*.spec')
# With this:
local_specs = $(call shell_real_build_only, find $(SPECS_DIR)/ -type f -name '*.spec')
```

To make this function available to the build-tag and specs makefile components the code was refactored so it would be parsed after `util.mk`.

Since the version variables have moved a set of variable printing functions has been added.
```bash
var=$(make -qn printvar-MY_VAR)
var=$(make -qn get-dist-tag)
```

###### Change Log  

- Restore dependency tracking of `$(local_spec_sources)` for `input-srpms` target
- Add `shell_real_build_only` function to conditionally disable calls to `$(shell ...)` during dry-runs.
- Add variable printing support to the makefile

**NO**

###### Test Methodology

- Local builds

### DIFF
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -221,6 +221,5 @@ printvar-all-vars: ;
 printvar-verbose-all-vars: $(foreach var,$(interesting_variables),printvar-verbose-$(var))
 printvar-verbose-all-vars: ;
 
-#$(print_targets): ; $(info $($(subst print-,,$@)))
 printvar-%: ; $(info $($(subst printvar-,,$@)))
 printvar-verbose-%: ; $(info $(subst printvar-verbose-,,$@): $($(subst printvar-verbose-,,$@)))
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -192,6 +192,35 @@ clean:
 	rm -rf $(BUILD_DIR)
 	rm -rf $(toolkit_root)/out
 
-# output version number
-get-version:
-	@echo $(RELEASE_VERSION)
+######## VARIABLE PRINTING ########
+
+# Some common version information that is useful to gather. Generally should be run with the Make flag --quiet
+get-version: printvar-RELEASE_VERSION
+get-dist-tag: printvar-DIST_TAG
+get-release-major: printvar-RELEASE_MAJOR_ID
+
+# Make an easy way to print out the build variables. These must be the last entries in the makefile so that all other
+# files have their variables inluded
+
+# Print out all variables to stdout, either or of the form "<VALUE>" or the verbose form "varname: <VALUE>"
+#   printvar-all-vars
+#   printvar-verbose-all-vars
+
+# Print a specific variable to stdout, using the same format as above
+#   printvar-*
+#   printvar-verbose-*
+
+# Use these targets like so:  `my-var=$(make printvar-MY_VAR --quiet)`
+# The --quiet flag is important to avoid printing extra output
+.PHONY: printvar-all-vars printvar-verbose-all-vars
+
+# Gather the variables we want to print out, removing any automatic .* variables, and the self reference
+interesting_variables = $(filter-out .% interesting_variables, $(sort $(.VARIABLES)))
+printvar-all-vars: $(foreach var,$(interesting_variables),printvar-$(var))
+printvar-all-vars: ;
+printvar-verbose-all-vars: $(foreach var,$(interesting_variables),printvar-verbose-$(var))
+printvar-verbose-all-vars: ;
+
+#$(print_targets): ; $(info $($(subst print-,,$@)))
+printvar-%: ; $(info $($(subst printvar-,,$@)))
+printvar-verbose-%: ; $(info $(subst printvar-verbose-,,$@): $($(subst printvar-verbose-,,$@)))
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -124,38 +124,9 @@ ifeq ($(USE_PREVIEW_REPO),y)
    endif
 endif
 
-ifneq ($(strip $(SRPM_PACK_LIST)),)
-LOCAL_SPECS = $(wildcard $(addprefix $(SPECS_DIR)/*/,$(addsuffix .spec,$(strip SRPM_PACK_LIST))))
-else # Empty pack list, build all under $(SPECS_DIR)
-LOCAL_SPECS = $(shell find $(SPECS_DIR)/ -type f -name '*.spec')
-endif
-
-LOCAL_SPEC_DIRS = $(foreach spec,$(LOCAL_SPECS),$(dir $(spec)))
-
 CA_CERT     ?=
 TLS_CERT    ?=
 TLS_KEY     ?=
-
-# Build defines
-DIST_TAG           ?= .cm2
-# Running 'git' as the owner of the repo, so it doesn't complain about the repo not belonging to root.
-BUILD_NUMBER       ?= $(shell runuser -u $$(stat -c "%U" $(PROJECT_ROOT)) -- git rev-parse --short HEAD)
-# an empty BUILD_NUMBER breaks the build later on
-ifeq ($(BUILD_NUMBER),)
-   BUILD_NUMBER = non-git
-endif
-RELEASE_MAJOR_ID   ?= 2.0
-# use minor ID defined in file (if exist) otherwise define it
-# note this file must be single line
-ifneq ($(wildcard $(OUT_DIR)/version-minor-id.config),)
-   RELEASE_MINOR_ID ?= .$(shell cat $(OUT_DIR)/version-minor-id.config)
-else
-   RELEASE_MINOR_ID ?= .$(shell date +'%Y%m%d.%H%M')
-endif
-RELEASE_VERSION    ?= $(RELEASE_MAJOR_ID)$(RELEASE_MINOR_ID)
-
-# Image tag - empty by default. Does not apply to the initrd.
-IMAGE_TAG          ?=
 
 # panic,fatal,error,warn,info,debug,trace
 LOG_LEVEL          ?= info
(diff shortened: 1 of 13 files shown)


toolkit/Makefile
```diff
@@ -120,39 +120,10 @@ ifeq ($(USE_PREVIEW_REPO),y)
...
-ifneq ($(wildcard $(OUT_DIR)/version-minor-id.config),)
-   RELEASE_MINOR_ID ?= .$(shell cat $(OUT_DIR)/version-minor-id.config)
-else
-   RELEASE_MINOR_ID ?= .$(shell date +'%Y%m%d.%H%M')
-endif
-RELEASE_VERSION    ?= $(RELEASE_MAJOR_ID)$(RELEASE_MINOR_ID)
-
-# Image tag - empty by default. Does not apply to the initrd.
-IMAGE_TAG          ?=
-
```
PawelWMS, line 155: I'd be careful about moving these variables around, as we already have some scripts depending on them being here. From the ones I was able to find so far:

1. This repo:

- `toolkit/scripts/rpmops.sh`

2. Our pipelines: look for `grep.*Makefile` across all files.

It seems like we're currently extracting `DIST_TAG` and `RELEASE_MAJOR_ID`. The problem in my opinion is on the side of how these variables are extracted, so if we can add `make` targets, which return these values, are quick and don't produce any extra garbage, then that would be the ideal way forward.

toolkit/scripts/srpm_pack.mk
```diff
@@ -19,6 +19,14 @@ SRPM_BUILD_LOGS_DIR = $(LOGS_DIR)/pkggen/srpms
 toolchain_spec_list = $(toolchain_build_dir)/toolchain_specs.txt
 srpm_pack_list_file = $(BUILD_SRPMS_DIR)/pack_list.txt
 
+ifneq ($(strip $(SRPM_PACK_LIST)),)
```
PawelWMS, line 22: I'm curious: is there a way to re-assign `SRPM_PACK_LIST` with its stripped version, so we don't have to repeat this everywhere? Not a blocker for this PR, this wasn't introduced by you here anyway, but if that is possible we might as well clean-up a little.

toolkit/scripts/utils.mk
```diff
@@ -20,9 +20,25 @@ build_arch := $(shell uname -m)
...
+# which is the short form of --dry-run.
+#
+# $1 - The full command to run, if we are not doing --dry-run
+ifeq (n,$(findstring n,$(firstword $(MAKEFLAGS))))
+define shell_real_build_only
+endef
+else # ifeq (n,$(findstring...
+define shell_real_build_only
+$(shell $1)
+endef
```
PawelWMS, line 37: Since it's so short, what do you think about this:
```suggestion
shell_real_build_only =
else # ifeq (n,$(findstring...
shell_real_build_only = $(shell $1)
endef
```

toolkit/scripts/utils.mk
```diff
@@ -20,9 +20,25 @@ build_arch := $(shell uname -m)
...
+endif # ifeq (n,$(findstring...
+
+myvar = $(call shell_real_build_only, ls)
+
```
PawelWMS, line 41: Debugging leftover?
```suggestion
```

PawelWMS (changes requested):
I really like this change. Just a few small comments and I think we definitely need to address the move of a few variables to a separate file before we merge to not break other tooling and our builds.

toolkit/Makefile
dmcilvaney: Will do, I'll tinker around and see what I can figure out.

toolkit/scripts/toolkit.mk
```diff
@@ -88,7 +88,7 @@ rpms-snapshot: $(rpms_snapshot)
 $(rpms_snapshot): $(rpms_snapshot_per_specs) $(depend_SPECS_DIR)
 	cp $(rpms_snapshot_per_specs) $(rpms_snapshot)
 
-$(rpms_snapshot_per_specs): $(go-rpmssnapshot) $(chroot_worker) $(LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(SPECS_DIR)
+$(rpms_snapshot_per_specs): $(go-rpmssnapshot) $(chroot_worker) $(local_specs) $(local_spec_dirs) $(SPECS_DIR)
```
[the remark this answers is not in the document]
dmcilvaney: That is a bit of a grey area for the makefiles, our docs callout ALL_CAPS as user configurable variables while lower_case are internal only variables... but generally the internal only variables are all localized to a single file as well.

toolkit/scripts/utils.mk
dmcilvaney: That was my first implementation as well, unfortunately Make doesn't parse it correctly. The ordering of the defines and if/else are important, although I'm not 100% sure what the exact issue is.

PawelWMS: Not sure, if we're looking at the same thing. My suggestion is not about ordering but about replacing `define...endef` with a one-liner.  However, if it cannot easily be done in a single line, then there's no point to waste time on this.

toolkit/Makefile
```diff
@@ -214,6 +188,35 @@ clean:
...
+.PHONY: printvar-all-vars printvar-verbose-all-vars
+
+# Gather the variables we want to print out, removing any automatic .* variables, and the self reference
+interesting_variables = $(filter-out .% interesting_variables, $(sort $(.VARIABLES)))
+printvar-all-vars: $(foreach var,$(interesting_variables),printvar-$(var))
+printvar-all-vars: ;
+printvar-verbose-all-vars: $(foreach var,$(interesting_variables),printvar-verbose-$(var))
+printvar-verbose-all-vars: ;
+
+#$(print_targets): ; $(info $($(subst print-,,$@)))
```
dmcilvaney, line 220: Remove commented code
```suggestion
```

dmcilvaney:
This PR has corresponding changes to the ADO pipelines (PR #10813) to support both the new layout of the dist/release tags, and the legacy locations.

toolkit/scripts/utils.mk
dmcilvaney: oh, I see, got confused by the diff a bit.

I don't think there is a way to get around using the define operator, since `$(call` is what populates the `$1` arg and I want this to be a drop-in replacement for a call to `$(shell <arg>)`.

oliviacrain (approved):

toolkit/scripts/srpm_pack.mk
```diff
@@ -19,13 +19,19 @@ SRPM_BUILD_LOGS_DIR = $(LOGS_DIR)/pkggen/srpms
 toolchain_spec_list = $(toolchain_build_dir)/toolchain_specs.txt
 srpm_pack_list_file = $(BUILD_SRPMS_DIR)/pack_list.txt
 
-ifneq ($(strip $(SRPM_PACK_LIST)),)
+# Configure the list of packages we want to process into SRPMs
+custom_srpm_pack_list = $(strip $(SRPM_PACK_LIST))
```
PawelWMS, line 23: Why not use "[override](https://www.gnu.org/software/make/manual/make.html#Override-Directive)" like we did with `REPO_LIST`? The word "custom" seems a bit misleading, like it's not related to the original `SRPM_PACK_LIST` or we modified to a greater degree than by simply stripping the whitespaces.
```suggestion
override SRPM_PACK_LIST = $(strip $(SRPM_PACK_LIST))
```

toolkit/Makefile
PawelWMS: I think we have a fix for that now in our pipelines, unblocking on this comment.

toolkit/scripts/toolkit.mk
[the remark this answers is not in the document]
PawelWMS: I see, you're right. Nevermind then, looks like there's no simple perfect solution here right now.;)

PawelWMS (approved):
Discussed remaining comments offline, unblocking.

toolkit/scripts/srpm_pack.mk
dmcilvaney: Switched to using override. Needed to be `override SRPM_PACK_LIST := ...` to avoid circular definition of the variable.

dmcilvaney:
As an FYI, I have a pr out that touches the pack list: https://github.com/microsoft/CBL-Mariner/pull/4181/files#diff-71dbe794406e4f45d19637b785e46e1b1806fffbe71ae6a780f7fa6cb944f481

Feel free to go ahead with this, I'm still waiting to figure out when we want to merge my PRs in and will adjust as needed.

dmcilvaney:
Small rebase to include recently added call to $(shell, also missed one before which I've added

PR #4210, opened 2022-11-11
Move toolchain RPMs to a dedicated location in ./build/toolchain_rpms

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Toolchain RPMs located in either `./out/RPMS` or `./build/rpm_cache` would often be inadvertently updated during other operations (generally the fetchers seem to be the major cause of this). These files are effectively static, so it does not make sense to invalidate large parts of the build due to their timestamps being updated.

If they are updated, it causes the worker chroot .tar.gz to be rebuilt, which in turn causes the majority of the build to rerun. 

Moving the toolchain RPMs into a dedicated folder allows much better management of their lifetime and state. It also allows us to prioritize using the toolchain RPMs over other sources of packages during fetching more reliably.

These changes will mean that most basic build operations will no longer cause undesirable rebuilds of components, and will only build the parts that have actually changed.

###### Change Log  

- Rename `reserved-file-list-file`, `ReadReservedFilesList` and associated flags/functions for better clarity of their intended use
- Add dedicated directory for toolchain RPMs
  - Swap dynamic `$(toolchain_rpms_dir)` for static `$(TOOLCHAIN_RPMS_DIR)` in makefiles
  - toolchain will place RPMs into `./build/toolchain_rpms`, and if `REBUILD_TOOLCHAIN=y` it will also place a copy into `./out/RPMS/`.
  - `specreader` will flag toolchain packages as such, and update their rpm paths to point to the new location (so the 'is prebuilt' checks can find the .rpms).
  - Fetcher tools (`rpmrepocloner`) gain an optional `toolchain-repo` local repo which can point to the local toolchain directory. It is the preferred source of packages and will be checked before any other source.
  - `pkgworker` may now install RPMs from the toolchain directory

**YES**

###### Test Methodology

- Local builds, will need additional testing before merge after rebasing on previous


dmcilvaney:
I would recommend splitting the commits when reviewing, the first one is just a rename while the actual logic occurs in the second commit.

oliviacrain (approved):

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -41,6 +42,7 @@ var (
 	repoFile             = app.Flag("repo-file", "Full path to local.repo").Required().ExistingFile()
 	rpmsDirPath          = app.Flag("rpm-dir", "The directory to use as the local repo and to submit RPM packages to").Required().ExistingDir()
 	srpmsDirPath         = app.Flag("srpm-dir", "The output directory for source RPM packages").Required().String()
+	toolchainDirPath     = app.Flag("toolchain-rpm-dir", "Directory that contains already built toolchain RPMs. Should contain top level directories for architecture.").Required().ExistingDir()
```
MateuszMalisz, line 45: nit: consider changing to "directories for architectures" or "directories per architecture".

toolkit/tools/scheduler/scheduler.go
```diff
@@ -69,7 +70,7 @@ var (
 	noCleanup            = app.Flag("no-cleanup", "Whether or not to delete the chroot folder after the build is done").Bool()
 	noCache              = app.Flag("no-cache", "Disables using prebuilt cached packages.").Bool()
 	stopOnFailure        = app.Flag("stop-on-failure", "Stop on failed build").Bool()
-	reservedFileListFile = app.Flag("reserved-file-list-file", "Path to a list of files which should not be generated during a build").ExistingFile()
+	toolchainManifest    = app.Flag("toolchain-manifest", "Path to a list of RPMs which are created by the toolchain. Will mark RPMs from this list as prebuilt.").ExistingFile()
```
MateuszMalisz, line 73: Nit: Consider using present tense instead of a future one.

toolkit/tools/scheduler/scheduler.go
```diff
@@ -120,24 +121,25 @@ func main() {
 		logger.Log.Fatalf("Unable to generate package build list, error: %s", err)
 	}
 
-	var reservedFiles []string
-	if len(reservedFileListFile) > 0 {
-		reservedFiles, err = schedulerutils.ReadReservedFilesList(reservedFileListFile)
+	var toolchainPackages []string
+	if len(toolchainManifest) > 0 {
+		toolchainPackages, err = schedulerutils.ReadReservedPackageManifest(toolchainManifest)
```
MateuszMalisz, line 126: I would consider changing the name of this function as well. The only times you use it is to process toolchain manifest, so might as well name it `ReadToolchainPackageManifest` or similar.

toolkit/tools/scheduler/schedulerutils/buildlist.go
```diff
@@ -123,3 +126,15 @@ func ReadReservedFilesList(path string) (reservedFiles []string, err error) {
 
 	return reservedFiles, nil
 }
+
+// rpmsToFind should be an absolute path to the expected RPM, while toolchainRPMs is a list of '<name>-<version>-<release>.rpm'.
+func IsToolchainRPM(rpmPath string, reservedRPMs []string) bool {
```
MateuszMalisz, line 131: All exported functions should have a documentation string before their definition.

toolkit/tools/specreader/specreader.go
```diff
@@ -63,13 +66,23 @@ func main() {
 		logger.Log.Panicf("Value in --workers must be greater than zero. Found %d", *workers)
 	}
 
-	err := parseSPECsWrapper(*buildDir, *specsDir, *rpmsDir, *srpmsDir, *distTag, *output, *workerTar, *workers, *runCheck)
+	var toolchainRPMs []string
+	var err error
+	toolchainManifest := *toolchainManifest
```
MateuszMalisz, line 71: I understand the shadowing but I would recommend using a slightly different name to avoid confusion.

toolkit/scripts/pkggen.mk
```diff
@@ -214,7 +211,7 @@ $(STATUS_FLAGS_DIR)/build-rpms.flag: $(preprocessed_file) $(chroot_worker) $(go-
 		--packages="$(PACKAGE_BUILD_LIST)" \
 		--rebuild-packages="$(PACKAGE_REBUILD_LIST)" \
 		--image-config-file="$(CONFIG_FILE)" \
-		--reserved-file-list-file="$(TOOLCHAIN_MANIFEST)" \
```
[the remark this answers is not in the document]
dmcilvaney: Probably missed something in a rebase, I've done a pass to make sure I've got quotes on everything else as well.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -309,6 +309,7 @@ func (c *Chroot) UnsafeRun(toRun func() error) (err error) {
 	}
 	defer originalWd.Close()
 
+	logger.Log.Debug("Entering Chroot")
```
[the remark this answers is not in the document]
dmcilvaney: Done

toolkit/tools/pkgworker/pkgworker.go
dmcilvaney: New: "Directory that contains already built toolchain RPMs. Should contain a top level directory for each architecture"

toolkit/tools/scheduler/scheduler.go
dmcilvaney: new: "Path to a list of RPMs which are created by the toolchain. RPMs from this list will are considered 'prebuilt' and will not be rebuilt"

toolkit/tools/scheduler/scheduler.go
dmcilvaney: Fair, updated.

toolkit/tools/scheduler/schedulerutils/buildlist.go
dmcilvaney: Added

toolkit/scripts/toolchain.mk
```diff
@@ -297,3 +305,23 @@ $(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(depend_REBUILD_TOOLCHAIN)
...
+# the toolchain RPMs should also be placed into out/RPMs.
+
+# ./out/SRPMS are still handled by build_official_toolchain_rpms.sh as a side effect of building the toolchain tar.gz.
+ifeq ($(REBUILD_TOOLCHAIN),y)
+$(RPMS_DIR): $(toolchain_out_rpms)
+
+# For each toolchain rpm in ./out/RPMS, add a dependency on its counterpart in the normal toolchain directory
+$(foreach toolchain_out_rpm,$(toolchain_out_rpms),$(eval $(toolchain_out_rpm): $(filter %/$(notdir $(toolchain_out_rpm)),$(toolchain_rpms))))
+$(toolchain_out_rpms): ; $(warning $^)
+	@if [ ! -f "$@" \
```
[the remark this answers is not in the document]
dmcilvaney: Done

toolkit/tools/specreader/specreader.go
dmcilvaney: Done

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -65,7 +65,7 @@ func main() {
 	var toolchainPackages []string
 	toolchainManifest := *toolchainManifest
 	if len(toolchainManifest) > 0 {
-		toolchainPackages, err = schedulerutils.ReadReservedFilesList(toolchainManifest)
+		toolchainPackages, err = schedulerutils.ReadReservedPackageManifest(toolchainManifest)
 		if err != nil {
 			logger.Log.Fatalf("unable to read toolchain manifest file '%s': %s", toolchainManifest, err)
 		}
```
PawelWMS, line 72: Could we also extract this logic into a separate function and remove the duplication from this file and `scheduler.go`? I think we're doing the exact same thing in both, so we could just do `toolchainPackages := schedulerutils.ReadReservedPackageManifest(...)`. I'm guessing all we'd have to do, is to allow `schedulerutils.ReadReservedPackageManifest` to accept an empty string as input.

toolkit/scripts/imggen.mk
```diff
@@ -105,6 +105,7 @@ $(image_package_cache_summary): $(go-imagepkgfetcher) $(chroot_worker) $(imggen_
 		--log-file=$(LOGS_DIR)/imggen/imagepkgfetcher.log \
 		--rpm-dir=$(RPMS_DIR) \
 		--tmp-dir=$(image_fetcher_tmp_dir) \
+		--toolchain-rpm-dir="$(TOOLCHAIN_RPMS_DIR)" \
```
PawelWMS, line 108: Let's be consistent about the plural form and use `toolchain-rpms-dir`.

toolkit/scripts/toolchain.mk
```diff
@@ -297,3 +305,23 @@ $(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(depend_REBUILD_TOOLCHAIN)
...
+			-o "$^" -nt "$@" ] ; then \
+		echo "Placing built toolchain RPM $(notdir $@) into $(RPMS_DIR)" && \
+		cp $^ $@; \
+	fi || $(call print_error, Failed to duplicate '$^' to '$@' )
+else
+$(toolchain_out_rpms):
+	@$(call print_error, Cannot build toolchain rpm ($@) unless REBUILD_TOOLCHAIN=y is set)
```
PawelWMS, line 331: Do we need the `else` part? In other words: is it even possible for any target to require `toolchain_out_rpms` if `REBUILD_TOOLCHAIN != y`? Unless the user manually tries to `make [toolchain_rpm]` but that's not something they should do and expect to work.

toolkit/scripts/toolchain.mk
```diff
@@ -297,3 +305,23 @@ $(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(depend_REBUILD_TOOLCHAIN)
...
+# ./out/RPMS is reserved for RPMs generated by the tooling, all other RPMs are cached in the ./build folder. If REBUILD_TOOLCHAIN=y is set
+# the toolchain RPMs should also be placed into out/RPMs.
+
+# ./out/SRPMS are still handled by build_official_toolchain_rpms.sh as a side effect of building the toolchain tar.gz.
+ifeq ($(REBUILD_TOOLCHAIN),y)
+$(RPMS_DIR): $(toolchain_out_rpms)
+
+# For each toolchain rpm in ./out/RPMS, add a dependency on its counterpart in the normal toolchain directory
+$(foreach toolchain_out_rpm,$(toolchain_out_rpms),$(eval $(toolchain_out_rpm): $(filter %/$(notdir $(toolchain_out_rpm)),$(toolchain_rpms))))
+$(toolchain_out_rpms): ; $(warning $^)
```
PawelWMS, line 323: Why print a warning for all toolchain out RPMs? It seems noisy and confusing. Or am I getting the `$(warning $^)` part wrong?

toolkit/scripts/toolchain.mk
```diff
@@ -297,3 +305,23 @@ $(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(depend_REBUILD_TOOLCHAIN)
...
+
+# ./out/RPMS is reserved for RPMs generated by the tooling, all other RPMs are cached in the ./build folder. If REBUILD_TOOLCHAIN=y is set
+# the toolchain RPMs should also be placed into out/RPMs.
+
+# ./out/SRPMS are still handled by build_official_toolchain_rpms.sh as a side effect of building the toolchain tar.gz.
+ifeq ($(REBUILD_TOOLCHAIN),y)
+$(RPMS_DIR): $(toolchain_out_rpms)
+
+# For each toolchain rpm in ./out/RPMS, add a dependency on its counterpart in the normal toolchain directory
+$(foreach toolchain_out_rpm,$(toolchain_out_rpms),$(eval $(toolchain_out_rpm): $(filter %/$(notdir $(toolchain_out_rpm)),$(toolchain_rpms))))
```
PawelWMS, line 322: This is insanely hard to read, took me a while to understand what's going on. But if we're going to rebuilt the toolchain, shouldn't we always expect `$(toolchain_out_rpms)` to depend on the whole set of `$(toolchain_rpms)`? Something along these lines:
```make
$(toolchain_out_rpms): $(toolchain_rpms)
```

PawelWMS: Or I'm completely misunderstanding this and the following lines and how they work together. I think I understand what we want to do, just not how this these lines work and get us there.

toolkit/scripts/toolchain.mk
```diff
@@ -70,8 +78,8 @@ clean-toolchain-rpms:
 	for f in $(toolchain_rpms_noarch); do rm -vf $(RPMS_DIR)/noarch/$$f; done
 
 copy-toolchain-rpms:
```
PawelWMS, line 79: I'm curious if this now could just depend on `$(toolchain_out_rpms)` and we could remove the whole logic below?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -78,24 +79,28 @@ func New() *RpmRepoCloner {
...
+		chrootLocalToolchainDir = "/toolchainrpms"
 
-		overlayWorkDirectory  = "/overlaywork/workdir"
-		overlayUpperDirectory = "/overlaywork/upper"
-		overlaySource         = "overlay"
+		overlayWorkDirectoryRpms       = "/overlaywork/workdir_rpms"
+		overlayUpperDirectoryRpms      = "/overlaywork/upper_rpms"
+		overlayWorkDirectoryToolchain  = "/overlaywork/workdir_toolchain"
+		overlayUpperDirectoryToolchain = "/overlaywork/upper_toolchain"
+		overlaySource                  = "overlay"
```
PawelWMS, line 103: Since we're here, let's at least add some alphabetical order to it (while still keeping the upper-workdir grouping):
```suggestion
		overlaySource                  = "overlay"
		overlayUpperDirectoryRpms      = "/overlaywork/upper_rpms"
		overlayWorkDirectoryRpms       = "/overlaywork/workdir_rpms"
		overlayUpperDirectoryToolchain = "/overlaywork/upper_toolchain"
		overlayWorkDirectoryToolchain  = "/overlaywork/workdir_toolchain"
```

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -137,12 +142,19 @@ func (r *RpmRepoCloner) Initialize(destinationDir, tmpDir, workerTar, existingRp
...
-	overlayMount, overlayExtraDirs := safechroot.NewOverlayMountPoint(r.chroot.RootDir(), overlaySource, chrootLocalRpmsDir, existingRpmsDir, overlayUpperDirectory, overlayWorkDirectory)
+	outRpmsOverlayMount, overlayExtraDirs := safechroot.NewOverlayMountPoint(r.chroot.RootDir(), overlaySource, chrootLocalRpmsDir, existingRpmsDir, overlayUpperDirectoryRpms, overlayWorkDirectoryRpms)
 	extraMountPoints := []*safechroot.MountPoint{
-		overlayMount,
+		outRpmsOverlayMount,
 		safechroot.NewMountPoint(destinationDir, chrootDownloadDir, bindFsType, safechroot.BindMountPointFlags, bindData),
 	}
 
+	// Optionally include the special toolchain package directory if it is provided
+	if len(chrootLocalToolchainDir) > 0 {
```
PawelWMS, line 152: This looks fishy. The check doesn't make sense, since `chrootLocalToolchainDir` is a constant. Also, I'd expect the toolchain to always be pre-populated before this tool is run.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -155,6 +167,9 @@ func (r *RpmRepoCloner) Initialize(destinationDir, tmpDir, workerTar, existingRp
 	// use overlay so cache repo must be explicitly initialized.
 	// We make sure it's present during all builds to avoid noisy TDNF error messages in the logs.
 	reposToInitialize := []string{chrootLocalRpmsDir, chrootDownloadDir, cacheRepoDir}
+	if len(chrootLocalToolchainDir) > 0 {
```
PawelWMS, line 170: Similar story for this check - looks suspicious, since this is always true.

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -148,11 +154,12 @@ func buildSRPMInChroot(chrootDir, rpmDirPath, workerTar, srpmFile, repoFile, rpm
...
+	outRpmsOverlayMount, outRpmsOverlayExtraDirs := safechroot.NewOverlayMountPoint(chroot.RootDir(), overlaySource, chrootLocalRpmsDir, rpmDirPath, chrootLocalRpmsDir, overlayWorkDirRpms)
+	toolchainRpmsOverlayMount, toolchainRpmsOverlayExtraDirs := safechroot.NewOverlayMountPoint(chroot.RootDir(), overlaySource, chrootLocalToolchainDir, toolchainDirPath, chrootLocalToolchainDir, overlayWorkDirToolchain)
 	rpmCacheMount := safechroot.NewMountPoint(*cacheDir, chrootLocalRpmsCacheDir, "", safechroot.BindMountPointFlags, "")
 	ccacheMount := safechroot.NewMountPoint(*ccacheDir, chrootCcacheDir, "", safechroot.BindMountPointFlags, "")
-	mountPoints := []*safechroot.MountPoint{overlayMount, rpmCacheMount, ccacheMount}
-	extraDirs := append(overlayExtraDirs, chrootLocalRpmsCacheDir, chrootCcacheDir)
+	mountPoints := []*safechroot.MountPoint{outRpmsOverlayMount, toolchainRpmsOverlayMount, rpmCacheMount, ccacheMount}
+	extraDirs := append(outRpmsOverlayExtraDirs, append(toolchainRpmsOverlayExtraDirs, chrootLocalRpmsCacheDir, chrootCcacheDir)...)
```
PawelWMS, line 162: Could we break up the double `append`? Seems a bit hard to read.
```suggestion
	extraDirs := append(outRpmsOverlayExtraDirs, chrootLocalRpmsCacheDir, chrootCcacheDir)
	extraDirs = append(extraDirs, toolchainRpmsOverlayExtraDirs...)
```

toolkit/tools/scheduler/schedulerutils/buildlist.go
```diff
@@ -123,3 +126,16 @@ func ReadReservedFilesList(path string) (reservedFiles []string, err error) {
 
 	return reservedFiles, nil
 }
+
+// IsToolchainRPM determines if a given RPM path is found in a list of reserved RPMs.
+// rpmsToFind should be an absolute path to the expected RPM, while toolchainRPMs is a list of '<name>-<version>-<release>.rpm'.
+func IsToolchainRPM(rpmPath string, reservedRPMs []string) bool {
```
PawelWMS, line 132: Seeing how the caller must pass the list of reserved RPMs anyway, I'd re-name it to something more generic like `IsReservedRPM`. Or even `IsReservedFile`. The caller can still give more context about the returned value with a correct choice of a variable name.

toolkit/tools/scheduler/schedulerutils/buildlist.go
```diff
@@ -123,3 +126,16 @@ func ReadReservedFilesList(path string) (reservedFiles []string, err error) {
 
 	return reservedFiles, nil
 }
+
+// IsToolchainRPM determines if a given RPM path is found in a list of reserved RPMs.
+// rpmsToFind should be an absolute path to the expected RPM, while toolchainRPMs is a list of '<name>-<version>-<release>.rpm'.
```
PawelWMS, line 131: The comment doesn't match the argument names. Also, it seems that both `rpmPath` and strings from `reservedRPMs` may be either file paths or simple file names, so we should make the comment reflect that.

toolkit/tools/specreader/specreader.go
```diff
@@ -63,13 +66,23 @@ func main() {
 		logger.Log.Panicf("Value in --workers must be greater than zero. Found %d", *workers)
 	}
 
-	err := parseSPECsWrapper(*buildDir, *specsDir, *rpmsDir, *srpmsDir, *distTag, *output, *workerTar, *workers, *runCheck)
+	var toolchainRPMs []string
+	var err error
+	toolchainManifestPath := *toolchainManifest
+	if len(toolchainManifestPath) > 0 {
```
PawelWMS, line 72: Is the check necessary, since we're using `ExistingFile()` on the `toolchain-manifest` argument?

toolkit/tools/specreader/specreader.go
```diff
@@ -63,13 +66,23 @@ func main() {
...
+		toolchainRPMs, err = schedulerutils.ReadToolchainPackageManifest(toolchainManifestPath)
+		if err != nil {
+			logger.Log.Fatalf("unable to read toolchain manifest file '%s': %s", toolchainManifestPath, err)
+		}
+	}
```
PawelWMS, line 77: Is the check necessary, since we're using `ExistingFile()` on the `toolchain-manifest` argument?
```suggestion
	toolchainRPMs, err := schedulerutils.ReadToolchainPackageManifest(*toolchainManifest)
	if err != nil {
		logger.Log.Fatalf("unable to read toolchain manifest file '%s': %s", toolchainManifestPath, err)
	}
```

toolkit/tools/specreader/specreader.go
```diff
@@ -63,13 +66,23 @@ func main() {
...
+
+	err = parseSPECsWrapper(*buildDir, *specsDir, *rpmsDir, *srpmsDir, toolchainRPMs, *existingToolchainRpmDir, *distTag, *output, *workerTar, *workers, *runCheck)
 	logger.PanicOnError(err)
 }
 
 // parseSPECsWrapper wraps parseSPECs to conditionally run it inside a chroot.
 // If workerTar is non-empty, parsing will occur inside a chroot, otherwise it will run on the host system.
-func parseSPECsWrapper(buildDir, specsDir, rpmsDir, srpmsDir, distTag, outputFile, workerTar string, workers int, runCheck bool) (err error) {
+func parseSPECsWrapper(buildDir, specsDir, rpmsDir, srpmsDir string, toolchainRPMs []string, toolchainDir, distTag, outputFile, workerTar string, workers int, runCheck bool) (err error) {
```
PawelWMS, line 85: No reason to repeat argument type for `string`:
```suggestion
func parseSPECsWrapper(buildDir, specsDir, rpmsDir, srpmsDir, toolchainDir, distTag, outputFile, workerTar string, toolchainRPMs []string, workers int, runCheck bool) (err error) {
```

toolkit/tools/specreader/specreader.go
```diff
@@ -182,7 +195,7 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
 }
 
 // parseSPECs will parse all specs in specsDir and return a summary of the SPECs.
-func parseSPECs(specsDir, rpmsDir, srpmsDir, distTag, arch string, workers int, runCheck bool) (packageRepo *pkgjson.PackageRepo, err error) {
+func parseSPECs(specsDir, rpmsDir, srpmsDir, toolchainDir string, toolchainRPMs []string, distTag, arch string, workers int, runCheck bool) (packageRepo *pkgjson.PackageRepo, err error) {
```
PawelWMS, line 198: Same here:
```suggestion
func parseSPECs(specsDir, rpmsDir, srpmsDir, toolchainDir distTag, arch string, toolchainRPMs []string, workers int, runCheck bool) (packageRepo *pkgjson.PackageRepo, err error) {
```

toolkit/tools/specreader/specreader.go
```diff
@@ -442,12 +456,18 @@ func parseProvides(rpmsDir, srpmPath string, list []string) (providerlist []*pkg
 				return
 			}
 
+			isToolchain := schedulerutils.IsToolchainRPM(rpmPath, toolchainRPMs)
+			if isToolchain {
+				rpmPath = convertToToolchainRpmPath(rpmPath, packagearch, toolchainDir)
+			}
+
```
PawelWMS, line 463:
```suggestion
			if isToolchain := schedulerutils.IsToolchainRPM(rpmPath, toolchainRPMs); isToolchain {
				rpmPath = convertToToolchainRpmPath(rpmPath, packagearch, toolchainDir)
			}

```

toolkit/tools/specreader/specreader.go
```diff
@@ -606,3 +626,12 @@ func filterOutDynamicDependencies(pkgVers []*pkgjson.PackageVer) (filteredPkgVer
 
 	return
 }
+
+// convertToToolchainRpmPath updates the .rpm path to point the the toolchain directory rather than the normal out/RPMs
+// directory
+func convertToToolchainRpmPath(currentRpmPath, arch string, toolchainDir string) (toolchainPath string) {
```
PawelWMS, line 632:
```suggestion
// convertToToolchainRpmPath updates the RPM path to point the the toolchain directory rather than the normal out/RPMs
// directory
func convertToToolchainRpmPath(currentRpmPath, arch, toolchainDir string) (toolchainPath string) {
```

toolkit/tools/specreader/specreader.go
```diff
@@ -606,3 +626,12 @@ func filterOutDynamicDependencies(pkgVers []*pkgjson.PackageVer) (filteredPkgVer
...
+	dir := filepath.Join(toolchainDir, arch)
+	toolchainPath = filepath.Join(dir, filepath.Base(currentRpmPath))
```
PawelWMS, line 634:
```suggestion
	rpmFileName := filepath.Base(currentRpmPath)
	toolchainPath = filepath.Join(toolchainDir, arch)
	toolchainPath = filepath.Join(toolchainPath, rpmFileName)
```

toolkit/tools/specreader/specreader.go
```diff
@@ -606,3 +626,12 @@ func filterOutDynamicDependencies(pkgVers []*pkgjson.PackageVer) (filteredPkgVer
...
+	logger.Log.Debugf("Toolchain changing %s to %s", currentRpmPath, toolchainPath)
```
PawelWMS, line 635: Let's stick to surrounding variables one way or another:

```suggestion
	logger.Log.Debugf("Toolchain changing '%s' to '%s'.", currentRpmPath, toolchainPath)
```

or:
```suggestion
	logger.Log.Debugf("Toolchain changing (%s) to (%s).", currentRpmPath, toolchainPath)
```

toolkit/tools/specreader/specreader.go
```diff
@@ -442,12 +456,18 @@ func parseProvides(rpmsDir, srpmPath string, list []string) (providerlist []*pkg
...
+				rpmPath = convertToToolchainRpmPath(rpmPath, packagearch, toolchainDir)
+			}
+
 			providerPkgVer := &pkgjson.Package{
 				Provides:     newProviderVer[0],
 				SrpmPath:     srpmPath,
 				RpmPath:      rpmPath,
 				Architecture: packagearch,
 				Requires:     reqlist,
+				IsToolchain:  isToolchain,
```
PawelWMS, line 470: Is this new field ever used?

toolkit/tools/specreader/specreader.go
```diff
@@ -410,6 +423,7 @@ func parseProvides(rpmsDir, srpmPath string, list []string) (providerlist []*pkg
 
 		if listEntry[tag] == "rpm" {
 			logger.Log.Trace("rpm ", listEntry[value])
+			// We will decide if this is a toolchain package later, and update the path if so.
```
PawelWMS, line 426: Man, this is a terrible function to read. Am I getting this right, that on one loop we'll set `rpmPath` and then we expect another run where the tag will be "provides", so we can use the `rpmPath` we just set there? Are we even guaranteed to end with a "provides" for each `rpmPath` we set?

toolkit/scripts/pkggen.mk
```diff
@@ -125,12 +124,13 @@ ifeq ($(STOP_ON_FETCH_FAIL),y)
 graphpkgfetcher_extra_flags += --stop-on-failure
 endif
 
-$(cached_file): $(graph_file) $(go-graphpkgfetcher) $(chroot_worker) $(pkggen_local_repo) $(depend_REPO_LIST) $(REPO_LIST) $(call shell_real_build_only, find $(CACHED_RPMS_DIR)/) $(pkggen_rpms) $(TOOLCHAIN_MANIFEST)
```
PawelWMS, line 128: Is `pkggen_rpms` even still used anywhere?

toolkit/tools/scheduler/buildagents/definition.go
```diff
@@ -9,13 +9,14 @@ import "fmt"
...
-	RepoFile  string
-	RpmDir    string
-	SrpmDir   string
-	CacheDir  string
-	CCacheDir string
+	WorkDir      string
+	WorkerTar    string
+	RepoFile     string
+	RpmDir       string
+	ToolchainDir string
```
PawelWMS, line 16: Is `ToolchainDir` used for anything?

toolkit/scripts/pkggen.mk
```diff
@@ -177,10 +177,6 @@ clean-compress-srpms:
 	rm -rf $(srpms_archive)
 
 ifeq ($(REBUILD_PACKAGES),y)
-# If we are responsible for building a toolchain, make sure those RPMs are also present in the output directory
-ifeq ($(REBUILD_TOOLCHAIN),y)
-$(RPMS_DIR): $(toolchain_rpms)
-endif
```
PawelWMS, line 183: Why we removed this dependency? Don't we want to have the toolchain in the `out/RPMS` directory if we're asking to rebuild it? Or is this already happening through other dependencies?

toolkit/tools/scheduler/schedulerutils/buildlist.go
```diff
@@ -101,8 +102,9 @@ func filterLocalPackagesOnly(packageVersionsInConfig []*pkgjson.PackageVer, inpu
 	return
 }
 
-// ReadReservedFilesList updates the list of reserved files from the manifest file passed in.
-func ReadReservedFilesList(path string) (reservedFiles []string, err error) {
+// ReadToolchainPackageManifest updates the list of reserved files (such as toolchain RPMs) from the manifest file passed in.
+// Entries will be returned in the form '<rpm>-<version>-<release>.rpm' with any preceding path removed.
+func ReadToolchainPackageManifest(path string) (reservedFiles []string, err error) {
```
PawelWMS, line 107: Seeing how it's used to read any file with a list of packages, I think the original generic name was more accurate. Especially since the `depsearch.go:isFilteredFile()` function doesn't really care what kind of entries (toolchain packages, a list of groceries, etc.) the list contains.

toolkit/docs/building/building.md
```diff
@@ -569,7 +568,7 @@ These are the useful build targets:
 | clean-*                          | Most targets have a `clean-<target>` target which selectively cleans the target's output.
 | compress-rpms                    | Compresses all RPMs in `../out/RPMS` into `../out/rpms.tar.gz`. See `hydrate-rpms` target.
 | compress-srpms                   | Compresses all SRPMs in `../out/SRPMS` into `../out/srpms.tar.gz`.
-| copy-toolchain-rpms              | Copy all toolchain RPMS from `../build/rpm_cache/cache` to  `../out/RPMS`.
+| copy-toolchain-rpms              | **Generally no longer needed** Copy all toolchain RPMS from `../build/rpm_cache/cache` to  `../out/RPMS`.
```
PawelWMS, line 601: Could we remove its usage from our pipelines in that case? This would be a good test if it's really no longer needed.

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
dmcilvaney: Sure, that seems reasonable, will update.

toolkit/scripts/imggen.mk
dmcilvaney: Done

toolkit/scripts/toolchain.mk
dmcilvaney: Probably not needed, but I wanted a bit of hint on what might be happening if things go wrong so I can debug more easily.

toolkit/scripts/toolchain.mk
dmcilvaney: Don't recall the exact reason for it (its been a while...), I think we can drop it though. There are some weird syntax with adding $(warnings...) to the makefile. I'll just pull the warning, best I can tell it's just extra debug code.

toolkit/scripts/toolchain.mk
dmcilvaney: I front loaded the filter operation to the makefile foreach so I could do `cp $^ $@` (ie `cp src.rpm dest.rpm`). I can implement that inside the bash instead if that is more readable, or at the lest I can add more to the comment.

dmcilvaney: Ok, I think I've found a more readable way to do this.
```makefile
# For each toolchain rpm in ./out/RPMS, add a dependency on the counterparts in the normal toolchain directory:
# Each path in $(toolchain_out_rpms) corresponds to a .rpm file we expect to have been built by the toolchain target and made available in ./out/RPMS.
# Those rpms however are placed by default in ./build/toolchain/* (listed in $(toolchain_rpms)). So if we want a copy placed in ./out/RPMS 
# we will need to copy it over. We can filter the list of toolchain rpms $(toolchain_rpms) to find the source that matches the target ($@),
# then copy it over.
$(toolchain_out_rpms): $(toolchain_rpms)
	@src_rpm='$(filter %/$(notdir $@),$(toolchain_rpms))'  && \
	if [ ! -f "$@" \
			-o "$$src_rpm" -nt "$@" ] ; then \
		echo "Placing built toolchain RPM $(notdir $@) into $(RPMS_DIR)" && \
		cp $$src_rpm $@; \
	fi || $(call print_error, Failed to duplicate '$$src_rpm' to '$@' )
else
$(toolchain_out_rpms):
	@$(call print_error, Cannot build toolchain rpm ($@) unless REBUILD_TOOLCHAIN=y is set)
endif
```

dmcilvaney: I still use the make pre-processor since the grep command ended up being quite a mess.

toolkit/scripts/toolchain.mk
dmcilvaney: I was debating what to do with copy-toolchain-rpms, I decided on leaving it unchanged since it could break a lot of pipelines if it went wrong.

If we want to do that I can move `$(toolchain_out_rpms)` out of the `ifeq ($(REBUILD_TOOLCHAIN),y)` block and only leave `$(RPMS_DIR): $(toolchain_out_rpms)` in the conditional.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Good catch. This was meant to be `toolchainRpmsDir`. I was thinking its reasonable that the cloner may be used in the future without a full toolchain available. I'll go ahead and drop it though, because at the rework to split the tools up more we were considering back then hasn't gone anywhere (and we can just add back later if needed).

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: also removed

toolkit/tools/pkgworker/pkgworker.go
dmcilvaney: Done

toolkit/tools/scheduler/schedulerutils/buildlist.go
dmcilvaney: Done

toolkit/tools/scheduler/schedulerutils/buildlist.go
dmcilvaney: Done

toolkit/tools/specreader/specreader.go
dmcilvaney: done

toolkit/tools/specreader/specreader.go
dmcilvaney: done

toolkit/tools/specreader/specreader.go
dmcilvaney: done

toolkit/tools/specreader/specreader.go
dmcilvaney: done

toolkit/tools/specreader/specreader.go
dmcilvaney: done

toolkit/tools/specreader/specreader.go
dmcilvaney: done

toolkit/tools/specreader/specreader.go
dmcilvaney: Hmm, no. I think this is left over from a previous implementation where the toolchain state of a package was passed to the grapher. Will remove.

toolkit/scripts/pkggen.mk
dmcilvaney: Hmm, looks like not... I feel like `$(RPMS_DIR):` should maybe depend on it though? If we manually add an rpm, we should retrigger an image build since one of its component RPMs may have changed.

I'm going to leave it for now and look at that in the future.

toolkit/scripts/pkggen.mk
dmcilvaney: Moved to https://github.com/microsoft/CBL-Mariner/pull/4210/files#diff-eef96fc8d3aac83583f5507a002236b9b08b6ce1d2fd7d9f0f04b06bd84d6a7eR57

toolkit/tools/scheduler/buildagents/definition.go
dmcilvaney: Its passed to the build agent via `serializeChrootBuildAgentConfig()` (chrootagent.go:84)

toolkit/tools/scheduler/schedulerutils/buildlist.go
dmcilvaney: Updated

toolkit/docs/building/building.md
dmcilvaney: Not easily, since we still want to support old versions of 2.0 and all 1.0 builds. I'd also like to not break downstream builds that may use it. Maybe for 3.0 we can fully drop it since that is a good if/else condition that we probably already are going to be using.

toolkit/tools/specreader/specreader.go
dmcilvaney: Will give this some thought, might return later to clean it up.

toolkit/docs/building/building.md
PawelWMS: You're right, that it's not trivial. But in that case I'd maybe use this wording to send a clear message of our intentions:
```suggestion
| copy-toolchain-rpms              | **DEPRECATED** Copy all toolchain RPMS from `../build/rpm_cache/cache` to  `../out/RPMS`.
```

toolkit/tools/scheduler/schedulerutils/buildlist.go
```diff
@@ -101,8 +102,15 @@ func filterLocalPackagesOnly(packageVersionsInConfig []*pkgjson.PackageVer, inpu
 	return
 }
 
-// ReadReservedFilesList updates the list of reserved files from the manifest file passed in.
+// ReadReservedFilesList updates the list of reserved files (such as toolchain RPMs) from the manifest file passed in.
```
PawelWMS, line 105: >updates [](http://example.com/codeflow?start=25&length=7)

Not part of your PR, but since we're here: the comment is incorrect. We're reading the list of reserved files, not updating it.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -137,12 +142,17 @@ func (r *RpmRepoCloner) Initialize(destinationDir, tmpDir, workerTar, existingRp
...
 	// files into it.
-	overlayMount, overlayExtraDirs := safechroot.NewOverlayMountPoint(r.chroot.RootDir(), overlaySource, chrootLocalRpmsDir, existingRpmsDir, overlayUpperDirectory, overlayWorkDirectory)
+	outRpmsOverlayMount, overlayExtraDirs := safechroot.NewOverlayMountPoint(r.chroot.RootDir(), overlaySource, chrootLocalRpmsDir, existingRpmsDir, overlayUpperDirectoryRpms, overlayWorkDirectoryRpms)
 	extraMountPoints := []*safechroot.MountPoint{
-		overlayMount,
+		outRpmsOverlayMount,
 		safechroot.NewMountPoint(destinationDir, chrootDownloadDir, bindFsType, safechroot.BindMountPointFlags, bindData),
 	}
 
+	// Include the special toolchain package directory
```
PawelWMS, line 151: >package [](http://example.com/codeflow?start=34&length=7)

NIt

PawelWMS, line 151: Nit-pick:
```suggestion
	// Include the special toolchain packages directory.
```

toolkit/tools/scheduler/schedulerutils/buildlist.go
PawelWMS, line 105: Not part of your PR, but since we're here:
```suggestion
// ReadReservedFilesList reads the list of reserved files (such as toolchain RPMs) from the manifest file passed in.
```

toolkit/tools/specreader/specreader.go
PawelWMS: I think you missed this one. Or just forgot to commit/push.;)

toolkit/scripts/toolchain.mk
PawelWMS: Git it. I'd still remove it in the final version. I think we shouldn't try to safeguard against any possible misuse of `make`. We'd probably end up with piles of warning and code that's never used.

toolkit/scripts/toolchain.mk
```diff
@@ -297,3 +307,26 @@ $(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(depend_REBUILD_TOOLCHAIN)
...
+# the toolchain RPMs should also be placed into out/RPMs.
+# ./out/SRPMS are still handled by build_official_toolchain_rpms.sh as a side effect of building the toolchain tar.gz.
+ifeq ($(REBUILD_TOOLCHAIN),y)
+$(RPMS_DIR): $(toolchain_out_rpms)
+
+# For each toolchain rpm in ./out/RPMS, add a dependency on the counterparts in the normal toolchain directory:
+# Each path in $(toolchain_out_rpms) corresponds to a .rpm file we expect to have been built by the toolchain target and made available in ./out/RPMS.
+# Those rpms however are placed by default in ./build/toolchain/* (listed in $(toolchain_rpms)). So if we want a copy placed in ./out/RPMS 
+# we will need to copy it over. We can filter the list of toolchain rpms $(toolchain_rpms) to find the source that matches the target ($@),
+# then copy it over.
```
PawelWMS, line 321: Nit-pick for the sake of consistency:
```suggestion
# For each toolchain RPM in ./out/RPMS, add a dependency on the counterparts in the normal toolchain directory:
# Each path in $(toolchain_out_rpms) corresponds to a .rpm file we expect to have been built by the toolchain target and made available in ./out/RPMS.
# Those RPMs however are placed by default in ./build/toolchain/* (listed in $(toolchain_rpms)). So if we want a copy placed in ./out/RPMS 
# we will need to copy it over. We can filter the list of toolchain rpms $(toolchain_rpms) to find the source that matches the target ($@),
# then copy it over.
```

toolkit/scripts/toolchain.mk
```diff
@@ -297,3 +307,26 @@ $(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(depend_REBUILD_TOOLCHAIN)
...
+$(toolchain_out_rpms): $(toolchain_rpms)
+	@src_rpm='$(filter %/$(notdir $@),$(toolchain_rpms))'  && \
+	if [ ! -f "$@" \
+			-o "$$src_rpm" -nt "$@" ] ; then \
+		echo "Placing built toolchain RPM $(notdir $@) into $(RPMS_DIR)" && \
+		cp $$src_rpm $@; \
```
PawelWMS, line 327: Are these two commands being printed in addition to the echo message itself? If so, I think we could silence them to reduce the noise and rely on the echo message itself. I already forgot how `make` handles this, so I apologize if I'm not making sense.:)

toolkit/tools/scheduler/buildagents/definition.go
PawelWMS: Yes, but keeping it, just so we can print it seems pointless. And I think it's not used by the code in any other way.

dmcilvaney: Sorry, I didn't clarify fully. It gets passed to the serializer, which passes that flag to the build agents so they can mount the toolchain directory and install RPMs from it during a build.

It gets passed through to `pkgworker.go:158` where it is used to create an overlay.

toolkit/scripts/toolchain.mk
dmcilvaney: Sure, I'll drop it.

toolkit/docs/building/building.md
dmcilvaney: Added a few details: `**[DEPRECATED]: This should no longer be needed as a work around in core repo builds. Will be removed in future versions.** Copy all toolchain RPMS from `../build/toolchain_rpms` to  `../out/RPMS`.`

toolkit/scripts/toolchain.mk
dmcilvaney: The entire command is chained with `&& \`, the first line has a `@` which should suppress the rest of the output.

toolkit/tools/specreader/specreader.go
dmcilvaney: I think I'll leave this as is for now, this function needs a re-write for sure but I'd rather not do that in this PR.

toolkit/tools/specreader/specreader.go
dmcilvaney: Ok, should be actually included now.

PawelWMS (approved):
:shipit:

dmcilvaney:
Rebased to accomodate #4666 #4751 #4949.

dmcilvaney:
Got a successful build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=333861&view=results

PR #4214, opened 2022-11-12
Prioritize already cached RPMs before using online repos

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

The fetchers operate in two modes, depending on if the build is using reusable container friendly chroot dirs, or dynamically generated chroot dirs.

The fetchers incrementally enable more repos from their priority list until they find a matching package. In the reusable chroot case, the fetchers will query the `cacheRepoID` repo to check for files before enabling the `allRepoIDs` (`"*"`) repos. This represents the local cache as generated by the current and previous runs of the fetcher. When running in the normal dynamic chroot mode however, we instead place the rpms into `fetcherRepoID`, which we do not enable. We should enable this ID in the same way we enable `cacheRepoID`.

Enabling this means the fetchers can likely find a local copy of the package prior to needing to reach out to the upstream servers, avoiding network operations which are generally very slow.

Also, the ordering of repos checked is an implementation detail of `tdnf` which we take advantage of (it checks in the order they appear in the repo files). To manage this we add each repo in order of priority to a single repo file.

We currently leave the existing upstream repo files intact in the repos folder however, and we cannot guarantee which order those files are parsed when we enable `"*"` repos. Instead merge the existing repo files into our custom build, prioritized repo file so that they appear after all the local repos we create.

Together this should speed up the fetcher tools by an order of magnitude when running on a hydrated environment (ie after the 1st successful build). Testing shows speedups from around 10 minutes to ~30-60 seconds.

###### Change Log  

- Check the fetcher catch for existing RPMs when working with normal chroots
- Remove any non-determinism in the order we enable all repos

**NO**

###### Test Methodology

- Local testing


oliviacrain (approved):
This is going to be very impactful- excited to see how it affects our build times (especially for buddy and delta builds)

CC @anphel31, was performance the only blocker keeping us from discontinuing use of the PMC downloader script in the delta build pipelines?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -290,7 +325,7 @@ func (r *RpmRepoCloner) Clone(cloneDeps bool, packagesToClone ...*pkgjson.Packag
 		err = r.chroot.Run(func() (err error) {
 			var chrootErr error
 			// Consider the built RPMs first, then the already cached (e.g. tooolchain), and finally all remote packages.
-			repoOrderList := []string{builtRepoID, cacheRepoID, allRepoIDs}
+			repoOrderList := []string{builtRepoID, fetcherRepoID, cacheRepoID, allRepoIDs}
```
PawelWMS, line 328: The source of this mess with using different folders for different build type is an issue with how our build pipelines are set-up and in general is a bug/misconfiguration. We need your change here and below as a workaround until that bug gets fixed. To make it easier to clean-up the code afterwards, though, could I ask you to make a small change, so we keep the workarounds closer to each other?

My ask: instead of adding `fetcherRepoID` to the `repoOrderList`, could we instead create an `effectiveCacheRepoID` variable and set it to either `cacheRepoID` or `fetcherRepoID` through that `if !buildpipeline.IsRegularBuild()` few lines above?

I'd have a similar ask for the changes in `WhatProvides`.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -229,6 +250,20 @@ func (r *RpmRepoCloner) initializeRepoDefinitions(repoDefinitions []string) (err
...
 
+	// Add each previously existing repofile to the end of the new file, then delete the original.
+	// We want to try our custom mounted repos before reaching out to the upstream servers.
+	for _, repoFile := range existingRepoFiles {
+		path := filepath.Join(fullRepoDirPath, repoFile.Name())
+		err = appendRepoFile(path, dstFile)
+		if err != nil {
+			return
+		}
+		err = os.Remove(path)
```
PawelWMS, line 261:
```suggestion
		originalRepoFilePath := filepath.Join(fullRepoDirPath, repoFile.Name())
		err = appendRepoFile(path, dstFile)
		if err != nil {
			return
		}
		err = os.Remove(originalRepoFilePath)
```

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -203,16 +205,35 @@ func (r *RpmRepoCloner) initializeRepoDefinitions(repoDefinitions []string) (err
...
 
-	fullRepoFilePath := filepath.Join(r.chroot.RootDir(), chrootRepoFile)
+	fullRepoDirPath := filepath.Join(r.chroot.RootDir(), chrootRepoDir)
+	fullRepoFilePath := filepath.Join(fullRepoDirPath, chrootRepoFile)
 
-	// Create the directory for the repo file
-	err = os.MkdirAll(filepath.Dir(fullRepoFilePath), os.ModePerm)
+	// Get a list of the existing repofiles that are part of the chroot, if any
+	existingRepoFiles := []fs.DirEntry{}
+	hasRepoDir, err := file.DirExists(fullRepoDirPath)
```
PawelWMS, line 218: I'd suggest just always creating the directory, since then the existence change is unnecessary and `existingRepoFiles` should still be populated correctly (if the repo didn't exist, then the list will simply be empty).

PawelWMS (approved):
Minor refactoring comments. Looks good, though!

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Sounds reasonable, will do!

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Will do.

PawelWMS (approved):

oliviacrain (approved):

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -229,6 +240,20 @@ func (r *RpmRepoCloner) initializeRepoDefinitions(repoDefinitions []string) (err
 		}
 	}
 
+	// Add each previously existing repofile to the end of the new file, then delete the original.
+	// We want to try our custom mounted repos before reaching out to the upstream servers.
+	for _, repoFile := range existingRepoFiles {
```
MateuszMalisz, line 245: I suggest filtering the `existingRepoFiles` to files that end with `.repo` only.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -324,8 +351,14 @@ func (r *RpmRepoCloner) WhatProvides(pkgVer *pkgjson.PackageVer) (packageNames [
 	}
 
 	foundPackages := make(map[string]bool)
+
+	effectiveCacheRepo := fetcherRepoID
```
MateuszMalisz, line 355: I would consider creating a simple function instead of this assignment to avoid potential mistakes in the future when someone might miss a place where `cacheRepoId` is calculated. I do not anticipate that issue to arise, but it might be a good idea to be defensive anyway. If you consider having a check for `IsRegularBuild` locally more readable, feel free to leave it as is.

MateuszMalisz (approved):
LGTM; minor suggestions included.

oliviacrain (approved):

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Done, using filepath.Glob() now

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Added ` selectCorrectCacheRepoID()`

dmcilvaney:
Minor merge conflict with `WhatProvides()`, no changes to the new code.

PR #4232, opened 2022-11-15
update generation 2 boot order during iso installation

SeanDougherty:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Current behavior:
Mariner installs itself onto Disk and reboots. Leaving boot order untouched, which causes seconds of failed booting into PXE after the post-install reboot and subsequent reboots.

Introduced behavior:
When in an EFI boot type'd iso install, we have the ability to set our boot order via the EFI boot manager (i.e., efibootmgr). Using this tool's functionality, and the knowledge specified by our image config files (e.g., unattended.json configs, attendedconfig.json, etc) we can set the first target in the boot order to our shim 'bootx64.efi' which typically lives in /boot/efi

###### Change Log  

Primary:
 - When installing a gen2 image via ISO, set the newly installed shim (bootx64.efi) as the first target in the EFI boot order
 - All 'installFunc' implementations must now pass back a new InstallationDetails object 

Supporting Changes:
 - Change configuration.setDefaultConfig to an exported function, so that config objects not generated by configuration.Load() (i.e., attended installers) can have their default config properly set.
 - Added a new utility function to configuration.go to find the first partition that has the Boot flag set.
 - Split updateBootOrder() into multiple functions, ejectDisk() and updateBootOrder()
 - Added a new function, removeOldMarinerBootTargets(), that removes older entries of the Mariner shim from the efi boot order when attempting to install.

No

###### Associated issues  

- [(Work Item)/(Bug) 34925438](https://microsoft.visualstudio.com/OS/_sprints/taskboard/Mariner%20Platform/OS/2211?workitem=34925438)

###### Test Methodology

- Multiple local builds of ISOs (attended and unattended)
- Pipeline build: ([ARM64 - PASSED](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=274138&view=results)) ([AMD64 - PASSED](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=274222&view=results))

### DIFF
--- a/toolkit/tools/imagegen/configuration/configuration.go
+++ b/toolkit/tools/imagegen/configuration/configuration.go
@@ -304,7 +304,7 @@ func Load(configFilePath string) (config Config, err error) {
 		return
 	}
 
-	config.setDefaultConfig()
+	config.SetDefaultConfig()
 
 	return
 }
@@ -405,7 +405,7 @@ func resolveBaseDirPath(baseDirPath, configFilePath string) (absoluteBaseDirPath
 	return filepath.Abs(baseDirPath)
 }
 
-func (c *Config) setDefaultConfig() {
+func (c *Config) SetDefaultConfig() {
 	c.DefaultSystemConfig = &c.SystemConfigs[0]
 	for i, systemConfig := range c.SystemConfigs {
 		if systemConfig.IsDefault {
--- a/toolkit/tools/imagegen/configuration/configuration.go
+++ b/toolkit/tools/imagegen/configuration/configuration.go
@@ -99,7 +99,7 @@ func (c *Config) GetBootPartition() (partitionIndex int, partition *Partition) {
 	for i, d := range c.Disks {
 		for j, p := range d.Partitions {
 			if p.HasFlag(PartitionFlagBoot) {
-				return j, &c.Disks[i].Partitions[j] 
+				return j, &c.Disks[i].Partitions[j]
 			}
 		}
 	}
--- a/toolkit/tools/imagegen/configuration/configuration.go
+++ b/toolkit/tools/imagegen/configuration/configuration.go
@@ -95,6 +95,17 @@ func (c *Config) GetDiskContainingPartition(partition *Partition) (disk *Disk) {
 	return nil
 }
 
+func (c *Config) GetBootPartition() (partitionIndex int, partition *Partition) {
+	for i, d := range c.Disks {
+		for j, p := range d.Partitions {
+			if p.HasFlag(PartitionFlagBoot) {
+				return j, &c.Disks[i].Partitions[j] 
+			}
+		}
+	}
+	return
+}
+
 // GetKernelCmdLineValue returns the output of a specific option setting in /proc/cmdline
 func GetKernelCmdLineValue(option string) (cmdlineValue string, err error) {
 	const cmdlineFile = "/proc/cmdline"
--- a/toolkit/tools/imagegen/configuration/systemconfig.go
+++ b/toolkit/tools/imagegen/configuration/systemconfig.go
@@ -70,6 +70,9 @@ func (s *SystemConfig) IsValid() (err error) {
 	// IsDefault must be validated by a parent struct
 
 	// Validate BootType
+	// if (s.BootType != "efi" && s.BootType != "legacy") {
+	// 	return fmt.Errorf("invalid [BootType]: %s. Expecting values of either 'efi' or 'legacy'.", s.Hostname)
+	// }
 
 	// Validate HostName
 	if (!govalidator.IsDNSName(s.Hostname) || strings.Contains(s.Hostname, "_")) && s.Hostname != "" {
--- a/toolkit/tools/imagegen/configuration/systemconfig.go
+++ b/toolkit/tools/imagegen/configuration/systemconfig.go
@@ -69,11 +69,6 @@ func (s *SystemConfig) GetMountpointPartitionSetting(mountPoint string) (partiti
 func (s *SystemConfig) IsValid() (err error) {
 	// IsDefault must be validated by a parent struct
 
-	// Validate BootType
-	// if (s.BootType != "efi" && s.BootType != "legacy") {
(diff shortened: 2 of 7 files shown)


toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -70,6 +70,9 @@ func (s *SystemConfig) IsValid() (err error) {
 	// IsDefault must be validated by a parent struct
 
 	// Validate BootType
+	// if (s.BootType != "efi" && s.BootType != "legacy") {
```
BettyRain, line 73: Is it to uncomment one day? Or just forgot to remove?

BettyRain (approved):

toolkit/tools/liveinstaller/liveinstaller.go
```diff
@@ -119,19 +125,69 @@ func installerFactory(forceAttended bool, configFile, templateConfigFile string)
...
+		return
+	}
+
+	err = removeOldMarinerBootTargets()
+	if (err != nil) {
+		return
+	}
+
+	logger.Log.Info("Setting Boot Order")
+	const squashErrors = false
```
dmcilvaney, line 152: nit: I think we generally put const variables at the top of the function (unless you have seen this elsewhere, then ignore)

SeanDougherty: Will change make this change! Thank you!

toolkit/tools/imagegen/configuration/systemconfig.go
SeanDougherty: Ahh, thank you! This was from when I was trying to test config validation. It didn't seem appropriate for this PR though. I will remove it, thanks Betty!

dmcilvaney: Could you also add a test case for these in either `systemconfig_test` or `imageconfigvalidator_test` (You can check the test coverage with `sudo make go-test-coverage`, it spits out a nice html file with everything).

toolkit/tools/imagegen/configuration/configuration.go
```diff
@@ -95,6 +95,17 @@ func (c *Config) GetDiskContainingPartition(partition *Partition) (disk *Disk) {
 	return nil
 }
 
+func (c *Config) GetBootPartition() (partitionIndex int, partition *Partition) {
```
dmcilvaney, line 98: Could you add a test case here for this (see other comment for details)?

toolkit/tools/imagegen/configuration/systemconfig.go
dmcilvaney: This seems like a legitimate thing to validate, if you think there is value feel free to leave it in place. Ideally our whole schema should be validated (its a long ways to go still though).

toolkit/tools/liveinstaller/liveinstaller.go
```diff
@@ -119,19 +125,69 @@ func installerFactory(forceAttended bool, configFile, templateConfigFile string)
...
+	program, commandArgs := formatBootEntryCreationCommand(installDetails)
+	err = shell.ExecuteLive(squashErrors, program, commandArgs...)
+
+	return
+}
+
+func formatBootEntryCreationCommand(installDetails installationDetails) (program string, commandArgs []string) {
+	program = "efibootmgr"
```
Camelron, line 160: This logic seems sound to me. Question: how do we know that the userspace app efibootmgr will be installed at the point that the installer is running? I think I'm misunderstanding when this liveinstaller.go code is running.

toolkit/tools/imagegen/configuration/systemconfig.go
SeanDougherty: How should we handle the situation where BootType is not set in the imageconfig? As the validation is currently "written", it will throw a fit if BootType is not set and I imagine this will break a lot of people's imageconfigs and upset a bunch of people

toolkit/tools/liveinstaller/liveinstaller.go
SeanDougherty: runliveinstaller runs liveinstaller.go and is an initrd environment created with the packages from iso-initrd-packages.json.   It seems efibootmgr has been part of that package list for a while, which thankfully meant we don't need to add it :) thank you for checking!

toolkit/tools/liveinstaller/liveinstaller.go
```diff
@@ -119,19 +125,69 @@ func installerFactory(forceAttended bool, configFile, templateConfigFile string)
...
 }
 
-func updateBootOrder() (err error) {
+func updateBootOrder(installDetails installationDetails) (err error) {
+	if (installDetails.finalConfig.DefaultSystemConfig.BootType != "efi") {
+		logger.Log.Info("No BootType of 'efi' detected. Not attempting to set EFI boot order.")
+		return
+	}
+
+	err = removeOldMarinerBootTargets()
```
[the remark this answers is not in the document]
SeanDougherty: good point! I will refactor so that they both make calls to shell.executelive() inside of their function body. Removing this awkward code splitting that's done with the Creation command! Thank you Cameron!

toolkit/tools/liveinstaller/liveinstaller.go
Camelron: Ah, that clears things up. Thanks for the info!

Camelron (approved):

toolkit/tools/imagegen/configuration/systemconfig.go
dmcilvaney: Talked offline, "" is a valid config in many situations. Maybe add some logic to check what those situations are. (maybe check if the install target is set for example?)

SeanDougherty: Validation added, and BootType testcase added as well!

SeanDougherty:
Kicked off new pipeline run, awaiting results before merging.

PR #4233, opened 2022-11-15
Add: Python package python3-grpcio

rikenm1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

One of the golden Container requires this python package.

###### Change Log  

- Change: Add second source tar to include git submodules.
- Change: Added python binding for gRPC. 

**YES/NO**
NO

###### Test Methodology

- Pipeline build id: 264839


SPECS/grpc/generate_source_tarball.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+
+set -e
+
+SRC_TARBALL="grpc-1.42.0.tar.gz"
```
nicogbg, line 5: why not using $PKG_VERSION instead of hard coding version number ?

SPECS/grpc/grpc.spec
```diff
@@ -1,30 +1,39 @@
...
-Release:        2%{?dist}
+Release:        3%{?dist}
 License:        ASL 2.0
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 Group:          Applications/System
 URL:            https://www.grpc.io
 Source0:        https://github.com/grpc/grpc/archive/v%{version}/%{name}-%{version}.tar.gz
-
+Source1:        %{name}-%{version}-submodules.tar.gz
```
nicogbg, line 11: in term of auto-upgrade it would be easier to incorporate the submodules into the Source0 tarball (1 tarball instead of using 2 tarballs)

nicogbg (changes requested):

SPECS/grpc/grpc.spec
rikenm1: I actually did that for my first PR but I was suggested by Pawel and Olivia to create a second source for the git submodules.

SPECS/grpc/grpc.spec
```diff
@@ -67,12 +88,29 @@ cmake ../.. -DgRPC_INSTALL=ON                \
...
-cd cmake/build
-%make_install
+pushd cmake/build
+%cmake_install
 find %{buildroot} -name '*.cmake' -delete
+popd
+#python
+pushd '%{buildroot}'
+PYROOT="${PWD}"
+popd
```
PawelWMS, line 111: I think I don't understand this part. Why do we need to go to `%{buildroot}` just to get the directory path? And then we set `PYROOT` but since it's not exported or set through `PYTROOT=abc %py3_install`, it's not being utilized by anything anyway. Could you check, if we actually need this part?
```suggestion
```

SPECS/grpc/grpc.spec
```diff
@@ -101,7 +139,16 @@ find %{buildroot} -name '*.cmake' -delete
 %license LICENSE
 %{_bindir}/grpc_*_plugin
 
+#imported from fedora
```
PawelWMS, line 142: Please just mention that we've taken the spec changes from Fedora X (where "X" is the version number) inside your changelog entry. We should also mention their license, which is MIT for Fedora specs.

PawelWMS (changes requested):
Main comment regarding source attribution to Fedora.

nicogbg (approved):

SPECS/grpc/grpc.spec
rikenm1: added comment in my change log

SPECS/grpc/generate_source_tarball.sh
rikenm1: made it an empty string.

SPECS/grpc/grpc.spec
rikenm1: removed..

SPECS/grpc/grpc.spec
```diff
@@ -101,7 +136,17 @@ find %{buildroot} -name '*.cmake' -delete
...
+
+%files -n python3-grpcio
+%license LICENSE
+%{python3_sitearch}/grpc
+%{python3_sitearch}/grpcio-%{version}-py%{python3_version}.egg-info
+
 %changelog
+* Wed Nov 09 2022 Riken Maharjan <rmaharjan@microsoft.com> - 1.42.0-3
+- Add python binding for gRPC.
+- CBL-Mariner import from Fedora 37 (license: MIT).
```
PawelWMS, line 148: A small clarification, because I agree that our way of adding specs is confusing at times: this particular wording is reserved for specs we've taken from other distros in whole. This is not a "full" import, we just added the `python3-grpcio` basing on Fedora's spec, so to avoid confusion about the source, let's be clearer on the change introduced in this PR. Here's my suggestion:
```suggestion
- Add 'python3-grpcio' subpackage using Fedora 37 spec for guidance.
```

PawelWMS (approved):

PR #4243, opened 2022-11-16
Add CONFIG_INITRD variable to configure stand-alone initrd image

gjswalling:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 
Needed for MarinerMOS build to configure initrd artifact, which will be directly booted from RAM using PXE.  This allows configuring the rootfs with startup services, authorized public keys, etc..

###### Change Log  
- Change toolkit, add CONFIG_INITRD variable to configure stand-alone initrd image

**NO** only toolkit, not toolchain

###### Test Methodology
- Pipeline build id: 1868


PawelWMS (changes requested):
Blocking to have time to discuss this offline. I feel like we might not need these changes, but I'm not entirely sure, what's the feature we're adding here.

toolkit/scripts/imggen.mk
```diff
@@ -166,7 +171,7 @@ $(image_external_package_cache_summary): $(cached_file) $(go-imagepkgfetcher) $(
 # Stand alone target to build just the initrd, should not be used in conjunction with other targets. Use the 'iso' target instead.
 initrd: $(go-liveinstaller) $(go-imager)
```
dmcilvaney, line 172: Add a dependency on `$(depends_CONFIG_INITRD)` here Make knows which components need to be rebuilt if the user changes the value of `CONFIG_INITRD`.

toolkit/scripts/utils.mk
```diff
@@ -42,7 +42,7 @@ endef
 ######## VARIABLE DEPENDENCY TRACKING ########
 
 # List of variables to watch for changes.
-watch_vars=PACKAGE_BUILD_LIST PACKAGE_REBUILD_LIST PACKAGE_IGNORE_LIST REPO_LIST CONFIG_FILE STOP_ON_PKG_FAIL TOOLCHAIN_ARCHIVE REBUILD_TOOLCHAIN SRPM_PACK_LIST SPECS_DIR
+watch_vars=PACKAGE_BUILD_LIST PACKAGE_REBUILD_LIST PACKAGE_IGNORE_LIST REPO_LIST CONFIG_FILE CONFIG_INITRD STOP_ON_PKG_FAIL TOOLCHAIN_ARCHIVE REBUILD_TOOLCHAIN SRPM_PACK_LIST SPECS_DIR
 # Current list: $(depend_PACKAGE_BUILD_LIST) $(depend_PACKAGE_REBUILD_LIST) $(depend_PACKAGE_IGNORE_LIST) $(depend_REPO_LIST) $(depend_CONFIG_FILE) $(depend_STOP_ON_PKG_FAIL)
 #					$(depend_TOOLCHAIN_ARCHIVE) $(depend_REBUILD_TOOLCHAIN) $(depend_SRPM_PACK_LIST) $(depend_SPECS_DIR)
```
dmcilvaney, line 47: Could you add `$(depends_CONFIG_INITRD)` to the comment here so it shows up in when you search for it (since the actual value is dynamicly generated)?

dmcilvaney (commented):
Could you clarify what the end goal here is? If you need an ISO installer with a non-standard initrd then this does seem to be a valid flow, but from my reading the goal here is to just build a customized initrd to boot and do arbitrary work.

If that is the case a call to `make image CONFIG_FILE=path/to/your/initrd_conf.json` should suffice (see `toolkit/resources/imageconfigs/iso_initrd.json` for an example of configuring an initrd image, specifically note the `StartupCommand:` entry and `/init` additional file).

Assuming the goal is the first option, I've added a few comments.

toolkit/scripts/imggen.mk
gjswalling: thanks!  Actually the 2nd option is what I'm going for and Pawel also suggested "make image" on my earlier PR attempt.  Closing this PR now as "make image" seems to be doing the trick!

PR #4249, opened 2022-11-17
Gate systemd's preset-all so it runs only on first install

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

When upgrading systemd, sometimes user customized services would be enabled/disabled unexpectedly. Our systemd package was calling `systemctl preset-all` during upgrade, which would clear any user customizations and reset all services back to their default enabled/disabled state.

We should only run this during the first install of systemd to guard against transaction ordering issues (say a transaction requires both a service and systemd, but rpm may occasionally choose a poor ordering even if the service has `Requires(post): systemd` if the transactions are complex). Once systemd is installed all future attempts to configure services during install should function normally.

###### Change Log  

- Gate `systemd`'s `%post: systemctl preset-all` behind a conditional that only runs on first install.

**YES**

###### Test Methodology

- Testing to follow

### DIFF
--- a/SPECS/systemd/systemd-bootstrap.spec
+++ b/SPECS/systemd/systemd-bootstrap.spec
@@ -1,7 +1,7 @@
 Summary:        Bootstrap version of systemd. Workaround for systemd circular dependency.
 Name:           systemd-bootstrap
 Version:        250.3
-Release:        7%{?dist}
+Release:        8%{?dist}
 License:        LGPLv2+ AND GPLv2+ AND MIT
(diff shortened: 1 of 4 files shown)


nicogbg (approved):

hbeberman (approved):

ddstreet:
re: requested review, this LGTM, I approve

I don't appear to be on the mariner devs github team so I don't think I am able to give official review yet ;-)

PR #4253, opened 2022-11-17
bash: change requires to package names

anphel31:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Modify bash requires to use package names instead of hard-coded paths (like /bin/mv and /bin/cp). Should resolve build break in libguestfs package that was caused by a recent change in coreutils (the /bin/mv binary moved to /usr/bin/mv)

###### Change Log  

- Change bash.spec to require coreutils (for mv and cp binaries) and grep (for grep binary)

**YES**

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/SPECS/bash/bash.spec
+++ b/SPECS/bash/bash.spec
@@ -14,8 +14,8 @@ BuildRequires:  readline
 Requires:       readline
 Requires(post): coreutils
 Requires(post): grep
-Requires(postun): grep
 Requires(postun): coreutils
+Requires(postun): grep
 Provides:       /bin/sh
 Provides:       /bin/bash
 Provides:       %{_bindir}/sh
--- a/SPECS/bash/bash.spec
+++ b/SPECS/bash/bash.spec
@@ -1,7 +1,7 @@
 Summary:        Bourne-Again SHell
 Name:           bash
 Version:        5.1.8
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv3
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -12,10 +12,10 @@ Source1:        bash_completion
 Patch0:         bash-5.1.patch
 BuildRequires:  readline
 Requires:       readline
-Requires(post): /bin/cp
-Requires(post): /bin/grep
-Requires(postun): /bin/grep
-Requires(postun): /bin/mv
+Requires(post): coreutils
+Requires(post): grep
+Requires(postun): grep
+Requires(postun): coreutils
 Provides:       /bin/sh
 Provides:       /bin/bash
 Provides:       %{_bindir}/sh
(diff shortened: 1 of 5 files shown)


neha170 (commented):
Hi Andrew, thanks for fixing the break. I found /bin/grep in many other spec files e.g. WALinuxAgent, dracut, tcsh, zsh and resource-agents. Do you think those should be included too in the change? Similarly, /bin/mv in perl-XML-LibXML, thanks

rmhsawyer (commented):
Thanks for fixing this. Looks like /bin/ln is also used in those packages and needs to be changed 
../SPECS/cracklib/cracklib.spec:17:Requires:         /bin/ln
../SPECS/cracklib/cracklib.spec:18:Requires(post):   /bin/ln
../SPECS/ca-certificates/ca-certificates.spec:73:BuildRequires:  /bin/ln

dmcilvaney:
Moving to `/usr/bin` brings us in alignment with Fedora here: https://docs.fedoraproject.org/en-US/packaging-guidelines/#_file_and_directory_dependencies

They do say that you should NOT use directory macros here (`%{_bindir}` etc ) so that isn't a way forward.

dmcilvaney:
I think using the path versions are useful when we have multiple packages providing the same file (busybox etc?). But it seems like we don't have multiple copies of `/bin/mv`.

SPECS/bash/bash.spec
```diff
@@ -12,10 +12,10 @@ Source1:        bash_completion
 Patch0:         bash-5.1.patch
 BuildRequires:  readline
 Requires:       readline
-Requires(post): /bin/cp
-Requires(post): /bin/grep
-Requires(postun): /bin/grep
-Requires(postun): /bin/mv
+Requires(post): coreutils
```
PawelWMS, line 15: I'm fine with this change but file paths are used across many specs and in general should work, so we might need to think of a more future-proof solution, so that our package declare provides for directories, that are just symbolic links like `/bin` or `/lib`. An update to the steps ran after the `%install` section might do the trick. That's out of scope for this PR, though, but I was curious of your opinion. If you agree, I can add a bug for us to address this in the future.

PawelWMS (approved):

SPECS/bash/bash.spec
dmcilvaney: The Fedora packaging guide is fine with paths, but ONLY to `/usr/bin` (among some others).  I think the big issue here is we have a mixture of `/bin` and `/usr/bin`. Taking with someone on another team who ran into this a while ago `dnf` handles these mismatches properly, and either follows the symlinks or has some other logic to figure this out. `tdnf` however does not implement t his.

anphel31:
I'm holding off on this PR for now since the breaking change in #4255 was reverted

PR #4272, opened 2022-11-22
change binary install location in coretuils package

rmhsawyer:
###### Merge Checklist  

- [x `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [] Ready to merge

---

###### Summary 

- This PR is trying to move all the binaries back into default install location from /bin/ to /usr/bin/
- We moved all the binaries to /bin at frist place as we followed Centos coreutils.spec but later they moved all the binaries into /usr/bin/ [upstream PR](https://github.com/vmware/photon/commit/3a7b668080aefefa61f30f42d165d51b7283a5f0#diff-2c1479545d88b270d5949996da652f11dc9feae915beba7eaef20c2efc2970af)

###### Change Log  

- remove the code that moves binaries from /usr/bin to /bin
- update the requires section in other related packages that using arbitrary binary path to coreutils package name 

YES

###### Associated issues  

- [Bug 41031409](https://microsoft.visualstudio.com/OS/_workitems/edit/41031409): tdnf coreutils package dependency resolution

###### Test Methodology

- Local toolchain and package build 
- Pipeline build id: in progress


SPECS/coreutils/coreutils.spec
```diff
@@ -92,7 +87,6 @@ LANGUAGE=en_US.UTF-8 LC_ALL=en_US.UTF-8 make -k check
 %files
 %defattr(-,root,root)
 %license COPYING
-/bin/*
```
PawelWMS, line 95: When I think about it now: what's the problem with `/bin`? It seems we have now 25 packages providing something under the `/bin` path, 33 requiring something from that path for run-time and 2 for build-time. Some from the same spec (18 core specs in total). Some of the specs we import from other distros might also have similar dependencies.

With that in mind, I'm curious, if it wouldn't be safer for the duration of 2.0 to provide both `/bin` and `/usr/bin` instead of updating the dependencies. Would it work, if you linked `ln -s /usr/bin %{buildroot}/bin` and left the `/bin/*` in the `%files` section intact? It's possible you might need to also `%exclude /bin` in case the build will complain that the directory itself is not listed.

rmhsawyer: I agree is a risky change and I am going to close this PR and create a new one with changes you suggested

PR #4279, opened 2022-11-23
Add srpmdownloader to address DOWNLOAD_SRPMS 

rlmenge:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add `srpmdownloader` tool to fix the DOWNLOAD_SRPMS flag. 
This flag can be used with SRPM_URL_LIST and/or SRPM_PACK_LIST. 
`srpmdownloader` is needed so that the rpm query for the spec's
version is able to parse the macros within the spec file. In order to
successfully parse the macros, we need to be within a chroot that has
Mariner's macros installed.

In order to avoid code duplication between `srpmpacker` and `srpmdownloader`, 
create `srpm` and tests within `srpm_test`. 

`srpmpacker` functions that moved and were given tests
- removeDuplicateStrings -> sliceutils library function RemoveDuplicateStrings
- findSPECFiles -> srpm library function FindSPECFiles 
- parsePackListFile ->  srpm library function ParsePackListFile 
- createChroot ->  srpm library function CreateSRPMChroot

###### Change Log  

- Add srpm library package and tests
- Add srpmdownloader.go

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/40942962/

###### Test Methodology

- AMD: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=268205&view=results
- ARM: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=268240&view=results
- local tests

### DIFF
--- a/toolkit/scripts/srpm_pack.mk
+++ b/toolkit/scripts/srpm_pack.mk
@@ -51,23 +51,21 @@ $(BUILD_SRPMS_DIR): $(STATUS_FLAGS_DIR)/build_srpms.flag
 	@echo Finished updating $@
 
 ifeq ($(DOWNLOAD_SRPMS),y)
-$(STATUS_FLAGS_DIR)/build_srpms.flag: $(LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(SPECS_DIR)
-	for spec in $(LOCAL_SPECS); do \
-		spec_file=$${spec} && \
-		srpm_file=$$(rpmspec -q $${spec_file} --srpm --define='with_check 1' --define='dist $(DIST_TAG)' --queryformat %{NAME}-%{VERSION}-%{RELEASE}.src.rpm) && \
-		for url in $(SRPM_URL_LIST); do \
-			wget $${url}/$${srpm_file} \
-				-O $(BUILD_SRPMS_DIR)/$${srpm_file} \
-				--no-verbose \
-				$(if $(TLS_CERT),--certificate=$(TLS_CERT)) \
-				$(if $(TLS_KEY),--private-key=$(TLS_KEY)) \
-				&& \
-			touch $(BUILD_SRPMS_DIR)/$${srpm_file} && \
-			break; \
-		done || $(call print_error,Loop in $@ failed) ; \
-		{ [ -f $(BUILD_SRPMS_DIR)/$${srpm_file} ] || \
-			$(call print_error,Failed to download $${srpm_file});  } \
-	done || $(call print_error,Loop in $@ failed) ; \
+$(STATUS_FLAGS_DIR)/build_srpms.flag: $(chroot_worker) $(LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(SPECS_DIR) $(go-srpmdownloader) $(srpm_pack_list_file)
+	srpm_file=$$($(go-srpmdownloader) \
+		--dir=$(SPECS_DIR) \
+		--output-dir=$(BUILD_SRPMS_DIR) \
+		--dist-tag=$(DIST_TAG) \
+		--srpm-url-list="$(SRPM_URL_LIST)" \
+		--ca-cert=$(CA_CERT) \
+		--tls-cert=$(TLS_CERT) \
+		--tls-key=$(TLS_KEY) \
+		--build-dir=$(SRPM_BUILD_CHROOT_DIR) \
+		--worker-tar=$(chroot_worker)  \
+		$(if $(SRPM_PACK_LIST),--srpm-list=$(srpm_pack_list_file)) \
+		$(if $(filter y,$(RUN_CHECK)),--run-check) \
+		--log-file=$(SRPM_BUILD_LOGS_DIR)/srpmdownloader.log \
+		--log-level=$(LOG_LEVEL) )  && \
 	touch $@
 
 # Since all the SRPMs are being downloaded by the "input-srpms" target there is no need to differentiate toolchain srpms.
--- a/toolkit/scripts/srpm_pack.mk
+++ b/toolkit/scripts/srpm_pack.mk
@@ -51,7 +51,7 @@ $(BUILD_SRPMS_DIR): $(STATUS_FLAGS_DIR)/build_srpms.flag
 	@echo Finished updating $@
 
 ifeq ($(DOWNLOAD_SRPMS),y)
-$(STATUS_FLAGS_DIR)/build_srpms.flag: $(chroot_worker) $(LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(SPECS_DIR) $(go-srpmdownloader) $(srpm_pack_list_file)
+$(STATUS_FLAGS_DIR)/build_srpms.flag: $(chroot_worker) $(LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(SPECS_DIR) $(depend_SPEC_DIRS) $(go-srpmdownloader) $(srpm_pack_list_file)
 	srpm_file=$$($(go-srpmdownloader) \
 		--dir=$(SPECS_DIR) \
 		--output-dir=$(BUILD_SRPMS_DIR) \
--- a/toolkit/tools/internal/sliceutils/sliceutils.go
+++ b/toolkit/tools/internal/sliceutils/sliceutils.go
@@ -55,3 +55,21 @@ func StringsSetToSlice(inputSet map[string]bool) []string {
 
 	return outputSlice[:index]
 }
+
+// RemoveDuplicateStrings will remove duplicate entries from a string slice
+func RemoveDuplicateStrings(packList []string) (deduplicatedPackList []string) {
+	var (
+		packListSet = make(map[string]struct{})
+		exists      = struct{}{}
+	)
+
+	for _, entry := range packList {
+		packListSet[entry] = exists
+	}
+
+	for entry := range packListSet {
+		deduplicatedPackList = append(deduplicatedPackList, entry)
+	}
+
+	return
+}
--- a/toolkit/tools/internal/sliceutils/sliceutils.go
+++ b/toolkit/tools/internal/sliceutils/sliceutils.go
@@ -58,18 +58,9 @@ func StringsSetToSlice(inputSet map[string]bool) []string {
 
 // RemoveDuplicateStrings will remove duplicate entries from a string slice
 func RemoveDuplicateStrings(packList []string) (deduplicatedPackList []string) {
-	var (
-		packListSet = make(map[string]struct{})
-		exists      = struct{}{}
-	)
-
+	packListSet := map[string]bool{}
 	for _, entry := range packList {
-		packListSet[entry] = exists
+		packListSet[entry] = true
 	}
-
-	for entry := range packListSet {
-		deduplicatedPackList = append(deduplicatedPackList, entry)
-	}
-
-	return
+	return StringsSetToSlice(packListSet)
 }
--- a/toolkit/tools/internal/srpm/srpm.go
+++ b/toolkit/tools/internal/srpm/srpm.go
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
 package srpm
 
 import (
--- a/toolkit/tools/internal/srpm/srpm.go
+++ b/toolkit/tools/internal/srpm/srpm.go
@@ -1,0 +1,147 @@
+package srpm
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/buildpipeline"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/directory"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/sliceutils"
+)
+
+// ParsePackListFile will parse a list of packages to pack if one is specified.
+// Duplicate list entries in the file will be removed.
+func ParsePackListFile(packListFile string) (packList []string, err error) {
+	if packListFile == "" {
+		return
+	}
+
+	file, err := os.Open(packListFile)
+	if err != nil {
+		return
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if line != "" {
+			packList = append(packList, line)
+		}
+	}
+
+	if len(packList) == 0 {
+		err = fmt.Errorf("cannot have empty pack list (%s)", packListFile)
+	}
+
+	packList = sliceutils.RemoveDuplicateStrings(packList)
+
+	return
+}
+
+// FindSPECFiles finds all SPEC files that should be considered for packing.
+// Takes into consideration a packList if provided.
+func FindSPECFiles(specsDir string, packList []string) (specFiles []string, err error) {
+	if len(packList) == 0 {
+		specSearch := filepath.Join(specsDir, "**/*.spec")
+		specFiles, err = filepath.Glob(specSearch)
+	} else {
+		for _, specName := range packList {
+			var specFile []string
+
+			specSearch := filepath.Join(specsDir, fmt.Sprintf("**/%s.spec", specName))
+			specFile, err = filepath.Glob(specSearch)
+
+			// If a SPEC is in the pack list, it must be packed.
+			if err != nil {
+				return
+			}
+			if len(specFile) != 1 {
+				if strings.HasPrefix(specName, "msopenjdk-11") {
+					logger.Log.Debugf("Ignoring missing match for '%s', which is externally-provided and thus doesn't have a local spec.", specName)
+					continue
+				} else {
+					err = fmt.Errorf("unexpected number of matches (%d) for spec file (%s)", len(specFile), specName)
+					return
+				}
+			}
+
+			specFiles = append(specFiles, specFile[0])
+		}
+	}
+
+	return
+}
+
+// createChroot creates a chroot to pack SRPMs inside of.
+func CreateChroot(workerTar, buildDir, outDir, specsDir string) (chroot *safechroot.Chroot, newBuildDir, newOutDir, newSpecsDir string, err error) {
+	const (
+		chrootName       = "srpmpacker_chroot"
+		existingDir      = false
+		leaveFilesOnDisk = false
+
+		outMountPoint    = "/output"
+		specsMountPoint  = "/specs"
+		buildDirInChroot = "/build"
+	)
+
+	extraMountPoints := []*safechroot.MountPoint{
+		safechroot.NewMountPoint(outDir, outMountPoint, "", safechroot.BindMountPointFlags, ""),
+		safechroot.NewMountPoint(specsDir, specsMountPoint, "", safechroot.BindMountPointFlags, ""),
+	}
+
+	extraDirectories := []string{
+		buildDirInChroot,
+	}
+
+	newBuildDir = buildDirInChroot
+	newOutDir = outMountPoint
+	newSpecsDir = specsMountPoint
+
+	chrootDir := filepath.Join(buildDir, chrootName)
+	chroot = safechroot.NewChroot(chrootDir, existingDir)
+
+	err = chroot.Initialize(workerTar, extraDirectories, extraMountPoints)
+	if err != nil {
+		return
+	}
+
+	defer func() {
+		if err != nil {
+			closeErr := chroot.Close(leaveFilesOnDisk)
+			if closeErr != nil {
+				logger.Log.Errorf("Failed to close chroot, err: %s", closeErr)
+			}
+		}
+	}()
+
+	// If this is container build then the bind mounts will not have been created.
+	if !buildpipeline.IsRegularBuild() {
+		// Copy in all of the SPECs so they can be packed.
+		specsInChroot := filepath.Join(chroot.RootDir(), newSpecsDir)
+		err = directory.CopyContents(specsDir, specsInChroot)
+		if err != nil {
+			return
+		}
+
+		// Copy any prepacked srpms so they will not be repacked.
+		srpmsInChroot := filepath.Join(chroot.RootDir(), newOutDir)
+		err = directory.CopyContents(outDir, srpmsInChroot)
+		if err != nil {
+			return
+		}
+	}
+
+	// Networking support is needed to download sources.
+	files := []safechroot.FileToCopy{
+		{Src: "/etc/resolv.conf", Dest: "/etc/resolv.conf"},
+	}
+
+	err = chroot.AddFiles(files...)
+	return
+}
--- a/toolkit/tools/internal/srpm/srpm.go
+++ b/toolkit/tools/internal/srpm/srpm.go
@@ -14,36 +14,6 @@ import (
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/sliceutils"
 )
 
-// ParsePackListFile will parse a list of packages to pack if one is specified.
-// Duplicate list entries in the file will be removed.
-func ParsePackListFile(packListFile string) (packList []string, err error) {
-	if packListFile == "" {
-		return
-	}
-
-	file, err := os.Open(packListFile)
-	if err != nil {
-		return
-	}
-	defer file.Close()
-
-	scanner := bufio.NewScanner(file)
-	for scanner.Scan() {
-		line := strings.TrimSpace(scanner.Text())
-		if line != "" {
-			packList = append(packList, line)
-		}
-	}
-
-	if len(packList) == 0 {
-		err = fmt.Errorf("cannot have empty pack list (%s)", packListFile)
-	}
-
-	packList = sliceutils.RemoveDuplicateStrings(packList)
-
-	return
-}
-
 // FindSPECFiles finds all SPEC files that should be considered for packing.
 // Takes into consideration a packList if provided.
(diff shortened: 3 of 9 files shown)


toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,233 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
rlmenge, line 2: Any tests would require setting some sort of localhost server as such: https://medium.com/zus-health/mocking-outbound-http-requests-in-go-youre-probably-doing-it-wrong-60373a38d2aa . But not sure how to guarantee security. Also may be better suited for network.go tests

toolkit/scripts/srpm_pack.mk
```diff
@@ -51,23 +51,21 @@ $(BUILD_SRPMS_DIR): $(STATUS_FLAGS_DIR)/build_srpms.flag
...
-				$(if $(TLS_CERT),--certificate=$(TLS_CERT)) \
-				$(if $(TLS_KEY),--private-key=$(TLS_KEY)) \
-				&& \
-			touch $(BUILD_SRPMS_DIR)/$${srpm_file} && \
-			break; \
-		done || $(call print_error,Loop in $@ failed) ; \
-		{ [ -f $(BUILD_SRPMS_DIR)/$${srpm_file} ] || \
-			$(call print_error,Failed to download $${srpm_file});  } \
-	done || $(call print_error,Loop in $@ failed) ; \
+$(STATUS_FLAGS_DIR)/build_srpms.flag: $(chroot_worker) $(LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(SPECS_DIR) $(go-srpmdownloader) $(srpm_pack_list_file)
```
PawelWMS, line 54: I know it used to be this way, but I'm curious if we shouldn't add a few `depend_*` targets as dependencies.
```suggestion
$(STATUS_FLAGS_DIR)/build_srpms.flag: $(chroot_worker) $(LOCAL_SPECS) $(depend_LOCAL_SPECS) $(LOCAL_SPEC_DIRS) $(depend_LOCAL_SPEC_DIRS)  $(SPECS_DIR) $(depend_SPECS_DIR) $(go-srpmdownloader) $(srpm_pack_list_file)
```

Not a blocker.

toolkit/tools/internal/sliceutils/sliceutils.go
```diff
@@ -55,3 +55,21 @@ func StringsSetToSlice(inputSet map[string]bool) []string {
...
+
+	for _, entry := range packList {
+		packListSet[entry] = exists
+	}
+
+	for entry := range packListSet {
+		deduplicatedPackList = append(deduplicatedPackList, entry)
+	}
+
+	return
```
PawelWMS, line 74: An attempt to simplify:
```suggestion
	packListSet := map[string]bool{}

	for _, entry := range packList {
		packListSet[entry] = true
	}

	return StringsSetToSlice(packListSet)
```

toolkit/scripts/srpm_pack.mk
dmcilvaney: `LOCAL_SPECS_DIR` shouldn't be something set by the user (ie probably lowercase). I've made that change as part of this PR https://github.com/microsoft/CBL-Mariner/pull/4181/files. I'll fixt his when I merge that in.

Agreed that we could use `$(depends_SPECS_DIR)` though.

toolkit/tools/internal/sliceutils/sliceutils.go
rlmenge: Tested and confirmed this works :)

SeanDougherty (commented):
This looks good to me! I will let someone more senior give the official approval, as I am still ramping up on our tools.

toolkit/tools/internal/sliceutils/sliceutils.go
```diff
@@ -55,3 +55,12 @@ func StringsSetToSlice(inputSet map[string]bool) []string {
...
 }
+
+// RemoveDuplicateStrings will remove duplicate entries from a string slice
+func RemoveDuplicateStrings(packList []string) (deduplicatedPackList []string) {
+	packListSet := map[string]bool{}
+	for _, entry := range packList {
+		packListSet[entry] = true
+	}
+	return StringsSetToSlice(packListSet)
+}
```
PawelWMS, line 66: One thing I missed with my previous suggestion is re-naming the variables to something more generic, since `sliceutils` is just a helper for all kinds of slices:
```suggestion
func RemoveDuplicateStrings(stringsSlice []string) []string {
	stringsSet := map[string]bool{}
	for _, entry := range stringsSlice {
		stringsSet[entry] = true
	}
	return StringsSetToSlice(stringsSet)
}
```

toolkit/tools/internal/srpm/srpm.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if len(packList) == 0 {
+		err = fmt.Errorf("cannot have empty pack list (%s)", packListFile)
+	}
+
+	packList = sliceutils.RemoveDuplicateStrings(packList)
+
+	return
+}
```
PawelWMS, line 82: Just a suggestion, since I noticed an opportunity to fix something that's been bugging me for a while.:)

```suggestion
func ParsePackListFile(packListFile string) (packList []string, err error) {
	if packListFile == "" {
		return
	}
	
	packSet := map[string]bool{}

	file, err := os.Open(packListFile)
	if err != nil {
		return
	}
	defer file.Close()

	scanner := bufio.NewScanner(file)
	for scanner.Scan() {
		line := strings.TrimSpace(scanner. Text())
		if line != "" {
			packSet[line] = true
		}
	}

	if len(packSet) == 0 {
		err = fmt.Errorf("cannot have empty pack list (%s)", packListFile)
	}

	packList = sliceutils.StringsSetToSlice(packSet)

	return
}
```

toolkit/tools/internal/srpm/srpm.go
```diff
@@ -0,0 +1,150 @@
...
+
+// createSRPMChroot creates a chroot to pack SRPMs inside of.
```
PawelWMS, line 84:
```suggestion
// CreateSRPMChroot creates a chroot to pack SRPMs inside of.
```

toolkit/tools/internal/srpm/srpm_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	fileAPtr, err := os.Create(filepathA)
+	assert.NoError(t, err)
+	defer os.Remove(filepathA)
+	defer fileAPtr.Close()
+
+	fileBPtr, err := os.Create(filepathB)
+	assert.NoError(t, err)
+	defer os.Remove(filepathB)
+	defer fileBPtr.Close()
```
PawelWMS, line 42: For test files we tend to use just checked-in ones. This way we can also reduce the amount of boilerplate and clean-up code in the tests and focus better on the tested scenario. If you don't mind, could we convert this set-up into 2 dummy spec files?

toolkit/tools/internal/srpm/srpm_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	os.Mkdir(specDir, 0777)
+	defer os.Remove(specDir)
+
+	fileAPtr, err := os.Create(filepathA)
+	assert.NoError(t, err)
+	err = os.WriteFile(filepathA, d1, 0644)
+	assert.NoError(t, err)
+	defer os.Remove(filepathA)
+	defer fileAPtr.Close()
```
PawelWMS, line 110: Same comment here.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+	if *caCertFile != "" {
+		newCACert, err := ioutil.ReadFile(*caCertFile)
+		if err != nil {
+			logger.Log.Panicf("Invalid CA certificate (%s), error: %s", *caCertFile, err)
+		}
+
+		templateSrcConfig.caCerts.AppendCertsFromPEM(newCACert)
+	}
+
+	if *tlsClientCert != "" && *tlsClientKey != "" {
```
PawelWMS, line 88: >if *tlsClientCert != "" && *tlsClientKey != "" { [](http://example.com/codeflow?start=1&length=48)

Might be worth panicking/printing an error when only one of the values is set. I'm guessing that is invalid.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		cert, err := tls.LoadX509KeyPair(*tlsClientCert, *tlsClientKey)
+		if err != nil {
+			logger.Log.Panicf("Invalid TLS client key pair (%s) (%s), error: %s", *tlsClientCert, *tlsClientKey, err)
```
PawelWMS, line 91:
```suggestion
			logger.Log.Panicf("Invalid TLS client cert-key pair (%s)-(%s), error: %s", *tlsClientCert, *tlsClientKey, err)
```

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+
+const (
+	defaultBuildDir    = "./build/SRPMS"
+	defaultWorkerCount = "10"
+)
+
+type sourceRetrievalConfiguration struct {
+	sourceURLs []string
+	caCerts    *x509.CertPool
+	tlsCerts   []tls.Certificate
```
PawelWMS, line 37: Does this need to be an array? It seems like we have one cert at most.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		cert, err := tls.LoadX509KeyPair(*tlsClientCert, *tlsClientKey)
+		if err != nil {
+			logger.Log.Panicf("Invalid TLS client key pair (%s) (%s), error: %s", *tlsClientCert, *tlsClientKey, err)
+		}
+
+		templateSrcConfig.tlsCerts = append(templateSrcConfig.tlsCerts, cert)
+
+	}
+	if *srpmUrlList != "" {
+		// Assumes that srpmUrlList come in as ',' seperated
```
PawelWMS, line 98: >// Assumes that srpmUrlList come in as ',' seperated [](http://example.com/codeflow?start=2&length=52)

Now I'm confused. Is the input separate with commas or spaces?

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+	// A pack list may be provided, if so only pack this subset.
+	// If none is provided, pack all srpms.
+	srpmList, err := srpm.ParsePackListFile(*srpmListFile)
+	logger.PanicOnError(err)
+
+	logger.Log.Infof("SRPM list %s", srpmList)
+
+	err = getSRPMQueryWrapper(*specsDir, *distTag, *buildDir, *outDir, *workerTar, *workers, *runCheck, srpmList, templateSrcConfig)
+	logger.PanicOnError(err)
+
```
PawelWMS, line 112:
```suggestion
```

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		if err != nil {
+			return
+		}
+		defer chroot.Close(leaveFilesOnDisk)
+	}
+
+	doCreateAll := func() error {
+		err = getSRPMQuery(specsDir, distTag, buildDir, outDir, workers, runCheck, srpmList, templateSrcConfig)
+		return err
+	}
```
PawelWMS, line 132:
```suggestion
	doCreateAll := func() error {
		return getSRPMQuery(specsDir, distTag, buildDir, outDir, workers, runCheck, srpmList, templateSrcConfig)
	}
```

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// If this is container build then the bind mounts will not have been created.
+	// Copy the chroot output to host output folder.
+	if !buildpipeline.IsRegularBuild() {
+		srpmsInChroot := filepath.Join(chroot.RootDir(), outDir)
+		err = directory.CopyContents(srpmsInChroot, originalOutDir)
+		if err != nil {
+			return err
+		}
```
PawelWMS, line 153: Simplification idea, but I see benefits of the current approach as well. Just something to consider:
```suggestion
```

PawelWMS: Alternatively, I'd print an error message saying we failed to perform the copy + print the source and destination paths.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	// A pack list may be provided, if so only pack this subset.
+	// If none is provided, pack all srpms.
+	srpmList, err := srpm.ParsePackListFile(*srpmListFile)
+	logger.PanicOnError(err)
+
+	logger.Log.Infof("SRPM list %s", srpmList)
+
+	err = getSRPMQueryWrapper(*specsDir, *distTag, *buildDir, *outDir, *workerTar, *workers, *runCheck, srpmList, templateSrcConfig)
```
PawelWMS, line 110: Have mercy, don't use a noun as a name for the function. Especially, if it begins with "get".;) To my main point: please use a name describing what's happening when we run this function.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+)
+
+type sourceRetrievalConfiguration struct {
+	sourceURLs []string
+	caCerts    *x509.CertPool
+	tlsCerts   []tls.Certificate
+}
+
+var (
+	app = kingpin.New("srpmpacker", "A tool to package a SRPM.")
```
PawelWMS, line 41: Now I'm confused. Do we run this tool to pack the specs into SRPMs, or to download pre-packed SRPMs from an external source?

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		return err
+	}
+
+	if chroot != nil {
+		logger.Log.Info("Grabbing SRPMs URL inside a chroot environment")
+		err = chroot.Run(doCreateAll)
+	} else {
+		logger.Log.Info("Grabbing SRPMs URL SRPMs in the host environment")
+		err = doCreateAll()
+	}
```
PawelWMS, line 140: Let's reduce the number of `if` checks a bit and keep const/var declarations in one place:
```suggestion
	const leaveFilesOnDisk = false
	var chroot *safechroot.Chroot

	originalOutDir := outDir

	doCreateAll := func() error {
		return getSRPMQuery(specsDir, distTag, buildDir, outDir, workers, runCheck, srpmList, templateSrcConfig)
	}

	if workerTar == "" {
		logger.Log.Info("Grabbing SRPMs' URLs in the host environment.")
		err = doCreateAll()
	else {
		chroot, buildDir, outDir, specsDir, err = srpm.CreateSRPMChroot(workerTar, buildDir, outDir, specsDir)
		if err != nil {
			return
		}
		defer chroot.Close(leaveFilesOnDisk)

		logger.Log.Info("Grabbing SRPMs' URLs inside a chroot environment.")
		err = chroot.Run(doCreateAll)
	}
```

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	// If this is container build then the bind mounts will not have been created.
+	// Copy the chroot output to host output folder.
+	if !buildpipeline.IsRegularBuild() {
+		srpmsInChroot := filepath.Join(chroot.RootDir(), outDir)
```
PawelWMS, line 149: It looks like we expect `chroot != nil` for container builds. If that's supposed to be always the case, let's add a check somewhere to make sure `buildpipeline.IsRegularBuild() || workerTar != ""`.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		err = directory.CopyContents(srpmsInChroot, originalOutDir)
+		if err != nil {
+			return err
+		}
+	}
+
+	return
+}
+
+// getSRPMQuery queries for the name, version and release of the SRPM
```
PawelWMS, line 159: We use the singular form here but it looks like we're actually querying multiple specs. Let's update the comment to avoid confusion.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+func getSRPMQuery(specsDir, distTag, buildDir, outDir string, workers int, runCheck bool, srpmList []string, srcConfig sourceRetrievalConfiguration) (err error) {
```
PawelWMS, line 160: >nc getSRPMQuer [](http://example.com/codeflow?start=2&length=14)

The function name is misleading. We're not returning any queries here, we're downloading SRPMs. At least that's what I think we're doing.;)

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+	const (
+		emptyQueryFormat = ``
```
PawelWMS, line 162: Unused const.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		querySrpm        = `%{NAME}-%{VERSION}-%{RELEASE}.src.rpm`
```
PawelWMS, line 163: >querySrpm [](http://example.com/codeflow?start=2&length=9)

Nit-pick: I think this is closer to `srpmNameQueryFormat` or something along these lines. Also, if not the first word in the variable name, please use all-caps for abbreviations.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		packageSRPMs, err = rpm.QuerySPEC(specfile, sourcedir, querySrpm, arch, defines, rpm.QueryHeaderArgument)
+		if err != nil {
+			logger.Log.Warn(err)
+			return err
+		}
+		packageSRPM := packageSRPMs[0]
+		downloadSRPM(packageSRPM, outDir, srcConfig)
+	}
+	return err
+}
```
PawelWMS, line 191: [Ignorable] Minor clean-up:
```suggestion
	const (
		emptyQueryFormat = ``
		querySrpm        = `%{NAME}-%{VERSION}-%{RELEASE}.src.rpm`
	)
	// Find the general SPEC query arguments
	defines := rpm.DefaultDefines(runCheck)
	defines[rpm.DistTagDefine] = distTag

	arch, err := rpm.GetRpmArch(runtime.GOARCH)
	if err != nil {
		logger.Log.Warn(err)
		return
	}

	specFiles, err := srpm.FindSPECFiles(specsDir, srpmList)
	if err != nil {
		return
	}

	// Parse each SPEC for name and version and hydrate SRPM
	for _, specfile := range specFiles {
		var packageSRPMs []string
		sourcedir := filepath.Dir(specfile)
		packageSRPMs, err = rpm.QuerySPEC(specfile, sourcedir, querySrpm, arch, defines, rpm.QueryHeaderArgument)
		if err != nil {
			logger.Log.Warn(err)
			return err
		}

		downloadSRPM(packageSRPMs[0], outDir, srcConfig)
	}

	return
}
```

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+	// Parse each SPEC for name and version and hydrate SRPM
+	for _, specfile := range specFiles {
+		sourcedir := filepath.Dir(specfile)
+		var packageSRPMs []string
+		packageSRPMs, err = rpm.QuerySPEC(specfile, sourcedir, querySrpm, arch, defines, rpm.QueryHeaderArgument)
+		if err != nil {
+			logger.Log.Warn(err)
+			return err
+		}
+		packageSRPM := packageSRPMs[0]
```
PawelWMS, line 187: It might be worth adding a check for `len(packageSRPMs) != 1` to make sure we didn't get an unexpected number of results.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		downloadSRPM(packageSRPM, outDir, srcConfig)
+	}
+	return err
+}
+
+// downloadSRPM downloads SRPM trying each of srcConfig's urls
+func downloadSRPM(fileName string, newSourceDir string, srcConfig sourceRetrievalConfiguration) {
```
PawelWMS, line 194: I think `newSourceDir` is closer to `outputDir`.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+	destinationFile := filepath.Join(newSourceDir, fileName)
+
+	var err error
+	var urlChosen string
+	srpmDownloaded := false
+	for _, urlSrc := range srcConfig.sourceURLs {
+		if srpmDownloaded {
+			break
+		}
+		logger.Log.Debugf("Trying (%s) from (%s)", fileName, urlSrc)
```
PawelWMS, line 208: Missing word? "Trying to download"?

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		url := network.JoinURL(urlSrc, fileName)
+
+		err = retry.Run(func() error {
+			err := network.DownloadFile(url, destinationFile, srcConfig.caCerts, srcConfig.tlsCerts)
+			urlChosen = url
+			if err != nil {
+				logger.Log.Warnf("Failed to download (%s). Error: %s", url, err)
```
PawelWMS, line 215: I think failure is expected if we're trying multiple sources, so I wouldn't print this on the warning level. Right now we're going to end up generating a lot of noise when everything is working as intended. I'd suggest switching to debug level.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+				os.Remove(destinationFile)
```
PawelWMS, line 216: Not your fault but this is straight-up wrong. Could I ask you to update `DownloadFile`, so that it never leaves a file on the disk if the download failed? I'd like to prevent this way of using the function from spreading.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+			} else {
+				srpmDownloaded = true
+			}
+
+			return err
+		}, downloadRetryAttempts, downloadRetryDuration)
+	}
+
+	if !srpmDownloaded {
+		logger.Log.Panicf("All attempts failed to download (%s). Error: %s", fileName, err)
```
PawelWMS, line 226: >logger.Log.Panicf [](http://example.com/codeflow?start=2&length=17)

We're usually returning an error from helper functions deeper in the stack and allow the caller to decide what to do with it.

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+		return
+	}
+
+	logger.Log.Debugf("Hydrated (%s) from (%s)", fileName, urlChosen)
+
```
PawelWMS, line 231:
```suggestion
```

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+			if err != nil {
+				logger.Log.Warnf("Failed to download (%s). Error: %s", url, err)
+				os.Remove(destinationFile)
+			} else {
+				srpmDownloaded = true
+			}
+
+			return err
+		}, downloadRetryAttempts, downloadRetryDuration)
+	}
```
PawelWMS, line 223: Slight refactoring **suggestion**, mostly because it immediately triggered suspicion in my head to see a `if srpmDownloaded` check right after `srpmDownloaded := false`.;)
```suggestion
	for _, urlSrc := range srcConfig.sourceURLs {
		logger.Log.Debugf("Trying (%s) from (%s)", fileName, urlSrc)
		url := network.JoinURL(urlSrc, fileName)

		err = retry.Run(func() error {
			err := network.DownloadFile(url, destinationFile, srcConfig.caCerts, srcConfig.tlsCerts)
			urlChosen = url
			srpmDownloaded = (err == nil)
			if !srpmDownloaded {
				logger.Log.Warnf("Failed to download SRPM from URL (%s). Error: %s", url, err)
				os.Remove(destinationFile)
			}

			return err
		}, downloadRetryAttempts, downloadRetryDuration)

		if srpmDownloaded {
			break
		}
	}
```

PawelWMS (changes requested):
:clock1:

toolkit/tools/srpmdownloader/srpmdownloader.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// Create a template configuration that all packed SRPM will be based on.
+	var templateSrcConfig sourceRetrievalConfiguration
+	// Setup remote source configuration
+	var err error
+	templateSrcConfig.caCerts, err = x509.SystemCertPool()
+	logger.PanicOnError(err, "Received error calling x509.SystemCertPool(). Error: %v", err)
+	if *caCertFile != "" {
+		newCACert, err := ioutil.ReadFile(*caCertFile)
+		if err != nil {
```
PawelWMS, line 81: I think we can replace the `err != nil` check with `PanicOnError()`. Similar comment for the TLS cert check below.

PR #4326, opened 2022-11-30
Reduce initrd image size in Mariner 2.0

chalamalasetty:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Reduce initrd image size in Mariner 2.0 default ISO.

  Mariner 1.0 ISO initrd.img : ~449 MB
  Mariner 2.0 ISO initrd.img: ~ 854 MB

Llvm (110 MB), Linux-firmware (123 MB), Mesa-dri-drivers (99 MB) are bigger chunks getting added to 2.0 initrd.img.
o	Avoid pulling mesa-dri-drivers which includes llvm by removing recommends from qt5-qtbase-gui. 
o	Split linux-firmware into sub packages and add “qlogic” firmware into ISO initrd image which is needed.
o	Clean up doc, man, locales (expect en-US & defaults) etc.
o	Removed OpenSSH-Server, wget, few tools which are specific to PXE boot and not needed in default ISO.

With these changes Mariner 2.0 initrd.img size is down to ~444 MB from 854 MB. Verified ISO was able to load with 1GB RAM on a Gen2 VM and installed Mariner OS on the disk with UI installer.

Steps to build PXE boot needed initrd image from ISO:

•	Remove “calamares” and add  "awk", "gawk", "iputils", "mariner-repos-shared", "mdadm" and "wget" to [iso-initrd-packages.json](https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json). To add these two PRs (https://github.com/microsoft/CBL-Mariner/pull/3012) & (https://github.com/microsoft/CBL-Mariner/pull/4019)
•	Build ISO with Mariner toolkit.  ([Steps to build ISO](https://github.com/microsoft/CBL-MarinerTutorials/blob/main/docs/building/building.md#build-a-demo-iso))

###### Change Log  

- Change
- Change
- Change

**NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: Local build.


SPECS/calamares/calamares.spec
```diff
@@ -217,8 +219,8 @@ install -p -m 644 %{SOURCE53} %{buildroot}%{_sysconfdir}/calamares/mariner-eula
 %{_libdir}/cmake/Calamares/
 
 %changelog
-* Mon Jul 25 2022 Minghe Ren <mingheren@microsoft.com> - 3.2.11-39
-- Modify users.conf to imporve security
+* Mon Nov 28 2022 Suresh Babu Chalamalasetty <schalam@microsoft.com> - 3.2.11-40
```
christopherco, line 222: Preserve older change log entry as well

SPECS/linux-firmware/linux-firmware.spec
```diff
@@ -36,17 +60,40 @@ cp iwlwifi-8000C-*.ucode %{buildroot}%{_firmwarepath}
 
 %files
 %defattr(-,root,root)
-%license WHENCE
```
christopherco, line 39: WHENCE file should be included for licensing reasons. Probably should be included in all subpackages.

SPECS/qt5-qtbase/qt5-qtbase.spec
```diff
@@ -203,7 +206,7 @@ Requires: %{name}-devel%{?_isa} = %{version}-%{release}
 # debating whether to do 1 subpkg per library or not -- rex
 %package gui
 Summary: Qt5 GUI-related libraries
-Recommends: mesa-dri-drivers
+# Recommends: mesa-dri-drivers
```
christopherco, line 209: nit - remove commented out recommends

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
```diff
@@ -34,27 +32,20 @@
...
         "lz4",
         "ncurses",
         "ncurses-term",
-        "net-tools",
         "nspr",
         "nss",
-        "openssh-server",
         "openssl",
         "mariner-release",
-        "mariner-repos-shared",
```
christopherco, line 54: Do we need this `mariner-repos-shared` package for GPG key support in RPM/DNF?

christopherco (dismissed):

SPECS/linux-firmware/linux-firmware.spec
chalamalasetty: Included in all packages from linux-firmware.

SPECS/linux-firmware/linux-firmware.spec
```diff
@@ -17,6 +17,30 @@ BuildArch:      noarch
 %description
```
christopherco, line 17: Question - Today if a user installs `linux-firmware` RPM, they'll get all the firmware blobs. Is it possible to preserve this behavior now that we're splitting the firmwares out into sub-packages? (i.e., `linux-firmware` package itself has `Requires: linux-firmware-XXXXX` for each sub-package?)

Doing something like this may help prevent a breaking change since the user was originally expecting all the firmware blobs to be installed with `sudo dnf install linux-firmware`.

SPECS/calamares/calamares.spec
```diff
@@ -217,6 +219,9 @@ install -p -m 644 %{SOURCE53} %{buildroot}%{_sysconfdir}/calamares/mariner-eula
 %{_libdir}/cmake/Calamares/
 
 %changelog
+* Mon Nov 28 2022 Suresh Babu Chalamalasetty <schalam@microsoft.com> - 3.2.11-40
+- Avoid pulling unneeded dependencies by setting AutoReq to no
```
PawelWMS, line 223: RPM is usually pretty good at figuring out correct dependencies. It can get things wrong if we pack test and/or sample files having a dependency on some exotic tool, which is not needed for the core functionality, but then it's better to have RPM ignore only these specific files with `%global __requires_exclude_from` (example in `qt5-qtbase.spec`) instead of adding a blanked `AutoReq` to drop all auto-generated deps.

What were the unneeded packages in this case and what was pulling them in?

SPECS/qt5-qtbase/qt5-qtbase.spec
```diff
@@ -61,8 +61,11 @@ BuildRequires: zlib-devel
 BuildRequires: qt5-rpm-macros
 
 Requires:         icu
+Requires:         pcre2
+Requires:         dbus
 Requires(post):   chkconfig
 Requires(postun): chkconfig
+AutoReq:        no
```
PawelWMS, line 68: Similar question about `AutoReq` as in `calamares.spec`.

SPECS/qt5-qttools/qt5-qttools.spec
```diff
@@ -390,6 +394,11 @@ fi
 
 
 %changelog
+* Mon Nov 28 2022 Suresh Babu Chalamalasetty <schalam@microsoft.com> - 5.12.5-6
+- Add AutoReq to no to avoid pulling unneeded runtime dependencies and add build deps
```
PawelWMS, line 398:
```suggestion
- Add AutoReq to avoid pulling unneeded runtime dependencies and add build deps
```

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
henryli001: I would recommend to add a statement in the PR description to mention the next-step action items for supporting PXE boot scenarios so that other developers or customers won't get confused when they see that we're removing PXE-related components within this change.

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
```diff
@@ -65,13 +56,9 @@
         "shim",
         "sqlite",
         "systemd",
-        "tar",
         "tdnf",
-        "usbutils",
         "util-linux",
         "veritysetup",
-        "vim",
-        "wget",
```
henryli001, line 74: In the runliveinstaller script, we have implementation for PXE boot support: https://github.com/microsoft/CBL-Mariner/blob/32cf2a208bddb0bd9d3f5fdaac541ed6b20a2d6b/toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller#L72. And within this implementation, we're actually calling wget. Since we're removing PXE boot support from the default ISO and plan to add new ISO or instructions for PXE boot scenario, does it still make sense to keep this implementation within our runliveinstaller? For PXE boot, should we consider using a PXE-specific installer script?

SPECS/linux-firmware/linux-firmware.spec
chalamalasetty: I tried this approach adding all sub package as Requires but toolkit was throwing error while building graph for linux-firmware package.

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
chalamalasetty: Our plan is to create or provide instructions to customers on how to build ISO for PXE boot scenarios. Mostly provide instructions on how to build PXE boot ISO until we finish our investigations to separate out rootfs filesystem to squashfs format.

SPECS/calamares/calamares.spec
chalamalasetty: Mariner 2.0 toolkit is behaving slightly different than 1.0. Even if a dependency is added for BuildRequires it is pulling by default to Requires. To avoid this we need to add AutoReq to no.

SPECS/mesa/mesa.spec
```diff
@@ -106,6 +106,7 @@ BuildRequires:  pkgconfig(xxf86vm)
 BuildRequires:  pkgconfig(zlib) >= 1.2.3
 BuildRequires:  python3-devel
 BuildRequires:  python3-mako
+AutoReq:        no
```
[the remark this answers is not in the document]
chalamalasetty: Mariner 2.0 toolkit is behaving slightly different than 1.0. Even if a dependency is added for BuildRequires it is pulling by default to Requires. To avoid this we need to add AutoReq to no.

SPECS/qt5-qtbase/qt5-qtbase.spec
chalamalasetty: Provided update above is applicable here also.

SPECS/qt5-qttools/qt5-qttools.spec
```diff
@@ -1,27 +1,31 @@
...
+BuildRequires: dbus-devel
+
 BuildRequires: qt5-qtbase-private-devel
 # Qt macros
 BuildRequires: qt5-rpm-macros >= %{version}
 BuildRequires: qt5-qtbase-static >= %{version}
 BuildRequires: qt5-qtdeclarative-static >= %{version}
 
 Requires: %{name}-common = %{version}-%{release}
+AutoReq:        no
```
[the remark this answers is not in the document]
chalamalasetty: Provided update above is applicable here also.

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
chalamalasetty: By default it looks for default ISO else if look for imageconfig command. For this default ISO it will not go to else if and when customers use PXE boot supported ISO this is needed. Hence we can keep it this way.

SPECS/qt5-qtbase/qt5-qtbase.spec
```diff
@@ -738,6 +740,10 @@ fi
 %{_qt5_libdir}/cmake/Qt5Gui/Qt5Gui_QXdgDesktopPortalThemePlugin.cmake
 
 %changelog
+* Mon Nov 28 2022 Suresh Babu Chalamalasetty <schalam@microsoft.com> - 5.12.11-4
+- Add AutoReq to no to avoid pulling unneeded runtime dependencies
```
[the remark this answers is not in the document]
chalamalasetty: Updated

SPECS/qt5-qttools/qt5-qttools.spec
chalamalasetty: updated.

SPECS/linux-firmware/linux-firmware.spec
christopherco: Can we bottom out on the graph error? Thinking through the scenario more, I think we should find a way to preserve the current behavior where updated `linux-firmware` package still distributes the firmware blobs that were present before.

Consider the situation where this updated `linux-firmware` package with minimal firmware blobs is published, a functioning production system that performs a package update will lose its firmware blobs since the older `linux-firmware` package will be uninstalled on the update operation. One of those firmware blobs could be required for the system's networking or storage driver to function.

SPECS/calamares/calamares.spec
PawelWMS: That is not the case, as far as I could tell. Latest `calamares` package in PMC doesn't require any of the packages from the BRs (I like to run `dnf repoquery --latest-limit 1 --resolve --requires calamares` to get a nice list).

Building with `AutoReq: no` removes the following dependencies (ran the same DNF command but against the built RPM file):
```bash
calamares-libs
glibc
libgcc
libstdc++
qt5-qtbase
qt5-qtbase-gui
```

Especially worrying ones are `calamares-libs` and the `qt5-qtbase*`, since they are not part of Mariner's base images and thus their absence will cause run-time issues for anyone trying to use this package.

I think we're not taking the right approach here and RPM is working as intended identifying the dependencies. We should be using `AutoReq: no` very carefully/never and instead identify the package's components generating these dependencies and going from there. We can move them to a separate subpackage (if they are not required for the core functionality to work properly), try modifying the build options to rely on different packages/disable some features (if acceptable by our users), or look for yet another solution.

PawelWMS (changes requested):
Blocking on `AutoReq: no` - this will cause run-time issues and regressions.

SPECS/linux-firmware/linux-firmware.spec
chalamalasetty: Made required change and looks like it is working and no repro of unresolved dependency error. We need Requires: linux-firmware-XXXXX = %{version}-%{release}

SPECS/calamares/calamares.spec
chalamalasetty: Verified with local built ISO with these changes and no regressions seen. OS able to install with UI and terminal installer with no issues. Calamares-libs, qt5-qtbase and other runtime dependencies pulled correctly and isolated llvm which is not needed.

PawelWMS: Ah, yes, I missed the fact that the updated package still brings in `calamares-libs` - I apologize for the confusion.

You're right, installing the updated `calamares` will still install the current dependencies (minus `llvm`) but now indirectly, because `calamares-libs` still has a dependency on them (`AutoReq: no` applies only to the `calamares` package, not the other ones from `calamares.spec`). The binaries from the `calamares` package, however, still need the now-removed dependencies, though. Here's an example for `/usr/bin/calamares` (part of the `calamares` package) and the Qt5 libraries:

image

I think that `AutoReq: no` is a too heavy hammer for what we're trying to do here. While `calamares` and the ISO installation works for the scenario you've tested, the current dependency on `llvm` came from somewhere (it's probably not `calamares` itself) and whatever needed it will now be at risk of not functioning properly or at all. And I agree, I'm also curious of @dmcilvaney's opinion.

To summarize: if the goal is to remove the dependency on `llvm`, we should understand what's bringing it in (`dnf repoquery --requires --tree --resolve calamares` should help immensely) and remove that dependency in a surgical manner (move the offending script/binary to a subpackage, remove it completely, change the package's build steps, etc.), while allowing the RPM's automatic dependency resolution to do its work.

dmcilvaney: I've talked offline, agreed that a blanket `AutoReq:no` is too broad. It may seem to work, but I'd be much happier knowing exactly what files are causing this, and fixing those specific files. Using ldd is one way (as shown by Pawel, also see https://www.steventwheeler.com/linux/2018/10/18/debugging-rpm-dependencies.html). We can also set https://rpm-software-management.github.io/rpm/manual/dependency_generators.html `--rpmfcdebug` to debug the generators. I feel like the full llvm package should not be getting pulled, even if some related libraries ARE needed.

PawelWMS: One more thing I had time to look at: I think the dependency on `llvm` was being brought by `mesa-dri-drivers`. Since that dependency is being dropped from the Qt5 packages in this PR, there's a chance nothing more needs to be done to get rid of the unnecessary `llvm` dependency.

image

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
```diff
@@ -65,13 +56,9 @@
         "shim",
         "sqlite",
         "systemd",
-        "tar",
         "tdnf",
-        "usbutils",
         "util-linux",
         "veritysetup",
-        "vim",
```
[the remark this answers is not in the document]
SeanDougherty: Apologies for double-commenting on this with Daniel. I had forgotten to move my review from pending to submitted

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
```diff
@@ -65,13 +56,9 @@
         "shim",
         "sqlite",
         "systemd",
-        "tar",
         "tdnf",
```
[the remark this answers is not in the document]
chalamalasetty: tdnf is needed to for OS installation from initrd image.

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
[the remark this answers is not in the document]
chalamalasetty: I agree on this. Getting nano into initrd image.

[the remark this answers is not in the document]
chalamalasetty: Adding nano to initrd image

SPECS/calamares/calamares.spec
chalamalasetty: Yes, after further investigation it seems that llvm is brought by mesa-dri-drivers with qt5 gui subpackage, Updated PR to simplify changes.

PawelWMS (approved):

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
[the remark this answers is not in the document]
dmcilvaney: Some background here: The initrd installer acts the same way our normal installer works for offline builds. It configures a 'fake' package repo locally, and tdnf can install packages from that local repo instead of using online repos.

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
chalamalasetty: Added steps on how to build PXE boot needed initrd image with default ISO.

toolkit/resources/imageconfigs/packagelists/iso-initrd-packages.json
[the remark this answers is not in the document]
chalamalasetty: Added nano to initrd image,

toolkit/resources/imageconfigs/postinstallscripts/iso_initrd/cleanup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
dmcilvaney, line 1: Is it worth cleaning up the rpm cache here as well? We do that for containers.

dmcilvaney (approved):

toolkit/resources/imageconfigs/postinstallscripts/iso_initrd/cleanup.sh
chalamalasetty: Not seeing cache under /var in initrd image.

PR #4350, opened 2022-12-02
CGroupsV2 Toggle Optionality for Mariner 2.0 Images

aditjha-msft:
Draft PR for introducing config flag that would allow turning on cgroupv2 by default on boot for Mariner 2.0 images.

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR introduces the ability for Mariner 2.0 images to be built with cgroupv2 enabled. This is necessary because right now Mariner does not allow customers the option to use cgroupv2, but this has been enabled in Ubuntu for a while now. There is customer request and demand to start using cgroupv2 with Mariner. It allows for unified controller structure and important performance improvements with Kubernetes and in general usage.

###### Change Log  

The approach for having the ability for turning "on" cgroupv2 is through config flag changes. Just like there is a flag options for SELinux that goes under the _KernelCommandLine_ block in image configs, there will be a CGroup flag with options for cgroupv1 (default) or cgroupv2. The necessary code and workflow is part of this PR, and at the very end the flag option set in the image config gets placed in the grub.cfg file that is used to boot Mariner.

YES

###### Test Methodology

- Pipeline build id: xxxx
- This PR was tested with the ` "CGroup": "version_two"` injected in the image configs for mariner gen2 and full iso images, and was ran successfully in a pipeline build. The generated artifacts were tested and the images booted up with the correct cgroup version that was requested. Local builds were also completed initially to make sure the CGroup config flag is functional throughout the system workflow.


Camelron (approved):

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -2073,6 +2079,32 @@ func setGrubCfgSELinux(grubPath string, kernelCommandline configuration.KernelCo
...
+		cgroup = fmt.Sprintf("%s", cgroupv2FlagValue)
+	case configuration.CGroupV1:
+		cgroup = fmt.Sprintf("%s", cgroupv1FlagValue)
+	case configuration.CGroupDefault:
+		cgroup = ""
+	}
+
+	logger.Log.Debugf("Adding CGroupConfiguration('%s') to '%s'", cgroup, grubPath)
+	err = sed(cgroupPattern, cgroup, kernelCommandline.GetSedDelimeter(), grubPath)
+	if err != nil {
```
SeanDougherty, line 2101: Great work! This is a good opportunity to include some test coverage for this new installutils.go function

Once we get some tests written, coverage can be tracked by running `sudo make go-test-coverage` from your toolkit and downloading the Coverage report html to get confirmation that the new tests cover this new function

Some test recommendations I have are: "TestShouldReturnErrorWhenCGroupPatternIsMissingFromGrubCfg" and "TestShouldSetCorrectCgroupFlagValueForConfiguration"

SeanDougherty (changes requested):

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -2073,6 +2079,32 @@ func setGrubCfgSELinux(grubPath string, kernelCommandline configuration.KernelCo
 	return
 }
 
+func setGrubCfgCGroup(grubPath string, kernelCommandline configuration.KernelCommandLine) (err error) {
+	const (
+		cgroupPattern     = "{{.CGroup}}"
+		cgroupv1FlagValue = "systemd.unified_cgroup_hierarchy=0"
+		cgroupv2FlagValue = "systemd.unified_cgroup_hierarchy=1"
+	)
+	var cgroup string
```
neha170, line 2088: Where do we check if kernelCommandline.CGroup is valid or not? What's the value of the string cgroup when there is no match in the switch cases? Do we want to add a 'default' to the switch, and call sed only for the correct cgroup values? Thanks

neha170 (commented):
Could we also test for cgroup="version_one" and cgroup="", thanks

toolkit/tools/imagegen/installutils/installutils.go
aditjha-msft: @neha170 the validity of CGroup entry is checked in the kernelcommandline.go file. The function is called isValid() and it gets utilized in the UnmarshalJSON function. If I'm understanding your second question correctly, I did add a default to the switch it is the last case, and it sets the value of string cgroup to "" as an empty string, so the sed call happening after the switch case is determined handles that as well. let me know if I'm misunderstanding something. Thanks!

toolkit/tools/imagegen/installutils/installutils.go
SeanDougherty: Based on our discussion, marking this resolved until we get feedback from team members

SeanDougherty (approved):

dmcilvaney (approved):

PR #4353, opened 2022-12-02
Work around llvm bug causing periodic crashes during DWARF finalization

sjep:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This is a longstanding issue in the LLVM project: [issue #49955](https://github.com/llvm/llvm-project/issues/49955). It only occurs when llvm is compiled from gcc and is due to a particular optimization in gcc: strict-aliasing. This workaround simply disables strict-aliasing during llvm compilation.

###### Change Log  

- Change
Added a compile flag to llvm (`-fno-strict-aliasing`)

**NO** (I think mariner toolchain is gcc-based, if clang is required in the toolchain let me know the steps to verify.

###### Test Methodology

- I built the llvm package, installed it on an existing mariner docker instance and ran clang on the bad file 10 times alongside another docker file with the original clang.
  Old: 10 runs, 3 fails
  New: 10 runs, 0 fails


henryli001 (commented):
Please change the targeted check-in branch to main since that's the dev branch Mariner team is using. 2.0 is for official releases. Also I'm wondering what does the ptest result look like after this change? Do all the tests pass?

sjep:
Sure - give me a second to rebase off main

sjep:
I rebased onto main. I'm not sure how to run ptest or any test for that matter other than the go-test-coverage target. How do I kick off the test suite?

henryli001:
> I rebased onto main. I'm not sure how to run ptest or any test for that matter other than the go-test-coverage target. How do I kick off the test suite?

Thanks for doing that!

go-test-coverage is executing the unit tests of our go tooling whereas the ptest here refers to the test suite of the llvm package.  I do see that there's a %check section in the llvm spec file, so we need to run them to make sure no tests are regressed. You can run the ptest suite by adding **RUN_CHECK=y** in your package build command and after the build finishes, you'll be able to see the test result within the build log of llvm.

SPECS/llvm/llvm.spec
```diff
@@ -37,6 +37,7 @@ export CXXFLAGS="`echo " %{build_cxxflags} " | sed 's/ -g//'`"
 mkdir -p build
 cd build
 cmake -G Ninja                              \
+      -DCMAKE_CXX_FLAGS="-fno-strict-aliasing" \
```
PawelWMS, line 40: Instead of disabling strict aliasing for the whole build, could we instead use [the patch](https://github.com/NixOS/nixpkgs/pull/152321) mentioned in the [49955 issue](https://github.com/llvm/llvm-project/issues/49955) you referenced?

PawelWMS (changes requested):
Ask to use a fix instead of disabling strict aliasing globally.

SPECS/llvm/llvm.spec
sjep: If you're comfortable with the minimalist approach I can include the same patch in this commit and apply it instead of what I'm currently doing, you're right that it appears to be sufficient though I have no idea why.

sjep:
Running build/checks now for patched llvm

SPECS/llvm/llvm.spec
```diff
@@ -8,6 +8,7 @@ Distribution:   Mariner
 Group:          Development/Tools
 URL:            https://llvm.org/
 Source0:        https://github.com/llvm/llvm-project/releases/download/llvmorg-%{version}/%{name}-%{version}.src.tar.xz
+Patch1:         llvm-12.0.1-issue-49955-workaround.patch
```
PawelWMS, line 11: Thanks for the update. Now we missed incrementing the `Release` tag and adding a changelog entry. Aside from that, I think we're good to go.

PawelWMS (changes requested):

SPECS/llvm/llvm.spec
sjep: Oops when I switched techniques I forgot to pull over that diff. Will fix this now.

PawelWMS (approved):

PawelWMS:
> > I rebased onto main. I'm not sure how to run ptest or any test for that matter other than the go-test-coverage target. How do I kick off the test suite?
> 
> Thanks for doing that!
> 
> go-test-coverage is executing the unit tests of our go tooling whereas the ptest here refers to the test suite of the llvm package. I do see that there's a %check section in the llvm spec file, so we need to run them to make sure no tests are regressed. You can run the ptest suite by adding **RUN_CHECK=y** in your package build command and after the build finishes, you'll be able to see the test result within the build log of llvm.

Very good catch, @henryli001! However, I have just checked and for this particular package the ptests are still regularly failing, so there's no regression to look for, yet. We'll be fixing that on our end but out of scope of this PR.

PawelWMS:
> > I rebased onto main. I'm not sure how to run ptest or any test for that matter other than the go-test-coverage target. How do I kick off the test suite?
> 
> Thanks for doing that!
> 
> go-test-coverage is executing the unit tests of our go tooling whereas the ptest here refers to the test suite of the llvm package. I do see that there's a %check section in the llvm spec file, so we need to run them to make sure no tests are regressed. You can run the ptest suite by adding **RUN_CHECK=y** in your package build command and after the build finishes, you'll be able to see the test result within the build log of llvm.

Very good observationm Henry!

Issue #3956, opened 2022-10-08
[2.0] Add support for LLVM's libc++


hcoona:
# Summary

I cannot build any c++ sources with clang++ in Mariner 2.0 Docker image.

I think it's because we upgrade clang to v12.0.1 but still leave libstdc++-devel v11.2.0.

# How to reproduce

```bash
# image id: 9981cdf1c8ec
docker run --net=host --rm -it mcr.microsoft.com/cbl-mariner/base/core:2.0 /bin/bash
tdnf install -y clang
cat > a.cc << EOF
#include <iostream>
int main() {
  std::cout << "Hello!" << std::endl;
  return 0;
}
EOF
clang++ -std=c++17 -stdlib=libc++ -c a.cc
```

![image](https://user-images.githubusercontent.com/712433/194713381-3d88e961-120b-4db8-92af-9797823cfe4d.png)

![image](https://user-images.githubusercontent.com/712433/194713454-24e65d4c-56aa-446c-b537-5b17e9141bc8.png)

![image](https://user-images.githubusercontent.com/712433/194713486-6f434c76-41bf-41db-bea1-42b3bda48a0b.png)

hcoona:
I think it's introduced by https://github.com/microsoft/CBL-Mariner/pull/1428, @christopherco could you help to take a look?

mfrw:
@hcoona 
I managed to get it working by installing `build-essential`

`tdnf install build-essential`
![image](https://user-images.githubusercontent.com/4929861/194804239-9f3ff09e-bddc-4b68-bdd3-d3083f7a5068.png)

Also, IMHO, `libstdc++ 11.2.0-2` is shipped by `gcc`.
Wondering if the issue is related.

hcoona:
@mfrw you are right, `libstdc++` is not `libc++`, my fault.

But it still not works for me.

```bash
# image id: 9981cdf1c8ec
docker run --net=host --rm -it mcr.microsoft.com/cbl-mariner/base/core:2.0 /bin/bash
tdnf install -y build-essential clang
cat > a.cc << EOF
#include <iostream>
int main() {
  std::cout << "Hello!" << std::endl;
  return 0;
}
EOF
clang++ -std=c++17 -stdlib=libc++ -c a.cc
```

![image](https://user-images.githubusercontent.com/712433/194831152-df202452-abe1-42ed-b061-a638ac3be935.png)

mfrw:
[...]

PR #4388, opened 2022-12-07
Add compiler-rt package for clang options like code coverage to work

sjep:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

The clang compiler has options for code coverage: `clang ++ -fprofile-instr-generate -fcoverage-mapping test.cc`.

Without certain static libraries this option fails with a fairly cryptic message:
```
/usr/bin/ld: cannot find /usr/lib/clang/12.0.1/lib/linux/libclang_rt.profile-x86_64.a: No such file or directory
```

This package provides the above missing archive, among other tools in the [compiler-rt project](https://compiler-rt.llvm.org/) within llvm.

###### Change Log  

- New package with spec file: compiler-rt which depends on llvm and provides libraries associated with llvm.

**NO**

###### Test Methodology

It's a bit difficult to run clang by itself right now due to issue #3956. However, after installing g++ alongside clang, binutils, kernel-headers and glibc you can run this test program:
```
cat > test.cc << EOF
#include <iostream>
int main() {
  std::cout << "Hello!" << std::endl;
  return 0;
}
EOF
clang++ -fprofile-instr-generate -fcoverage-mapping test.cc
```


SPECS/compiler-rt/compiler-rt.spec
```diff
@@ -0,0 +1,72 @@
+%global maj_ver 12
```
jslobodzian, line 1: Do we need all the Macros?

SPECS/compiler-rt/compiler-rt.spec
```diff
@@ -0,0 +1,72 @@
...
+%global min_ver 0
+%global patch_ver 1
+
+%global compiler_rt_srcdir %{name}-%{version}.src
```
jslobodzian, line 5: Is "_srcdir" the right suffix for this macro?  Seems like this is actually the "source" name since it's used in "Source0"

SPECS/compiler-rt/compiler-rt.spec
```diff
@@ -0,0 +1,72 @@
...
+%{_includedir}/profile
+%{_includedir}/sanitizer
+%{_includedir}/xray
+%{_libdir}/clang/%{version}/lib/linux/*
+%{_libdir}/clang/%{version}/share/*
+%{_bindir}/hwasan_symbolize
+
+%changelog
+* Wed Dec 07 2022 Adam Schwab <adschwab@microsoft.com>
+- Influenced by Fedora's spec from version fc35 (license: MIT)
```
sameluch, line 69: Since these two change log entries are for the same version, they should be collapsed into a single entry.

SPECS/compiler-rt/compiler-rt.spec
```diff
@@ -0,0 +1,72 @@
...
+into compiled binaries. Profile is a library used to collect coverage information and
+BlocksRuntime is an implementation of Apple "blocks" interface.
+
+%prep
+%setup -q -n %{compiler_rt_srcdir}
+
+%build
+# Disable symbol generation
+export CFLAGS="`echo " %{build_cflags} " | sed 's/ -g//'`"
+export CXXFLAGS="`echo " %{build_cxxflags} " | sed 's/ -g//'`"
```
PawelWMS, line 35: Just curious: why do we want to disable symbol generation? Is this causing build issues?

SPECS/compiler-rt/compiler-rt.spec
```diff
@@ -0,0 +1,72 @@
...
+
+mkdir -p build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=%{_prefix}   \
+      -DCMAKE_BUILD_TYPE=Release          \
+      -Wno-dev ..
```
PawelWMS, line 41: I'd give `%cmake` a shot. This way we have better control over build arguments of our packages and can update them in one place (and it already has `DCMAKE_INSTALL_PREFIX` set the way we want it):
```suggestion
%cmake -DCMAKE_BUILD_TYPE=Release  \
       -Wno-dev ..
```

SPECS/compiler-rt/compiler-rt.spec
sjep: This spec is designed to mimic the clang spec (which is its cousin, given it comes from the same llvm repo) - this is how the clang spec specified its version. If you want me to simplify it that makes sense but shouldn't these be similar to each other?

SPECS/compiler-rt/compiler-rt.spec
sjep: I wanted to do this but wasn't sure it would pass the checker, I guess it has to be exactly:
```
- Original version for CBL-Mariner. Influenced by Fedora's spec from version fc35 (license: MIT)
```
This should pass the test, I can change it to this.

SPECS/compiler-rt/compiler-rt.spec
sjep: This has been in all the specs I've looked at so far. I copied it here because I assumed it was wanted for any c/c++ compilation. Please advise.

SPECS/LICENSES-AND-NOTICES/data/licenses.json
```diff
@@ -1990,6 +1990,7 @@
                 "clamav",
                 "cloud-hypervisor",
                 "cmake-fedora",
+                "compiler-rt",
```
PawelWMS, line 1993: Kind general request for future reviews: please minimize overwriting commit history (through rebasing or `git commit --amend`) and instead just append new commits. This way it's a lot easier to track what changed since the last time you looked at the PR.

SPECS/compiler-rt/compiler-rt.spec
```diff
@@ -0,0 +1,64 @@
...
+%{_includedir}/profile
+%{_includedir}/sanitizer
+%{_includedir}/xray
+%{_libdir}/clang/%{version}/lib/linux/*
+%{_libdir}/clang/%{version}/share/*
+%{_bindir}/hwasan_symbolize
+
+%changelog
+* Tue Dec 06 2022 Adam Schwab <adschwab@microsoft.com>
+- Original version for CBL-Mariner. Influenced by Fedora's spec from version fc35 (license: MIT) - 12.0.1-1
```
PawelWMS, line 64:
```suggestion
* Tue Dec 06 2022 Adam Schwab <adschwab@microsoft.com> - 12.0.1-1
- Original version for CBL-Mariner. Influenced by Fedora's spec from version fc35 (license: MIT)
```

SPECS/compiler-rt/compiler-rt.spec
PawelWMS: Since these 3 macros are used only once, there's really no gain here, so removing them will reduce clutter.

When it comes to similarity: keeping specs similar to the original we used from a different distro may make future updates easier. That mostly applies to the build, installation, test, and file sections, though. Here we can just drop these macros.

SPECS/compiler-rt/compiler-rt.spec
```diff
@@ -0,0 +1,61 @@
...
+%{_includedir}/sanitizer
+%{_includedir}/xray
+%{_libdir}/clang/%{version}/lib/linux/*
+%{_libdir}/clang/%{version}/share/*
+%{_bindir}/hwasan_symbolize
+
+%changelog
+* Tue Dec 06 2022 Adam Schwab <adschwab@microsoft.com> - 12.0.1-1
+- Original version for CBL-Mariner. License verified.
+- Influenced by Fedora's spec from version fc35 (license: MIT)
```
PawelWMS, line 61: After an offline discussion, this will probably better reflect the genesis of the spec:
```suggestion
- Initial CBL-Mariner import from Fedora 35 (license: MIT). License verified.
```

PR #4489, opened 2022-12-22
TensorFlow: add missing deps

rikenm1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

- Adding missing TensorFlow dependencies. 

###### Change Log  

- Change : Add Tensorboard, Tensorboard plugins, and add missing runtime dep for python3-virtualvenv spec

**YES/NO**
NO

###### Test Methodology

- Pipeline build id: [AMD] [286180](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=286180&view=results)
[ARM]: [286058](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=286058&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=d723e8ff-9fe0-5a1d-bed2-08b8dc44a520)


SPECS/python-virtualenv/python-virtualenv.spec
```diff
@@ -25,6 +25,9 @@ BuildRequires:  python3-pip
 %endif
 
 Requires:       python3
+Requires:       python3-filelock
+Requires:       python3-platformdirs = 2.0.0
```
jslobodzian, line 29: Are you sure you want to set these as =
We should only have the one version in mariner.  And if we upgrade the version in mariner, it's alsmost always for a CVE reason.  So this explicit assignment drags in older CVE riddled packages.

SPECS/tensorflow/tensorflow.spec
```diff
@@ -64,6 +65,43 @@ Requires:       python3-zipp
 %description -n python3-tensorflow
 Python 3 version.
 
+%package -n python3-tf-nightly
```
jslobodzian, line 68: What does this package do?

SPECS/python-tensorflow-io-gcs/python-tensorflow-io-gcs.spec
```diff
@@ -0,0 +1,50 @@
+%global pypi_name tensorflow-io-gcs
+%global _description %{expand:
```
jslobodzian, line 2: I think you should dump this Macro.. Copy this content to the  %description section.

Also...I would change the Summary: below to say simply "Extended TensorFlow File Formats"

SPECS/python-tensorflow-io-gcs/python-tensorflow-io-gcs.spec
```diff
@@ -0,0 +1,50 @@
+%global pypi_name tensorflow-io-gcs
```
jslobodzian, line 1: What does gcs stand for?

SPECS/python-virtualenv/python-virtualenv.spec
rikenm1: Jon, these had upper bound on the version so if we update these runtime packages outside of that upper bound, it might not work so I had this explicit version with the dep. 
image

SPECS/python-tensorflow-io-gcs/python-tensorflow-io-gcs.spec
jslobodzian: Maybe the summary is just whatever the definition of gcs is,

SPECS/tensorflow/tensorflow.spec
rikenm1: to remove the circular dep among the packages had to include this.

SPECS/python-tensorflow-estimator/python-tensorflow-estimator.spec
```diff
@@ -0,0 +1,65 @@
+%global pypi_name tensorflow-estimator
+%global _description %{expand:
```
jslobodzian, line 2: I would dump this macro.  It's only used in one place.

SPECS/python-tensorflow-estimator/python-tensorflow-estimator.spec
```diff
@@ -0,0 +1,65 @@
...
+A high-level TensorFlow API that greatly simplifies machine learning programming}
+%define _enable_debug_package 0
+%global debug_package %{nil}
+
+Summary:        A high-level TensorFlow API that greatly simplifies machine learning programming
+Name:           python-%{pypi_name}
+Version:        2.11.0
```
jslobodzian, line 9: Are you sure this version is correct. The previous spec file has the exact same version number.

SPECS/python-tensorboard/python-tensorboard.spec
```diff
@@ -0,0 +1,110 @@
+%global pypi_name tensorboard
+%global _description %{expand:
+TensorBoard is a suite of web applications for inspecting and understanding your TensorFlow runs and graphs}
+%define _enable_debug_package 0
+%global debug_package %{nil}
+
+Summary:        TensorBoard is a suite of web applications for inspecting and understanding your TensorFlow runs and graphs
+Name:           python-%{pypi_name}
+Version:        2.11.0
```
jslobodzian, line 9: Just double checking that this version number is correct.  It's the same version as the next spec file.

SPECS/python-tensorboard-plugin-wit/python-tensorboard-plugin-wit.spec
```diff
@@ -0,0 +1,47 @@
+%global pypi_name tensorboard-plugin-wit
+%global _description %{expand:
+The What-If Tool (WIT) provides an easy-to-use interface for expanding understanding of a black-box classification or regression ML model}
+%define _enable_debug_package 0
+%global debug_package %{nil}
+Summary:        The What-If Tool (WIT) provides an easy-to-use interface for expanding understanding of a black-box classification or regression ML model
```
jslobodzian, line 6: Consider shortening this to the Machine Learning What If Tool?

SPECS/python-tensorflow-estimator/python-tensorflow-estimator.spec
rikenm1: TensorFlow, Tensorboard, Keras, and most of the dep of Tensorflow made by the TensorFlow team have the same version. For example in this case 2.11.0

SPECS/python-tensorboard/python-tensorboard.spec
rikenm1: TensorFlow, Tensorboard, Keras, and most of the dep of Tensorflow made by the TensorFlow team have the same version. For example in this case 2.11.0

SPECS/keras/keras.spec
```diff
@@ -0,0 +1,74 @@
+%define _enable_debug_package 0
+%global debug_package %{nil}
+Summary:        Keras is a high-level neural networks API, written in Python and capable of running on top of TensorFlow, CNTK, or Theano.
```
jslobodzian, line 3: Consider shortening to "A High Level Neural Networks API"

SPECS/python-distlib/python-distlib.signatures.json
```diff
@@ -0,0 +1,5 @@
+{
+    "Signatures": {
+     "distlib-0.3.6.tar.gz": "14bad2d9b04d3a36127ac97f30b12a19268f211063d8f8ee4f47108896e11b46"
```
osamaesmailmsft, line 3: Not 100% sure, but I thought our blob tarballs were supposeds to use the package name (in this case, python-distlib)

SPECS/keras/keras.spec
```diff
@@ -0,0 +1,74 @@
...
+
+%description -n python3-keras
+Python 3 version.
+
+%prep
+%autosetup -p1
+
+
+%build
+# tar -xf %{SOURCE1} -C /root/
```
jslobodzian, line 45: Dead code?  Probably should remove

SPECS/keras/keras.spec
```diff
@@ -0,0 +1,74 @@
...
+
+ln -s %{_bindir}/python3 %{_bindir}/python
+
+bazel --batch build  --verbose_explanations //keras/tools/pip_package:build_pip_package
+# ---------
+# steps to create the cache tar. network connection is required to create the cache.
+#----------------------------------
+# pushd /root
```
jslobodzian, line 53: Should add the script for this

SPECS/keras/keras.spec
```diff
@@ -0,0 +1,74 @@
...
+BuildRequires:  pyproject-rpm-macros
+BuildRequires:  python3-devel
+BuildRequires:  python3-numpy
+BuildRequires:  python3-packaging
+BuildRequires:  python3-pip
+BuildRequires:  python3-requests
+BuildRequires:  python3-wheel
+BuildRequires:  tar
+BuildRequires:  which
+BuildRequires:  python3-tf-nightly = 2.11.0
```
jslobodzian, line 27: I guess this is version locked?

SPECS/keras/keras.spec
PawelWMS: Alternatively I'd double-check we're using the cache correctly, because at the moment `Source1` is not used anywhere in this spec. If we don't need the cache to build the package (keep in mind that test runs enable network access, so they might pass, because Bazel is able to download other dependencies), let's remove the cache. If we do need it, I'm guessing we need this line uncommented.

SPECS/python-distlib/python-distlib.signatures.json
PawelWMS: The most general guideline is "make sure the name is clear enough to understand, which package it goes to, and that there's little to no risk of collisions". In this case, "distlib" may be somewhat risky, if there's a non-Python version of this package, so I'd also suggest a re-name.

SPECS/python-distlib/python-distlib.spec
```diff
@@ -0,0 +1,192 @@
...
+
+%install
+%pyproject_install
+%pyproject_save_files %{srcname}
+
+%if %{with check}
+%check
+export PYTHONHASHSEED=0
+# Some tests require network access
+export SKIP_ONLINE=1
```
PawelWMS, line 61: In our case we can run tests requiring network access.
```suggestion
```

SPECS/python-distlib/python-distlib.spec
```diff
@@ -0,0 +1,192 @@
...
+# test_sequencer_basic test fails due to relying
+# on the ordering of the input, hence disabling it.
+# https://github.com/pypa/distlib/issues/161
+%pytest -k "not test_sequencer_basic"
+%endif # with_tests
+
+%files -n python%{python3_pkgversion}-%{srcname} -f %pyproject_files
+%doc README.rst
```
PawelWMS, line 69: Please find a license file and add it with the `%license` macro.

SPECS/python-distlib/python-distlib.spec
```diff
@@ -0,0 +1,192 @@
+%global srcname distlib
+%bcond_without check
+Summary:        Low-level components of distutils2/packaging, augmented with higher-level APIs
+Name:           python-distlib
+Version:        0.3.6
+Release:        1%{?dist}
```
PawelWMS, line 6: Forgot to increment the `Release` tag:
```suggestion
Release:        2%{?dist}
```

SPECS/python-platformdirs/python-platformdirs.spec
```diff
@@ -0,0 +1,61 @@
...
+a "user data dir".}
+Summary:        Python module for determining appropriate platform-specific dirs
+Name:           python-%{srcname}
+Version:        2.0.0
+Release:        2%{?dist}
+License:        MIT
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+URL:            https://github.com/platformdirs/platformdirs
+Source0:        %{url}/archive/%{version}/%{srcname}-%{version}.tar.gz
```
PawelWMS, line 13: `Source0` written this way will make `rpmbuild` look for `%{srcname}-%{version}.tar.gz`, which doesn't match the file mentioned in the signatures JSON. Please make sure all of these file names align.

SPECS/python-tensorboard-plugin-wit/python-tensorboard-plugin-wit.spec
```diff
@@ -0,0 +1,47 @@
...
+%package -n python3-%{pypi_name}
+Summary:        %{summary}
+
+%description -n python3-%{pypi_name} %{_description}
+
+%prep
+%autosetup -p1 -n what-if-tool-%{version}
+
+%build
+ln -s %{_bindir}/python3 %{_bindir}/python
```
PawelWMS, line 32: Just curious: why was this needed, if we're running `python3` in the next line anyway?

SPECS/python-tensorboard/python-tensorboard.signatures.json
```diff
@@ -0,0 +1,6 @@
+{
+    "Signatures": {
+     "tensorboard-2.11.0.tar.gz": "4045f3eca77ef335dcd70879ff9aecc2d2ba09e3fb5f39e501f868b4ab23c2b2",
+     "python-tensorboard-2.11.0-cache.tar.gz": "77b35869c63fc15b5cdeeff3dc6ac472e79a62739a69537bf60a38fa902b66a4"
```
PawelWMS, line 4: I'd suggest keeping the file name format the same, so either:
```suggestion
     "python-tensorboard-2.11.0.tar.gz": "4045f3eca77ef335dcd70879ff9aecc2d2ba09e3fb5f39e501f868b4ab23c2b2",
     "python-tensorboard-2.11.0-cache.tar.gz": "77b35869c63fc15b5cdeeff3dc6ac472e79a62739a69537bf60a38fa902b66a4"
```

or:
```suggestion
     "tensorboard-2.11.0.tar.gz": "4045f3eca77ef335dcd70879ff9aecc2d2ba09e3fb5f39e501f868b4ab23c2b2",
     "tensorboard-2.11.0-cache.tar.gz": "77b35869c63fc15b5cdeeff3dc6ac472e79a62739a69537bf60a38fa902b66a4"
```

Personally, I'm a bigger fan of the former, because it has a lower chance of collisions with other package source.

SPECS/python-tensorboard/python-tensorboard.spec
```diff
@@ -0,0 +1,110 @@
...
+Requires:   python3-absl-py
+Requires:   python3-google-auth-oauthlib = 0.5.3
+Requires:   python3-google-auth = 2.6.6
+Requires:   python3-grpcio 
+Requires:   python3-markdown
+Requires:   python3-numpy
+Requires:   python3-protobuf = 3.17.3
+Requires:   python3-requests = 2.27.1
+Requires:   python3-setuptools
+Requires:   python3-tensorflow-estimator = 2.11.0
```
PawelWMS, line 44: Why do we need to be so strict on these package versions? This may cause upgrade issues (a different package may require a newer versions of one of these for instance) and leave us vulnerable to security bugs fixed in newer versions of these packages.

SPECS/python-tensorboard/python-tensorboard.spec
```diff
@@ -0,0 +1,110 @@
...
+
+ln -s /usr/bin/python3 /usr/bin/python
+
+#tensorboard-data-server
+pushd tensorboard/data/server/pip_package
+python3 setup.py -q bdist_wheel 
+popd
+mkdir -p pyproject-wheeldir/ && cp tensorboard/data/server/pip_package/dist/*.whl pyproject-wheeldir/
+
+#tensorboard buildt usin bazel
```
PawelWMS, line 72: Nit-pick:
```suggestion
#tensorboard built using bazel
```

SPECS/python-tensorboard/python-tensorboard.spec
```diff
@@ -0,0 +1,110 @@
...
+bazel --batch build //tensorboard/pip_package:build_pip_package
+#cache 
+# ---------
+# steps to create the cache tar. network connection is required to create the cache.
+#----------------------------------
+# bazel clean
+# pushd /root
+# tar -czvf %{name}-%{version}-cache.tar.gz .cache  #creating the cache using the /root/.cache directory
+# popd
+# mv /root/%{name}-%{version}-cache.tar.gz /usr/
```
PawelWMS, line 82: Please move to a script - same ask as what Jon mentioned in another spec.

SPECS/python-tensorboard/python-tensorboard.spec
```diff
@@ -0,0 +1,110 @@
...
+
+# sleep 200
+#tensorboard build script has an error
```
PawelWMS, line 85: Please remove these lines or provide more explanation. It's unclear to me, what are they for right now.

SPECS/python-tensorflow-estimator/python-tensorflow-estimator.signatures.json
```diff
@@ -0,0 +1,6 @@
+{
+    "Signatures": {
+     "tensorflow-estimator-2.11.0.tar.gz": "922f9187de79e8e7f7d7a5b2d6d3aabc81bbbd6ba5f12a4f52967dd302214a43",
+     "python-tensorflow-estimator-2.11.0-cache.tar.gz": "d10ac3f8470dd3e26ecca88b399d8cd927ab69b70ac876ce702063b648047cee"
```
PawelWMS, line 4: Similar comment here: let's stick to one file naming format.

SPECS/python-tensorflow-estimator/python-tensorflow-estimator.spec
```diff
@@ -0,0 +1,65 @@
...
+
+# ---------
+# steps to create the cache tar. network connection is required to create the cache.
+#----------------------------------
+# bazel clean
+# pushd /root
+# tar -czvf %{name}-%{version}-cache.tar.gz .cache  #creating the cache using the /root/.cache directory
+# popd
+# mv /root/%{name}-%{version}-cache.tar.gz /usr/
+# sleep 240
```
PawelWMS, line 49: Please move to a script.

SPECS/python-tensorflow-estimator/python-tensorflow-estimator.spec
```diff
@@ -0,0 +1,65 @@
...
+Summary:        A high-level TensorFlow API that greatly simplifies machine learning programming
+Name:           python-%{pypi_name}
+Version:        2.11.0
+Release:        1%{?dist}
+License:        ASL 2.0
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+URL:            https://github.com/tensorflow/estimator
+Source0:        https://github.com/tensorflow/estimator/archive/refs/tags/v%{version}.tar.gz#/%{pypi_name}-%{version}.tar.gz
+Source1:        %{name}-%{version}-cache.tar.gz
```
PawelWMS, line 16: We have the cache source, but I don't see it being unpacked anywhere in the spec.

SPECS/python-tensorflow-io-gcs/python-tensorflow-io-gcs.spec
```diff
@@ -0,0 +1,50 @@
...
+License:        ASL 2.0
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+URL:            https://github.com/tensorflow/io
+Source0:        https://github.com/tensorflow/io/archive/refs/tags/v%{version}.tar.gz#/%{pypi_name}-%{version}.tar.gz
+BuildRequires:  python3-setuptools
+BuildRequires:  python3-pip
+BuildRequires:  python3-wheel
+BuildRequires:  python3-six
+ExclusiveArch:  x86_64
```
PawelWMS, line 20: The package seems to contain only Python scripts. Why is it exclusively aimed at AMD64?

SPECS/python-virtualenv/python-virtualenv.spec
PawelWMS: It still looks incorrect, though, according to your screenshot.
```suggestion
Requires:       python3-platformdirs < 3
```

SPECS/python-virtualenv/python-virtualenv.spec
```diff
@@ -25,6 +25,9 @@ BuildRequires:  python3-pip
...
+Requires:       python3-distlib = 0.3.6
```
PawelWMS, line 30:
```suggestion
Requires:       python3-distlib < 1
```

SPECS/tensorflow/tensorflow.spec
```diff
@@ -102,6 +140,11 @@ bazel --batch build  --verbose_explanations //tensorflow/tools/pip_package:build
 %{_bindir}/toco
 %{_bindir}/toco_from_protos
 
+%files -n python3-tf-nightly
+%license LICENSE
+%{python3_sitelib}/*
```
PawelWMS, line 145: This seems wrong, since `python3-tensorflow` also provides `%{python3_sitelib}/*`. Please double-check, which Python artifacts should belong to `python3-tf-nightly` and split these artifacts between the packages accordingly. You can then add `Requires: python3-tf-nightly = %{version}-%{release}` to `pythno3-tensorflow`, so that `python3-tensorflow` still provides the same set of features.

SPECS/tensorflow/tensorflow.spec
```diff
@@ -102,6 +140,11 @@ bazel --batch build  --verbose_explanations //tensorflow/tools/pip_package:build
...
+
+
 %changelog
 * Thu Dec 08 2022 Riken Maharjan <rmaharjan@microsoft> - 2.11.0-2
```
PawelWMS, line 149: Needs a changelog entry.

SPECS/python-virtualenv/python-virtualenv.spec
rikenm1: @PawelWMS My bad I screenshotted the latest virtualenv instead of the version we had. 
image

SPECS/tensorflow/tensorflow.spec
rikenm1: this is the intended design. 
The reason behind this is as follows:
Some of the run time dependencies of Tensorflow like Keras, Tensboard, and  Tensorflow-estimator need TensorFlow as a build time requirement. The build would not even succeed without TensorFlow itself being installed. The tensorFlow team usages nightly package like I did to satisfy this requirement.

SPECS/python-tensorboard/python-tensorboard.spec
rikenm1: image
 yes this is locked.

SPECS/python-tensorflow-estimator/python-tensorflow-estimator.spec
rikenm1: I talked with Nicholas last time about this. Adding these commands to the script file will confuse others as it won't run as intended because it needs Bazel runtime with build requirements from this spec file.  
Also, these commands need to run precisely after the Bazel build is successful which is happening in this case.

SPECS/python-tensorboard/python-tensorboard.spec
rikenm1: I talked with Nicholas last time about this. Adding these commands to the script file will confuse others as it won't run as intended because it needs Bazel runtime with build requirements from this spec file.
Also, these commands need to run precisely after the Bazel build is successful which is happening in this case.

SPECS/python-tensorflow-io-gcs/python-tensorflow-io-gcs.spec
oliviacrain: It stands for Google cloud storage, judging by the namespace defines in [their code](https://github.com/tensorflow/io/blob/master/tensorflow_io_gcs_filesystem/core/gcs_filesystem.cc#L36). @rikenm1 are we sure this package is necessary for our needs?

oliviacrain (approved):

PR #4578, opened 2023-01-10
Fix logwatch, rear and dmraid to use chronie pacakge, Remove crontabs package.

osamaesmailmsft:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

I removed `crontabs` from the packages since it was redundant (`cronie` already handles `crontab`). I also replaced the `crontabs` BuildRequires in the packages that have it.

###### Change Log  

- Removed `crontabs`
- Replaced `crontabs` with `cronie` in `logwatch`, `dmraid`, and `rear`

NO

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: 290213 (AMD64) 290214 (ARM64)
- ^ BOTH ARE IN PROGRESS RIGHT NOW


SPECS-EXTENDED/rear/rear.spec
```diff
@@ -91,7 +91,7 @@ Requires: mkisofs
 
 ### On RHEL/Fedora the genisoimage packages provides mkisofs
 %if %{?centos_version:1}%{?fedora:1}%{?rhel_version:1}0
```
jslobodzian, line 93: What is this macro doing on Mariner?  Are the Requires that follow actually being included in the Mariner Build?  Also the %if 0%{?rhel} a few lines down is suspicious.  At a glance I think we want to remove those for Mariner.

cgmanifest.json
```diff
@@ -2509,7 +2509,7 @@
         "other": {
           "name": "dmraid",
           "version": "1.0.0.rc16",
-          "downloadUrl": "https://people.redhat.com/heinzm/sw/dmraid/src/dmraid-1.0.0.rc16.tar.bz2"
+          "downloadUrl": "https://people.redhat.com/heinzm/sw/dmraid/src/dmraid-1.0.0.rc16-3.tar.bz2"
```
jslobodzian, line 2512: The -3 here doesn't seem to coincide with the version that the dmraid SPEC file has called out.  Are they really still on a Release Candidate?

SPECS-EXTENDED/dmraid/dmraid.spec
```diff
@@ -9,7 +9,7 @@ Distribution:   Mariner
 Summary: Device-mapper RAID tool and library
 Name: dmraid
 Version: 1.0.0.rc16
```
jslobodzian, line 11: Shouldn't there be a -3 here....instead of below where you added it on line 28.
Also, do we really still need a release candidate?

osamaesmailmsft: I wasn't sure how to do release candidate since it was conflicting with our release number, but I found dmraid-1.0.0.rc16 anyway so this should be fine.

SPECS-EXTENDED/rear/rear.spec
osamaesmailmsft: I think you're right that none of these are being included. However, it seems that the checks for each distro are because everyone adds `mkisofs` from different packages. We're using cdrkit, so should I add a BuildRequires for that when removing the flags?

osamaesmailmsft: I was able to remove the comments and add `cdrkit` to the Requires, but I saw `lsb` referenced by other distros. I found lsb_release, but I can't find a package that has LSB.

SPECS-EXTENDED/rear/rear.spec
```diff
@@ -65,43 +65,14 @@ Requires: bc
...
 %if %{?mandriva_version:1}0
-Requires: iproute2
-### Mandriva switched from 2008 away from mkisofs,
-### and as a specialty call the package cdrkit-genisoimage!
-%if 0%{?mandriva_version} >= 2008
-Requires: cdrkit-genisoimage
-%else
-Requires: mkisofs
-%endif
 #Requires: lsb
```
oliviacrain, line 69: For what it's worth, [newer versions of the spec from Fedora](https://src.fedoraproject.org/rpms/rear/blob/rawhide/f/rear.spec) don't require any sort of `lsb` package.

PR #4751, opened 2023-02-02
Parse %check section when RUN_CHECK=y to add %check passing as a requirement

osamaesmailmsft:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Since some tests are timing dependent and inconsistently pass/fail, we are requiring at least 3 build attempts when running ptests.

###### Change Log  

- buildworker sets the number of build attempts to at least 3 if RUN_CHECK=y.
- Also, if RUN_CHECK=y, we parse the rpm build log to verify the test results.

YES

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: 310085


toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -193,11 +197,44 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
 	)
 
 	logBaseName := filepath.Base(srpmFile) + ".log"
-	err = retry.Run(func() (buildErr error) {
-		builtFiles, logFile, buildErr = agent.BuildPackage(srpmFile, logBaseName, outArch, dependencies)
-		return
-	}, buildAttempts, retryDuration)
-
+	logger.Log.Debugf("osamatest: with check is (%t)", agent.Config().RunCheck)
+	maxAttempts := 3
```
osamaesmailmsft, line 201: Should I keep this here or make the change from time.sleep to waitgroups in retry.run so that it applies project-wide?

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -4,14 +4,18 @@
...
+	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/pkggraph"
-	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/retry"
+	// "github.com/microsoft/CBL-Mariner/toolkit/tools/internal/retry"
```
PawelWMS, line 18:
```suggestion
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
PawelWMS, line 201: For constant values, please use `const`.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -193,11 +197,44 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
 	)
 
 	logBaseName := filepath.Base(srpmFile) + ".log"
-	err = retry.Run(func() (buildErr error) {
-		builtFiles, logFile, buildErr = agent.BuildPackage(srpmFile, logBaseName, outArch, dependencies)
-		return
-	}, buildAttempts, retryDuration)
-
+	logger.Log.Debugf("osamatest: with check is (%t)", agent.Config().RunCheck)
```
PawelWMS, line 200: Please clean-up debug messages (in this case "osamatest" seems out of place).

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -193,11 +197,44 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
...
+	maxAttempts := 3
+	numAttempts := buildAttempts
+	if agent.Config().RunCheck && buildAttempts < maxAttempts {
+		numAttempts = maxAttempts
+	}
+	wg := new(sync.WaitGroup)
```
PawelWMS, line 206: What's the point of using WaitGroups here? It looks like we're not running any additional threads. I'm a bit confused.

osamaesmailmsft: When I ran the retry.Run part of the code with 3 attempts, it kicks off one build, then starts another one a second later, then a third after that. The same applied when I made this for loop without the waitgroups. With the waitgroups, the second build attempt doesn't start until after the first. I'm not sure if it's due to some parallelization or something of the like.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -193,11 +196,41 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
 	)
 
 	logBaseName := filepath.Base(srpmFile) + ".log"
-	err = retry.Run(func() (buildErr error) {
-		builtFiles, logFile, buildErr = agent.BuildPackage(srpmFile, logBaseName, outArch, dependencies)
-		return
-	}, buildAttempts, retryDuration)
-
+	const maxAttempts := 3
+	numAttempts := buildAttempts
```
PawelWMS, line 200: Instead of hard-coding values, I feel like we should differentiate between regular build failures, for which `buildAttempts` is used, and successful builds with failing tests. We shouldn't be silently overriding the rebuild count passed by the user.

My suggestions:
- Add a new parameter controlling the number of retries we want to perform in case of test failures. We could then first apply the build retries in case of a build failure and if the build succeeds, then we'd look into test results and their retry counts.
- [**My preference**] Remove retrying failed package builds and use only test retries. I think we never actually used anything than a single build re-try (the package build have been pretty stable) and in general, I believe we shouldn't be hiding flaky package builds and instead we should be fixing them.

toolkit/tools/scheduler/schedulerutils/buildworker.go
PawelWMS: I see. That was actually going to be my next question: how is the updated approach different from what `retry.Run` was doing? It seems like we're still running `agent.BuildPackage` and continue with the code once it exits. Of course, you've also added the log parsing code, but that should have no effect on the parallel vs sequential builds issue. I think more investigation is in order, to understand why the parallel builds were happening. Or why they have now disappeared.

I'd start by checking if `agent.BuildPackage` is exiting before the package actually finished building.

When it comes to WorkGroups, they make no sense if `Add` and `Done` are called form the same thread. In a simples case, the idea is to call `Add` before you start a thread, which will then call `Done` once it has finished. See: [example](https://gobyexample.com/waitgroups). And `Wait` is meant to halt the main thread (the one which started all the other ones) until all started threads call `Done`.

PawelWMS (changes requested):
We should understand the package build race better, before we're done with this PR.

toolkit/tools/scheduler/schedulerutils/buildworker.go
osamaesmailmsft: Essentially, the tests are failing due to timing out, but I haven't found any way to *only* run the %check section of a spec file, which has made it hard to only run tests again without building. I guess I could theoretically use the `rpmbuild` flags `-bi --short-circuit`, but that would still be re-installing before running the tests again.

To make sure I'm understanding what you're saying, are you saying that I should leave the `retry.Run` as is and let the package build run, and then parse the tests after, and rebuild if and only if the tests fail? Or do you mean that I should change some of my if-case logic to only check for test failures? Either way, I don't see a way to test without rebuilding/reinstalling the built RPM.

toolkit/tools/scheduler/schedulerutils/buildworker.go
osamaesmailmsft: To be completely honest, I tried the WaitGroups in the same section of the code because I wasn't sure if that would help with the package build retries being simultaneous, but I wasn't sure what else to try. I'm not sure why it seems to help where retry.Run didn't. I'll look into agent.BuildPackage and see how that function is returning, or if there is some parallelism somewhere that's causing this.

osamaesmailmsft: So, it appears that the `chrootAgent` that is used for `agent.BuildPackage` is initialized with multithreading, according to `toolkit/scripts/toolchain/container/toolchain_build_in_chroot.sh`. I'm investigating further to see if that's actually leading to `rpmbuild` running in parallel

toolkit/tools/scheduler/schedulerutils/buildworker.go
PawelWMS: Yes, I was thinking about the former, so re-running the full build in case of a test failure (and still failing immediately after first build failure). You're right that there's no easy way to retry only the `%check` section. Or at least no easy way I can think of right now. Anyway, since test builds are not time-critical, we might be fine with full rebuilds for now. If we end up timing out the whole build, then we'll need to improve.

osamaesmailmsft: Okay, that makes sense. I was able to get retry.Run working correctly now, which makes me think something might have been wrong with `redis` and not `retry.Run`, so that's being handled in its own PR.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -193,11 +197,37 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
 	)
 
 	logBaseName := filepath.Base(srpmFile) + ".log"
+	const checkAttempts = 3
```
PawelWMS, line 200: I still don't like the hard-coding of the attempts count. Here's my current suggestion:

Let's extend the meaning of `PACKAGE_BUILD_RETRIES` (found inside the `Makefile` file) to also force a rebuild in case of a test failure. Then let's change its default from 1 to 3. This should allow us to directly use `buildAttempts` in this function and completely drop `checkAttempts` along with some of the code. We'd also then have to slightly update `building.md` to mention, that the rebuilds will also happen for failed package tests.

I'm aware this is not the perfect solution, but the best, no overly complicated idea that came to my mind, considering we don't currently have a way to re-run only the `%check` section for otherwise successful builds. I'm open to other ideas, I don't want to hard-code these retries, though.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -193,11 +197,37 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
...
+						buildErr = errors.New(currLine)
+						if os.Rename(logFile, fmt.Sprintf("%s-fail-%d.log", filepath.Base(srpmFile), time.Now().UnixMilli())) != nil {
+							logger.Log.Debugf("Log file rename failed")
+						}
+						break
+					}
+				}
+			}
+			file.Close()
+		}
```
PawelWMS, line 228: A few clean-up steps:
- Let's always use `defer` to close files and set it up immediately after a successful `os.Open` call. Look for `defer file.Close()` in other places for examples.
- Minimizing the number of indents tends to lead to a code, that's easier to read. Also, we tend to immediately leave the function after encountering an error, thus my suggested change in the `if logErro != nil` logic.
- We can simplify using the scanner a bit.
- Let's add some more debug logging - this is of immense help, when something goes wrong.
- Let's keep and print the re-name error.
- Please extract the log file parsing logic to a separate function. This function is getting a bit bulky.
- Extra credit: add tests for the log parsing function.

**NOTE**: the code change below might not compile, although I did my best. It's meant to visualize my points above.;)
```suggestion
		// If the package builds with no errors and RUN_CHECK=y, check logs to see if the %check section passed, and if not, return as the build error.
		if buildErr != nil {
			return
		}

		if agent.Config().RunCheck {
			file, logErr := os.Open(logFile)
			// If we can't open the log file, that's a build error.
			if logErr != nil {
				logger.Log.Errorf("Failed to open log file '%s' while checking package test results. Error: %v", logfile, logErr)
				buildErr = logErr				
				return
			}
			defer file.Close()

			for scanner := bufio.NewScanner(file); scanner.Scan() {
				currLine := scanner.Text()
				// Anything besides 0 is a failed test
				if strings.Contains(currLine, "CHECK DONE") && !strings.Contains(currLine, "EXIT STATUS 0") {
					failedLogFile := fmt.Sprintf("%s-FAILED_TEST-%d.log", filepath.Base(srpmFile), time. Now().UnixMilli())
					buildErr = os.Rename(logfile, failedLogFile)
					if buildErr != nil {
						logger.Log.Errorf("Log file rename failed. Error: %v", buildErr)
						return
					}

					buildErr = fmt.Errorf("package test failed. Test status line: %s", currLine)
					break
				}
			}
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
osamaesmailmsft: Would I change the default from 1 to 3 for all scenarios, or would it be something like default is 3 if RUN_CHECK=y, else 1.

I think that there's a way to run %check without running %build, and that would be using `rpmbuild -bi --short-circuit`, which skips all steps before %install. If we do that, though, that would likely involve restructuring the way we run `rpmbuild`.

PawelWMS: Yes, ideally I'd only change the default for `RUN_CHECK=y`. Good idea!

And my gut is telling me, that trying to modify how we run `rpmbuild` might be a rabbit hole, so I didn't want to suggest it for this change. Having said that, if you can give it a try and see if it can be done within a few short hours, then go for it! If that won't work, or there's no time for this now, please create a bug summarizing what we did here and what might be a potential improvement, so at least we have a good starting point for future work.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -193,11 +197,46 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
...
+				buildErr = logErr				
+				return
+			}
+			defer file.Close()
+
+			for scanner := bufio.NewScanner(file); scanner.Scan() {
+				currLine := scanner.Text()
+				// Anything besides 0 is a failed test
+				if strings.Contains(currLine, "CHECK DONE") && !strings.Contains(currLine, "EXIT STATUS 0") {
+					failedLogFile := fmt.Sprintf("%s-FAILED_TEST-%d.log", filepath.Base(srpmFile), time. Now().UnixMilli())
```
PawelWMS, line 226: Looks like I made an autocorrect error. Sorry about that.
```suggestion
					failedLogFile := fmt.Sprintf("%s-FAILED_TEST-%d.log", filepath.Base(srpmFile), time.Now().UnixMilli())
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -193,11 +197,46 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
 	)
 
 	logBaseName := filepath.Base(srpmFile) + ".log"
```
PawelWMS, line 199: Another thing I've noticed just now: we're calling `filepath.Base(srpmFile)` twice now. Let's instead call it once and save the result. This way, if the log file's base name ever changes, it's going to be easier to make the update.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -192,12 +196,48 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
...
+			file, logErr := os.Open(logFile)
+			// If we can't open the log file, that's a build error.
+			if logErr != nil {
+				logger.Log.Errorf("Failed to open log file '%s' while checking package test results. Error: %v", logfile, logErr)
+				buildErr = logErr				
+				return
+			}
+			defer file.Close()
+			scanner := bufio.NewScanner(file)
+			for scanner.Scan() {
```
PawelWMS, line 223: Ah, sorry about that. Looks like I missed a semicolon when I was providing the example. If you're still interested [**NOT** important, just an FYI]:
```suggestion
			for scanner := bufio.NewScanner(file); scanner.Scan(); {
```

And btw. I got the idea from [here](https://github.com/microsoft/CBL-Mariner/blob/225fbf0b815e0ddc19cc39c43473f5fe44b71303/toolkit/tools/internal/logger/log.go#L139).

osamaesmailmsft: Ahh, I was curious what was causing the issue with that because it looked fine. Didn't realize the semicolon would fix that :)

toolkit/tools/scheduler/schedulerutils/buildworker.go
PawelWMS: Update: discussed offline and my idea wouldn't work well, since it would make the build process somewhat surprising for test runs due to additional rebuilds. Current plan is to modify the rebuilds through setting the `PACKAGE_BUILD_RETRIES` argument during test runs.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,6 +189,35 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
+			}
+			checkErr = fmt.Errorf("Package test failed. Test status line: %s", currLine)
+			return
+		}
+		if strings.Contains(currLine, "CHECK DONE") && strings.Contains(currLine, "EXIT STATUS 0") {
+			return
+		}
+	}
+	return
+}
```
PawelWMS, line 219: Ah, with the extracted function we get the additional benefit of being able to (re)use only one error variable:
```suggestion
func parseCheckSection(logFile string) (err error) {
	file, err := os.Open(logFile)
	// If we can't open the log file, that's a build error.
	if err != nil {
		logger.Log.Errorf("Failed to open log file '%s' while checking package test results. Error: %v", logFile, err)
		return
	}
	defer file.Close()
	for scanner := bufio.NewScanner(file); scanner.Scan(); {
		currLine := scanner.Text()
		// Anything besides 0 is a failed test
		if strings.Contains(currLine, "CHECK DONE") && !strings.Contains(currLine, "EXIT STATUS 0") {
			failedLogFile := fmt.Sprintf("%s-FAILED_TEST-%d.log", logFile[:strings.Index(logFile, ".log")], time.Now().UnixMilli())
			err = os.Rename(logFile, failedLogFile)
			if err != nil {
				logger.Log.Errorf("Log file rename failed. Error: %v", err)
				return
			}
			err = fmt.Errorf("Package test failed. Test status line: %s", currLine)
			return
		}
		if strings.Contains(currLine, "CHECK DONE") && strings.Contains(currLine, "EXIT STATUS 0") {
			return
		}
	}
	return
}
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,6 +189,35 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
+		currLine := scanner.Text()
+		// Anything besides 0 is a failed test
+		if strings.Contains(currLine, "CHECK DONE") && !strings.Contains(currLine, "EXIT STATUS 0") {
+			failedLogFile := fmt.Sprintf("%s-FAILED_TEST-%d.log", logFile[:strings.Index(logFile, ".log")], time.Now().UnixMilli())
+			checkErr = os.Rename(logFile, failedLogFile)
+			if checkErr != nil {
+				logger.Log.Errorf("Log file rename failed. Error: %v", checkErr)
+				return
+			}
+			checkErr = fmt.Errorf("Package test failed. Test status line: %s", currLine)
```
PawelWMS, line 211: We try to follow conventions suggested by `go-staticcheck`, which in this case asks to not begin errors with upper case. And to not add a period at the end, but that's not a problem here.
```suggestion
			checkErr = fmt.Errorf("package test failed. Test status line: %s", currLine)
```

If you're using VS Code, then I think [this plug-in](https://marketplace.visualstudio.com/items?itemName=golang.Go) should help with spotting these suggestions.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,6 +189,35 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
+		logger.Log.Errorf("Failed to open log file '%s' while checking package test results. Error: %v", logFile, logErr)
+		checkErr = logErr
+		return
+	}
+	defer file.Close()
+	for scanner := bufio.NewScanner(file); scanner.Scan(); {
+		currLine := scanner.Text()
+		// Anything besides 0 is a failed test
+		if strings.Contains(currLine, "CHECK DONE") && !strings.Contains(currLine, "EXIT STATUS 0") {
+			failedLogFile := fmt.Sprintf("%s-FAILED_TEST-%d.log", logFile[:strings.Index(logFile, ".log")], time.Now().UnixMilli())
```
PawelWMS, line 205: Nit-pick: while I like one-liners, I think we're overdoing it here a bit and lose clarity in the end. Here's my suggestion to break it down a bit:
```suggestion
			failedLogFile := strings.TrimSuffix(logFile, ".log")
			failedLogFile = fmt.Sprintf("%s-FAILED_TEST-%d.log", failedLogFile, time.Now().UnixMilli())
```

I also replaced using string indeces in favour of `strings.TrimSuffix` to make it clearer what our goal here was.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,6 +189,35 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
+			if checkErr != nil {
+				logger.Log.Errorf("Log file rename failed. Error: %v", checkErr)
+				return
+			}
+			checkErr = fmt.Errorf("Package test failed. Test status line: %s", currLine)
+			return
+		}
+		if strings.Contains(currLine, "CHECK DONE") && strings.Contains(currLine, "EXIT STATUS 0") {
+			return
+		}
```
PawelWMS, line 216: Good idea, but let's combine it with the check above, since both begin with `strings.Contains(currLine, "CHECK DONE")`.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -195,6 +227,14 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
 	logBaseName := filepath.Base(srpmFile) + ".log"
 	err = retry.Run(func() (buildErr error) {
 		builtFiles, logFile, buildErr = agent.BuildPackage(srpmFile, logBaseName, outArch, dependencies)
+		// If the package builds with no errors and RUN_CHECK=y, check logs to see if the %check section passed, and if not, return as the build error.
+		if buildErr != nil {
+			return
+		}
+
+		if agent.Config().RunCheck {
+			buildErr = parseCheckSection(logFile)
```
PawelWMS, line 236: Ah, one more thing I just realized: we don't want to fail the whole build if we find a test failure. At least not yet. Once all of our package tests start passing reliably, we can enable that. We might want to also consider adding a build variable like `IGNORE_TEST_FAILURES`, but that's out of scope of this PR.
```suggestion
			parseCheckSection(logfile)
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -195,6 +227,14 @@ func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, ou
...
+		}
```
PawelWMS, line 237: Ah, one more thing I just realized: we don't want to fail the whole build if we find a test failure. At least not yet. Once all of our package tests start passing reliably, we can enable that. We might want to also consider adding a build variable like `IGNORE_TEST_FAILURES`, but that's out of scope of this PR.

We still want to re-run the build on a test failure, so I suggest we make a distinction between build and test failures by introducing a new "flag" variable.

```suggestion
		// If the package builds with no errors and RUN_CHECK=y, check logs to see if the %check section passed, and if not, return as the build error.
		if buildErr != nil {
			return
		}

		if agent.Config().RunCheck {
			buildErr = parseCheckSection(logfile)
			testFailed = buildErr != nil
		}
```

osamaesmailmsft: Hmmm, in this instance then, do I just append to the log stating that there was a failed check and then don't rebuild it?

PawelWMS: Adding a log wouldn't hurt.

I'm not sure what you mean by "don't rebuild it". The idea was to still re-run the build on failing tests but use the `testFailed` flag later to know what caused the build failure and return without errors if the package built fine and the only issue was a test failure.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,17 +189,66 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
 	)
 
 	logBaseName := filepath.Base(srpmFile) + ".log"
-	err = retry.Run(func() (buildErr error) {
+	const checkAttempts = 3
+	build_and_check := func() (buildErr error) {
 		builtFiles, logFile, buildErr = agent.BuildPackage(srpmFile, logBaseName, outArch, dependencies)
+		// If the package builds with no errors and RUN_CHECK=y, check logs to see if the %check section passed, and if not, return as the build error.
+		if buildErr != nil {
+			buildErr = fmt.Errorf("buildErr: %v", buildErr)
```
PawelWMS, line 232: No need, caller will print the error if needed anyway.
```suggestion
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,17 +189,66 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
+			err = os.Rename(logFile, failedLogFile)
+			if err != nil {
+				logger.Log.Errorf("Log file rename failed. Error: %v", err)
+				return
+			}
+			err = fmt.Errorf("package test failed. Test status line: %s", currLine)
+			return
+		} else if strings.Contains(currLine, "CHECK DONE") {
+			return
+		}
```
PawelWMS, line 215: Simplification idea, since we're calling `strings.Contains(currLine, "CHECK DONE")` twice:
```suggestion
		if strings.Contains(currLine, "CHECK DONE") {
			if strings.Contains(currLine, "EXIT STATUS 0") {
				return
			}

			failedLogFile := strings.TrimSuffix(logFile, ".log")
			failedLogFile = fmt.Sprintf("%s-FAILED_TEST-%d.log", failedLogFile, time.Now().UnixMilli())
			err = os.Rename(logFile, failedLogFile)
			if err != nil {
				logger.Log.Errorf("Log file rename failed. Error: %v", err)
				return
			}
			err = fmt.Errorf("package test failed. Test status line: %s", currLine)
			return
		}
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,17 +189,65 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
+
 // buildSRPMFile sends an SRPM to a build agent to build.
 func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, outArch string, dependencies []string) (builtFiles []string, logFile string, err error) {
 	const (
 		retryDuration = time.Second
 	)
 
 	logBaseName := filepath.Base(srpmFile) + ".log"
-	err = retry.Run(func() (buildErr error) {
+	const checkAttempts = 3
```
jslobodzian, line 228: Please use a command line argument for this.  Create a new one like TEST_BUILD_RETRIES and use that for checkAttempts

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,17 +189,65 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
+
+		if agent.Config().RunCheck {
+			buildErr = parseCheckSection(logFile)
+			if buildErr != nil {
+				buildErr = fmt.Errorf("checkErr: %v", buildErr)
+			}
+		}
 		return
-	}, buildAttempts, retryDuration)
+	}
```
jslobodzian, line 243: please add some comments

jslobodzian (approved):

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,18 +189,60 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
 	err = retry.Run(func() (buildErr error) {
 		builtFiles, logFile, buildErr = agent.BuildPackage(srpmFile, logBaseName, outArch, dependencies)
+		// If the package builds with no errors and RUN_CHECK=y, check logs to see if the %check section passed, and if not, return as the build error.
+		if buildErr != nil {
+			return
+		}
+
+		if agent.Config().RunCheck {
+			checkErr := (parseCheckSection(logFile) != nil)
+			if checkErr != nil {
```
PawelWMS, line 240: This won't work:
- `retry. Run` uses `buildErr` to determine if it should retry or not, so we must assign errors to that variable.
- `checker` is a boolean because `(parseCheckSection(logFile) != nil)` is a logical statement, so it will always be `!= nil`.

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -186,18 +189,70 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
-func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, srpmFile, outArch string, dependencies []string) (builtFiles []string, logFile string, err error) {
+func buildSRPMFile(agent buildagents.BuildAgent, buildAttempts int, checkAttempts int, srpmFile, outArch string, dependencies []string) (builtFiles []string, logFile string, err error) {
 	const (
 		retryDuration = time.Second
 	)
 
+	// checkFailed is a flag to see if a non-null buildErr is from the %check section
+	checkFailed := false
 	logBaseName := filepath.Base(srpmFile) + ".log"
+	// temporary solution; potential fix: build normally for buildAttempts, then run rmpbuild -bi --short-circuit to just do the checks
```
PawelWMS, line 231: Thanks for the comment. Might be worth to also add a link to the bug/task tracking this work.

toolkit/tools/scheduler/scheduler.go
```diff
@@ -65,6 +66,7 @@ var (
 	distroBuildNumber    = app.Flag("distro-build-number", "The distro build number that the SRPM will be built with.").Required().String()
 	rpmmacrosFile        = app.Flag("rpmmacros-file", "Optional file path to an rpmmacros file for rpmbuild to use.").ExistingFile()
 	buildAttempts        = app.Flag("build-attempts", "Sets the number of times to try building a package.").Default(defaultBuildAttempts).Int()
+	checkAttempts        = app.Flag("check-attempts", "Sets the minimum number of times to test a package.").Default(defaultCheckAttempts).Int()
```
PawelWMS, line 69: Let's rephrase this is a bit to be clear that the retries only happen in case of a test failures.

PawelWMS (commented):
Looks good. Just 3 minor asks to update comments/docs.

PawelWMS (approved):

PR #4788, opened 2023-02-07
qemu: Add, don't delete, romfiles when building for aarch64

vinceaperri:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Adds missing romfiles to system-aarch64-core package.

###### Change Log  

- Adds missing romfiles to system-aarch64-core package.

**NO**

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/SPECS/qemu/qemu.spec
+++ b/SPECS/qemu/qemu.spec
@@ -217,7 +217,7 @@ Obsoletes: %{name}-system-unicore32-core <= %{version}-%{release}
 Summary:        QEMU is a FAST! processor emulator
 Name:           qemu
 Version:        6.2.0
-Release:        13%{?dist}
+Release:        14%{?dist}
 License:        BSD AND CC-BY AND GPLv2+ AND LGPLv2+ AND MIT
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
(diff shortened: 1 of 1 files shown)


SPECS/qemu/qemu.spec
```diff
@@ -2099,6 +2093,8 @@ useradd -r -u 107 -g qemu -G kvm -d / -s %{_sbindir}/nologin \
 %{_bindir}/qemu-system-aarch64
 %{_datadir}/systemtap/tapset/qemu-system-aarch64*.stp
 %{_mandir}/man1/qemu-system-aarch64.1*
+%{_datadir}/%{name}/pxe*rom
+%{_datadir}/%{name}/efi*rom
```
PawelWMS, line 2097: Double-checking: it seems that now both `qemu-system-aarch64-core` and `qemu-system-x86-core` subpackages will install the same files. Have you verified they don't conflict with each other when installed at the same time?

Also, it might be worth considering moving the common files into a separate subpackage and add `Requires` on it to `qemu-system-aarch64-core` and `qemu-system-x86-core`. That way we are sure to avoid conflicts and remove duplication as a bonus.

PawelWMS (changes requested):
I'd just like to confirm we don't end up with a file conflict when we install `qemu-system-aarch64-core` and `qemu-system-x86-core` at the same time.

SPECS/qemu/qemu.spec
vinceaperri: Thank you for the review and recommendation. Yes, even with the duplication, you are still able to `tdnf install qemu-system-aarch64-core qemu-system-x86-core`. Removing each individually does not remove the romfiles either. However, I do think it is a good idea to remove the duplication anyway. I will create a subpackage named ipxe for this purpose.

PawelWMS (approved):

PR #4832, opened 2023-02-10
set disable_root true in cloud-init.cfg

rmhsawyer:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR is to change disable_root option to true in cloud-init.cfg to prevent SSH logging in for root access. The Penguinator test called verify_no_pre_exist_users() which also failed due to this as when disable_root is set to false. 
From cloud-init: [Root login can be enabled/disabled using the disable_root config key.](https://canonical-cloud-init.readthedocs-hosted.com/en/latest/reference/modules.html)
Verify_no_pre_exist_users() checks if [matching pattern string](https://github.com/microsoft/lisa/blob/ae6f8998fafc4b8c8b746ad67f1c5f7d9f11acf9/microsoft/testsuites/core/azure_image_standard.py#L892) exits in /root/.ssh/authorized_keys and if not, fails the test. disable_root is responsible to add that matching string.

###### Change Log  

- set disable_root to true in cloud-init.cfg. This will add extra strings in /root/.ssh/authorized_keys: 
image

YES

###### Associated issues  

- [Bug 42949053](https://microsoft.visualstudio.com/OS/_workitems/edit/42949053): [Mariner 2] root user has authorized_keys file when deployed to Azure
- [Bug 43185302](https://microsoft.visualstudio.com/OS/_workitems/edit/43185302): [Penguinator Distro test] - verify_no_pre_exist_users failing

###### Test Methodology

- local image build 
- Verify "verify_no_pre_exist_users()" Penguinator test passed:
https://penguinatorstarter.powerappsportals.com/mytests/myjobs/results/?id=80c1d1ea-c7a8-ed11-a81c-0022484b8df1
https://penguinatorstarter.powerappsportals.com/mytests/myjobs/results/?id=f932a4f0-cfa8-ed11-a81c-0022484b8df1


christopherco (changes requested):
Change itself seems reasonable but in the PR description, please explain the original failure and any relevant error messages, and include what effect we should see with this change (i.e., what are the exact string changes we can expect in /root/.ssh/authorized_keys)

PR #4846, opened 2023-02-13
Add dracut-megaraid package

vinceaperri:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR adds the dracut-megaraid package that contains configuration for dracut to add megaraid drivers to initramfs for HCI scenarios

###### Change Log  

- Add dracut-megaraid package
- Add regeneration hook for it in initramfs package

**NO**

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/SPECS/initramfs/initramfs.spec
+++ b/SPECS/initramfs/initramfs.spec
@@ -1,7 +1,7 @@
 Summary:        initramfs
 Name:           initramfs
 Version:        2.0
-Release:        10%{?dist}
+Release:        11%{?dist}
 License:        Apache License
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -19,7 +19,7 @@ install -D -m644 %{SOURCE0} %{buildroot}%{_sysconfdir}/dracut.conf.d/
 install -d -m755 %{buildroot}%{_localstatedir}/lib/initramfs/kernel
 
 %define watched_path %{_sbindir} %{_libdir}/udev/rules.d %{_libdir}/systemd/system /lib/modules %{_sysconfdir}/dracut.conf.d
-%define watched_pkgs e2fsprogs, systemd, kpartx, device-mapper-multipath, verity-read-only-root, dracut-fips
+%define watched_pkgs e2fsprogs, systemd, kpartx, device-mapper-multipath, verity-read-only-root, dracut-fips, dracut-megaraid
 
 %define removal_action() rm -rf %{_localstatedir}/lib/rpm-state/initramfs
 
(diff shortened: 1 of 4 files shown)


SPECS/initramfs/initramfs.spec
```diff
@@ -88,15 +88,17 @@ echo "initramfs (re)generation" %* >&2
 cat > /dev/null \
 if [ -f %{_localstatedir}/lib/rpm-state/initramfs/regenerate ]; then \
     echo "(re)generate initramfs for all kernels," %* >&2 \
-    mkinitrd -q \
+    mkinitrd --preload megaraid_sas --preload megaraid -q \
     mv /boot/initrd.img-*mshv* /boot/efi/ \
 elif [ -d %{_localstatedir}/lib/rpm-state/initramfs/pending ]; then \
     for k in `ls %{_localstatedir}/lib/rpm-state/initramfs/pending/`; do \
         echo "(re)generate initramfs for $k," %* >&2 \
         if [[ $k == *mshv* ]]; then \
```
romoh, line 96: what's mshv? 
Mainly trying to understand if preloading the megaraid drivers there makes sense.

vinceaperri: kernel-mshv. This PR adds megaraid* drivers regardless of the kernel variant installed.

SPECS/initramfs/initramfs.spec
```diff
@@ -88,15 +88,17 @@ echo "initramfs (re)generation" %* >&2
 cat > /dev/null \
 if [ -f %{_localstatedir}/lib/rpm-state/initramfs/regenerate ]; then \
     echo "(re)generate initramfs for all kernels," %* >&2 \
-    mkinitrd -q \
+    mkinitrd --preload megaraid_sas --preload megaraid -q \
```
dmcilvaney, line 91: I have some concerns about forcing these drivers into every image's initramfs. Would it be acceptable to move this to the dracut conf files? You can either add that line in a post install script + invoke mkinird again, or if you want something a bit more stable maybe we could consider adding a package? (I had to add drivers for my verity module here:  https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS/verity-read-only-root/verity.conf . There are both normal driver directives, and a driver_force option.

dmcilvaney, line 91: I have concnerns about forcing these drivers on all images. dracut has ways to be more selective using conf files.

Off the top of my head I can think of:
Option 1: Using a post-install script, add the drivers to the dracut conf file, then trigger a rebuild of the  initramfs
Option 2: Add a package that includes a dedicated conf file in dracut.conf.d/

Following the path from --preload, it looks like this maps to just adding a driver to dracut normally. Dracut also offers a force_drivers option  but I haven't used it personally.

dmcilvaney (changes requested):

SPECS/initramfs/initramfs.spec
dmcilvaney: I had to add some drivers as part of the verity module: https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS/verity-read-only-root/verity.conf

vinceaperri: I like option 2, but I am not inclined to add a top-level package for this.. Would another option be to add a subpackage to dracut? It would be similar to what dracut-fips currently does.

dmcilvaney: That seems like a valid option.

romoh (approved):

PR #4859, opened 2023-02-15
Add maven-jdk17 package version 3.8.7

Sumynwa:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
The default behavior when installing maven is to also install its dependency of msopenjdk-11.
However, having dependency on another msopenjdk-* version without breaking the default
behavior is not available as of now.
For ex:
(a) Requires = <common binary/virtual package>
     By default, this also installs msopenjdk-11. However, in order to pick another msopenjdk-* version, one needs to manually
     install jdk before installing maven. This deviates from default behavior.

(b) Recommends = <>    Weak dependency
      This also has the same behavior as (a)

(c) Suggests = <>  Weak dependency
      This does not cause any msopenjdk to install, requiring manual install.

Since maven needs JAVA_HOME environment to be set, jdk needs to be installed in order to run maven.
Hence, adding a new package to have dependency on msopenjdk-17. 

###### Change Log  
- Add install dependency on msopenjdk-17
- lint spec
- add package entry in cgmanifest
- add package entry in licenses

NO

###### Test Methodology

- Pipeline build id: Buddy Build
                              AMD64
                              [ARM64](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=309626)
                              [Package is building. The Buddy Build fails due to unmet dependency: msopenjdk-17]


cgmanifest.json
```diff
@@ -12436,6 +12436,16 @@
         }
       }
     },
+    {
```
jslobodzian, line 12439: No need to add this twice

SPECS/maven-jdk17/maven_build_caches.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
```
jslobodzian, line 1: No need to duplicate the script (Unless you changed something in it)

SPECS/maven-jdk17/maven-jdk17.spec
```diff
@@ -0,0 +1,247 @@
+%global debug_package %{nil}
```
jslobodzian, line 1: I think you can move this spec file to the regular maven folder.  You don’t need to separate

jslobodzian: See what we did for golang:
https://github.com/microsoft/CBL-Mariner/tree/2.0/SPECS/golang

Sumynwa: We can move the spec file to regular maven folder. But since the difference b/w golang and maven is NEW package name (SAME version), we need to add the same in cgmanifest as well as in licenses mapping..

PR #4878, opened 2023-02-17
Upgrade erlang to version 25.2 to support rabbitmq

sameluch:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

###### Change Log  

- Upgrade erlang to version 25.2

**NO**

###### Test Methodology

- AMD64 Buddy Build ID: 310109, 310883, 316516
- ARM64 Buddy Build ID: 310856, 310890, 316517

### DIFF
--- a/SPECS/erlang/erlang.spec
+++ b/SPECS/erlang/erlang.spec
@@ -1,16 +1,19 @@
 %define  debug_package %{nil}
-Name:         erlang
-Summary:      erlang
-Version:      24.2
-Release:      1%{?dist}
-Group:        Development/Languages
+Summary:        erlang
+Name:           erlang
+Version:        25.2
+Release:        1%{?dist}
+License:        ASL 2.0
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
-License:      ASL 2.0
-URL:          https://erlang.org
-Source0:      https://github.com/erlang/otp/archive/OTP-%{version}/otp-OTP-%{version}.tar.gz
+Group:          Development/Languages
+URL:            https://erlang.org
+Source0:        https://github.com/erlang/otp/archive/OTP-%{version}/otp-OTP-%{version}.tar.gz
+BuildRequires:  unzip
+BuildRequires:  ncurses-devel
+BuildRequires:  openssl-devel
+BuildRequires:  unixODBC
 
-BuildRequires: unzip
 %description
 erlang programming language
 
@@ -19,14 +22,12 @@ erlang programming language
 
 %build
 export ERL_TOP=`pwd`
-./otp_build autoconf
-sh configure --disable-hipe --prefix=%{_prefix}
-
+./configure
 make
 
 %install
 
-make install DESTDIR=%{buildroot}
+%make_install
 
 %post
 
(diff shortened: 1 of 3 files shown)


SPECS/erlang/erlang.spec
```diff
@@ -19,52 +24,59 @@ erlang programming language
...
 %post
 
 %files
 %defattr(-,root,root)
 %license LICENSE.txt
-%{_bindir}/*
-%{_libdir}/*
-%exclude /usr/src
-%exclude %{_libdir}/debug
+%{usr_local_bin}/ct_run
```
PawelWMS, line 39: Do you know why we suddenly moved to the local `bin` and `lib` directories? I currently don't think it's a big deal but for the sake of consistency, I'm curious, if we can avoid that by restoring the `--prefix=%{_prefix}` switch in `./configure`.

PawelWMS (commented):
Looks good. I'm only curious, if we can keep the original installation directories, but that's a minor request.

SPECS/erlang/erlang.spec
sameluch: I will give this a try and see if there is any impact/differences. /usr/local/bin and /usr/local/lib are simply the default install locations when ./configure is run with no args, hence the switch of install directories. I suspect there isn't much of a functional difference in this case, but having the consistency between versions is a good catch.

SPECS/erlang/erlang.spec
```diff
@@ -1,16 +1,21 @@
...
-Version:      24.2
-Release:      1%{?dist}
-Group:        Development/Languages
+%define  usr_local_bin %{_prefix}/local/bin
+%define  usr_local_lib %{_prefix}/local/lib
+Summary:        erlang
+Name:           erlang
+Version:        25.2
+Release:        1%{?dist}
+License:        ASL 2.0
```
oliviacrain, line 8: Can we use an SPDX license expression here?

```suggestion
License:        Apache-2.0
```

SPECS/erlang/erlang.spec
```diff
@@ -19,52 +24,59 @@ erlang programming language
 
 %build
 export ERL_TOP=`pwd`
-./otp_build autoconf
-sh configure --disable-hipe --prefix=%{_prefix}
-
+./configure
```
oliviacrain, line 27: Can we call the `%configure` macro here? It will set our hardened build flags and should fix the `/usr/local/bin` problem Pawel pointed out.

sameluch: After updating the spec file to use this `%configure`, I ran both an AMD64 and ARM64 buddy build, with successful results, indicating that in fact was the issue causing the use of the default install location of `/usr/local/bin` and `/usr/local/lib`. The define tags have been removed accordingly and files listed in the `%files` section of the spec have been updated to use `_bindir` and `_libdir` respectively.

SPECS/erlang/erlang.spec
sameluch: Updated

SPECS/erlang/erlang.spec
sameluch: As noted below by Olivia, changing the `./configure` to instead use `%configure` has resolved the need for the use of the `/usr/local` directories.

oliviacrain (approved):

PawelWMS (approved):

oliviacrain (approved):

PR #4977, opened 2023-03-02
Only conditionally move kernel-mshv initrd if it exists

Camelron:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- conditionally move kernel-mshv initrd if it exists

NO

###### Test Methodology

- rpm installed on a kata image, verified it reboots and systemd reinstall works. 

Fixes: 
- https://github.com/microsoft/CBL-Mariner/pull/4436

### DIFF
--- a/SPECS/initramfs/initramfs.spec
+++ b/SPECS/initramfs/initramfs.spec
@@ -89,7 +89,7 @@ cat > /dev/null \
 if [ -f %{_localstatedir}/lib/rpm-state/initramfs/regenerate ]; then \
     echo "(re)generate initramfs for all kernels," %* >&2 \
     mkinitrd -q \
-    [ ! -f /boot/initrd.img-*mshv* ] || mv /boot/initrd.img-*mshv* /boot/efi/ \
+    mv /boot/initrd.img-*mshv* /boot/efi/ >/dev/null 2>&1 || : \
 elif [ -d %{_localstatedir}/lib/rpm-state/initramfs/pending ]; then \
     for k in `ls %{_localstatedir}/lib/rpm-state/initramfs/pending/`; do \
         echo "(re)generate initramfs for $k," %* >&2 \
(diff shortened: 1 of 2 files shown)


anphel31 (approved):

SPECS/initramfs/initramfs.spec
```diff
@@ -89,7 +89,7 @@ cat > /dev/null \
 if [ -f %{_localstatedir}/lib/rpm-state/initramfs/regenerate ]; then \
     echo "(re)generate initramfs for all kernels," %* >&2 \
     mkinitrd -q \
-    mv /boot/initrd.img-*mshv* /boot/efi/ \
+    [ ! -f /boot/initrd.img-*mshv* ] || mv /boot/initrd.img-*mshv* /boot/efi/ \
```
mfrw, line 92: nitpick
Should we have a simple `if [check]; then` instead of this :)
It took me a while to understand what was happening.

Thanks

SPECS/initramfs/initramfs.spec
```diff
@@ -107,7 +107,7 @@ echo "initramfs" %{version}-%{release} "posttrans" >&2
 %removal_action
 mkinitrd -q
 # Move initrd generated for kernel-mshv to /boot/efi, where linuxloader expects to find it
-mv /boot/initrd.img-*mshv* /boot/efi/
+[ ! -f /boot/initrd.img-*mshv* ] || mv /boot/initrd.img-*mshv* /boot/efi/
```
mfrw, line 110: Same comment as above

SPECS/initramfs/initramfs.spec
hbeberman: The double negative tripped me up too.

Instead of doing any if conditionals, I suggest doing the following to just suppress the failure exit code from the `mv` if the file didn't exist. Seems to be a fairly common pattern in other install/uninstall RPM scriptlets.
```
mv /boot/initrd.img-*mshv* /boot/efi/ || :
```

rakshaa2000 (approved):

SPECS/initramfs/initramfs.spec
hbeberman: Thought about this some more, and that'd still result in the error being printed during RPM transactions, which isn't great. Looks like a lot of other specs are using a `>/dev/null 2>&1 || :` pattern in this scenario.

```
mv /boot/initrd.img-*mshv* /boot/efi/ >/dev/null 2>&1 || :
```

PR #4992, opened 2023-03-06
Separating buildAttempts & checkAttempts in build tooling & edge case fix

osamaesmailmsft:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Other parts of the pipeline couldn't parse the log names to mark test pass/fail because of the log renaming, so the last attempt isn't getting renamed anymore. Also, the logic has been fleshed out more for buildAttempts / checkAttempts.

###### Change Log  

- Doing retry.Run multiple times to separate buildAttempts, checkAttempts, and the last buildAttempt (not renaming log file on check failure for this single attempt).

NO

###### Test Methodology

- Pipeline build id: 322456

### DIFF
--- a/toolkit/tools/scheduler/schedulerutils/buildworker.go
+++ b/toolkit/tools/scheduler/schedulerutils/buildworker.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/file"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/pkggraph"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/retry"
@@ -191,14 +192,14 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
 
 // parseCheckSection reads the package build log file to determine if the %check section passed or not
 func parseCheckSection(logFile string) (err error) {
-	file, err := os.Open(logFile)
+	logFileObject, err := os.Open(logFile)
 	// If we can't open the log file, that's a build error.
 	if err != nil {
 		logger.Log.Errorf("Failed to open log file '%s' while checking package test results. Error: %v", logFile, err)
 		return
 	}
-	defer file.Close()
-	for scanner := bufio.NewScanner(file); scanner.Scan(); {
+	defer logFileObject.Close()
+	for scanner := bufio.NewScanner(logFileObject); scanner.Scan(); {
 		currLine := scanner.Text()
 		// Anything besides 0 is a failed test
 		if strings.Contains(currLine, "CHECK DONE") {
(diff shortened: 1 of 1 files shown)


PawelWMS (changes requested):
I think for now we could go for something simpler and keep the code cleaner: instead of moving the log file for errors, let's only copy it. This way we will keep the logs for all failing builds as well as maintain the original behaviour where the original log file path pointed to the most recent build.

Here's what I have in mind:
- Revert the current changes.
- Copy the log file instead of renaming it, which should only require replacing `os.Rename` with `file.Copy`.
    **NOTE**: `file.Copy` is our internal function. Yes, default Go libraries don't have a function for copying files. Or at least they didn't in the previous versions. :)
- Let's also save a copy of the log for regular build failures. I'd suggest extracting to log copying logic into a separate function and calling that build and test failures.

osamaesmailmsft:
> I think for now we could go for something simpler and keep the code cleaner: instead of moving the log file for errors, let's only copy it. This way we will keep the logs for all failing builds as well as maintain the original behaviour where the original log file path pointed to the most recent build.
> 
> Here's what I have in mind:
> 
> * Revert the current changes.
> * Copy the log file instead of renaming it, which should only require replacing `os.Rename` with `file.Copy`.
>   **NOTE**: `file.Copy` is our internal function. Yes, default Go libraries don't have a function for copying files. Or at least they didn't in the previous versions. :)
> * Let's also save a copy of the log for regular build failures. I'd suggest extracting to log copying logic into a separate function and calling that build and test failures.

I reverted the previous version that was a bit complicated / messy; does this line up more with what you were suggesting?

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -207,7 +208,7 @@ func parseCheckSection(logFile string) (err error) {
 			}
 			failedLogFile := strings.TrimSuffix(logFile, ".log")
 			failedLogFile = fmt.Sprintf("%s-FAILED_TEST-%d.log", failedLogFile, time.Now().UnixMilli())
-			err = os.Rename(logFile, failedLogFile)
+			err = file.Copy(logFile, failedLogFile)
```
jslobodzian, line 211: I think this is clever, but isn't the logfile still open at the point?  Is the content guaranteed to be flushed to disk before making the copy?

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -191,14 +192,14 @@ func getBuildDependencies(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, g
...
+	logFileObject, err := os.Open(logFile)
 	// If we can't open the log file, that's a build error.
 	if err != nil {
 		logger.Log.Errorf("Failed to open log file '%s' while checking package test results. Error: %v", logFile, err)
 		return
 	}
-	defer file.Close()
-	for scanner := bufio.NewScanner(file); scanner.Scan(); {
+	defer logFileObject.Close()
+	for scanner := bufio.NewScanner(logFileObject); scanner.Scan(); {
```
jslobodzian, line 202: Is the logfile re-written on the next loop?  It seems to me that the logfileObject will not have been closed, or flushed to disk.  I am worried this is going to either append to the logfileObject from the previous loop or there's an error here because the logfile is already open.

toolkit/tools/scheduler/schedulerutils/buildworker.go
PawelWMS: The log file is still open but that shouldn't be an issue, since copying it should only read the file.

toolkit/tools/scheduler/schedulerutils/buildworker.go
osamaesmailmsft: `logFileObject` isn't written to at any point during this function; the loop just feeds the log file one line at a time to make parsing cleaner.

PawelWMS (approved):
Looks good to me. Leaving the suggestion to save logs for build failures for extra credit, but I think we're done with the issue we were trying to resolve.

jslobodzian (approved):

PR #5061, opened 2023-03-10
Dev cross main 

jslobodzian:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


SPECS-CROSS/cross-gcc/cross-gcc.spec
```diff
@@ -0,0 +1,161 @@
+%global debug_package %{nil}
```
jslobodzian, line 1: There is inconsistency in the naming of these cross compiler packages.  Why cross-gcc instead of gcc-cross?

Also I noticed that Fedora prefers a syntax like gcc-aarch64-linux so that you can see what kind of cross compiler it is.

Since we are setting the precedent here, we should decide on some uniform conventions for all of our cross compilers

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+%global _target_arch    aarch64
+%global _tuple          %{_target_arch}-%{_vendor}-linux-gnu
+%global _cross_name     %{_target_arch}-%{_vendor}-linux-gnu
+
+# Folders which should be in our macro file
+%global _opt                /opt/
+%global _crossdir           /opt/cross/
+
+# Generally we include '/usr' in most paths.
+# Can we also use '/usr' for our paths? This will bring us in line with the
```
jslobodzian, line 13: Note that the comment here is a question.  We probably should lead out with an actual position, and remove the comment,

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+# %%configure macro which sets these.
+%global _bindir            /bin
```
jslobodzian, line 15: Aren't most of these macros already define globally here:
https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS/mariner-rpm-macros/macros

jslobodzian: let's clean them up

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+# need logic to switch modes something like this:
+%if "%{_target_arch}" != "%{_host_arch}"
+%global _cross_prefix       %{_crossdir}%{_tuple}/
+%global _cross_sysroot      %{_crossdir}%{_tuple}/sysroot/
+%global _cross_includedir   /usr/%{_host}/%{_tuple}/include/
+%global _cross_infodir      %{_crossdir}%{_tuple}/share/info
+%global _cross_bindir       %{_tuple}/bin
+%global _cross_libdir       %{_tuple}/lib
+%global _tuple_name         %{_tuple}-
+%else
```
jslobodzian, line 42: Isn't this dead code?  target_arch and host_arch will always be different

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+%global _cross_sysroot      %{nil}
+%global _cross_includedir   %{_includedir}
+%global _cross_infodir      %{_infodir}
+%global _cross_bindir       %{_bindir}
+%global _cross_libdir       %{_libdir}
+%global _tuple_name         %{nil}
+%endif
+
+Summary:        Contains a linker, an assembler, and other tools
+Name:           %{_cross_name}-binutils
```
jslobodzian, line 53: Have we considered sharing this SPEC file with standard binutils?  Otherwise we need to make double changes whenever binutils (or glibc or gcc) are updated.

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+make %{?_smp_mflags}
+
+%install
+make %{?_smp_mflags} DESTDIR=%{buildroot} install
+
+find %{buildroot} -name '*.la' -delete
+rm -rf %{buildroot}/%{_cross_infodir}
+%find_lang %{name} --all-name
+
+%files -f %{name}.lang
```
jslobodzian, line 93: Why .lang?

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
+# Globals which should be in a macro file.
+# These should be set programatically in the future.
+%global _host_arch      x86_64
+%global _target_arch    aarch64
+%global _tuple          %{_target_arch}-%{_vendor}-linux-gnu
+%global _cross_name     %{_target_arch}-%{_vendor}-linux-gnu
+
+# Folders which should be in our macro file
+%global _opt                /opt/
```
jslobodzian, line 9: Why are these files installed in the /opt folder?

opt is traditionally for 3rd party software.

https://linuxhandbook.com/linux-directory-structure/#:~:text=%2Fopt%20%E2%80%93%20Optional%20software%20Traditionally%2C%20the%20%2Fopt%20directory,so%20that%20all%20the%20users%20can%20run%20it.

SPECS-CROSS/binutils_cross/binutils_cross.spec
PawelWMS, line 53: >%{_cross_name}-binutils [](http://example.com/codeflow?start=16&length=23)

We try to keep the `Name` tag and the spec file name the same.

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+%{_cross_prefix}%{_bindir}/%{_tuple_name}elfedit
+%{_cross_prefix}%{_bindir}/%{_tuple_name}ranlib
+%{_cross_prefix}%{_bindir}/%{_tuple_name}readelf
+%{_cross_prefix}%{_bindir}/%{_tuple_name}strip
+
+%{_cross_prefix}%{_cross_libdir}/ldscripts/*
+#%%{_cross_prefix}%{_cross_libdir}/bfd-plugins/libdep.so
+"/opt/cross/aarch64-mariner-linux-gnu/lib/bfd-plugins/libdep.so"
+
+%{_cross_prefix}%{_mandir}/man1/%{_tuple_name}readelf.1
```
jslobodzian, line 139: should the documentation go into a separate package?

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+%{_cross_prefix}%{_mandir}/man1/%{_tuple_name}nm.1
+%{_cross_prefix}%{_mandir}/man1/%{_tuple_name}ar.1
+%{_cross_prefix}%{_mandir}/man1/%{_tuple_name}ld.1
+%{_cross_prefix}%{_mandir}/man1/%{_tuple_name}dlltool.1
+%{_cross_prefix}%{_mandir}/man1/%{_tuple_name}addr2line.1
+%{_cross_prefix}%{_mandir}/man1/%{_tuple_name}windres.1
+%{_cross_prefix}%{_mandir}/man1/%{_tuple_name}size.1
+%{_cross_prefix}%{_mandir}/man1/%{_tuple_name}objdump.1
+
+# These are the -devel files, we are not generating them.
```
jslobodzian, line 158: There is a lot of commented out content here.  Essentially dead code.  Do we need the headers?

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+./configure \
+            --prefix=%{_cross_prefix} \
+            --target=%{_tuple} \
+            --disable-multilib \
+            --with-sysroot=%{_cross_sysroot}
+
+make %{?_smp_mflags}
+
+%install
+make %{?_smp_mflags} DESTDIR=%{buildroot} install
```
PawelWMS, line 87:
```suggestion
%make_install
```

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+
+%build
+# Ideally we would like to model this after the %%configure macro in the future.
+./configure \
+            --prefix=%{_cross_prefix} \
+            --target=%{_tuple} \
+            --disable-multilib \
+            --with-sysroot=%{_cross_sysroot}
+
+make %{?_smp_mflags}
```
PawelWMS, line 84: Can we use our macros or is this way of building required for the cross-compilation bits to be built correctly?
```suggestion
%make_build
```

toolkit/scripts/toolchain/cross/launch_qemu.sh
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) Microsoft Corporation.
```
jslobodzian, line 1: what is the purpose of this script?

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
+# Globals which should be in a macro file.
```
PawelWMS, line 1: Very general question: how and when are these packages built? I'm curious, if we need the two new spec folder now or maybe we can keep the new specs with the core ones.

toolkit/scripts/toolchain/cross/kernel_cross.sh
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+kernelVersion="5.15.48.1"
```
jslobodzian, line 4: Should this kernel version be fixed?  How do we keep it up to date with the rolling kernel in mariner?

toolkit/scripts/toolchain/cross/kernel_cross.sh
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) Microsoft Corporation.
```
jslobodzian, line 1: what is the purpose of this script?

SPECS-CROSS/cross-gcc/cross-gcc.spec
PawelWMS: +1 for making the tuple part of the package name. If we ever expand the number of supported architectures, then we won't have to worry about conflicts. Also, this way it's clearer what architecture we're targeting with these tools.

toolkit/scripts/toolchain/cross/inittab
```diff
@@ -0,0 +1 @@
+::respawn:-/bin/sh
```
jslobodzian, line 1: What is the purpose of this file?
Is it needed?  Copyright notice?

toolkit/scripts/toolchain/cross/generate_minimal_rootfs.sh
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) Microsoft Corporation.
```
jslobodzian, line 1: What is the purpose of this script?

SPECS-CROSS/binutils_cross/binutils_cross.spec
```diff
@@ -0,0 +1,248 @@
...
+Group:          System Environment/Base
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+ExclusiveArch:  x86_64
+Source0:        http://ftp.gnu.org/gnu/binutils/binutils-%{version}.tar.xz
+# Patch was derived from source: https://src.fedoraproject.org/rpms/binutils/blob/f34/f/binutils-export-demangle.h.patch
+Patch0:         export-demangle-header.patch
+# Patch1 Source https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=6b86da53d5ee2022b9065f445d23356190380746
+Patch1:         linker-script-readonly-keyword-support.patch
+Patch2:         thin_archive_descriptor.patch
```
PawelWMS, line 67: Should we conflict with regular `binutils`? Or can both live together in harmony?:)

SPECS-CROSS/cross-gcc/cross-gcc.spec
PawelWMS, line 1: Why did we need to drop the debug package? We can skip stripping the symbols, but we'll still be missing the sources, which make debugging a lot easier.

toolkit/scripts/toolchain/cross/gcc_cross_test.sh
```diff
@@ -0,0 +1,53 @@
+# This script assumes that gcc_cross.sh has been run with the gcc extracted sources at $HOME/cross/gcc-9.1.0 and the gcc cross compilation binaries at /opt/cross/bin
```
jslobodzian, line 1: Copyright?
What is the purpose fo this script?
Note that the comment says gcc-9.1.0 but Mariner is on 11.2.  Is the version relevant?

toolkit/scripts/toolchain/cross/gcc_cross_test.sh
```diff
@@ -0,0 +1,53 @@
...
+# delete previous test results
+rm -rf ${gccBuildDir}/gcc/testsuite
+
+# version 2021a installed via apt-get
+# version 2019c in tzdata spec file
+DEBIAN_FRONTEND="noninteractive" apt-get -y install tzdata
+
+# version 1.6.1-1 installed via apt-get
+# version 1.6.2 in dejagnu spec file
+sudo apt-get install -y dejagnu
```
jslobodzian, line 16: apt-get is a debian thing?  Is this script needed for Mariner?

toolkit/scripts/toolchain/cross/gcc_cross.sh
```diff
@@ -0,0 +1,103 @@
...
+sudo rm -rf ${sysroot}
+
+mkdir -p ${buildDir}
+mkdir -p ${installDir}
+mkdir -p "${sysroot}/lib"
+
+cd ${buildDir}
+
+# Download source tarballs
+wget http://ftp.gnu.org/gnu/binutils/binutils-2.32.tar.xz
```
jslobodzian, line 37: We cannot download content from the internet.  The sources for these should be in the Mariner blob store.

Also, if this is really needed, each file should be sha256sum checked.

toolkit/scripts/toolchain/cross/gcc_cross.sh
```diff
@@ -0,0 +1,103 @@
...
+wget https://github.com/microsoft/CBL-Mariner-Linux-Kernel/archive/rolling-lts/mariner-2/5.15.48.1.tar.gz
+wget https://ftp.gnu.org/gnu/glibc/glibc-2.28.tar.xz
+wget https://ftp.gnu.org/gnu/mpfr/mpfr-4.0.1.tar.gz
+wget http://ftp.gnu.org/gnu/gmp/gmp-6.1.2.tar.xz
+wget https://ftp.gnu.org/gnu/mpc/mpc-1.1.0.tar.gz
+
+# Unzip source tarballs
+for f in *.tar*; do tar xf $f; done
+
+cd gcc-9.1.0
```
jslobodzian, line 48: Mariner does not have gcc 9

SPECS-CROSS/cross-gcc/cross-gcc.spec
```diff
@@ -0,0 +1,161 @@
...
+
+# If we want our cross compile aware packges to also support native, we
+# need logic to switch modes something like this:
+%if "%{_target_arch}" != "%{_host_arch}"
+%global _cross_prefix       %{_crossdir}%{_tuple}/
+%global _cross_sysroot      %{_crossdir}%{_tuple}/sysroot/
+%global _cross_includedir   /usr/%{_host}/%{_tuple}/include/
+%global _cross_infodir      %{_crossdir}%{_tuple}/share/info
+%global _cross_bindir       %{_tuple}/bin
+%global _cross_libdir       %{_tuple}/lib
```
PawelWMS, line 39: Some of these macros (`_cross_libdir` for instance) seem unused in this spec. Also, we do we re-define them in each spec? I'm curious, if we could keep them all in one place in a separate "cross-macros" package and only manipulate the value of `_tuple` depending on the architecture we build for.

toolkit/scripts/toolchain/cross/gcc_cross.sh
```diff
@@ -0,0 +1,103 @@
+# Copyright (c) Microsoft Corporation.
```
jslobodzian, line 1: What's the purpose of this script?

toolkit/scripts/toolchain/cross/busybox_cross.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Update the username based on your environment
+# sudo chown username ${busyboxInstallDir}
+
+export PATH="${installDir}/bin":$PATH
+
+cd ${buildDir}
+
+# Download source tarballs
+wget https://www.busybox.net/downloads/busybox-1.32.0.tar.bz2
```
jslobodzian, line 20: cannot wget from the internet per secure supply chain requirements.  Take from Mariner Blob store and run sha256sum

toolkit/scripts/toolchain/cross/busybox_cross.sh
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) Microsoft Corporation.
```
jslobodzian, line 1: What's the purpose of this script?

toolkit/scripts/toolchain/cross/busybox_container.sh
```diff
@@ -0,0 +1,69 @@
...
+            cross_containers="$(docker container ls -aq --filter ancestor=${image})"
+            echo "Removing containers which use that image: ${cross_containers}"
+            [ -n "${cross_containers}" ] && docker rm ${cross_containers}
+            docker image rm -f ${image}
+        done
+    fi
+fi
+
+echo "Configuring QEMU multi arch support"
+docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
```
jslobodzian, line 58: Where is this container procured from?   In production this cannot come from 3rd party.

toolkit/scripts/toolchain/cross/aarch64-linux-gnu.xfail
```diff
@@ -0,0 +1,27 @@
+FAIL: gcc.target/aarch64/fuse_adrp_add_1.c scan-assembler adrp\tx.*, fixed_regs\n\tadd\tx.*, x.*fixed_regs
```
jslobodzian, line 1: What's the purpose of this file?

SPECS-CROSS/cross-gcc/cross-gcc.spec
```diff
@@ -0,0 +1,161 @@
...
+BuildRequires:  %{_cross_name}-glibc-bootstrap2
+BuildRequires:  %{_cross_name}-gcc-bootstrap3
+BuildRequires:  grep
+AutoReqProv:    no
+ExclusiveArch:  x86_64
+Conflicts:      %{_cross_name}-gcc-bootstrap
+Conflicts:      %{_cross_name}-gcc-bootstrap2
+Conflicts:      %{_cross_name}-gcc-bootstrap3
+Conflicts:      %{_cross_name}-glibc-bootstrap
+Conflicts:      %{_cross_name}-glibc-bootstrap2
```
PawelWMS, line 73: Why do we need so many bootstrap packages? Btw. for the sake of building Mariner on top of Ubuntu we've been building our bootstrap packages inside `toolkit/scripts/toolchain/container/toolchain_build_in_chroot.sh`. I like relying solely on specs, but I think we shouldn't multiply these bootstrap packages, as they are hard to maintain. See: recent issue with `cyrus-sasl-bootstrap` and `openldap`. This might be a bigger discussion, but my initial 3 ideas for this are:
- Duplicate the approach from `toolkit/scripts/toolchain/container/toolchain_build_in_chroot.sh`, or
- use `Obsoletes` instead of `Conflicts` similar to what `cyrus-sasl` is doing with `cyrus-sasl-bootstrap` or `systemd` with `systemd-bootstrap`, or
- **[bigger item]** extend our toolkit to support bootstrap packages built from one spec through the use of an additional `bootstrap` macro - Fedora already does something similar in some of their packages, I believe.

SPECS-CROSS/cross-gcc/cross-gcc.spec
```diff
@@ -0,0 +1,161 @@
...
+License:        GPLv2+
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+Group:          Development/Tools
+BuildRequires:  %{_cross_name}-binutils
+BuildRequires:  %{_cross_name}-kernel-headers
+BuildRequires:  %{_cross_name}-glibc-bootstrap2
+BuildRequires:  %{_cross_name}-gcc-bootstrap3
+BuildRequires:  grep
+AutoReqProv:    no
```
PawelWMS, line 67: Why is `AutoReqProv: no` needed?

toolkit/do.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
jslobodzian, line 1: why is this necessary? Once the packages are built and published I think this will no longer be useful and it probably should not be on the main branch

SPECS-DEV/aarch64-none-linux-gnu-toolchain/aarch64-none-linux-gnu-toolchain.spec
```diff
@@ -0,0 +1,32 @@
+Summary:        Prebuilt aarch64-none-linux-gnu cross compiling toolchain
```
jslobodzian, line 1: Is this needed?

SPECS-DEV/aarch64-mariner-linux-gnu-prebuilt-toolchain/aarch64-mariner-linux-gnu-prebuilt-toolchain.spec
```diff
@@ -0,0 +1,31 @@
+Summary:        Prebuilt aarch64-mariner-linux-gnu cross compiling toolchain
```
jslobodzian, line 1: Is this needed?

SPECS-CROSS/kernel-headers_cross/kernel-headers_cross.spec
```diff
@@ -0,0 +1,192 @@
...
+%global _cross_includedir   %{_includedir}
+%global _cross_infodir      %{_infodir}
+%global _cross_bindir       %{_bindir}
+%global _cross_libdir       %{_libdir}
+%global _tuple_name         %{nil}
+%endif
+
+Summary:        Linux API header files
+Name:           %{_cross_name}-kernel-headers
+Version:        5.15.57.1
```
jslobodzian, line 54: How do we keep this current with the latest rolling kernel?

SPECS-CROSS/kernel-headers_cross/kernel-headers_cross.spec
```diff
@@ -0,0 +1,192 @@
...
+Release:        3%{?dist}
+License:        GPLv2
+URL:            https://github.com/microsoft/CBL-Mariner-Linux-Kernel
+Group:          System Environment/Kernel
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+#Source0:       https://github.com/microsoft/CBL-Mariner-Linux-Kernel/archive/rolling-lts/mariner-2/%%{version}.tar.gz
+Source0:        kernel-%{version}.tar.gz
```
PawelWMS, line 62:
```suggestion
Source0:        https://github.com/microsoft/CBL-Mariner-Linux-Kernel/archive/rolling-lts/mariner-2/%{version}.tar.gz#/kernel-%{version}.tar.gz
```

SPECS-DEV/aarch64-mariner-linux-gnu-prebuilt-toolchain/aarch64-mariner-linux-gnu-prebuilt-toolchain.spec
```diff
@@ -0,0 +1,31 @@
...
+Name:           aarch64-mariner-linux-gnu-prebuilt-toolchain
+Version:        0.1.0
+Release:        1%{?dist}
+License:        GPLv2+
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+Group:          Development/Tools
+AutoReqProv:    no
+ExclusiveArch:  x86_64
+Source0:        %{name}-%{version}.tar.gz
```
PawelWMS, line 11: Where is this source coming from? Is it built through a separate step elsewhere similar to how building packages inside `SPECS-SIGNED` work?

toolkit/do.sh
PawelWMS, line 1: It feels like this logic could be converted to a new target in our tooling instead of being kept as a separate package. What do you think?

toolkit/scripts/toolchain/cross/gcc_cross.sh
```diff
@@ -0,0 +1,103 @@
...
+cd ${buildDir}
+
+# Download source tarballs
+wget http://ftp.gnu.org/gnu/binutils/binutils-2.32.tar.xz
+wget https://ftp.gnu.org/gnu/gcc/gcc-9.1.0/gcc-9.1.0.tar.xz
+wget https://github.com/microsoft/CBL-Mariner-Linux-Kernel/archive/rolling-lts/mariner-2/5.15.48.1.tar.gz
+wget https://ftp.gnu.org/gnu/glibc/glibc-2.28.tar.xz
+wget https://ftp.gnu.org/gnu/mpfr/mpfr-4.0.1.tar.gz
+wget http://ftp.gnu.org/gnu/gmp/gmp-6.1.2.tar.xz
+wget https://ftp.gnu.org/gnu/mpc/mpc-1.1.0.tar.gz
```
PawelWMS, line 43: This looks fishy. We shouldn't be reaching upstream without validating the signatures. Also, the upstream might disappear at any time. I'd suggest relying on our blob store.

toolkit/scripts/toolchain/cross/busybox_cross.sh
PawelWMS, line 20: We should use verified sources. In our case that's usually a source downloaded from a blob storage, for which we already have a signature.

SPECS-CROSS/binutils_cross/binutils_cross.spec
dallasd1: This question is still unanswered. We originally decided not to combine the cross binutils/glibc/gcc/kernel-headers because there weren't many usecases to justify making that change and potential bugs that might come with it. 

Ideally we would have one SPEC that covers both native and cross building using `target_arch` and `host_arch` macros, but that seemed more of a second phase of integrating these packages. The other question is do we really want to make binutils/glibc/gcc/kernel-headers cross-enabled while all the other specs aren't cross-enabled?

SPECS-CROSS/binutils_cross/binutils_cross.spec
dallasd1: I think the problem with combining these specs with the core specs is that it may not be obvious that these are "cross" packages that are solely meant to be used to build the toolchain that builds the cross-compiler package `cross-gcc`. The packages in `SPECS-CROSS` are built sequentially due to certain dependencies to eventually build a cross compiler (cross-gcc).

Ideally, these would be built through a pipeline and a user that wants to cross-compile would be able to pull the gcc-cross package from PMC and cross-compile source code.

SPECS-CROSS/binutils_cross/binutils_cross.spec
dallasd1: Both should be able to live together in harmony. This binutils is placed in the "cross" directory defined with the `--prefix` option. To override the native binutils the PATH can be changed to the "cross" directory.

SPECS-CROSS/binutils_cross/binutils_cross.spec
dallasd1: This logic can be removed. It's just to show these changes can be merged into the native binutils spec and triggered by macros those macros.

dallasd1:
If all that's wanted is the cross-compiler functionality, only the files under `SPECS-CROSS` need to be merged. The changes from `SPECS-DEV` and `toolkit` can be ignored/removed from this PR.

jslobodzian:
This is no longer needed and was handled in another way.

PR #5065, opened 2023-03-13
Fix: K3s Containerd CVE-2023-25153 & CVE-2023-25173

guptavinayak11:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

This PR fixes K3s' Containerd CVE-2023-25153 & CVE-2023-25173. The fixes for this are not out on upstream so needed to patch in the changes to my own fork and then update the go.mod and dependencies accordingly. 

The patch has been added for source to k3s folder and new updated vendor tarball has been updated based on the fork created with patches applied.

###### Change Log  

Updated k3s vendor tarball to fix CVE-2023-25153 & CVE-2023-25173
[CVE-2023-25153](https://nvd.nist.gov/vuln/detail/CVE-2023-25153)
[CVE-2023-25173](https://nvd.nist.gov/vuln/detail/CVE-2023-25173)

**NO**

###### Test Methodology

Built locally with Mariner toolkit.
Ran conformance tests for k3s on Mariner VM. 

BuddyBuild: [PR-5065](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=326384&view=results)


SPECS/k3s/k3s-1.24.6.spec
```diff
@@ -1,11 +1,12 @@
...
 Name:           k3s
 Version:        1.24.6
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        ASL 2.0
 Group:          System Environment/Base
 URL:            http://k3s.io
-Source0:        https://github.com/k3s-io/%{name}/archive/refs/tags/v%{version}+k3s1.tar.gz#/%{name}-%{version}.tar.gz
+#Source0:        https://github.com/k3s-io/%{name}/archive/refs/tags/v%{version}+k3s1.tar.gz#/%{name}-%{version}.tar.gz
+Source0:        %{name}-%{version}-v2.tar.gz
```
mfrw, line 9: I would suggest that we leave source0 alone and apply patches.
If that is not something we can achieve, then probably we should follow this:

```suggestion
Source0:        https://github.com/k3s-io/%{name}/archive/refs/tags/v%{version}+k3s1.tar.gz#/%{name}-%{version}-v2.tar.gz
```

SPECS/k3s/k3s-1.24.6.spec
```diff
@@ -79,47 +80,10 @@ exit 0
 %{install_sh}
 
 %changelog
-* Tue Feb 07 2023 Vinayak Gupta <guptavinayak@microsoft.com> - 1.24.6-1
-- Updated k3s to 1.24.6
-
-* Fri Feb 03 2023 CBL-Mariner Servicing Account <cblmargh@microsoft.com> - 1.24.3-6
-- Bump release to rebuild with go 1.19.5
-
-* Wed Jan 18 2023 CBL-Mariner Servicing Account <cblmargh@microsoft.com> - 1.24.3-5
```
mfrw, line 88: Why are we removing the changelog lines ?

SPECS/k3s/k3s-1.24.6.spec
```diff
@@ -20,7 +21,7 @@ Source0:        https://github.com/k3s-io/%{name}/archive/refs/tags/v%{version}+
 # 8. git clone --single-branch --branch="v1.1.4" --depth=1 https://github.com/opencontainers/runc
 # 9. popd
 # 10. tar -cf %%{name}-%%{version}-vendor.tar.gz vendor
-Source1:        %{name}-%{version}-vendor.tar.gz
+Source1:        %{name}-%{version}-vendor-v2.tar.gz
```
mfrw, line 24: This looks good to me :)

SPECS/k3s/k3s-1.24.6.spec
```diff
@@ -1,11 +1,12 @@
 Summary:        Lightweight Kubernetes
 Name:           k3s
```
mfrw, line 2: If this remains the only spec in `k3s/` directory, then I would suggest renaming it to `k3s` and dropping the version suffix for this spec.

guptavinayak11: I will need to add to a few more versions in future. Like k3s 1.25.4 etc. That is why maintaining the version in spec file name would be good thing here.

SPECS/k3s/k3s-1.24.6-cve.patch
```diff
@@ -0,0 +1,74 @@
+diff --git a/go.mod b/go.mod
+index ae51a2598d..c402fe477b 100644
+--- a/go.mod
++++ b/go.mod
+@@ -7,7 +7,7 @@ replace (
+ 	github.com/Mirantis/cri-dockerd => github.com/k3s-io/cri-dockerd v0.2.4-0.20220729204449-bcbb75d5abff // k3s-master
+ 	github.com/cloudnativelabs/kube-router => github.com/k3s-io/kube-router v1.5.1-0.20220630214451-a43bcd8511d2
+ 	github.com/containerd/cgroups => github.com/containerd/cgroups v1.0.1
+-	github.com/containerd/containerd => github.com/k3s-io/containerd v1.5.13-k3s1 // k3s-release/1.5
++	github.com/containerd/containerd => github.com/guptavinayak11/containerd v1.5.14-0.20230311100541-d1bb18d6c070 // k3s-release/1.5
```
mfrw, line 10: Are we sure we want to do this ?

I would request to tag someone within the `k3s` stakeholder team also please take a look and vet this :)

guptavinayak11: Yes, so creating the fork is the only way to apply patches on the go.mod containerd dependency so that go mod vendor can package it correctly and also the go build command can build it. Is there any other way you feel we can do this?

nisamson: The go.mod container dependency is a red herring; the version mentioned in here is not at all relevant to the one used in the build, and changing the version mentioned in the go.mod won't fix the CVE as a result. 

https://github.com/k3s-io/k3s/blob/26e9405767263a2915723cb72b1ffd7f50687a8f/scripts/version.sh#L25-L27
The build process does some monkeypatching of containerd to avoid breaking dependencies. This means they use a forked version of containerd which lags behind the upstream quite a bit. You can see the monkey patching process in `scripts/download`. Their forked version does not currently have a fix for this CVE.

They are planning on fixing this in the [next patch version](https://github.com/k3s-io/k3s/issues/7092) of k3s. For now, I think we should either wait for this fix to be deployed, or apply a patch to the vendored tarball to match what their fix does.

nisamson, line 10: The go.mod container dependency is a red herring; the version mentioned in here is not at all relevant to the one used in the build, and changing the version mentioned in the go.mod won't fix the CVE as a result. 

https://github.com/k3s-io/k3s/blob/26e9405767263a2915723cb72b1ffd7f50687a8f/scripts/version.sh#L25-L27
The build process does some monkeypatching of containerd to avoid breaking dependencies. This means they use a forked version of containerd which lags behind the upstream quite a bit. You can see the monkey patching process in `scripts/download`. Their forked version does not currently have a fix for this CVE.

They are planning on fixing this in the [next patch version](https://github.com/k3s-io/k3s/issues/7092) of k3s. For now, I think we should either wait for this fix to be deployed, or apply a patch to the vendored tarball to match what their fix does.

nisamson (changes requested):

SPECS/k3s/k3s-1.24.6-cve.patch
guptavinayak11: Hi @nisamson, thanks for the comments. I raised this issue with k3s folks and they will take it in their next release. So, we will update this with the fix w/ next upcoming release.

guptavinayak11: I'm planning to close this PR for now. Let me know if u have any issues. Otherwise, I'll close it by tomorrow.

PR #229, opened 2020-10-16
Modify ip4, ip6 defaults from whitelist to blacklist

rlmenge:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 
Issue that ssh connections lost when firewall rules flushed with iptables -F.
Therefore, changed default to ACCEPT all packages instead of DROP for all chains.
This allows ssh connection to maintain connection without needing rule to grant
access again.

###### Change Log  
- Updated ip4save, ip6save defaults to be ACCEPT rather than DROP and added rules to drop known offenders
- Updated iptables.spec to use autosetup and bumped revision number
- Modified the ip4save, ip6save to match rules of other distros

###### Does this affect the toolchain?  

NO

###### Test Methodology

- local build

PR #5071, opened 2023-03-13
Update default iptable rules

AZaugg:
Change the default OUTPUT chain to target ACCEPT, a shortcut to not needing the Accept rule within the output chain.

Adding ICMP as a default ACCEPT rule within the INPUT chain, helps with debugging. In addition, ICMP is essentially for the proper operation of ipv6.

###### Merge Checklist  

- [X ] The toolchain has been rebuilt successfully (or no changes were made to it)
- [X ] The toolchain/worker package manifests are up-to-date
- [ X] Any updated packages successfully build (or no packages were changed)
- [X ] All package sources are available
- [X ] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [X ] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [X ] All source files have up-to-date hashes in the `*.signatures.json` files
- [X ] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [X ] Documentation has been updated to match any changes to the build system
- [X ] Ready to merge

---

###### Summary 

###### Change Log  

- Changed OUTPUT chain to have an ACCEPT Policy
- Accept ICMP traffic

**NO**

###### Test Methodology

- Added additional rules in /etc/sysconfig/ip[6]tables and restarted iptables service, verified by IP[6]tables -L that rules were added


SPECS/iptables/ip4save
```diff
@@ -2,12 +2,12 @@
 *filter
 :INPUT DROP [0:0]
 :FORWARD DROP [0:0]
-:OUTPUT DROP [0:0]
+:OUTPUT ACCEPT [0:0]
 # Allow local-only connections
 -A INPUT -i lo -j ACCEPT
 -A INPUT -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
+-A INPUT -p icmp -j ACCEPT
```
PawelWMS, line 9: This will change the default configuration for all Mariner images, which may or may not be a significant change for some of our users. @AZaugg, @rlmenge, can you guarantee that this is a benign change? I'm not an expert in this topic, unfortunately, but it's best to be careful about such system-wide changes.

@AZaugg, why is this change necessary for all of Mariner?

PawelWMS (changes requested):

SPECS/iptables/ip6save
```diff
@@ -2,12 +2,12 @@
 *filter
 :INPUT DROP [0:0]
 :FORWARD DROP [0:0]
-:OUTPUT DROP [0:0]
+:OUTPUT ACCEPT [0:0]
 # Allow local-only connections
 -A INPUT -i lo -j ACCEPT
 -A INPUT -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
+-A INPUT -p icmpv6 -j ACCEPT
```
PawelWMS, line 9: Similar concerns and questions here.

SPECS/iptables/iptables
```diff
@@ -43,11 +43,11 @@ echo 0 > /proc/sys/net/ipv4/tcp_ecn
 # Set a known state
 iptables -P INPUT   DROP
 iptables -P FORWARD DROP
-iptables -P OUTPUT  DROP
+iptables -P OUTPUT  ACCEPT
 
 ip6tables -P INPUT   DROP
 ip6tables -P FORWARD DROP
-ip6tables -P OUTPUT  DROP
+ip6tables -P OUTPUT  ACCEPT
```
PawelWMS, line 50: This seems like a big change to the default configuration. Why do we need it? Can this cause any security concerns?

SPECS/iptables/ip4save
```diff
@@ -2,12 +2,12 @@
 *filter
 :INPUT DROP [0:0]
 :FORWARD DROP [0:0]
-:OUTPUT DROP [0:0]
+:OUTPUT ACCEPT [0:0]
```
PawelWMS, line 5: To be honest, I like the idea of starting the config with a set of deny-all rules and then enabling only what's needed. I understand that in this case we end up accepting all output later on, but now we're introducing an inconsistency in the config for the sake of saving one line but at a cost of potentially allowing too much output if we ever want to change this policy at some point in the future.

SPECS/iptables/ip6save
AZaugg: So, from an IPv6 perspective, icmp is essential for v6. For example, arp is replaced by NDP with ndp operating over icmp

SPECS/iptables/ip4save
AZaugg: From my point of view it helps with debugging. We had a new server being provisioned and it included the iptables package, the system failed to build correctly and debugging was a lot more involved. ping inbound were being dropped, which lead us down several paths of "network" issues. with iptables, not the first place we looked.

In addition, certain icmp messages are important, for example, "icmp fragmentation required". I think it is used for MTU and MSS negotiation.

SPECS/iptables/ip6save
franckhlmartin: Correct, if icmpv6 is completely disabled the server cannot function properly as Router Advertisement and Neighbor Discovery is done over icmpv6

SPECS/iptables/ip4save
AZaugg: Apologies forgot to reply to this comment. From an ops point of view, setting the policy as DROP as in
iptables -P OUTPUT DROP 
can be a little problematic. If someone logs into the machine and does a iptables -F, it will clear all the rules, and the default policy is to close the machine down, and you would completely lose access.

If it is set to ACCEPT as a default policy, the machine would fail open, allowing access. So I guess it is less secure. 

From my point of view, most places have security by depth. So the tradeoff of it failing open is better from an operational point of view.

In addition, I would prefer to have the default policy on my chains as ACCEPT, and have explicit rules to drop traffic but via a REJECT as in
iptables -A [INPUT|OUTPUT] -j REJECT

This way if debugging a network issue, you would still get the TCP resets from your server. This is immensely useful to help debug server access issues.


Let me know, I am happy to set this to DROP

SPECS/iptables/ip6save
AZaugg: If you want, I can reduce it to the essential icmp messages for V6. I believe that the following are essential for ipv6 operation

```
-p udp  --dport 546 -d fe80::/64 
-A -p icmpv6 ---icmpv6-type 133 -j ACCEPT
-A -p icmpv6 --icmpv6-type 134  -j ACCEPT
-A -p icmpv6 --icmpv6-type 135  -j ACCEPT
-A -p icmpv6 --icmpv6-type 136  -j ACCEPT
-A -p icmpv6 --icmpv6-type 137  -j ACCEPT
```

https://www.iana.org/assignments/icmpv6-parameters/icmpv6-parameters.xhtml

franckhlmartin: Another option missing is type 2 Packet Too Big PTB, this is important for the host to receive it as it may have to retry the communication with a smaller packet.

SPECS/iptables/ip4save
rlmenge: We had a similar discussion a few years ago related to this PR incase it is helpful https://github.com/microsoft/CBL-Mariner/pull/229

SPECS/iptables/ip4save
rlmenge: Spoke with @chaitanyalala and this goes against our "secure by default" policy where customers (running on both bare metal or containers) should have the highest security possible by default and then can customize and accept ports / packets as needed.

SPECS/iptables/iptables
rlmenge: This introduces a (brief) period of time where the only rules are to ACCEPT everything. I would say this is a concern

rlmenge (changes requested):
Adding myself as a reviewer

rlmenge:
Hello! Many of these changes were merged in https://github.com/microsoft/CBL-Mariner/pull/5609 . I believe this PR can now be closed. Note that we are looking into changing the defaults for Mariner 3.0 including setting default policy.

PR #5131, opened 2023-03-21
Create kernel specific subpackages for specialized kernels

rlmenge:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

There was a bug where certain specialized kernels (`kernel-azure`, `kernel-hci`, `kernel-rt`) would create subpackages of the same name (`bpftool`, `python3-perf`) as the generic kernel. This will cause packages to be published and overwritten. **For now,** create specialized version names (ex. `kernel-azure-bpftool`) of the subpackages from specialized kernels. Explicitly call out the conflict with the generic kernel's subpackages. This should be investigated further for a less involved solution.

Subpackage changes from kernel-azure
`bpftool` -> `bpftool-azure` which conflicts with `bpftool` and variants
`python3-perf `-> `python3-perf-azure` which conflicts with `python3-perf` and variants

Subpackage changes from kernel-hci 
`bpftool` -> `bpftool-hci` which conflicts with `bpftool` and variants
`python3-perf` ->python3-perf-hci` which conflicts with `python3-perf` and variants

Subpackage changes from kernel-rt
`bpftool` -> `bpftool-rt` which conflicts with `bpftool` and variants
`python3-perf` ->python3-perf-rt` which conflicts with `python3-perf` and variants

###### Change Log  

- Add conflicts with the other subpackages

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/43814917/

###### Test Methodology

- (in progress) https://dev.azure.com/mariner-org/mariner/_build/results?buildId=330766&view=results
- Buddy Build Pipeline https://dev.azure.com/mariner-org/mariner/_build/results?buildId=330010&view=results
- Tested
image

### DIFF
--- a/SPECS-EXTENDED/kernel-rt/kernel-rt.spec
+++ b/SPECS-EXTENDED/kernel-rt/kernel-rt.spec
@@ -60,6 +60,11 @@ ExclusiveArch:  x86_64
 %ifarch x86_64
 BuildRequires:  pciutils-devel
 %endif
+Conflicts:      kernel
+Conflicts:      kernel-azure
+Conflicts:      kernel-hci
+Conflicts:      kernel-mshv
+Conflicts:      kernel-uvm
 # When updating the config files it is important to sanitize them.
 # Steps for updating a config file:
 #  1. Extract the linux sources into a folder
@@ -124,9 +129,7 @@ This package contains the 'perf' performance analysis tools for Linux kernel.
 %package -n     python3-perf-rt
 Summary:        Python 3 extension for perf tools
 Requires:       python3
-Conflicts:      python3-perf
-Conflicts:      python3-perf-azure
-Conflicts:      python3-perf-hci
+Requires:       %{name} = %{version}-%{release}
 
 %description -n python3-perf-rt
 This package contains the Python 3 extension for the 'perf' performance analysis tools for Linux kernel.
@@ -140,9 +143,7 @@ This package contains common device tree blobs (dtb)
 
 %package -n     bpftool-rt
 Summary:        Inspection and simple manipulation of eBPF programs and maps
-Conflicts:      bpftool
-Conflicts:      bpftool-azure
-Conflicts:      bpftool-hci
+Requires:       %{name} = %{version}-%{release}
 
 %description -n  bpftool-rt
 This package contains the bpftool, which allows inspection and simple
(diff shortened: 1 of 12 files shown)


SPECS-EXTENDED/kernel-rt/kernel-rt.spec
```diff
@@ -121,11 +121,12 @@ Requires:       audit
 %description tools
 This package contains the 'perf' performance analysis tools for Linux kernel.
 
-%package -n     python3-perf
+%package        python3-perf
 Summary:        Python 3 extension for perf tools
 Requires:       python3
+Conflicts:      python3-perf
```
PawelWMS, line 127: I'm not sure if it's so straightforward, since all of these subpackages conflict with each other. As in: we might need to list the subpackages from all kernel specs we have. And then keep it up-to-date with every new kernel spec.

Also, did we figure out, if we actually cannot these subpackages only from `kernel.spec`? Now I don't remember. That would simplify this mess a lot. I know we already don't produce these subpackages from some kernel specs anyway (example: `kernel-mshv.spec`).

SPECS-EXTENDED/kernel-rt/kernel-rt.spec
```diff
@@ -121,11 +121,12 @@ Requires:       audit
 %description tools
 This package contains the 'perf' performance analysis tools for Linux kernel.
 
-%package -n     python3-perf
+%package        python3-perf
```
mfrw, line 124: We should probably document it somewhere that we have specific `perf/bpftool` for `kernel-rt/hci`.

SPECS-EXTENDED/kernel-rt/kernel-rt.spec
mfrw: +1 on the additional overhead of maintaining the `Conflicts:`.
Also, should we add an additional `Conflicts` in the main kernel spec ?

SPECS/kernel-hci/kernel-hci.spec
```diff
@@ -136,11 +136,12 @@ Requires:       audit
 %description tools
 This package contains the 'perf' performance analysis tools for Linux kernel.
 
-%package -n     python3-perf
+%package        python3-perf
```
vinceaperri, line 139: I think naming this as python3-perf-hci would be more consistent with out kernel variant naming scheme than kernel-hci-python3-perf. That way, kernel would be to python3-perf as kernel-hci would be to python3-perf-hci.

Further, there's a good reason why kernel's python3-perf and bpftool subpackages aren't normally prefixed with "kernel-". In the case of python3-perf, it's because it's fundamentally a python3 module. Likewise, we shouldn't do this for our kernel variants.

PR #5139, opened 2023-03-23
Add the mstflint Kernel Module

ellie-di:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 
This PR adds the mstflint kernel module for secure boot. AODS had a requirement to be able to use the mstflint tool (with its signed kernel module) for debugging and diagnostic purposes.

**NO**

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=331631&view=artifacts&pathAsName=false&type=publishedArtifacts
- Tested the signed artifact from above pipeline and was able to install the module (insmod) when secure boot is enabled.


SPECS/mstflint/mstflint.spec
```diff
@@ -23,15 +23,35 @@ BuildRequires:  libxml2-devel
...
+Requires:       kernel
+
+# set package name
+%{!?_name: %global _name mstflint}
+
+%global kver %(/bin/rpm -q --queryformat '%{RPMTAG_VERSION}-%{RPMTAG_RELEASE}' $(/bin/rpm -q --whatprovides kernel-devel))
+%global ksrc %{_libdir}/modules/%{kver}/build
+%global debug_package %{nil}
+%global kernel_source() %{ksrc}
+%global kernel_release() %{kver}
```
PawelWMS, line 38: Why create these 2 functions? Couldn't we just use the 2 macros defined above directly?

SPECS/mstflint/mstflint.spec
```diff
@@ -40,13 +60,42 @@ find . -type f -iname '*.cpp' -exec chmod a-x '{}' ';'
...
 
+#Compile mstflint driver
+export EXTRA_CFLAGS='-DVERSION=\"%version\"'
+cd kernel
+set -- *
+mkdir source
+mv "$@" source/
+mkdir obj
+for flavor in %flavors_to_build; do
+  export K_BUILD=%{kernel_source $flavor}
```
PawelWMS, line 71: This looks fishy. We're passing an argument to a function, which doesn't do anything with it.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,48 @@
...
+%global install_mod_dir extra/%{_name}
+%define mstflint_name %(value="%{mstflint_unsigned_name}-%{version}-%{release}"; echo "${value//[^a-zA-Z0-9_]/_}")
+%define mstflint_module_name mstflint_access.ko
+%define mstflint_module_path %{install_mod_dir}/%{mstflint_module_name}
+
+%install
+export INSTALL_MOD_PATH=$RPM_BUILD_ROOT
+mkdir -p %{install_mod_dir}
+cd kernel
+for flavor in %{flavors_to_build}; do
```
PawelWMS, line 32: If there's only one flavour for now, I'd suggest getting rid of the `flavors_to_build` macro along with the `for` loops for the sake of simplicity.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,48 @@
...
+  export KSRC=%{kernel_source $flavor}
+  export KVERSION=%{kernel_release $KSRC}
```
PawelWMS, line 34: Do we need these environmental variables? Could we use the macros directly instead?

SPECS/mstflint/mstflint.spec
```diff
@@ -23,15 +23,35 @@ BuildRequires:  libxml2-devel
...
+BuildRequires:  kmod
 Requires:       python3
+Requires:       kernel
+
+# set package name
+%{!?_name: %global _name mstflint}
+
+%global kver %(/bin/rpm -q --queryformat '%{RPMTAG_VERSION}-%{RPMTAG_RELEASE}' $(/bin/rpm -q --whatprovides kernel-devel))
+%global ksrc %{_libdir}/modules/%{kver}/build
+%global debug_package %{nil}
```
PawelWMS, line 36: Removing debuginfo packages is a global setting for the whole spec, so it will also remove any debug packages for the other subpackages. This is a regression. Why do we want to remove the debuginfo packages?

SPECS/mstflint/mstflint.spec
```diff
@@ -23,15 +23,35 @@ BuildRequires:  libxml2-devel
 BuildRequires:  make
 BuildRequires:  openssl-devel
 BuildRequires:  zlib-devel
+BuildRequires:  kernel-devel
```
PawelWMS, line 26: We're introducing an out-of-tree kernel module in this spec. This is a serious maintenance commitment to keep it up-to-date with the changing internal kernel APIs.

1. Do we have a team, who is going to make sure this module continues to work?
2. Please limit the `BuildRequires` and `Requires` to the lowest and highest kernel versions we currently know this module works with. Otherwise, we will encounter build and/or run-time issues in the future as the kernel internal APIs change and the module falls behind.

PawelWMS (changes requested):
Maing concern: maintenance questions regarding an introduction of an out-of-tree kernel module.

Also a few clean-up remarks. Not fully done yet, will continue with the rest once we tackle the OOT question.

SPECS/mstflint/mstflint.spec
ellie-di: 1- Yes, Chaitanya's team. I'll be adding a test for this module in next iteration.
2- Sure.

SPECS/mstflint/mstflint.spec
ellie-di: @PawelWMS  I've tried to follow the original mstflint kernel spec as much as possible https://github.com/Mellanox/mstflint/blob/master/kernel/mstflint_kernel.spec

SPECS/mstflint/mstflint.spec
ellie-di: I looked into the way that the module is including the kernel headers, it seems the kernel header file inclusions has been handled for the older versions, so there is no point for checking the required lowest kernel version. And for the highest version check, the cost of maintenance would be more if we add the current version as the highest required version; since there is less chance that those APIs get changed on next kernel version, but that condition will definitely break the build or runtime execution. Then there is no point on limiting the highest kernel version as well. 
As I mentioned, I'm going to add a test for this module, and I think that's a better solution to ensure the module compatibility with kernel version.

ogalbxela, line 26: The list of BuildRequires is suppossed to be sorted alphabetically to make it easier to maintain the code. The newly added lines violate this convention, so please rearrange them accordingly.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,36 @@
...
+%global install_mod_dir extra/%{_name}
+
+%install
+export INSTALL_MOD_PATH=$RPM_BUILD_ROOT
+mkdir -p %{install_mod_dir}
+cd kernel
+install -d $INSTALL_MOD_PATH/lib/modules/%{kver}/%{install_mod_dir}
+cp $PWD/obj/mstflint_access.ko $INSTALL_MOD_PATH/lib/modules/%{kver}/%{install_mod_dir}
+cd -
+
```
ogalbxela, line 27: The usual way to handle kernel modules in Linux is through Dynamic Kernel Module Support (DKMS), and it is generally not recommended to install kernel modules manually.

PS: As for the Microsoft workflow and its relation to this project, I am not familiar with them and have no connections to this project. Please consider me as a random individual who has been asked to review a change in a project that I am not familiar with. However, I do find the installation of a kernel module object to be suspicious.

SPECS/mstflint/mstflint.spec
```diff
@@ -23,15 +23,31 @@ BuildRequires:  libxml2-devel
...
+%global install_mod_dir extra/%{_name}
 
 %description
 This package contains firmware update tool, vpd dump and register dump tools
 for network adapters based on Mellanox Technologies chips.
 
+%package kernel
+Summary:        mstflint Kernel Modules
+Group:          System Environment/Kernel
+Requires:       kernel-headers >= 5.15.0
```
ogalbxela, line 44: What is the reason for restricting the kernel version?
As far as I know, mstflint supports kernel versions 4.x as well. Is it something related to Microsoft?

SPECS/mstflint/mstflint.spec
```diff
@@ -23,15 +23,35 @@ BuildRequires:  libxml2-devel
 BuildRequires:  make
 BuildRequires:  openssl-devel
 BuildRequires:  zlib-devel
+BuildRequires:  kernel-devel
+BuildRequires:  kmod
 Requires:       python3
+Requires:       kernel
```
ogalbxela, line 29: same here, alphabetical order please

SPECS/mstflint/mstflint.spec
```diff
@@ -40,13 +56,35 @@ find . -type f -iname '*.cpp' -exec chmod a-x '{}' ';'
...
+mkdir source
+mv "$@" source/
+mkdir obj
+
+rm -rf obj
+cp -a source obj
+cd $PWD/obj
+make KPVER=%{kver}
+cd -
+cd -
```
ogalbxela, line 72: what is the second "cd -" for?

ogalbxela: why not DKMS?

chaitanyalala:
DKMS can build modules but not sign. DKMS is not a viable approach for
mariner.

ogalbxela:
> 
Could you please clarify if the mstflint-signed RPM package with the pre-compiled kernel module is intended to be shipped with the kernel it was compiled against or if it can be shipped separately?
It's important to ensure that the pre-compiled module is compatible with the target systems.

ellie-di:
> > 
> 
> Could you please tell me where in this PR the kernel module is being signed? I assume that it should happen after the compilation step, right? Perhaps I just have overlooked it. Please also clarify if the mstflint-signed RPM package with the pre-compiled kernel module is intended to be shipped with the kernel it was compiled against or if it can be shipped separately? It's important to ensure that the pre-compiled module is compatible with the target systems.

@ogalbxela Thanks for reviewing this PR. The signing process is in CBL-Mariner pipeline repo which is not public. mstflint tool and its kernel module will be published per each CBL-Mariner release in https://packages.microsoft.com/.

SPECS/mstflint/mstflint.spec
ellie-di: I think I agree with you. The reason for that limit is that Mariner 2.0 uses kernel 5.15.x and we wanted to limit the support scope of this OOT module.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
ellie-di: Thanks for your comments. That's a fair point, however as Chaitanya mentioned it's not supported by Mariner ecosystem.

SPECS/mstflint/mstflint.spec
```diff
@@ -15,6 +15,8 @@ BuildRequires:  automake
 BuildRequires:  boost-devel
 BuildRequires:  gcc
 BuildRequires:  gcc-c++
+BuildRequires:  kernel-devel
```
PawelWMS, line 18: Since we're using this to build a kernel module, let's be explicit about the version we're using with `BuildRequires: kernel-devel = <version>-<release>`. One idea might be to follow what we do with livepatches, which also deal with kernel modules. At the top of their specs we define:
```
%define kernel_version_release 5.15.87.1-1.cm2
```

and then use that macro in `BuildRequires: kernel-devel = %{kernel_version_release}`.

SPECS/mstflint/mstflint.spec
```diff
@@ -23,15 +25,29 @@ BuildRequires:  libxml2-devel
 BuildRequires:  make
 BuildRequires:  openssl-devel
 BuildRequires:  zlib-devel
+Requires:       kernel
```
PawelWMS, line 28: The `Requires` is actually a per-package information unlike `BuildRequires`, which give you build-time dependencies for the whole spec. What that means is that now you've defined a run-time requirement for `msft` package, while we're interested in the `msft-kernel` one. You'd need to move the `Requires` under `%package kernel`.

Btw. the published `msft-kernel` package will end up being built from the `msft-signed.spec` and not from this one, so the run-time requirements are not influencing anyone in this special case. Keeping the `Requires` in both specs, however, is probably going to be less confusing, so I'd still leave them.

And please also specify the exact kernel version the modules needs. Same idea as with the `BuildRequires: kernel-devel` comment above.

SPECS/mstflint/mstflint.spec
```diff
@@ -23,15 +25,29 @@ BuildRequires:  libxml2-devel
...
+%global install_mod_dir extra/%{_name}
+
 %description
 This package contains firmware update tool, vpd dump and register dump tools
 for network adapters based on Mellanox Technologies chips.
 
+%package kernel
+Summary:        mstflint Kernel Modules
+Group:          System Environment/Kernel
+Requires:       kernel-headers >= 5.15.0
```
PawelWMS, line 44: Is this true? Does a built kernel module need any headers? And if it does, please specify the exact version with `=` instead of `>=`.

SPECS/mstflint/mstflint.spec
```diff
@@ -40,13 +56,33 @@ find . -type f -iname '*.cpp' -exec chmod a-x '{}' ';'
 %configure --enable-fw-mgr --enable-adb-generic-tools
 %make_build
 
+# Compile mstflint driver
+export EXTRA_CFLAGS='-DVERSION=\"%version\"'
+cd kernel
+set -- *
+mkdir source
+mv "$@" source/
```
PawelWMS, line 64: Why do we need to overwrite the positional parameters? Wouldn't this work as well:
```suggestion
mkdir source
mv * source/
```

SPECS/mstflint/mstflint.spec
```diff
@@ -40,13 +56,33 @@ find . -type f -iname '*.cpp' -exec chmod a-x '{}' ';'
...
+mkdir obj
+
+rm -rf obj
```
PawelWMS, line 67: Why create a directory and immediately remove it?

SPECS/mstflint/mstflint.spec
```diff
@@ -40,13 +56,33 @@ find . -type f -iname '*.cpp' -exec chmod a-x '{}' ';'
...
+cp -a source obj
+cd $PWD/obj
+make KPVER=%{kver}
```
PawelWMS, line 70: I'm somewhat confused by this build procedure. It seems we're first copying the contents of `kernel` into the `source` subdirectory and then again copy that to `obj` (which may have been deleted a step earlier) just to call `make KPVER=%{kver}` from inside `obj`. I think I am missing something here. Couldn't we just call `EXTRA_CFLAGS='-DVERSION=\"%version\"' make -C kernel KPVER=%{kver}` and skip all the extra steps?

```suggestion
# Compile mstflint driver
EXTRA_CFLAGS='-DVERSION=\"%version\"' make -C kernel KPVER=%{kver}
```

SPECS/mstflint/mstflint.spec
```diff
@@ -40,13 +56,33 @@ find . -type f -iname '*.cpp' -exec chmod a-x '{}' ';'
...
+
 %install
 %make_install
 # Remove the devel files that we don't ship
 rm -fr %{buildroot}%{_includedir}
 find %{buildroot} -type f -name "*.la" -delete -print
 find %{buildroot} -type f -name '*.a' -delete
 
+export INSTALL_MOD_PATH=$RPM_BUILD_ROOT
```
PawelWMS, line 79: It looks like we might need to lint this and the other specs as we don't want to use `RPM_BUILD_ROOT`. [Here's a short doc](https://dev.azure.com/mariner-org/mariner/_wiki/wikis/mariner.wiki/1126/Spec-linter) about configuring the one we use. The linter might also clean-up other details.

**NOTE**: the linter is not perfect, so I suggest running it after committing all previous changes and then double-checking the modifications it has made.

SPECS/mstflint/mstflint.spec
```diff
@@ -40,13 +56,33 @@ find . -type f -iname '*.cpp' -exec chmod a-x '{}' ';'
...
+mkdir -p %{install_mod_dir}
```
PawelWMS, line 80: This folder seems unused.

SPECS/mstflint/mstflint.spec
```diff
@@ -57,7 +93,17 @@ find %{buildroot} -type f -name '*.a' -delete
 %{_datadir}/mstflint
 %{_mandir}/man1/*
 
+%post kernel
+depmod %{kver} -a
+
+%files kernel
+%defattr(-,root,root,-)
+/lib/modules/%{kver}/%{install_mod_dir}/
```
PawelWMS, line 101: It looks like we're repeating `/lib/modules/%{kver}/%{install_mod_dir}/` multiple times in the spec. Let's move it to a macro to avoid repetitions and accidental issues related to typos or forgetting to change one of the repeated paths, if ever needed in the future. I think the `install_mod_dir` macro is used only in this path, so we might replace it as well.

SPECS/mstflint/mstflint.spec
```diff
@@ -23,15 +25,29 @@ BuildRequires:  libxml2-devel
 BuildRequires:  make
 BuildRequires:  openssl-devel
 BuildRequires:  zlib-devel
+Requires:       kernel
 Requires:       python3
 
+# set package name
+%{!?_name: %global _name mstflint}
```
PawelWMS, line 32: What is this for?

SPECS/mstflint/mstflint.spec
```diff
@@ -23,15 +25,29 @@ BuildRequires:  libxml2-devel
...
+
+%global kver %(/bin/rpm -q --queryformat '%{RPMTAG_VERSION}-%{RPMTAG_RELEASE}' $(/bin/rpm -q --whatprovides kernel-devel))
```
PawelWMS, line 34: Let's just use the hard-coded version of `kernel-devel` we're building this spec with (see: comments about `BuildRequires`) and reduce the complexity.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,35 @@
...
+Summary:        Mellanox firmware burning tool
+Name:           %{mstflint_unsigned_name}-signed
+Version:        4.21.0
+Release:        4%{?dist}
+License:        Dual BSD/GPL
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+Group:          System Environment/Kernel
+URL:            https://github.com/Mellanox/%{name}
+Source0:        https://github.com/Mellanox/%{name}/releases/download/v%{version}-1/%{name}-%{version}-1.tar.gz
```
PawelWMS, line 12: In this case the source is only going to be the signed kernel module built from `mstflint.spec`.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,35 @@
+%define mstflint_unsigned_name mstflint-kernel
+
+Summary:        Mellanox firmware burning tool
+Name:           %{mstflint_unsigned_name}-signed
+Version:        4.21.0
+Release:        4%{?dist}
```
PawelWMS, line 6: This reminds me: we probably want to "entangle" this spec with `mstflint.spec`. The idea is to make it harder to miss propagating the updates made in `mstflint.spec` to this spec. If anything related to `mstflint-kernel` ever changes in `mstflint.spec`, then we'll need to bump the `Release` number in this spec, so that we produce a package with a new version and the changes actually get published. The "entangled" specs get validated by our PR checks, so we block them and avoid these mistakes.

Updating the check should only require adding another `frozenset` inside `toolkit/scripts/check_entangled_specs.py`. Similar to the `kernel-hci.spec` and `kernel-hci-signed.spec` pair.

**NOTE**: any changes inside `mstflint.spec` will then require a change in this one as well. Even the ones, which don't affect the `*-kernel` subpackage. One way to avoid that would be to split the `mstflint-kernel` package into a completely separate spec file. I'll need to think a bit more about the potential pros and cons, though.

SPECS/mstflint/mstflint.spec
ellie-di: Following our discussion, we are not going for this option.

SPECS/mstflint/mstflint.spec
ellie-di: yes, we are limiting this module to kernel 5.15.x

SPECS/mstflint/mstflint.spec
ellie-di: Removed.

SPECS/mstflint/mstflint.spec
ellie-di: Thanks for the suggestion. Applied all except the -C option, as mstflint kernel Makefile is dependent on {pwd} and it doesn't work if it's not issued inside the directory.

SPECS/mstflint/mstflint.spec
ellie-di: Removed.

SPECS/mstflint/mstflint.spec
ellie-di: Removed.

SPECS/mstflint/mstflint.spec
ellie-di: Removed.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
ellie-di: I don't think there would be frequent changes in the spec and both are working on the same source tarball. If there is any value in separating them, we can do it later.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,32 @@
...
+License:        Dual BSD/GPL
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+Group:          System Environment/Kernel
+URL:            https://github.com/Mellanox/%{name}
+Source0:        https://github.com/Mellanox/%{name}/releases/download/v%{version}-1/%{name}-%{version}-1.tar.gz
+Requires:       kernel >= 5.15.0 and kernel < 5.16.0
+Requires:       kernel-headers >= 5.15.0 and kernel-headers < 5.16.0
```
PawelWMS, line 15: This is a limitation of our tools, where I believe they don't handle the `and`, so we use this instead:
```suggestion
Requires:       kernel >= 5.15.0
Requires:       kernel < 5.16.0
Requires:       kernel-headers >= 5.15.0
Requires:       kernel-headers < 5.16.0
```

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,32 @@
+%define mstflint_unsigned_name mstflint-kernel
+%global kver %(/bin/rpm -q --queryformat '%{RPMTAG_VERSION}-%{RPMTAG_RELEASE}' $(/bin/rpm -q --whatprovides kernel-devel))
+%global install_mod_dir lib/modules/%{kver}/extra/%{name}
```
PawelWMS, line 3: The `%{name}` macro is the value of the `Name` tag below. In this case it would translate to "mstflint-kernel-signed". I see you're storing the actual package name in `mstflint_unsigned_name`, so let's use that:
```suggestion
%global install_mod_dir lib/modules/%{kver}/extra/%{mstflint_unsigned_name}
```

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,32 @@
+%define mstflint_unsigned_name mstflint-kernel
+%global kver %(/bin/rpm -q --queryformat '%{RPMTAG_VERSION}-%{RPMTAG_RELEASE}' $(/bin/rpm -q --whatprovides kernel-devel))
```
PawelWMS, line 2: This might be problematic, because the specs are parsed twice in our tooling and for this package in both cases the `kernel-devel` package is not going to be present in the build environment, so this will probably fail to do, what we'd like it to do. Let's chat offline, because I think I need to learn more about the expected installation directory for kernel modules.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,32 @@
...
+URL:            https://github.com/Mellanox/%{name}
+Source0:        https://github.com/Mellanox/%{name}/releases/download/v%{version}-1/%{name}-%{version}-1.tar.gz
+Requires:       kernel >= 5.15.0 and kernel < 5.16.0
+Requires:       kernel-headers >= 5.15.0 and kernel-headers < 5.16.0
+
+%install
+export INSTALL_MOD_PATH=%{buildroot}
+cd kernel
+install -d $INSTALL_MOD_PATH/%{install_mod_dir}
+cp $PWD/mstflint_access.ko $INSTALL_MOD_PATH/%{install_mod_dir}
```
PawelWMS, line 21: Just a small clean-up and aligning more to what we tend to gravitate towards in other specs:
```suggestion
install -dm 755 %{buildroot}%{install_mod_dir}
install -m 644 %{SOURCE0} %{buildroot}%{install_mod_dir}
```

The missing `/` in `%{buildroot}%{install_mod_dir}` is by design if we take in the `install_mod_dir` from a separate comment at the top of this spec.

The "signed" specs use pre-built sources as input, thus the removal of `cd kernel` and going straight to `%{SOURCE0}` - that macro holds the absolute path to the first source, which in this case is probably going to be that pre-built and signed kernel module.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
PawelWMS: Got it, that sounds good to me.

We'd still want to entangle the specs. The "signed" and regular specs split is just a way to produce packages, which need some of their components to be signed, but logically they are viewed as a single spec, so their versions and releases should be updated in lockstep.

SPECS/mstflint/mstflint.spec
```diff
@@ -1,7 +1,9 @@
...
+%global install_mod_dir lib/modules/%{kver}/extra/%{name}
```
PawelWMS, line 2: Same small update as in the "signed" spec:
```suggestion
%global install_mod_dir %{_libdir}/modules/%{kver}/extra/%{name}-kernel
```

SPECS/mstflint/mstflint.spec
```diff
@@ -29,9 +33,17 @@ Requires:       python3
 This package contains firmware update tool, vpd dump and register dump tools
 for network adapters based on Mellanox Technologies chips.
 
+%package kernel
+Summary:        mstflint Kernel Modules
+Group:          System Environment/Kernel
+Requires:       kernel >= 5.15.0 and kernel < 5.16.0
+Requires:       kernel-headers >= 5.15.0 and kernel-headers < 5.16.0
```
PawelWMS, line 40: Similar issue with our tooling not supporting `and`:
```suggestion
Requires:       kernel >= 5.15.0
Requires:       kernel < 5.16.0
Requires:       kernel-headers >= 5.15.0
Requires:       kernel-headers < 5.16.0
```

SPECS/mstflint/mstflint.spec
```diff
@@ -15,6 +17,8 @@ BuildRequires:  automake
 BuildRequires:  boost-devel
 BuildRequires:  gcc
 BuildRequires:  gcc-c++
+BuildRequires:  kernel-devel
```
PawelWMS, line 20: Let's also include the version constraints for `kernel-devel` for the sake of consistency, being explicit, and the fact, that the BRs are what actually dictates the run-time requirements due to how RPM figures out implicit run-time dependencies.
```suggestion
BuildRequires:  kernel-devel >= 5.15.0
BuildRequires:  kernel-devel < 5.16.0
```

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,32 @@
...
+Name:           %{mstflint_unsigned_name}-signed
+Version:        4.21.0
+Release:        4%{?dist}
+License:        Dual BSD/GPL
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+Group:          System Environment/Kernel
+URL:            https://github.com/Mellanox/%{name}
+Source0:        https://github.com/Mellanox/%{name}/releases/download/v%{version}-1/%{name}-%{version}-1.tar.gz
+Requires:       kernel >= 5.15.0 and kernel < 5.16.0
```
PawelWMS, line 14: Could we also convert the upper and lower bound for kernel version into 2 macros and use them instead? This way there's a lesser chance of making any mistakes during future updates and we avoid repetition.

SPECS/mstflint/mstflint.spec
PawelWMS, line 40: Also, a similar request to use macros for upper and lower bounds for the versions of the kernel packages.

SPECS/mstflint/mstflint.spec
```diff
@@ -15,6 +17,8 @@ BuildRequires:  automake
...
+BuildRequires:  kmod
```
PawelWMS, line 21: Double-checking: is `kmod` needed to build the module? Or is it only needed during run-time to load it?

SPECS/mstflint/mstflint.spec
```diff
@@ -40,13 +52,22 @@ find . -type f -iname '*.cpp' -exec chmod a-x '{}' ';'
...
 %make_install
 # Remove the devel files that we don't ship
 rm -fr %{buildroot}%{_includedir}
 find %{buildroot} -type f -name "*.la" -delete -print
 find %{buildroot} -type f -name '*.a' -delete
 
+export INSTALL_MOD_PATH=%{buildroot}
+cd kernel
+install -d $INSTALL_MOD_PATH/%{install_mod_dir}
+cp $PWD/mstflint_access.ko $INSTALL_MOD_PATH/%{install_mod_dir}
```
PawelWMS, line 69: Similar clean-up as in the "signed" spec:
```suggestion
install -dm 755 %{buildroot}%{install_mod_dir}
install -m 644 kernel/mstflint_access.ko %{buildroot}%{install_mod_dir}
```

SPECS/mstflint/mstflint.spec
ellie-di: @PawelWMS  If it doesn't support why didn't throw any error? I expect RPM version > 4.13 support it.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
PawelWMS: Discussed offline, seems like I was wrong and it's working fine. I'll double-check what the buddy builds return but no need to worry about it for now.

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
```diff
@@ -0,0 +1,32 @@
...
+Requires:       kernel-headers >= 5.15.0 and kernel-headers < 5.16.0
+
+%install
+export INSTALL_MOD_PATH=%{buildroot}
+cd kernel
+install -d $INSTALL_MOD_PATH/%{install_mod_dir}
+cp $PWD/mstflint_access.ko $INSTALL_MOD_PATH/%{install_mod_dir}
+
+
```
PawelWMS, line 23: Following our offline discussion so we don't forget:
```suggestion
%post -n %{mstflint_unsigned_name}
depmod %{kver} -a

%postun -n %{mstflint_unsigned_name}
depmod %{kver} -a
```

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
PawelWMS, line 15: And to make the `%post*` sections work I'm guessing we'll also need `Requires(post)` and `Requires(postun)` on `kmod`.

SPECS/mstflint/mstflint.spec
PawelWMS: It's not about RPM, it's about our own build tooling that parses the specs and builds a dependency diagram. RPM definitely supports that. It's possible that it supports some of these keywords, but to be on the safe side, I wanted to go with this option.

PawelWMS: Yeah, the tooling is too dumb and fails with issues like these:
```
(...)
INFO[0271] ---------------------------                  
INFO[0271] --------- Summary ---------                  
INFO[0271] ---------------------------                  
INFO[0271] Number of built SRPMs:             1         
INFO[0271] Number of prebuilt SRPMs:          0         
INFO[0271] Number of failed SRPMs:            0         
INFO[0271] Number of blocked SRPMs:           0         
INFO[0271] Number of unresolved dependencies: 1 
INFO[0271] Built SRPMs:                                 
INFO[0271] --> mstflint-4.21.0-4.cm2.src.rpm            
INFO[0271] Unresolved dependencies:                     
INFO[0271] --> and:C:''V:'',C2:''V2:'
...

For now we'd need to stick to separate lines.

SPECS/mstflint/mstflint.spec
ellie-di: Moved it to "Requires:".

SPECS-SIGNED/mstflint-driver-signed/mstflint-signed.spec
ellie-di: Done! Thanks!

SPECS/mstflint/mstflint.spec
PawelWMS: Could we do still do this clean-up? I see we already have that in the "signed" spec.

ellie-di: Done!

PawelWMS (approved):

PR #5176, opened 2023-03-29
Don't make go modules owned by root

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

When we invoke our build commands it is usually of the form `sudo make ...` since we need permissions to create mount points, overlays, chroots, etc. 

`go build` will automatically pull any missing go modules during the build. This has a side-effect of polluting the `$GOPATH` (default is `~/go`) with modules owned by the root user when we build the go tools (since we are invoking `sudo make ...`). This can be fixed via an occasional invocation of `sudo chown -R $USER:$USER ~/go` but it would be better if we just never broke the files in the first place.

We can query the variable `$SUDO_USER` to see if the command is being invoked via a call to `sudo`, it will be empty otherwise. If it is set, we can use `sudo -u $SUDO_USER` to temporarily switch back to the calling user and run `go get ./...` to explicitly download all our dependencies. 

We allow this command to fail and fall back to the previous behavior. 

###### Change Log  

- Try to avoid pulling go modules as the root user so it doesn't mess up the user's dev environment.

**NO**

###### Test Methodology

- Local builds

### DIFF
--- a/toolkit/scripts/tools.mk
+++ b/toolkit/scripts/tools.mk
@@ -96,7 +96,7 @@ $(BUILD_DIR)/tools/internal.test_coverage: $(go_internal_files) $(go_imagegen_fi
 
 # Downloads all the go dependencies without using sudo, so we don't break other go use cases for the user.
 # We can check if $SUDO_USER is set (the user who invoked sudo), and if so, use that user to run go get via sudo -u.
-# We allow the command to fail with || true, since we don't want to fail the build if the user has already
+# We allow the command to fail with || echo ..., since we don't want to fail the build if the user has already
 # downloaded the dependencies as root. The go build command will download the dependencies if they are missing (but as root).
 $(STATUS_FLAGS_DIR)/got_go_deps.flag: $(go_common_files)
 	@cd $(TOOLS_DIR)/ && \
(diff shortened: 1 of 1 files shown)


nicogbg (approved):

mfrw (approved):

toolkit/scripts/tools.mk
```diff
@@ -90,10 +90,25 @@ $(TOOL_BINS_DIR)/%: $(go_common_files)
...
 # Runs tests for common components
-$(BUILD_DIR)/tools/internal.test_coverage: $(go_internal_files) $(go_imagegen_files)
+$(BUILD_DIR)/tools/internal.test_coverage: $(go_internal_files) $(go_imagegen_files) $(STATUS_FLAGS_DIR)/got_go_deps.flag
 	cd $(TOOLS_DIR)/$* && \
 		go test -covermode=atomic -coverprofile=$@ ./...
 
+# Downloads all the go dependencies without using sudo, so we don't break other go use cases for the user.
+# We can check if $SUDO_USER is set (the user who invoked sudo), and if so, use that user to run go get via sudo -u.
+# We allow the command to fail with || echo ..., since we don't want to fail the build if the user has already
+# downloaded the dependencies as root. The go build command will download the dependencies if they are missing (but as root).
```
rlmenge, line 100: Nice comments!

rlmenge, line 100: Nice description!

rlmenge (approved):
LGTM

PR #4026, opened 2022-10-14
1.0 toolkit: skip ARCHIVE_TOOL on extraction

anphel31:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 

What does the PR accomplish, why was it needed?
Change toolkit scripts to not use ARCHIVE_TOOL when extracting tarballs (rpms.tar.gz, srpms.tar.gz, toolchain_built_rpms_all.tar.gz, toolchain_built_srpms_all.tar.gz). The tar command is smart enough to figure out which decompression library to use (see [gnu tar](https://www.gnu.org/software/tar/manual/tar.html#gzip) "Reading compressed archive is even simpler: you don’t need to specify any additional options as GNU tar recognizes its format automatically.". Removing the explicit ARCHIVE_TOOL makes the scripts more flexible, and they can handle cases where these tarballs might not be compressed.

Relevant example below. In this case the "-z" flag is equivalent to "-I $(ARCHIVE_TOOL)"
```
tar -xf compressed_rpms.tar.gz
    OK
tar -xzf compressed_rpms.tar.gz
    OK
tar -xf uncompressed_rpms.tar.gz
[...]

PR #3683, opened 2022-09-03
Skip compression on rpm/srpm archives

anphel31:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 

What does the PR accomplish, why was it needed?
We are currently compressing and decompressing RPM/SRPM files (which are already compressed) into archives.
These package tarballs (rpms.tar.gz, srpms.tar.gz, toolchain_built_rpms_all.tar.gz, toolchain_built_srpms_all.tar.gz) are currently being compressed/decompressed by gzip or pigz. Since the underlying rpm files are themselves already compressed, the archive is roughly the same size with or without compression. Skipping the compression step saves a significant amount of time.

The size differences in a recent main branch build with and without compression are shown below.
```
FILE                                  SIZE COMPRESSED              SIZE UNCOMPRESSED
toolchain_built_rpms_all.tar.gz       1601 MB                      1634 MB
toolchain_built_srpms_all.tar.gz      593 MB                       594 MB
rpms.tar.gz                           13175 MB                     13243 MB
srpms.tar.gz                          7893 MB                      8050 MB
```

[...]

PR #5199, opened 2023-03-31
Add QUICKBUILD flags for fast auto build

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

While it is fairly straightforward to build efficiently on the stable branches it is very challenging to find the correct combination of flags for unstable branches. Lets add a set of `QUICKBUILD_*` flags that auto configure the build for various use-cases.

`QUICKBUILD=y` will set:
 * `QUICKBUILD_TOOLCHAIN=y`
 * `QUICKBUILD_PACKAGES=y`

`QUICKBUILD_TOOLCHAIN=y` will set:
 * `REBUILD_TOOLCHAIN= y`
 * `INCREMENTAL_TOOLCHAIN= y`
 * `ALLOW_TOOLCHAIN_DOWNLOAD_FAIL= y`

`QUICKBUILD_PACKAGES=y` will set:
 * `DELTA_BUILD= y`
 * `USE_CCACHE= y`

Also update the raw toolchain build process to understand `INCREMENTAL_TOOLCHAIN`. If it is set, let's try and use a cached copy of the raw toolchain which is based on the contents of the source hashes and scripts.

Documentation will follow in a subsequent PR.

###### Change Log  

- Add `QUICKBUILD*` flags to auto set incremental fast build flags
- Enlighten `create_toolchain_in_container.sh` about `INCREMENTAL_TOOLCHAIN`

**YES**

###### Test Methodology
- Local build
- Needs pipeline testing


dmcilvaney:
Next step is to redo the toolchain download logic to be unified, and to completely skip the raw + final toolchain scripts if all the RPMs can just be downloaded.

dmcilvaney:
Note: The documentation for this is in https://github.com/microsoft/CBL-Mariner/pull/5218

toolkit/Makefile
```diff
@@ -19,16 +21,29 @@ PACKAGE_REBUILD_LIST    ?=
...
+
+# We generally want to use the DELTA_BUILD flag to determine whether to use the incremental toolchain, but to preserve
+# backwards compatibility, we'll still use the INCREMENTAL_TOOLCHAIN flag if it's set.
+DELTA_BUILD                     ?= n
+ifeq ($(origin DELTA_BUILD), undefined)
+# We didn't set DELTA_BUILD, so use the previous default behavior of INCREMENTAL_TOOLCHAIN for backwards compatibility.
 INCREMENTAL_TOOLCHAIN           ?= n
+else
+# We set DELTA_BUILD, so inherit value for INCREMENTAL_TOOLCHAIN.
+INCREMENTAL_TOOLCHAIN           ?= $(DELTA_BUILD)
```
jslobodzian, line 36: Thank you

jslobodzian: Any chance we can echo a message that "INCREMENTAL_TOOLCHAIN" is deprecated?

So we can yank it out in the future.  No reason to have both

toolkit/quickrebuild_autoconfig.mk
```diff
@@ -0,0 +1,55 @@
...
+# Licensed under the MIT License.
+
+# Contains:
+#	- build optimization flag auto configuration
+
+######## QUICKREBUILD AUTO CONFIGURE ########
+
+# The QUICKREBUILD* flags are special flags that will try to build the toolchain and packages as quickly as possible. They will
+# automatically set REBUILD_TOOLS, USE_CCACHE, REBUILD_TOOLCHAIN, DELTA_BUILD, INCREMENTAL_TOOLCHAIN, and ALLOW_TOOLCHAIN_DOWNLOAD_FAIL to 'y'.
+QUICKREBUILD           ?= n
```
PawelWMS, line 11: Nit-pick: could we be consistent with separating words in argument names, so in this case `QUICK_REBUILD` instead of `QUICKREBUILD`?

toolkit/quickrebuild_autoconfig.mk
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# Contains:
+#	- build optimization flag auto configuration
+
+######## QUICKREBUILD AUTO CONFIGURE ########
+
+# The QUICKREBUILD* flags are special flags that will try to build the toolchain and packages as quickly as possible. They will
+# automatically set REBUILD_TOOLS, USE_CCACHE, REBUILD_TOOLCHAIN, DELTA_BUILD, INCREMENTAL_TOOLCHAIN, and ALLOW_TOOLCHAIN_DOWNLOAD_FAIL to 'y'.
```
jslobodzian, line 10: mostly an FYI here. I think this change is fine, but the request I was receiveing from others  was to have some sort of configuration file to read some of these items.  So there were simple defaults.

I do think part of the issue has been that folks who are not sure how to use the toolchain pass too many arguments that already default to the right things.  So, maybe we don't need a configuration file anyway.

Still, I think we should look for an opportunity here to eliminate unneeded arguments.  How many of these are really required.

toolkit/Makefile
```diff
@@ -19,16 +21,29 @@ PACKAGE_REBUILD_LIST    ?=
...
+endif
```
PawelWMS, line 37: I think I'm missing a crucial piece of `make` magic. With `DELTA_BUILD ?= n` at the very top, in my local tests the `ifeq ($(origin DELTA_BUILD), undefined)` block was never executed. It was fine, if I moved the default `DELTA_BUILD` to somewhere after that check.

And what would fail with this approach:
```suggestion
DELTA_BUILD                     ?= n
INCREMENTAL_TOOLCHAIN           ?= $(DELTA_BUILD)
```

I think this still allows the user to set `INCREMENTAL_TOOLCHAIN` independently from `DELTA_BUILD` with the default being `n` for both.

toolkit/Makefile
```diff
@@ -19,16 +21,29 @@ PACKAGE_REBUILD_LIST    ?=
...
+
+REBUILD_TOOLCHAIN               ?= n
 ALLOW_TOOLCHAIN_DOWNLOAD_FAIL   ?= n
 UPDATE_TOOLCHAIN_LIST           ?= n
 REBUILD_PACKAGES                ?= y
 DOWNLOAD_SRPMS                  ?= n
 ALLOW_SRPM_DOWNLOAD_FAIL        ?= n
 REBUILD_TOOLS                   ?= n
 RUN_CHECK                       ?= n
-USE_CCACHE                      ?= n
```
christopherco, line 31: Where did this go?

toolkit/Makefile
PawelWMS: Btw. using `ifneq ($(origin DELTA_BUILD), command line)` seems to made `make` go into that block, if I didn't set `DELTA_BUILD` from the command line.

toolkit/Makefile
```diff
@@ -19,16 +21,29 @@ PACKAGE_REBUILD_LIST    ?=
...
+INCREMENTAL_TOOLCHAIN           ?= $(DELTA_BUILD)
+endif
+
+REBUILD_TOOLCHAIN               ?= n
 ALLOW_TOOLCHAIN_DOWNLOAD_FAIL   ?= n
 UPDATE_TOOLCHAIN_LIST           ?= n
 REBUILD_PACKAGES                ?= y
 DOWNLOAD_SRPMS                  ?= n
 ALLOW_SRPM_DOWNLOAD_FAIL        ?= n
 REBUILD_TOOLS                   ?= n
```
PawelWMS, line 45: Could also move this line to the new mk file? I think it's going to be less confusing to have the defaults set in one place.

PawelWMS, line 45: With the new mk file the `REBUILD_TOOLS` and a few other arguments now have two places where their defaults are set. I think this is making it even harder to understand what's going on. Could we keep the defaults in one place? In this case move these `?=` to the end of the new mk file, so that we have some values set for them, if they none of the quick build branches touched them?

toolkit/quickrebuild_autoconfig.mk
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) Microsoft Corporation.
```
christopherco, line 1: Please consider adding some very basic documentation on the new options in the docs folder

christopherco (changes requested):

toolkit/quickrebuild_autoconfig.mk
```diff
@@ -0,0 +1,55 @@
...
+ifneq ($(filter n,$(ALLOW_TOOLCHAIN_DOWNLOAD_FAIL)),)
+$(error QUICKREBUILD cannot be used with ALLOW_TOOLCHAIN_DOWNLOAD_FAIL explicitly set to 'n')
+endif
+
+# Don't care if REBUILD_TOOLS is set or not, doesn't matter to the quickbuild.
+
+REBUILD_TOOLCHAIN             = y
+INCREMENTAL_TOOLCHAIN         = y
+ALLOW_TOOLCHAIN_DOWNLOAD_FAIL = y
+REBUILD_TOOLS                 ?= y
```
PawelWMS, line 39: Why the `?=` in this case, if it's not set anywhere earlier?

toolkit/quickrebuild_autoconfig.mk
```diff
@@ -0,0 +1,55 @@
...
+$(error QUICKREBUILD cannot be used with REBUILD_TOOLCHAIN explicitly set to 'n')
+endif
+ifneq ($(filter n,$(INCREMENTAL_TOOLCHAIN)),)
+$(error QUICKREBUILD cannot be used with INCREMENTAL_TOOLCHAIN explicitly set to 'n')
+endif
+ifneq ($(filter n,$(ALLOW_TOOLCHAIN_DOWNLOAD_FAIL)),)
+$(error QUICKREBUILD cannot be used with ALLOW_TOOLCHAIN_DOWNLOAD_FAIL explicitly set to 'n')
+endif
+
+# Don't care if REBUILD_TOOLS is set or not, doesn't matter to the quickbuild.
```
PawelWMS, line 34: If it doesn't matter, why not remove setting it from this file completely?

toolkit/quickrebuild_autoconfig.mk
```diff
@@ -0,0 +1,55 @@
...
+ifneq ($(filter n,$(DELTA_BUILD)),)
+$(error DELTA_BUILD cannot be used with REBUILD_TOOLCHAIN explicitly  set to 'n')
+endif
+
+# Don't care if USE_CCACHE, or REBUILD_TOOLS are set or not, doesn't matter to the quickbuild.
+
+DELTA_BUILD    = y
+USE_CCACHE    ?= y
+REBUILD_TOOLS ?= y
+endif
```
PawelWMS, line 55: The `USE_CCACHE` argument seems tightly coupled with the idea of a quick build, so what do you think about this change and removing the default from `Makefile`, so we have it set in one place:
```suggestion
USE_CCACHE    = y
REBUILD_TOOLS ?= y
endif

USE_CCACHE    ?= n

```

PawelWMS, line 55: I'd also add a line for every argument that's set conditionally earlier in this file, so that we can move the defaults from `Makefile` and keep them in one place:

```suggestion
USE_CCACHE    ?= y
REBUILD_TOOLS ?= y
endif

REBUILD_TOOLCHAIN             ?= n
INCREMENTAL_TOOLCHAIN         ?= n
ALLOW_TOOLCHAIN_DOWNLOAD_FAIL ?= n
```

toolkit/scripts/toolchain/create_toolchain_in_container.sh
```diff
@@ -8,28 +8,38 @@ set -e
 MARINER_BUILD_DIR=$1
 MARINER_SPECS_DIR=$2
 MARINER_SOURCE_URL=$3
+INCREMENTAL_TOOLCHAIN=$4
+ARCHIVE_TOOL=$5
```
PawelWMS, line 12: I'd double-check that archive tool parameter with Andrew. I recall he's been removing these from some scripts a few months ago, but now I don't remember why and if it was here or elsewhere.

toolkit/scripts/toolchain/create_toolchain_in_container.sh
```diff
@@ -8,28 +8,38 @@ set -e
...
-# Binutils readonly patch
-cp -v $MARINER_SPECS_DIR/binutils/linker-script-readonly-keyword-support.patch ./container/linker-script-readonly-keyword-support.patch
-# RPM LD_FLAGS patch
-cp -v $MARINER_SPECS_DIR/rpm/define-RPM_LD_FLAGS.patch ./container/rpm-define-RPM-LD-FLAGS.patch
-
-# Create .bashrc file for lfs user in the container
-cat > ./container/.bashrc << EOF
+# Grab an identity for the raw toolchain components so we can avoid rebuilding it if it hasn't changed
+sha_component_tag=$(sha256sum ./container/toolchain-sha256sums ./container/toolchain_build_in_chroot.sh ./create_toolchain_in_container.sh | sha256sum | cut -d' ' -f1 )
+export sha_component_tag
```
PawelWMS, line 16: Why the `export`?

toolkit/scripts/toolchain/create_toolchain_in_container.sh
```diff
@@ -8,28 +8,38 @@ set -e
...
+    echo "No existing container with tag ${sha_component_tag}, building..."
+
+    ./toolchain_verify.sh $MARINER_BUILD_DIR
+
+    # Cleanup
+    docker images -a
+    docker ps -a
+    # docker system prune -f
+    # docker rmi $(docker images -a -q)
+    # docker rmi $(docker history marinertoolchain -q)
```
PawelWMS, line 29: Leftover debug commands?

toolkit/scripts/toolchain/create_toolchain_in_container.sh
```diff
@@ -8,28 +8,38 @@ set -e
...
-cp -v $MARINER_SPECS_DIR/coreutils/coreutils-fix-get-sys_getdents-aarch64.patch ./container
-# Binutils readonly patch
-cp -v $MARINER_SPECS_DIR/binutils/linker-script-readonly-keyword-support.patch ./container/linker-script-readonly-keyword-support.patch
-# RPM LD_FLAGS patch
-cp -v $MARINER_SPECS_DIR/rpm/define-RPM_LD_FLAGS.patch ./container/rpm-define-RPM-LD-FLAGS.patch
-
-# Create .bashrc file for lfs user in the container
-cat > ./container/.bashrc << EOF
+# Grab an identity for the raw toolchain components so we can avoid rebuilding it if it hasn't changed
+sha_component_tag=$(sha256sum ./container/toolchain-sha256sums ./container/toolchain_build_in_chroot.sh ./create_toolchain_in_container.sh | sha256sum | cut -d' ' -f1 )
```
PawelWMS, line 15: I think I have a better understanding of your idea with saving the Docker containers after looking at this now and I love saving us time by not re-running the toolchain build for every time we touch one of the files the `$(raw_toolchain)` target depends on. I think we don't need to save full Docker containers, though, and can thus save storage, time, and script complexity by making better use of `make` itself.

From this script it seems that all we care about is to eventually create the `toolchain_from_container.tar.gz` file. If that's correct, here's my idea:
1. Let's make this script save `toolchain_from_container_${sha_component_tag}.tar.gz` inside the `build` directory instead of saving full Docker containers.
2. Add an intermediate `toolchain_from_container_${sha_component_tag}.tar.gz` target inside `toolchain.mk`, which does exactly what `$(raw_toolchain)` does right now.
3. Have the `$(raw_toolchain)` target depend on the new `toolchain_from_container_${sha_component_tag}.tar.gz` in addition to what it already depends on but now the only thing it does, is copying `toolchain_from_container_${sha_component_tag}.tar.gz` over ``toolchain_from_container.tar.gz`.

Benefits:
- We don't messy up our scripts with tracking dependencies but have `make` do it for us.
- We only use up storage for the tarball we care about without any additional overhead of a full Docker container.
- We don't permanently create new Docker containers for each change touching the raw toolchain.

Let me know, if I'm getting this right and we can also chat offline to hammer out the details.

toolkit/scripts/toolchain/create_toolchain_in_container.sh
```diff
@@ -8,28 +8,38 @@ set -e
...
+export sha_component_tag
+
+# Check if we already have a committed container with the same sha_component_tag
+if [ "$INCREMENTAL_TOOLCHAIN" == "y" ] && [ -z "$(docker images -q marinertoolchain_populated:${sha_component_tag} 2>/dev/null)" ]; then
```
PawelWMS, line 19: >if [ "$INCREMENTAL_TOOLCHAIN" == "y" ] [](http://example.com/codeflow?start=0&length=38)

Wouldn't this break for clean repos, where we build the toolchain but without `INCREMENTAL_TOOLCHAIN=y`?

PawelWMS (dismissed):
:clock1:

toolkit/quickrebuild_autoconfig.mk
PawelWMS: I think Daniel's addressing that separately in #5218.

toolkit/Makefile
```diff
@@ -19,16 +21,29 @@ PACKAGE_REBUILD_LIST    ?=
 PACKAGE_IGNORE_LIST     ?=
 SRPM_PACK_LIST          ?=
 
-REBUILD_TOOLCHAIN               ?= n
+# Logic to auto configure build options for optimized builds
+# These will set the default values for: REBUILD_TOOLS, REBUILD_TOOLCHAIN, DELTA_BUILD, INCREMENTAL_TOOLCHAIN, and ALLOW_TOOLCHAIN_DOWNLOAD_FAIL
+include ./quickrebuild_autoconfig.mk
+
+# We generally want to use the DELTA_BUILD flag to determine whether to use the incremental toolchain, but to preserve
```
htaubenfeld, line 28:
```suggestion
# Generally, the DELTA_BUILD flag is used to determine whether to use the incremental toolchain, but to preserve
```

toolkit/Makefile
```diff
@@ -19,16 +21,29 @@ PACKAGE_REBUILD_LIST    ?=
...
+# backwards compatibility, we'll still use the INCREMENTAL_TOOLCHAIN flag if it's set.
```
htaubenfeld, line 29:
```suggestion
# backwards compatibility, use the INCREMENTAL_TOOLCHAIN flag if it's set.
```
*Nit*, try to not use "we".

htaubenfeld (approved):

toolkit/quickrebuild_autoconfig.mk
dmcilvaney: Would also love to have a config file, its listed in the tooling improvements document.

toolkit/scripts/toolchain/create_toolchain_in_container.sh
dmcilvaney: Your suggestion is definatly the correct `makefile` way to do things.

In this case though I'm concerned about avoiding throwing away this state with a `make clean`. That is a fairly common operation, and it would destroy the toolchain container.

While this does erode the self-contained nature of the build system I think its a reasonable compromise since the files should generally be static, and if we track the inputs correctly I don't think there will be any issues with stale state. (Also, it only triggers for `INCREMENTAL_TOOLCHAIN=y` so a full build will regenerate as needed.

toolkit/scripts/toolchain/create_toolchain_in_container.sh
dmcilvaney: Hmm, you are right I think the logic is inverted. It intent is to never use the cached version unless `INCREMENTAL_TOOLCHAIN=y` is set.

I think the correct version should be:

```bash
if [ "$INCREMENTAL_TOOLCHAIN" != "y" ] || [ -z "$(docker images -q marinertoolchain_populated:${sha_component_tag} 2>/dev/null)" ]; then
...
```

dmcilvaney: Hmm, you are right I think the logic is inverted. It intent is to never use the cached version unless `INCREMENTAL_TOOLCHAIN=y` is set.

I think the correct version should be:

```bash
if [ "$INCREMENTAL_TOOLCHAIN" != "y" ] || [ -z "$(docker images -q marinertoolchain_populated:${sha_component_tag} 2>/dev/null)" ]; then
...
```

toolkit/quickrebuild_autoconfig.mk
dmcilvaney: Done

toolkit/Makefile
dmcilvaney: Sure, I'll move everything related into the new file.

dmcilvaney: Maybe call it `incremental_builds.mk`.

toolkit/Makefile
dmcilvaney: Fixed

toolkit/quickrebuild_autoconfig.mk
dmcilvaney: The behavior of the quickbuild flags isn't truly affected by REBUILD_TOOLS=y/n (hence why it isn't error checked), but I'm trying to make this the start of a 'works by default' way of interacting with the tools. So why not just turn that on by default for people, one less flag to worry about.

dmcilvaney: I'll update comments for REBUILD_TOOLS and USE_CCACHE since they are both similar.

toolkit/quickrebuild_autoconfig.mk
dmcilvaney: See above comment as well, but I'd like these flags to be a new user-friendly way to run the tools. Going forward I think the REBUILD_TOOLS flag should just be defaulted to 'y' anyways.

toolkit/scripts/toolchain/create_toolchain_in_container.sh
dmcilvaney: Looks like https://github.com/microsoft/CBL-Mariner/pull/4026

In this case I think we should keep it.. We know the format of the archive, and this saves several minutes during the build process. (by default `tar` uses the single threaded version of `gzip` which is SUPER slow for this use case.

toolkit/Makefile
dmcilvaney: Cant see a reason your simple solution doesn't work, will use that.

toolkit/quickrebuild_autoconfig.mk
dmcilvaney: Done

toolkit/scripts/toolchain/create_toolchain_in_container.sh
dmcilvaney: We are exporting `${tag}`, but looking at it I'm not sure why. I'll remove this one at least.

toolkit/Makefile
dmcilvaney: Reworked file, but change is included.

toolkit/Makefile
dmcilvaney: Reworked file, but change is included.

toolkit/Makefile
```diff
@@ -19,16 +21,29 @@ PACKAGE_REBUILD_LIST    ?=
...
+DELTA_BUILD                     ?= n
+ifeq ($(origin DELTA_BUILD), undefined)
+# We didn't set DELTA_BUILD, so use the previous default behavior of INCREMENTAL_TOOLCHAIN for backwards compatibility.
```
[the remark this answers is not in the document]
dmcilvaney: Removed this section

toolkit/Makefile
```diff
@@ -19,16 +21,29 @@ PACKAGE_REBUILD_LIST    ?=
...
 INCREMENTAL_TOOLCHAIN           ?= n
+else
+# We set DELTA_BUILD, so inherit value for INCREMENTAL_TOOLCHAIN.
```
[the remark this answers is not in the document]
dmcilvaney: Removed this section

toolkit/quickrebuild_autoconfig.mk
```diff
@@ -0,0 +1,80 @@
...
+ifneq ($(filter n,$(INCREMENTAL_TOOLCHAIN)),)
+$(error REBUILD_TOOLCHAIN cannot be used with INCREMENTAL_TOOLCHAIN explicitly set to 'n')
+endif
+ifneq ($(filter n,$(DELTA_BUILD)),)
+$(error REBUILD_TOOLCHAIN cannot be used with DELTA_BUILD explicitly  set to 'n')
+endif
+ifneq ($(filter n,$(ALLOW_TOOLCHAIN_DOWNLOAD_FAIL)),)
+$(error REBUILD_TOOLCHAIN cannot be used with ALLOW_TOOLCHAIN_DOWNLOAD_FAIL explicitly set to 'n')
+endif
```
PawelWMS, line 37: Shouldn't the error messages in this section refer to `QUICK_REBUILD_TOOLCHAIN`? Similar to what we're printing for `QUICK_REBUILD_PACKAGES` below?

```suggestion
ifneq ($(filter n,$(REBUILD_TOOLCHAIN)),)
$(error QUICK_REBUILD_TOOLCHAIN cannot be used with REBUILD_TOOLCHAIN explicitly set to 'n')
endif
ifneq ($(filter n,$(INCREMENTAL_TOOLCHAIN)),)
$(error QUICK_REBUILD_TOOLCHAIN cannot be used with INCREMENTAL_TOOLCHAIN explicitly set to 'n')
endif
ifneq ($(filter n,$(DELTA_BUILD)),)
$(error QUICK_REBUILD_TOOLCHAIN cannot be used with DELTA_BUILD explicitly  set to 'n')
endif
ifneq ($(filter n,$(ALLOW_TOOLCHAIN_DOWNLOAD_FAIL)),)
$(error QUICK_REBUILD_TOOLCHAIN cannot be used with ALLOW_TOOLCHAIN_DOWNLOAD_FAIL explicitly set to 'n')
endif
```

toolkit/scripts/toolchain/create_toolchain_in_container.sh
PawelWMS: What you mentioned about `make clean` makes me an even stronger opponent of the current solution. We shouldn't leave any stage behind when that target is executed. It's definitely not expected behaviour and may lead to more confusion while debugging build issues.

We can chat in more detail offline, though. I'm guessing it might be faster.:)

toolkit/quickrebuild_autoconfig.mk
dmcilvaney: I've moved the docs here now

toolkit/quickrebuild_autoconfig.mk
dmcilvaney: For the logic: I feel like we should allow users to configure the tools to their needs where possible. In this case REBUILD_TOOLS=n won't stop the quick rebuild flow from working so I don't need to error check or override the user.

Make has some interesting precedence rules. While command line will always override the makefile, `?=` works as expected for exported environment variables (which I know some people use).
i.e.:
```makefile
FOO ?= baz
$(error echo $(FOO))
```
```bash
make
# baz

make FOO=bar
#bar

export FOO=fizz
make
#fizz
```
But using `FOO = baz` the last command would return `baz` rather than `fizz`.

PawelWMS: Scary stuff. Carry on then.:)

toolkit/scripts/toolchain/create_toolchain_in_container.sh
dmcilvaney: @anphel31 Thoughts?

anphel31: Here's the related PR for main: #3683 
I removed the compression on archives that contained only rpm or srpm files. I left the compression enabled on toolchain_from_container.tar.gz (since it contains individual toolchain files, not compressed rpms). So using the archive tool here should be fine.

christopherco (approved):

PR #5251, opened 2023-04-10
Add a script to update manifest files from pkg repo

0xba1a:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Adds a new script to update manifest RPM version based on PMC

**NO**

###### Test Methodology

For giving a toolkit to customers who wants to maintain some SPECS from their repo, they should be able to download the toolchain and pkggen RPMs from PMC. Neither main nor 2.0 branch have all the RPMs present in PMC all the time. Because when a 2.0 image build is in progress, those RPMs will not be available in PMC (not even in preview repo for some time).

So created this tool which will get the package names from the manifest files, get the corresponding PMC version using tdnf and update the manifest files.

Note: As the toolkit for customers is targeted to run only on Mariner, I created this script specific to Mariner.

### DIFF
--- a/toolkit/resources/manifests/package/update_manifest.py
+++ b/toolkit/resources/manifests/package/update_manifest.py
@@ -4,9 +4,52 @@ import os
 import sys
 
 
+# Check /etc/os-release for ID=mariner. Otherwise exit
+with open("/etc/os-release", "r") as f:
+    for line in f:
+        if "ID=mariner" in line:
+            break
+    else:
+        print("This script runs only on Mariner")
+        sys.exit(1)
+
+
 def usage():
     print("Usage: " + sys.argv[0] + " <arch>")
     sys.exit(1)
+
+
+def update_manifest_pkg_version(old_file, new_file):
+    # Remove the toolchain new files if they exist
+    if os.path.exists(new_file):
+        os.remove(new_file)
+    # Remove the pkggen new files if they exist
+    if os.path.exists(pkggen_file_new):
+        os.remove(pkggen_file_new)
+
+    # for each line in old_file
+    # split the line by last hypen and store it as the pkgname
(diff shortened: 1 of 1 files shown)


toolkit/resources/manifests/package/update_manifest.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python3
```
mfrw, line 1: We could probably do in the toolkit directory:
`git checkout 2.0-stable -- resources/manifests`

dmcilvaney: This is my goto command as well, however there are issues with it unfortunately. If one of the packages needed for cycle resolution is updated (and the cycle has an explicit version requirement in int) the core repo will fail to build (since it expects the toolchain to provide the exact package version needed to break the cycle, and the version may not match).

I've actually been working on a set of documents to help one of our customers work in the core repo. Its stiill early draft but it might be of interest: 
- https://github.com/microsoft/CBL-Mariner/pull/5199
- https://github.com/microsoft/CBL-Mariner/pull/5218
- https://github.com/microsoft/CBL-MarinerTutorials/pull/37

dmcilvaney: I have a section on getting stable commits here: https://github.com/microsoft/CBL-Mariner/blob/479bf4586a131a871bbdcd3f3a43e3e259e274c5/toolkit/docs/building/building.md#clone-and-sync-to-a-stable-commit

0xba1a: Yes. force fetch actually solves the problem. Let me abandon this PR. But let the branch be here, so this code can be used later if required in some other scenario.

PR #5297, opened 2023-04-13
Upgrade tdnf Version To 3.5.2

sameluch:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This PR is to upgrade tdnf to the latest version 3.5.2 to introduce the following changes from the upstream:
- include the --downloadonly flag instead of using the `tdnf download` command we have created through patches
- include the --no-deps flag from upstream for future use and in place of the `tdnf download-nodeps` command created via patch
- bring in additional upstream functionality such as `autoremove`

###### Change Log  

- Upgrade tdnf to version 3.5.2
- Modify go tools to use --downloadonly in place of `tdnf download`

**YES**

###### Associated issues  

- #xxxx

###### Test Methodology

-


SPECS/libguestfs/libguestfs.spec
```diff
@@ -755,9 +755,9 @@ cp %{SOURCE9} %{_sysconfdir}/yum.repos.d/allrepos.repo
 # Must keep in sync with BRs under "Build requirements for the appliance"
 # Download to
 mkdir -pv %{_var}/cache/tdnf
-tdnf download -y --disablerepo=* \
+tdnf isntall --downloadonly -y --disablerepo=* \
```
PawelWMS, line 758:
```suggestion
tdnf install --downloadonly -y --disablerepo=* \
```

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -299,7 +299,7 @@ func tdnfInstall(packages []string) (err error) {
 		return
 	}
 
-	installArgs := []string{"install", "-y", releaseverCliArg}
+	installArgs := []string{"install", "-y", releaseverCliArg, "--disablerepo=mariner-official-base"}
```
sameluch, line 302: This is one change that I am slightly skeptical of, though it doesn't seem to provide a negative impact to the build process. I added this since the worker chroot has not internet connection and therefore can't access the base repo anyway. By doing so, this allows us to narrow down errors a bit better than simply many cURL failures that can mask other problems within the tdnf install when building packages.

SPECS/cmake/cmake.spec
```diff
@@ -81,6 +81,9 @@ bin/ctest --force-new-ctest-process --rerun-failed --output-on-failure
 %{_prefix}/doc/%{name}-*/*
 
 %changelog
+* Wed May 04 2023 Sam Meluch <sammeluch@microsoft.com> - 3.21.4-5
```
anphel31, line 84: should be Wed May 03 ?

SPECS/shadow-utils/shadow-utils.spec
```diff
@@ -178,9 +176,6 @@ chmod 000 %{_sysconfdir}/shadow
 %{_libdir}/libsubid.so
 
 %changelog
-* Tue May 02 2023 CBL-Mariner Servicing Account <cblmargh@microsoft.com> - 4.9-12
```
anphel31, line 181: why is this shadow-utils patch being reverted?

SPECS/tdnf/tdnf.spec
```diff
@@ -90,6 +108,7 @@ python bindings for tdnf
 
 %package        autoupdate
 Summary:        systemd services for periodic automatic update
+Group:          Applciations/RPM
```
anphel31, line 111: typo: Applications

SPECS/tdnf/tdnf.spec
```diff
@@ -28,17 +31,22 @@ BuildRequires:  glibc-devel
 BuildRequires:  gpgme-devel
 BuildRequires:  libmetalink-devel
 BuildRequires:  libsolv-devel
+BuildRequires:  libxml2-devel
 BuildRequires:  make
 BuildRequires:  openssl-devel
```
anphel31, line 36: Should double check that the dependencies to build the new TDNF are satisfied in build_official_toolchain_rpms.sh.
Since this package is built in the toolchain environment, the BuildRequires in the spec are not parsed, and the toolchain packages build sequentially based on dependencies.
https://github.com/microsoft/CBL-Mariner/blob/main/toolkit/scripts/toolchain/build_official_toolchain_rpms.sh#L519

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -532,7 +532,7 @@ func initializeRpmDatabase(installRoot string, diffDiskBuild bool) (err error) {
 			return err
 		}
 	}
-	err = initializeTdnfConfiguration(installRoot)
+	//err = initializeTdnfConfiguration(installRoot)
```
anphel31, line 535: Can you add a comment here about the issue this line causes and why it can be removed

SPECS/shadow-utils/shadow-utils.spec
sameluch: This patch was causing problems in the build some time ago so I had reverted the change to help see my builds through in the pipelines prior to getting onto main/2.0. This commit should likely be taken out, given the fix was applied to 2.0/main a couple weeks ago here: https://github.com/microsoft/CBL-Mariner/commit/39e584023c93299673fff9b8ff6b3b7e3978926b

SPECS/tdnf/tdnf-default-mariner-release.patch
```diff
@@ -1,24 +1,30 @@
...
-Error(1022) : distroverpkg config entry is set to a package that is not installed. Check /etc/tdnf/tdnf.conf
-
-Fixes: Bug 24144834: TDNF does not support $basearch and $releasever in .repo files - distroverpkg error
-
-By default, determine $ basearch and $ releasever from this package:
-mariner_user@mariner-aks [ /etc/tdnf ]$ sudo tdnf --refresh list installed | grep mariner
-mariner-release.noarch                      1.0-1.cm1                @System
-
-Workaround: add the following to /etc/tdnf/tdnf.conf
-distroverpkg=mariner-release
```
PawelWMS, line 15: I'd leave this comment for future reference.

toolkit/resources/manifests/package/pkggen_core_aarch64.txt
```diff
@@ -193,10 +193,6 @@ nghttp2-1.46.0-2.cm2.aarch64.rpm
 curl-8.0.1-1.cm2.aarch64.rpm
 curl-devel-8.0.1-1.cm2.aarch64.rpm
 curl-libs-8.0.1-1.cm2.aarch64.rpm
-tdnf-3.2.2-4.cm2.aarch64.rpm
-tdnf-cli-libs-3.2.2-4.cm2.aarch64.rpm
-tdnf-devel-3.2.2-4.cm2.aarch64.rpm
-tdnf-plugin-repogpgcheck-3.2.2-4.cm2.aarch64.rpm
```
PawelWMS, line 199: I'd only update the package version but leave the lines where they are. I think we have a tool, which re-builds this file if needed, and it dictates the order of packages. Also, there's an order in which we want these packages installed and I'm not sure, if this is the source of this order. Might want to double-check with @anphel31.

SPECS/tdnf/tdnf.spec
PawelWMS:
```suggestion
Group:          Applications/RPM
```

SPECS/tdnf/tdnf.spec
```diff
@@ -47,14 +55,14 @@ BuildRequires:  glib
...
 BuildRequires:  python3-requests
-BuildRequires:  python3-setuptools
 BuildRequires:  python3-xml
 %endif
 
 %description
 tdnf is a yum/dnf equivalent which uses libsolv and libcurl
 
 %define _tdnfpluginsdir %{_libdir}/tdnf-plugins
+%define _tdnf_history_db_dir %{_libdir}/sysimage/tdnf
```
PawelWMS, line 65: Since we're here, could we move these macros to the top of the spec? It makes it easier to find them.

SPECS/tdnf/tdnf-default-mariner-release.patch
```diff
@@ -1,24 +1,30 @@
...
+index 602b249..59b058e 100644
 --- a/client/defines.h
 +++ b/client/defines.h
-@@ -121,7 +121,7 @@ typedef enum
+@@ -137,7 +137,7 @@ typedef enum
+    and configurable with "persistdir" at run time */
+ #define TDNF_DEFAULT_DB_LOCATION          HISTORY_DB_DIR
+ 
+-#define TDNF_DEFAULT_DISTROVERPKG         "system-release"
++#define TDNF_DEFAULT_DISTROVERPKG         "mariner-release"
```
PawelWMS, line 19: Btw. I feel now all of this could be avoided, if we just added `Provides: system-release = %{version}-%{release}` inside `mariner-release.spec` for the sake of compatibility. It looks like we already have `Provides: system-release(releasever)` there anyway, so if you have a minute, it might be worth checking, if we could completely drop this patch with that change. However, it's a P2 at this point.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -532,7 +532,6 @@ func initializeRpmDatabase(installRoot string, diffDiskBuild bool) (err error) {
 			return err
 		}
 	}
-	err = initializeTdnfConfiguration(installRoot)
```
PawelWMS, line 535: This looks like a significant change. I'd definitely verify the VHD, ISO, and container images, if they still build correctly and TDNF runs without issues.

toolkit/tools/pkgworker/pkgworker.go
PawelWMS: I share your skepticism.:) The main reason is that we should not hard-code anything external to the toolkit inside its code. The toolkit should only need to know about the internal RPM repos present in the chroot like the ones defined in `toolkit/resources/manifests/package/local.repo`. The `mariner-official-base` repo is not really part of the toolkit, may change at any time, and in general the toolkit should not have any dependency on such external entities. We do depend on the repo being in the chroot for the sake of fetching packages, but the dependency is on the chroot packages providing the repo configurations, not on the actual repo names.

Also, I wouldn't worry about the worker accidentally reaching out to any unwanted repositories. If I remember correctly, this code is run when we already downloaded/built all of the required packages, so they are available in the local RPM repos. If we see it reaching out elsewhere, that would probably (I might have forgotten some use case) be a bug, which we should address separately, because it would mean we have a more sever design flaw somewhere in our code. Your current change might hide that potential sever issue, so I'd prefer to revert it.

In other words: it's good to make sure we don't try to access the wrong repos, but I wouldn't include such "just in case" updates in this PR as that's our-of-scope and if there really is an issue for us to address here, it might require more investigation and a fix elsewhere in our design.

toolkit/resources/manifests/package/pkggen_core_aarch64.txt
sameluch: The build order needed to be updated for tdnf due to the change in dependencies for the new tdnf version that was causing a build error.

SPECS/tdnf/tdnf.spec
sameluch: Sure thing!

SPECS/tdnf/tdnf-default-mariner-release.patch
sameluch: sure I can readd it, it must have dropped when I regenerated the patch.

toolkit/tools/imagegen/installutils/installutils.go
sameluch: I have currently tested ISO and vhdx images locally with a change to how the `mariner-release` package is included in the image. Rather than installing `mariner-release` ahead of all packages using rpm, the package will instead be added to the list of packages to build giving the same functionality without the reliance on `tdnf` not detecting packages as already installed if done so via rpm.

toolkit/tools/imagegen/installutils/installutils.go
sameluch: With the removal of this line I added some context to where `mariner-release` is added to the package list in the imager instead to add clarification as to why the change was made/needed.

SPECS/tdnf/tdnf.spec
sameluch: Looks like everything that tdnf needs is still built/installed ahead of tdnf in this script.

SPECS/tdnf/tdnf.spec
oliviacrain: We don't use these tags, might as well just remove it if we're touching it at all.

SPECS/tdnf/tdnf.spec
```diff
@@ -28,17 +33,22 @@ BuildRequires:  glibc-devel
 BuildRequires:  gpgme-devel
 BuildRequires:  libmetalink-devel
```
oliviacrain, line 34: IIRC `libmetalink` is not required for the build anymore.

https://github.com/vmware/tdnf/pull/266

oliviacrain: Seems to be the case in the currently existing metalink plugin code, which has no dependency on libmetalink:

https://github.com/vmware/tdnf/blob/v3.5.2/plugins/metalink/CMakeLists.txt

We can go back later and remove libmetalink from the toolchain (since it's unmaintained, with a last release in 2015), but it'd be nice to get it out of the spec.

SPECS/tdnf/tdnf.spec
```diff
@@ -16,7 +20,9 @@ Source4:        tdnfrepogpgcheck.conf
 Patch0:         tdnf-mandatory-space-list-output.patch
 Patch1:         tdnf-default-mariner-release.patch
 Patch2:         tdnf-enable-plugins-by-default.patch
-Patch3:         tdnf-add-download-command.patch
+Patch3:         tdnf-printf-fix.patch
+Patch4:         tdnf-sqlite-library.patch
+Patch5:         tdnf-GetRepoMD-fix.patch
```
PawelWMS, line 25: Could you add a link to [the PR fixing this issue upstream](https://github.com/vmware/tdnf/pull/432), so we know what to track and can remove it whenever we update to a version of TDNF, which already contains the upstream fix?
```suggestion
// Patch to be removed once we upgrade to a version of TDNF containing the upstream fix:
// https://github.com/vmware/tdnf/pull/432
Patch5:         tdnf-GetRepoMD-fix.patch
```

SPECS/tdnf/tdnf.spec
```diff
@@ -72,9 +78,19 @@ Group:          Development/Libraries
 %description cli-libs
 Library providing cli libs for tdnf like clients.
 
+%package plugin-metalink
+Summary:        tdnf plugin providing metalink functionality for repo configurations
+Group:          Applications/RPM
+Requires:       tdnf = %{version}-%{release}
```
PawelWMS, line 84: Nit-pick: for requirements within the same spec, we try to use the macros. Just a consistency thing.
```suggestion
Requires:       %{name} = %{version}-%{release}
```

SPECS/tdnf/tdnf.spec
```diff
@@ -72,9 +78,19 @@ Group:          Development/Libraries
...
+Requires:       libxml2
+
+%description plugin-metalink
+tdnf plugin providing metalink functionality for repo configurations
+
 %package        plugin-repogpgcheck
 Summary:        tdnf plugin providing gpg verification for repository metadata
 Group:          Development/Libraries
+Requires:       tdnf = %{version}-%{release}
```
PawelWMS, line 93: Similar comment regarding macro usage:
```suggestion
Requires:       %{name} = %{version}-%{release}
```

SPECS/tdnf/tdnf.spec
```diff
@@ -99,51 +116,67 @@ systemd services for periodic automatic update
...
+pushd %{__cmake_builddir}/python
+%py3_install
+popd
 find %{buildroot} -name '*.pyc' -delete
 
-%ldconfig_scriptlets
+%post
+/sbin/ldconfig
+# Required step to ensure new history util from version 3.4.1 does not fail
+[ -f %{_tdnf_history_db_dir}/history.db ] || %{_libdir}/tdnf/tdnf-history-util init
```
PawelWMS, line 159: To be on the safe side, let's use `if`. With the current implementation running the post-install script with a setting like `set -e`, would actually error out if `%{_tdnf_history_db_dir}/history.db` is missing. I don't know, how the scripts are ran right now, but this should be more stable in case of any behaviour changes in the future:
```suggestion
if [ -f %{_tdnf_history_db_dir}/history.db ]; then
    %{_libdir}/tdnf/tdnf-history-util init
fi
```

SPECS/tdnf/tdnf.spec
```diff
@@ -99,51 +116,67 @@ systemd services for periodic automatic update
...
+%{_bindir}/tdnfj
+%{_bindir}/tdnf-config
 %{_bindir}/tdnf-cache-updateinfo
 %{_libdir}/libtdnf.so.3
 %{_libdir}/libtdnf.so.3.*
+%{_libdir}/tdnf/tdnf-history-util
 %config(noreplace) %{_sysconfdir}/tdnf/tdnf.conf
 %dir %{_var}/cache/tdnf
+%dir %{_tdnf_history_db_dir}
 %{_datadir}/bash-completion/completions/tdnf
```
PawelWMS, line 180: NIt-pick: could we put the files in alphabetical order? Makes it easier to read the specs.

toolkit/tools/imager/imager.go
```diff
@@ -144,6 +145,13 @@ func buildSystemConfig(systemConfig configuration.SystemConfig, disks []configur
 		return
 	}
 
+	// The old way of including mariner-release in images was reliant on some old behavior
+	// from tdnf, i.e. not detecting packages installed via rpm. Which caused an issue when
+	// tdnf was upgraded past 3.4.1 to include a package install history. Instead of pre-installing
+	// mariner-release ahead of calculating the packages + dependencies for the image, now we will
+	// simply add mariner-release to each image's package list to be installed.
```
PawelWMS, line 152: I'd say there's no need to reference any "old way" of doing things, since most readers won't be aware what the "old way" was from this comment. Instead, I'd focus on explaining what we're doing here given the current situation and why it is needed.

SPECS/tdnf/tdnf.spec
```diff
@@ -99,51 +116,67 @@ systemd services for periodic automatic update
...
 %{_bindir}/yum
+%{_bindir}/tdnfj
+%{_bindir}/tdnf-config
 %{_bindir}/tdnf-cache-updateinfo
 %{_libdir}/libtdnf.so.3
 %{_libdir}/libtdnf.so.3.*
+%{_libdir}/tdnf/tdnf-history-util
 %config(noreplace) %{_sysconfdir}/tdnf/tdnf.conf
 %dir %{_var}/cache/tdnf
+%dir %{_tdnf_history_db_dir}
```
dmcilvaney, line 179: Do we need to add anything here so the history doesn't get wiped when we update? Not sure exactly how this all works here, but I'd assume we want to leave the history db alone when updating tdnf versions.

sameluch: The history db should be left alone by default during an upgrade. The only thing that would change it would be the history-util init call that will happen on the initial install if the file is not already present.

SPECS/tdnf/tdnf.spec
sameluch: I changed this up to use an if statement as I agree its a bit more clear than using bash `||`. The goal of it in the first place is to only run the `tdnf-history-util init` if the history.db file is not in place already, given it makes all tdnf calls fail if it is not.

oliviacrain (approved):

dmcilvaney (approved):

SPECS/tdnf/tdnf.spec
PawelWMS: Huh, not sure why I though that it's going to bail out on `||` then. Sorry for the confusion.

PawelWMS (approved):

PR #4871, opened 2023-02-16
Install vmlinux with root executable permissions

rlmenge:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 

debuginfo packages are generated via find-debuginfo script. This script is packaged in the [debugedit rpm](https://github.com/microsoft/CBL-Mariner/blob/main/SPECS/debugedit/debugedit.spec). The script is run after the %install phase of an rpm. It looks at the BUILDROOT for its sources. Specifically it will look for all the executables in this [line](https://sourceware.org/git/?p=debugedit.git;a=blob;f=scripts/find-debuginfo.in;h=7dec3c3bfdc59a5b3cd53838f9cf748c0b1a932c;hb=HEAD#l422).  Because we were initally installing vmlinux with 644 permissions, it was not seen as an executable and not analyzed by find-debuginfo. Thus, only the modules had their debuginfo extracted leading to missing debug sources such as tcp.c.

Fix by giving vmlinux root executable permissions.
This increases the kernel-debuginfo package from about 828MB to 847MB (19MB difference)

###### Change Log  

- Install vmlinux with root executable permissions 

###### Does this affect the toolchain?  

**YES**

[...]

PR #5301, opened 2023-04-14
Disable recomputing build-ids for kernel debuginfo packages

rlmenge:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Because the kernel's vmlinux is now being scanned by `find-debuginfo` as a result of #4871, the default arguments set by `rpm` for `find-debuginfo` cause vmlinux's build-id to no longer match the vmlinuz shipped with the kernel. 

By default, `rpm` sets a series of macros for the arguments it passes to `find-debuginfo.sh`, the script which extracts debug symbols. One of these arguments is `_unique_build_ids` which sets `--build-id-seed`. This causes `find-debuginfo` to recompute the build-ids for binaries. Setting the `rpm` flag `_no_recompute_build_ids` (which sets `-n` for `find-debuginfo`)  is not enough as it conflicts with the build-id-seed and causes an error. This is not desired behavior and thus, turn off the relevant default settings for rpm's arguments to `find-debuginfo`.

The defaults for rpm's find-debuginfo call can be seen [here](https://github.com/rpm-software-management/rpm/blob/master/macros.in#L146)
The find-debuginfo.sh is from debugedit and the arguments can be seen upstream [here](https://sourceware.org/git/?p=debugedit.git;a=blob;f=scripts/find-debuginfo.in;h=7dec3c3bfdc59a5b3cd53838f9cf748c0b1a932c;hb=refs/heads/main#l28)

Change the argument values for the following rpm macros (affected find-debuginfo argument in parenthesis):
- `%undefine _unique_build_ids` (`--build-id-seed`) : removes build-id-seed
- `%undefine _unique_debug_names` (`--unique-debug-suffix`): unset because this macro requires the above unqiue_build_ids
- `%global _missing_build_ids_terminate_build 1`(`--strict-build-id`): ensures build-ids be present 
- `%global _no_recompute_build_ids 1` (`-n`): explicilty turns off recomputing build-ids

These flags cause the resulting find-debuginfo call to be
```
/bin/find-debuginfo -j16 --strict-build-id -n -m --unique-debug-src-base kernel-5.15.102.1-100.cm2.x86_64 /usr/src/mariner/BUILD/CBL-Mariner-Linux-Kernel-rolling-lts-mariner-2-5.15.102.1
```

Additional flags worth exploring:
`_include_minidebuginfo` (-m)
`_find_debuginfo_opts r` (-r)

###### Change Log  

- Disable recomputing build-ids for kernel debuginfo package

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/44126568

###### Test Methodology

- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=344017&view=results
- Buddy build with all kernels: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=346312&view=results
- confirmed that these settings do NOT affect other packages' settings
- repro steps for checking build id. These steps use[ extract-vmlinux](https://github.com/microsoft/CBL-Mariner-Linux-Kernel/blob/8e63ee223a45b872a289856007ee9bbc7fbaae9f/scripts/extract-vmlinux#L4) from the kernel scripts.
```
rpm2cpio kernel.rpm | cpio -idmv
rpm2cpio kernel-debuginfo.rpm | cpio -idmv

cp lib/modules/<version>/vmlinuz .
cp ./usr/lib/debug/lib/modules/<version>/vmlinux vmlinux-debug

bash extract-vmlinux vmlinuz > vmlinux-decompress

readelf -n vmlinux-decompress | grep "Build ID:"
    Build ID: a7e38fcf5a17a5ab5f92c768080234e5eb29a2b3
readelf -n vmlinux-debug | grep "Build ID:"
    Build ID: a7e38fcf5a17a5ab5f92c768080234e5eb29a2b3
```

### DIFF
--- a/SPECS/kernel/kernel.spec
+++ b/SPECS/kernel/kernel.spec
@@ -2,6 +2,12 @@
 %global sha512hmac bash %{_sourcedir}/sha512hmac-openssl.sh
 %define uname_r %{version}-%{release}
 
+# Edit default debuginfo settings
+%undefine _unique_build_ids
+%undefine _unique_debug_names
+%global _missing_build_ids_terminate_build 1
+%global _no_recompute_build_ids 1
+
 %ifarch x86_64
 %define arch x86_64
 %define archdir x86
@@ -18,7 +24,7 @@
 Summary:        Linux Kernel
 Name:           kernel
 Version:        5.15.107.1
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv2
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -410,6 +416,9 @@ ln -sf linux-%{uname_r}.cfg /boot/mariner.cfg
 %{_sysconfdir}/bash_completion.d/bpftool
 
 %changelog
+* Wed Apr 19 2023 Rachel Menge <rachelmenge@microsoft.com> - 5.15.107.1-2
+- Disable rpm's debuginfo defaults which generate build-ids
+
 * Tue Apr 18 2023 CBL-Mariner Servicing Account <cblmargh@microsoft.com> - 5.15.107.1-1
 - Auto-upgrade to 5.15.107.1
 
--- a/SPECS/kernel/kernel.spec
+++ b/SPECS/kernel/kernel.spec
@@ -2,7 +2,11 @@
 %global sha512hmac bash %{_sourcedir}/sha512hmac-openssl.sh
 %define uname_r %{version}-%{release}
 
-# Edit default debuginfo settings
+# find_debuginfo.sh arguments are set by default in rpm's macros.
+# The default arguments regenerate the build-id for vmlinux in the 
+# debuginfo package causing a mismatch with the build-id for vmlinuz in
+# the kernel package. Therefore, explicilty set the relevant default 
+# settings to prevent this behavior.
 %undefine _unique_build_ids
 %undefine _unique_debug_names
 %global _missing_build_ids_terminate_build 1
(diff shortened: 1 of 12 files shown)


SPECS/kernel/kernel.spec
```diff
@@ -2,6 +2,12 @@
 %global sha512hmac bash %{_sourcedir}/sha512hmac-openssl.sh
 %define uname_r %{version}-%{release}
 
+# Edit default debuginfo settings
```
PawelWMS, line 5: Seeing how this is uncommon, could you expand the comment to say what this does and why we do it?

ddstreet, line 5: for reference, fedora's corresponding section for these changes:
https://src.fedoraproject.org/rpms/kernel/blob/rawhide/f/kernel.spec#_671

shouldn't these be protected inside %if %{with_debuginfo} like fedora does? If not, wouldn't these changes make it impossible for us to build a kernel rpm without debuginfo (i.e. rpmbuild --without debuginfo would now fail, I think)

also as you noted, some of the other settings there might be useful and/or needed before merging this? At least a quick investigation for each param?

ddstreet (changes requested):

PawelWMS (commented):
Looks good to me, not approving just because I lack expertise here.:)

SPECS/kernel/kernel.spec
rlmenge: Mariner does not support these macros (%with_debuginfo) as of today. I did a quick test and running `rpmbuild -ba --nodebuginfo /usr/src/mariner/SPECS/kernel.spec 2>&1 | tee kernel-defconfig.log` did not fail due to the new macros but it DID fail because vmlinux is now unpackaged file. Fedora also protects the vmlinux install step: https://src.fedoraproject.org/rpms/kernel/blob/rawhide/f/kernel.spec#_2160

ddstreet: ah ok, so our build already didn't work without debuginfo, so this doesn't break anything that wasn't already broken. It doesn't seem like anyone was doing any non-debuginfo builds (or else someone presumably would have fixed it already) so it LGTM. Thanks!

ddstreet: +1 a bit more comment in the spec would likely be helpful in the future, otherwise LGTM

ddstreet (approved):

vinceaperri (approved):

anphel31 (approved):

PR #5305, opened 2023-04-15
Create kernel conflicts and add specialized kernel subpackages

rlmenge:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

rework of #5131

There was a bug where certain specialized kernels (`kernel-azure`, `kernel-hci`, `kernel-rt`) would create subpackages of the same name (`bpftool`, `python3-perf`) as the generic kernel. This will cause packages to be published and overwritten. Therefore, explicitly create conflicts between all the kernels, create specialized version names of the subpackages from specialized kernels (ex. `bpftool-azure`)  and add requires for the relevant kernel. This should be investigated further for a less involved solution.

Subpackage changes from kernel-azure
`bpftool` -> `bpftool-azure` which conflicts with `bpftool` and variants
`python3-perf `-> `python3-perf-azure` which conflicts with `python3-perf` and variants

Subpackage changes from kernel-hci 
`bpftool` -> `bpftool-hci` which conflicts with `bpftool` and variants
`python3-perf` ->python3-perf-hci` which conflicts with `python3-perf` and variants

Subpackage changes from kernel-rt
`bpftool` -> `bpftool-rt` which conflicts with `bpftool` and variants
`python3-perf` ->python3-perf-rt` which conflicts with `python3-perf` and variants

This also impacts rpm installation. In order to change your kernel package, you will need to uninstall your current kernel and install the desired kernel package. Version updates should continue as expected. Only subpackages built with that kernel will then be able to be installed.

###### Change Log  

- Add conflicts for the kernels
 
**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/43814917/

###### Test Methodology

- Buddy Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=343891&view=results
- Full Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=344494&view=results

### DIFF
--- a/SPECS-EXTENDED/kernel-rt/kernel-rt.spec
+++ b/SPECS-EXTENDED/kernel-rt/kernel-rt.spec
@@ -13,7 +13,7 @@
 Summary:        Realtime Linux Kernel
 Name:           kernel-rt
 Version:        5.15.55.1
-Release:        2%{?dist}
+Release:        3%{?dist}
 License:        GPLv2
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -60,6 +60,11 @@ ExclusiveArch:  x86_64
 %ifarch x86_64
 BuildRequires:  pciutils-devel
 %endif
+Conflicts:      kernel
+Conflicts:      kernel-azure
+Conflicts:      kernel-hci
+Conflicts:      kernel-mshv
+Conflicts:      kernel-uvm
 # When updating the config files it is important to sanitize them.
 # Steps for updating a config file:
 #  1. Extract the linux sources into a folder
@@ -121,11 +126,13 @@ Requires:       audit
 %description tools
 This package contains the 'perf' performance analysis tools for Linux kernel.
 
-%package -n     python3-perf
+%package -n     python3-perf-rt
 Summary:        Python 3 extension for perf tools
 Requires:       python3
+Requires:       %{name} = %{version}-%{release}
+Provides:       python3-perf
 
-%description -n python3-perf
+%description -n python3-perf-rt
 This package contains the Python 3 extension for the 'perf' performance analysis tools for Linux kernel.
 
 %package dtb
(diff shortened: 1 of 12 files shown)


nicogbg (approved):

Redent0r (approved):

SPECS-EXTENDED/kernel-rt/kernel-rt.spec
```diff
@@ -121,11 +126,13 @@ Requires:       audit
...
 
-%package -n     python3-perf
+%package -n     python3-perf-rt
 Summary:        Python 3 extension for perf tools
 Requires:       python3
+Requires:       %{name} = %{version}-%{release}
+Provides:       python3-perf
 
-%description -n python3-perf
+%description -n python3-perf-rt
```
PawelWMS, line 135: My concerns with this approach:
- These two packages seem to have no run-time dependency on any kernel binaries. Or at least nothing, that RPM could detect automatically - running `dnf repoquery --requires --resolve --recursive bpftool` returns no kernel subpackages. It seems to me, that it's very likely, they don't really have a dependency on any specific kernel version of build. Or in other words: these tools seem no different from tools produced by other specs.
- Our build tools are not smart enough to pick the right version of these packages when building an image. The tools check one dependency at a time, so a dependency on `python3-perf` would present multiple candidates. The candidates resolution mechanism (we ask RPM to pick the package it would prefer) is where we're going to probably end up picking packages at random, because at that point in the build it's unknown, which kernel we're going to need.
- I feel like we're introducing unnecessary redundancies with these two packages.

Considering that, I think we might be better off building them only from one kernel spec and removing them from all other ones.

PawelWMS (dismissed):
Blocking until we've figured out the questions of building separate `bpftools` and `python3-perf` packages vs building them from only one kernel spec.

PR #5424, opened 2023-04-28
Avoid JIT'd Perl in dracut to avoid SELinux errors

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Dracut uses `-P` when calling `grep` in one location. This requires executing JIT'd code out of memory which is blocked by SELinux. While not a fatal error it generates a large number of SELinux and audit reports. (see https://github.com/dracutdevs/dracut/issues/2323). Use a more basic call to `grep` and use `awk` to extract the result instead.

###### Change Log  

- Remove `dracut` call to `grep` `PCRE` and replace with an extra call to `awk` instead.

**NO**

###### Associated issues  

- https://github.com/dracutdevs/dracut/issues/2323

###### Test Methodology

- Local build and testing via VM


oliviacrain (approved):

SPECS/dracut/fix-grep-selinux-execmem.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/dracut-functions.sh b/dracut-functions.sh
```
christopherco, line 1: Can this be a git patch format that also has comments for why this patch is needed?

dmcilvaney: Just submitted the PR upstream, will use that commit instead.

oliviacrain (approved):

dmcilvaney:
> Hello - dracut upstream here. Thank you for your contribution upstream and sorry for the slightly off-topic note. Noticed that Mariner is on dracut v55, which is now about 2 years old (latest is v59). Most Linux distros - including Photon OS is on v59. Is there anything we can help to update dracut for Mariner ? What would be the best process to go about it ?
Hey @LaszloGombos , happy to chat! For Mariner we try to limit big package updates to our major releases (we went to v55 ~ around Mariner-2.0). 

Dracut is definitely on our todo list for the next release though, (alongside grabbing a newer systemd) since we are looking at supporting some new boot flows and features. I've made issue #5494 as a place to chat and track this. I'd be happy for any advice you have. We are certainly open to public contributions as well, if that is something you are interested in (although we don't have a branch for the next major release yet).

PR #5430, opened 2023-05-01
Correct FRR installation macro to resolve installation errors.

MsftSam:
… create_package (available) for FRR installation

###### Merge Checklist  

- [] Ready to merge

---

###### Summary 

This PR removes an unavailable macro (sysusers_create_compat) and replaces it with an available variant (sysusers_create_package) that allows FRR to be successfully installed from the extended repo.

###### Change Log  

Version was incremented, and the above mentioned macros were corrected, allowing successful installation.

This does not affect the toolchain.

###### Associated issues  

- Internally this is bug #44265925

###### Test Methodology

This change has been tested for package installation and execution on a clean Mariner VM
InstallSuccess
ExecSuccess

### DIFF
--- a/SPECS-EXTENDED/frr/frr.spec
+++ b/SPECS-EXTENDED/frr/frr.spec
@@ -6,7 +6,7 @@
 Summary:        Routing daemon
 Name:           frr
 Version:        8.4.2
-Release:        2%{?dist}
+Release:        3%{?dist}
 License:        GPLv2+
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -179,7 +179,7 @@ rm %{buildroot}%{_libdir}/frr/*.so
 rm -r %{buildroot}%{_includedir}/frr/
 
 %pre
-%sysusers_create_compat %{SOURCE2}
+%sysusers_create_package %{name} %{SOURCE2}
 
 %post
 %systemd_post frr.service
--- a/SPECS-EXTENDED/frr/frr.spec
+++ b/SPECS-EXTENDED/frr/frr.spec
@@ -266,6 +266,9 @@ rm tests/lib/*grpc*
 %endif
 
 %changelog
+* Wed May 3 2023 Samuel Mueller <samuelle@microsoft.com> - 8.4.2-2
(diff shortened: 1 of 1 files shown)


SPECS-EXTENDED/frr/frr.spec
```diff
@@ -179,7 +179,7 @@ rm %{buildroot}%{_libdir}/frr/*.so
 rm -r %{buildroot}%{_includedir}/frr/
 
 %pre
-%sysusers_create_compat %{SOURCE2}
+%sysusers_create_package %{name} %{SOURCE2}
```
PawelWMS, line 182: How do we know these two macros do the same thing? I want to make sure we're doing the right thing for this package. Also, since `sysusers_create_compat ` seems to be [a macro still in use by Fedora](https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/), could we instead bring in the package providing it to support the original macro instead? This would also make working with new specs we import from Fedora easier.

SPECS-EXTENDED/frr/frr.spec
```diff
@@ -6,7 +6,7 @@
 Summary:        Routing daemon
 Name:           frr
 Version:        8.4.2
-Release:        2%{?dist}
+Release:        3%{?dist}
```
PawelWMS, line 9: Please also update the changelog at the bottom of the spec.

SPECS-EXTENDED/frr/frr.spec
oliviacrain: I don't disagree that we should strive for compatibility, but we'd like to promote this package before the next release and this is the faster option. I'll create an ADO task to track the macro addition.

To answer for Sam: `%sysusers_create_compat` output by definition emulates the behavior of `%sysusers_create_package`, switching out the use of `systemd-sysusers` for `shadow-utils` utilities.

PawelWMS: Thanks, @oliviacrain! As long as the result is the same, I'm fine with this change.

oliviacrain (approved):
LGTM!

[Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=354733)

PR #5431, opened 2023-05-01
Allow PackageRepo field to have configurable GPG

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
We support defining custom package repo files for the ISO installer to use, and optionally install into the final system. Currently they only support the Microsoft GPG keys and enforce signature checking. This change will make that behavior customizable.

Currently we don't support custom keys for use in the ISO installer itself, since we don't currently have a mechanism to provision the keys into the installer. This can likely be done via an addition to the `AdditionalFiles` logic but is not critical right now.

###### Change Log  

- Add `GPGCheck` and `GPGKeys` fields to the `PackageRepo` object.
- Add some logic to check that custom keys aren't being used in the ISO installer.

**NO**

###### Associated issues  

- https://dev.azure.com/microsoft/OS/_workitems/edit/42608887

###### Test Methodology

- Local Builds


toolkit/tools/imagegen/configuration/configuration_test.go
```diff
@@ -456,14 +456,18 @@ var expectedConfiguration Config = Config{
...
+					Install:  true,
+					GPGCheck: true,
+					GPGKeys:  "file:///etc/pki/rpm-gpg/MY-CUSTOM-KEY",
+				},
+				{
+					Name:     "repo2",
+					BaseUrl:  "https://repo2.com",
+					Install:  false,
+					GPGCheck: false,
+					GPGKeys:  "file:///etc/pki/rpm-gpg/MICROSOFT-RPM-GPG-KEY file:///etc/pki/rpm-gpg/MICROSOFT-METADATA-GPG-KEY",
```
PawelWMS, line 470: Do we need `GPGKeys` if `GPGCheck` is set to `false`?

dmcilvaney: No, but the default if blank will be our MS keys. In this case parsing the .json which doesn't set the keys will result in this output. (Open to dissenting opinions on how to handle that, but I'm looking to not break anyone using this. Currently we hard-code the keys to those two MS keys).

toolkit/tools/imagegen/configuration/packagerepo.go
```diff
@@ -139,15 +165,13 @@ func (p *PackageRepo) repoUrlIsValid() (err error) {
 
 func writeAdditionalFields(stringBuilder *strings.Builder) (err error) {
 	const (
-		gpgKey       = "gpgkey=file:///etc/pki/rpm-gpg/MICROSOFT-RPM-GPG-KEY file:///etc/pki/rpm-gpg/MICROSOFT-METADATA-GPG-KEY\n"
 		enable       = "enabled=1\n"
-		gpgCheck     = "gpgcheck=1\n"
 		repogpgCheck = "repo_gpgcheck=1\n"
```
henryli001, line 169: Shouldn't the repogpgCheck field also be customizable? If this field is set to true, then yum will perform a signature check on the repodata and it will fail if there's no repo gpg key.

toolkit/tools/imagegen/configuration/packagerepo.go
```diff
@@ -59,6 +74,17 @@ func (p *PackageRepo) IsValid() (err error) {
...
 	}
 
+	// Repos for ISO use only (install = false) cannot support custom GPG keys if GPGCheck is enabled. Warn the user even
+	// if we are installing the repo file, as the repo file will not be usable in the installer.
+	if p.GPGCheck && p.GPGKeys != getDefaultGPGKeys() {
+		if !p.Install {
+			err = fmt.Errorf("invalid value for package repo '%s' [GPGKeys] (%s), custom GPG keys are only supported for repos that are installed", p.Name, p.GPGKeys)
+			return
+		} else {
+			logger.Log.Warnf("Warning: Custom GPG keys are only supported for repos that are installed. The repo file for '%s' will not be usable in the installer.", p.Name)
```
PawelWMS, line 84: I wouldn't expect the user to know what we mean by "installer" and this warning message may just be confusing for them. Could we re-word it to instead say that the installed repos will be available only in the built image, but not for the sake of populating it with packages during the image build? Or at least that's what I understand this warning is about.

toolkit/tools/imagegen/configuration/packagerepo.go
```diff
@@ -203,6 +227,28 @@ func createCustomRepoFile(fileName string, packageRepo PackageRepo) (err error)
...
+	}
+
+	// Write the repo GPGCheck field
+	var gpgVal string
+	if packageRepo.GPGCheck {
+		gpgVal = "1"
+	} else {
+		gpgVal = "0"
+	}
+	repoGpgCheck := fmt.Sprintf("gpgcheck=%s\n", gpgVal)
```
PawelWMS, line 245: Suggestion to reduce the amount of code:
```suggestion
	repoGpgCheck := "gpgcheck=0\n"
	if packageRepo.GPGCheck {
		repoGpgCheck = "gpgcheck=1\n"
	}
```

toolkit/tools/imagegen/configuration/packagerepo.go
```diff
@@ -203,6 +227,28 @@ func createCustomRepoFile(fileName string, packageRepo PackageRepo) (err error)
...
+	_, err = stringBuilder.WriteString(repoGpgCheck)
```
PawelWMS, line 246: Suggestion to reduce the amount of code:
```suggestion
	if packageRepo.GPGCheck {
			_, err = stringBuilder.WriteString(gpgcheck=1\n)
	} else {
			_, err = stringBuilder.WriteString(gpgcheck=0\n)
	}
```

toolkit/tools/imagegen/configuration/packagerepo.go
```diff
@@ -203,6 +227,28 @@ func createCustomRepoFile(fileName string, packageRepo PackageRepo) (err error)
...
+	if err != nil {
+		logger.Log.Errorf("Error writing repo GPGCheck: %s. Error: %s", repoGpgCheck, err)
+		return
+	}
```
PawelWMS, line 250: Suggestion to reduce the amount of code, since we have only two options here:
```suggestion
	if packageRepo.GPGCheck {
			_, err = stringBuilder.WriteString("gpgcheck=1\n")
	} else {
			_, err = stringBuilder.WriteString("gpgcheck=0\n")
	}
	if err != nil {
		logger.Log.Errorf("Error writing repo GPGCheck. Error: %s", err)
		return
	}
```

toolkit/tools/imagegen/configuration/packagerepo.go
```diff
@@ -25,15 +25,30 @@ import (
...
 }
 
 // UnmarshalJSON Unmarshals a PackageRepo entry
 func (p *PackageRepo) UnmarshalJSON(b []byte) (err error) {
 	// Use an intermediate type which will use the default JSON unmarshal implementation
 	type IntermediateTypePackageRepo PackageRepo
+
+	// Set default values
+	p.GPGCheck = getDefaultGPGCheck()
+	p.GPGKeys = getDefaultGPGKeys()
```
PawelWMS, line 50: Since this is the only place we use these two new functions, I think we can simplify to use `const` instead. This way we'd be also more consistent with how we use constants in other places of the code.

toolkit/tools/imagegen/configuration/packagerepo.go
dmcilvaney: Ah, missed that. Will also include it.

toolkit/tools/imagegen/configuration/packagerepo.go
dmcilvaney: There are a few spots in these files we do make these types of defaults publicly available, was just following that flow. Just using consts is better here though, agreed.

toolkit/tools/imagegen/configuration/packagerepo.go
```diff
@@ -25,15 +25,30 @@ import (
...
 type PackageRepo struct {
-	Name    string `json:"Name"`
-	BaseUrl string `json:"BaseUrl"`
-	Install bool   `json:"Install"`
+	Name         string `json:"Name"`
+	BaseUrl      string `json:"BaseUrl"`
+	Install      bool   `json:"Install"`
+	GPGCheck     bool   `json:"GPGCheck"`     // Default value is true
+	RepoGPGCheck bool   `json:"RepoGPGCheck"` // Default value is true
+	GPGKeys      string `json:"GPGKeys"`      // Default value is "file:///etc/pki/rpm-gpg/MICROSOFT-RPM-GPG-KEY file:///etc/pki/rpm-gpg/MICROSOFT-METADATA-GPG-KEY"
```
henryli001, line 33: May consider splitting this field into two where one points to rpm gpg check and the other points to repodata gpg check. There might be cases where the users only want to enable one of these two checks instead of both at the same time.

dmcilvaney: I don't see a corresponding field for the .repo files, all the keys just get bundled together for both the metadata and packages. I feel like we should try and stay in line with the behavior from the actual config file here.

As an aside, maybe this should have a key-value entry to support overwriting ANY .repo field...

henryli001 (approved):

PR #5446, opened 2023-05-02
Record performance of tooling using timestamps

trungams:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Enable recording of how long each component in the build tool takes during build process. During each build, a tracker runs in parallel with the main tool being run and records the time taken by each step of the build tool to a jsonl file. Later we can use the data from this file to visualize and further analyze how long each part of the build takes to run.

###### Change Log
- Update makefiles to include timestamp feature
- Add `timestamp.go`: defines data structures for timestamping
- Add `timestamp_mgr.go`: defines interface for recording timestamps in go tools
- Add `bldtracker`: defines interface for recording timestamps in shell scripts
- Add snippets to existing go tools to measure time of build steps

**NO**

###### Test Methodology

- Local build


trungams:
I will extract the timestamp additions to existing go source files to another PR so this one looks less cluttered

toolkit/Makefile
```diff
@@ -82,6 +82,7 @@ STATUS_FLAGS_DIR   ?= $(BUILD_DIR)/make_status
 CHROOT_DIR         ?= $(BUILD_DIR)/worker/chroot
 IMAGEGEN_DIR       ?= $(BUILD_DIR)/imagegen
 TOOLCHAIN_RPMS_DIR ?= $(BUILD_DIR)/toolchain_rpms
+TIMESTAMP_DIR      ?= $(BUILD_DIR)/timestamp
```
mfrw, line 85: This is super timely.

One interesting area as per my investigation is `lock-contention` w.r.t to our `logger`.
We should probably time a full build with `LOG_LEVEL=info` & `LOG_LEVEL=panic`
The above would give us a good pointer for further investigation.

dmcilvaney: I think we might need to tinker with the logging library a bit first. My recollection is that we always log at minimum `debug` level to the log files regardless of what the console output level is.

toolkit/tools/internal/timestamp/timestamp_mgr.go
```diff
@@ -0,0 +1,365 @@
...
+func (readMgr *TimeStampReadManager) buildTreeFromFile(fd *os.File) (err error) {
+	fd.Seek(0, 0)
+	scanner := bufio.NewScanner(fd)
+	for scanner.Scan() {
+		var ts TimeStamp
+		err = json.Unmarshal(scanner.Bytes(), &ts)
+		if err != nil {
+			logger.Log.Warnf("Error reading timestamp object from file %v", err)
+			continue
+		}
```
dmcilvaney, line 352:
```suggestion
		}
		ts.subSteps = make(map[string]*TimeStamp)
```

toolkit/tools/internal/timestamp/timestamp_mgr.go
```diff
@@ -0,0 +1,365 @@
...
+		readMgr.nodes[ts.ID] = &ts
+		if readMgr.root == nil {
+			readMgr.root = &ts
+		} else {
+			readMgr.nodes[ts.ParentID].addSubStep(&ts)
+		}
```
dmcilvaney, line 358: Ran into some nil ptr exceptions while tinkering with this. I think this works as expected now?
```suggestion
		} else if ts.ParentID != -1 {
			readMgr.nodes[ts.ParentID].addSubStep(&ts)
			readMgr.lastVisited = &ts
		}
```

toolkit/tools/internal/timestamp/timestamp_mgr.go
```diff
@@ -0,0 +1,365 @@
...
+		} else {
+			mgr.nodes[ts.ParentID].addSubStep(ts)
+			if mgr.lastVisited.ID == ts.ParentID {
+				mgr.lastVisited = ts
+			}
+		}
+	case EventStop:
+		ts := mgr.nodes[record.ID]
+		ts.complete(record.time)
+		if record.EventType == EventStop && ts.ID == mgr.lastVisited.ID {
```
dmcilvaney, line 334: Not sure if you want to handle this or not, its kind of user error in a way. This can leave dangling substeps without a stop event. i.e. if we start a bunch of steps, then call `CompleteTiming()` without stopping the stub steps first we get an unbalanced set of events.

toolkit/scripts/toolchain_download_timestamp.sh
```diff
@@ -0,0 +1,43 @@
...
+# Separate from toolchain.sh since it needs to handle asynchronous operations
+
+# $1 bldtracker tool
+# $2 out path
+# $3 rpm
+# $4 mode
+
+bldtracker="$1"
+script_name="download toolchain rpms"
+out_file="$2"
```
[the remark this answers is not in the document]
trungams: The existing code doesn't have support for multiple bldtracker programs running concurrently. Adding support means we might have to introduce file locking again.

toolkit/tools/internal/uidutils/uidutils.go
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Utilities for managing unique IDs
+
+package uidutils
```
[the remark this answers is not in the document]
trungams: I moved this to be a part of `TimeStampManager`, and fixed the logic so that we correctly use the next max integer when generating new IDs

toolkit/tools/internal/timestamp/timestamp_mgr.go
```diff
@@ -0,0 +1,365 @@
...
+//    defer timestamp.StopEvent(ts)
+//  }
+//
+//  schedulerTS, _ := timestamp.StartEvent("scheduler", nil   // starts recording "tool/scheduler"
+//  defer timestamp.StopEvent(schedulerTS)                    // stops recording "tool/scheduler" later
+//  for task := range allTasks {
+//    go worker(schedulerTS, task)                            // each worker will add a substep under "tool/scheduler"
+//  }
+
+package timestamp
```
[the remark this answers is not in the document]
trungams: I added `timestamp_test.go` with some test cases covering the basic usage

toolkit/tools/internal/timestamp/timestamp_mgr.go
trungams: Updated

toolkit/tools/internal/timestamp/timestamp_mgr.go
trungams: Do you have steps to replicate this error? With our current way of using timestamp manager I don't think we ever have more than one node without a parent ID

toolkit/tools/internal/timestamp/timestamp_mgr.go
trungams: There will be some incorrectness in the measurement whether or not we handle that. We can either have a check at the end and print out a warning if some start events do not have a corresponding stop events, or we can force stop all existing events when `CompleteTiming()` is called. I think we can go with either way.

toolkit/tools/internal/timestamp/timestamp_mgr.go
dmcilvaney: I created some nodes, saved them to a file, then  reloaded them from file and continued. I don't think I've got the code anymore though, was just a quick hack in a unit test.

toolkit/tools/internal/timestamp/timestamp_mgr.go
dmcilvaney: The original implementation did the second option, but I'm not sure its a big deal either way.

toolkit/scripts/toolchain_download_timestamp.sh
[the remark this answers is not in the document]
dmcilvaney: This script can just disappear completely then, that was the only place it was being used.

toolkit/tools/internal/timestamp/timestamp_test.go
```diff
@@ -0,0 +1,215 @@
+// Copyright (c) Microsoft Corporation.
```
dmcilvaney, line 1: Looks like `StopEventByPath` `PauseEvent` and `ResumeEvent` aren't covered here.

dmcilvaney: (From https://github.com/microsoft/CBL-Mariner/actions/runs/5063082281)

dmcilvaney (approved):

Issue #5460, opened 2023-05-05
crashdump memory is reserved even if kdump (kexec-tools package) is not installed/enabled


ddstreet:
The crashkernel= boot parameter causes the kernel to reserve a specific amount of memory. Currently this parameter is included in the Mariner kernel spec file, so all Mariner deployments reserve memory for a kdump kernel, on all boots. For systems where kexec-tools is not installed, no kdump kernel is loaded into the reserved memory, resulting in a loss of usable memory.

Camelron:
This is addressed for 3.0 in the following commit: https://github.com/microsoft/CBL-Mariner/pull/6281

It will apply to all 3.0 images once we deprecate and remove the legacy boot.cfg and go full grub2-mkconfig by default. I'll leave this issue open until this one resolves. 

For 2.0, we want to keep the crashkernel allocation in the cmdline as we cannot safely ship its removal without causing other customers who _do_ use kexec-tools.service to go red.

Malateshk007:
@Camelron has already provided an update on this matter, and I will proceed with closing this issue. Thank you for your engagement and please feel free to reach out if further discussion is needed.

PR #5462, opened 2023-05-05
Move crashkernel= parameter from kernel rpm to kexec-tools rpm

ddstreet:
Fixes: #5460

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Move crashkernel= parameter from kernel rpm to kexec-tools rpm to solve #5460 

###### Change Log  

Move crashkernel= parameter from kernel rpm to kexec-tools rpm

YES

###### Associated issues  

- #5460

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/SPECS-SIGNED/kernel-signed/kernel-signed.spec
+++ b/SPECS-SIGNED/kernel-signed/kernel-signed.spec
@@ -10,7 +10,7 @@
 Summary:        Signed Linux Kernel for %{buildarch} systems
 Name:           kernel-signed-%{buildarch}
 Version:        5.15.112.1
-Release:        2%{?dist}
+Release:        3%{?dist}
 License:        GPLv2
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -153,6 +153,9 @@ ln -sf linux-%{uname_r}.cfg /boot/mariner.cfg
 %exclude /module_info.ld
 
 %changelog
+* Wed Jun 07 2023 Dan Streetman <ddstreet@ieee.org> - 5.15.112.1-3
+- Bump release to match kernel
+
 * Wed May 24 2023 Rachel Menge <rachelmenge@microsoft.com> - 5.15.112.1-2
 - Bump release to match kernel
 
--- a/SPECS/kernel-headers/kernel-headers.spec
+++ b/SPECS/kernel-headers/kernel-headers.spec
@@ -1,7 +1,7 @@
 Summary:        Linux API header files
 Name:           kernel-headers
 Version:        5.15.112.1
-Release:        2%{?dist}
+Release:        3%{?dist}
 License:        GPLv2
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -36,6 +36,9 @@ cp -rv usr/include/* /%{buildroot}%{_includedir}
 %{_includedir}/*
 
 %changelog
+* Wed Jun 07 2023 Dan Streetman <ddstreet@ieee.org> - 5.15.112.1-3
+- Bump release to match kernel
+
 * Wed May 24 2023 Rachel Menge <rachelmenge@microsoft.com> - 5.15.112.1-2
 - Bump release to match kernel
 
--- a/SPECS/kernel/kernel.spec
+++ b/SPECS/kernel/kernel.spec
@@ -28,7 +28,7 @@
 Summary:        Linux Kernel
 Name:           kernel
 Version:        5.15.112.1
-Release:        2%{?dist}
+Release:        3%{?dist}
 License:        GPLv2
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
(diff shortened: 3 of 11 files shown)


dmcilvaney (approved):

oliviacrain (approved):

christopherco (changes requested):
Other than the kexec-tools signature file regeneration, rest LGTM

rlmenge (approved):
LGTM (once signatures are updated)

ddstreet:
Unfortunately, this isn't a very good approach due to our current image build process.

In Mariner, the `/boot/grub2/grub.cfg` file isn't actually owned by any package; it's generated by the image build tooling. This unfortunately means we have no (good) way to deliver a fix to this file for existing end users.

This prevents this PR from being mergable as it is now because, if it was, then end users who upgraded to the new kernel package would have the `crashkernel=` parameter removed from their `/boot/mariner.cfg` file, and (if they installed or upgraded to the new kexec-tools package) the `crashkernel=` parameter would be added to the new `/boot/kdump.cfg` file. However, since we have no way to update the `grub.cfg` file to include the new configuration from the `kdump.cfg` file,  their system would no longer include the `crashkernel=` parameter and they would not be able to capture any kernel crashdumps.

I don't think we can proceed with this PR until we change from our current design of generating the grub.cfg from our image build tooling over to properly using the standard `grub-mkconfig` tooling to generate the grub configuration based on the standard grub config in `/etc`. And even then, we need to carefully consider if existing users would be impacted by the change.

dmcilvaney:
> Unfortunately, this isn't a very good approach due to our current image build process.
> 
> In Mariner, the `/boot/grub2/grub.cfg` file isn't actually owned by any package; it's generated by the image build tooling. This unfortunately means we have no (good) way to deliver a fix to this file for existing end users.
> 
> This prevents this PR from being mergable as it is now because, if it was, then end users who upgraded to the new kernel package would have the `crashkernel=` parameter removed from their `/boot/mariner.cfg` file, and (if they installed or upgraded to the new kexec-tools package) the `crashkernel=` parameter would be added to the new `/boot/kdump.cfg` file. However, since we have no way to update the `grub.cfg` file to include the new configuration from the `kdump.cfg` file, their system would no longer include the `crashkernel=` parameter and they would not be able to capture any kernel crashdumps.
> 
> I don't think we can proceed with this PR until we change from our current design of generating the grub.cfg from our image build tooling over to properly using the standard `grub-mkconfig` tooling to generate the grub configuration based on the standard grub config in `/etc`. And even then, we need to carefully consider if existing users would be impacted by the change.

Can we add some logic to kernel.cfg to check for the presence of this file as a stopgap until m3, and use today's default otherwise?

```bash
# Inside kernel.cfg
if [ -f /boot/kdump.cfg ]
    load_env -f /boot/kdump.cfg
else
    kdump_cfg="crashkernel=256M"
fi

mariner_cmdline=init=/lib/systemd/systemd ro loglevel=3 no-vmw-sta $kdump_cfg
...

```

ddstreet:
> > Unfortunately, this isn't a very good approach due to our current image build process.
> > In Mariner, the `/boot/grub2/grub.cfg` file isn't actually owned by any package; it's generated by the image build tooling. This unfortunately means we have no (good) way to deliver a fix to this file for existing end users.
> > This prevents this PR from being mergable as it is now because, if it was, then end users who upgraded to the new kernel package would have the `crashkernel=` parameter removed from their `/boot/mariner.cfg` file, and (if they installed or upgraded to the new kexec-tools package) the `crashkernel=` parameter would be added to the new `/boot/kdump.cfg` file. However, since we have no way to update the `grub.cfg` file to include the new configuration from the `kdump.cfg` file, their system would no longer include the `crashkernel=` parameter and they would not be able to capture any kernel crashdumps.
> > I don't think we can proceed with this PR until we change from our current design of generating the grub.cfg from our image build tooling over to properly using the standard `grub-mkconfig` tooling to generate the grub configuration based on the standard grub config in `/etc`. And even then, we need to carefully consider if existing users would be impacted by the change.
> 
> Can we add some logic to kernel.cfg to check for the presence of this file as a stopgap until m3, and use today's default otherwise?
> 
> ```shell
> # Inside kernel.cfg
> if [ -f /boot/kdump.cfg ]
>     load_env -f /boot/kdump.cfg
> else
>     kdump_cfg="crashkernel=256M"
> fi
> 
> mariner_cmdline=init=/lib/systemd/systemd ro loglevel=3 no-vmw-sta $kdump_cfg
> ...
> ```

IMHO we shouldn't further hack things. We should do things right using grub-mkconfig and this PR should wait until that's done.

ddstreetmicrosoft:
Closing this in favor of the work @Camelron has been doing to enable grub2-mkconfig and move the crashkernel= parameter to the kexec-tools package.

PR #5465, opened 2023-05-06
Add timestamp code to existing tooling

trungams:
- Use the timestamp feature introduced in #5446 to start measuring runtime of certain functions in our build tools. This change enables build tools to start recording timestamps when they begin/complete some steps during the build process. The final results are stored in jsonl files and can be converted to other format for further analysis


christopherco:
Haven't reviewed yet, but what is the general thinking on piecemeal incorporating this feature into `main`?

dmcilvaney (approved):

PR #5476, opened 2023-05-09
Add ldap backend support into sudo

AZaugg:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Add the config option to sudo build to allow configuration of sudo via LDAP.

###### Change Log  

- Add openldap to buildrequires and adding --with-ldap option to sudo SPEC file

**NO**

###### Test Methodology

- Updated SPEC file with config option
- Ran a rebuild of the package 
```sudo  make build-packages -j$(nproc) REBUILD_TOOLS=y SRPM_PACK_LIST="sudo" PACKAGE_REBUILD_LIST="sudo" REFRESH_WORKER_CHROOT=n
```

### DIFF
--- a/SPECS/sudo/sudo.spec
+++ b/SPECS/sudo/sudo.spec
@@ -15,7 +15,7 @@ BuildRequires:  openssl-libs
 BuildRequires:  pam-devel
 BuildRequires:  sed
 BuildRequires:  zlib-devel
-BuildRequires:  openldap
+BuildRequires:  openldap-devel
 Requires:       audit-libs
 Requires:       openssl-libs
 Requires:       pam
@@ -38,7 +38,6 @@ the ability to run some (or all) commands as root or another user while logging 
     --with-env-editor \
     --with-pam \
     --with-ldap \
-    --with-ldap-conf-file=/etc/sudo-ldap.conf \
     --with-linux-audit \
     --enable-zlib=system \
     --with-passprompt="[sudo] password for %p: "


SPECS/sudo/sudo.spec
```diff
@@ -36,6 +37,8 @@ the ability to run some (or all) commands as root or another user while logging
     --docdir=%{_docdir}/%{name}-%{version} \
     --with-env-editor \
     --with-pam \
+    --with-ldap \
+    --with-ldap-conf-file=/etc/sudo-ldap.conf \
```
ddstreet, line 41: is there a reason/need for defaulting to a separate ldap conf file here? Or just copying how Debian/Ubuntu does it?

AZaugg: I essentially followed what was on the man page [here](https://linux.die.net/man/5/sudoers.ldap).

AZaugg:
> Can you provide some detail about how you actually tested this to make sure it works and does what you expect?

I added the options to the sudo.spec file, and built out the RPM, i then upgraded the local RPM and ensured the the config options were set:

```
azaugg@azaugg-ld77 [ ~ ]$ rpm -qa|grep sudo
sudo-1.9.13p3-1.cm2.x86_64
azaugg@azaugg-ld77 [ ~ ]$ sudo sudo -V
Sudo version 1.9.13p3
Configure options: --prefix=/usr --bindir=/usr/bin --libexecdir=/usr/lib --docdir=/usr/share/doc/sudo-1.9.13p3 --with-env-editor --with-pam --with-linux-audit --enable-zlib=system --with-passprompt=[sudo] password for %p:
```

Now after upgrade

```
azaugg@azaugg-ld77 [ ~ ]$ sudo rpm -U sudo-1.9.13p3-2.cm2.x86_64.rpm
/sbin/ldconfig: /lib/libwbclient.so.0 is not a symbolic link

/sbin/ldconfig: /lib/libwbclient.so.0 is not a symbolic link

azaugg@azaugg-ld77 [ ~ ]$ rpm -qa|grep sudo
sudo-1.9.13p3-2.cm2.x86_64
azaugg@azaugg-ld77 [ ~ ]$ sudo sudo -V
Sudo version 1.9.13p3
Configure options: --prefix=/usr --bindir=/usr/bin --libexecdir=/usr/lib --docdir=/usr/share/doc/sudo-1.9.13p3 --with-env-editor --with-pam --with-ldap --with-linux-audit --enable-zlib=system --with-passprompt=[sudo] password for %p:
```

My other colleague took the RPM and configured the ldap server and was able to bind against the directory server and control sudo via it

SPECS/sudo/sudo.spec
trungams: The man page also mentions this
> Also note that on systems using the OpenLDAP libraries, default values specified in /etc/openldap/ldap.conf or the user's .ldaprc files are not used.
so I think leaving it to the current path is fine.

trungams (approved):
Looks good to me!

SPECS/sudo/sudo.spec
```diff
@@ -15,6 +15,7 @@ BuildRequires:  openssl-libs
 BuildRequires:  pam-devel
 BuildRequires:  sed
 BuildRequires:  zlib-devel
+BuildRequires:  openldap
```
trungams, line 18: The documentation suggests `openldap-devel`: https://www.sudo.ws/docs/install/#linux

SPECS/sudo/sudo.spec
trungams: There seems to be a slight difference between the man page you linked and the official documentation: https://www.sudo.ws/docs/man/1.9.13/sudoers.ldap.man/#Configuring_ldap.conf. The default value is `/etc/ldap.conf` and it's shared among other ldap clients. Unless users have specific requirements for using sudo with ldap, they don't need to create a separate config file. Let's remove that parameter and leave it to default to `/etc/ldap.conf`

trungams (changes requested):

trungams:
@AZaugg I added some comments. In addition, could you also provide some information on how testing was done to verify sudo works fine with ldap? An example `/etc/ldap.conf` and a brief summary of the steps taken to get sudo working with openldap would be good.

SPECS/sudo/sudo.spec
AZaugg: Looking at sudo build options when enabling ldap, it requires access to iber.h and ldap.h, I took a look at the mariner-supplied packages, both these headerfiles are provided in openldap

```
azaugg@azaugg-ld77 [ ~/CBL-Mariner/SPECS/sudo ]$ sudo repoquery -l openldap|grep inc
Last metadata expiration check: 0:01:52 ago on Thu May 18 15:26:41 2023.
- Fix CVE-2022-29155
/usr/include/lber.h
/usr/include/lber_types.h
/usr/include/ldap.h
/usr/include/ldap_cdefs.h
/usr/include/ldap_features.h
/usr/include/ldap_schema.h
/usr/include/ldap_utf8.h
/usr/include/ldif.h
/usr/include/openldap.h
/usr/include/slapi-plugin.h
```

AZaugg: Hmm, come to think of it, I don't think Mariner ships with a devel package for openldap
```
azaugg@azaugg-ld77 [ ~/CBL-Mariner/SPECS/sudo ]$ sudo yum info openldap-devel
[sudo] password for azaugg:
[using capability match for 'openldap-devel']
Name          : openldap
Arch          : x86_64
Epoch         : 0
Version       : 2.4.57
Release       : 8.cm2
Install Size  :   1.65M 1732185 (1732185)
Repo          : @System
...
...
...
```

SPECS/sudo/sudo.spec
AZaugg: Sounds good.

SPECS/sudo/sudo.spec
trungams: Hmm you are right. Our [openldap spec file](https://github.com/microsoft/CBL-Mariner/blob/main/SPECS/openldap/openldap.spec) doesn't specify a devel package. Let me check back on this

oliviacrain: I would still use the `openldap-devel` name that's currently attached to the main package. It makes less work for us when we inevitably split the package properly in a future release.

AZaugg: Done, referenced to openldap-devel. I do want to point out that if a devel package is needed, we must have the includes iber.h and ldap.h in the devel file

trungams (approved):

ddstreet:
@trungams sounds like your investigation into openldap conf shows the separate conf file is not needed (by default), and from @AZaugg comment it sounds like the actual testing with the modified rpm that defaults to the normal ldap conf file does work (if i'm reading the comment right), so it all LGTM, thanks!

ddstreet (approved):
as mentioned in last comment, LGTM

PR #5480, opened 2023-05-09
Revert "Promote containernetworking-plugins from extended to core"

BettyRain:
Reverts microsoft/CBL-Mariner#5471


christopherco (approved):
Thanks for posting this - we're still evaluating whether our current CNI / CNI-plugins package is sufficient.

hbeberman (approved):

PR #5481, opened 2023-05-09
Promote containernetworking-plugins from extended to core

BettyRain:
Promote containernetworking-plugins from extended to core: https://github.com/microsoft/CBL-Mariner/pull/5471


christopherco (changes requested):
Requesting to see if our `cni-plugins` and `cni` packages could be used instead.

PR #5497, opened 2023-05-12
Add mstflint Driver to the kernel-hci

ellie-di:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR adds the mstflint kernel driver for secure boot as a patch into the kernel-hci source tree.

**NO**

###### Test Methodology

- Pipeline build id: [359641](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=359641&view=results)
- Upgraded kernel with kernel-hci-5.15.110.1-2.cm2.x86_64.rpm, disabled the secure boot and checked the module is loaded.
image

### DIFF
--- a/SPECS/kernel-hci/kernel-hci.spec
+++ b/SPECS/kernel-hci/kernel-hci.spec
@@ -3,9 +3,9 @@
 %define uname_r %{version}-%{release}
 
 # find_debuginfo.sh arguments are set by default in rpm's macros.
-# The default arguments regenerate the build-id for vmlinux in the 
+# The default arguments regenerate the build-id for vmlinux in the
 # debuginfo package causing a mismatch with the build-id for vmlinuz in
-# the kernel package. Therefore, explicilty set the relevant default 
+# the kernel package. Therefore, explicilty set the relevant default
 # settings to prevent this behavior.
 %undefine _unique_build_ids
 %undefine _unique_debug_names
@@ -15,11 +15,10 @@
 %define arch x86_64
 %define archdir x86
 %define config_source %{SOURCE1}
-
 Summary:        Linux Kernel for HCI
 Name:           kernel-hci
 Version:        5.15.110.1
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv2
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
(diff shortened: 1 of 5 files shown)


SPECS/kernel-hci/kernel-hci.spec
```diff
@@ -247,7 +246,7 @@ for MODULE in `find %{buildroot}/lib/modules/%{uname_r} -name *.ko` ; do \
 %define __spec_install_post\
     %{?__debug_package:%{__debug_install_post}}\
     %{__arch_install_post}\
-    %{__os_install_post}\
+    %__os_install_post\
```
jslobodzian, line 249: is removing the { } the right call here?

jslobodzian (approved):

SPECS/kernel-hci/kernel-hci.spec
```diff
@@ -68,15 +59,22 @@ BuildRequires:  libmspack-devel
...
 Requires(post): coreutils
 Requires(postun): coreutils
 ExclusiveArch:  x86_64
+# find_debuginfo.sh arguments are set by default in rpm's macros.
+# The default arguments regenerate the build-id for vmlinux in the
+# debuginfo package causing a mismatch with the build-id for vmlinuz in
+# the kernel package. Therefore, explicilty set the relevant default
+# settings to prevent this behavior.
+%undefine _unique_build_ids
+%undefine _unique_debug_names
```
rlmenge, line 77: My guess is you moved these because of the linter. These are actually tied to the other globals

%global _missing_build_ids_terminate_build 1
%global _no_recompute_build_ids 1

Feel free to ignore the linter's recommendations for the kernel.

SPECS/kernel-hci/kernel-hci.spec
ellie-di: I ran the spec linter and that removed it. I can put it back to be consistent with the above line.

romoh (approved):
curious, what is the size increase with that change?

christopherco (approved):

SPECS/kernel-hci/kernel-hci.spec
christopherco: @jslobodzian this one is a long-standing issue with the spec-linter. We lived with it since the folks changing the kernels were a very limited few who knew the quirks, but we're starting to scale this out, so things like this will start causing more problems.
If we have a backlog for issues about the spec-linter, we should add a note there to tackle when we have time. If not, we probably should start one so we can get the spec-linter into a reliable state.

ellie-di:
> curious, what is the size increase with that change?

Not much, I compared the kernel-hci version 5.15.110 with 5.15.107 and the diff was ~11KB increase.

PR #5505, opened 2023-05-15
Add rootfs partition name in gen2 market place image

miz060:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add rootfs partition name in gen2 market place image to enable kata image's grub.cfg to find mshv-bootloader file.

###### Change Log  

- Add rootfs partition name in amd64 gen2 market place image 
- Add rootfs partition name in aarch64 gen2 market place image (for future images only)

**NO**

###### Test Methodology

- Manual test done with https://dev.azure.com/mariner-org/mariner/_build/results?buildId=361715&view=results

### DIFF
--- a/toolkit/imageconfigs/marketplace-gen2.json
+++ b/toolkit/imageconfigs/marketplace-gen2.json
@@ -28,6 +28,7 @@
                 },
                 {
                     "ID": "rootfs",
+                    "Name": "rootfs",
                     "Start": 565,
                     "End": 0,
                     "FsType": "ext4"
@@ -51,7 +52,8 @@
                 },
                 {
                     "ID": "rootfs",
-                    "MountPoint": "/"
+                    "MountPoint": "/",
+                    "MountIdentifier": "partlabel"
                 }
             ],
             "PackageLists": [
--- a/toolkit/imageconfigs/marketplace-gen2.json
+++ b/toolkit/imageconfigs/marketplace-gen2.json
@@ -52,8 +52,7 @@
                 },
                 {
                     "ID": "rootfs",
-                    "MountPoint": "/",
-                    "MountIdentifier": "partlabel"
+                    "MountPoint": "/"
                 }
             ],
             "PackageLists": [
--- a/toolkit/imageconfigs/marketplace-gen2-aarch64.json
+++ b/toolkit/imageconfigs/marketplace-gen2-aarch64.json
@@ -28,6 +28,7 @@
                 },
                 {
                     "ID": "rootfs",
+                    "Name": "rootfs",
                     "Start": 565,
                     "End": 0,
                     "FsType": "ext4"
@@ -51,7 +52,8 @@
                 },
                 {
                     "ID": "rootfs",
-                    "MountPoint": "/"
+                    "MountPoint": "/",
+                    "MountIdentifier": "partlabel"
                 }
             ],
             "PackageLists": [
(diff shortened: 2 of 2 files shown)


Camelron:
lgtm

Camelron (approved):

christopherco (changes requested):
This change has the potential to impact all of our Azure VM use cases, so I'd like to make sure this is fully reviewed and tested before merging. Requesting a few reviews from a few additional devs.

And can we get run results for this change through LISAv3/Penguinator?

toolkit/imageconfigs/marketplace-gen2.json
```diff
@@ -51,7 +52,8 @@
                 },
                 {
                     "ID": "rootfs",
-                    "MountPoint": "/"
+                    "MountPoint": "/",
+                    "MountIdentifier": "partlabel"
```
hbeberman, line 56: One thing to double check here is if cloud-init growpart still functions with this change. In the past we've seen growpart fail because it couldn't identify the root partition when specified with PARTLABEL.

Please create an Azure VM with this new image and an explicit "--os-disk-size-gb 100" and make sure that the rootfs has been automatically expanded to fill all of that disk space.

hbeberman (changes requested):

dmcilvaney:
I'm wondering if we should split this into a dedicated image definition. While it might be unlikely anyone has taken a depenency on the exact formatting of the marketplace fstab file I'm leery of changing it mid-release.

Two options off the top of my head:
1) We split a dedicated base image with this change and take the hit in infrastructure needed to build/test/release it.
2) Take just the `"name": ...` change (which will make the partition labelled) but make the change to the fstab/boot arg in Packer. 

I'm leaning towards 2) for now until we do our next major release where we can switch stuff up.

toolkit/imageconfigs/marketplace-gen2.json
miz060: I tested with "--os-disk-size-gb 100" and the rootfs was indeed automatically expanded to fill all of the disk space.

But as Daniel suggested below, I figured out a way to make kata's grub.cfg work with only partition name and no partlabel. So I just reverted the partlabel change,

miz060:
> I'm wondering if we should split this into a dedicated image definition. While it might be unlikely anyone has taken a depenency on the exact formatting of the marketplace fstab file I'm leery of changing it mid-release.
> 
> Two options off the top of my head:
> 
> 1. We split a dedicated base image with this change and take the hit in infrastructure needed to build/test/release it.
> 2. Take just the `"name": ...` change (which will make the partition labelled) but make the change to the fstab/boot arg in Packer.
> 
> I'm leaning towards 2) for now until we do our next major release where we can switch stuff up.

Thanks for the suggestion Daniel! I figured out the packer change to make things work with only the partition name and no partlabel.

christopherco (changes requested):
Changes LGTM. Just looking for Penguinator results for these latest changes to be added.

miz060:
> Changes LGTM. Just looking for Penguinator results for these latest changes to be added.

This is the LISAv3 passing run with this change:
https://penguinatorstarter.powerappsportals.com/mytests/myjobs/testdetails/?id=ff58d86f-ccf5-ed11-a81c-6045bdb1047c

christopherco (approved):

hbeberman (approved):

PR #5522, opened 2023-05-17
Dropping package requirement of GeoIP-data in GeoIP

AZaugg:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

GeoIP is a package that can no longer be cleanly installed due to geoip-data no longer being available, dropping that package requirement as a hard requirement.
###### Change Log  

- Remove require on GeoIP

**NO**

###### Test Methodology

- Built new RPM
- Installed new RPM 
- Ran tool manually


SPECS-EXTENDED/GeoIP/GeoIP.spec
```diff
@@ -15,7 +15,6 @@ BuildRequires:	gcc
 BuildRequires:	make
 BuildRequires:	sed
 BuildRequires:	zlib-devel
-Requires:	GeoIP-data
```
mfrw, line 18: I was wondering if we can use `GeoIp-Lite` in here. 
As per my initial research, it seems the above package supersedes this package as per [this](https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS-EXTENDED/GeoIP-GeoLite-data/GeoIP-GeoLite-data.spec#L23)

AZaugg: I can do. My opinion was for it to not have any hard dependencies and let users feed in a data source

AZaugg: So, I don't think this will work. Looking at the Mariner upstream [repos](https://packages.microsoft.com/cbl-mariner/2.0/prod/extended/x86_64/Packages/g/), i don't see GeoIp-Lite. I do see Geolite-[asn|city|country]. Which is provided by the [spec file ](https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS-EXTENDED/geolite2/geolite2.spec). Do you want me to add

```
Requires GeoIP-data | GeoIp-Lite | Geolite-[asn|city|country]
```

Instead?

oliviacrain: If the library is functional with a user-provided data source, then I think we're the clear to remove the requirement on a data package and consider adding a `Recommends` link to `Geolite` data at a later date. Let's defer that decision to unblock @AZaugg.

oliviacrain (approved):

SPECS-EXTENDED/GeoIP/GeoIP.spec
mfrw: Apologies for the late response - I was out on vacation. I concur with @oliviacrain.
Thanks OC

PR #5524, opened 2023-05-17
[main] update libdrm to 2.4.114

hideyukn88:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

This PR updates libdrm to version 2.4.114. This is required to update Mesa to newer version with GPU selection fixes.

###### Change Log  

- Update libdrm to 2.4.114
- Port SPEC file from Fedora 39
- Merge applicable SPEC file changes previously made for Mariner.
 - This change impacts libdrm, libdrm-devel, libdrm-debuginfo and drm-utils packages

**NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- local build


SPECS/libdrm/libdrm.spec
```diff
@@ -207,125 +199,144 @@ mkdir -p %{buildroot}%{_docdir}/libdrm
...
-
 %if %{with etnaviv}
 %{_includedir}/libdrm/etnaviv_*.h
 %{_libdir}/libdrm_etnaviv.so
 %{_libdir}/pkgconfig/libdrm_etnaviv.pc
 %endif
-
+%{_includedir}/libsync.h
+%{_includedir}/xf86drm.h
+%{_includedir}/xf86drmMode.h
```
PawelWMS, line 265: Could we move these 3 headers next to the other ones at the beginning of `%files devel`? It makes it easier to read, when they are together. In this case that would give us the additional benefit that they wouldn't be squeezed in between the `%if %{with X}` sections - this way it's clearer that they are always part of the `devel` subpackage.

SPECS/libdrm/libdrm.spec
```diff
@@ -207,125 +199,144 @@ mkdir -p %{buildroot}%{_docdir}/libdrm
...
 %files -n drm-utils
+%if %{with amdgpu}
+%{_bindir}/amdgpu_stress
+%endif
 %{_bindir}/drmdevice
+%if %{with etnaviv}
 %exclude %{_bindir}/etnaviv_*
+%endif
+%if %{with exynos}
 %exclude %{_bindir}/exynos_*
```
PawelWMS, line 281: This looks fishy. If we're building with `etnaviv` or `exynos`, then why would we want to exclude their binaries in such cases? Should we use the `%{without X}` macro instead?

SPECS/libdrm/libdrm.spec
```diff
@@ -207,125 +199,144 @@ mkdir -p %{buildroot}%{_docdir}/libdrm
...
-%{_bindir}/kms-steal-crtc
-%{_bindir}/kms-universal-planes
-
-%if %{with libkms}
-%{_bindir}/kmstest
 %endif
-
+%if %{with tegra}
+%exclude %{_bindir}/tegra-*
+%endif
```
PawelWMS, line 285: Similar question about excluding the binary when building with `tegra`.

SPECS/libdrm/libdrm.spec
```diff
@@ -207,125 +199,144 @@ mkdir -p %{buildroot}%{_docdir}/libdrm
...
-- Removing BR on 'marinerui-rpm-macros'. Using macros from the build env.
-
-* Fri Dec 18 2020 Pawel Winogrodzki <pawelwi@microsoft.com> - 2.4.102-3
-- Initial CBL-Mariner import from Fedora 33 (license: MIT).
 - Added an explicit 'pkgconfig(libdrm)' for 'libdrm-devel'.
-- Added build-time dependency on 'marinerui-rpm-macros'.
 - Added the "LICENSE.PTR" file.
 - Removed manual pages.
 - Replaced ldconfig scriptlets with explicit calls to ldconfig.
 - Removed sections related to %%bcond_with for 'cairo_tests' and 'freedreno_kgsl'.
```
PawelWMS, line 301: For changes to a spec we already had in our repo, please leave the old changelog entries and just mention that for your update you've used Fedora X for guidance. Look for `spec (license: MIT) for guidance` across the specs to see what I have in mind.

PawelWMS (changes requested):
Looks pretty good, just a few comments and questions.

SPECS/libdrm/libdrm.spec
hideyukn88: Updated.

SPECS/libdrm/libdrm.spec
hideyukn88: That's how done in Fedora committed at https://src.fedoraproject.org/rpms/libdrm/c/c883cf30d9a67e4bca04425abb71bc8a4ca36dd9

SPECS/libdrm/libdrm.spec
hideyukn88: Same as above.

SPECS/libdrm/libdrm.spec
hideyukn88: Yes, this is what I wonder, since Mariner's change after fork from original Fedora, has own history and timeline/version can't be sorted/aligned. Do you mean I don't need to incorporate the new history (below line 304 ~ 339) from Fedora but just add that one line saying which version of Fedora used for guidance?

hideyukn88: Updated

SPECS/libdrm/libdrm.spec
PawelWMS: From that link it looks like Fedora **includes** the files, while with the current spec we **exclude** them: Fedora doesn't use the `%exclude` macro, while we do. I think we still need a bit more clean-up here.

PawelWMS: Nevermind, I think I see your point now. Looks good.

hideyukn88: I'm not sure we are seeing same things, but it is "exclude", please check line 275 and 276 at that commit. Also, for at current rawhide, see https://src.fedoraproject.org/rpms/libdrm/blob/rawhide/f/libdrm.spec#_273, https://src.fedoraproject.org/rpms/libdrm/blob/rawhide/f/libdrm.spec#_276 and https://src.fedoraproject.org/rpms/libdrm/blob/rawhide/f/libdrm.spec#_279.

SPECS/libdrm/libdrm.spec
PawelWMS: For `tegra` it looks like Fedora doesn't explicitly exclude it anywhere. Could you only double-check, that we even need this line and the package builds on ARM64? I've started a [buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=370841&view=results) for you to verify that.

PawelWMS (approved):
Looks good, please only verify the ARM64 build.

SPECS/libdrm/libdrm.spec
hideyukn88: Do you mean https://src.fedoraproject.org/rpms/libdrm/blob/rawhide/f/libdrm.spec#_279 ?

PR #5528, opened 2023-05-18
Remove umask handling from bash.spec and change it in filesystem.spec

tobiasb-ms:
Refactor how we configure `umask`.

###### Merge Checklist  
- [x] Ready to merge

---

###### Summary
CIS recommends setting `umask` to `027` in `/etc/profile` and `/etc/bashrc` to ensure that files created by users are not group- or world-writable. This PR refactors how we configure `umask` to make it easier to change the value in the future. Note that we already set `umask` to `027` but we do it in a way that is not easily discoverable, and the CIS audit script trips on it.

Before this change, `umask` was configured thusly:
1. In `filesystem.spec`, we set `umask` to `027` in `/etc/profile` *after* sourcing everything in `/etc/profile.d`. This means any scripts in `/etc/profile.d` that set `umask` will be overridden.
2. In `bash.spec`, we create a file `/etc/profile.d/umask.sh` that sets `umask`, but to values we don't want. But this is then overridden by item 1.

While this results in a value of `umask` that we want, it is not very discoverable or idiomatic.

###### Change Log
- Have `bash.spec` take a direct dependency on `filesystem`. In practice, `bash.spec` was never built without `filesystem` anyway, so this is essentially a no-op.
- Remove `/etc/profile.d/umask.sh` from `bash.spec`. This was being ovewritten by `/etc/profile` anyway, so this is another no-op.
- Remove the `umask` setting from `/etc/profile` in `filesystem.spec`, moving it to `/etc/profile.d/99-umask.sh`. This appears to be a more idiomatic way to do this and allows better customization for customers, but is still a no-op.

**NO**

###### Test Methodology
Validated locally that the build rpms had the correct files.
Updated these packages on a marketplace VM and verified everything was correct.
Buddy Builds:
- Buddy [AMD64](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=365176&view=results)
- Buddy [ARM64](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=365178&view=results)
- Toolchain: [AMD64](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=363164&view=results)
- Toolchain: [ARM64](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=363165&view=results)

### DIFF
--- a/SPECS/bash/bash.spec
+++ b/SPECS/bash/bash.spec
@@ -1,7 +1,7 @@
 Summary:        Bourne-Again SHell
 Name:           bash
 Version:        5.1.8
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv3
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -16,6 +16,7 @@ Requires(post): /bin/cp
 Requires(post): /bin/grep
 Requires(postun): /bin/grep
 Requires(postun): /bin/mv
+Requires:       filesystem
 Provides:       /bin/sh
 Provides:       /bin/bash
 Provides:       %{_bindir}/sh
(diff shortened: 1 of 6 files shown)


dmcilvaney (approved):
Oh man does the linter not like filesystem.spec... the bsah suggestion is fairly benign so maybe do grab that one though (alphabetical re-ordering of the new Requires, github won't let me do an inline suggestion for it unfortunately).

hbeberman (approved):

tobiasb-ms:
> Oh man does the linter not like filesystem.spec... the bsah suggestion is fairly benign so maybe do grab that one though (alphabetical re-ordering of the new Requires, github won't let me do an inline suggestion for it unfortunately).

I'd rather do that as a separate change so-as not to confuse the issue I'm solving here.

invidian:
I have a question since I recently run into issues related to `umask`. Why default umask isn't changed via `/etc/login.defs` file, like in other distros? This is where pam_umask.so may read it from, as `/etc/profile.d` may not be read in all cases.

PR #5530, opened 2023-05-18
Add workflow to check for required kernel configs

rlmenge:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

In an effort to reduce relying on my memory and to ensure that critical customer configs are not modified, add a new PR check which checks if
1. any configs with known required values have been changed to undesired values/removed
2. new configs are being added and have not added to the json of required kernel configs

Required configs are currently only set for the "kernel" spec but other kernels can created fields in the json to account for their configs.

For case 1, a failure looks like
```
===============================================================================
== Results for SPECS/kernel/config ==
===============================================================================

----------------- Kernel config verification FAILED -----------------
CONFIG_DRM_AMDGPU_SI is "", expected ['y'].
Reason: Needed for customer
PR: https://github.com/microsoft/CBL-Mariner/pull/54[16](https://github.com/rlmenge/CBL-Mariner/actions/runs/5010776690/jobs/8980970694?pr=30#step:9:17)

CONFIG_TLS is " is not set", expected ['m', 'y'].
Reason: Needed for customer
PR: https://github.com/microsoft/CBL-Mariner/pull/5406

CONFIG_X86_PLATFORM_DRIVERS_DELL is "m", expected ['y'].
Reason: Needed for customer
PR: https://github.com/microsoft/CBL-Mariner/pull/[26](https://github.com/rlmenge/CBL-Mariner/actions/runs/5010776690/jobs/8980970694?pr=30#step:9:27)71

CONFIG_INIT_ON_FREE_DEFAULT_ON is "y", expected ['is not set', ''].
Reason: Creates increased boot times and errors on large memory systems
PR: https://github.com/microsoft/CBL-Mariner/pull/48[29](https://github.com/rlmenge/CBL-Mariner/actions/runs/5010776690/jobs/8980970694?pr=30#step:9:30)
```

For case 2 a failure looks like
```
====================== Kernel new config verification FAILED for ARM64 ======================
New configs detected for ARM64. Please add the following to toolkit/scripts/mariner-required-configs.json
CONFIG_CRYPTO_ECDSA
```
There is also a verbose flag when checking required configs  (1)
```
===============================================================================
== Results for /home/rachel/repos/CBL-Mariner-main/SPECS/kernel/config_aarch64 ==
===============================================================================
Option                          Required Arch       Expected Value      Comment  
-------------------------------------------------------------------------------
CONFIG_INIT_ON_FREE_DEFAULT_ON  ['AMD64', 'ARM64']  ['is not set', '']  OK       
CONFIG_CC_CAN_LINK_STATIC       ['AMD64', 'ARM64']  ['']                OK       
CONFIG_SECURITY_SMACK           ['AMD64', 'ARM64']  ['is not set', '']  OK       
CONFIG_BPF_UNPRIV_DEFAULT_OFF   ['AMD64', 'ARM64']  ['y']               OK     
```

###### Change Log  

- Create new workflow to check for critical configs

**NO**

###### Test Methodology

- PR which tests new config: https://github.com/rlmenge/CBL-Mariner/pull/29
- PR which tests modifying required config: https://github.com/rlmenge/CBL-Mariner/pull/30
- PR to test changing kernel-azure config: https://github.com/rlmenge/CBL-Mariner/pull/32
- this pr does not run the check because the configs did not change :)

### DIFF
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -1,9 +1,9 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-# This action checks that certain groups of specs have matching tags.
-# The main use case is to ensure that signed specs have the same Version and
-#   Release tags as their unsigned counterparts
+# This action checks that required kernel configs have not been removed or
+# modified to an undesirable value. It also checks that new configs added to
+# the kernel config files are in the list of required configs.
 name: Kernel Required Configs Check
 
 on:
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -40,7 +40,6 @@ jobs:
           echo "base_sha=${{ github.event.before }}" >> $GITHUB_ENV
           echo "Merging ${{ github.sha }} into ${{ github.event.before }}"
 
-      # For consistency, we use the same major/minor version of Python that CBL-Mariner ships
       - name: Setup Python 3.11
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -1,0 +1,74 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# This action checks that certain groups of specs have matching tags.
+# The main use case is to ensure that signed specs have the same Version and
+#   Release tags as their unsigned counterparts
+name: Kernel Required Configs Check
+
+on:
+  push:
+    branches: [main, dev, 1.0*, 2.0*]
+  pull_request:
+    branches: [main, dev, 1.0*, 2.0*]
+
+jobs:
+  check:
+    name: Kernel configs check
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checkout the branch of our repo that triggered this action
+      - name: Workflow trigger checkout
+        uses: actions/checkout@v2
+        
+      - name: Get base commit for PRs
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          git fetch origin ${{ github.base_ref }}
+          echo "base_sha=$(git rev-parse origin/${{ github.base_ref }})" >> $GITHUB_ENV
+          echo "Merging ${{ github.sha }} into ${{ github.base_ref }}"
+
+      - name: Get base commit for Pushes
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          git fetch origin ${{ github.event.before }}
+          echo "base_sha=${{ github.event.before }}" >> $GITHUB_ENV
+          echo "Merging ${{ github.sha }} into ${{ github.event.before }}"
+
+
+      # For consistency, we use the same major/minor version of Python that CBL-Mariner ships
+      - name: Setup Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+
+      - name: Get Python dependencies
+        run: python3 -m pip install -r toolkit/scripts/requirements.txt
+
+      # Check kernel configs
+      - name: Get the changed files and configs
+        run: |
+          echo "Files changed: '$(git diff-tree --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }})'"
+          changed_configs=$(git diff-tree --diff-filter=d  --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }} | { grep "SPECS/kernel/config.*$" || test $? = 1; })
+          echo "Files to validate: '${changed_configs}'"
+          echo "updated-configs=$(echo ${changed_configs})" >> $GITHUB_ENV
+
+          # Check for new configs in the PR
+          # and check if they are in required configs
+          if [[ -n $changed_configs ]]; then
+            config_changes=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- SPECS/kernel/config)
+            echo "AMD config changes:"
+            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${changed_configs}"
+            
+            echo "ARM config changes:"
+            config_aarch64_changes=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- SPECS/kernel/config_aarch64)
+            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${config_aarch64_changes}"
+          fi
+
+      - name: Run kernel config checking script AMD64
+        run: python3 toolkit/scripts/check_required_kernel_configs.py --required_configs toolkit/scripts/mariner-required-configs.json --config_file SPECS/kernel/config
+      
+      - name: Run kernel config checking script ARM64
+        run: python3 toolkit/scripts/check_required_kernel_configs.py --required_configs toolkit/scripts/mariner-required-configs.json --config_file SPECS/kernel/config_aarch64
+
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -46,26 +46,29 @@ jobs:
       - name: Get Python dependencies
         run: python3 -m pip install -r toolkit/scripts/requirements.txt
 
-      # Check kernel configs
-      - name: Get the changed files and configs
+      # Check if kernel configs changed
+      - name: Check if config files changed
         run: |
           echo "Files changed: '$(git diff-tree --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }})'"
           changed_configs=$(git diff-tree --diff-filter=d  --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }} | { grep "SPECS/kernel/config.*$" || test $? = 1; })
           echo "Files to validate: '${changed_configs}'"
           echo "updated-configs=$(echo ${changed_configs})" >> $GITHUB_ENV
 
+      # Check if new configs were added
+      - name: Parse for new config changes
+        run: |
           # Check for new configs in the PR
           # and check if they are in required configs
-          if [[ -n $changed_configs ]]; then
-            config_changes=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- SPECS/kernel/config)
-            echo "AMD config changes:"
-            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${changed_configs}"
-            
-            echo "ARM config changes:"
-            config_aarch64_changes=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- SPECS/kernel/config_aarch64)
-            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${config_aarch64_changes}"
+          if [[ -n "${{ env.updated-configs }}" ]]; then
+            holder="${{ env.updated-configs }}"
+            for word in $holder; do
+              config_diff=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- ${word})
+              echo "config_diff for ${word} : ${config_diff}"
+              python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_file="${word}" --config_str="${config_diff}"
+            done
           fi
 
+      # Check if required configs were removed or modified
       - name: Run kernel config checking script AMD64
         run: python3 toolkit/scripts/check_required_kernel_configs.py --required_configs toolkit/scripts/mariner-required-configs.json --config_file SPECS/kernel/config
       
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -41,10 +41,10 @@ jobs:
           echo "Merging ${{ github.sha }} into ${{ github.event.before }}"
 
       # For consistency, we use the same major/minor version of Python that CBL-Mariner ships
-      - name: Setup Python 3.7
+      - name: Setup Python 3.11
         uses: actions/setup-python@v4
         with:
-          python-version: 3.7
+          python-version: 3.11
 
       - name: Get Python dependencies
         run: python3 -m pip install -r toolkit/scripts/requirements.txt
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -10,11 +10,11 @@ on:
   push:
     branches: [main, dev, 1.0*, 2.0*]
     paths:
-      - 'SPECS/kernel/config*'
+      - 'SPECS/kernel*/config*'
   pull_request:
     branches: [main, dev, 1.0*, 2.0*]
     paths:
-      - 'SPECS/kernel/config*'
+      - 'SPECS/kernel*/config*'
 
 jobs:
   check:
@@ -53,7 +53,7 @@ jobs:
       - name: Check if config files changed
         run: |
           echo "Files changed: '$(git diff-tree --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }})'"
-          changed_configs=$(git diff-tree --diff-filter=d  --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }} | { grep "SPECS/kernel/config.*$" || test $? = 1; })
+          changed_configs=$(git diff-tree --diff-filter=d  --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }} | { grep "SPECS/kernel.*/config.*$" || test $? = 1; })
           echo "Files to validate: '${changed_configs}'"
           echo "updated_configs=$(echo ${changed_configs})" >> $GITHUB_ENV
 
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       # Checkout the branch of our repo that triggered this action
       - name: Workflow trigger checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         
       - name: Get base commit for PRs
         if: ${{ github.event_name == 'pull_request' }}
@@ -36,10 +36,9 @@ jobs:
           echo "base_sha=${{ github.event.before }}" >> $GITHUB_ENV
           echo "Merging ${{ github.sha }} into ${{ github.event.before }}"
 
-
       # For consistency, we use the same major/minor version of Python that CBL-Mariner ships
       - name: Setup Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
 
@@ -52,19 +51,19 @@ jobs:
           echo "Files changed: '$(git diff-tree --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }})'"
           changed_configs=$(git diff-tree --diff-filter=d  --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }} | { grep "SPECS/kernel/config.*$" || test $? = 1; })
           echo "Files to validate: '${changed_configs}'"
-          echo "updated-configs=$(echo ${changed_configs})" >> $GITHUB_ENV
+          echo "updated_configs=$(echo ${changed_configs})" >> $GITHUB_ENV
 
       # Check if new configs were added
       - name: Parse for new config changes
         run: |
           # Check for new configs in the PR
           # and check if they are in required configs
-          if [[ -n "${{ env.updated-configs }}" ]]; then
-            holder="${{ env.updated-configs }}"
-            for word in $holder; do
-              config_diff=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- ${word})
-              echo "config_diff for ${word} : ${config_diff}"
-              python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_file="${word}" --config_str="${config_diff}"
+          if [[ -n "${{ env.updated_configs }}" ]]; then
+            holder="${{ env.updated_configs }}"
+            for file in $holder; do
+              config_diff=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- ${file})
+              echo "config_diff for ${file} : ${config_diff}"
+              python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_file="${file}" --config_diff="${config_diff}"
             done
           fi
 
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -9,8 +9,12 @@ name: Kernel Required Configs Check
 on:
   push:
     branches: [main, dev, 1.0*, 2.0*]
+    paths:
+      - 'SPECS/kernel/config*'
   pull_request:
     branches: [main, dev, 1.0*, 2.0*]
+    paths:
+      - 'SPECS/kernel/config*'
 
 jobs:
   check:
@@ -58,19 +62,17 @@ jobs:
         run: |
           # Check for new configs in the PR
           # and check if they are in required configs
-          if [[ -n "${{ env.updated_configs }}" ]]; then
-            holder="${{ env.updated_configs }}"
-            for file in $holder; do
-              config_diff=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- ${file})
-              echo "config_diff for ${file} : ${config_diff}"
-              python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_file="${file}" --config_diff="${config_diff}"
-            done
-          fi
+          holder="${{ env.updated_configs }}"
+          for file in $holder; do
+            config_diff=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- ${file})
+            echo "config_diff for ${file} : ${config_diff}"
+            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_file="${file}" --config_diff="${config_diff}"
+          done
 
       # Check if required configs were removed or modified
-      - name: Run kernel config checking script AMD64
-        run: python3 toolkit/scripts/check_required_kernel_configs.py --required_configs toolkit/scripts/mariner-required-configs.json --config_file SPECS/kernel/config
-      
-      - name: Run kernel config checking script ARM64
-        run: python3 toolkit/scripts/check_required_kernel_configs.py --required_configs toolkit/scripts/mariner-required-configs.json --config_file SPECS/kernel/config_aarch64
-
+      - name: Run kernel config checking script
+        run: |
+          holder="${{ env.updated_configs }}"
+          for file in $holder; do
+            python3 toolkit/scripts/check_required_kernel_configs.py --required_configs toolkit/scripts/mariner-required-configs.json --config_file "${file}"
+          done
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -66,7 +66,7 @@ jobs:
           for file in $holder; do
             config_diff=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- ${file})
             echo "config_diff for ${file} : ${config_diff}"
-            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_file="${file}" --config_diff="${config_diff}"
+            python3 toolkit/scripts/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_file="${file}" --config_diff="${config_diff}"
           done
 
       # Check if required configs were removed or modified
--- a/.github/workflows/check-kernel-config.yml
+++ b/.github/workflows/check-kernel-config.yml
@@ -8,11 +8,11 @@ name: Kernel Required Configs Check
 
 on:
   push:
-    branches: [main, dev, 1.0*, 2.0*]
+    branches: [main, 1.0*, 2.0*]
     paths:
       - 'SPECS/kernel*/config*'
   pull_request:
-    branches: [main, dev, 1.0*, 2.0*]
+    branches: [main, 1.0*, 2.0*]
     paths:
       - 'SPECS/kernel*/config*'
 
--- a/.github/workflows/check_new_kernel_configs.py
+++ b/.github/workflows/check_new_kernel_configs.py
@@ -9,6 +9,11 @@ import argparse
 import sys
 import re
 
+def get_data_from_config(input_file):
+    with open(input_file, 'r') as file:
+        input_config_data = file.read()
+    return input_config_data
+
 def check_kernel(input_file):
     match = re.search(r'SPECS/(.*?)/', input_file)
     if match:
@@ -16,12 +21,10 @@ def check_kernel(input_file):
     else:
         return None
 
-def check_config_arch(input_file):
-    with open(input_file, 'r') as file:
-        contents = file.read()
-    if "Linux/x86_64" in contents:
+def check_config_arch(input_config_data):
+    if "Linux/x86_64" in input_config_data:
         return "AMD64"
-    elif "Linux/arm64" in contents:
+    elif "Linux/arm64" in input_config_data:
         return "ARM64"
     else:
         return None
@@ -73,11 +76,12 @@ if kernel == None:
     print("Kernel not found in config filepath")
     sys.exit(1)
 
-arch = check_config_arch(config_file)
+input_config_data = get_data_from_config(config_file)
+
+arch = check_config_arch(input_config_data)
 if arch == None:
     print("Architecture not found in config file")
     sys.exit(1)
-
 
 missing_configs = find_missing_configs(required_configs, kernel, arch, config_diff)
 if missing_configs == None:
--- a/.github/workflows/check_new_kernel_configs.py
+++ b/.github/workflows/check_new_kernel_configs.py
@@ -22,51 +22,51 @@ def check_config_arch(input_file):
 
 # find the lines in the diff that contain +/- CONFIG
 # ignore the lines that contain @
+# return the set of words that follow +/-
 def find_matching_lines(input_string):
     pattern = r'(?!.*@)[+-]\s*.*CONFIG.*'
     matching_lines = re.findall(pattern, input_string)
-
-    return matching_lines
+    holder=[re.sub(r"\+|\-|=y|=m|\#|is not set", r"", s).strip() for s in matching_lines]
+    config_set = set(holder)
+    return config_set
 
 # parse diff for new kernel configs
 # check if they are in required configs
-def find_missing_words(json_file, arch, config_diff):
+def find_missing_configs(json_file, arch, config_diff):
     # Load the JSON object
     with open(json_file, 'r') as file:
         data = json.load(file)
-    configData = data['required-configs']
+    config_data = data['required-configs']
 
     # Extract the words from the string
     config_words = find_matching_lines(config_diff)
-    holder = [ s.replace('+', '').replace('-', '').replace('=y','').replace('=m','').replace(' is not set', '').replace('#', '').strip() for s in config_words]
-    wordSet = set(holder)
 
     # Find the missing words
-    missing_words = []
-    for word in wordSet:
-        if word not in configData or arch not in configData[word]["arch"] :
-            missing_words.append(word)
-    return missing_words
+    missing_configs = []
+    for word in config_words:
+        if word not in config_data or arch not in config_data[word]["arch"] :
+            missing_configs.append(word)
+    return missing_configs
 
 
 parser = argparse.ArgumentParser(
-description="Tool for checking if known required kernel configs are present.")
-parser.add_argument('--required_configs', help='path to json of required configs', required=True)
+description="Tool for checking new configs are present in required configs JSON.")
+parser.add_argument('--required_configs', help='path to JSON of required configs', required=True)
 parser.add_argument('--config_file', help='path to config', required=True)
-parser.add_argument('--config_str', help='config diff', required=True)
+parser.add_argument('--config_diff', help='config diff', required=True)
 args = parser.parse_args()
-requiredConfigs = args.required_configs
-configFile = args.config_file
-configDiff = args.config_str
-arch = check_config_arch(configFile)
+required_configs = args.required_configs
+config_file = args.config_file
+config_diff = args.config_diff
+arch = check_config_arch(config_file)
 
-missing_words = find_missing_words(requiredConfigs, arch, configDiff)
-print("Missing words:", missing_words)
-if len(missing_words) == 0:
+missing_configs = find_missing_configs(required_configs, arch, config_diff)
+print("Missing words:", missing_configs)
+if len(missing_configs) == 0:
     print("All configs are present in required configs")
 else:
-    print ("====================== Kernel new config verification FAILED for {0} ======================".format(arch))
-    print("New configs detected for {0}. Please add the following to toolkit/scripts/mariner-required-configs.json".format(arch))
-    for word in missing_words:
-        print('{0}'.format(word))
+    print (f"====================== Kernel new config verification FAILED for {arch} ======================")
+    print(f"New configs detected for {arch}. Please add the following to toolkit/scripts/mariner-required-configs.json")
+    for word in missing_configs:
+        print(word)
     sys.exit(1)
--- a/.github/workflows/check_new_kernel_configs.py
+++ b/.github/workflows/check_new_kernel_configs.py
@@ -9,6 +9,17 @@ import argparse
 import sys
 import re
 
+def check_config_arch(input_file):
+    with open(input_file, 'r') as file:
+        contents = file.read()
+    file.close()
+    if "Linux/x86_64" in contents:
+        return "AMD64"
+    elif "Linux/arm64" in contents:
+        return "ARM64"
+    else:
+        return None
+
 # find the lines in the diff that contain +/- CONFIG
 # ignore the lines that contain @
 def find_matching_lines(input_string):
@@ -19,40 +30,43 @@ def find_matching_lines(input_string):
 
 # parse diff for new kernel configs
 # check if they are in required configs
-def find_missing_words(json_file, string):
+def find_missing_words(json_file, arch, config_diff):
     # Load the JSON object
     with open(json_file, 'r') as file:
         data = json.load(file)
     configData = data['required-configs']
 
     # Extract the words from the string
-    config_words = find_matching_lines(string)
+    config_words = find_matching_lines(config_diff)
     holder = [ s.replace('+', '').replace('-', '').replace('=y','').replace('=m','').replace(' is not set', '').replace('#', '').strip() for s in config_words]
     wordSet = set(holder)
 
     # Find the missing words
     missing_words = []
     for word in wordSet:
-        if word not in configData:
+        if word not in configData or arch not in configData[word]["arch"] :
             missing_words.append(word)
-
     return missing_words
 
 
 parser = argparse.ArgumentParser(
 description="Tool for checking if known required kernel configs are present.")
 parser.add_argument('--required_configs', help='path to json of required configs', required=True)
-parser.add_argument('--config_str', help='path to config being checked', required=True)
+parser.add_argument('--config_file', help='path to config', required=True)
+parser.add_argument('--config_str', help='config diff', required=True)
 args = parser.parse_args()
 requiredConfigs = args.required_configs
-configStr = args.config_str
+configFile = args.config_file
+configDiff = args.config_str
+arch = check_config_arch(configFile)
 
-missing_words = find_missing_words(requiredConfigs, configStr)
+missing_words = find_missing_words(requiredConfigs, arch, configDiff)
 print("Missing words:", missing_words)
 if len(missing_words) == 0:
     print("All configs are present in required configs")
 else:
-    print ("====================== Kernel config verification FAILED ======================")
+    print ("====================== Kernel new config verification FAILED for {0} ======================".format(arch))
+    print("New configs detected for {0}. Please add the following to toolkit/scripts/mariner-required-configs.json".format(arch))
     for word in missing_words:
-        print('{0} is new. Please add to toolkit/scripts/mariner-required-configs.json'.format(word))
+        print('{0}'.format(word))
     sys.exit(1)
--- a/.github/workflows/check_new_kernel_configs.py
+++ b/.github/workflows/check_new_kernel_configs.py
@@ -9,10 +9,16 @@ import argparse
 import sys
 import re
 
+def check_kernel(input_file):
+    match = re.search(r'SPECS/(.*?)/', input_file)
+    if match:
+        return match.group(1)
+    else:
+        return None
+
 def check_config_arch(input_file):
     with open(input_file, 'r') as file:
         contents = file.read()
-    file.close()
     if "Linux/x86_64" in contents:
         return "AMD64"
     elif "Linux/arm64" in contents:
@@ -32,11 +38,14 @@ def find_matching_lines(input_string):
 
 # parse diff for new kernel configs
 # check if they are in required configs
-def find_missing_configs(json_file, arch, config_diff):
+def find_missing_configs(json_file, kernel, arch, config_diff):
     # Load the JSON object
     with open(json_file, 'r') as file:
         data = json.load(file)
-    config_data = data['required-configs']
+        if kernel not in data:
+            print(f"Kernel {kernel} not found in {json_file}")
+            return None
+        config_data = data[kernel]['required-configs']
 
     # Extract the words from the string
     config_words = find_matching_lines(config_diff)
@@ -58,9 +67,23 @@ args = parser.parse_args()
 required_configs = args.required_configs
 config_file = args.config_file
 config_diff = args.config_diff
-arch = check_config_arch(config_file)
 
-missing_configs = find_missing_configs(required_configs, arch, config_diff)
+kernel = check_kernel(config_file)
+if kernel == None:
+    print("Kernel not found in config filepath")
+    sys.exit(1)
+
+arch = check_config_arch(config_file)
+if arch == None:
+    print("Architecture not found in config file")
+    sys.exit(1)
+
+
+missing_configs = find_missing_configs(required_configs, kernel, arch, config_diff)
+if missing_configs == None:
+    print(f"Could not find required configs for {kernel}")
+    sys.exit(0)
+
 print("Missing words:", missing_configs)
 if len(missing_configs) == 0:
     print("All configs are present in required configs")
--- a/.github/workflows/check_new_kernel_configs.py
+++ b/.github/workflows/check_new_kernel_configs.py
@@ -48,16 +48,16 @@ def find_missing_configs(json_file, kernel, arch, config_diff):
         if kernel not in data:
             print(f"Kernel {kernel} not found in {json_file}")
             return None
-        config_data = data[kernel]['required-configs']
+        required_configs_data = data[kernel]['required-configs']
 
     # Extract the words from the string
     config_words = find_matching_lines(config_diff)
 
     # Find the missing words
     missing_configs = []
-    for word in config_words:
-        if word not in config_data or arch not in config_data[word]["arch"] :
-            missing_configs.append(word)
+    for config_option in config_words:
+        if config_option not in required_configs_data or arch not in required_configs_data[config_option]["arch"] :
+            missing_configs.append(config_option)
     return missing_configs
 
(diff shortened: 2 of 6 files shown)


toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
...
+        },
+        "CONFIG_BINFMT_MISC": {
+            "value": [
+                "y"
+            ],
+            "arch": [
+                "AMD64"
+            ],
+            "comment": "Needed for customer",
+            "PR": null
```
PawelWMS, line 85:
```suggestion
            "PR": https://github.com/microsoft/CBL-Mariner/pull/3300
```

toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
...
+            ],
+            "comment": "Needed for auoms metrics",
+            "PR": "https://github.com/microsoft/CBL-Mariner/pull/5292"
+        },
+        "CONFIG_INIT_ON_FREE_DEFAULT_ON": {
+            "value": [
+                "is not set",
+                ""
+            ],
+            "arch": [
```
christopherco, line 406: Not needed for this PR, but consider that this arch field could extend to more architectures in the future

toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
...
+        },
+        "CONFIG_SCHEDSTATS": {
+            "value": [
+                "y"
+            ],
+            "arch": [
+                "AMD64"
+            ],
+            "comment": "Needed for customer",
+            "PR": null
```
PawelWMS, line 95: >"PR": null [](http://example.com/codeflow?start=12&length=10)

Suggestion: for changes, that were done before Mariner GA'ed and got moved to GitHub, we could either mention the PR number (1693 in this case) from those times or just say "Initial Mariner kernel configuration".

toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
...
+        },
+        "CONFIG_IO_URING": {
+            "value": [
+                "y"
+            ],
+            "arch": [
+                "AMD64"
+            ],
+            "comment": "Needed for customer",
+            "PR": null
```
PawelWMS, line 166: >"PR": null [](http://example.com/codeflow?start=12&length=10)

Pre-GA ADO PR 427.

toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
...
+        "CONFIG_SENSORS_DELL_SMM": {
+            "value": [
+                "m",
+                "y"
+            ],
+            "arch": [
+                "AMD64"
+            ],
+            "comment": "Needed for customer",
+            "PR": null
```
PawelWMS, line 219:
```suggestion
            "PR": https://github.com/microsoft/CBL-Mariner/pull/5406
```

toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
...
+        "CONFIG_NVME_RDMA": {
+            "value": [
+                "m",
+                "y"
+            ],
+            "arch": [
+                "AMD64"
+            ],
+            "comment": "Needed for customer",
+            "PR": null
```
PawelWMS, line 241:
```suggestion
            "PR": "https://github.com/microsoft/CBL-Mariner/pull/5283"
```

toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
...
+        "CONFIG_NVME_TCP": {
+            "value": [
+                "m",
+                "y"
+            ],
+            "arch": [
+                "AMD64"
+            ],
+            "comment": "Needed for customer",
+            "PR": null
```
PawelWMS, line 293:
```suggestion
            "PR": "https://github.com/microsoft/CBL-Mariner/pull/5283"
```

toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
...
+        },
+        "CONFIG_TCG_TPM": {
+            "value": [
+                "y"
+            ],
+            "arch": [
+                "AMD64"
+            ],
+            "comment": "Needed for customer",
+            "PR": null
```
PawelWMS, line 357:
```suggestion
            "PR": "https://github.com/microsoft/CBL-Mariner/pull/135"
```

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,74 @@
...
+          python-version: 3.7
+
+      - name: Get Python dependencies
+        run: python3 -m pip install -r toolkit/scripts/requirements.txt
+
+      # Check kernel configs
+      - name: Get the changed files and configs
+        run: |
+          echo "Files changed: '$(git diff-tree --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }})'"
+          changed_configs=$(git diff-tree --diff-filter=d  --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }} | { grep "SPECS/kernel/config.*$" || test $? = 1; })
```
PawelWMS, line 53: >"SPECS/kernel/config.*$" [](http://example.com/codeflow?start=136&length=24)

Don't we want to scan configs for other kernels as well? Or is this a V1 for now?

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,74 @@
...
+          echo "Files to validate: '${changed_configs}'"
+          echo "updated-configs=$(echo ${changed_configs})" >> $GITHUB_ENV
+
+          # Check for new configs in the PR
+          # and check if they are in required configs
+          if [[ -n $changed_configs ]]; then
+            config_changes=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- SPECS/kernel/config)
+            echo "AMD config changes:"
+            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${changed_configs}"
```
PawelWMS, line 62:
```suggestion
            echo "AMD config changes:"
            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${config_changes}"
```

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Get Python dependencies
+        run: python3 -m pip install -r toolkit/scripts/requirements.txt
+
+      # Check kernel configs
+      - name: Get the changed files and configs
+        run: |
+          echo "Files changed: '$(git diff-tree --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }})'"
+          changed_configs=$(git diff-tree --diff-filter=d  --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }} | { grep "SPECS/kernel/config.*$" || test $? = 1; })
+          echo "Files to validate: '${changed_configs}'"
+          echo "updated-configs=$(echo ${changed_configs})" >> $GITHUB_ENV
```
PawelWMS, line 55: Seems unnecessary, since we're not accessing that variable in other steps:
```suggestion
```

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,74 @@
...
+            config_aarch64_changes=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- SPECS/kernel/config_aarch64)
+            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${config_aarch64_changes}"
+          fi
+
+      - name: Run kernel config checking script AMD64
+        run: python3 toolkit/scripts/check_required_kernel_configs.py --required_configs toolkit/scripts/mariner-required-configs.json --config_file SPECS/kernel/config
+      
+      - name: Run kernel config checking script ARM64
+        run: python3 toolkit/scripts/check_required_kernel_configs.py --required_configs toolkit/scripts/mariner-required-configs.json --config_file SPECS/kernel/config_aarch64
+
```
PawelWMS, line 74: Suggestion: it looks like we're first checking if any config files were touched in this PR and then go one-by-one inside the `if` statement anyway, even if only one of the configs was modified. While I believe that it's unlikely for only one of the configs to change at a time, I'm thinking we could make this code more flexible by moving away from the repeated steps in the `if` statement and just storing a list of configs we want to verify. Similarly, I think we could also completely drop the `changed_configs` variable:

```suggestion
    branches: [main, dev, 1.0*, 2.0*]

env:
  CHECKED_CONFIGS: ("SPECS/kernel/config SPECS/kernel/config_aarch64")

jobs:
  check:
    name: Kernel configs check
    runs-on: ubuntu-latest

    steps:
      # Checkout the branch of our repo that triggered this action
      - name: Workflow trigger checkout
        uses: actions/checkout@v2
        
      - name: Get base commit for PRs
        if: ${{ github.event_name == 'pull_request' }}
        run: |
          git fetch origin ${{ github.base_ref }}
          echo "base_sha=$(git rev-parse origin/${{ github.base_ref }})" >> $GITHUB_ENV
          echo "Merging ${{ github.sha }} into ${{ github.base_ref }}"

      - name: Get base commit for Pushes
        if: ${{ github.event_name == 'push' }}
        run: |
          git fetch origin ${{ github.event.before }}
          echo "base_sha=${{ github.event.before }}" >> $GITHUB_ENV
          echo "Merging ${{ github.sha }} into ${{ github.event.before }}"


      # For consistency, we use the same major/minor version of Python that CBL-Mariner ships
      - name: Setup Python 3.7
        uses: actions/setup-python@v2
        with:
          python-version: 3.7

      - name: Get Python dependencies
        run: python3 -m pip install -r toolkit/scripts/requirements.txt

      - name: Check new kernel configs entries
        run: |
          for checked_config in $CHECKED_CONFIGS; do
            echo "Checking for unexpected new entries inside ($checked_config)."

            config_changes=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- "$checked_config")
            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${config_changes}"            
          done

      - name: Check required kernel config entries
        run: |
          for checked_config in $CHECKED_CONFIGS; do
            echo "Checking for required entries inside ($checked_config)."

            python3 toolkit/scripts/check_required_kernel_configs.py --required_configs toolkit/scripts/mariner-required-configs.json --config_file "$checked_config"
          done

```

Note 1: we could even use a matrix to run these in parallel for each config, but I think that's a bit of an overkill.
Note 2: I'm not entirely sure how failure is detected in these pipelines. It's possible we might need to use `set -e` or just exit manually whenever we detect a failure for steps that aren't just a single command.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) Microsoft Corporation.
```
PawelWMS, line 1: Curious: why keep this script here and the other one inside `toolkit/scripts`?

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,58 @@
...
+
+import json
+import argparse
+import sys
+import re
+
+# find the lines in the diff that contain +/- CONFIG
+# ignore the lines that contain @
+def find_matching_lines(input_string):
+    pattern = r'(?!.*@)[+-]\s*.*CONFIG.*'
```
PawelWMS, line 15: >pattern = r'(?!.*@)[+-]\s*.*CONFIG.*' [](http://example.com/codeflow?start=4&length=37)

Suggestion: compile once by moving this to a global and using:

```python
config_line_patter = re.compile(r'(?!.*@)[+-]\s*.*CONFIG.*')
```

.github/workflows/check_new_kernel_configs.py
rlmenge: I can't really see a need for someone to use the check for a new config outside of the context for the workflow

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,58 @@
...
+        if word not in configData:
+            missing_words.append(word)
+
+    return missing_words
+
+
+parser = argparse.ArgumentParser(
+description="Tool for checking if known required kernel configs are present.")
+parser.add_argument('--required_configs', help='path to json of required configs', required=True)
+parser.add_argument('--config_str', help='path to config being checked', required=True)
```
PawelWMS, line 45: >path to config being checked [](http://example.com/codeflow?start=42&length=28)

Somewhat confusing help, since it's not a file but a diff string.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,58 @@
...
+    for word in wordSet:
+        if word not in configData:
+            missing_words.append(word)
+
+    return missing_words
+
+
+parser = argparse.ArgumentParser(
+description="Tool for checking if known required kernel configs are present.")
+parser.add_argument('--required_configs', help='path to json of required configs', required=True)
```
PawelWMS, line 44: Nit-pick:

```suggestion
parser.add_argument('--required_configs', help='path to JSON of required configs', required=True)
```

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,58 @@
...
+configStr = args.config_str
+
+missing_words = find_missing_words(requiredConfigs, configStr)
+print("Missing words:", missing_words)
+if len(missing_words) == 0:
+    print("All configs are present in required configs")
+else:
+    print ("====================== Kernel config verification FAILED ======================")
+    for word in missing_words:
+        print('{0} is new. Please add to toolkit/scripts/mariner-required-configs.json'.format(word))
```
PawelWMS, line 57: Suggestion: I'd use [Python's f-strings](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python) to shorten the lines. I also think it makes it easier to see what goes where.
```suggestion
        print(f'{word} is new. Please add to toolkit/scripts/mariner-required-configs.json')
```

.github/workflows/check_new_kernel_configs.py
PawelWMS, line 15: >r'(?!.*@)[+-]\s*.*CONFIG.*' [](http://example.com/codeflow?start=14&length=27)

What are we searching for exactly? If we only care about new entries, then I'm thinking we might be able to simplify the pattern. How about `^\+CONFIG_.*`?

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,74 @@
...
+          # Check for new configs in the PR
+          # and check if they are in required configs
+          if [[ -n $changed_configs ]]; then
+            config_changes=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- SPECS/kernel/config)
+            echo "AMD config changes:"
+            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${changed_configs}"
+            
+            echo "ARM config changes:"
+            config_aarch64_changes=$(git diff-tree -p -r ${{ env.base_sha }} ${{ github.sha }} -- SPECS/kernel/config_aarch64)
+            python3 .github/workflows/check_new_kernel_configs.py --required_configs=toolkit/scripts/mariner-required-configs.json --config_str="${config_aarch64_changes}"
```
PawelWMS, line 66: >check_new_kernel_configs [](http://example.com/codeflow?start=38&length=24)

Do we really want to catch every new config entry? With newer kernels there's going to be newer configs and for some of them we're going to take the default value because we just don't have any specific requirements. At least that's my guess. If our policy is to track every single config entry, then I think this is fine, but otherwise, we're going to bloat that JSON file and potentially cause a lot of headache for people without a good reason.

Related question: what about config entries, which were there in the past and now a config update is robust enough, for git to think their lines got removed and added elsewhere? In such cases we're going to flag an pre-existing config as a new one.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,58 @@
...
+# ignore the lines that contain @
+def find_matching_lines(input_string):
+    pattern = r'(?!.*@)[+-]\s*.*CONFIG.*'
+    matching_lines = re.findall(pattern, input_string)
+
+    return matching_lines
+
+# parse diff for new kernel configs
+# check if they are in required configs
+def find_missing_words(json_file, string):
```
PawelWMS, line 22: Could we use a different variable name than `string`? It's a keyword in so many languages, that this can be a bit confusing. Also, it this name doesn't really provide a lot of information to the reader.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,58 @@
...
+    # Extract the words from the string
+    config_words = find_matching_lines(string)
+    holder = [ s.replace('+', '').replace('-', '').replace('=y','').replace('=m','').replace(' is not set', '').replace('#', '').strip() for s in config_words]
+    wordSet = set(holder)
+
+    # Find the missing words
+    missing_words = []
+    for word in wordSet:
+        if word not in configData:
+            missing_words.append(word)
```
PawelWMS, line 37: I'm curious if using a [set difference](https://www.geeksforgeeks.org/python-set-difference/) would work.

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,81 @@
...
+    parser = argparse.ArgumentParser(
+        description="Tool for checking if known required kernel configs are present.")
+
+
+    parser.add_argument('--required_configs', help='path to json of required configs', required=True)
+    parser.add_argument('--config_file', help='path to config being checked', required=True)
+    args = parser.parse_args()
+    requiredConfigs = args.required_configs
+    configFile = args.config_file
+
```
PawelWMS, line 67: Please let's stick to one variable-naming convention.

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,81 @@
...
+        contents = file.read()
+    file.close()
+    if "Linux/x86_64" in contents:
+        return "AMD64"
+    elif "Linux/arm64" in contents:
+        return "ARM64"
+    else:
+        return None
+
+def check_strings_in_file(json_file, input_file):
```
PawelWMS, line 22: Could we use more precise names for the input arguments? I think the ones you're using when you're calling this function are quite good.

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,81 @@
...
+    with open(json_file, 'r') as file:
+        data = json.load(file)
+        configData = data['required-configs']
+
+    with open(input_file, 'r') as file:
+        contents = file.read().split("\n")
+
+    missing_configs = {}
+
+    for key, value in configData.items():
```
PawelWMS, line 33: Please use more descriptive names for the key and value variables. Otherwise it's getting somewhat hard to read the code.

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  check:
+    name: Kernel configs check
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checkout the branch of our repo that triggered this action
+      - name: Workflow trigger checkout
+        uses: actions/checkout@v2
```
oliviacrain, line 23: Please update to `actions/checkout@v3`- that version is based on a newer node runtime that'll be supported for longer

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,81 @@
...
+        # check for arch
+        if arch not in value['arch']:
+            continue
+        found = False
+        for line in contents:
```
PawelWMS, line 38: Instead of re-parsing the contents for every JSON entry, I'd suggest extracting all of the config values from `contents` into a confing-to-value map, removing the `for` loop and using a simple `if key in kernel_configs` followed by `config_value = kernel_configs[key]`. Then you can check if `config_value` is one of the allowed ones.

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,74 @@
...
+# This action checks that required kernel configs have not been removed or
+# modified to an undesirable value. It also checks that new configs added to
+# the kernel config files are in the list of required configs.
+name: Kernel Required Configs Check
+
+on:
+  push:
+    branches: [main, dev, 1.0*, 2.0*]
+  pull_request:
+    branches: [main, dev, 1.0*, 2.0*]
```
oliviacrain, line 13: We might benefit from [filtering based on changed files](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore). That would allow us to get rid of all of the changed file logic, and prevent this from running on unrelated PRs.

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,74 @@
...
+          git fetch origin ${{ github.event.before }}
+          echo "base_sha=${{ github.event.before }}" >> $GITHUB_ENV
+          echo "Merging ${{ github.sha }} into ${{ github.event.before }}"
+
+
+      # For consistency, we use the same major/minor version of Python that CBL-Mariner ships
+      - name: Setup Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
oliviacrain, line 44: Should probably go for a version of Python that isn't EOL in a little over a month. I'm working on a cleanup PR to fix this for other actions, so no need to change this elsewhere.

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,81 @@
...
+            # check for config in line (without extra _VALUE)
+            if "{0}=".format(key) in line or "{0} is not set".format(key) in line:
+                for val in value['value']:
+                    if val in line and val != "":
+                        found = True
+                        break
+                else:
```
PawelWMS, line 45: I think there's something wrong with this indent.

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,81 @@
...
+                    missing_configs[key] = (line.split(key)[1].replace('=',''), value['value'], value['comment'], value['PR'])
```
PawelWMS, line 46: >missing_configs [](http://example.com/codeflow?start=20&length=15)

Could we use an object with well-named fields instead? That way it's going to be easier to understand what's going on instead of reading raw indeces.

Example:
```python
missing_configs[key] = {"config_name" = key, "allowed_values" = value.value, "comment" = value.comment, "pr" = value.PR}
```

.github/workflows/check-kernel-config.yml
PawelWMS: I had no idea that's a thing. Sounds like a great idea, I need to update other PR checks now.;)

PawelWMS (changes requested):
Left a few comments.

.github/workflows/check-kernel-config.yml
rlmenge: Now we are :)

.github/workflows/check-kernel-config.yml
oliviacrain: I should say- you can only get rid of all the changed file logic if you have a static set of files you're looking for. For, say, the linter, where we need to know what specific subset of specs to scan, we still need it.

.github/workflows/check_new_kernel_configs.py
rlmenge: So this regex is looking for lines that do not have @ and do have + or - and CONFIG. I will look into expanding this further and expanding my regex powers : P

toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
+{
```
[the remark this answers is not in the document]
rlmenge: for now, I expanded the JSON's first value be the kernel for which we are checking

toolkit/scripts/mariner-required-configs.json
```diff
@@ -0,0 +1,640 @@
...
+        },
+        "CONFIG_HIST_TRIGGERS": {
+            "value": [
+                "y"
+            ],
+            "arch": [
+                "AMD64"
+            ],
+            "comment": "Needed for auoms metrics",
+            "PR": "https://github.com/microsoft/CBL-Mariner/pull/5292"
```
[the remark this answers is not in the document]
rlmenge: Made the PR field a list. Comments can just be longer strings imo

.github/workflows/check-kernel-config.yml
rlmenge: jk I have made it so the json can be extended to support them

.github/workflows/check_new_kernel_configs.py
rlmenge: had to expand the check to account for the arch (config could be in AMD but not ARM and thus is still missing)

.github/workflows/check-kernel-config.yml
rlmenge: failures trigger a workflow failure :) I have seen enough now to know this :P

.github/workflows/check-kernel-config.yml
rlmenge: This was an ask from Chris to ensure the required configs stay updated.

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,78 @@
...
+          changed_configs=$(git diff-tree --diff-filter=d  --no-commit-id --name-only -r ${{ env.base_sha }} ${{ github.sha }} | { grep "SPECS/kernel.*/config.*$" || test $? = 1; })
+          echo "Files to validate: '${changed_configs}'"
+          echo "updated_configs=$(echo ${changed_configs})" >> $GITHUB_ENV
+
+      # Check if new configs were added
+      - name: Parse for new config changes
+        run: |
+          # Check for new configs in the PR
+          # and check if they are in required configs
+          holder="${{ env.updated_configs }}"
```
rlmenge, line 65: holder is needed for the bas to properly regster the string and parse each word inside. Open to better suggestions if known

.github/workflows/check-kernel-config.yml
PawelWMS: I played around with this for other PR checks. It looks like required checks will block a PR instead of being skipped when we use `paths` but none of the config files were changes for a given PR. I haven't had time to see how to configure this better, so that irrelevant PR checks are skipped correctly, but unless we do that, we cannot make this PR check a required one (it would have a similar effect as the linter check).

.github/workflows/check-kernel-config.yml
PawelWMS: As long as Chris, you, and everyone else involved in managing the kernel accept this additional workload, then who am I to stop you?:)

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,78 @@
...
+          for file in $holder; do
```
PawelWMS, line 66: Nit-pick: do you really need this additional `holder` variable? The name isn't giving us any more context (if anything it feels like it's making it more cryptic) and we could save two lines by using `${{ env.updated_configs }}` directly (just without the double quotes around it).

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,99 @@
...
+    with open(input_file, 'r') as file:
+        input_config_data = file.read()
+    return input_config_data
+
+def check_kernel(input_file):
+    match = re.search(r'SPECS/(.*?)/', input_file)
+    if match:
+        return match.group(1)
+    else:
+        return None
```
PawelWMS, line 22: Two things:
- I'd think about the method name a bit. Right now it sounds like you're trying to check the actual kernel and it's a bit confusing. If I understand this correctly, you just want to verify if the file is a kernel config.
- We might be overdoing it here a bit with trying to verify if the given file is indeed a kernel config file. Seeing that it's a very simple format, the best we could do is to check if it contains at least one line beginning with `CONFIG_SOMETHING=(y|n|m)`. But even that wouldn't be a really reliable check. With that said, I think this check might be an overkill and I'd just drop it.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,99 @@
...
+
+def check_config_arch(input_config_data):
```
PawelWMS, line 24: >check_config_arch [](http://example.com/codeflow?start=4&length=17)

General naming observation: functions beginning with "check" make the reader assume we are verifying something and potentially returning/throwing an error or returning a boolean saying if the check succeeded or not. At least that would be my guess. In this case, we're trying to determine the architecture from the config, so I'd suggest something closer to `extract_config_arch`, `determine_config_arch`, etc.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,99 @@
...
+        return "AMD64"
+    elif "Linux/arm64" in input_config_data:
+        return "ARM64"
+    else:
+        return None
+
+# find the lines in the diff that contain +/- CONFIG
+# ignore the lines that contain @
+# return the set of words that follow +/-
+def find_matching_lines(input_string):
```
PawelWMS, line 35: >find_matching_lines [](http://example.com/codeflow?start=4&length=19)

Another naming suggestion: the main purpose of this function seems to be to extract or of the config variables, so I'd mention that in the name. Right now the function's name is so generic, it's impossible to tell what to expect from its output.

.github/workflows/check_new_kernel_configs.py
PawelWMS: Nevermind, I see now that it's actually supposed to extract the kernel folder name from the config path. I'm still going to pretend it's not me paying too little attention to the code and blame it on the name of the function instead.:)

Anyway: I'd re-name it to something closer to `read_kernel_name`, `extract_kernel_dir_name`, etc. so the name better reflects what's happening.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,99 @@
...
+    pattern = r'(?!.*@)[+-]\s*.*CONFIG.*'
+    matching_lines = re.findall(pattern, input_string)
+    holder=[re.sub(r"\+|\-|=y|=m|\#|is not set", r"", s).strip() for s in matching_lines]
+    config_set = set(holder)
+    return config_set
+
+# parse diff for new kernel configs
+# check if they are in required configs
```
PawelWMS, line 43: Nit-pick: let's be consistent with how we write comments. My personal preference: begin with a capital letter and use proper punctuation. To the best of our abilities that is.;)

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,99 @@
...
+def find_missing_configs(json_file, kernel, arch, config_diff):
```
PawelWMS, line 44: >json_file [](http://example.com/codeflow?start=25&length=9)

Argument naming suggestion: `config_json_path` might give the reader a better idea of what this is.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,99 @@
...
+    # Load the JSON object
+    with open(json_file, 'r') as file:
```
PawelWMS, line 46: >json_file [](http://example.com/codeflow?start=14&length=9)

General comment for both scripts: when within one script we're variables representing both file paths and file objects (like the `file` variable below), I'd advise making it clear in the variable names. Example from this piece of code:
- `json_file` -> `config_file_path`
- `file` -> `config_file`.

This way it's easier to avoid mistakes and the contents of the variables are clearer to the reader.

PawelWMS: And I'd in general avoid generic variable names like `file`, `data`, `json`, etc. With one exception for well-established conventions within a given language like `with(...) as f`.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,99 @@
...
+        required_configs_data = data[kernel]['required-configs']
+
+    # Extract the words from the string
+    config_words = find_matching_lines(config_diff)
+
+    # Find the missing words
+    missing_configs = []
+    for config_option in config_words:
+        if config_option not in required_configs_data or arch not in required_configs_data[config_option]["arch"] :
+            missing_configs.append(config_option)
```
PawelWMS, line 60: Small suggestion to shorten the line:
```python
            missing_configs += config_option
```

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,99 @@
...
+input_config_data = get_data_from_config(config_file)
+
+arch = check_config_arch(input_config_data)
+if arch == None:
+    print("Architecture not found in config file")
+    sys.exit(1)
+
+missing_configs = find_missing_configs(required_configs, kernel, arch, config_diff)
+if missing_configs == None:
+    print(f"Could not find required configs for {kernel}")
```
PawelWMS, line 88: The interface (comment) for `find_missing_configs` says nothing about how to interpret a result of `None`. Without such explicit information, the caller can at most say that something went wrong.

I'd suggest either expanding the comment for `find_missing_configs` to explicitly say that a result of `None` means that there were no required configs found for the provided kernel (and thus establish a more robust contract between the caller and the callee), or make this error message more generic and have the function take care of printing more details about the cause of the error.

.github/workflows/check_new_kernel_configs.py
```diff
@@ -0,0 +1,99 @@
...
+    sys.exit(0)
+
+print("Missing configs:", missing_configs)
+if len(missing_configs) == 0:
```
PawelWMS, line 92: Seems unnecessary if there were no missing configs. I think we can just remove this line, since we're going to print them later if, indeed, some configs were missing.

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,155 @@
...
+    with open(input_file, 'r') as file:
+        input_config_data = file.read()
+    return input_config_data
+
+def check_kernel(input_file):
+    match = re.search(r'SPECS/(.*?)/', input_file)
+    if match:
+        return match.group(1)
+    else:
+        return None
```
PawelWMS, line 33: It looks like you have some duplicated functions here and in the other script. I'd suggest placing them in a shared module, so we don't repeat ourselves. This also makes it harder to update one but forget the other.

.github/workflows/check_new_kernel_configs.py
rlmenge: I added additional output. LMK if this was along the lines of what you were thinking.

.github/workflows/check_new_kernel_configs.py
PawelWMS: Discussion note: finding configs and stripping unnecessary parts with a single regex.

toolkit/scripts/check_required_kernel_configs.py
PawelWMS: Discussion notes: removing re-parsing of the config.

toolkit/scripts/check_required_kernel_configs.py
PawelWMS: Erm, what I had in mind, is just making the field names visible, so the reader understands what they are for with more ease.:) I think my comment was too vague - I apologize.

toolkit/scripts/check_required_kernel_configs.py
PawelWMS: Discussion notes: low-pri, might want to consider to consolidate both scripts into one "kernel config checker".

toolkit/scripts/check_new_kernel_configs.py
```diff
@@ -0,0 +1,81 @@
...
+import argparse
+import sys
+import re
+from kernel_sources_analysis import get_data_from_config, extract_kernel_dir_name, extract_config_arch
+
+# Find the lines in the diff that contain +/- CONFIG
+# Ignore the lines that contain @
+# Return the set of configs that follow +/-
+def extract_modified_configs(input_string):
+    pattern = r'(?!.*@)[+-]\s*.*CONFIG.*'
```
PawelWMS, line 17: Try `^[-+](?:# )?(CONFIG_[^= ]+)`. It should grab the config names in the capture group.

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,78 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# This action checks that required kernel configs have not been removed or
+# modified to an undesirable value. It also checks that new configs added to
+# the kernel config files are in the list of required configs.
+name: Kernel Required Configs Check
+
+on:
```
oliviacrain, line 9: If it's in your time budget for this, I'd recommend adding a trigger for `workflow_dispatch`. This would allow you to [run the workflow manually](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually), say, in cases where you just want to check the validity of each config independent of any changes.

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,78 @@
...
+          echo "Merging ${{ github.sha }} into ${{ github.base_ref }}"
+
+      - name: Get base commit for Pushes
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          git fetch origin ${{ github.event.before }}
+          echo "base_sha=${{ github.event.before }}" >> $GITHUB_ENV
+          echo "Merging ${{ github.sha }} into ${{ github.event.before }}"
+
+      # For consistency, we use the same major/minor version of Python that CBL-Mariner ships
```
oliviacrain, line 43: Comment no longer relevant

.github/workflows/check-kernel-config.yml
oliviacrain: Bash is wizardry to me, no better suggestions here, I'd just recommend putting a comment explaining this so nobody retraces your steps trying to remove this variable.

toolkit/scripts/check_new_kernel_configs.py
```diff
@@ -0,0 +1,80 @@
...
+from kernel_sources_analysis import get_data_from_config, extract_kernel_dir_name, extract_config_arch
+
+# Regex for finding config options
+# Matches words that start with +/-CONFIG_ or +/-# CONFIG_
+CONFIG_REGEX=r'(?:(?<=[-+])|(?<=[-+]# ))CONFIG_\w+'
+
+def extract_modified_configs(input_string):
+    matching_lines = re.findall(CONFIG_REGEX, input_string)
+    config_set = set(matching_lines)
+    return config_set
```
PawelWMS, line 20: Nice update. Suggestion for a final touch:

```suggestion
# Regex for finding config options
# Matches words that start with +/-CONFIG_ or +/-# CONFIG_
CONFIG_REGEX=re.compile(r'(?:(?<=[-+])|(?<=[-+]# ))CONFIG_\w+')

def extract_modified_configs(input_string):
    return set(CONFIG_REGEX.findall(input_string))
```

or:

```suggestion
# Regex for finding config options
# Matches words that start with +/-CONFIG_ or +/-# CONFIG_
CONFIG_REGEX=re.compile(r'(?:(?<=[-+])|(?<=[-+]# ))CONFIG_\w+')

def extract_modified_configs(input_string):
    matching_configs = CONFIG_REGEX.findall(input_string)
    return set(matching_configs)
```

toolkit/scripts/check_new_kernel_configs.py
```diff
@@ -0,0 +1,80 @@
...
+# Usage: python3 check_new_kernel_configs.py --required_configs <path to json of required configs> --config_str <string of diff for config file>
+
+import json
+import argparse
+import sys
+import re
+from kernel_sources_analysis import get_data_from_config, extract_kernel_dir_name, extract_config_arch
+
+# Regex for finding config options
+# Matches words that start with +/-CONFIG_ or +/-# CONFIG_
```
PawelWMS, line 14: I'd suggest to also add a few lines of examples with lines that will match and the ones that won't. Usually that helps a lot with understanding regular expressions. Since we're using backward lookups here, I'd include also the final string that's going to end up in the match. Something along these lines
```
Examples:
"+CONFIG_ABC=y" -> "CONFIG_ABC"
"-# CONFIG_XYZ is not..." -> "CONFIG_XYZ"
```

toolkit/scripts/check_new_kernel_configs.py
```diff
@@ -0,0 +1,82 @@
...
+missing_configs = find_missing_configs(required_configs, kernel, arch, config_diff)
+if missing_configs == None:
+    print(f"Could not find required configs for {kernel}")
+    sys.exit(0)
+
+if len(missing_configs) == 0:
+    print("All configs are present in required configs")
+else:
+    print (f"====================== Kernel new config verification FAILED for {arch} ======================")
+    print(f"New configs detected for {arch}. Please add the following to toolkit/scripts/mariner-required-configs.json")
```
PawelWMS, line 79: Since we're passing `config_file` as input, I'd suggest either hard-coding it instead or replacing the hard-coded path in this message to `{config_file}`.

toolkit/scripts/kernel_sources_analysis.py
```diff
@@ -0,0 +1,31 @@
...
+
+import re
+
+def get_data_from_config(input_file):
+    with open(input_file, 'r') as file:
+        input_config_data = file.read()
+    return input_config_data
+
+def extract_kernel_dir_name(input_file):
+    match = re.search(r'SPECS/(.*?)/', input_file)
```
PawelWMS, line 15: >match = re.search(r'SPECS/(.*?)/', input_file) [](http://example.com/codeflow?start=4&length=46)

When I think about it now, using a regular expression here might be a bit of an overkill. Also, this now depends on the kernel spec being somewhere inside a directory called "SPECS", which makes this function more vulnerable to failures in case of changes in the future (a kernel spec placed under a different path).

If it is, however, required that all paths passed into this function contain the "SPECS" directory, then I agree this needs to be there. But in such case, I'd also add a check for that requirement.

Otherwise, I'd rely on a simple `os.path.basename(os.path.dirname(input_file))` (example: [here]([https://note.nkmk.me/en/python-os-basename-dirname-split-splitext/#:~:text=dirname()-,Use%20os.,name)%20from%20a%20path%20string.&text=If%20you%20want%20to%20get,above%20the%20file%2C%20use%20os](https://note.nkmk.me/en/python-os-basename-dirname-split-splitext/#:~:text=dirname()-,Use%20os.,name)%20from%20a%20path%20string.&text=If%20you%20want%20to%20get,above%20the%20file%2C%20use%20os).)).

Also, since the function will work (and I'm guessing is expected to work) for any file path, not only a path to a kernel spec, I think we can make the name more generic `extract_file_dir_name`.

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,134 @@
...
+# Script to check for required kernel configs in a kernel config file
+# Usage: python3 check_required_kernel_configs.py --required_configs <path to json of required configs> --config_file <path to config being checked>
+
+import json
+import argparse
+import sys
+from kernel_sources_analysis import get_data_from_config, extract_kernel_dir_name, extract_config_arch
+
+# Define a class
+    #incorrect configs is map: {config: (newValue, expectedValue, comment, PR)}
```
PawelWMS, line 13: Refactoring leftover? I'm not sure what this comment is about.

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,134 @@
...
+    # Calculate maximum width for each column
+    column_widths = [max(len(str(row[i])) for row in print_data) for i in range(len(print_data[0]))]
+
+    # Print columns
+    for j, row in enumerate(print_data):
+        for i, column in enumerate(row):
+            print(str(column).ljust(column_widths[i] + 2), end='')
+        print()
+        if j == 0:
+            print("-------------------------------------------------------------------------------")
```
PawelWMS, line 81: I don't know what's going on here, it's already past my bedtime, I trust we want this and the output is just stunning to behold.:)

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,134 @@
...
+import json
+import argparse
+import sys
+from kernel_sources_analysis import get_data_from_config, extract_kernel_dir_name, extract_config_arch
+
+# Define a class
+    #incorrect configs is map: {config: (newValue, expectedValue, comment, PR)}
+class IncorrectConfig:
+    def __init__(self, name, newValue, expectedValue, comment, PR):
+        self.name = name
```
PawelWMS, line 16: Do we ever use `name`?

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,134 @@
...
+        self.newValue = newValue
+        self.expectedValue = expectedValue
+        self.comment = comment
+        self.PR = PR
+
+def check_strings_in_file(json_file, kernel, arch, input_config_data):
```
PawelWMS, line 22: >check_strings_in_file [](http://example.com/codeflow?start=4&length=21)

I think we could be more specific in naming this function. I think some complex systems could be briefly described as "checking strings in files". What I'm getting at: we're not getting much info about the purpose and potential result of calling this function from the current name.;)

.github/workflows/check-kernel-config.yml
```diff
@@ -0,0 +1,77 @@
...
+# Licensed under the MIT License.
+
+# This action checks that required kernel configs have not been removed or
+# modified to an undesirable value. It also checks that new configs added to
+# the kernel config files are in the list of required configs.
+name: Kernel Required Configs Check
+
+on:
+  push:
+    branches: [main, dev, 1.0*, 2.0*]
```
oliviacrain, line 11: Probably don't need to trigger on `dev` branch

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,134 @@
...
+    with open(json_file, 'r') as req_config_json_file:
+        config_json_data = json.load(req_config_json_file)
```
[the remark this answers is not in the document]
rlmenge: One reads the contents as is while this loads it into json format. I created a new common function to help for this case :)

toolkit/scripts/check_required_kernel_configs.py
rlmenge: nope. Removed

toolkit/scripts/check_required_kernel_configs.py
rlmenge: Removed

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,134 @@
...
+                        break
+                # config was found but value is not correct
+                # mark as found and add to incorrect_configs
+                if not found:
+                    incorrect_configs[config_option] = (IncorrectConfig(config_option, line.split(config_option)[1].replace('=',''), value['value'], value['comment'], value['PR']))
+                    found = True
+        if not found:
+            # check if config can be missing
+            if "" not in value['value']:
+                incorrect_configs[config_option] = (IncorrectConfig(config_option, line, value['value'], value['comment'], value['PR']))
```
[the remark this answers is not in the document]
rlmenge: done

toolkit/scripts/check_new_kernel_configs.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # Find the missing configs
+    missing_configs = []
+    for config_option in config_set:
+        if config_option not in required_configs_data or arch not in required_configs_data[config_option]["arch"] :
+            missing_configs.append(config_option)
+    return missing_configs
+
+
+parser = argparse.ArgumentParser(
```
trungams, line 48: should this part of the code be moved to a main function, or under `if __name__ == '__main__':` block?

toolkit/scripts/check_new_kernel_configs.py
PawelWMS: Replying just to try and force to make GitHub show this comment again.:)

toolkit/scripts/kernel_sources_analysis.py
PawelWMS: Replying just to try and force to make GitHub show this comment again.:)

toolkit/scripts/kernel_sources_analysis.py
```diff
@@ -0,0 +1,46 @@
...
+def create_map_of_config_values(input_config_data):
+    config_map = {}
+    for line in input_config_data.split('\n'):
+        if "=" in line:
+            config_map[line.split('=')[0]] = line.split('=')[1]
+        # Find configs that are not set
+        # Example: # CONFIG_FOO is not set
+        elif "is not set" in line:
+            config_map[line.split()[1]] = "is not set"
+    return config_map
```
PawelWMS, line 46: **[Just having fun, no action needed/expected]**

Looks good, thanks for the update! But I wouldn't be myself, if I didn't suggest a regex alternative:

```suggestion
# Regex matching pairs of kernel config name and its value.
#
# Sample input:
#   CONFIG_ABC=y
#   CONFIG_XYZ=""
#   # CONFIG_111 is not set
#
# Result:
#   {
#     "CONFIG_ABC": "y",
#     "CONFIG_XYZ": "",
#     "CONFIG_111": "is not set",
#   }
CONFIG_TO_VALUE_REGEX=re.compile(r'(CONFIG_\w+)[ =](.*)')
def create_map_of_config_values(input_config_data):
    return dict(CONFIG_REGEX.findall(input_config_data))}
```

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,123 @@
...
+                }
+        # check if required configs removed from the kernel's config file
+        # Note that some required configs are required to be missing
+        elif "" not in req_value['value']: 
+                incorrect_configs[config_option] = {
+                    "newValue": "MISSING",
+                    "expectedValue": req_value['value'],
+                    "comment": f"Config not found.",
+                    "PR": req_value['PR']
+                }
```
PawelWMS, line 46: One thing I'd like to double-check: in the config JSON how do we encode configs with values set to empty strings? For example `CONFIG_DEFAULT_HOSTNAME`? Is it going to be `value: ['""', "some other valid value"]` and `value: [""]` is reserved for configs, which may be missing from the config file altogether?

toolkit/scripts/check_required_kernel_configs.py
```diff
@@ -0,0 +1,123 @@
...
+    else:
+        if result == {}:
+            print("All required configs are present")
+        else:
+            print()
+            print ("----------------- Kernel config verification FAILED -----------------")
+            for config_option, value in result.items():
+                print(f'{config_option} is "{result[config_option]["newValue"]}", expected {result[config_option]["expectedValue"]}.\nReason: {result[config_option]["comment"]}')
+                print(f'PR: {result[config_option]["PR"]}')
+                print()
```
PawelWMS, line 122: One more thing I remembered: if you want to do multi-line prints, you can also do this to avoid multiple `print` calls, using `\n`, and make the code reflect the console output a bit closer:
```suggestion
                print(f'''
{config_option} is "{result[config_option]["newValue"]}", expected {result[config_option]["expectedValue"]}.
Reason: {result[config_option]["comment"]}')
PR: {result[config_option]["PR"]}'

```

toolkit/scripts/check_required_kernel_configs.py
rlmenge: This is a really good question. If you want `CONFIG_X="" ` then the required value has to be `"\"\""` because json only recognises double quotes for values ("") and the config file uses double quotes ("").

toolkit/scripts/kernel_sources_analysis.py
rlmenge: Done. I also added the optional argument "kernel" to pass the kernel name explicitly to the program

toolkit/scripts/check_new_kernel_configs.py
rlmenge: nice catch! Done

toolkit/scripts/kernel_sources_analysis.py
rlmenge: Slight typo in this (CONFIG_TO_VALUE_REGEX vs CONFIG_REGEX) but otherwise works like a charm! I have commited it in your honor

toolkit/scripts/check_new_kernel_configs.py
```diff
@@ -0,0 +1,87 @@
...
+    config_file = args.config_file
+    config_diff = args.config_diff
+    if args.kernel:
+        kernel = args.kernel
+    else:
+        kernel = extract_kernel_dir_name(config_file)
+    if kernel == None:
+        print("ERROR: Kernel name not found. Please provide kernel name using --kernel flag or ensure config file path is correct")
+        sys.exit(1)
+    else:
```
PawelWMS, line 65: Nit-pick: no need for `else` at this point, since the `if` is going to exit anyway.

toolkit/scripts/kernel_sources_analysis.py
```diff
@@ -0,0 +1,53 @@
...
+        input_config_data = file.read()
+    return input_config_data
+
+def get_jsondata_from_jsonfile(input_json_file):
+    with open(input_json_file, 'r') as req_config_json_file:
+        config_json_data = json.load(req_config_json_file)
+    return config_json_data
+
+def extract_kernel_dir_name(input_file):
+    match =  os.path.basename(os.path.dirname(input_file))
```
PawelWMS, line 22: >   [](http://example.com/codeflow?start=11&length=2)

Nit-pick: extra space.

.github/workflows/check-kernel-config.yml
rlmenge: was having trouble with this. Will look into in the future

toolkit/scripts/kernel_sources_analysis.py
PawelWMS: Thank you for catching the typo and taking in the change!:)

christopherco (approved):
:shipit:

PR #5535, opened 2023-05-18
make RPM cache flat, improve tdnf arg formatting

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

When we place files into the RPM cache (`./build/rpm_cache`) we download them into the base directory. We then post-process the directory to split the packages by arch. This means that tdfn doesn't think it has already downloaded the rpm even though it is in the cache. While it IS able to see it in the cache it still duplicates the file.

Why don't we just leave all the .rpms in the base directory in the cache? This is not a user-accessed directory so the organization isn't important. Also, each .rpm will be marked with its architecture in the filename.

Also, recent versions of tdnf can handle version arguments such as `tdnf install "pkg-name <= 1.2.3-4"` etc. Previously we used the format `pkg-name=1.2.3-4`. The hard-coded package versions still work (i.e., packagelist.json: `pkg-name=1.2.3-4.cm2`).

The previous implementation was breaking on some edge cases were we wanted to find rpms especially for <, and <= during 'WhatProvides()` calls.

This is a precursor PR to implementing automatic downloading of delta rpms during a build.

###### Change Log  

- Make RPM caches flat instead of split by arch
- Allow tdnf to use the full `tdnf  install "<pkgname> <condition> <version>"`

**YES/NO**

###### Test Methodology

- Local builds
- ~https://dev.azure.com/mariner-org/mariner/_build/results?buildId=363115~ Failed due to unrelated issue in parent commits
- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=364004


toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -587,27 +574,25 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string, enabledRepoOrder ...stri
...
 	// TDNF does not accept versioning information on implicit provides.
 	if pkgVer.IsImplicitPackage() {
 		if pkgVer.Condition != "" {
 			logger.Log.Warnf("Discarding version constraint for implicit package: %v", pkgVer)
 		}
+		tdnfArg = pkgVer.Name
 		return
 	}
 
 	// Treat <= as =
```
oliviacrain, line 586: Comment is no longer relevant

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -412,21 +412,8 @@ func (r *RpmRepoCloner) WhatProvides(pkgVer *pkgjson.PackageVer) (packageNames [
...
+// Packages will be placed in a flat directory.
 func (r *RpmRepoCloner) ConvertDownloadedPackagesIntoRepo() (err error) {
-	srcDir := filepath.Join(r.chroot.RootDir(), chrootDownloadDir)
-	repoDir := srcDir
-
-	if !buildpipeline.IsRegularBuild() {
-		// Docker based build doesn't use overlay so repo folder
-		// must be explicitely set to the RPMs cache folder
-		repoDir = filepath.Join(r.chroot.RootDir(), cacheRepoDir)
-	}
```
oliviacrain, line 423: What's the rationale behind removing this?

oliviacrain (approved):

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -587,27 +574,25 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string, enabledRepoOrder ...stri
...
-	case "=":
-		tdnfArg = fmt.Sprintf("%s-%s", tdnfArg, pkgVer.Version)
-	case "":
-		break
+	case "<=", "<", "=":
+		tdnfArg = fmt.Sprintf("%s %s %s", pkgVer.Name, pkgVer.Condition, pkgVer.Version)
+	case "", ">", ">=":
+		tdnfArg = pkgVer.Name
 	default:
 		logger.Log.Warnf("Discarding '%s' version constraint for: %v", pkgVer.Condition, pkgVer)
```
PawelWMS, line 594: Are there any more valid options left for `default`? If not, with the current changes, I think this case should be treated as an error.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -587,27 +574,25 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string, enabledRepoOrder ...stri
...
-		logger.Log.Warnf("Treating '%s' version constraint as '=' for: %v", pkgVer.Condition, pkgVer)
-		fallthrough
-	case "=":
-		tdnfArg = fmt.Sprintf("%s-%s", tdnfArg, pkgVer.Version)
-	case "":
-		break
+	case "<=", "<", "=":
+		tdnfArg = fmt.Sprintf("%s %s %s", pkgVer.Name, pkgVer.Condition, pkgVer.Version)
+	case "", ">", ">=":
+		tdnfArg = pkgVer.Name
```
PawelWMS, line 592: I think we can simplify further by:
1. completely removing this case,
2. reverting back the removal of `tdnfArg = pkgVer.Name` at the beginning of this function, and
3. reverting the new `tdnfArg = pkgVer.Name` line inside `if pkgVer.IsImplicitPackage()`.

In short: it looks like the only time we want to use something different than the plain `pkgVer.Name` is for the `"<=", "<", "="` case.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -587,27 +574,25 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string, enabledRepoOrder ...stri
 }
 
 func convertPackageVersionToTdnfArg(pkgVer *pkgjson.PackageVer) (tdnfArg string) {
-	tdnfArg = pkgVer.Name
 	// TDNF does not accept versioning information on implicit provides.
 	if pkgVer.IsImplicitPackage() {
 		if pkgVer.Condition != "" {
 			logger.Log.Warnf("Discarding version constraint for implicit package: %v", pkgVer)
```
PawelWMS, line 580: Since we handle TDNF handles version constraints better now, do we even want to treat implicit packages differently? Maybe we can remove this whole `if` statement?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -587,27 +574,25 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string, enabledRepoOrder ...stri
...
 		}
+		tdnfArg = pkgVer.Name
 		return
 	}
 
 	// Treat <= as =
 	// Treat > and >= as "latest"
```
PawelWMS, line 587: Idea: since TDNF handles version constraints better now, we might as well just add them for the >= and > cases as well. This would have the advantage of consistency, so we might avoid some developer confusion during debugging.

Also, we could then have one comment less and people wouldn't need to wonder why these constraints are special when we re-visit this function in the future.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -412,21 +412,8 @@ func (r *RpmRepoCloner) WhatProvides(pkgVer *pkgjson.PackageVer) (packageNames [
...
-	srcDir := filepath.Join(r.chroot.RootDir(), chrootDownloadDir)
-	repoDir := srcDir
-
-	if !buildpipeline.IsRegularBuild() {
-		// Docker based build doesn't use overlay so repo folder
-		// must be explicitely set to the RPMs cache folder
-		repoDir = filepath.Join(r.chroot.RootDir(), cacheRepoDir)
-	}
-
-	err = rpmrepomanager.OrganizePackagesByArch(srcDir, repoDir)
```
PawelWMS, line 425: This is the only place `OrganizePackagesByArch` is used. We can remove that function as well now. No need to keep it, if we're not using it.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: The logic to pick a specific version chokes very badly if we offer too many choices (since it downloads them all, then picks one). I made this change to work around a specific rpm resolution issue for the delta downloader, but when I noticed it started downloading a lot more things. So this was a concious choice to only fix what I needed. 

If we want to commit to improving the performance of the WhatProvides()... et. al. functions I'm happy to make it general (like it probably should be).

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: If you mean the implicit provides comment, tdnf still completely ignores it unfortunately.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Assuming we do the optimization you suggested we will still hit the default case for '>' and '>=' so I think its good to keep.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Updated

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Fair point, I've cleaned it up.

dmcilvaney:
@PawelWMS, @oliviacrain fyi I merged in some changes from main. I integrated some code form @anphel31 to do rpm path validation. I believe it should have the same result still.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: I've split out so we have both the warning for removal, and an error for unsupported.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: See above, I think I've settled on a compromise.

toolkit/tools/internal/packagerepo/repomanager/rpmrepomanager/rpmrepomanager.go
```diff
@@ -46,65 +45,48 @@ func CreateRepo(repoDir string) (err error) {
...
 }
 
-// OrganizePackagesByArch will recursively move RPMs from srcDir into architecture folders under repoDir
-func OrganizePackagesByArch(srcDir, repoDir string) (err error) {
-	const noArch = "noarch"
-
-	logger.Log.Debugf("Organizing RPM packages from (%s) to (%s)", srcDir, repoDir)
-
-	stdout, stderr, err := shell.Execute("uname", "-m")
+// ValidateRpmPaths d
```
oliviacrain, line 48: Incomplete doc comment?

dmcilvaney:
```suggestion
// ValidateRpmPaths checks for any rpm filenames that don't match the expected output according to 'rpm -qp ...'.  It
// will return an error with all the mismatched pairs if it finds any.
```

dmcilvaney: Thanks, fixed

anphel31 (approved):

toolkit/tools/internal/packagerepo/repomanager/repomanager.go
```diff
@@ -7,7 +7,5 @@ package repomanager
 type RepoManager interface {
 	// CreateRepo will create a package repository at repoDir
 	CreateRepo(repoDir string) error
-
-	// OrganizePackagesByArch will move packages in flatDir into architecture folders under repoDir
-	OrganizePackagesByArch(flatDir, repoDir string) error
+	ValidateRpmPaths(repoDir string) error
```
dmcilvaney, line 10:
```suggestion

	// ValidateRpmPaths will check if all .rpm files have the correct name as returned from an rpm query
	ValidateRpmPaths(repoDir string) error
```

PR #5570, opened 2023-05-23
Unified `*_LIST` arguments usage to accept spec names

PawelWMS:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

The change makes sure that all of the following tools arguments can uniformly work with spec file names:
- `PACKAGE_BUILD_LIST`,
- `PACKAGE_IGNORE_LIST`,
- `PACKAGE_REBUILD_LIST`,
- `SRPM_PACK_LIST`.

In Mariner 2.0 accepted values for these arguments diverged, where some worked with **spec file** names and the others with **package** names. This caused a lot of grief and confusion for our user in case of specs, which didn't produce the default package (example `python-werkzeug.spec` building `python3-werkzeug`). In addition to that, it was impossible to provide a single set of packages to (re)build for our buddy builds, since `python3-werkzeug` would break the SRPM packing stage, while `python-werkzeug` would break the RPM building stage.

We can only reliably require all arguments to accept spec file names due to the fact, that we don't know what the produced packages names are going to be until we parse all specs. The `SRPM_PACK_LIST` is the input to the spec parsing step, thus it cannot rely on package names.

Since the dependency graph operates on **package** names, if the value provided in the `PACKAGE_*_LIST` argument doesn't match a package, we will now retrieve all packages built by a given spec and consider that whole set as the input.

###### Change Log  

- Unified use of the `*_LIST` arguments.
- Updated the documentation.
- Small fix in `pkggen.mk` to recognize changes in the `PACKAGE_IGNORE_LIST` argument.
- Added a matched for `PackageVer` to the `sliceutils` package.
- Updated `pkggraph.ReadDOTGraphFile`, so it now doesn't require the caller to always manually create an empty graph.

No.

###### Test Methodology

- Local package build tests.
- [Full build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=366132&view=results).


toolkit/tools/internal/sliceutils/sliceutils.go
```diff
@@ -37,11 +41,40 @@ func FindMatches(slice []string, isMatch func(string) bool) []string {
...
 func StringMatch(expected, given interface{}) bool {
+	if checkValid, checkResult := nilCheck(expected, given); checkValid {
+		return checkResult
+	}
+
 	return expected.(string) == given.(string)
 }
 
+// PackageVerMatch is intended to be used with "Contains" and "Find" for slices of *pkgjson.PackageVers.
+func PackageVerMatch(expected, given interface{}) bool {
```
dmcilvaney, line 54: Huh, I'm actually surprised we don't have something like this. Could this live in pkgjson.go as a method of the packagever type maybe?

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1079,8 +1079,9 @@ func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer,
 		} else {
 			logger.Log.Warnf("Could not goal package %+v", pkg)
 			if strict {
-				logger.Log.Warnf("Missing %+v", pkg)
+				logger.Log.Errorf("Missing %+v", pkg)
 				err = fmt.Errorf("could not find all goal nodes with strict=true")
+				return
```
dmcilvaney, line 1084: I'll miss the lovely work around we could get by flooding the sets with enough valid hits so we were probabilistically likely to not get an error here...

(Good riddens)

toolkit/tools/scheduler/schedulerutils/buildlist.go
```diff
@@ -21,84 +23,73 @@ import (
...
+			for _, pkg := range nodes {
+				packageVersMap[pkg.VersionedPkg] = true
+			}
 		} else {
-			logger.Log.Debugf("Found external package to filter out: %v.", pkgVer)
+			foundNode, err := dependencyGraph.FindBestPkgNode(&pkgjson.PackageVer{Name: packageOrSpecName})
+			if err != nil {
+				logger.Log.Errorf("Failed while searching the dependency graph for package '%s', error: %s", packageOrSpecName, err)
+				return nil, err
+			}
```
dmcilvaney, line 74: I've been thinking we should use wrapped errors more, there are several spots in the code were we completely lose our error state and overwrite it. (We also can compare against the wrapped errors, although we don't really do that much in our code since generally any error is effectively fatal) Thoughts?
```suggestion
			if err != nil {
				err = fmt.Errorf("failed while searching the dependency graph for package '%s': %w", packageOrSpecName, err)
				return nil, err
			}
```

toolkit/tools/internal/sliceutils/sliceutils.go
PawelWMS: Well this package is for working on slices but matchers can live closer to what they are supposed to match. In other words: I see it as reasonable to move it to `pkgjson.go`. I'll take a closer look in a minute.

PawelWMS: OK, I'll leave the matcher inside `sliceutils` for now. Moving that matcher to `pkgjson` creates a cyclic dependency, since we use the `nilCheck` function in the matcher.

Also, I want to keep all of the `*SetToSlice` functions inside `sliceutils`, so they are all inside one package, so we'd be left with a dependency on `pkgjson` anyway. Bear in mind, that this part is supposed to be temporary - once we upgrade to Goland 1.18+, I'll replace these functions with a generic `SetToSlice[T...]` and we'll be able to drop this weird dependency.

Once the generics are there, then we will no longer have `sliceutils` depend on other packages and we will be free to distribute the matchers to sit next to the types they are meant to match.

toolkit/tools/scheduler/schedulerutils/buildlist.go
PawelWMS: In general I don't really mind the way we print errors in most places right now - the "stacked" approach. It is similar to printing a call stack and we don't run into a risk of printing one, insanely long error message, which can make it hard to read. 

As far as I remember, we intentionally don't try to parse the errors and then decide how to treat them. This would have the effect of producing exception-/error-driven code (not to be confused with exception-driven development), which doesn't help with code clarity. In general, I'm strongly enforcing having as few ignorable errors as possible, since this way the developer doesn't have to consider the possibility if a returned error is something they can tolerate or not. Instead, it is a clear message that something went wrong and we have to bail.

dmcilvaney: Agreed about not trying to parse the errors if at all possible. I've just noticed a lot of places where we do
```
if err != nil {
    // Forgot to write a logger.Log.Errorf()
    return err
}
```
Or some variation. The problem I have with using the print statements is if we aren't super verbose we can lose track of where the caller is. Not saying %w is a magic bullet but I like it more so far.

dmcilvaney (approved):

toolkit/tools/scheduler/schedulerutils/buildlist.go
PawelWMS: OK, now I think I better understand what you had in mind when you mentioned the unwrapped errors. I think I also like the idea of wrapping error, let me see how this may looks like here.

PR #5605, opened 2023-05-31
Add specarchchecker tool to allow arch validation

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add a tool that can be used to check if the build tools will evaluate a given `.spec` name as buildable or not for the current architecture. It uses a simple chroot based tool to do this. The tool can be invoked directly after calling `make go-specarchchecker`, or automatically by calling `make run-specarchchecker REBUILD_TOOLS=y PACKAGE_BUILD_LIST="..." PACKAGE_REBUILD_LIST="..."`. The make command will place the results into `./build/valid_arch_spec_names/valid_arch_spec_names.txt`

It processes values passed in via `PACKAGE_BUILD_LIST` and `PACKAGE_REBUILD_LIST` and treats them as <name>.spec. It will not process other styles of matching (ie .rpm or subpackage names).

Invoke it like: `make run-specarchchecker REBUILD_TOOLS=y PACKAGE_BUILD_LIST="shim-unsigned-aarch64"` on an x86_64 system and `./build/valid_arch_spec_names/valid_arch_spec_names.txt` will be empty, but on an aarch64 system it will include `shim-unsigned-aarch64`

###### Change Log  

- Break out some common spec handling functionality into `./toolkit/tools/pkg/simplechroottool.go`
  - Refactor rpmssnapshot to use this package
- Add specarchchecker that also uses this package

**NO**

###### Test Methodology
- Local
- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=370793&view=results


dmcilvaney:
See also:
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=370793&view=results
https://dev.azure.com/mariner-org/mariner/_git/CBL-Mariner-Pipelines/pullrequest/14728

toolkit/scripts/toolkit.mk
```diff
@@ -26,6 +26,10 @@ rpms_snapshot_build_dir = $(BUILD_DIR)/$(rpms_snapshot_dir_name)
 rpms_snapshot_logs_path = $(LOGS_DIR)/$(rpms_snapshot_dir_name)/rpms_snapshot.log
 rpms_snapshot_per_specs = $(rpms_snapshot_build_dir)/$(specs_dir_name)_$(rpms_snapshot_name)
 
+valid_arch_spec_names_build_dir = $(BUILD_DIR)/valid_arch_spec_names
+valid_arch_spec_names           =  $(valid_arch_spec_names_build_dir)/valid_arch_spec_names.txt
```
anphel31, line 30: nit: spacing

anphel31 (approved):

toolkit/scripts/toolkit.mk
```diff
@@ -101,3 +105,21 @@ $(rpms_snapshot_per_specs): $(go-rpmssnapshot) $(chroot_worker) $(local_specs) $
 
 print-build-summary:
 	sed -E -n 's:^.+level=info msg="Built \(([^\)]+)\) -> \[(.+)\].+$:\1\t\2:gp' $(LOGS_DIR)/pkggen/rpmbuilding/* | tee $(LOGS_DIR)/pkggen/build-summary.csv
+
+# This will run the spec arch checker and generate a list of valid spec names suitable for passing to PACKAGE_BUILD_LIST et. al.
+.PHONY: prepare-spec-arch-checker
```
PawelWMS, line 110: I think `prepare-spec-arch-checker` is not a target defined anywhere in the makefiles.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
```diff
@@ -0,0 +1,146 @@
...
+	WorkerTarPath string
+	chroot        *safechroot.Chroot
+	chrootSpecDir string
+}
+
+// Chroot returns the chroot environment we are working in, call Chroot().Run() to execute commands inside the chroot. Call
+// InitializeChroot() before calling this function.
+func (s *SimpleChrootTool) Chroot() *safechroot.Chroot {
+	return s.chroot
+}
```
PawelWMS, line 40: Let's not expose "internals". Let's maybe go directly to `RunInChroot` or something.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
```diff
@@ -0,0 +1,146 @@
...
+	const leaveFilesOnDisk = false
+
+	if s.chroot != nil {
+		s.chroot.Close(leaveFilesOnDisk)
+	}
+}
+
+// BuildDefines creates a map of RPM build defines for the crhoot. This function should be called after InitializeChroot()
+// and from a context inside a call to the chroot's .Run() function.
+func (s *SimpleChrootTool) BuildDefines(distTag string) map[string]string {
```
PawelWMS, line 84: >BuildDefines [](http://example.com/codeflow?start=27&length=12)

`DefaultDefines`.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
```diff
@@ -0,0 +1,146 @@
...
+	const runCheck = true
```
PawelWMS, line 85: Let's move enabling/disabling checks to the init.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
```diff
@@ -0,0 +1,146 @@
...
+	defines := rpm.DefaultDefines(runCheck)
+	defines[rpm.DistTagDefine] = distTag
+
+	return defines
+}
+
+// BuildCompatibleSpecsList builds a list of spec files in the chroot's SPECs directory that are compatible with the build arch. Paths
+// are relative to the chroot's base directory. This function should be called after InitializeChroot() and from a context inside a call
+// to the chroot's .Run() function.
+func (s *SimpleChrootTool) BuildCompatibleSpecsList(inputSpecPaths []string, defines map[string]string) (filteredSpecPaths []string, err error) {
```
PawelWMS, line 96: >, defines map[string]strin [](http://example.com/codeflow?start=75&length=26)

For public functions, let's remove the `defines` argument and add an optional `SetDefines` method.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
```diff
@@ -0,0 +1,146 @@
...
+
+// SimpleChrootTool is a tool for creating a basic spec/rpm parsing environment inside a chroot. When creating a new instance, the
+// following fields must be set:
+//   - BuildDirPath: The path to the directory where the chroot will be created and run from.
+//   - WorkerTarPath: The path to the tarball containing the chroot environment.
+// The following fields will be set by InitializeChroot():
+//   - chroot: The chroot environment we are working in, call Chroot.Run() to execute commands inside the chroot.
+//   - chrootSpecDir: The path to the SPECS directory inside the chroot "/SPECS."
+
+type SimpleChrootTool struct {
```
PawelWMS, line 29: From offline discussion: think what we want to achieve with the new tool to not make it so generic, that it is duplicate of `safechroot` but also not limit its purpose to narrowly.

One idea you mentioned: separate the `BuildCompatibleSpecsList` function into a separate module and make this become just a way to define a more robust chroot environment with pre-defined mount points and directories for general use.

Suggestion: in this case it might be possible that the default defines may need to moved out along with `BuildCompatibleSpecsList`, since they will no longer make sense for a general-purpose chroot.

toolkit/scripts/toolkit.mk
dmcilvaney: Oops, missed the rename on that.

toolkit/scripts/toolkit.mk
```diff
@@ -101,3 +105,21 @@ $(rpms_snapshot_per_specs): $(go-rpmssnapshot) $(chroot_worker) $(local_specs) $
...
+run-specarchchecker: $(valid_arch_spec_names)
+	@cat $(valid_arch_spec_names) && echo "" # File doesn't have a newline at the end, so add one via echo.
```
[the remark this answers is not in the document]
dmcilvaney: I think it will be handy for debugging build logs in the pipelines. Makes it easier to figure out why the pipeline did what it did.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
dmcilvaney: With moving the spec functions to a new package this no longer makes sense. Will need to pass along.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
dmcilvaney: I'm thinking that part of this well-defined chroot is the `distTag` and `runChecks` state. I'm thinking we leave the default defines as part of the chroot wrapper package since it needs to track those.

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
```diff
@@ -53,42 +52,38 @@ const (
...
-// New creates an unitialized RPMs snapshot generator.
-func New(buildDirPath, workerTarPath string) *SnapshotGenerator {
-	return &SnapshotGenerator{
-		buildDirPath:  buildDirPath,
-		workerTarPath: workerTarPath,
-	}
+// New creates a new snapshot generator.
+func New(buildDirPath, workerTarPath, specsDirPath, distTag string) (newSnapshotGenerator *SnapshotGenerator, err error) {
+	newSnapshotGenerator = &SnapshotGenerator{}
+	err = newSnapshotGenerator.InitializeChroot(buildDirPath, chrootName, workerTarPath, specsDirPath, distTag, runChecks)
```
PawelWMS, line 67: If we're initializing the chroot here, then we'd need to also expose a `CleanUp` function for the caller. Alternatively, we could still keep the init and clean-up inside `GenerateSnapshot` but I see pros and cons of both without any strong preferences.

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
```diff
@@ -53,42 +52,38 @@ const (
...
 }
 
-// New creates an unitialized RPMs snapshot generator.
-func New(buildDirPath, workerTarPath string) *SnapshotGenerator {
-	return &SnapshotGenerator{
-		buildDirPath:  buildDirPath,
-		workerTarPath: workerTarPath,
-	}
+// New creates a new snapshot generator.
+func New(buildDirPath, workerTarPath, specsDirPath, distTag string) (newSnapshotGenerator *SnapshotGenerator, err error) {
```
PawelWMS, line 65: > specsDirPath, distTag string [](http://example.com/codeflow?start=37&length=29)

Comment related to the one about the init-clean-up one: if we init here and already pass the spec path and dist tag, then `GenerateSnapshot` shouldn't need these arguments and we can remove them from that method. And, similarly, we can also move everything back to `GenerateSnapshot`. Either way sounds good to me.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
```diff
@@ -0,0 +1,103 @@
...
+// ChrootRootDir returns the root directory where the chroot was created. Call InitializeChroot() before calling this function.
+func (s *SimpleChrootTool) ChrootRootDir() string {
+	if s.chroot == nil {
+		return ""
+	}
+	return s.chroot.RootDir()
+}
+
+// ChrootSpecDir returns the directory inside the chroot where the specs dir is mounted. Call InitializeChroot() before calling this function.
+func (s *SimpleChrootTool) ChrootRelativeSpecDir() string {
```
PawelWMS, line 43: >ChrootRelativeSpecDir [](http://example.com/codeflow?start=27&length=21)

Nit-pick: I guess `ChrootSpecDir` would already be clear enough. The word "relative" might suggest that the path is a relative one within the context of the chroot.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
```diff
@@ -0,0 +1,103 @@
...
+
+// SimpleChrootTool is a tool for creating an environment inside a chroot suitable for basic tasks like parsing RPMs.
+// The following fields will be set by InitializeChroot():
+//   - chroot: The chroot environment we are working in, call SimpleChrootTool.RunInChroot() to execute commands inside the chroot.
+//   - chrootSpecDir: The relative path to the SPECS directory inside the chroot: "/SPECS."
+//   - runChecks: Whether or not to parse with check sections enabled.
+//   - distTag: The distribution tag to use when parsing RPMs.
+type SimpleChrootTool struct {
+	chroot        *safechroot.Chroot
+	chrootSpecDir string
```
PawelWMS, line 29: Is this needed as a separate member, since we're using the constant? Do we want it to be settable somehow?

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
```diff
@@ -53,42 +52,38 @@ const (
...
+
+	chrootName = "rpmssnapshot_chroot"
+	runChecks  = false
 )
 
 type SnapshotGenerator struct {
-	chroot        *safechroot.Chroot
-	buildDirPath  string
-	workerTarPath string
+	simplechroottool.SimpleChrootTool
```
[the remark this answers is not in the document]
PawelWMS: Alternatively, if we want to make clear that it's an extension to `SimpleChrootTool`, we could do `type SnapshotGenerator simplechroottool.SimpleChrootTool`. But I think that's not the case - `simplechroottool.SimpleChrootTool` is a pre-defined chroot environment, while the generator is conceptually more than just an environment.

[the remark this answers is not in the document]
PawelWMS: One more thing: I think the generator has no reason to expose the `SimpleChrootTool` public functions. It's just supposed to be usable for the sole purpose of generating a snapshot. The user shouldn't be able to call arbitrary functions inside the chroot the generator is using. Similarly, they should have no need to access the chroot's root itself.

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
dmcilvaney: Good callout, these were only being used for logging so I just moved the log calls outside the function.

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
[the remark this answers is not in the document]
dmcilvaney: While it's not a super common flow in our code base as far as I know this is a rasonablly common go idiom of adding an interface to a struct.

I'm not super tied to this, but my understanding is that this is fairly common go coding practice. The `type SnapshotGenerator simplechroottool.SimpleChrootTool` approach isn't very flexible since you can't add more context to the `SnapshotGenerator` type anymore, so I'm not a fan of that one. 

All that said, I don't mind moving them to an explicit member of the struct rather than an interface. I don't think it will be functionally different here.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
dmcilvaney: Open to naming suggestions. I want to be clear that this is the '/SPECS/' part of '/path/to/chroot/**SPECS**/' and not the rest of the root folder.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/rpm"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safechroot"
+)
+
+const (
+	chrootSpecDirPath = "/SPECS"
+)
+
+// SimpleChrootTool is a tool for creating an environment inside a chroot suitable for basic tasks like parsing RPMs.
```
PawelWMS, line 21: >SimpleChrootTool [](http://example.com/codeflow?start=3&length=16)

Idea: `SimpleToolsChroot`.

toolkit/tools/pkg/simplechroottool/simplechroottool.go
```diff
@@ -0,0 +1,101 @@
...
+func (s *SimpleChrootTool) RunInChroot(toRun func() error) (err error) {
+	if s.chroot == nil {
+		return fmt.Errorf("chroot has not been initialized")
+	}
+	return s.chroot.Run(toRun)
+}
+
+// DefaultDefines creates a map of RPM build defines for the chroot. This function should be called after InitializeChroot()
+// and from a context inside a call to the chroot's .Run() function.
+func (s *SimpleChrootTool) DefaultDefines() map[string]string {
```
PawelWMS, line 96: Let's move this to `rpm.go`.

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
```diff
@@ -53,42 +52,33 @@ const (
...
+
+	chrootName = "rpmssnapshot_chroot"
+	runChecks  = false
 )
 
 type SnapshotGenerator struct {
-	chroot        *safechroot.Chroot
-	buildDirPath  string
-	workerTarPath string
+	SimpleChrootTool simplechroottool.SimpleChrootTool
```
PawelWMS, line 61: >SimpleChrootTool simplechroottool.SimpleChrootTool [](http://example.com/codeflow?start=1&length=50)

Make private, add `CleanUp`.

toolkit/tools/pkg/simpletoolchroot/simpletoolchroot.go
```diff
@@ -0,0 +1,87 @@
...
+	return
+}
+
+// CleanUp cleans up the chroot environment. This function should be called after all other functions in this package have been
+// called (likely in a defer statement)
+func (s *SimpleToolChroot) CleanUp() {
+	const leaveFilesOnDisk = false
+
+	if s.chroot != nil {
+		s.chroot.Close(leaveFilesOnDisk)
```
PawelWMS, line 77: >s.chroot.Close(leaveFilesOnDisk) [](http://example.com/codeflow?start=2&length=32)

Let's return the error from `Close` and leave it up to the caller to ignore it or not.

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
```diff
@@ -53,42 +52,38 @@ const (
...
 }
 
-// GenerateSnapshot generates a snapshot of all packages built from the specs inside the input directory.
-func (s *SnapshotGenerator) GenerateSnapshot(specsDirPath, outputFilePath, distTag string) (err error) {
-	err = s.initializeChroot(specsDirPath)
-	if err != nil {
-		return
-	}
-	defer s.cleanUp()
+// New creates a new snapshot generator.
```
PawelWMS, line 64: Maybe let's also explicitly mention that `CleanUp` must be called if this returns successfully.

toolkit/tools/pkg/specarchchecker/specarchchecker.go
```diff
@@ -0,0 +1,101 @@
...
+// buildAllSpecsListFromNames builds a list of all spec file paths from a list of spec names. Paths are relative to the SPECS
+// directory in the chroot.
+func (a *ArchChecker) buildAllSpecsListFromNames(specNames []string) (specPaths []string, err error) {
+	for _, specName := range specNames {
+		var fullSpecPath []string
+		specFilesGlob := filepath.Join(a.simpleToolChroot.ChrootRelativeSpecDir(), "**", fmt.Sprintf("%s.spec", specName))
+
+		fullSpecPath, err = filepath.Glob(specFilesGlob)
+		if err != nil {
+			err = fmt.Errorf("failed while trying to enumerate all spec files with (%s). Error: %w", specFilesGlob, err)
```
PawelWMS, line 64: >(%s). Error: %w", specFilesGlob, err) [](http://example.com/codeflow?start=74&length=37)

Let's switch all `Error: %w` to `Error:\n%w`.

toolkit/tools/pkg/simpletoolchroot/simpletoolchroot.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+// Run executes a given function inside the tool's chroot environment.
```
PawelWMS, line 81: >Run [](http://example.com/codeflow?start=3&length=3)

Refactoring leftover: the comment should now start with "RunInChroot".

toolkit/tools/pkg/simpletoolchroot/simpletoolchroot.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// InitializeChroot initializes the chroot environment so .RunInChroot() can be used to execute commands inside the chroot. This function
+// should be called before any other functions in this package. CleanUp() must be called (likely via defer) after InitializeChroot() is used.
+//   - buildDir: The path to the directory where the chroot will be created
+//   - chrootName: The name of the chroot to create
+//   - workerTarPath: The path to the tar file containing the worker files
+//   - specsDirPath: The path to the directory containing the spec files
+//   - distTag: The distribution tag to use when parsing RPMs
+//   - runChecks: Whether or not to parse with check sections enabled
```
PawelWMS, line 50: Let's remove the `distTag` and `runChecks` arguments from this description.

toolkit/tools/pkg/specarchchecker/specarchchecker.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+type ArchChecker struct {
+	simpleToolChroot simpletoolchroot.SimpleToolChroot
+}
+
+const (
+	chrootName = "specarchchecker_chroot"
+	runChecks  = false
+)
```
PawelWMS, line 25: For consts, that are used only in one function, could we move them into the scope of the function and out of the global scope?

PawelWMS (approved):
:shipit:

toolkit/tools/pkg/specarchchecker/specarchchecker.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/rpm"
+	simpletoolchroot "github.com/microsoft/CBL-Mariner/toolkit/tools/pkg/simpletoolchroot"
+)
+
+type ArchChecker struct {
+	simpleToolChroot simpletoolchroot.SimpleToolChroot
+}
+
+const (
+	chrootName = "specarchchecker_chroot"
```
PawelWMS, line 23: >chrootName = "specarchchecker_chroot" [](http://example.com/codeflow?start=1&length=37)

Btw. `chrootName` can now also be a local constant.

toolkit/tools/pkg/rpmssnapshot/rpmssnapshot.go
```diff
@@ -53,42 +52,37 @@ const (
 	rpmSpecBuiltRPMRegexDistributionIndex
 	rpmSpecBuiltRPMRegexArchitectureIndex
 	rpmSpecBuiltRPMRegexMatchesCount
+
+	chrootName = "rpmssnapshot_chroot"
```
PawelWMS, line 56: This one may also be made function-local.

Btw. the constants above are global only for the sake of keeping them closer to the regular expression, so it's easier to see which capture groups the indeces refer to. I keep going back and forth on what's the best place to keep such consts.

PR #5607, opened 2023-05-31
Add flag to disable PMC repository

trungams:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

- The `REPO_LIST` variable defined in our makefile is used to specify package repositories to be used by `tdnf` in addition to existing PMC repository. Add an option to explicitly disable PMC so that `tdnf` will only use the repos specified in `REPO_LIST`.

###### Change Log  

- Add new flag in Makefile `DISABLE_MARINER_REPOS` which defaults to false
- Add logic to `rpmrepocloner` to disable PMC repos when the flag is passed

**NO**

###### Test Methodology

- Local build
  * Create a local file `custom.repo`, which is a copy of https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS/mariner-repos/mariner-official-base.repo
  * Run a make command to build packages or create image with `DISABLE_MARINER_REPOS=y`. The command will fail because `tdnf` cannot find some packages.
  * Run the same command with an additional variable `REPO_LIST="/path/to/custom.repo"`. The build command should work.


toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -109,6 +111,11 @@ func (r *RpmRepoCloner) Initialize(destinationDir, tmpDir, workerTar, existingRp
 		logger.Log.Info("Enabling preview repo")
 	}
 
+	r.disableMarinerRepos = disableMarinerRepos
+	if disableMarinerRepos {
+		logger.Log.Info("Disabling Mariner repos")
```
dmcilvaney, line 116: Thoughts about making this a warning, along with some details about what repos we are disabling?
Something like `logger.Log.Warnf("Disabling Mariner repos (repo ids:  %s", marinerReposGlob)`?

dmcilvaney:
Can you add an entry to the end of building.md for this arg?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
trungams: That's fair, but this might require some extra work to get the full names of the repo ids. Either we have to parse the final repo file or parse the output of `tdnf repolist`, which isn't nicely formatted (?)

toolkit/tools/internal/packagerepo/repocloner/repocloner.go
```diff
@@ -22,7 +22,7 @@ type RepoPackage struct {
 // It is capable of generate a local repository consisting of a set of request packages
 // and their dependencies.
 type RepoCloner interface {
-	Initialize(destinationDir, tmpDir, workerTar, existingRpmsDir, toolchainRpmsDir string, usePreviewRepo bool, repoDefinitions []string) error
+	Initialize(destinationDir, tmpDir, workerTar, existingRpmsDir, toolchainRpmsDir string, usePreviewRepo bool, repoDefinitions []string, disableMarinerRepos bool) error
```
PawelWMS, line 25: Small request: could you re-order the arguments, so each type is grouped together and we can thus not repeat `bool`?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
trungams: Added this warning. The message on an example build is like this
```
WARN[0008] Disabling Mariner repos: [mariner-official-base mariner-official-base-debuginfo mariner-official-base-source]
```

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -24,10 +25,13 @@ import (
 
 const (
 	allRepoIDs             = "*"
+	marinerReposGlob       = "mariner*"
```
0xba1a, line 28: This method disables any repository passed using REPO_LIST but starting "mariner". For example if I use this flag and pass `mariner-extended`, it will be disabled.

0xba1a (approved):

0xba1a (changes requested):
This disables any repo that starts with mariner. Can you explicitely specify the default repos used by toolkit?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Just had a thought... could we call `rpm -e mariner-repos-*` or something similar? Just fully remove the repo files that way... The user call always explicitly re-add any repo file they wish that way.

PawelWMS: This might have unexpected consequences due to removal of `mariner-repos-shared` (package holds GPG keys, which may be needed). But I think @0xba1a's main concern is to make sure we remove only the things we need. My one idea would be to modify our logic, where for `DISABLE_MARINER_REPOS=y` we first remove all .repo files except for the ones built into the chroot (the ones representing the local caches) and then proceed to add the ones specified by the user.

What do you think?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -56,6 +60,14 @@ var (
 	//   - version:         1.1b.8_X-22~rc1
 	//   - dist:            cm1
 	listedPackageRegex = regexp.MustCompile(`^\s*([[:alnum:]_+-]+)\.([[:alnum:]_+-]+)\s+([[:alnum:]._+~-]+)\.([[:alnum:]_+-]+)`)
+
+	// The line specifying a yum repo ID is enclosed in square brackets, and can only contain the following characters, see `man dnf.conf(5)`:
```
PawelWMS, line 64: Please also provide examples similar to what we did for the regular expressions above.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -181,6 +194,21 @@ func (r *RpmRepoCloner) Initialize(destinationDir, tmpDir, workerTar, existingRp
...
 
+	r.disableMarinerRepos = disableMarinerRepos
+	if r.disableMarinerRepos {
+		repoList, err := r.getRepoList()
+		if err != nil {
+			return err
+		}
+		disabledMarinerRepos := make([]string, 0)
+		for _, repoID := range repoList {
+			if matched, err := filepath.Match(marinerReposGlob, repoID); err == nil && matched {
```
PawelWMS, line 205: Please error out if `err != nil`.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -181,6 +194,21 @@ func (r *RpmRepoCloner) Initialize(destinationDir, tmpDir, workerTar, existingRp
 		return
 	}
 
+	r.disableMarinerRepos = disableMarinerRepos
+	if r.disableMarinerRepos {
```
PawelWMS, line 198: I think we may be able to simplify by modifying `initializeRepoDefinitions`. That function already gathers the user-provided and built-in repos into one file. All we'd need to do, is skip adding the built-in repos whenever `r.disableMarinerRepos` is true. In other words: skip the `for _, originalRepoFilePath := range existingRepoFiles` loop.

This way we wouldn't need to parse the repo files at all, I believe.

**NOTE**: the `repoDefinitions` variable already contains the built-in repo files `fetcher.repo` and `local.repo`, which are required for the chroot to function properly with the local caches.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: Update: I think we may both simplify and make sure we include the right repos by only modifying the `initializeRepoDefinitions` function. More details in my comment around `if r.disableMarinerRepos`.

0xba1a: Instead of removing the repos based on re, can we introduce a flag (`DISABLE_REPO_LIST`) using which the user can pass repo-ids she wants to disable explicitly?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
trungams: I think we'd have to remove the existing repo files instead of skipping the copy to `allrepos.repo`. This approach makes sense if there's a guarantee that the existing repo files consist only of PMC repos

trungams: also, the `getRepoList` function is only used to get the repo IDs that are going to be disabled

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
trungams: removed the regex to simplify the implementation

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
trungams: I updated my code to follow this logic

trungams:
@0xba1a @PawelWMS I updated the logic so that we remove the existing repo files instead of using wildcard to disable repo IDs, as suggested by Pawel. Please let me know what you think of this approach.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -28,6 +28,8 @@ const (
 	toolchainRepoId        = "toolchain-repo"
 	cacheRepoID            = "upstream-cache-repo"
 	squashChrootRunErrors  = false
+	chrootRepoDir          = "/etc/yum.repos.d/"
+	chrootRepoFile         = "allrepos.repo"
```
PawelWMS, line 32: Could we move these back where they were, since they don't need to be globals anymore?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -69,9 +71,10 @@ const (
 
 // RpmRepoCloner represents an RPM repository cloner.
 type RpmRepoCloner struct {
-	chroot         *safechroot.Chroot
-	usePreviewRepo bool
-	cloneDir       string
+	chroot              *safechroot.Chroot
+	usePreviewRepo      bool
+	disableMarinerRepos bool
```
PawelWMS, line 76: We don't need `disableMarinerRepos` to be a member of this struct anymore, since we only need it inside `Initialize`, where we have it as an input variable and can further pass it into `initializeRepoDefinitions`.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -255,9 +258,13 @@ func (r *RpmRepoCloner) initializeRepoDefinitions(repoDefinitions []string) (err
...
-		err = appendRepoFile(originalRepoFilePath, dstFile)
-		if err != nil {
-			return
+		if !r.disableMarinerRepos {
+			err = appendRepoFile(originalRepoFilePath, dstFile)
+			if err != nil {
+				return
+			}
+		} else {
+			logger.Log.Infof("Disabling repositories listed in %s", originalRepoFilePath)
```
PawelWMS, line 267: >logger.Log.Infof("Disabling repositories listed in %s", originalRepoFilePath) [](http://example.com/codeflow?start=3&length=77)

The users are not familiar with the files we're removing here, so these messages may be confusing to them and just add unnecessary noise. I'd punt this down to a debug log.

For the user we can just print one generic "Disabling default upstream RPM repositories" message or something along these lines.

PawelWMS (changes requested):
Just a few clean-up comments, otherwise looks good.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
trungams: updated

PawelWMS (approved):
:shipit:

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -255,9 +259,13 @@ func (r *RpmRepoCloner) initializeRepoDefinitions(repoDefinitions []string) (err
 	// Add each previously existing repofile to the end of the new file, then delete the original.
 	// We want to try our custom mounted repos before reaching out to the upstream servers.
 	for _, originalRepoFilePath := range existingRepoFiles {
-		err = appendRepoFile(originalRepoFilePath, dstFile)
-		if err != nil {
-			return
+		if !disableMarinerRepos {
```
dmcilvaney, line 262: Could we add a bit more to the comment here explaining how the repos will get ordered and added/removed with this flag? Its a bit magic right now. (At least mention that the chroot ships with the mariner repos and we want to remove them).

dmcilvaney (approved):
Had one comment, but looks good otherwise.

toolkit/scripts/imggen.mk
```diff
@@ -93,6 +93,10 @@ ifeq ($(DISABLE_UPSTREAM_REPOS),y)
 imagepkgfetcher_extra_flags += --disable-upstream-repos
 endif
 
+ifeq ($(DISABLE_MARINER_REPOS),y)
+imagepkgfetcher_extra_flags += --disable-mariner-repos
```
0xba1a, line 97: Please change the flag name to `--disable-default-repos`

toolkit/scripts/pkggen.mk
```diff
@@ -118,6 +118,10 @@ ifeq ($(DISABLE_UPSTREAM_REPOS),y)
 graphpkgfetcher_extra_flags += --disable-upstream-repos
 endif
 
+ifeq ($(DISABLE_MARINER_REPOS),y)
+graphpkgfetcher_extra_flags += --disable-mariner-repos
```
0xba1a, line 122: Same here. Change it to `--disable-default-repos` or something relevant.

toolkit/Makefile
```diff
@@ -37,6 +37,7 @@ ALLOW_SRPM_DOWNLOAD_FAIL        ?= n
 RUN_CHECK                       ?= n
 USE_PREVIEW_REPO                ?= n
 DISABLE_UPSTREAM_REPOS          ?= n
+DISABLE_MARINER_REPOS           ?= n
```
0xba1a, line 40: Flag name can be more relevant and explanatory

0xba1a:
Please change the flag to some relevant name. Otherwise everything looks fine.

0xba1a (approved):
LGTM

toolkit/scripts/imggen.mk
PawelWMS: Good call, more generic.:)

PR #5609, opened 2023-06-01
Allow DHCP6 related packets through firewall

rlmenge:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

These changes are needed for Mariner clients to recieve DHCPv6 packets. This PR heavily borrows from https://github.com/microsoft/CBL-Mariner/pull/5071

As stated in the DHCPv6 Protocol (https://datatracker.ietf.org/doc/html/rfc8415) 
- udp on port 546 is needed for clients
- icmpv6 is needed. In an effort to reduce the changes as much as possible, only add types that are known to be needed. However, unclear if this is actually a worth while effort or if it will to more problems rather than allowing all icmpv6 packets

[NOTES]
- %post section does not seem to actually restart the iptables.service as the new rules are not seen when using `ip6tables -L`. Requires manually restarting the service or restarting the machine
- rerunning the iptables.service is not enough for the new ip local addresses to be seen. Need to also restart networkd

###### Change Log  

- Add rules to accept DHCPv6 packets

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/44612207

###### Test Methodology

- local build and install on Azure VM

Test method:
On fresh Azure Linux VM which supports dual stack (as seen here: https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/create-vm-dual-stack-ipv6-cli#code-try-0)
1. check that ip6tables -L has no icmpv6 rules
2. run tcpdump -n -i eth0 ip6 --> see no dhcp6 solicit or neighbor solicitation
3. run ip addr and see that eth0 has only link local addresses
4. load new rules and restart networkd and iptables
5. run tcpdump -n -i eth0 ip6 --> see dhcp6 solicit and neighbor solicitation messages
6. run ip addr and see that eth0 has additional local addresses


SPECS/iptables/ip6save
```diff
@@ -9,5 +9,20 @@
 #keep commented till upgrade issues are sorted
 #-A INPUT -j LOG --log-prefix "FIREWALL:INPUT "
 -A INPUT -p tcp -m tcp --dport 22 -j ACCEPT
+# Allow DHCPv6 client requests
+# Support SLACC
```
franckhlmartin, line 13: It is called SLAAC

franckhlmartin (commented):
This document is a good guide:
https://www.rfc-editor.org/rfc/rfc4890#section-4.4

So may be missing:
- Destination Unreachable (Type 1)
- Time Exceeded (Type 3) 
- Parameter Problem (Type 4)
- Inverse Neighbor Discovery Solicitation (Type 141)
- Inverse Neighbor Discovery Advertisement (Type 142)

I tend to allow echo request (128) and echo response (129) as it makes life easier to debug network routing issues, but if this is not enabled on IPv4 then keep it the same here.

aadhar-agarwal:
How are you testing the new iptable rules?

rlmenge:
> How are you testing the new iptable rules?

See updated description:)

SPECS/iptables/ip6save
rlmenge: done :)

franckhlmartin (approved):
Looks good to me

aadhar-agarwal (approved):

ddstreet (approved):
LGTM, thanks!

PR #5635, opened 2023-06-06
Fix libesmtp buffer overflow for CVE-2019-19977

ddstreet:
Fix buffer overflow for CVE-2019-19977

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

Fix buffer overflow for CVE-2019-19977

NO

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2019-19977

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=372547&view=results


dmcilvaney:
Looks like this package was bad when it was added, wondering if our git checks didn't catch it or if it predates them. Will need to address cgmanifest, license etc.

Otherwise looks good.

SPECS-EXTENDED/libesmtp/libesmtp.spec
```diff
@@ -5,14 +5,15 @@ Distribution:   Mariner
...
-Release:        20%{?dist}
+Release:        21%{?dist}
 License:        LGPLv2+
 #Source0:        https://github.com/libesmtp/libESMTP/archive/refs/tags/v1.0.6.tar.gz
 Source:         https://github.com/libesmtp/libESMTP/archive/refs/tags/%{name}-%{version}.tar.bz2
 URL:            https://github.com/libesmtp
 BuildRequires:  gcc
 BuildRequires:  openssl-devel pkgconfig autoconf automake libtool
 Patch0: libesmtp-1.0.6-openssl-1.1.patch
+Patch1: CVE-2019-19977-avoid-stach-overrwrite-6.patch
```
mfrw, line 16: NIT: I think we should match the spacing with the above line 14, for both line 15 & 16

ddstreet:
> Looks like this package was bad when it was added, wondering if our git checks didn't catch it or if it predates them. Will need to address cgmanifest, license etc.
> 
> Otherwise looks good.

fixed upstream link and verified license

ddstreet:
> Looks like this package was bad when it was added, wondering if our git checks didn't catch it or if it predates them. Will need to address cgmanifest,

we should ignore this failure; whoever added the tarball blob used a bz2-compressed version, but upstream provides only gz-compressed tarballs. So the spec file is correct in listing the Source: line as ending in bz2 (that's what is in our blob store).

Fixing that check would require uploading a new blob with gz compression, which is totally unnecessary (and has the potential to introduce unexpected changes). If/when the actual version is updated (mariner 3 maybe), then the new version tarball should be uploaded as a tar.gz and the spec source line updated to correctly point to a tar.gz.

> license etc.
added 'license verified' to changelog

dmcilvaney:
> > Looks like this package was bad when it was added, wondering if our git checks didn't catch it or if it predates them. Will need to address cgmanifest,
> 
> we should ignore this failure; whoever added the tarball blob used a bz2-compressed version, but upstream provides only gz-compressed tarballs. So the spec file is correct in listing the Source: line as ending in bz2 (that's what is in our blob store).
> 
> Fixing that check would require uploading a new blob with gz compression, which is totally unnecessary (and has the potential to introduce unexpected changes). If/when the actual version is updated (mariner 3 maybe), then the new version tarball should be uploaded as a tar.gz and the spec source line updated to correctly point to a tar.gz.
> 
> > license etc.
> > added 'license verified' to changelog

I think we will need an admin to bypass the check then.

dmcilvaney (approved):

dmcilvaney:
> > > > Looks like this package was bad when it was added, wondering if our git checks didn't catch it or if it predates them. Will need to address cgmanifest,
> > > 
> > > 
> > > we should ignore this failure; whoever added the tarball blob used a bz2-compressed version, but upstream provides only gz-compressed tarballs. So the spec file is correct in listing the Source: line as ending in bz2 (that's what is in our blob store).
> > > Fixing that check would require uploading a new blob with gz compression, which is totally unnecessary (and has the potential to introduce unexpected changes). If/when the actual version is updated (mariner 3 maybe), then the new version tarball should be uploaded as a tar.gz and the spec source line updated to correctly point to a tar.gz.
> > > > license etc.
> > > > added 'license verified' to changelog
> > 
> > 
> > I think we will need an admin to bypass the check then.
> 
> we can just push to main, no admin needed for that

Fairly sure the manifests check is mandatory, so will need an admin to force the merge in.
@jslobodzian

oliviacrain:
@ddstreet You can also add this spec to the `ignore_known_issues` list in `.github/workflows/validate-cg-manifest.sh` with a comment explaining the issue.

ddstreet:
> > > > > Looks like this package was bad when it was added, wondering if our git checks didn't catch it or if it predates them. Will need to address cgmanifest,
> > > > 
> > > > 
> > > > we should ignore this failure; whoever added the tarball blob used a bz2-compressed version, but upstream provides only gz-compressed tarballs. So the spec file is correct in listing the Source: line as ending in bz2 (that's what is in our blob store).
> > > > Fixing that check would require uploading a new blob with gz compression, which is totally unnecessary (and has the potential to introduce unexpected changes). If/when the actual version is updated (mariner 3 maybe), then the new version tarball should be uploaded as a tar.gz and the spec source line updated to correctly point to a tar.gz.
> > > > > license etc.
> > > > > added 'license verified' to changelog
> > > 
> > > 
> > > I think we will need an admin to bypass the check then.
> > 
> > 
> > we can just push to main, no admin needed for that
> 
> Fairly sure the manifests check is mandatory, so will need an admin to force the merge in. @jslobodzian

what i meant was, it might be mandatory for using the web ui, but we can all just push directly to main, without any extra requirements.

ddstreet:
> @ddstreet You can also add this spec to the `ignore_known_issues` list in `.github/workflows/validate-cg-manifest.sh` with a comment explaining the issue.

thanks, updated that, rebased, and pushed to main

PR #5640, opened 2023-06-07
toolkit: add feature to profile & trace to have better diagnostics

mfrw:

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Add profiling and tracing support to the toolkit. The go toolchain has inbuilt tools to diagnose logic and performance problems in Go programs. This PR integrates that support for our toolkit.

This PR introduces 3 additional flags, by default all the three flags are disabled. 
- `ENABLE_TRACE`: Setting this flag to `y`, instruments the various tools in the toolkit to record `trace` data. Tracing is a way to instrument code to analyze latency throughout the lifecycle of a call or user request. Traces provide an overview of how much latency each component contributes to the overall latency in a system. Traces can span multiple Go processes.
- `ENABLE_MEM_PROFILE`: Setting this flag to `y` instruments the toolkit to record the memory profile.
- `ENABLE_CPU_PROFILE`: Setting this flag to `y` instruments the toolkit to record the cpu profile.

Profiling tools analyze the complexity and costs of a Go program such as its memory usage and frequently called functions to identify the expensive sections of a Go program.

It is advisable to only turn one of the flags for a run to get data that does not have noise due to other profiling/tracing flags.

The trace/profile data will be created in the `build/profile` directory. 
To use the profile data: `go tool pprof -http=:8080 file.profile`
To use the trace data: `go tool trace -http=:8080 file.trace`

###### Change Log  

- NA

**NO**

###### Associated issues  

- NA

###### Links to CVEs  
- NA

###### Test Methodology

- Local Build


toolkit/tools/grapher/grapher.go
```diff
@@ -36,6 +45,46 @@ func main() {
 	app.Version(exe.ToolkitVersion)
 	kingpin.MustParse(app.Parse(os.Args[1:]))
 
+	// CPU Profiling
```
dmcilvaney, line 48: I was thinking about this again, could we add a new `profiler` type, and have a "constructor" and "destructor" for it in the exe package?
Something like
```go
type Profiler stuct {
	cpf, mpf, tf *os.File	 
}

func CreateProfiler(cpfFile, mpfFile, tfFile string) p *Profiler {
	p = &Profiler{}
	if cpfFile != "" {
		p.cpf, err = os.Create(cpfFile)
		// TODO: error check and start
	}
	if mpfFile != "" {
		p.mpf, err := os.Create(mpfFile)
		// TODO: error check and start
	}
	// etc...
	return p
}

func (p* Profiler) StopProfiler {
	if p.cpf != nil {
                pprof.StopCPUProfile()
		p.cpf.Close()
	}
        // etc...
}
```
Then we could call it like:
```go
profiler, err := exe.CreateProfiler(*cpuProfFile, *memProfFile, *traceFile)
// ...
defer profiler.StopProfiler()
```

dmcilvaney: (I didn't actually compile any of this... so ymmv but the idea is there)

toolkit/scripts/pkggen.mk
```diff
@@ -86,6 +86,12 @@ $(specs_file): $(chroot_worker) $(BUILD_SPECS_DIR) $(build_specs) $(build_spec_d
 		--worker-tar $(chroot_worker) \
 		$(if $(filter y,$(RUN_CHECK)),--run-check) \
 		$(logging_command) \
+		--cpu-prof-file=$(PROFILE_DIR)/specreader.cpu.pprof \
```
dmcilvaney, line 89: This could probably  be done the same way we define `$(logging_command)` since its all the 'same' for each tool.

toolkit/tools/grapher/grapher.go
mfrw: Great idea, let me try to implement it this way.

toolkit/scripts/pkggen.mk
mfrw: I tried to do it as generically as possible, but my make skills are not on point, and although it worked, but the resultant filenames were not self evident of the stages.
e.g for the spec reader, it created a file called `specs.trace`, i would prefer `specreader.trace` though.

Should we could keep it as is, what do you think ?
@dmcilvaney

dmcilvaney: While you would still end up with the ugly naming, I think this would work if dropped into `profile.mk`:
```make
PROFILE_DIR    ?= $(BUILD_DIR)/profile
traceing_file   = $(PROFILE_DIR)/$(notdir $@)
tracing_command = --cpu-prof-file=$(traceing_file).cpu.pprof \
                  --mem-prof-file=$(traceing_file).mem.pprof \
                  --trace-file=$(traceing_file).trace \
                  $(if $(filter y,$(ENABLE_CPU_PROFILE)),--enable-cpu-prof) \
                  $(if $(filter y,$(ENABLE_MEM_PROFILE)),--enable-mem-prof) \
                  $(if $(filter y,$(ENABLE_TRACE)),--enable-trace)
```
You could then call it via
```make
	$(go-imageconfigvalidator) \
		--input=$(CONFIG_FILE) \
		--dir=$(CONFIG_BASE_DIR) \
		$(tracing_command ) \
		--timestamp-file=$(TIMESTAMP_DIR)/imageconfigvalidator.jsonl && \
	touch $@
```

mfrw: This is exactly what I did, the names ended up not being pleasing to the idea, or give the intent of what it was.
e.g for the specreader, the file was `specs.trace`

Inn anycase, let me read more on it

dmcilvaney: Its the naming convention we ended up with for our log files, so it won't be unique (but agreed, the names sort of suck).

toolkit/tools/grapher/grapher.go
```diff
@@ -26,6 +27,12 @@ var (
 	strictGoals      = app.Flag("strict-goals", "Don't allow missing goal packages").Bool()
 	strictUnresolved = app.Flag("strict-unresolved", "Don't allow missing unresolved packages").Bool()
 	timestampFile    = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
+	enableCpuProf    = app.Flag("enable-cpu-prof", "Enable CPU pprof data collection.").Bool()
```
dmcilvaney, line 30: Another brainwave... why have all these flags listed here, we are going to turn around and immediately call `profile.StartProfiling()` with all of them in main and they are always the same for every tool. I wonder if we can just hide this all inside something like `exe.InputFlag(app, string)`

```go
// in exe.go
func ProfilingFlag(k *kingpin.Application)  *Profiler
	enableCpuProf = k.Flag("enable-cpu-prof", "Enable CPU pprof data collection.").Bool()
	enableMemProf    = k.Flag("enable-mem-prof", "Enable Memory pprof data collection.").Bool()
	enableTrace      = k.Flag("enable-trace", "Enable trace data collection.").Bool()
	cpuProfFile      = k.Flag("cpu-prof-file", "File that stores CPU pprof data.").String()
	memProfFile      = k.Flag("mem-prof-file", "File that stores Memory pprof data.").String()
	traceFile        = k.Flag("trace-file", "File that stores trace data.").String()

	p, err := profile.StartProfiling(*cpuProfFile, *memProfFile, *traceFile, *enableCpuProf, *enableMemProf, *enableTrace)
	if err != nil {
		// Unlcear what to do here, panic? Maybe kingping has a nice way to complain to the user?
	}
	return p
```

dmcilvaney: Maybe we should treat this like we do the timestaping code, never break the program due to a timing failure. Just complain to the user and keep going with it turned off.

toolkit/tools/pkg/profile/profile.go
```diff
@@ -0,0 +1,76 @@
...
+	"runtime/trace"
+)
+
+type Profiler struct {
+	cpuProfFile *os.File
+	memProfFile *os.File
+	traceFile   *os.File
+}
+
+func StartProfiling(cpuProfFile, memProfFile, traceFile string, enableCpuProf, enableMemProf, enableTrace bool) (*Profiler, error) {
```
dmcilvaney, line 17: Doesn't need to be now, but adding a readme on how to consume this data would be nice as well.

mfrw: Sure, will make another PR

toolkit/tools/grapher/grapher.go
mfrw: This is a cleaner approach. Thanks

toolkit/tools/pkg/profile/profile.go
dmcilvaney, line 17: We could clean up the signature here even more, since we are already creating this container  when we call `SetupProfileFlags()`
```suggestion
func StartProfiling(profileFlags exe.ProfileFlags) (*Profiler, error) {
```

Sumynwa (approved):

toolkit/tools/pkg/profile/profile.go
```diff
@@ -0,0 +1,81 @@
...
+	p := &Profiler{
+		cpuProfFile: nil,
+		memProfFile: nil,
+		traceFile:   nil,
+	}
+
+	if *pf.EnableCpuProf {
+		cpf, err := os.Create(*pf.CpuProfFile)
+		if err != nil {
+			return nil, fmt.Errorf("Unable to create cpu-pprof file: %s", err)
```
dmcilvaney, line 29: Linter will probably be grumpy about the capitalization of "Unable". It likes error strings to be all lowercase.

toolkit/tools/grapher/grapher.go
```diff
@@ -35,10 +37,15 @@ func main() {
...
-	var err error
 	logger.InitBestEffort(*logFile, *logLevel)
 
+	prof, err := profile.StartProfiling(profFlags)
+	if err != nil {
+		logger.Log.Warnf("Could not start profiling: %s", err)
+		return
+	}
+	defer prof.StopProfiler()
+
```
dmcilvaney, line 48: Do we want to make success here fully optional? This isn't a critical flow for the tools
```suggestion
	prof, err := profile.StartProfiling(profFlags)
	if err != nil {
		logger.Log.Warnf("Could not start profiling: %s", err)
	} else {
		defer prof.StopProfiler()
	}

```

dmcilvaney (approved):
The big block of flags in the Makefile still seems messy, but not a deal breaker. 
Added one suggestion to make the error 'optional' but I think this looks ok.

PR #5660, opened 2023-06-09
Fix against race condition in unattended install

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

The ISO installer will start on every available console (currently `tty1` and `ttyS0`). For normal install flows this is fine since they will both wait for the user to select the desired installer flow. The unattended installer however will proceed automatically. Since we have two instances they will both attempt to install at the same time generally causing one to fail to partition. Anecdotally this is usually the serial console failing making it look like the install has failed, however the `tty1` installer is actually running successfully in the background. If left alone the image will generally install correctly and restart after a minute or two.

Lets prioritize` ttyS0` as the 'correct' installer, and have all other instances give up and just follow the log output from the running copy.

Also add a `installer-initrd` PHONY target that builds the ISO image.

###### Change Log  

- Prioritize unattended installer on `ttyS0` and block all others.
- Add `installer-initrd` target

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/44974720

###### Test Methodology

- Local testing against VMs

### DIFF
--- a/toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller
+++ b/toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller
@@ -109,11 +109,12 @@ if [[ -f "$UNATTENDED_CONFIG_FILE" ]]; then
     # If the file descriptor is already locked bail out immediately and just follow the other installer's log. The tail
     # command will follow the log file and should never exit. We must ensure that this instance of the installer does
     # not run any further so we always exit 0 just in case.
-    flock -x -n 200 || \
-        { echo "Unattended install already running, following other installer's log instead" ; \
-        sleep 5 ; \
-        tail -f -n 10000 /installer/log.txt ; \
-        exit 0 ; }
+    if ! flock -x -n 200; then
+        echo "Unattended install already running, following other installer's log instead" 
+        sleep 5
+        tail -f -n 10000 /installer/log.txt 
+        exit 0
+    fi
 fi
 
 if [[ ! -f "$UNATTENDED_CONFIG_FILE" ]]; then
--- a/toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller
+++ b/toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller
@@ -8,6 +8,7 @@ export TERM=mariner-installer
 ISO_ROOT=/mnt/cdrom
 CONFIG_ROOT=$ISO_ROOT/config
 UNATTENDED_CONFIG_FILE=$CONFIG_ROOT/unattended_config.json
+UNATTENDED_LOCK_FILE=/var/lock/unattended.lock
 
 # PXE boot setup
 CMDLINE=/proc/cmdline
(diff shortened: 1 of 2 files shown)


hbeberman (approved):
:shipit:

toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller
```diff
@@ -33,6 +34,39 @@ while getopts ":c:u:" OPTIONS; do
   esac
 done
 
+# If we are running unattended, we need to make sure that we only run once.
+# We can use the unattended config file via flock to synchronize the unattended
+# install flows.
+if [[ ! -f "$UNATTENDED_CONFIG_FILE" ]]; then
```
trungams, line 40: shouldn't the `!` be removed?

dmcilvaney: Huh... will need to dig into this to see why this is working  at all then...

dmcilvaney: Ah, I see the issue. `$ISO_ROOT` isn't mounted yet so this always returns true. This logic needs to move down after we complete the mount process.

(Unrelated, but this race condition illuminates why the mount functionality may have started failing earlier)

trungams: Could you explain a bit what would have caused the mount failure? I think I've seen it once or twice before when trying ISO install on both devices. Not sure if this change also help fix it.

dmcilvaney: The old retry code would loop until the `mount` command returned code 0. But... the mount command will return non-zero when the mount point is already mounted. This would mean that if the mount was already mounted it would get stuck in the retry loop until it timed out. This likely started being an issue when we added the parallel installers since only one would be able to successfully call the mount command, the other would end up stuck in the retry loop. I added the grep to escape early if it was already mounted but I couldn't for the life of me figure out what was causing the mount to already be present. In hindsight... it was the other installer running on ttyS0.

trungams (approved):

trungams:
:shipit:

toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller
```diff
@@ -83,6 +84,38 @@ else
     retry "{ mount -L $LABEL -o ro $ISO_ROOT || grep -qs $ISO_ROOT /proc/mounts ; }"
 fi
 
+# If we are running unattended (i.e., have $UNATTENDED_CONFIG_FILE), we need to make sure that we only run once.
+# We can use flock to synchronize the unattended install flows.
+if [[ -f "$UNATTENDED_CONFIG_FILE" ]]; then
```
PawelWMS, line 89: Could we combine this `if` with the one behind it with an `else`? That way it should be a bit clearer, that there's nothing special happening in this `if` (like removal of the unattended config file), that would influence the result of the second check.

toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller
```diff
@@ -83,6 +84,38 @@ else
...
+    echo "Synchronizing unattended install flows... current sessions are:"
+    who
+    thisTerminal=$(who -m | xargs | cut -d' ' -f2)
+    echo "This instance is on '$thisTerminal'"
+
+    # Generally we want to prioritize ttyS0 since that will be used for debugging. The output from the log file will
+    # contain the timestamp and color codes so it is still useful for debugging and showing liveness but it will look
+    # messy.
+    if [[ "$thisTerminal" != "ttyS0" ]]; then
```
PawelWMS, line 98: If we're sure that `ttyS0` is doing the same thing, why not just exit here immediately?

trungams: I think it's expected, since `runliveinstaller` is the entrypoint script for each user session: https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/resources/imageconfigs/iso_initrd.json#L50

dmcilvaney: I never want to leave us in the state where the installer doesn't run at all. I don't know if we can guarantee the exact consoles we have won't ever change. This is much more defensive than explicitly bailing out on a specific console.

oliviacrain (approved):

toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller
```diff
@@ -83,6 +84,38 @@ else
...
+    exec 200>> "$UNATTENDED_LOCK_FILE"
+
+    # If the file descriptor is already locked bail out immediately and just follow the other installer's log. The tail
+    # command will follow the log file and should never exit. We must ensure that this instance of the installer does
+    # not run any further so we always exit 0 just in case.
+    flock -x -n 200 || \
+        { echo "Unattended install already running, following other installer's log instead" ; \
+        sleep 5 ; \
+        tail -f -n 10000 /installer/log.txt ; \
+        exit 0 ; }
```
PawelWMS, line 116: Refactoring suggestion to achieve two goals:
- Drop the need for repeated `; \` at the end of every line.
- Be explicit about ignoring non-zero exit codes from `flock`. Thanks to that, this code should be more resilient if we ever add `set -e` to this script in the future. We tend to already do that in our scripts and I wouldn't be very surprised, if we ever wanted to introduce it here to fail fast.

```suggestion
    if ! flock -x -n 200; then
        echo "Unattended install already running, following other installer's log instead" 
        sleep 5
        tail -f -n 10000 /installer/log.txt 
        exit 0
    fi
```

PawelWMS (approved):
Small refactoring suggestions but looks good.

toolkit/resources/imageconfigs/additionalfiles/iso_initrd/root/runliveinstaller
dmcilvaney: I considered it, but I feel these are really unrelated bits of code. I found it much more readable to keep them split into their own blocks.

PR #5677, opened 2023-06-13
Toolkit: Add MAX_CPU flag to limit number of CPUS used for package building

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add MAX_CPU flag to limit number of CPUS used for package building. When set, this value is used to set the `%_smp_ncpus_max` in each package build chroot. This value sets an upper limit on CPUs to use for all packages that build with `make %{?_smp_mflags}`.
Our immediate need for this is to prevent OOM errors on ARM64 machines, which are typically 32 core, and regularly run out of memory during full package builds.

###### Change Log  

- Add MAX_CPU option to toolkit
- Update documentation

**NO**

###### Test Methodology

- Local build test - setting MAX_CPU to 3:
- -  time sudo make build-packages SRPM_PACK_LIST="tree" REBUILD_TOOLS=y REFRESH_WORKER_CHROOT=n USE_CCACHE=y RUN_CHECK=n MAX_CPU=3
- - Observed in tree-1.8.0-2.cm2.src.rpm.log on 8 core machine: cd tree-1.8.0; make -j3

- Local build test - not specifying MAX_CPU at all
- -  time sudo make build-packages SRPM_PACK_LIST="tree" REBUILD_TOOLS=y REFRESH_WORKER_CHROOT=n USE_CCACHE=y RUN_CHECK=n
- - Observed in tree-1.8.0-2.cm2.src.rpm.log on 8 core machine: cd tree-1.8.0; make -j8

### DIFF
--- a/toolkit/tools/pkgworker/pkgworker.go
+++ b/toolkit/tools/pkgworker/pkgworker.go
@@ -54,6 +54,7 @@ var (
 	packagesToInstall    = app.Flag("install-package", "Filepaths to RPM packages that should be installed before building.").Strings()
 	outArch              = app.Flag("out-arch", "Architecture of resulting package").String()
 	useCcache            = app.Flag("use-ccache", "Automatically install and use ccache during package builds").Bool()
+	maxCPU               = app.Flag("max-cpu", "Max number of CPUs used for package building").String()
 
 	logFile  = exe.LogFileFlag(app)
 	logLevel = exe.LogLevelFlag(app)
@@ -93,6 +94,7 @@ func main() {
 	if *useCcache {
 		defines[rpm.MarinerCCacheDefine] = "true"
 	}
+	defines[rpm.MaxCPUDefine] = *maxCPU
 
 	builtRPMs, err := buildSRPMInChroot(chrootDir, rpmsDirAbsPath, toolchainDirAbsPath, *workerTar, *srpmFile, *repoFile, *rpmmacrosFile, *outArch, defines, *noCleanup, *runCheck, *packagesToInstall, *useCcache)
 	logger.PanicOnError(err, "Failed to build SRPM '%s'. For details see log file: %s .", *srpmFile, *logFile)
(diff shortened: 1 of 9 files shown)


toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -54,6 +54,7 @@ var (
 	packagesToInstall    = app.Flag("install-package", "Filepaths to RPM packages that should be installed before building.").Strings()
 	outArch              = app.Flag("out-arch", "Architecture of resulting package").String()
 	useCcache            = app.Flag("use-ccache", "Automatically install and use ccache during package builds").Bool()
+	maxCPU               = app.Flag("max-cpu", "Max number of CPUs used for package building").String()
```
oliviacrain, line 57: Can we either mark this as required or provide a default value of `0`, if not provided?

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -93,6 +94,7 @@ func main() {
 	if *useCcache {
 		defines[rpm.MarinerCCacheDefine] = "true"
 	}
+	defines[rpm.MaxCPUDefine] = *maxCPU
```
oliviacrain, line 97: Would love to see input validation ensuring we don't have negative inputs.

toolkit/tools/pkgworker/pkgworker.go
anphel31: That's a good idea; I'll fix up the default case. will check for when nothing is provided, and not set the macro at all in that case.
Should not need to mark `max-cpu` as required, since it's just an override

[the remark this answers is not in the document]
anphel31: Is there any benefit? I looked into this and made the changes to Int locally where I could, but it still needs to be converted back to a string to set the macro in the chroot. Seems to be simpler to always handle it as a string

oliviacrain (approved):

PR #5693, opened 2023-06-15
rpm: remove unnecessary brp-strip scripts

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Remove redundant brp-strip operations from our default `__os_install_post` definition.
Get rid of custom scripts bundled in our rpm-build package: `brp-strip-debug-symbols` & `brp-strip-unneeded`
Call the upstream rpm versions of each brp-strip* script, while still keeping some Mariner customizations related to ccache and embedded ELF metadata.

###### Change Log  

- rpm: Remove unnecessary script: `brp-strip-debug-symbols`. This does the exact same thing as `brp-strip`, but in a less efficient way - it searches for all ELF binaries and sequentially removes debug symbols. `brp-strip` removes the debug symbols from multiple files in parallel.
- rpm: Remove unnecessary script: `brp-strip-unneeded`. This calls "strip --strip-unneeded -g" on every ELF binary. The upstream rpm macro does not call --strip-unneeded, and the `brp-strip-comment-note` script appears to strip the same content as this script.
- mariner-rpm-macros: remove references to `brp-strip-debug-symbols`, `brp-strip-unneeded` in `__os_install_post`
- mariner-rpm-macros: use standard brp-macro naming definitions from upstream rpm
- update toolchain manifests

**YES**

###### Test Methodology

- Full X64 build: 2.0.20230615-anphel-378724
- Full ARM64 build: 2.0.20230616-anphel-379014
- Buddy build to just build rpm and mariner-rpm-macros packages: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=378541&view=results
- I have spot-checked binaries built with this change from the following packages: kernel, moby-cli, tree. I compared binaries built with and without these changes, comparing the file size and headers (using `objdump -x binary`). I did not detect any differences in these packages. The overall size of the debuginfo packages with and without the changes is also consistent.


SPECS/mariner-rpm-macros/macros
```diff
@@ -69,19 +69,26 @@
...
+    %{?__brp_strip_static_archive} \
     %{?py_auto_byte_compile:%{?__brp_python_bytecompile}} \
     find %{buildroot} -name "*.pc" | xargs -I{} sed -i -e 's@-Wl,-dT,%{_topdir}/BUILD/module_info.ld@ @' {} \
     %{nil}
 
 %__os_install_post    \
     %{__os_install_post_leave_signatures} \
-    %{_rpmconfigdir}/brp-strip %{__strip} \
-    %{_rpmconfigdir}/brp-strip-debug-symbols %{__strip} \
-    %{_rpmconfigdir}/brp-strip-unneeded %{__strip} \
```
anphel31, line 84: note: removed brp-strip-static-archive due to errors like the below seen in toolchain build:
```
error: File not found: /usr/src/mariner/BUILDROOT/cracklib-2.9.7-5.cm2.x86_64/usr/lib/libcrack.la

" error: File not found: /usr/src/mariner/BUILDROOT/gcc-11.2.0-4.cm2.x86_64/usr/lib64/libstdc++.la 
" error: File not found: /usr/src/mariner/BUILDROOT/gcc-11.2.0-4.cm2.x86_64/usr/lib64/libsupc++.la
```

SPECS/mariner-rpm-macros/macros
```diff
@@ -69,19 +69,26 @@
...
+    %{?__brp_strip_comment_note} \
+    %{?__brp_strip_static_archive} \
     %{?py_auto_byte_compile:%{?__brp_python_bytecompile}} \
     find %{buildroot} -name "*.pc" | xargs -I{} sed -i -e 's@-Wl,-dT,%{_topdir}/BUILD/module_info.ld@ @' {} \
     %{nil}
 
 %__os_install_post    \
     %{__os_install_post_leave_signatures} \
-    %{_rpmconfigdir}/brp-strip %{__strip} \
-    %{_rpmconfigdir}/brp-strip-debug-symbols %{__strip} \
```
anphel31, line 83: Note: removed brp-elfperms due to error seen building rpms with modified toolchain:

```
chroot: failed to run command ‘rpm’: Permission denied 
```

anphel31: upstream __os_install_post for reference: https://github.com/rpm-software-management/rpm/blob/8ec9cfc81c3f8326a1c7ee5b795ef517cd48e6b4/platform.in#L64

SPECS/mariner-rpm-macros/macros
anphel31: upstream __os_install_post for reference: https://github.com/rpm-software-management/rpm/blob/8ec9cfc81c3f8326a1c7ee5b795ef517cd48e6b4/platform.in#L64

reubeno: I think you meant `brp-remove-la-files`? Reading up on the original [PR](https://github.com/rpm-software-management/rpm/pull/1674) it looks like the upstream thinking was that some components would need to opt-out of this behavior -- perhaps such as the ones you hit errors in -- possibly by redefining the macro wrapping this script to no-op it? I think I'd agree this doesn't need to be done now; do you think it would be worth a follow-up later?

SPECS/mariner-rpm-macros/macros
reubeno: Curious... I'm a bit fuzzy as to what the impact is of us not having this script run, but seems like something that could be deferred?

anphel31: It's something we never had before, so should be fine to leave out. From the upstream commit, it looks like an optimization to reduce the amount of binaries scanned for dependency generation by removing the execute bit. We can probably evaluate later on, but for now it definitely causes some problems. Upstream commit: https://github.com/rpm-software-management/rpm/commit/80818e4f902ba3cf85e4cfcd8a7a4c71c601f3cf

SPECS/mariner-rpm-macros/macros
anphel31: Yes, I did mean to say `__brp_remove_la_files`. Seems worthy of follow up, but a quick grep of our core specs shows 200+ packages that perform something along the lines of `find %{buildroot} -type f -name "*.la" -delete -print` already.

reubeno (approved):
Great to see us making these changes! I'm assuming we'll want to spot-check a few packages to make sure that the resulting RPMs's binaries still match what we'd expect to see (i.e., appropriately stripped of debug info, etc.)?

SPECS/rpm/rpm.spec
```diff
@@ -285,6 +281,9 @@ popd
 %{python3_sitelib}/*
 
 %changelog
+* Thu Jun 15 2023 Andrew Phelps <anphel@microsoft.com> - 4.18.0-3
```
oliviacrain, line 284: Can't comment on the relevant line, but can we add a version constraint `>= 2.0-22` on the `mariner-rpm-macros` requirement in the `rpm-build` subpackage?

anphel31: sure, will add this

anphel31:
> Great to see us making these changes! I'm assuming we'll want to spot-check a few packages to make sure that the resulting RPMs's binaries still match what we'd expect to see (i.e., appropriately stripped of debug info, etc.)?

Yes, I've spot checked some of the built packages. Will update the test section of the description with details.

oliviacrain (approved):

PawelWMS (approved):

PR #5708, opened 2023-06-19
Adding missing binaries for cdi v1.55+

Adub17030MS:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR is adding the tools/ binaries to rpm spec (these are needed for Nexus GA). The following binaries are required to be included by the Nexus 

###### Change Log  

- Adding tools/ binaries to spec

**NO**

###### Test Methodology

- Buddybuild: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=381930&view=results
- Validation will occur on the Nexus side, so we are good to merge on this end


SPECS/containerized-data-importer/containerized-data-importer.spec
```diff
@@ -155,7 +168,7 @@ install -m 0644 _out/manifests/release/cdi-cr.yaml %{buildroot}%{_datadir}/cdi/m
 %files api
 %license LICENSE
 %doc README.md
-%{_bindir}/virt-cdi-apiserver
+%{_bindir}/cdi-apiserver
```
PawelWMS, line 171: This way we're effectively removing the binaries we have already made public and potentially introducing regressions fo whoever has already been using this package. Are we only re-naming the files but they are actually the same executables or are the new ones different from the old ones?

PawelWMS (changes requested):
Please add more details in the summary. Right now it's unclear why we're making this change and in the current state the modifications are a regression, so we should have a good justification.

If we're only re-naming the executables, then please also explain why we're doing that and make sure the old ones are still available to prevent breaking anyone, who might have been using the ones from the previous versions of the package.

SPECS/containerized-data-importer/containerized-data-importer.spec
oliviacrain: Agreed, would like to see symlinks from the old names to the new.

romoh: Same binaries, just renames. 
We had a bad name to begin with. Those were created mainly for [cdi containers](https://github.com/kubevirt/containerized-data-importer/) so less likely a chance of breaking anyone.

PawelWMS: Yes, but only if the new files are still the same executables, just with a different name. If these are different tools, then we'd need to keep both.

SPECS/containerized-data-importer/containerized-data-importer.spec
```diff
@@ -18,13 +18,14 @@
...
 Version:        1.55.0
-Release:        10%{?dist}
+Release:        11%{?dist}
 License:        ASL 2.0
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 Group:          System/Packages
 URL:            https://github.com/kubevirt/containerized-data-importer
 Source0:        https://github.com/kubevirt/containerized-data-importer/archive/refs/tags/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
+Patch0:         allow-dynamic-build.patch
```
PawelWMS, line 28: In order to apply the patch, you'll also need `%setup` to take it into account. For this spec the easiest way would probably be to add a `%patch0 -p1` line somewhere under `%setup`, but I also encourage you to try switching to `%autosetup -p1`

jslobodzian:
PR has been opened for over 2 years against Mariner 2.0.  Mariner 2.0 support has ended.  Closing as won't fix.

PR #5716, opened 2023-06-20
systemd: Enable audit integration.

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Enabled audit integration in systemd package. This is needed so that systemd will report any SELinux violations for service control operations (start/stop/etc.) to the audit logs. This will ensure that audits of a system's security can be done accurately. In addition, it greatly helps with debugging SELinux issues.

###### Change Log  

- Enabled audit integration in systemd package.

NO

###### Test Methodology

1. Built the package locally.
2. Installed the updated package in a test VM.
3. Ran a `systemctl restart` command that I knew would hit an SELinux violation.
4. Ensured that the expected SELinux violation showed up in the audit logs.

### DIFF
--- a/SPECS/systemd/systemd.spec
+++ b/SPECS/systemd/systemd.spec
@@ -48,7 +48,6 @@ BuildRequires:  tpm2-tss-devel
 BuildRequires:  util-linux-devel
 BuildRequires:  xz-devel
 Requires:       %{name}-rpm-macros = %{version}-%{release}
-Requires:       audit
 Requires:       glib
 Requires:       kmod
 Requires:       libcap
--- a/SPECS/systemd/systemd.spec
+++ b/SPECS/systemd/systemd.spec
@@ -1,7 +1,7 @@
 Summary:        Systemd-250
 Name:           systemd
 Version:        250.3
-Release:        15%{?dist}
+Release:        16%{?dist}
 License:        LGPLv2+ AND GPLv2+ AND MIT
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -25,6 +25,7 @@ Patch3:         CVE-2022-3821.patch
 Patch4:         CVE-2022-45873.patch
 Patch5:         backport-helper-util-macros.patch
 Patch6:         CVE-2022-4415.patch
+BuildRequires:  audit-devel
 BuildRequires:  cryptsetup-devel
 BuildRequires:  docbook-dtd-xml
 BuildRequires:  docbook-style-xsl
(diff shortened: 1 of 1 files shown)


romoh:
LGTM

SPECS/systemd/systemd.spec
```diff
@@ -47,6 +48,7 @@ BuildRequires:  tpm2-tss-devel
 BuildRequires:  util-linux-devel
 BuildRequires:  xz-devel
 Requires:       %{name}-rpm-macros = %{version}-%{release}
+Requires:       audit
```
hbeberman, line 51: We intentionally don't have audit installed on a few configurations. Can we remove this Requires and still have systemd operate both with and without audit installed?

hbeberman (changes requested):

SPECS/systemd/systemd.spec
PawelWMS: That's a good point. If we build with `audit-devel` and `-Daudit=true` we will most probably end up with binaries having a dependency on dynamic libraries from `audit`. Depending on the binaries now loading `audit` libs during run-time we may or may not be able to break the dependency. If the `systemd` main executable ends up with the new dependency, then this might be unbreakable. If only some non-critical tools/executables have the new dependency, then we would have the option of moving them to a separate subpackage, which would keep the default `systemd` clean. I wouldn't be surprised, however, if it's going to turn out, that the dependency is unbreakable this way.

An alternative would be to now build `systemd` twice with and without the `-Daudit=true` switch and put the resulting binaries in separate subpackages.

cwize1: Looks like systemd happily ignores failures to connect the audit socket.

There is a dependency on `audit-libs`. But that dependency has already been declared.

PawelWMS: And is there still no implicit (added by RPM) dependency between `systemd` and `audit` after this change? This should be visible by comparing the output of `rpm -q --requires <systemd_rpm>` before and after the change.

cwize1: Running that command showed no dependency on the `audit` package. (Also, during testing I was able to remove `audit` without `dnf` wanting to remove `systemd`.)

There is a dependency on `libaudit.so`. But that is provided by `audit-libs` not `audit.`

PawelWMS (approved):
Looks good to me. I understand Henry's tests will give us the final "yay" or "nay", but from our offline discussions it seems like this is good.

hbeberman (approved):
I've confirmed on one of the non-audit enabled images that this change isn't unexpectedly pulling audit in and that systemd is continuing to operate normally. LGTM.

PR #5719, opened 2023-06-20
Package cmake modules in grpc-devel

reubeno:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
I've encounted some source code that expects to find gRPCConfig.cmake under /usr/lib/cmake. The grpc upstream package typically provides this, but the Mariner spec was deleting these cmake files for some reason.

###### Change Log  
Removes the logic to delete the *.cmake files, and packages them up in grpc-devel.

No.

###### Test Methodology

In progress.

### DIFF
--- a/SPECS/grpc/grpc.spec
+++ b/SPECS/grpc/grpc.spec
@@ -1,7 +1,7 @@
 Summary:        Open source remote procedure call (RPC) framework
 Name:           grpc
 Version:        1.42.0
-Release:        5%{?dist}
+Release:        6%{?dist}
 License:        ASL 2.0
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -98,7 +98,6 @@ popd
 %install
 pushd cmake/build
 %cmake_install
-find %{buildroot} -name '*.cmake' -delete
 popd
 
 #python
@@ -135,6 +134,7 @@ export GRPC_PYTHON_BUILD_SYSTEM_ABSL=True
 %{_libdir}/libgrpcpp_channelz.so
 %{_libdir}/libupb.so
 %{_libdir}/pkgconfig/*.pc
+%{_libdir}/cmake/*
 
 %files plugins
 %license LICENSE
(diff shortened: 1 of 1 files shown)


reubeno:
Thanks for the gentle reminder :). I've added a changelog entry.

I also see spec linting isn't happy, but presumably isn't happy with what's already checked into `main`. What's the recommendation for changes like this? Is it preferred that I keep the change more minimal (i.e., as-is) or is it requested to get clean since I'm going in and touching the spec anyhow?

SPECS/grpc/grpc.spec
```diff
@@ -98,7 +98,6 @@ popd
 %install
 pushd cmake/build
 %cmake_install
-find %{buildroot} -name '*.cmake' -delete
```
oliviacrain, line 101: @neha170 It's been a while, but do you remember why you excluded these files from packaging when writing the original spec? No worries if not, just want to make sure there isn't something else we need to address before taking this change.

oliviacrain:
> I also see spec linting isn't happy, but presumably isn't happy with what's already checked into `main`. What's the recommendation for changes like this? Is it preferred that I keep the change more minimal (i.e., as-is) or is it requested to get clean since I'm going in and touching the spec anyhow?

The generally accepted view is that linting is appreciated but not necessary for small, targeted changes. Sometimes the linter gives us some bad suggestions (_mea culpa_, working on addressing those), so I'd take suggestions with a grain of salt if they are questionable/don't seem to add much value. If you find a bad suggestion, feel free to open a bug in ADO and assign it to me for fixing.

In this case all of the suggestions look safe to take- fixing whitespace issues and alphabetizing build requirements shouldn't break anything here!

SPECS/grpc/grpc.spec
neha170: @oliviacrain unfortunately, I don't remember at all. I am guessing the linter suggested this.

oliviacrain: I don't believe that's the case- the linter only corrects already-present `*.la` removal commands to use a standardized command.

Thanks for taking a look! Resolving, since it looks like there's no concerns over merging this.

oliviacrain (approved):
[Ran a buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=382641&view=results), all looks good there. Thanks for the fix!

PR #5744, opened 2023-06-27
Fixed the `PACKAGE_CACHE_SUMMARY` build option.

PawelWMS:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Restoring the `PACKAGE_CACHE_SUMMARY` feature. The feature required adjustment to work with how the toolkit has changed since Mariner's 1.0 version. In order to prevent further bloat in `RpmRepoCloner`, the class has been refactored, so that enabling/disabling of a specific package repository/set of repositories can be done from one place (see: `SetEnabledRepos`), and that it can be updated after the cloner has been initialized.

**QUESTION:** the `RepoCloner` interface seems unnecessary. We only have one implementation and there seems to be no need for any other one in the foreseeable future. Can we remove it?

###### Change Log  

- `rpmrepocloner.go`:
  - Updated cloning to use batches of up most 50 packages instead of cloning one at a time.
  - Fixed a bug where `initializeMountedChrootRepo` wouldn't refresh the package manager's cache after rebuilding the local repositories with `createrepo`. This caused the cloner to randomly find or miss cached packages.
  - Moved enabling/disabling repositories out of the init function and into `SetEnabledRepos` method.
  - Replaced disabling network access with disabling upstream repositories with `SetEnabledRepos` and the `RepoFlagUpstream` flag.
  - Moved dynamic discovery of the cache folder into the init function.
  - Changed the logic for disabling default Mariner repositories. Instead of manipulating the `.repo` files, the logic now depends on repo IDs and has been moved to `SetEnabledRepos` and the set of `RepoFlag` flags.
  - Changed the logic for incremental enabling of searched repositories for `Clone` and `WhatProvides`. Now the class uses the `reposArgsList` member variable, which is an ordered list of TDNF `enablerepo` and `disablerepo` switches. Calls to `SetEnabledRepo` update this variable to reflect the currently enabled repositories.
  - `ClonedRepoContents` will now not list duplicate packages.
  - Updated `convertPackageVersionToTdnfArg` to use `-` for the `=` condition to handle both summary files from before and after this change. Using `=` for packages with epoch greater than 0 makes TDNF find no packages.
- `graphpkgfetcher.go`:
  - Updated the node resolution logic when we want to run a build using a cache summary file. In the old solution the fetcher skipped matching the unresolved nodes if a summary file was used. Now we download the files from the summary, switch the cloner to use only the toolkit and cached packages, and attempt to find packages for the unresolved nodes.
  - Fixed `assignRPMPath` to not updated the node in case of a failure.
  - Reordered methods and name to start with public methods, then private ones, then local functions. Each group in alphabetical order.
- `graphpkgfetcher.go` and `imagepkgfetcher.go`: switched to using the new cloner constructor and repo enabling/disabling logic.
- `repocloner.go`: added an `ID()` function for the `RepoPackage` class to be able to use unique identifiers in sets. This simplifies removing duplicates during package restoration from the package summaries.
- `repoutils.go`:
  - Updated to handle package duplicates in the summary files.
  - Moved skipping already downloaded packages and final verification into separate function.
- `rpm.go`:
  - Updated listing built packages from a spec to include the epoch. Packages with epochs greater than 0 confuse TDNF, if both 0 and higher epoch versions of the same package are available in the repository.
  - Updated relevant tests.

No.

###### Test Methodology

- [Full pipeline build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=384854&view=results).
- Local livepatch builds restoring the whole PMC repository using RPM snapshots from Mariner's monthly releases.


toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -69,325 +92,266 @@ const (
...
-	r.chroot = safechroot.NewChroot(tmpDir, isExistingDir)
+		finalArgs := make([]string, len(constantArgs)+len(currentPackageNames))
+		finalArgs = append(constantArgs, currentPackageNames...)
 
-	r.cloneDir = destinationDir
+		logger.Log.Debugf("Cloning batch %d to %d: %v", i, batchEndIndex, currentPackageNames)
+		err = r.chroot.Run(func() (err error) {
+			var chrootErr error
+			// Consider the toolchain RPMs first, then built RPMs, then the already cached, and finally all remote packages.
+			preBuilt, chrootErr = r.clonePackage(finalArgs)
```
dmcilvaney, line 192: We will stomp on the value of `preBuilt` here. We need to move this to a map, list maybe. We use this to report if a given package is cloned from the toolchain repo.

PawelWMS: That's true. It's a bug we have in the current code and I didn't focus on fixing this particular issue here. Having said that, I'll take a look, if I can resolve with a minor adjustemnt. Otherwise, I'll leave it as is.

dmcilvaney: This should be good for now.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -23,17 +24,26 @@ import (
...
-	toolchainRepoId        = "toolchain-repo"
-	cacheRepoID            = "upstream-cache-repo"
-	squashChrootRunErrors  = false
-	chrootDownloadDir      = "/outputrpms"
-	leaveChrootFilesOnDisk = false
-	updateRepoID           = "mariner-official-update"
-	previewRepoID          = "mariner-preview"
-	fetcherRepoID          = "fetcher-cloned-repo"
-	cacheRepoDir           = "/upstream-cached-rpms"
+	RepoFlagMarinerDefaults = uint64(1) << iota
```
dmcilvaney, line 27: I don't think we use bit flags very much, could we add a quick comment here for future readers?
```suggestion
	// Use bit shifted iotas to generate a set of bit flags (ie 0x001, 0x010, 0x100, ...)
	RepoFlagMarinerDefaults = uint64(1) << iota
```

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -23,17 +24,26 @@ import (
...
+	RepoFlagAll = RepoFlagToolchain | RepoFlagLocalBuilds | RepoFlagDownloadedCache | RepoFlagPreview | RepoFlagMarinerDefaults | RepoFlagUpstream
+)
+
+const (
+	chrootCloneDirContainer = "/upstream-cached-rpms"
+	chrootCloneDirRegular   = "/outputrpms"
+
+	repoIDAll            = "*"
+	repoIDBuilt          = "local-repo"
+	repoIDCacheContainer = "upstream-cache-repo"
```
dmcilvaney, line 43: Probably not now, but these names really need to be sanitized... `upstream-cache-repo` vs `fetcher-cloned-repo`, and `/upstream-cached-rpms` vs `/outputrpms` are the same thing for different environments but have such different naming conventions.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -406,23 +437,15 @@ func (r *RpmRepoCloner) WhatProvides(pkgVer *pkgjson.PackageVer) (packageNames [
...
+	for _, reposArgs := range r.reposArgsList {
+		logger.Log.Debugf("Using repos args: %v", reposArgs)
 
 		err = r.chroot.Run(func() (err error) {
-			completeArgs := append(baseArgs, fmt.Sprintf("--enablerepo=%s", repoID))
-
-			if !r.usePreviewRepo {
-				completeArgs = append(completeArgs, fmt.Sprintf("--disablerepo=%s", previewRepoID))
-			}
+			completeArgs := append(baseArgs, reposArgs...)
```
dmcilvaney, line 448: It looks like we never use more than one arg at a time here. It took me a sec to figure out what was going on. I'd recommend we switch this to just a `[]string` instead of `[][]string`.

```suggestion
			completeArgs := append(baseArgs, reposArgs)
```

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -69,60 +92,56 @@ const (
...
-	chroot         *safechroot.Chroot
-	usePreviewRepo bool
-	cloneDir       string
+	chroot                *safechroot.Chroot
+	chrootCloneDir        string
+	defaultMarinerRepoIDs []string
+	externalRepoIDs       map[string]bool
+	mountedCloneDir       string
+	repoIDCache           string
+	reposArgsList         [][]string
```
dmcilvaney, line 101: I think we can get away with just `[]string` here.
```suggestion
	reposArgsList         []string
```

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -658,25 +664,139 @@ func convertPackageVersionToTdnfArg(pkgVer *pkgjson.PackageVer) (tdnfArg string)
...
+	if RepoFlagDownloadedCache&reposFlags != 0 {
+		previousReposList = append(previousReposList, fmt.Sprintf("--enablerepo=%s", r.repoIDCache))
+		r.reposArgsList = append(r.reposArgsList, previousReposList)
+	}
+
+	// Options past this point are only valid if upstream repos are enabled.
+	if RepoFlagUpstream&reposFlags == 0 {
+		return
+	}
+
```
dmcilvaney, line 721: Do we want to preferentially enable the user's selected repos before resorting to `"*"`?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -292,20 +320,20 @@ func (r *RpmRepoCloner) initializeRepoDefinitions(disableDefaultRepos bool, repo
...
-			if err != nil {
-				return
-			}
-		} else {
-			logger.Log.Debugf("Disabling repositories listed in %s", originalRepoFilePath)
+		repoIDs, err := readRepoIDs(originalRepoFilePath)
+		if err != nil {
+			return err
+		}
+		r.defaultMarinerRepoIDs = append(r.defaultMarinerRepoIDs, repoIDs...)
```
dmcilvaney, line 328: I think we are poluting the default repos here with any custom repo files the user passes in.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -658,25 +664,139 @@ func convertPackageVersionToTdnfArg(pkgVer *pkgjson.PackageVer) (tdnfArg string)
...
+}
+
+func (r *RpmRepoCloner) reposArgsHaveOnlyLocalSources(reposArgs []string) bool {
+	const repoIDIndex = 1
+
+	for _, repoArg := range reposArgs {
+		if strings.Contains(repoArg, "--enablerepo=") {
+			repoID := strings.Split(repoArg, "=")[repoIDIndex]
+			if r.externalRepoIDs[repoID] {
+				return false
```
dmcilvaney, line 796: I think we need to explicitly track the toolchain + out/RPMS repos since the user can supply arbitrarily many external repos.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Per offline discussion, we do in fact want to keep a `[][]string` so we can preserve an inremental enabling of repos without recalculating the arguments.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Keep, `[][]string`

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Per offline discussion, we rely on passing tooling related repos alongside custom repos and it is non-trivial to differentiate them.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Could you also add a comment for each one inidcating its intended use?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: When the call to `disabledDefaultMarinerReposArgs()` is made inside `SetEnabledRepos()` I think it will also turn off the custom repo files.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Per offline discussion I get the logic now. It may be more clear to explicitly check against the out/RPMs or toolchain repo IDs just for readibility however.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: I agree. For now I tried to map them to more uniformly named variables. Seeing how we're moving away from using container builds, maybe we'll be able to completely remove one of these variants, since the container support was implemented only for internal use.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: Fortunately, we're not. The default repos are read from inside the chroot's directory. The user-defined ones are never copied into the chroot and we only parse them from the paths provided by the user.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: Sounds good.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: I've made the slight change of saying that `preBuilt` means **ALL** packages were pre-built. This is consistent with the current behaviour. I wasn't sure which way we may want to go for the actual fix and it seemed too risky and robust regardless of the direction to add it to this PR.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Unfortunately customers are not requesting docker based builds for dev use... so they are boomaranging RIGHT back again :P

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
dmcilvaney: Allright, maybe add a comment on the order of operations but seems reasonable.

toolkit/tools/internal/packagerepo/repoutils/repoutils.go
```diff
@@ -107,3 +71,89 @@ func SaveClonedRepoContents(cloner repocloner.RepoCloner, dstFile string) (err e
...
+
+func removeDownloadedPackages(packages []*repocloner.RepoPackage, cloneDirectory string) []*pkgjson.PackageVer {
+	const packageCondition = "="
+
+	packageVers := make([]*pkgjson.PackageVer, len(packages))
+	index := 0
+	for _, pkg := range packages {
+		pkgVersion := fmt.Sprintf("%s.%s", pkg.Version, pkg.Distribution)
+
+		// Skip packages that are already present, this is expected for the toolchain
```
dmcilvaney, line 100: The toolchain RPMs should no longer be present in the chache, they have their own directory now. I think the only reason a package would be here is if it was downloaded as a dependency of another package (but we set that to `false`...). I think if we want to enforce a clean state this could be a hard error.

dmcilvaney: Although since there is a call to `verifyClonedRepoContents()` can we just relax the requirement of a clean checkout here? It could just delete any unexpected RPMs?

dmcilvaney (approved):

toolkit/tools/internal/packagerepo/repoutils/repoutils.go
PawelWMS: I've added this function to handle cases of multiple rebuilds with the same summary file. Requiring an empty state seemed overly restrictive and time-consuming. I realized that it's not an error to have pre-existing packages as long as the final set of available packages equaled the ones in the summary file.

I like the idea of removing any unexpected RPMs for builds with summary files. It would be a change in the contract (error out vs silently remove packages) but since this feature hasn't been used by anyone, I think we could get away with it and that's a nice improvement. Having said that, would you mind if I left that change for an undetermined future? Reason: my human limitations. The change seems to be in a stable state now and my mind needs a reset.

PawelWMS: Quick improvement: I may re-name the function to `filterOutDownloadedPackages`. The word "remove" is hinting at deleting the packages instead of removing them from the list.

PR #5795, opened 2023-07-06
glibc: restore glibc-debuginfo package

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Our glibc package does not generate debuginfo, but since it could be useful to debug some glibc components, I've modified the os postinstall script so that symbols are not removed.

###### Change Log  

- Restore debuginfo package in glibc
- Add missing BR on gawk in glibc.spec
- Change toolchain manifests for glibc
- Change packages that build with glibc-static to use the newest version

**YES**

###### Test Methodology

- Delta build x64: 389270
- Full build: 2.0.20230710-anphel-391291


oliviacrain (approved):

oliviacrain:
Just curious- what sort of size impact does this change have on the relevant packages? Are they big enough that we might trip some "image too big" BVTs that we need to look for in shiproom?

bossmc:
This fix feels wrong, we should just build the debuginfo package for glibc (see line 3 of the spec file where it's disabled for these packages with no comment).  The RPM ecosystem has good support for debug-symbol management and deliberately sidestepping it seems like a source of maintenance headache.

I don't know if Mariner publishes its debuginfo packages to the standard repository or to a separate one (Ubuntu and Fedora both split their repos so you don't have to fetch all the debuginfo metadata if you're just using the distro, not debugging it) but it feel like the right fix is to allow a `glibc-debuginfo` package to be built and then install it in order to debug glibc issues.  This also resolves Oliva's worry about image size bloat.

bossmc:
As an aside the `glibc` specfile is missing a `BuildRequires` for `gawk` (it's mandated by the configure stage to be present)

anphel31:
> Just curious- what sort of size impact does this change have on the relevant packages? Are they big enough that we might trip some "image too big" BVTs that we need to look for in shiproom?

That's a good question. I checked, and the main glibc.rpm size does increase from 4MB to 12MB with this change.

SPECS/busybox/busybox.spec
```diff
@@ -94,6 +94,9 @@ install -m 644 docs/busybox.petitboot.1 %{buildroot}/%{_mandir}/man1/busybox.pet
 %{_mandir}/man1/busybox.petitboot.1.gz
 
 %changelog
+* Wed Jul 05 2023 Andrew Phelps <anphel@microsoft.com> - 1.35.0-5
+- Bump release to rebuild against glibc 2.35-4
+
 * Fri Oct 07 2022 Andy Caldwell <andycaldwell@microsoft.com> - 1.35.0-4
```
anphel31, line 100: Note: it looks like the release was not actually bumped to 4 with this previous change. I've skipped release 4 here and gone directly to 5.

anphel31:
> This fix feels wrong, we should just build the debuginfo package for glibc (see line 3 of the spec file where it's disabled for these packages with no comment). The RPM ecosystem has good support for debug-symbol management and deliberately sidestepping it seems like a source of maintenance headache.
> 
> I don't know if Mariner publishes its debuginfo packages to the standard repository or to a separate one (Ubuntu and Fedora both split their repos so you don't have to fetch all the debuginfo metadata if you're just using the distro, not debugging it) but it feel like the right fix is to allow a `glibc-debuginfo` package to be built and then install it in order to debug glibc issues. This also resolves Oliva's worry about image size bloat.

I agree that having a debuginfo package would probably be best. The reason we don't have a debuginfo package for glibc is because of an old bug (`Bug 26939136: Mariner gdb can't debug multithreaded applications by default`). The developer who made the change has left and the bug details are inaccessible now. However, that bug was for an older gdb and glibc version than we have now. I'll see if a simple multithreaded app can be debugged with gdb.

SPECS/busybox/busybox.spec
bossmc: Oh, my bad, sorry, must have missed it.  It's not worth cutting a `-4` jumping to `-5` is fine.

PR #5798, opened 2023-07-07
Support N+1 goal nodes for scheduler

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When we update a package we will want to run package tests (PTests) on it. (aka %check sections). This gives us confidence we have not broken that package with the change. To expand on this further we would also like to know if there are any unexpected impacts to downstream packages (i.e., those packages which depend on the updated package).

When we select a package to build we do so by adding a goal into  our dependency graph. This feature will allow the graph to include not only the selected package, but also any `n` layers of the graph above the selected packages. 

A 'layer' is defined by a run node (i.e., we will not count meta, build, etc. nodes towards the `n` layers).

The scheduler will remain responsible for determining which packages must be rebuild to support the requested set (ie dependencies). Each additional layer will likely pull in more dependencies to build.

```mermaid
---
title: Request to build Node "D"
---
graph TB
classDef doNotBuild fill:#bb6666
classDef doBuild fill:#66bb66
class A-Build,B-Build,C-Build doNotBuild;

GOAL:::doBuild
A-Run:::doNotBuild
A-Build:::doNotBuild
B-Run:::doNotBuild
B-Build:::doNotBuild
C-Run:::doNotBuild
C-Build:::doNotBuild
meta:::doNotBuild
D-Run:::doBuild
D-Build:::doBuild

GOAL ---> D-Run

A-Run ---> A-Build
A-Run ---> B-Run
B-Run ---> B-Build
C-Run ---> C-Build
C-Build ---> D-Run
D-Run ---> D-Build
B-Run ---> meta
meta ---> D-Run
```

```mermaid
---
title: Request to build Node "D" with N+1
---
graph TB
classDef doNotBuild fill:#bb6666
classDef doBuild fill:#11bb11
classDef doBuildOld fill:#66bb66
class A-Build,B-Build,C-Build doNotBuild;

GOAL:::doBuildOld
A-Run:::doNotBuild
A-Build:::doNotBuild
B-Run:::doBuild
B-Build:::doBuild
C-Run:::doBuild
C-Build:::doBuild
meta:::doBuild
D-Run:::doBuildOld
D-Build:::doBuildOld

GOAL ---> D-Run

A-Run ---> A-Build
A-Run ---> B-Run
B-Run ---> B-Build
C-Run ---> C-Build
C-Build ---> D-Run
D-Run ---> D-Build
B-Run ---> meta
meta ---> D-Run
```

```mermaid
---
title: Request to build Node "D" N+2
---
graph TB
classDef doNotBuild fill:#bb6666
classDef doBuild fill:#11bb11
classDef doBuildOld fill:#66bb66
class A-Build,B-Build,C-Build doNotBuild;

GOAL:::doBuildOld
A-Run:::doBuild
A-Build:::doBuild
B-Run:::doBuildOld
B-Build:::doBuildOld
C-Run:::doBuildOld
C-Build:::doBuildOld
meta:::doBuildOld
D-Run:::doBuildOld
D-Build:::doBuildOld

GOAL ---> D-Run

A-Run ---> A-Build
A-Run ---> B-Run
B-Run ---> B-Build
C-Run ---> C-Build
C-Build ---> D-Run
D-Run ---> D-Build
B-Run ---> meta
meta ---> D-Run
```

###### Change Log  

- `AddGoalNode()` now supports an `extraLayers` argument that expands the nodes it is connected to.

**NO**

###### Test Methodology

- Pipeline build id: xxxx


toolkit/Makefile
```diff
@@ -48,6 +48,7 @@ INITRD_CACHE_SUMMARY            ?=
 PACKAGE_ARCHIVE                 ?=
 PACKAGE_BUILD_RETRIES           ?= 1
 CHECK_BUILD_RETRIES             ?= 1
+EXTRA_BUILD_LAYERS              ?= 0
```
PawelWMS, line 51: Since we're here and it's related, I think it's time to remove `REBUILD_DEP_CHAINS` both from this file and the `building.md` one.

toolkit/docs/building/building.md
```diff
@@ -803,6 +803,7 @@ To reproduce an ISO build, run the same make invocation as before, but set:
 | ALLOW_TOOLCHAIN_REBUILDS      | n                                                                                                      | Do not treat rebuilds of toolchain packages during regular package build phase as errors.
 |  PACKAGE_BUILD_RETRIES        | 1                                                                                                      | Number of build retries for each package
 | CHECK_BUILD_RETRIES           | 1                                                                                                      | Minimum number of check section retries for each package if RUN_CHECK=y and tests fail.
+| EXTRA_BUILD_LAYERS            | 0                                                                                                      | How many additional layers of the build graph to build beyond the requested packages (useful for testing changes in dependent packages)
```
PawelWMS, line 806: Won't the value of 0 change the current default behaviour where we're rebuilding everything, that depends on an updated package?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -120,6 +120,7 @@ func ConstructCloner(destinationDir, tmpDir, workerTar, existingRpmsDir, toolcha
 	err = r.initialize(destinationDir, tmpDir, workerTar, existingRpmsDir, toolchainRpmsDir, repoDefinitions)
 	if err != nil {
 		err = fmt.Errorf("failed to prep new rpm cloner:\n%w", err)
+		return
```
PawelWMS, line 123: Thanks for catching that!

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1019,8 +1019,22 @@ func (g *PkgGraph) AddMetaNode(from []*PkgNode, to []*PkgNode) (metaNode *PkgNod
 	return
 }
 
-// AddGoalNode adds a goal node to the graph which links to existing nodes. An empty package list will add an edge to all nodes
+// AddGoalNodeWithLayers adds a goal node to the graph which links to existing nodes. An empty package list will add an edge to all nodes
```
PawelWMS, line 1022:
```suggestion
// AddGoalNode adds a goal node to the graph which links to existing nodes. An empty package list will add an edge to all nodes
```

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1092,6 +1106,91 @@ func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer,
...
 
+	// Expand the goal node if requested
+	if extraLayers > 0 {
+		g.addGoalNodeLayers(goalNode, extraLayers)
+	}
+
+	return
+}
+
+// addGoalNodeLayers will expand a goal node by one layer. For example, if the goaled node is "foo" (ie the goal node
```
PawelWMS, line 1117:
```suggestion
// addGoalNodeLayers will expand a goal node by one layer. For example, if the goaled node is "foo" (i.e. the goal node
```

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1092,6 +1106,91 @@ func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer,
...
+// points to "foo") and extraLevels is 1, the goal node will now link to all nodes which depend on "foo" as well as
+// "foo" itself. A node is considered to depend on "foo" if it is a run node that has edges connecting it to "foo"
+// without any other run nodes in between.
+//
+//	i.e., if "foo_run" -> "foo_build" -> "<Some Meta Node>" -> "bar_run" -> "bar_build", and we are expanding from "bar"
+//	with layers=1, only "foo_run" will be added to the goal nodes since "foo_build" and "<Some Meta Node>" are not run nodes.
```
PawelWMS, line 1123: Let's keep the example consistent with the description above. Right now we've changed to dependency order.
```suggestion
//	E.g., if "bar_run" -> "bar_build" -> "<Some Meta Node>" -> "foo_run" -> "foo_build", and we are expanding from "foo"
//	with layers=1, only "bar_run" will be added to the goal nodes since "bar_build" and "<Some Meta Node>" are not run nodes.
```

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1019,8 +1019,22 @@ func (g *PkgGraph) AddMetaNode(from []*PkgNode, to []*PkgNode) (metaNode *PkgNod
...
+}
+
+// AddGoalNodeWithExtraLayers adds a goal node to the graph which links to existing nodes. An empty package list will add an edge to all nodes
+//   - goalName: The name of the goal node to add
+//   - packages: A list of packages to add to link the goal node to. If empty, all nodes will be added to the goal node
+//   - strict: If true, the goal node will fail if any of the packages are not found
+//   - extraLayers: The number of levels to expand the goal node. Each level will add one more layer of packages beyond
+//     the goal node. For example, if the goal node is "foo" and extraLevels is 1, the goal node will link to all nodes
+//     which depend on "foo" as well as "foo" itself (Specifically run nodes, all other nodes are stepped over)
+func (g *PkgGraph) AddGoalNodeWithExtraLayers(goalName string, packages []*pkgjson.PackageVer, strict bool, extraLayers int) (goalNode *PkgNode, err error) {
```
PawelWMS, line 1037: Question about a scenario we want to support.

Context:
- We've just published all latest packages to PMC.
- We're now fixing a critical CVE in `foo`.

The delta build will now run `make build-packages CONFIG_FILE= DELTA_FETCH=y EXTRA_BUILD_LAYERS=1`. It is not expected to try and figure out, which packages changed, thus no `PACKAGE_BUILD_LIST` argument. Will this implementation allow us to build `foo` and its closest dependents?

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1092,6 +1106,91 @@ func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer,
...
+
+// addGoalNodeLayers will expand a goal node by one layer. For example, if the goaled node is "foo" (ie the goal node
+// points to "foo") and extraLevels is 1, the goal node will now link to all nodes which depend on "foo" as well as
+// "foo" itself. A node is considered to depend on "foo" if it is a run node that has edges connecting it to "foo"
+// without any other run nodes in between.
+//
+//	i.e., if "foo_run" -> "foo_build" -> "<Some Meta Node>" -> "bar_run" -> "bar_build", and we are expanding from "bar"
+//	with layers=1, only "foo_run" will be added to the goal nodes since "foo_build" and "<Some Meta Node>" are not run nodes.
+func (g *PkgGraph) addGoalNodeLayers(goalNode *PkgNode, layers int) {
+	logger.Log.Debugf("Expanding goal node %s by %d layers", goalNode.GoalName, layers)
```
PawelWMS, line 1125: Can we surround the goal node's name with brackets or quotes? This makes it easier to debug. Especially if the string is empty or has spaces in it.

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1092,6 +1106,91 @@ func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer,
...
+	}
+
+	// For each node in the current layer, add all of the nodes that depend on it, then repeat as many times as requested
+	expandedGoalNodes = initialGoalNodes
+	for i := 0; i < layers; i++ {
+		expandedGoalNodes = append(expandedGoalNodes, g.getNextGoalLayer(expandedGoalNodesSet, expandedGoalNodes)...)
+	}
+
+	// Add the new edges if they are missing
+	for _, n := range expandedGoalNodes {
```
PawelWMS, line 1144: Let's use full words like "node", "expandedGoalNode", etc.

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1092,6 +1106,91 @@ func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer,
...
+		if n == goalNode {
```
PawelWMS, line 1145: Should we also check that `n` is not inside `initialGoalNodes`? I think it won't matter for the build itself, but the graph will be a bit easier to read.

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1092,6 +1106,91 @@ func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer,
...
+// points to "foo") and extraLevels is 1, the goal node will now link to all nodes which depend on "foo" as well as
+// "foo" itself. A node is considered to depend on "foo" if it is a run node that has edges connecting it to "foo"
+// without any other run nodes in between.
+//
+//	i.e., if "foo_run" -> "foo_build" -> "<Some Meta Node>" -> "bar_run" -> "bar_build", and we are expanding from "bar"
+//	with layers=1, only "foo_run" will be added to the goal nodes since "foo_build" and "<Some Meta Node>" are not run nodes.
+func (g *PkgGraph) addGoalNodeLayers(goalNode *PkgNode, layers int) {
+	logger.Log.Debugf("Expanding goal node %s by %d layers", goalNode.GoalName, layers)
+
+	var expandedGoalNodes []*PkgNode
```
PawelWMS, line 1127: I think we can go ahead and drop slices entirely in favour of "sets". Maps are passed by reference, so `getNextGoalLayer` can work directly on its input.

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1092,6 +1106,91 @@ func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer,
...
+			expandedGoalNodesSet[goalNode] = true
+		}
+
+		// Add all the nodes that depend on this node to the expanded goal nodes list. If the dependant node is a run
+		// node we can stop expanding from it (A subsequent call to expandGoalNodesOnce() will expand from it further if
+		// needed). If the dependant node is a build, meta, etc. node we need to keep expanding from it since we only
+		// care about adding goals to run nodes. We ignore goal nodes since they should have no dependents, and may
+		// potentially pull in unrelated parts of the graph.
+		dependentNodes := graph.NodesOf(g.To(goalNode.ID()))
+		for _, dependentNeighbor := range dependentNodes {
```
PawelWMS, line 1179: I'd avoid calling these "neighbours". While technically correct, let's be clear we mean dependents or parent nodes, so there's no mistake about the relationship between the two.

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1092,6 +1106,91 @@ func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer,
...
+// - currentGoalNodes: The current layer of goal nodes to expand from.
+//
+// Returns a list of additional nodes that connect to currentGoalNodes (but not any nodes that are already in currentGoalNodes)
+func (g *PkgGraph) getNextGoalLayer(expandedGoalNodesSet map[*PkgNode]bool, currentGoalNodes []*PkgNode) (expandedGoalNodes []*PkgNode) {
+
+	// We will iterate over the current goal nodes and add all the nodes that depend on them to the expanded goal nodes.
+	// The expandedGoalNodesSet will ensure we don't add the same node twice.
+	for _, goalNode := range currentGoalNodes {
+		if expandedGoalNodesSet[goalNode] {
+			logger.Log.Tracef("Already expanded from %s, skipping", goalNode.FriendlyName())
```
PawelWMS, line 1166: Similar request for all log messages here: let's surround node names with something, so it's clear where they begin and end, and what is the remainder of the message.

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -245,14 +265,7 @@ func checkTestGraph(t *testing.T, g *PkgGraph) {
...
 	}
-	for _, mustHave := range component1 {
-		found := false
-		for _, n := range g.AllNodesFrom(a.RunNode) {
-			found = found || mustHave.Equal(n)
-		}
-		assert.True(t, found)
-	}
-	assert.Equal(t, len(component1), len(g.AllNodesFrom(a.RunNode)))
+	checkEqualComponents(t, component1, g.AllNodesFrom(a.RunNode))
```
PawelWMS, line 268: I think we're not creating clones of the test nodes anyway, so working on pointers should be fine. If that's the case, won't this work instead of `checkEqualComponents`?
```suggestion
	assert.ElementsMatch(t, component1, g.AllNodesFrom(a.RunNode))
```

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -840,6 +919,84 @@ func TestMetaNodeAddPkg(t *testing.T) {
 	assert.Equal(t, 5, len(g.AllNodesFrom(c.RunNode)))
 }
 
+func TestGoalWithLevelOneAndMeta(t *testing.T) {
+	g, err := buildTestGraphHelper()
+	assert.NoError(t, err)
+	assert.NotNil(t, g)
+
+	c1, _ := g.FindBestPkgNode(&pkgC)
+	c2, _ := g.FindBestPkgNode(&pkgC2)
```
PawelWMS, line 928: Why ignore the returned `err` instead of asserting they are not nil?

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -840,6 +919,84 @@ func TestMetaNodeAddPkg(t *testing.T) {
...
+		pkgD4Unresolved,
+		pkgD5Unresolved,
+		pkgD6Unresolved,
+		meta,
+		goal,
+	}
+	checkEqualComponents(t, expectedGoalPackages2meta, g.AllNodesFrom(goal))
+}
+
+func TestGoalWithMultipleGoals(t *testing.T) {
```
PawelWMS, line 963: I think we're testing also the fact we're adding more layers here. Suggestion:
```suggestion
func TestGoalWithMultipleGoalsAndOneExtraLayer(t *testing.T) {
```

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -840,6 +919,84 @@ func TestMetaNodeAddPkg(t *testing.T) {
...
+		pkgD2Unresolved,
+		pkgD3Unresolved,
+		pkgC2Run,
+		pkgC2Build,
+		pkgD4Unresolved,
+		pkgD5Unresolved,
+		pkgD6Unresolved,
+		goal,
+	}
+	checkEqualComponents(t, expectedGoalPackages2meta, g.AllNodesFrom(goal))
```
PawelWMS, line 997:
```suggestion
	expectedGoalPackages2Meta := []*PkgNode{
		pkgBRun,
		pkgBBuild,
		pkgCRun,
		pkgCBuild,
		pkgD2Unresolved,
		pkgD3Unresolved,
		pkgC2Run,
		pkgC2Build,
		pkgD4Unresolved,
		pkgD5Unresolved,
		pkgD6Unresolved,
		goal,
	}
	checkEqualComponents(t, expectedGoalPackages2Meta, g.AllNodesFrom(goal))
```

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -1019,8 +1019,22 @@ func (g *PkgGraph) AddMetaNode(from []*PkgNode, to []*PkgNode) (metaNode *PkgNod
...
 func (g *PkgGraph) AddGoalNode(goalName string, packages []*pkgjson.PackageVer, strict bool) (goalNode *PkgNode, err error) {
+	return g.AddGoalNodeWithExtraLayers(goalName, packages, strict, 0)
+}
+
+// AddGoalNodeWithExtraLayers adds a goal node to the graph which links to existing nodes. An empty package list will add an edge to all nodes
+//   - goalName: The name of the goal node to add
+//   - packages: A list of packages to add to link the goal node to. If empty, all nodes will be added to the goal node
+//   - strict: If true, the goal node will fail if any of the packages are not found
+//   - extraLayers: The number of levels to expand the goal node. Each level will add one more layer of packages beyond
+//     the goal node. For example, if the goal node is "foo" and extraLevels is 1, the goal node will link to all nodes
```
cwize1, line 1035: nit: Probably better to use "x" instead of "foo".

While using "foo" and "bar" aren't strictly prohibited because of how commonly they are used within code documentation, it is discouraged due to their meanings in other contexts.

toolkit/tools/internal/pkggraph/pkggraph.go
cwize1, line 1117: "one layer"? Seems like the number of layers is a parameter.

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -225,7 +225,27 @@ func buildTestGraphHelper() (g *PkgGraph, err error) {
 	return
 }
 
+// Checks if two lists of nodes are equivalent (ignoring order and graph edges)
+func checkEqualComponents(t *testing.T, expected, actual []*PkgNode) {
+	t.Helper()
+	foundPackage := make(map[*PkgNode]bool)
+	for _, mustHave := range expected {
```
cwize1, line 232: Is the `foundPackage` map needed?

Seems like you could just write:

```golang
for _, mustHave := range expected {
	found := false
	for _, n := range actual {
		found = found || mustHave.Equal(n)
	}
	assert.True(t, found, "expected to find %s in actual", mustHave.String())
}
```

cwize1 (changes requested):

toolkit/docs/building/building.md
dmcilvaney: This is how many additional layers we should add, we always have an implicit layer 0 which includes the packages that are directly requested.

toolkit/tools/internal/pkggraph/pkggraph.go
dmcilvaney: This implementation will have no meaningful impact in this scenario. We will be using the automatic ALL goal node which points to all packages, so adding addional layers will have no affect here. A follow on feature will be needed to better optimize this use case (by limiting the cascading rebuilds).

toolkit/tools/internal/pkggraph/pkggraph.go
dmcilvaney: It should be same to assume that all nodes in `initialGoaldNodes` DO have an edge to `g` already, which means they should be skipped.

toolkit/tools/internal/pkggraph/pkggraph_test.go
dmcilvaney: Fair point, I think this was the remains of a cut-down function with more output before.

toolkit/tools/internal/pkggraph/pkggraph_test.go
dmcilvaney: Unfortunately not, they underlying objects are completely separate graph objecst so a custom equality check is needed (we use `PkgNode.Equal()`)

toolkit/tools/internal/pkggraph/pkggraph_test.go
dmcilvaney: Cruft from an older larger test probably, I'll rename all of them

toolkit/tools/internal/pkggraph/pkggraph.go
dmcilvaney: Gave it a shot, its not straight forward from what I can see. Its hard to keep track of the state (ie what is the difference between a node we haven't seen one we have visited and are done with, or one we need to check still). I'm sure there is a way to do it but getting the recursive call to jump over build nodes right is not trivial from what I can see. It tends to either stop too early, or recurse indefinately, without additional meta data.

cwize1 (approved):

toolkit/tools/internal/pkggraph/pkggraph.go
PawelWMS: Ah, I see, I initially misunderstood. Now a reverse question came to my mind: if my understanding is correct, all true goal nodes (of type `TypeGoal`) are never in a cycle. And `goalNode` is a true goal node, so do we even need this check?

toolkit/tools/internal/pkggraph/pkggraph.go
PawelWMS: Unresolving - forgot to push a commit? And if you disagree, please at least tell us why.;)

toolkit/tools/internal/pkggraph/pkggraph.go
PawelWMS: Yeah, I see the complexity now, makes sense.

toolkit/tools/internal/pkggraph/pkggraph.go
dmcilvaney: Renamed from `neighbourNode` -> `dependentNode`. I'll rename the outer variables as well.

toolkit/tools/internal/pkggraph/pkggraph.go
dmcilvaney: Its to make sure we don't add a self-edge from goal -> goal. When we get the list of all dependents of the goal'd nodes, we will get the goal node itself back again. No reason to add an edge from the goal to the goal.

I'll add a comment to this affect.

dmcilvaney: I'm still going to leave it in with an eye to another PR which will allow arbitrary goal node creation from a list of other nodes as a defensive check.

toolkit/tools/internal/pkggraph/pkggraph.go
dmcilvaney: (We did this in https://github.com/microsoft/CBL-Mariner/pull/5993)

dmcilvaney:
Merged latest main

PR #5800, opened 2023-07-07
rpm: change default binary compression to zstd

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

RPMs in our build system are currently being sequentially compressed with gzip level 9 by default. We can save time building packages, and reduce rpm file sizes, by switching to a different compression type. This PR sets `%_binary_payload w7T0.zstdio` in our macro file to change the default to zstd level 7 multithreaded compression. zstd compression is very fast, and can reduce RPM file sizes from ~3% - ~15%.

Reasoning for choosing zstd level 7:
zstd seems to give us the best balance of reduced build time, as well as reduced file size. I ran some tests with zstd compression levels of 3 (the default), 19 (the value used in Fedora 31), and 7 (something in-between). I found that using 19 provides smaller file sizes, at the expense of much worse build times. Level 7 reduces the file size significantly while requiring only slightly more build time than 3. See below for a comparison of compression type/levels when building our kernel package (kernel-5.15.118.1-1.cm2.x86_64) (which has among the longest build times of our core packages). Note that xz-7 compression has an advantage in the smallest file size, but is still much slower than zstd-7.

| Compression Type | Compression level | Kernel-debuginfo RPM size | Time to generate kernel-debuginfo |
|--|--|--|--|
| gzip | 9 | 913 MB | 14 min 16 sec |
| xz | 7 | 447 MB | 6 min 12 sec |
| zstd | 3 | 718 MB | 39 sec |
| zstd | 7 | 650 MB | 43 sec |
| zstd | 19 | 529 MB | 21 min |

For reference, Fedora changed from gzip to xz in Fedora 12 [details](https://fedoraproject.org/wiki/Features/XZRpmPayloads). They changed to zstd, with compression level 19, in Fedora 31 [details](https://fedoraproject.org/wiki/Changes/Switch_RPMs_to_zstd_compression)

Description and sample compression options: [macros.in](https://github.com/rpm-software-management/rpm/blob/4588bc3f994338502d2770ad24cbfcdaa6c335ec/macros.in#L339)

Build time improvements:
These build times are for x64 OB buddy builds of some core mariner packages that take the longest to build:
kernel - 34 min (gzip) - 20 min (zstd)
pytorch - 69 min (gzip) - 57 min (zstd)
nodejs18 - 50 min (gzip) - 48 min (zstd)

File size reduction:
rpms.tar.gz file size changes, from main branch. This contains all core mariner RPM packages;
gzip - 21,010 MB
zstd - 17,884 MB

toolchain_built_rpms_all.tar.gz
gzip - 1,600 MB
zstd - 1,462 MB

###### Change Log  

- Change mariner-rpm-macros to set `%_binary_payload w7T0.zstdio`
- Change toolchain manifests

**YES**

###### Test Methodology

- Full build: 2.0.20230707-anphel-390037

To check the compression:
```
$ rpm -qp --qf "%{PAYLOADCOMPRESSOR} %{PAYLOADFLAGS} %{PAYLOADFORMAT}\n"    https://packages.microsoft.com/cbl-mariner/2.0/prod/base/x86_64/Packages/k/kernel-5.15.112.1-1.cm2.x86_64.rpm
gzip 9 cpio

$ rpm -qp --qf "%{PAYLOADCOMPRESSOR} %{PAYLOADFLAGS} %{PAYLOADFORMAT}\n"  ../out/RPMS/x86_64/kernel-5.15.116.1-12.cm2.x86_64.rpm
zstd 7T0 cpio
```


oliviacrain:
Does `rpm` have a dependency on `zstd`? Do we ship `zstd` on every image that ships `rpm`? Are there any other risks of taking this?

anphel31:
> Does `rpm` have a dependency on `zstd`? Do we ship `zstd` on every image that ships `rpm`?

Yes, our rpm-libs package [requires](https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS/rpm/rpm.spec#L61C21-L61C21)  zstd. And we've built rpm with zstd support since Mariner 1.0.

oliviacrain (approved):
LGTM!

anphel31:
> This generally looks great to me! I just have these 2 questions:
> 
> * zstd seems like a clear winner. Could you help educate us on how we arrived on level 7 compression as the default? Is that worth benchmarking as well to find the right compression time vs. decompression time vs. size trade-off? (Or perhaps that's something that was already done / already exists somewhere?)
> * From the referenced links, it sounds like this has a chance of speeding up decompression/install times too! I'd be curious to know if any non-negligible impact was seen in image build, say, as that's a process that does quite a bit of package installation.

I've updated the PR description with an explaination of why I chose zstd level 7. For the image build times, it's possible they will improve. However, with the test runs I performed so far, the build time seems roughly the same.

dmcilvaney:
Do you know if there is any way to swap in `pigz` here if its available? That was one of the biggest improvements to tooling stuff we found (outside rpm build).

dmcilvaney:
> Do you know if there is any way to swap in and test `pigz` here? That was one of the biggest improvements to tooling stuff we found (outside rpm build). We would need to add it to our chroot but its WAY faster than normal gzip (pigz = Parallel Implementation of Gzip)
> 
> edit: Looking at the macro files it seems like its a bit black-box unfortunately.

Had a dig, looks like this is using parallel zstd so we are not likely to see much improvement beyond this (the `T0` in `w7T0.zstdio` means # cpu threads).

SPECS/mariner-rpm-macros/macros
```diff
@@ -52,6 +52,9 @@
 %{?mariner_ccache_enabled:CCACHE_DIR="/ccache-dir" ; export CCACHE_DIR ;}\
 %{nil}
 
+# use zstd compression for binary package payloads
+%_binary_payload w7T0.zstdio
```
dmcilvaney, line 56: Do we want to set `%_source_payload` as well? Looks like that affects .src.rpm compression.

anphel31:
> > Do you know if there is any way to swap in and test `pigz` here? That was one of the biggest improvements to tooling stuff we found (outside rpm build). We would need to add it to our chroot but its WAY faster than normal gzip (pigz = Parallel Implementation of Gzip)
> > edit: Looking at the macro files it seems like its a bit black-box unfortunately.
> 
> Had a dig, looks like this is using parallel zstd so we are not likely to see much improvement beyond this (the `T0` in `w7T0.zstdio` means # cpu threads).

Right, even if we could use pigz, the zstd compression shrinks the file sizes quite a bit over gzip.

SPECS/mariner-rpm-macros/macros
anphel31: I briefly looked at changing the `%_source_payload`, but the source files do not compress very well, so I think the benefit would be minimal at best.

dmcilvaney (approved):

SPECS/mariner-rpm-macros/macros
anphel31: I performed a local build of "make input-srpms" with `%_source_payload w7T0.zstdio`. The size of the src.rpm files was nearly identical. Does not seem worth it to change this parameter.

For example, `kernel-5.15.116.1-2.cm2.src.rpm`, which is around 187 MB, shrunk by only 234 KB.
kernel-5.15.116.1-2.cm2.src.rpm - gzip 9 cpio   -  195699714 bytes
kernel-5.15.116.1-2.cm2.src.rpm - zstd 7T0 cpio -  195459770 bytes

PR #5802, opened 2023-07-07
Add functionality to serve stale DNS records

ddstreet:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

adds functionality to serve stale DNS records

###### Change Log  

add config param to systemd-resolved to retain dns records after expiry, and serve those stale records to clients if upstream ns is unreachable or reports error

NO

###### Test Methodology

- Pipeline build id: xxxx
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=390272&view=results

### DIFF
--- a/SPECS/systemd/systemd.spec
+++ b/SPECS/systemd/systemd.spec
@@ -1,7 +1,7 @@
 Summary:        Systemd-250
 Name:           systemd
 Version:        250.3
-Release:        16%{?dist}
+Release:        17%{?dist}
 License:        LGPLv2+ AND GPLv2+ AND MIT
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -25,6 +25,8 @@ Patch3:         CVE-2022-3821.patch
 Patch4:         CVE-2022-45873.patch
 Patch5:         backport-helper-util-macros.patch
 Patch6:         CVE-2022-4415.patch
+Patch7:         serve-stale-0001-resolved-added-serve-stale-feature-implementation-of.patch
+Patch8:         serve-stale-0002-resolved-Initialize-until_valid-while-storing-negati.patch
 BuildRequires:  audit-devel
 BuildRequires:  cryptsetup-devel
 BuildRequires:  docbook-dtd-xml
@@ -281,6 +283,9 @@ fi
 %files lang -f %{name}.lang
 
 %changelog
+* Fri Jul 07 2023 Dan Streetman <ddstreet@ieee.org> - 250.3-17
+- Add support to systemd-resolved to serve stale dns data
+
 * Tue Jun 20 2023 Chris Gunn <chrisgun@microsoft.com> - 250.3-16
 - Enable audit integration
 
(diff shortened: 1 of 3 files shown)


tobiasb-ms (approved):

SPECS/systemd/systemd.spec
```diff
@@ -281,6 +283,9 @@ fi
 %files lang -f %{name}.lang
 
 %changelog
+* Fri Jul 07 2023 Dan Streetman <ddstreet@ieee.org> - 250.3-17
+- Enable audit integration
```
hbeberman, line 287: Please update the changelog entry

dmcilvaney: This is a personal thing, but I like to link back to the [PR/commit](https://github.com/systemd/systemd/pull/27681) we pull backports from (either as a comment in the spec, or in the changelog). Makes updating the spec easier since we can quick tell if we need to preserve a patch or if we can just drop it when we version bump.

Not a deal breaker, just a personal preference.

ddstreet: thanks! updated

ddstreet: > This is a personal thing, but I like to link back to the [PR/commit](https://github.com/systemd/systemd/pull/27681) we pull backports from (either as a comment in the spec, or in the changelog). Makes updating the spec easier since we can quick tell if we need to preserve a patch or if we can just drop it when we version bump.
> 
> Not a deal breaker, just a personal preference.

Patch files always have the commit in the first line, e.g.:
```
$ head -1 serve-stale-0001-resolved-added-serve-stale-feature-implementation-of.patch 
From 5ed91481abea382dc486507556e5cdf0f36b796f Mon Sep 17 00:00:00 2001

```

Generally I find it easier to keep the provenance info in the patch file instead of the spec/changelog, as you get the former for free (if you are handling the patch correctly). Although I will admit I think manually listing each patch file - with an arbitrary number to enforce ordering - is totally insane to me. The deb approach of simply using a directory to store all patches with quilt series file makes so much more sense.

Anyway, I agree it would be good to adopt some standard for these kind of things; for example Debian/Ubuntu generally asks uploaders to add [DEP3](https://dep-team.pages.debian.net/deps/dep3/) info to patch file headers. I find most of the DEP3 requirements to be way more than is needed - especially when using a backported patch - but the Source: line can be useful sometimes.

What I typically did at Canonical is, since a bug is required for all (stable release) patches, I prefixed all patch files with the bug number; for example, if I was backporting upstream patch '0001-do_stuff.patch' for launchpad bug 12345, I would name the patch `lp12345-do_stuff.patch`. For multiple patches I created a dir `lp12345` and dropped the numbered patches in (e.g. 0001-do_stuff.patch, 0002-more.patch, etc). That immediately makes clear where to find all the backstory on what and why the patches were added. Since we don't necessarily do that (i.e. we rarely have a open github bug that corresponds to patches) it doesn't exactly help us, currently...maybe in future releases we may move in that direction.

hbeberman (approved):

PR #5813, opened 2023-07-11
Filter implicit run nodes before passing to collapse.

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Filter implicit run nodes before passing to `pkgGraph.CreateCollapsedNode()`. The new delta fetcher will create build nodes that point into the cache (prebuilt nodes). When we parse these prebuilt nodes to see if we can resolve any implicit nodes we match based on `node.RpmPath`. Previously those paths would always point into `out/RPMS`, and cache nodes will never point into that directory. With the new delta nodes some build nodes will now point into the cache. These can now match against remote cache nodes, which are not a valid node to squash implicit provides into.

Filter to only real run nodes and discard cache nodes.

###### Change Log  

- In `replaceNodesWithProvides()` filter out cach nodes before passing them to `CreateCollapsedNode()`.

**NO**

###### Test Methodology

- Local build

### DIFF
--- a/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
+++ b/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
@@ -56,12 +56,12 @@ func InjectMissingImplicitProvides(res *BuildResult, pkgGraph *pkggraph.PkgGraph
 func replaceNodesWithProvides(res *BuildResult, pkgGraph *pkggraph.PkgGraph, provides *pkgjson.PackageVer, nodes []*pkggraph.PkgNode, rpmFileProviding string) (err error) {
 	var parentNode *pkggraph.PkgNode
 
-	// Find a run node that is backed by the same rpm as the one providing the implicit provide.
+	// Find a local run node that is backed by the same rpm as the one providing the implicit provide.
 	// Make this node the parent node for the new implicit provide node.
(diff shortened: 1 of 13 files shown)


toolkit/tools/scheduler/schedulerutils/implicitprovides.go
```diff
@@ -60,7 +60,8 @@ func replaceNodesWithProvides(res *BuildResult, pkgGraph *pkggraph.PkgGraph, pro
 	// Make this node the parent node for the new implicit provide node.
 	// - By making a run node the parent node, it will inherit the identical runtime dependencies of the already setup node.
 	for _, node := range pkgGraph.AllRunNodes() {
-		if rpmFileProviding == node.RpmPath {
+		// We need to filter out any non-run nodes since cached remote nodes are considered "Run" nodes for the graph.
```
PawelWMS, line 63: Does that mean the term "run node" is overloaded and means different things from the point of the view of the graph and the node itself? And will other places calling `AllRunNodes` (and maybe also `AllBuildNodes`) still get the nodes they expect? In other words: I'm curious, if it would make sense to move the filtering into the the `All*Nodes` methods?

PawelWMS (approved):

PR #5816, opened 2023-07-11
Update kata-containers-cc to 0.6.0

dallasd1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Update kata-containers-cc to version 0.6.0. This includes updates to the policy tool, tardev-snapshotter, and configuration to enable SEV-SNP.

###### Change Log  

- Update kata-containers-cc to version 0.6.0.

**NO**

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=399398&view=results


SPECS/kata-containers-cc/kata-containers-cc.spec
```diff
@@ -161,7 +161,9 @@ install -D -m 0755 kata-runtime %{buildroot}%{coco_bin}/kata-runtime
 install -D -m 0755 data/kata-collect-data.sh %{buildroot}%{coco_bin}/kata-collect-data.sh
 
 install -D -m 0644 config/configuration-clh.toml %{buildroot}/%{defaults_kata}/configuration-clh.toml
+install -D -m 0644 config/configuration-clh-snp.toml %{buildroot}/%{defaults_kata}/configuration-clh-snp.toml
```
ms-mahuber, line 164: the vanilla kata package should install its own CLH config, while the CC package should install its own CLH SNP config, right?

dallasd1: The vanilla kata package does not install a "-clh" config. Its config file is named configuration.toml.

SPECS/kata-containers-cc/kata-containers-cc.spec
```diff
@@ -135,6 +134,7 @@ mkdir -p %{buildroot}%{share_kata}
 mkdir -p %{buildroot}%{coco_path}/libexec
 mkdir -p %{buildroot}/etc/systemd/system/containerd.service.d/
 ln -s /usr/bin/cloud-hypervisor               %{buildroot}%{coco_bin}/cloud-hypervisor
```
ms-mahuber, line 136: with decoupling the configs, we can most likely split out things that only kata-containers needs vs things that kata-containers-cc needs: Both packages should only install/uninstall whatever components (binaries, configurations) they need.

SPECS/kata-containers-cc/kata-containers-cc.spec
ms-mahuber: In other words: Why do we need two configurations? :)

dallasd1: This allows us to have the standard kata-cc config along with the SNP config. Vanilla kata actually installs many different configs so it's similar in that sense. If we're never going to use the standard kata-cc config anymore then it can be removed, but having flexibility here might be advantageous for comparing both cases.

SPECS/kata-containers-cc/kata-containers-cc.spec
dallasd1: Are we 100% certain that we will only have one CLH on the host and that it has SNP enabled? If so, we can remove one of these. If not, we may want to keep both symlinks here in case we use the non-snp config, which would point to the non-snp CLH.

dallasd1: Leaving here discussed offline

ms-mahuber (approved):

PR #5825, opened 2023-07-13
Fixed package post-install calls to `ldconfig`.

PawelWMS:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

The `tdnf-cli-libs` package wasn't calling `ldconfig` and it was breaking downgrades of `tdnf`. While working on the fix, I realized that the `%ldconfig*` macros never worked and we were only lucky we didn't hit more issues.

Why the macros didn't work:
All `%ldconfig_*` macros evaluate to an empty line if `%ldconfig` is not defined, which was the case for Mariner due to a commented-out `%ldconfig /sbin/ldconfig` line inside the `SPECS/mariner-rpm-macros/macros` file.

Using [Fedora's solution](https://src.fedoraproject.org/rpms/glibc/blob/rawhide/f/glibc.spec#_512) to this issue as suggested by @oliviacrain. Now the `glibc` package (owner of `ldconfig`) will call it every time another package installs a file into any of the standard library directories.

###### Change Log  

- Added files triggers to `glibc`.

Yes.

###### Test Methodology

- Local package builds.
- Local verification that the `%ldconfig*` macros now work.
- [Full pipeline build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=402496&view=results).


oliviacrain:
So, if these macros never worked in the first place, we have a chance to start fresh and do this the right way. Instead of fixing the macros, why don't we just leave them be and use file triggers in the `glibc` package to call `ldconfig` whenever a new library is installed? [That's what Fedora does](https://src.fedoraproject.org/rpms/glibc/blob/rawhide/f/glibc.spec#_512) and it's a much more scalable solution.

PawelWMS:
> So, if these macros never worked in the first place, we have a chance to start fresh and do this the right way. Instead of fixing the macros, why don't we just leave them be and use file triggers in the `glibc` package to call `ldconfig` whenever a new library is installed? [That's what Fedora does](https://src.fedoraproject.org/rpms/glibc/blob/rawhide/f/glibc.spec#_512) and it's a much more scalable solution.

@oliviacrain, this is week was not wasted, you just helped me learn about this way of doing things and so far I like. Thank you. I'll take a closer look at what Fedora is doing but at the moment it certainly looks a lot more reasonable.

PR #5827, opened 2023-07-13
build nginx with http_gunzip_module

intlabs:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This updates the nginx package, adding the capability to serve gzip compressed static files to clients that cannot decode on the fly. This enables us to significantly reduce the size of static content on disk transparently for consumers. This functionality is the default in many distrobutions.

###### Change Log  

- Enable building with http_gunzip_module

**NO**

###### Associated issues  

- n/a

###### Links to CVEs  
- n/a

###### Test Methodology

- local build, validated `gunzup on;` configuration works as expected.

### DIFF
--- a/SPECS/nginx/nginx.spec
+++ b/SPECS/nginx/nginx.spec
@@ -6,7 +6,7 @@ Name:           nginx
 # Currently on "stable" version of nginx from https://nginx.org/en/download.html.
 # Note: Stable versions are even (1.20), mainline versions are odd (1.21)
 Version:        1.22.1
-Release:        5%{?dist}
+Release:        6%{?dist}
 License:        BSD-2-Clause
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -59,6 +59,7 @@ sh configure \
     --sbin-path=%{_sbindir}/nginx                 \
     --user=%{nginx_user} \
     --with-http_auth_request_module \
+    --with-http_gunzip_module \
     --with-http_gzip_static_module \
     --with-http_realip_module \
     --with-http_ssl_module \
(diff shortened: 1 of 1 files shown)


SPECS/nginx/nginx.spec
```diff
@@ -59,6 +59,7 @@ sh configure \
     --sbin-path=%{_sbindir}/nginx                 \
     --user=%{nginx_user} \
     --with-http_auth_request_module \
+    --with-http_gunzip_module \
```
romoh, line 62: Pardon my ignorance here. 
Does that mean that the nginx web server will always use rely on compression/decompression? or is that configurable? 
Mainly concerned about any performance regressions if any for this change. (since this is a commonly ised package as you may think)

intlabs: no problem :)

It won't always rely on compression/decompression, this simply enables the build of the gunzip module (see: https://nginx.org/en/docs/http/ngx_http_gunzip_module.html)

Here's a super simplistic example of how this module can be used in combination with other modules already built with the current mariner nginx config:
```
http {
       ...
        server {
         ...
         gzip on;
         server_name _;
         location / {
           gzip_static always;
           gunzip on; #<<-- what this updated rpm would support
         }
        }
}
```

Here any static file eg `hello.gz` can be requested by simple clients eg `wget` buy requesting `/hello`, nginx will transparently serve the decompressed contents of hello.gz.

This build configuration is very common amongst distros, eg:
 * https://src.fedoraproject.org/rpms/nginx//blob/rawhide/f/nginx.spec#_316
 * http://hg.nginx.org/pkg-oss/file/tip/rpm/SPECS/Makefile#l77
 
There is no performance impact if the module is not used/specified in the configuration - this is not a breaking change. If the module is used and made use of the CPU and memory consumption does increase over simple serving of static files, though this is offset by significant savings in on disk space sondumed by the content.

romoh: Thanks for the detailed explanation, Pete.

romoh (approved):

oliviacrain:
Kicked off a test build [here](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=393062&view=results)

oliviacrain (approved):
Test build passes, LGTM! Thanks for the patch!

PR #5841, opened 2023-07-18
Add shortest path print of unsolvable nodes when using trace level logging

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Debugging builds with unresolved nodes can be challenging, let us print a shortest path from the node to the goal at trace level logging. 

###### Change Log  

- Add shortest path from each unresolved node to the goal node.

**NO**

###### Test Methodology

- Local Builds

### DIFF
--- a/toolkit/scripts/tools.mk
+++ b/toolkit/scripts/tools.mk
@@ -42,7 +42,8 @@ go_module_files = $(TOOLS_DIR)/go.mod $(TOOLS_DIR)/go.sum
 go_internal_files = $(shell find $(TOOLS_DIR)/internal/ -type f -name '*.go')
 go_pkg_files = $(shell find $(TOOLS_DIR)/pkg/ -type f -name '*.go')
 go_imagegen_files = $(shell find $(TOOLS_DIR)/imagegen/ -type f -name '*.go')
-go_common_files = $(go_module_files) $(go_internal_files) $(go_imagegen_files) $(go_pkg_files) $(STATUS_FLAGS_DIR)/got_go_deps.flag $(BUILD_DIR)/tools/internal.test_coverage
+go_scheduler_files = $(shell find $(TOOLS_DIR)/scheduler -type f -name '*.go')
+go_common_files = $(go_module_files) $(go_internal_files) $(go_imagegen_files) $(go_pkg_files) $(go_scheduler_files) $(STATUS_FLAGS_DIR)/got_go_deps.flag $(BUILD_DIR)/tools/internal.test_coverage
 # A report on test coverage for all the go tools
 test_coverage_report=$(TOOL_BINS_DIR)/test_coverage_report.html
 
--- a/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
+++ b/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
@@ -107,14 +107,18 @@ func fetchPackages(dependencyGraph *pkggraph.PkgGraph, hasUnresolvedNodes, tryDo
 	}
 	defer cloner.Close()
 
+	toolchainPackages, err := schedulerutils.ReadReservedFilesList(*toolchainManifest)
+	if err != nil {
+		err = fmt.Errorf("unable to read toolchain manifest file '%s':\n%w", *toolchainManifest, err)
+		return
+	}
+	toolchainPackageSet := make(map[string]bool)
+	for _, pkg := range toolchainPackages {
+		toolchainPackageSet[pkg] = true
+	}
+
 	if hasUnresolvedNodes {
-		var toolchainPackages []string
 		logger.Log.Info("Found unresolved packages to cache, downloading packages")
-		toolchainPackages, err = schedulerutils.ReadReservedFilesList(*toolchainManifest)
-		if err != nil {
-			err = fmt.Errorf("unable to read toolchain manifest file '%s':\n%w", *toolchainManifest, err)
-			return
-		}
 
 		err = resolveGraphNodes(dependencyGraph, *inputSummaryFile, toolchainPackages, cloner, *stopOnFailure)
 		if err != nil {
@@ -128,7 +132,7 @@ func fetchPackages(dependencyGraph *pkggraph.PkgGraph, hasUnresolvedNodes, tryDo
 	// Optional delta build cache hydration
 	if tryDownloadDeltaRPMs {
 		logger.Log.Info("Attempting to download delta RPMs for build nodes")
-		err = downloadDeltaNodes(dependencyGraph, cloner)
+		err = downloadDeltaNodes(dependencyGraph, cloner, toolchainPackageSet)
 		if err != nil {
 			err = fmt.Errorf("failed to download delta RPMs:\n%w", err)
 			return
@@ -188,7 +192,7 @@ func setupCloner() (cloner *rpmrepocloner.RpmRepoCloner, err error) {
 //   - pkgsToIgnore: The list of packages to ignore
 //   - imageConfig: The image config to use to find the packages we need to build
 //   - baseDirPath: The base directory to use to find the packages we need to build
-func downloadDeltaNodes(dependencyGraph *pkggraph.PkgGraph, cloner *rpmrepocloner.RpmRepoCloner) (err error) {
+func downloadDeltaNodes(dependencyGraph *pkggraph.PkgGraph, cloner *rpmrepocloner.RpmRepoCloner, toolchainPackageSet map[string]bool) (err error) {
 	const (
 		useImplicitForOptimization = true
 	)
@@ -211,7 +215,7 @@ func downloadDeltaNodes(dependencyGraph *pkggraph.PkgGraph, cloner *rpmrepoclone
 		return
 	}
 
-	isGraphOptimized, deltaPkgGraphCopy, _, err := schedulerutils.PrepareGraphForBuild(deltaPkgGraphCopy, packageVersToBuild, useImplicitForOptimization)
+	isGraphOptimized, deltaPkgGraphCopy, _, err := schedulerutils.PrepareGraphForBuild(deltaPkgGraphCopy, packageVersToBuild, useImplicitForOptimization, toolchainPackageSet)
 	if err != nil {
 		err = fmt.Errorf("failed to initialize graph for delta package downloading:\n%w", err)
 		return
--- a/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
+++ b/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
@@ -107,18 +107,14 @@ func fetchPackages(dependencyGraph *pkggraph.PkgGraph, hasUnresolvedNodes, tryDo
 	}
 	defer cloner.Close()
 
-	toolchainPackages, err := schedulerutils.ReadReservedFilesList(*toolchainManifest)
-	if err != nil {
-		err = fmt.Errorf("unable to read toolchain manifest file '%s':\n%w", *toolchainManifest, err)
-		return
-	}
-	toolchainPackageSet := make(map[string]bool)
-	for _, pkg := range toolchainPackages {
-		toolchainPackageSet[pkg] = true
-	}
-
 	if hasUnresolvedNodes {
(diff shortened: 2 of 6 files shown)


toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -107,14 +107,18 @@ func fetchPackages(dependencyGraph *pkggraph.PkgGraph, hasUnresolvedNodes, tryDo
...
 
+	toolchainPackages, err := schedulerutils.ReadReservedFilesList(*toolchainManifest)
+	if err != nil {
+		err = fmt.Errorf("unable to read toolchain manifest file '%s':\n%w", *toolchainManifest, err)
+		return
+	}
+	toolchainPackageSet := make(map[string]bool)
+	for _, pkg := range toolchainPackages {
+		toolchainPackageSet[pkg] = true
+	}
```
PawelWMS, line 118: Let's just make `resolveGraphNodes` accept a set instead. I think using a list there was a mistake to begin with and we'll simplify a bit of the code both here and there this way.

PawelWMS: After thinking about this more: since we want to restore the reverted toolkit changes and they are quite robust, I take that back - the refactoring can wait to minimize merge conflicts.

toolkit/tools/scheduler/schedulerutils/depsolver.go
```diff
@@ -40,6 +48,22 @@ func CanSubGraph(pkgGraph *pkggraph.PkgGraph, node *pkggraph.PkgNode, useCachedI
...
+			} else {
+				path, _ := pt.To(pkgNode.ID())
+				logger.Log.Tracef("Path between %v and %v:", node, pkgNode)
+				for _, n := range path {
+					logger.Log.Warnf("  %v", n.(*pkggraph.PkgNode))
+				}
+			}
+		}
+
 		foundUnsolvableNode = true
```
PawelWMS, line 67: Since we return if `BellmanFordFrom` fails but that's only logging, I think we should set `foundUnsolvableNode` first. Also, I think we can drop `else` due to that `return`:
```suggestion
		foundUnsolvableNode = true

		// If we are in trace mode, print the path from the root node to the unsolvable node
		if logger.Log.IsLevelEnabled(logrus.TraceLevel) {
			pt, ok := path.BellmanFordFrom(node, pkgGraph)
			if !ok {
				logger.Log.Warnf("Could not run BellmanFordFrom on %v", node)
				return
			}

			path, _ := pt.To(pkgNode.ID())
			logger.Log.Tracef("Path between %v and %v:", node, pkgNode)
			for _, n := range path {
				logger.Log.Warnf("  %v", n.(*pkggraph.PkgNode))
			}
		}
```

toolkit/tools/scheduler/schedulerutils/depsolver.go
```diff
@@ -40,6 +48,22 @@ func CanSubGraph(pkgGraph *pkggraph.PkgGraph, node *pkggraph.PkgNode, useCachedI
...
+		if logger.Log.IsLevelEnabled(logrus.TraceLevel) {
+			pt, ok := path.BellmanFordFrom(node, pkgGraph)
+			if !ok {
+				logger.Log.Warnf("Could not run BellmanFordFrom on %v", node)
+				return
+			} else {
+				path, _ := pt.To(pkgNode.ID())
+				logger.Log.Tracef("Path between %v and %v:", node, pkgNode)
+				for _, n := range path {
+					logger.Log.Warnf("  %v", n.(*pkggraph.PkgNode))
```
PawelWMS, line 62: Why log the nodes on the warning level if we run all of this only if we're logging on trace level? I understand the failure above, but this seems like something we expect to log.

toolkit/tools/scheduler/schedulerutils/depsolver.go
```diff
@@ -120,6 +148,12 @@ func FindUnblockedNodesFromResult(res *BuildResult, pkgGraph *pkggraph.PkgGraph,
 	return
 }
 
+func isToolchainPackage(pkgNode *pkggraph.PkgNode, toolchainPackages map[string]bool) bool {
```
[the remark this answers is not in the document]
PawelWMS: Nevermind, I missed the fact that `IsReservedFile` works on a slice. Another thing to refactor later and I see your point here now.

toolkit/tools/scheduler/schedulerutils/depsolver.go
```diff
@@ -76,8 +100,12 @@ func LeafNodes(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMutex, goalNode *
...
 		// a normal node via InjectMissingImplicitProvides().
 		if !useCachedImplicit && pkgNode.Implicit && pkgNode.State == pkggraph.StateCached {
-			logger.Log.Debugf("Skipping cached implicit provide leaf node: %v", pkgNode)
-			return
+			if isToolchainPackage(pkgNode, toolchainPackages) {
+				logger.Log.Debugf("Allowing cached implicit toolchain leaf node: %v", pkgNode)
+			} else {
+				logger.Log.Debugf("Skipping cached implicit provide leaf node: %v", pkgNode)
+				return
+			}
```
PawelWMS, line 108: This will probably print a lot of output. What do you think about making it trace level?

toolkit/tools/scheduler/schedulerutils/depsolver.go
```diff
@@ -76,8 +100,12 @@ func LeafNodes(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMutex, goalNode *
 		// Ideally we will wait for the actual provider of the implicit node to be built and convert the implicit node to
 		// a normal node via InjectMissingImplicitProvides().
 		if !useCachedImplicit && pkgNode.Implicit && pkgNode.State == pkggraph.StateCached {
-			logger.Log.Debugf("Skipping cached implicit provide leaf node: %v", pkgNode)
-			return
+			if isToolchainPackage(pkgNode, toolchainPackages) {
```
PawelWMS, line 103: With this change we started to make a distinction between toolchain and non-toolchain implicit nodes and that distinction forces us to pass around the list of toolchain packages to this and other functions. Wouldn't it make sense to find toolchain nodes once when we build the graph and encode that knowledge in the node itself by creating a new type? Also, some of the checks above like `useCachedImplicit` and `pkgNode.State == pkggraph.StateCached` don't really apply to toolchain packages - they are always available in the cache. If they weren't, that would be a bug somewhere a lot earlier in the code, which might be another indication that toolchain and non-toolchain implicit nodes are different enough to give them separate types.

dmcilvaney:
The core issue was a merge conflict resolution gone bad. Resolved in https://github.com/microsoft/CBL-Mariner/pull/5851.

With that fix the implicit nodes will have been converted to prebuilt nodes in the fetcher, and special handling will not be needed.

toolkit/scripts/tools.mk
```diff
@@ -42,7 +42,8 @@ go_module_files = $(TOOLS_DIR)/go.mod $(TOOLS_DIR)/go.sum
 go_internal_files = $(shell find $(TOOLS_DIR)/internal/ -type f -name '*.go')
 go_pkg_files = $(shell find $(TOOLS_DIR)/pkg/ -type f -name '*.go')
 go_imagegen_files = $(shell find $(TOOLS_DIR)/imagegen/ -type f -name '*.go')
-go_common_files = $(go_module_files) $(go_internal_files) $(go_imagegen_files) $(go_pkg_files) $(STATUS_FLAGS_DIR)/got_go_deps.flag $(BUILD_DIR)/tools/internal.test_coverage
+go_scheduler_files = $(shell find $(TOOLS_DIR)/scheduler -type f -name '*.go')
```
dmcilvaney, line 45: I'd like to fix this regardless.

toolkit/tools/scheduler/schedulerutils/depsolver.go
dmcilvaney: Ah, forgot to commit the switch to trace.

toolkit/tools/scheduler/schedulerutils/depsolver.go
dmcilvaney: That was probably muscle memory tbh, I'm thinking we just ignore the failure and keep going.

toolkit/tools/scheduler/schedulerutils/implicitprovides.go
```diff
@@ -61,16 +61,16 @@ func replaceNodesWithProvides(res *BuildResult, pkgGraph *pkggraph.PkgGraph, pro
 	// - By making a run node the parent node, it will inherit the identical runtime dependencies of the already setup node.
 	for _, node := range pkgGraph.AllRunNodes() {
 		// We need to filter out any non-local run nodes since cached remote nodes are not acceptable parent nodes for collapsed nodes.
-		if node.Type == pkggraph.TypeLocalRun && rpmFileProviding == node.RpmPath {
-			logger.Log.Debugf("Linked implicit provide (%s) to run node (%s)", provides, node.FriendlyName())
+		if (node.Type == pkggraph.TypeLocalRun || node.Type == pkggraph.TypePreBuilt) && rpmFileProviding == node.RpmPath {
```
dmcilvaney, line 64: @PawelWMS thoughts on keeping this? I can't think of a reason we would end up with implicit nodes that link back to toolchain nodes now that the prebuilt stuff is fixed.

toolkit/tools/scheduler/schedulerutils/depsolver.go
```diff
@@ -40,6 +42,21 @@ func CanSubGraph(pkgGraph *pkggraph.PkgGraph, node *pkggraph.PkgNode, useCachedI
 
 		// This node is not yet solvable
 		logger.Log.Warnf("Could not subgraph due to node: %v", pkgNode)
+
+		// If we are in trace mode, print the path from the root node to the unsolvable node
+		if logger.Log.IsLevelEnabled(logrus.TraceLevel) {
+			pt, ok := path.BellmanFordFrom(node, pkgGraph)
```
trungams, line 48: Out of curiosity but would Bellman-Ford help us here. Since the build graph is unweighted, BFS works, or if there's no support for finding path via BFS then I think Dijkstra has similar effect/complexity on this type of graph.

toolkit/tools/scheduler/schedulerutils/implicitprovides.go
PawelWMS: I also can't think of a reason for that but I'm not too familiar with the logic in this function. I think we can leave the original version for now but my reason would be to not introduce more changes until we're sure the builds have stabilized a bit.

toolkit/tools/scheduler/schedulerutils/depsolver.go
dmcilvaney: tbh it was the first agorithm that had example code in the package, I have no particular interest in the specific algorithm. I'll have a quick look to see if there is a dfs option to generate the path data but I didn't recall seeing one when I looked.

dmcilvaney: Found a single shot shortest path alg instead of using the full BF, switched to that.

trungams: hmm I think Yen's algorithm uses Dijkstra internally, so we can just use dijkstra to simplify it. It's also provided by gonum I think. Should help if our build graph may have up to 100k edges

PawelWMS (approved):

toolkit/tools/scheduler/schedulerutils/depsolver.go
dmcilvaney: API wise the call to the Yen's function is simpler, we don't need to do the two stage process. We just get a list back. Since this is trace level I'm really not concerned about performance here. If we do this for real at debug/info level we can optimize by caching the shortest path data maybe, but I'm satisfied with yens(k=1) which simplifies down to a simple shortest path.

trungams: sounds good to me!

PR #5863, opened 2023-07-21
Resolved cyclic ptest dependencies

PawelWMS:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Updating the way we handle `RUN_CHECK=y` builds. Some specs define additional `BuildRequires` for test runs and these requirements create unresolvable cyclic dependencies. In order to resolve this issue for offline builds, this PR splits the packages' build nodes into a test and a non-test build nodes. This will allow the non-test nodes to build without issues first as non-test builds are required by our policy to have no circular dependencies except for the toolchain packages. The built non-test nodes will then unblock the test ones to build again with their extra dependencies.

# Sample build output
Details:
- All packages pre-built.
- `SRPM_PACK_LIST="python-py pytest python-wheel python-setuptools_scm"`
- `PACKAGE_BUILD_LIST="python-py pytest python-wheel python-setuptools_scm"`
- `PACKAGE_REBUILD_LIST=`
- Deleted one subpackages of `python-wheel`.
- The `python-setuptools_scm` package doesn't require `python-wheel` to **build** but requires it for the ptests.

image
 

###### Change Log  

- Ptests are now represented by a new `TypeTest` node. The `TypeLocalBuild` nodes are reserved for "regular" builds now.
- Added 3 new `TEST_*_LIST` build arguments analogous to the existing set of `PACKAGE_BUILD_LIST`, `PACKAGE_REBUILD_LIST`, and `PACKAGE_IGNORE_LIST`.
- Test node differences from a build node:
   - Only added if `RUN_CHECK=y` **and** the spec file contains the `%check` section.
   - The test build skips creation of the RPM files. The `%prep`, `%build`, `%install`, and `%check` sections are still executed.
   - The test node has a virtual dependency on the build node. This forces the scheduler to first consider the build node as a build candidate. At that point in time we establish if the build node needs to be rebuilt or we can declare it as a (delta) pre-built node. This information is used to establish is the test node needs to run or can be skipped.
   - The virtual dependency is removed once the build node has been analyzed and this makes it possible to queue the test and the build in parallel, assuming that the build node was the last remaining dependency of the test node.
   - Nothing ever depends on a test node.
- The build agents are no longer pre-configured to run the tests or not and instead accept a new argument when asked to build an SRPM.
- The scheduler no longer needs the `run-check` flag. That's now encoded in the existence of the test nodes.
- Replaced `IsSRPMPrebuilt` with `FindRPMFiles`. The function no longer returns the `isPrebuilt` boolean as it was redundant when the `missingFiles` output provided the same information and more.
- Moved literals such as `<NO_SRPM_PATH>` into constants.
- `rpm.go`:
   - Added `SpecHasCheckSection` to check if a spec constaing the `%check` section.
   - Added `TestRPMFromSRPM` to run a test build of an SRPM by adding the `-ri` flag to `rpmbuild`, which causes the tool to stop before it tries to create the RPM files.

No.

###### Test Methodology

- [Full pipeline build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=409101&view=results).


toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -83,8 +77,7 @@ func main() {
 	srpmsDirAbsPath, err := filepath.Abs(*srpmsDirPath)
 	logger.PanicOnError(err, "Unable to find absolute path for SRPMs directory '%s'", *srpmsDirPath)
 
-	srpmName := strings.TrimSuffix(filepath.Base(*srpmFile), ".src.rpm")
-	chrootDir := filepath.Join(*workDir, srpmName)
+	chrootDir := buildChrootDirPath(*workDir, *srpmFile, *runCheck)
```
PawelWMS, line 80: Need a separate chroot for parallel "regular" and test builds.

toolkit/tools/scheduler/scheduler.go
```diff
@@ -69,7 +69,6 @@ var (
 	rpmmacrosFile              = app.Flag("rpmmacros-file", "Optional file path to an rpmmacros file for rpmbuild to use.").ExistingFile()
 	buildAttempts              = app.Flag("build-attempts", "Sets the number of times to try building a package.").Default(defaultBuildAttempts).Int()
 	checkAttempts              = app.Flag("check-attempts", "Sets the minimum number of times to test a package if the tests fail.").Default(defaultCheckAttempts).Int()
-	runCheck                   = app.Flag("run-check", "Run the check during package builds.").Bool()
```
PawelWMS, line 72: No longer needed - presence of test nodes indicates we're running checks.

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -317,45 +343,35 @@ func (g *PkgGraph) addToLookup(pkgNode *PkgNode, deferSort bool) (err error) {
...
 	}
 
 	switch pkgNode.Type {
 	case TypeLocalBuild:
-		if existingLookup.BuildNode == nil {
-			existingLookup.BuildNode = pkgNode.This
-		} else {
-			err = duplicateError
-			return
-		}
```
PawelWMS, line 345: We're already running `validateNodeForLookup`, so this was redundant.

toolkit/scripts/pkggen.mk
```diff
@@ -75,7 +75,7 @@ analyze-built-graph: $(go-graphanalytics)
 	fi
 
 # Parse all specs in $(BUILD_SPECS_DIR) and generate a specs.json file encoding all dependency information
-$(specs_file): $(chroot_worker) $(BUILD_SPECS_DIR) $(build_specs) $(build_spec_dirs) $(go-specreader)
+$(specs_file): $(chroot_worker) $(BUILD_SPECS_DIR) $(build_specs) $(build_spec_dirs) $(go-specreader) $(depend_RUN_CHECK)
```
dmcilvaney, line 78: Do we care about this if it doesn't affect the behavior of the specreader?
```suggestion
$(specs_file): $(chroot_worker) $(BUILD_SPECS_DIR) $(build_specs) $(build_spec_dirs) $(go-specreader)
```

toolkit/tools/grapher/grapher.go
```diff
@@ -110,47 +110,80 @@ func addUnresolvedPackage(g *pkggraph.PkgGraph, pkgVer *pkgjson.PackageVer) (new
...
+		if nodes.TestNode != nil {
+			newTestNode = nodes.TestNode
+		}
 	}
 
 	if newRunNode == nil {
 		// Add "Run" node
-		newRunNode, err = g.AddPkgNode(pkgVer, pkggraph.StateMeta, pkggraph.TypeLocalRun, pkg.SrpmPath, pkg.RpmPath, pkg.SpecPath, pkg.SourceDir, pkg.Architecture, "<LOCAL>")
-		logger.Log.Debugf("Adding run node %s with id %d\n", newRunNode.FriendlyName(), newRunNode.ID())
+		newRunNode, err = g.AddPkgNode(pkg.Provides, pkggraph.StateMeta, pkggraph.TypeLocalRun, pkg.SrpmPath, pkg.RpmPath, pkg.SpecPath, pkg.SourceDir, pkg.Architecture, "<LOCAL>")
```
dmcilvaney, line 143: Constant for `"<LOCAL>"`?

toolkit/tools/grapher/grapher.go
```diff
@@ -110,47 +110,80 @@ func addUnresolvedPackage(g *pkggraph.PkgGraph, pkgVer *pkgjson.PackageVer) (new
...
 			return
 		}
+		logger.Log.Debugf("Adding run node %s with id %d\n", newRunNode.FriendlyName(), newRunNode.ID())
 	}
 
 	if newBuildNode == nil {
 		// Add "Build" node
-		newBuildNode, err = g.AddPkgNode(pkgVer, pkggraph.StateBuild, pkggraph.TypeLocalBuild, pkg.SrpmPath, pkg.RpmPath, pkg.SpecPath, pkg.SourceDir, pkg.Architecture, "<LOCAL>")
-		logger.Log.Debugf("Adding build node %s with id %d\n", newBuildNode.FriendlyName(), newBuildNode.ID())
+		newBuildNode, err = g.AddPkgNode(pkg.Provides, pkggraph.StateBuild, pkggraph.TypeLocalBuild, pkg.SrpmPath, pkg.RpmPath, pkg.SpecPath, pkg.SourceDir, pkg.Architecture, "<LOCAL>")
```
dmcilvaney, line 152: Constant for `"<LOCAL>"`?

toolkit/tools/grapher/grapher.go
```diff
@@ -110,47 +110,80 @@ func addUnresolvedPackage(g *pkggraph.PkgGraph, pkgVer *pkgjson.PackageVer) (new
...
-	// A "run" node has an implicit dependency on its coresponding "build" node, encode that here.
+	// A "run" node has an implicit dependency on its corresponding "build" node, encode that here.
 	err = g.AddEdge(newRunNode, newBuildNode)
 	if err != nil {
-		logger.Log.Errorf("Adding edge failed for %+v", pkgVer)
+		logger.Log.Errorf("Adding run -> build edge failed for %+v", pkg.Provides)
+		return
+	}
+
+	if !pkg.RunTests {
```
dmcilvaney, line 166: Do we want to always inject test nodes into the graph? I'm not sure either way... If we always inject them we can re-use the graph artifacts for a build with check on/off, but the graph files also get larger... and it makes the scheduler simpler if the nodes just don't exist.

toolkit/scripts/pkggen.mk
dmcilvaney: Actually, probably better to continue to pass the below `--run-check` through and keep `$(depend_RUN_CHECK)`

toolkit/scripts/pkggen.mk
```diff
@@ -257,7 +257,6 @@ $(STATUS_FLAGS_DIR)/build-rpms.flag: $(preprocessed_file) $(chroot_worker) $(go-
 		--timestamp-file=$(TIMESTAMP_DIR)/scheduler.jsonl \
 		--toolchain-manifest="$(TOOLCHAIN_MANIFEST)" \
 		$(if $(CONFIG_FILE),--base-dir="$(CONFIG_BASE_DIR)") \
-		$(if $(filter y,$(RUN_CHECK)),--run-check) \
```
dmcilvaney, line 260: I think this should stay?

toolkit/tools/specreader/specreader.go
```diff
@@ -315,82 +315,82 @@ func readSpecWorker(requests <-chan string, results chan<- *parseResult, cancel
...
-			result.err = err
-			results <- result
+			sendEmptyResult(results, err)
 			continue
 		}
 
 		srpmPath := filepath.Join(srpmsDir, srpmResults[0])
 
-		isCompatible, err := rpm.SpecArchIsCompatible(specfile, sourcedir, arch, defines)
+		isCompatible, err := rpm.SpecArchIsCompatible(specFile, sourceDir, arch, noCheckDefines)
```
dmcilvaney, line 333: Do you think there is ever a chance a check section is arch dependent?

toolkit/tools/scheduler/buildagents/chrootagent.go
```diff
@@ -36,7 +36,7 @@ func (c *ChrootAgent) Initialize(config *BuildAgentConfig) (err error) {
 // - logName is the file name to save the package build log to.
 // - outArch is the target architecture to build for.
```
dmcilvaney, line 37:
```suggestion
// - outArch is the target architecture to build for.
// - runCheck indicates if the package should be built, or just tested
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -48,7 +49,7 @@ type BuildResult struct {
 	Err            error
 	LogFile        string
 	Node           *pkggraph.PkgNode
-	Skipped        bool
+	Ignored        bool
 	UsedCache      bool
 	WasDelta       bool
 }
```
dmcilvaney, line 55: I've gotten confused by some of these, was thining maybe we should add a comment for each member of the req/res objects?
```suggestion
	Node           *pkggraph.PkgNode // Main node that was built
	Ignored        bool              // ???
	UsedCache      bool              // Used a cached component (expand on this later?)
	WasDelta       bool              // Used a pre-built published component
}
```

toolkit/tools/scheduler/schedulerutils/preparerequest.go
```diff
@@ -23,51 +23,149 @@ func ConvertNodesToRequests(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMute
...
+	if !isNodeUnblocked(pkgGraph, buildState, testNode) {
+		return
+	}
+
+	return &BuildRequest{
+		Node:           testNode,
+		PkgGraph:       pkgGraph,
+		AncillaryNodes: ancillaryTestNodes,
+		IsDelta:        isDelta,
+		UseCache:       buildUsesCache && canUseCacheForNode(pkgGraph, testNode, packagesToRebuild, buildState),
```
dmcilvaney, line 146: Do we need to grab the value of `canUseCacheForNode()` prior to deleting the edges? That function checks rebuliding dependencies based on those edges, and I think the goal here is to ensure we re-run the tests in the even the build node is triggered to rebuild?

toolkit/scripts/pkggen.mk
PawelWMS: Ah, crap! I wanted to remove this from the scheduler, not the reader.:/ Thanks for catching this!

PawelWMS: It's not used anymore.

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -35,30 +35,42 @@ type NodeState int
...
+
+// Constants representing empty member fields of the 'PkgNode' struct.
+const (
+	NoArchitecture = "<NO_ARCHITECTURE>"
+	NoName         = "<NO_NAME>"
+	NoRPMPath      = "<NO_RPM_PATH>"
+	NoSourceDir    = "<NO_SOURCE_DIR>"
+	NoSourceRepo   = "<NO_SOURCE_REPO>"
+	NoSpecPath     = "<NO_SPEC_PATH>"
+	NoSRPMPath     = "<NO_SRPM_PATH>"
```
dmcilvaney, line 73:
```suggestion
	NoSRPMPath     = "<NO_SRPM_PATH>"
    FromLocalRepo  = "<LOCAL>"
```

toolkit/tools/grapher/grapher.go
dmcilvaney:
```suggestion
		newRunNode, err = g.AddPkgNode(pkg.Provides, pkggraph.StateMeta, pkggraph.TypeLocalRun, pkg.SrpmPath, pkg.RpmPath, pkg.SpecPath, pkg.SourceDir, pkg.Architecture, pkg.FromLocalRepo)
```

toolkit/tools/grapher/grapher.go
dmcilvaney:
```suggestion
		newBuildNode, err = g.AddPkgNode(pkg.Provides, pkggraph.StateBuild, pkggraph.TypeLocalBuild, pkg.SrpmPath, pkg.RpmPath, pkg.SpecPath, pkg.SourceDir, pkg.Architecture, pkg.FromLocalRepo)
```

toolkit/tools/scheduler/schedulerutils/preparerequest.go
PawelWMS: This is tricky, but we need to call `canUseCacheForNode()` after removing that edge. Since the partner build node is not built yet, then this edge would make `canUseCacheForNode()` always return `false`. At this point the partner build node is being sent to the build worker, so it won't get analyzed again in the future.

In other word the process for the test nodes is:
- Hold it's analysis with the virtual B -> T edge.
- Once the build node is unblocked, analyze its partner test node at the same time. That's this function here. We remove the virtual edge and the test node either gets immediately queued or is blocked on some extra dependencies - that's the part decided by `canUseCacheForNode()`.
- If the test node ends up being blocked, it gets re-analyzed later once these other dependencies are done. These unblocked nodes end up inside `testNodes` from `ConvertNodesToRequests()` and are queued for building in the separate `testNodesToRequests()` function. At this point the partner build nodes for these test nodes have either already finished building or are being built, thus the check for active and cached SRPMs inside that function.

toolkit/tools/specreader/specreader.go
PawelWMS: Good point, but I think this should be treated as a spec misconfiguration. If a package builds for a specific architecture, then it should also be testable for it. Right now, the only exception that comes to my mind, is that the test dependencies might not be available for a specific architecture. In such cases I'd lean toward skipping the `%check` section instead of meddling with the `BuildArch` or `ExclusiveArch` tags, since they are related to building the package itself and the author of the spec shouldn't use them to manipulate tests.

toolkit/tools/grapher/grapher.go
PawelWMS: I think I'd continue with what we've been doing so far. I think the graph didn't contain any test-only information for `RUN_CHECK=n`. Or at least we didn't include the test-only BRs and that's the concept I relied on while adding this change. In other words: for regular builds, I'm currently favour a graph, which contains only the nodes relevant for the build we're running right now. Otherwise, I think things may get unnecessarily confusing when we debug build issues.

toolkit/tools/scheduler/schedulerutils/preparerequest.go
dmcilvaney: Ok, I think I follow. Can we add something similar to the code as a comment for future me, who will also undoubtedly become confused?

toolkit/tools/grapher/grapher.go
```diff
@@ -267,10 +302,8 @@ func populateGraph(graph *pkggraph.PkgGraph, repo *pkgjson.PackageRepo) (err err
 
 	// Scan and add each package we know about
 	logger.Log.Infof("Adding all packages from %s", *input)
-	// NOTE: range iterates by value, not reference. Manually access slice
-	for idx := range packages {
-		pkg := packages[idx]
-		err = addLocalPackage(graph, pkg)
+	for _, pkg := range packages {
```
dmcilvaney, line 305: I don't recall exactly what was going wrong, but if I recall right it ended up being that having two different versions of the `pkg` object floating around with different pointers ended up causing a bunch of headaches (I think it because we want to operate on the original copies of the packages in the repo object, not clones of them?)

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -69,6 +82,13 @@ const (
 	dotKeyFill         = "style"
 )
 
+var lookupNodesTypes = map[NodeType]bool{
```
dmcilvaney, line 85:
```suggestion
// Determines if a type of node is valid for inclusion in the lookup tables.
var lookupNodesTypes = map[NodeType]bool{
```

toolkit/tools/grapher/grapher.go
PawelWMS: I'm not sure I follow. The `pkg` variable is a pointer regardless of the way we initiate it here - we're still operating on the same object. As far as I could, I also believe that both ways of initialization are equivalent in terms of result - no cloning is happening anywhere.

toolkit/tools/scheduler/schedulerutils/preparerequest.go
PawelWMS: Very reasonable request, I should have added it anyway. Thanks!

toolkit/tools/internal/pkggraph/pkggraph.go
```diff
@@ -83,6 +103,7 @@ type PkgNode struct {
 	SourceRepo   string              // The location this package was acquired from
 	GoalName     string              // Optional string for goal nodes
 	Implicit     bool                // If the package is an implicit provide
+	KeepChroot   bool                // If the chroot should be kept after building the package
```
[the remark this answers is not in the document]
PawelWMS: No, I thought I removed it. Thanks for catching that!

toolkit/tools/scheduler/buildagents/testagent.go
```diff
@@ -28,7 +28,7 @@ func (t *TestAgent) Initialize(config *BuildAgentConfig) (err error) {
 }
 
 // BuildPackage simply sleeps and then returns success for TestAgent.
-func (t *TestAgent) BuildPackage(inputFile, logName, outArch string, dependencies []string) (builtFiles []string, logFile string, err error) {
+func (t *TestAgent) BuildPackage(inputFile, logName, outArch string, runCheck bool, dependencies []string) (builtFiles []string, logFile string, err error) {
```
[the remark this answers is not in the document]
PawelWMS: This class is used for the sake of the tests to mock the actual agent. Many of these arguments are going to end up being unused. At some point we might consider using them to print more details about the test's execution or validate expected inputs, but for this PR I'll keep it simple.

toolkit/tools/grapher/grapher.go
```diff
@@ -110,47 +110,80 @@ func addUnresolvedPackage(g *pkggraph.PkgGraph, pkgVer *pkgjson.PackageVer) (new
...
+	nodes, err := g.FindExactPkgNodeFromPkg(pkg.Provides)
 	if err != nil {
 		return
 	}
+
 	if nodes != nil {
-		logger.Log.Warnf(`Duplicate package name for package %+v read from SRPM "%s" (Previous: %+v)`, pkgVer, pkg.SrpmPath, nodes.RunNode)
-		err = nil
+		logger.Log.Warnf(`Duplicate package name for package %+v read from SRPM "%s" (Previous: %+v)`, pkg.Provides, pkg.SrpmPath, nodes.RunNode)
 		if nodes.RunNode != nil {
```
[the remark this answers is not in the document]
PawelWMS: Right, nice catch.

And I believe hitting this part of the code should actually be an error. I didn't want to include more potentially breaking changes into this PR, though.

toolkit/tools/scheduler/buildagents/testagent.go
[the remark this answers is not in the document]
cwize1: Oh, sorry. I should have noticed that.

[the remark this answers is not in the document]
PawelWMS: No problem.;)

jslobodzian (approved):

toolkit/tools/grapher/grapher.go
dmcilvaney: Maybe this is from before we had everything defined as pointers... seems ok looking again.

dmcilvaney (approved):

dmcilvaney (approved):

dmcilvaney (approved):

PR #5864, opened 2023-07-21
Update delta paths prior to implicit handling

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Currently we fail to link `font(fontawesome)` to its associated delta node. The issue is that a delta node will point into the cache, but a newly build node will instead point into `out/RPMS`. The halding for linking an implicit node expects an exact match for the rpm path. Fix the delta path for nodes first, before trying to link implicit nodes. Since we won't have fixed the implict nodes yet, we will also need to allow the delta path update code to skip over implicit nodes (since they will be unexpectedly pointing into the cache rather than `out/RPMS`).

This command should now work: `sudo make build-packages CONFIG_FILE="" PACKAGE_REBUILD_LIST="fontawesome-fonts" PACKAGE_BUILD_LIST="python3-sphinx_rtd_theme" QUICK_REBUILD=y`

###### Change Log  

- Reorder delta node updates to be before implicit node updates when processing a build result
- Allow delta updates to ignore implicit nodes that will point into the cache

**NO**

###### Test Methodology

- Local builds

### DIFF
--- a/toolkit/tools/scheduler/scheduler.go
+++ b/toolkit/tools/scheduler/scheduler.go
@@ -351,9 +351,21 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild []*pkgjson
 
 		if !stopBuilding {
 			if res.Err == nil {
+				if res.Node.Type == pkggraph.TypeLocalBuild && res.WasDelta {
+					logger.Log.Tracef("This is a delta result, update the graph with the new delta files for '%v'.", res.Node)
+					// We will need to update the graph with paths to any delta files that were actually rebuilt.
+					err = setAssociatedDeltaPaths(res, pkgGraph, graphMutex)
+					if err != nil {
+						// Failures to manipulate the graph are fatal. The ancillary delta nodes may be in an invalid state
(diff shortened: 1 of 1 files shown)


toolkit/tools/scheduler/scheduler.go
```diff
@@ -351,9 +351,21 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild []*pkgjson
 
 		if !stopBuilding {
 			if res.Err == nil {
```
PawelWMS, line 353: I think it's time to extract this logic to a separate function, so we can return on errors instead of setting `stopBuilding` and then having to check the variable later. The caller of the new function can set `stopBuilding` if `err != nil && stopOnFailure`.

PawelWMS (approved):

toolkit/tools/scheduler/scheduler.go
dmcilvaney: Agreed, I'll make a note to look at a refactor here.

PR #5867, opened 2023-07-24
Image customizer prototype

cwize1:
This is a proof of concept for the image customizer proposal. It mainly meant to show integrating with the existing utility libraries and successfully modifying an image.

### DIFF
--- a/toolkit/tools/imagecustomizer/api/systemconfig.go
+++ b/toolkit/tools/imagecustomizer/api/systemconfig.go
@@ -7,23 +7,42 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+type FileConfigList []FileConfig
+
 // SystemConfig defines how each system present on the image is supposed to be configured.
 type SystemConfig struct {
-	AdditionalFiles map[string]FileConfig `yaml:"AdditionalFiles"`
+	AdditionalFiles map[string]FileConfigList `yaml:"AdditionalFiles"`
 }
 
 // Configuration of files that are to be added to the image.
 type FileConfig struct {
-	SourceFile      string `yaml:"SourceFile"`
+	DestinationFile string `yaml:"DestinationFile"`
 	FilePermissions string `yaml:"FilePermissions"`
 }
 
-// Allow FileConfig to be parsed from just a string containing the source file.
+// Allow FileConfigList to be parsed from just a single value.
+func (l *FileConfigList) UnmarshalYAML(value *yaml.Node) error {
+	// Try to parse the value as a single value.
+	var fileConfig FileConfig
+	if err := value.Decode(&fileConfig); err == nil {
+		*l = append(*l, fileConfig)
+		return nil
+	}
+
+	// Try to parse the value as a list.
+	if err := value.Decode((*[]FileConfig)(l)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Allow FileConfig to be parsed from just a string containing the destination file.
 func (f *FileConfig) UnmarshalYAML(value *yaml.Node) error {
 	// Try to parse the value as a string.
-	var sourceFile string
-	if err := value.Decode(&sourceFile); err == nil {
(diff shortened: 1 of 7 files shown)


toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -229,7 +243,21 @@ func (c *Chroot) Initialize(tarPath string, extraDirectories []string, extraMoun
...
-		allMountPoints := append(defaultMountPoints(), extraMountPoints...)
+		allMountPoints := []*MountPoint{}
+
+		for _, mountPoint := range extraMountPoints {
+			if mountPoint.mountBeforeDefaults {
+				allMountPoints = append(allMountPoints, mountPoint)
+			}
+		}
+
+		allMountPoints = append(allMountPoints, defaultMountPoints()...)
```
cwize1, line 254: When building an image from scratch, the rootfs is just a normal directory. But when modifying and existing image, the rootfs has to be mounted before all the other partitions. Hence this change.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -262,7 +290,13 @@ func (c *Chroot) AddFiles(filesToCopy ...FileToCopy) (err error) {
 	for _, f := range filesToCopy {
 		dest := filepath.Join(c.rootDir, f.Dest)
 		logger.Log.Debugf("Copying '%s' to worker '%s'", f.Src, dest)
-		err = file.Copy(f.Src, dest)
+
+		if f.FileMode != nil {
```
cwize1, line 294: Setting the file permissions of AdditionalFiles is a missing feature of the Mariner Toolkit.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -533,7 +567,7 @@ func (c *Chroot) createMountPoints(allMountPoints []*MountPoint) (err error) {
 
 		err = unix.Mount(mountPoint.source, fullPath, mountPoint.fstype, mountPoint.flags, mountPoint.data)
 		if err != nil {
-			logger.Log.Errorf("Mount failed on (%s). Error: %s", fullPath, err)
+			logger.Log.Errorf("Mount of (%s) to (%s) failed. Error: %s", mountPoint.source, fullPath, err)
```
cwize1, line 570: Just improving debuggability.

toolkit/tools/imagecustomizerlib/imagecustomizer.go
```diff
@@ -0,0 +1,231 @@
+// Copyright (c) Microsoft Corporation.
```
cwize1, line 1: Having a clean separation between the CLI and the actual business logic is goodness.

toolkit/tools/imagecustomizer/api/systemconfig.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import (
+	"gopkg.in/yaml.v3"
```
cwize1, line 7: YAML is a superset of JSON. So, JSON files will also be accepted.

toolkit/tools/imagecustomizer/api/systemconfig.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// Configuration of files that are to be added to the image.
+type FileConfig struct {
+	DestinationFile string `yaml:"DestinationFile"`
+	FilePermissions string `yaml:"FilePermissions"`
+}
+
+// Allow FileConfigList to be parsed from just a single value.
+func (l *FileConfigList) UnmarshalYAML(value *yaml.Node) error {
```
cwize1, line 24: This is kind of cool in that I can maintain backwards-compatibility with the existing `AdditionalFiles` schema while adding support for both setting the file permissions and for a single source file being copied to 2 different locations.

PR #5459, opened 2023-05-05
Use tools, bpftool, and python3-perf subpackages from kernel.spec for all kernels

rlmenge:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 

rework of #5131, #5305

There was a bug where certain specialized kernels (`kernel-azure`, `kernel-hci`) would create subpackages of the same name (`bpftool`, `python3-perf`) as the generic kernel. This will cause packages to be published and overwritten. However, these packages along with the tools subpackage are not dependent on the configs and can therefore be produced from only the kernel.spec.

Notes:
- photon and [fedora](https://src.fedoraproject.org/rpms/kernel/blob/rawhide/f/kernel.spec#_1059) both have the generic kernel produce the tools and bpftool packages
- subpackages should be installed using `sudo tdnf install -y kernel-tools-$(uname -r)` (as stated in https://microsoft.visualstudio.com/OS/_workitems/edit/44344054) which should account for version issues

###### Change Log  

- Use tools, bpftool, and python3-perf subpackages from kernel.spec for all kernels
 
###### Does this affect the toolchain?  

**Yes**

###### Associated issues  

[...]

PR #5897, opened 2023-07-28
Add kernel conflicts and produce specialized versions of bpftool and python3-perf

rlmenge:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

rework of #5131, #5305, #5459

There was a [bug ](https://microsoft.visualstudio.com/OS/_workitems/edit/43814917/?triage=true)where certain specialized kernels (`kernel-azure`, `kernel-hci`) would create subpackages of the same name (`bpftool`, `python3-perf`) as the generic kernel. This will cause packages to be published and overwritten. To prevent this, create specialized packages. The packages can be accessed using the following names

|kernel | bpftool | python3-perf|
|-----|----|-------|
|kernel| bpftool | python3-perf |
|kernel-azure| kernel-azure-bpftool, bpftool-azure| kernel-azure-python3-perf, python3-perf-azure |
|kernel-hci| kernel-hci-bpftool, bpftool-hci | kernel-hci-python3-perf, python3-perf-hci|
|kernel-rt| kernel-rt-bpftool, bpftool-rt| kernel-rt-python3-perf, python3-perf-rt|

This will affect building and installing of kernels. The following cases for example for kernel-azure and bpftool should behave in the following way in regards to how the bpftool package is named. 

This change would cause the following tasks to behave as follows when certain package names are used
|task| kernel used | "bpftool" | "kernel-azure-bpftool"| "bpftool-azure" |
|-----|------|----|-------|-----|
|image build w/ package| kernel| Success| Error|Error| 
|tdnf install| kernel| success|Error|Error|
|image build w/ package| kernel-azure| Error| Success|Success| 
|tdnf install| kernel-azure| Error| Success|Success| 

On systems which already have "kernel" installed and "bpftool"
|task| result| steps|
|-----|----|-------|
|upgrade bpftool| Success | n/a|
|upgrade kernel| Success| n/a|
|install kernel-azure | Error| conflicts requires uninstalling kernel|

On systems which already have "kernel-azure" installed and "bpftool"
|task| result | steps|
|-----|----|-------|
|upgrade bpftool| Error| requires using bpftool-azure|
|upgrade kernel-azure | Success| n/a - old bpftool does not require kernel-azure|
|install kernel | Error| conflicts requires uninstalling kernel-azure|

Notes:
-  these packages along with the tools subpackage may not be dependent on the configs and can therefore be produced from only the kernel.spec. This is an investigation that is happening [here](https://microsoft.visualstudio.com/OS/_workitems/edit/44120370). This would reduce overhead significantly. Until then, lean on the safer side.

- subpackages should be installed using `sudo tdnf install -y kernel-tools-$(uname -r)` (as stated in https://microsoft.visualstudio.com/OS/_workitems/edit/44344054) which should account for version issues

###### Change Log  

- Add new requires for bpftool and python3-perf for specfic kernel
- Add kernel conflicts

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/43814917/?triage=true

###### Test Methodology

Built packages and image locally.
Validated above use cases in hyperv vm
Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=399534&view=results (both ARM and AMD images succeeded)
After rebase: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=400737&view=results

### DIFF
--- a/SPECS-EXTENDED/kernel-rt/kernel-rt.spec
+++ b/SPECS-EXTENDED/kernel-rt/kernel-rt.spec
@@ -3,6 +3,7 @@
 %global rt_version rt48
 %define uname_r %{version}-%{rt_version}-%{release}
 %define version_upstream %(echo %{version} | rev | cut -d'.' -f2- | rev)
+%define short_name rt
 
 # find_debuginfo.sh arguments are set by default in rpm's macros.
 # The default arguments regenerate the build-id for vmlinux in the 
@@ -136,13 +137,13 @@ Requires:       audit
 %description tools
 This package contains the 'perf' performance analysis tools for Linux kernel.
 
-%package -n     python3-perf-rt
+%package        python3-perf
 Summary:        Python 3 extension for perf tools
 Requires:       python3
 Requires:       %{name} = %{version}-%{release}
-Provides:       python3-perf
+Provides:       python3-perf-%{short_name}
 
-%description -n python3-perf-rt
+%description    python3-perf
 This package contains the Python 3 extension for the 'perf' performance analysis tools for Linux kernel.
 
 %package dtb
@@ -152,12 +153,12 @@ Group:          System Environment/Kernel
 %description dtb
 This package contains common device tree blobs (dtb)
 
-%package -n     bpftool-rt
+%package        bpftool
 Summary:        Inspection and simple manipulation of eBPF programs and maps
 Requires:       %{name} = %{version}-%{release}
-Provides:       bpftool
+Provides:       bpftool-%{short_name}
 
-%description -n  bpftool-rt
+%description    bpftool
 This package contains the bpftool, which allows inspection and simple
 manipulation of eBPF programs and maps.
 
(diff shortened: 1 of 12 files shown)


vinceaperri:
Have you considered what would happen if tuned were installed on a machine with kernel-hci? Since it [requires](https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS/tuned/tuned.spec#L37) python3-perf.

rlmenge:
> Have you considered what would happen if an installation of tuned were attempted on a machine with a specialized kernel such as kernel-rt? Since tuned [requires](https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS/tuned/tuned.spec#L37) python3-perf, which requires kernel, this may fail due to a conflict between kernel and any specialized kernel currently installed.

This is defintely an issue. I will look into how to update tuned.spec for this case.

ddstreetmicrosoft:
I *highly* recommend using separate kernel version strings for each kernel variation, instead of the same version for multiple different builds that have different configs (and maybe different sources?).

For example, you should be able to tell exactly what kernel someone is running just by the output from 'uname -r'.

That also solves the problem of package files colliding/conflicting.

rlmenge:
Closing as current Mariner 2.0 toolkit does not support the ability to choose which kernel to build and instead grabs a random one and fails if it chose wrong

PR #5907, opened 2023-07-31
Extend `AdditionalFiles` config

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary

This change extends the `AdditionalFiles` field of the image system config to support copying to multiple destination files and changing the file permissions of the destination files.

###### Change Log

- Change `AdditionalFiles` to an array of `FileConfig` structs and add support for changing the destination file permissions.
- Add a custom [un]marshaler for `AdditionalFiles` so that the existing schema is still supported.
- Extend and add new image configuration tests to cover the new options for `AdditionalFiles`.
- Add a test for `installutils` to verify file copying behavior during image builds.

NO

###### Test Methodology

- All existing toolkit tests pass.
- New toolkits tests were added to test the new features.
- Ran a local build of `marketplace-gen2.json` and manually ensured all the `AdditionalFiles` were present in the image,

### DIFF
--- a/toolkit/docs/formats/imageconfig.md
+++ b/toolkit/docs/formats/imageconfig.md
@@ -230,16 +230,41 @@ FinalizeImageScripts provide the opportunity to run shell scripts to customize t
 
 ### AdditionalFiles
 
-The `AdditionalFiles` list provides a mechanism to add arbitrary files to the image. The elemments are are `"src": "dst"` pairs. `src` is relative to the image config `.json` file, while `dst` is an absolute path on the installed system.
+The `AdditionalFiles` list provides a mechanism to add arbitrary files to the image. The elements are are `"src": "dst"` pairs. `src` is relative to the image config `.json` file, while `dst` is an absolute path on the installed system.
 
-ISO installers will include the files on the installation media and will place them  into the final installed image.
+The `dst` can be one of:
+
+- A string representing the destination absolute path, OR
+- A object (`FileConfig`) containing the destination absolute path and other file options, OR
+- An array containing a mixture of strings and objects, which allows a single source file to be copied to multiple destination paths.
+
+ISO installers will include the files on the installation media and will place them into the final installed image.
 
 ```json
     "AdditionalFiles": [
         "../../out/tools/imager": "/installer/imager",
-        "additionalconfigs": "/etc/my/config.conf"
+        "additionalconfigs": [
+            "/etc/my/config.conf",
+            {
+                "Path": "/etc/yours/config.conf",
+                "Permissions": "664"
+            }
+        ]
     ]
 ```
+
+#### FileConfig
+
+`FileConfig` provides options to modify metadata of files copied using `AdditionalFiles`.
+
+Fields:
+
+- `Path`: The destination absolute path.
+- `Permissions`: The file permissions to apply to the file.
+
+  Supported value formats:
+
+  - Octal string: A JSON string containing an octal number. e.g. `"664"`
 
 ### Networks
 
--- a/toolkit/docs/formats/imageconfig.md
+++ b/toolkit/docs/formats/imageconfig.md
@@ -230,12 +230,14 @@ FinalizeImageScripts provide the opportunity to run shell scripts to customize t
 
 ### AdditionalFiles
 
-The `AdditionalFiles` list provides a mechanism to add arbitrary files to the image. The elements are are `"src": "dst"` pairs. `src` is relative to the image config `.json` file, while `dst` is an absolute path on the installed system.
+The `AdditionalFiles` list provides a mechanism to add arbitrary files to the image. The elements are are `"src": "dst"` pairs.
+
+`src` is relative to the image config `.json` file.
 
 The `dst` can be one of:
 
 - A string representing the destination absolute path, OR
-- A object (`FileConfig`) containing the destination absolute path and other file options, OR
+- An object (`FileConfig`) containing the destination absolute path and other file options, OR
 - An array containing a mixture of strings and objects, which allows a single source file to be copied to multiple destination paths.
 
 ISO installers will include the files on the installation media and will place them into the final installed image.
--- a/toolkit/tools/imagegen/configuration/configuration_test.go
+++ b/toolkit/tools/imagegen/configuration/configuration_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/ptrutils"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -413,9 +414,13 @@ var expectedConfiguration Config = Config{
 			KernelOptions: map[string]string{
 				"default": "kernel",
 			},
-			AdditionalFiles: map[string]string{
-				"local/path/file1": "/final/system/path",
-				"local/path/file2": "/final/system/path/renamedfile2",
+			AdditionalFiles: map[string]FileConfigList{
+				"local/path/file1": {{Path: "/final/system/path"}},
+				"local/path/file2": {{Path: "/final/system/path/renamedfile2"}},
+				"local/path/file3": {{Path: "/final/system/path/file3"}},
+				"local/path/file4": {{Path: "/final/system/path/file4", Permissions: ptrutils.PtrTo(FilePermissions(0o664))}},
+				"local/path/file5": {{Path: "/final/system/path/file5"}},
+				"local/path/file6": {{Path: "/final/system/path/file6"}, {Path: "/final/system/path/file6_copy"}},
 			},
 			Hostname: "Mariner-Test",
 			BootType: "efi",
--- a/toolkit/tools/imagegen/configuration/fileconfig.go
+++ b/toolkit/tools/imagegen/configuration/fileconfig.go
@@ -1,0 +1,119 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//
+
+package configuration
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// DestinationFileConfigList is a list of destination files to copy.
+type FileConfigList []FileConfig
+
+// FileConfig specifies options for how a file is copied in the target OS.
+type FileConfig struct {
+	// The file path in the target OS that the file will be copied to.
+	Path string `json:"Path"`
+
+	// The file permissions to set on the file.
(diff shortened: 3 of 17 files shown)


toolkit/tools/imagegen/configuration/configuration_test.go
```diff
@@ -413,9 +414,13 @@ var expectedConfiguration Config = Config{
...
 				"default": "kernel",
 			},
-			AdditionalFiles: map[string]string{
-				"local/path/file1": "/final/system/path",
-				"local/path/file2": "/final/system/path/renamedfile2",
+			AdditionalFiles: map[string]FileConfigList{
+				"local/path/file1": {{Path: "/final/system/path"}},
+				"local/path/file2": {{Path: "/final/system/path/renamedfile2"}},
+				"local/path/file3": {{Path: "/final/system/path/file3"}},
+				"local/path/file4": {{Path: "/final/system/path/file4", Permissions: ptrutils.PtrTo(FilePermissions(0o664))}},
```
dmcilvaney, line 421: I'm open to considering this (`ptrutils.PtrTo`), but maybe in a subsequent PR? If we are going to add this we should switch the existing code. We generally just use `&` in our code to get addresses.

toolkit/tools/imagegen/configuration/fileconfig.go
```diff
@@ -0,0 +1,119 @@
...
+//
+
+package configuration
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// DestinationFileConfigList is a list of destination files to copy.
```
dmcilvaney, line 13:
```suggestion
// DestinationFileConfigList is a list of destination files to generate in the final image.
```

dmcilvaney: Not critical, this just gets the idea across better to me.

toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -16,29 +16,29 @@ import (
...
+	IsKickStartBoot      bool                      `json:"IsKickStartBoot"`
+	IsIsoInstall         bool                      `json:"IsIsoInstall"`
+	BootType             string                    `json:"BootType"`
+	Hostname             string                    `json:"Hostname"`
+	Name                 string                    `json:"Name"`
+	PackageLists         []string                  `json:"PackageLists"`
+	Packages             []string                  `json:"Packages"`
+	KernelOptions        map[string]string         `json:"KernelOptions"`
+	KernelCommandLine    KernelCommandLine         `json:"KernelCommandLine"`
+	AdditionalFiles      map[string]FileConfigList `json:"AdditionalFiles"`
```
dmcilvaney, line 29: Not sure which I prefer here.. `map[string]FileConfigList` or `map[string][]FileConfig`

toolkit/tools/imagegen/configuration/fileconfig.go
```diff
@@ -0,0 +1,119 @@
...
+type FileConfigList []FileConfig
```
dmcilvaney, line 14: We haven't used types to mask lists before, but I'm not opposed to it if you feel it makes things cleaner.

toolkit/tools/imagegen/configuration/fileconfig.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (l *FileConfigList) UnmarshalJSON(b []byte) error {
+	var err error
+
+	// Try to parse as a single value.
+	var fileConfig FileConfig
+	err = json.Unmarshal(b, &fileConfig)
+	if err == nil {
+		*l = FileConfigList{fileConfig}
```
dmcilvaney, line 47: Does it make sense to also call `IsValid()` here? I'm thinking probably now with the current implementation but what if we add more checking to the validation function?

toolkit/tools/imagegen/configuration/fileconfig.go
```diff
@@ -0,0 +1,119 @@
...
+	return nil
+}
+
+func (f *FileConfig) IsValid() (err error) {
+	// Path
+	if f.Path == "" {
+		return fmt.Errorf("invalid [Path] value: empty string")
+	}
+
+	// Permissions
```
dmcilvaney, line 73: Missing validation here?

toolkit/tools/imagegen/configuration/fileconfig.go
cwize1: This type only exists to allow me to attach a custom marshaller to it.

toolkit/tools/imagegen/configuration/fileconfig.go
dmcilvaney: Generally we call recursively into the members' `IsValid()`

toolkit/tools/imagegen/configuration/configuration_test.go
cwize1: Using `&` only works when referring to a named variable or a struct initializer. For things like integers and strings, you have to use a temporary variable, which is annoying. I added `ptrutils.PtrTo` to avoid needing to use temporary variables.

I don't think we need to use `ptrutils.PtrTo` everywhere. Only where `&` doesn't work.

toolkit/tools/imagegen/configuration/systemconfig.go
cwize1: `FileConfigList` only exists to allow me to attach a custom marshaller to it.

toolkit/tools/imagegen/configuration/fileconfig.go
cwize1: Oh, I saw cases where `IsValid()` wasn't being called recursively. Also, doesn't this mean that `IsValid()` will be called multiple times? Once for every `UnmarshalJSON()` of every parent object?

toolkit/docs/formats/imageconfig.md
```diff
@@ -230,17 +230,42 @@ FinalizeImageScripts provide the opportunity to run shell scripts to customize t
...
 
-The `AdditionalFiles` list provides a mechanism to add arbitrary files to the image. The elemments are are `"src": "dst"` pairs. `src` is relative to the image config `.json` file, while `dst` is an absolute path on the installed system.
+The `AdditionalFiles` list provides a mechanism to add arbitrary files to the image. The elements are are `"src": "dst"` pairs. `src` is relative to the image config `.json` file, while `dst` is an absolute path on the installed system.
 
-ISO installers will include the files on the installation media and will place them  into the final installed image.
+The `dst` can be one of:
+
+- A string representing the destination absolute path, OR
+- A object (`FileConfig`) containing the destination absolute path and other file options, OR
+- An array containing a mixture of strings and objects, which allows a single source file to be copied to multiple destination paths.
```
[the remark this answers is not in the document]
cwize1: It is actually easier to write the custom marshallers if I allow both.

toolkit/docs/formats/imageconfig.md
```diff
@@ -230,17 +230,42 @@ FinalizeImageScripts provide the opportunity to run shell scripts to customize t
 
 ### AdditionalFiles
 
-The `AdditionalFiles` list provides a mechanism to add arbitrary files to the image. The elemments are are `"src": "dst"` pairs. `src` is relative to the image config `.json` file, while `dst` is an absolute path on the installed system.
+The `AdditionalFiles` list provides a mechanism to add arbitrary files to the image. The elements are are `"src": "dst"` pairs. `src` is relative to the image config `.json` file, while `dst` is an absolute path on the installed system.
```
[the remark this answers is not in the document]
cwize1: Done.

toolkit/docs/formats/imageconfig.md
```diff
@@ -230,17 +230,42 @@ FinalizeImageScripts provide the opportunity to run shell scripts to customize t
...
 
-ISO installers will include the files on the installation media and will place them  into the final installed image.
+The `dst` can be one of:
+
+- A string representing the destination absolute path, OR
+- A object (`FileConfig`) containing the destination absolute path and other file options, OR
```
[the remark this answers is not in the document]
cwize1: Done.

toolkit/tools/imagegen/configuration/fileconfig.go
cwize1: Fixed.

jiria: Isnt this validated as part of FileConfig UnmarshalJSON?

cwize1: @jiria This is validating the list type not the individual items in the list.

toolkit/tools/imagegen/configuration/fileconfig.go
cwize1: I changed it slightly.

toolkit/tools/imagegen/configuration/fileconfig.go
cwize1: Oh, this is kind of a mess. Not only do you call `IsValid()` during unmarshalling, but you also call it after the unmarshalling has completed. It should really be one or the other.

And if we do the validity checks during unmarshalling, then a parent object should not do validity checks on child objects with their own custom unmarshallers. Otherwise, there will be duplicate validity checking, which will unnecessarily slow down the code.

cwize1: Fixed.

toolkit/tools/imagegen/configuration/configuration_test.go
dmcilvaney: ok, fair enough. Creating the dummy variables does get annoying.

toolkit/tools/imagegen/configuration/fileconfig.go
dmcilvaney: Yup, it's a bit ungainly, but perf isn't a huge issue for us since we generally only load these structs a single time. It allows us a few benefits. Obviously we run it while unmarshalling, but it also allows us to explicitly validate a structure at any time, and at any level. Also the unit tests are a bit more powerful since you can ensure there aren't any hidden issues further inside nested values, but also directly validate the IsValid() calls you care about.

I did the initial version of this stuff in a bit of a rush though, so happy for a better pattern.

cwize1: I think it is fine for now. The config files are pretty small. So, the fact that unmarhalling them isn't O(n) probably doesn't matter too much. If it ever becomes a problem, we can fix then.

dmcilvaney (approved):

jiria (approved):

PR #5924, opened 2023-08-01
Make specreader tool run in parallel with graphpkgfetcher tool.

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When running the makefile we take `./**/*.spec` -> `*.src.rpm` -> `*.spec` -> `specs.json`, this is a bit silly. Why don't we just directly parse the specs as they exist on the disk? `*/**/*.spec` -> `specs.json`. This means we can start running the fetcher while we are also running the packer. They are both fairly light-weight resource wise, so if we can run them in parallel we can likely save some time.

While we are at it, we should also increase the number of workers since we can more heavily saturate the available network connections.

###### Change Log  

- Parse `*.spec` files directly from `./SPECS/` instead of waiting on the intermediate srpms.

**NO**

###### Test Methodology

- Local builds


toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -67,7 +67,7 @@ const (
 
 const (
 	defaultBuildDir    = "./build/SRPMS"
-	defaultWorkerCount = "10"
+	defaultWorkerCount = "80"
```
mfrw, line 70: Does this change and the above `100` affect our total time ?
I am wondering if there is a chance to get rate-limited by the blobstore ?

dmcilvaney: I've never managed to saturate the connection on my dev box. I've also got a more advanced change that scales this to the available cores but this is a stopgap to help speed up the new parallel builds in the meantime.

dmcilvaney:
Added a filter to the specreader tool that will only parse the content of the passed spec files

toolkit/tools/specreader/specreader.go
```diff
@@ -200,8 +244,40 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
+		if err != nil {
+			logger.Log.Errorf("Failed to find *.spec files. Check that %s is the correct directory. Error: %v", specsDir, err)
+			return nil, err
+		}
+	} else {
+		for specName := range parseList {
+			var specFile []string
+
+			specSearch := filepath.Join(specsDir, fmt.Sprintf("**/%s.spec", specName))
+			specFile, err = filepath.Glob(specSearch)
```
cwize1, line 263: Perhaps `matchingSpecFiles` instead of `specFile`? It is strange seeing a variable being used as a list when it has a name that isn't plural.

toolkit/tools/specreader/specreader.go
```diff
@@ -200,8 +244,40 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
 
+func findSpecFiles(specsDir string, parseList map[string]bool) (specFiles []string, err error) {
+	// Find the filepath for each spec in the SPECS directory.
+	if len(parseList) == 0 {
+		specSearch, err := filepath.Abs(filepath.Join(specsDir, "**/*.spec"))
+		if err == nil {
+			specFiles, err = filepath.Glob(specSearch)
+		}
+		if err != nil {
+			logger.Log.Errorf("Failed to find *.spec files. Check that %s is the correct directory. Error: %v", specsDir, err)
```
cwize1, line 255: Out of interest, why do you sometimes use `logger.Log.Errorf` to log extra information and in other cases use `fmt.Errorf` to wrap the error?

cwize1:
Is the title of this PR accurate? As best as I can tell, the specs were already being parsed in parallel. But this change does two things. Firstly, it increases the parallelization of the specs parsing. And secondly, it removes intermediate build steps where the .spec files are placed in SRPMs, only to be extracted again. Is this description accurate?

Perhaps a better title might be something like: `Optimize .spec file parsing`?

toolkit/tools/specreader/specreader.go
```diff
@@ -82,13 +84,55 @@ func main() {
...
+	specsAbsDir, err := filepath.Abs(*specsDir)
+	logger.PanicOnError(err, "Unable to get absolute path for specs directory '%s': %s", *specsDir, err)
+
+	err = parseSPECsWrapper(*buildDir, specsAbsDir, *rpmsDir, *srpmsDir, *existingToolchainRpmDir, *distTag, *output, *workerTar, parseList, toolchainRPMs, *workers, *runCheck)
+	logger.PanicOnError(err)
+}
+
+// parsePackListFile will parse a list of packages to pack if one is specified.
+// Duplicate list entries in the file will be removed.
+func parseParseListFile(parseListFile string) (parseList map[string]bool, err error) {
```
cwize1, line 102: Might be worth mentioning that the `bool` is a dummy value and that you are treating the `map` as a set.

toolkit/tools/specreader/specreader.go
```diff
@@ -200,8 +244,40 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
 	return
 }
 
+func findSpecFiles(specsDir string, parseList map[string]bool) (specFiles []string, err error) {
+	// Find the filepath for each spec in the SPECS directory.
+	if len(parseList) == 0 {
+		specSearch, err := filepath.Abs(filepath.Join(specsDir, "**/*.spec"))
+		if err == nil {
```
cwize1, line 251: Using non-standard error handling patterns can make the code harder to read. If you want to attach a single error log to multiple calls, then you should place those multiple calls in a helper function.

That being said, I don't think `filepath.Abs` will ever actually fail (at least on a normal OS). So, you probably only need to attach the error log to the `filepath.Glob` call.

cwize1 (approved):

toolkit/tools/specreader/specreader.go
```diff
@@ -44,6 +45,7 @@ type parseResult struct {
 var (
 	app                     = kingpin.New("specreader", "A tool to parse spec dependencies into JSON")
 	specsDir                = exe.InputDirFlag(app, "Directory to scan for SPECS")
+	parseListFile           = app.Flag("parse-list", "Path to a list of SPECs to parse. If empty will parse all SPECs.").ExistingFile()
```
neha170, line 48: Do we want to rename this to specListFile? Accordingly, we'll rename parse-list to spec-list

toolkit/tools/specreader/specreader.go
neha170, line 102: parseParseListFile sounds weird to me. Maybe parseSpecListFile? This will naturally flow from above renaming I suggested
Is this a generic parser we're making, which will be used for other purposes besides just parsing the packages here? Then, we could rename to parseListFile?

toolkit/tools/specreader/specreader.go
dmcilvaney: Just very old style. This is the very first go tool we ever built. Wanted to just keep consistent with the existing code here, but generally we prefer wrapping errors with `fmt.Errorf("...%w",err)` these days.

toolkit/tools/specreader/specreader.go
dmcilvaney: This was a dirty copy/paste of a similar function to quickly get this going... agreed this function is a bit odd. I'll rework this to follow the more standard `if err != nil  {` pattern we use.

toolkit/tools/specreader/specreader.go
dmcilvaney: For better or worse we use this pattern all over the codebase, I tend to name them 'sets' when I use them. I'll update this function to follow that pattern, although I'm going to leave the original version of this function alone for now.

dmcilvaney:
> Is the title of this PR accurate? As best as I can tell, the specs were already being parsed in parallel. But this change does two things. Firstly, it increases the parallelization of the specs parsing. And secondly, it removes intermediate build steps where the .spec files are placed in SRPMs, only to be extracted again. Is this description accurate?
> 
> Perhaps a better title might be something like: `Optimize .spec file parsing`?
Our outer makefile serialized the processing of the graph cache, and the spec parsing. So in that way we can now run both tasks in parallel. I'll massage the PR a bit to make it more clear.

toolkit/tools/specreader/specreader.go
dmcilvaney: Sure, I can rename that if it reads better. This is in contrast to the existing `packList` variables used in the srpm packer (set via SRPM_PACK_LIST=...)

toolkit/tools/specreader/specreader.go
```diff
@@ -82,13 +84,55 @@ func main() {
...
+	specsAbsDir, err := filepath.Abs(*specsDir)
+	logger.PanicOnError(err, "Unable to get absolute path for specs directory '%s': %s", *specsDir, err)
+
+	err = parseSPECsWrapper(*buildDir, specsAbsDir, *rpmsDir, *srpmsDir, *existingToolchainRpmDir, *distTag, *output, *workerTar, specListSet, toolchainRPMs, *workers, *runCheck)
+	logger.PanicOnError(err)
+}
+
+// parsePackListFile will parse a list of packages to pack if one is specified.
+// Duplicate list entries in the file will be removed.
+func parseSpecListFile(specListFile string) (specListSet map[string]bool, err error) {
```
PawelWMS, line 102: I think we're duplicating code from `srpmpacker.go`. Let's move it to a shared module, so there's only one copy.

toolkit/tools/specreader/specreader.go
```diff
@@ -200,8 +244,44 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
+	if len(specListSet) == 0 {
+		var specSearch string
+		specSearch, err = filepath.Abs(filepath.Join(specsDir, "**/*.spec"))
+		if err != nil {
+			err = fmt.Errorf("invalid spec path: '%s'. Error: %w", specsDir, err)
+			return nil, err
+		}
+		specFiles, err = filepath.Glob(specSearch)
+		if err != nil {
+			err = fmt.Errorf("failed to find *.spec files. Check that %s is the correct directory. Error: %w", specsDir, err)
```
PawelWMS, line 258:
```suggestion
			err = fmt.Errorf("failed to find *.spec files. Check that '%s' is the correct directory. Error:\n%w", specsDir, err)
```

toolkit/tools/specreader/specreader.go
```diff
@@ -200,8 +244,44 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
+			return nil, err
+		}
+		return
```
PawelWMS, line 261:
```suggestion
```

toolkit/tools/specreader/specreader.go
```diff
@@ -200,8 +244,44 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
+
+			specSearch := filepath.Join(specsDir, fmt.Sprintf("**/%s.spec", specName))
+			matchingSpecFiles, err = filepath.Glob(specSearch)
+
+			// If a SPEC is in the parse list, it should be parsed.
+			if err != nil {
+				return nil, err
+			}
+			if len(matchingSpecFiles) != 1 {
+				err = fmt.Errorf("unexpected number of matches (%d) for spec file (%s)", len(matchingSpecFiles), specName)
```
PawelWMS, line 274: Nit-pick: let's be consistent and use either brackets or single quotes when surrounding file paths.

toolkit/tools/specreader/specreader.go
```diff
@@ -200,8 +244,44 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
+	} else {
+		for specName := range specListSet {
+			var matchingSpecFiles []string
+
+			specSearch := filepath.Join(specsDir, fmt.Sprintf("**/%s.spec", specName))
+			matchingSpecFiles, err = filepath.Glob(specSearch)
+
+			// If a SPEC is in the parse list, it should be parsed.
+			if err != nil {
+				return nil, err
```
PawelWMS, line 271: Should we log an error here as well to indicate which spec name we've had issues with or is glob's error already informative enough?

toolkit/tools/specreader/specreader.go
```diff
@@ -200,8 +244,44 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
+			err = fmt.Errorf("failed to find *.spec files. Check that %s is the correct directory. Error: %w", specsDir, err)
+			return nil, err
+		}
+		return
+	} else {
+		for specName := range specListSet {
+			var matchingSpecFiles []string
+
+			specSearch := filepath.Join(specsDir, fmt.Sprintf("**/%s.spec", specName))
+			matchingSpecFiles, err = filepath.Glob(specSearch)
```
PawelWMS, line 267: Suggestion: since we're already explicitly mentioning `err` in the returns, we can drop the variable declarations.
```suggestion
		specSearch, err := filepath.Abs(filepath.Join(specsDir, "**/*.spec"))
		if err != nil {
			err = fmt.Errorf("invalid spec path: '%s'. Error: %w", specsDir, err)
			return nil, err
		}
		specFiles, err = filepath.Glob(specSearch)
		if err != nil {
			err = fmt.Errorf("failed to find *.spec files. Check that %s is the correct directory. Error: %w", specsDir, err)
			return nil, err
		}
		return
	} else {
		for specName := range specListSet {
			specSearch := filepath.Join(specsDir, fmt.Sprintf("**/%s.spec", specName))
			matchingSpecFiles, err := filepath.Glob(specSearch)
```

PawelWMS (commented):
Looks good, I'd like to remove the duplicate `parseSpecListFile` and everything else is just suggestions for minor tweaks.

toolkit/scripts/pkggen.mk
```diff
@@ -59,6 +59,7 @@ clean-spec-parse:
 	@echo Verifying no mountpoints present in $(parse_working_dir)
 	$(SCRIPTS_DIR)/safeunmount.sh "$(parse_working_dir)" && \
 	rm -rf $(parse_working_dir)
+	rm -rf $(specs_file)*
```
PawelWMS, line 62: Nit-pick: do we need the `*`?
```suggestion
	rm -rf $(specs_file)
```

PawelWMS:
I guess the `Make specs parsing parallel with fetching if SRPM_PACK_LIST isn't set.` in the PR's description is no longer needed.;)

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: Superseded by the semaphore PR, reverting.
```suggestion
	defaultWorkerCount = "10"
```

toolkit/tools/specreader/specreader.go
dmcilvaney:
```suggestion
				err = fmt.Errorf("unexpected number of matches '%d' for spec file '%s'", len(matchingSpecFiles), specName)
```

toolkit/tools/specreader/specreader.go
```diff
@@ -200,8 +244,43 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
+		if err != nil {
+			err = fmt.Errorf("failed to find *.spec files. Check that '%s' is the correct directory. Error:\n%w", specsDir, err)
+			return nil, err
+		}
+	} else {
+		for specName := range specListSet {
+			var matchingSpecFiles []string
+
+			specSearch := filepath.Join(specsDir, fmt.Sprintf("**/%s.spec", specName))
+			matchingSpecFiles, err = filepath.Glob(specSearch)
```
dmcilvaney, line 266:
```suggestion
		specSearch, err := filepath.Abs(filepath.Join(specsDir, "**/*.spec"))
		if err != nil {
			err = fmt.Errorf("invalid spec dir: '%s'. Error:\n%w", specsDir, err)
			return nil, err
		}
		specFiles, err = filepath.Glob(specSearch)
		if err != nil {
			err = fmt.Errorf("failed to find *.spec files. Check that '%s' is the correct directory. Error:\n%w", specsDir, err)
			return nil, err
		}
	} else {
		for specName := range specListSet {
			specSearch := filepath.Join(specsDir, fmt.Sprintf("**/%s.spec", specName))
			matchingSpecFiles, err := filepath.Glob(specSearch)
```

toolkit/tools/specreader/specreader.go
dmcilvaney: Fixed

toolkit/tools/specreader/specreader.go
dmcilvaney: Refactored to a new package now

toolkit/tools/specreader/specreader.go
dmcilvaney: I'd expect this to be vanishingly rare, and will likely have nothing to do with which spec we are working on. I've tried a bunch of weird names and haven't managed to break this so far. You would need to actually break the glob search itself, and given linux will take almost anything imaginable as input...

That said, I'll add the search string we are trying to use to the error.

PawelWMS (approved):

PR #5926, opened 2023-08-02
Add simple pre-cache downloader

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Query all the packages available on the upstream repos, and try to download them into the cache based on the contents of the .spec files available locally.

###### Change Log  

- Add `make pre-cache` target to pre-load the rpm cache.

**NO**

###### Test Methodology

- Local tests


toolkit/docs/building/building.md
```diff
@@ -522,6 +522,16 @@ If that is not desired all remote sources can be disabled by clearing the follow
 
 > Try to download pre-built packages if the versions match the local spec files.
 
+#### `PRECACHE=...`
+
+##### `PRECACHE=`**`n`** *(default)*
+
+> Don't pre-load the cache from upstream sources
+
+##### `PRECACHE=n`
```
cwize1, line 531: `y`?

toolkit/scripts/incremental_building.mk
```diff
@@ -62,6 +62,7 @@ DELTA_BUILD    = y
 USE_CCACHE    ?= y
 REBUILD_TOOLS ?= y
 DELTA_FETCH   ?= y
+PRECACHE	  ?= y
```
cwize1, line 65: nit: Convert tab to spaces to ensure whitespace is consistent.

dmcilvaney: Forgot to push my final commit for this, should be fixed.

cwize1 (approved):

dmcilvaney:
Re-worked to use xargs so we can limit parallelism (the build VMs start dropping connections if we queue up too many at once).

cwize1 (approved):

toolkit/scripts/precache.sh
```diff
@@ -33,58 +37,103 @@ else
...
+    # Every 5 seconds calculate the current progress and print it
+    while true; do
+        if [[ -f "$downloaded_files" ]]; then
+            current_count=$( wc -l < "$downloaded_files")
+        fi
+        if [[ -f "$skipped_file" ]]; then
+            skipped_count=$( wc -l < "$skipped_file")
+        fi
+
+        percent=$(( (current_count + skipped_count) * 100 / expected_count ))
```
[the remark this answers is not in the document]
dmcilvaney: Unfortunately if the script is killed via `ctrl+c` or similar the background process will remain.
Adding the logic to the parallel workers isn't really cleaner, since one of the requests was to limit output and I'm not aware of an easy way to pass state into the exported functions running in a sub-process (ie only print out if you are %100==0 type thing)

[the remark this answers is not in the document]
cwize1: Fair

toolkit/scripts/precache.sh
```diff
@@ -0,0 +1,111 @@
...
+rm -f "$downloaded_files"
+rm -f "$skipped_file"
+
+# For each base url, pull the meta data and write it to the repo summary file
+echo "" > $repo_summary_file
+for base_url in "${base_urls[@]}"; do
+    repo_unique="repo-$(echo $(((RANDOM%99999)+10000)))"
+    repo_name="mariner-precache-$repo_unique"
+    echo "Querying repo $base_url via 'repoquery $REPOQUERY_OS_ARGS --repofrompath=$repo_name,$base_url -a --qf="%{location}" >> $repo_summary_file'"
+    if [[ "$ID" == "mariner" ]]; then
```
PawelWMS, line 51: Since we now have some time before we need the performance upgrades, could we move this logic to be more aligned with the rest of our tooling to a Go tool? Either as an extension to our current fetcher or something along the lines of `specarchchecker` or `rpmssnapshot`?

PawelWMS (changes requested):
I think we should consider converting the script into a Go tool to be more aligned with most of the toolkit and benefit from features like more consistent logging and error checking.

toolkit/scripts/precache.sh
dmcilvaney: Sure, I can do that if we don't mind waiting on this. The script is getting really ugly anyways.

dmcilvaney: Ok, and now in golang.

dmcilvaney:
Relies on https://github.com/microsoft/CBL-Mariner/pull/5953

toolkit/scripts/precache.sh
PawelWMS: About waiting for this: that was my worry too, so I discussed our schedule with @jslobodzian and we're fine to have this finished in September. The main point was you were able to prove we can run faster and now we're given time to implement it in a cleaner way.

cwize1 (changes requested):

toolkit/scripts/precache.mk
```diff
@@ -0,0 +1,40 @@
...
+# it does, so add the phony target as a dependency to the flag file
+.PHONY: precache_always_run_phony
+pre-cache: $(STATUS_FLAGS_DIR)/precache.flag
+$(STATUS_FLAGS_DIR)/precache.flag: $(go-precacher) $(rpms_snapshot) precache_always_run_phony 
+	@if [ "$(DISABLE_UPSTREAM_REPOS)" = "y" ]; then \
+		echo "ERROR: Upstream repos are disabled (DISABLE_UPSTREAM_REPOS=y), cannot precache rpms"; \
+		exit 1; \
+	fi
+	[ -f $@ ] || touch $@ # Create the flag file if it doesn't exist in case we don't download anything
+	#$(SCRIPTS_DIR)/precache.sh "$(rpms_snapshot)" "$(precached_state_dir)" "$(CACHED_RPMS_DIR)/cache" "$(downloaded_files)" $(PACKAGE_URL_LIST) && \
```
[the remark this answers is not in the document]
dmcilvaney: Leaving for reference for now, was going to clean up once I got feedback on the go tool.

toolkit/tools/internal/file/file.go
```diff
@@ -127,6 +127,25 @@ func Write(data string, dst string) (err error) {
 	return
 }
 
+// WriteLines writes each string to the same file, separated by sep (e.g. "\n").
+func WriteLines(data []string, dst string, sep string) (err error) {
```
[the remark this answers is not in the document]
dmcilvaney: Fair point, updated to `WriteLines(dataLines []string, lineSeparator string, destinationPath string)`

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,323 @@
...
+
+// getRepoPackages returns a list of packages available in the given repoUrl by running repoquery
+func getRepoPackages(repoUrl string, isMariner bool) (packages []string, err error) {
+	const (
+		reqoqueryTool    = "repoquery"
+		randomNameLength = 10
+	)
+	var (
+		queryCommonArgList = []string{"-a", "--qf", "%{location}"}
+		marinerArgList     = []string{"-y", "-q", "--disablerepo=*"}
```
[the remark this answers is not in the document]
dmcilvaney: I'm not sure Ubuntu 22.04 even has `rpm`... let me see when it goes away. (we allways call out 18.04 as our validated env).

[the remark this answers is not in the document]
dmcilvaney: So it looks like modern ubuntu doesn't provide this command, not sure what the path here is. (On old versions `yum-utils` provides the tool. It looks like 20.04 has nothing, and 22.04 has a related python file inside `python3-dnf`.  I'll see if I can find a better way to query the repos, maybe via dnf directly?

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,323 @@
...
+		ubuntuArgList      = []string{"--show-duplicates", "--tempcache"}
+
+		finalArgList []string
+	)
+
+	logger.Log.Infof("Getting pre-caching package data from %s", repoUrl)
+	if isMariner {
```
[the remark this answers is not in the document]
dmcilvaney: I've moved this to a chroot. I was hoping to avoid it since it takes ~5-10 seconds to build up the chroot and the way this tool runs it will always be invoked even if there is nothing to build, so it's a sunk cost for every build with PRECACHE=y set, but... I think it is the right approach. This will isolate us from any host specifics.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,323 @@
...
+	return
+}
+
+func precachePackage(pkg *repocloner.RepoPackage, availablePackages map[string]string, outDir string, wg *sync.WaitGroup, results chan<- downloadResult, netOpsSemaphore chan struct{}) {
+	const (
+		downloadRetryAttempts = 2
+		downloadRetryDuration = time.Second
+	)
+
+	wg.Add(1)
```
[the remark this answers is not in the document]
dmcilvaney: Good catch, given how many goroutines we start the probabliliy is very small, but... I'm sure that would come back to annoy me at some point in the future.

toolkit/scripts/pkggen.mk
```diff
@@ -26,7 +26,8 @@ pkggen_rpms     = $(call shell_real_build_only, find $(RPMS_DIR)/*  2>/dev/null
 cache_working_dir      = $(PKGBUILD_DIR)/tdnf_cache_worker
 parse_working_dir      = $(BUILD_DIR)/spec_parsing
 rpmbuilding_logs_dir   = $(LOGS_DIR)/pkggen/rpmbuilding
-rpm_cache_files        = $(call shell_real_build_only, find $(CACHED_RPMS_DIR))
+rpm_cache_cache_dir    = $(CACHED_RPMS_DIR)/cache
```
dmcilvaney, line 29: @pawelwms Noticed you were modifying this bit, any issues with this change? (if so, its easy to move the precacher working dir somehwere else. The `find` command was picking up all the random files inside the worker chroot and causing headaches since it was looking for everything inside the root `CACHED_RPMS_DIR/` rather than `CACHED_RPMS_DIR/cache`.

toolkit/tools/internal/shell/shell.go
```diff
@@ -18,6 +18,7 @@ import (
 
 // ShellProgram is the default shell program used by the tooling.
 const ShellProgram = "/bin/bash"
+const marinerReleaseFile = "/etc/mariner-release"
```
dmcilvaney, line 21:
```suggestion
```

toolkit/docs/building/building.md
```diff
@@ -522,6 +522,16 @@ If that is not desired all remote sources can be disabled by clearing the follow
...
 
+#### `PRECACHE=...`
+
+##### `PRECACHE=`**`n`** *(default)*
+
+> Don't pre-load the cache from upstream sources
+
+##### `PRECACHE=y`
+
+> Load the cache with as many rpms from the usptream repos before starting to build.
```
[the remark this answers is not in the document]
dmcilvaney:
```suggestion
> Load the cache with as many rpms from the upstream repos before starting to build.
```

toolkit/tools/precacher/precacher.go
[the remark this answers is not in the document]
cwize1: Considering that this is a new feature, IMHO I don't think it would be too problematic to say it requires Ubuntu 22.04 to use.

cwize1 (approved):

toolkit/scripts/pkggen.mk
PawelWMS: Ignore my changes. I've just reverted them and they were purely cosmetic anyway. I can easily restore them after this PR is merged. Thanks for asking!

toolkit/docs/building/building.md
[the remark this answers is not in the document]
PawelWMS:
```suggestion
> Load the cache with RPMs from the upstream repos before starting to build.
```

toolkit/scripts/precache.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
PawelWMS, line 1: Is this file still needed?

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	// See what is available online from the upstream repos
+	availablePackages, err := getAllRepoData(*repoUrls, *workerTar, *buildDir)
+	if err != nil {
+		logger.PanicOnError(err)
+	}
+
+	logger.Log.Infof("Pre-caching: Found %d available packages", len(availablePackages))
+	if logger.Log.IsLevelEnabled(logrus.DebugLevel) {
```
PawelWMS, line 95: Why the check? You don't want this to end up in the log file by default?

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+		logger.PanicOnError(err)
+	}
+
+	// See what is available online from the upstream repos
+	availablePackages, err := getAllRepoData(*repoUrls, *workerTar, *buildDir)
+	if err != nil {
+		logger.PanicOnError(err)
+	}
+
+	logger.Log.Infof("Pre-caching: Found %d available packages", len(availablePackages))
```
PawelWMS, line 94: I like the idea of adding information where the log is coming from. Having said that, I think we shouldn't be adding that manually or encourage people to do that as well. What do you think about using the logger's [default fields](https://github.com/sirupsen/logrus#default-fields)? Something for a separate PR, but applicable globally.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+		logger.Log.Infof("Pre-caching: Creating 1st time chroot directory %s", buildDir)
+		err = os.MkdirAll(buildDir, 0755)
+		if err != nil {
+			err = fmt.Errorf("failed to create directory %s: %w", buildDir, err)
+			return nil, err
+		}
+	}
+	queryChroot, err := createChroot(workerTar, buildDir, leaveChrootOnDisk)
+	if err != nil {
+		err = fmt.Errorf("failed to create chroot: %w", err)
```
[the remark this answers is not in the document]
cwize1: Did you mean "newlines" instead of "spaces"?

toolkit/tools/precacher/precacher.go
cwize1: I believe this is just an optimization. The `logger.Log.Debugf` call won't log if "Debug" logs aren't enabled anyway. So, you don't need to iterate through the list if the "Debug" logs aren't enabled.

toolkit/scripts/precache.mk
```diff
@@ -0,0 +1,50 @@
...
+precache_chroot_dir = $(precache_state_dir)/chroot
+
+$(call create_folder,$(precache_state_dir))
+
+clean-cache: clean-precache
+clean: clean-precache
+clean-precache:
+	@echo Verifying no mountpoints present in $(precache_chroot_dir)
+	$(SCRIPTS_DIR)/safeunmount.sh "$(precache_chroot_dir)" && \
+	rm -rf $(precache_chroot_dir)
```
PawelWMS, line 20: If we're running `clean`, we don't want to leave any lingering cache behind. I'd update `clean-precache`:

```suggestion
	rm -rf $(precache_state_dir)
```

and if we want to have an option to keep the cache, we can create a separate `clean-precache-chroot` target or something along these lines.

toolkit/scripts/precache.mk
```diff
@@ -0,0 +1,50 @@
...
+		--cpu-prof-file=$(PROFILE_DIR)/precacher.cpu.pprof \
+		--mem-prof-file=$(PROFILE_DIR)/precacher.mem.pprof \
+		--trace-file=$(PROFILE_DIR)/precacher.trace \
+		$(if $(filter y,$(ENABLE_CPU_PROFILE)),--enable-cpu-prof) \
+		$(if $(filter y,$(ENABLE_MEM_PROFILE)),--enable-mem-prof) \
+		$(if $(filter y,$(ENABLE_TRACE)),--enable-trace) \
+		--timestamp-file=$(TIMESTAMP_DIR)/precacher.jsonl && \
+	if [ -s "$(precache_downloaded_files)" ]; then \
+		touch $@; \
+	fi
```
PawelWMS, line 50: If we always want to run `pre-cache`, could we just drop the status flag? Other targets depending on the pre-cache being present can depend on the `pre-cache` target as well, I believe. Or am I missing something here?
```suggestion
# We always want to run the precache script, it will decide if it needs to download anything.
.PHONY: pre-cache
pre-cache: $(go-precacher) $(chroot_worker) $(rpms_snapshot)
```

```suggestion
# We always want to run the precache script, it will decide if it needs to download anything.
.PHONY: pre-cache
pre-cache: $(go-precacher) $(chroot_worker) $(rpms_snapshot)
	@if [ "$(DISABLE_UPSTREAM_REPOS)" = "y" ]; then \
		echo "ERROR: Upstream repos are disabled (DISABLE_UPSTREAM_REPOS=y), cannot precache rpms"; \
		exit 1; \
	fi
	$(go-precacher) \
		--snapshot "$(rpms_snapshot)" \
		--output-dir "$(rpm_cache_cache_dir)" \
		--output-summary-file "$(precache_downloaded_files)" \
		$(foreach url,$(PACKAGE_URL_LIST), --repo-url "$(url)") \
		--worker-tar $(chroot_worker) \
		--worker-dir $(precache_chroot_dir) \
		--log-file=$(SRPM_BUILD_LOGS_DIR)/precacher.log \
		--log-level=$(LOG_LEVEL) \
		--cpu-prof-file=$(PROFILE_DIR)/precacher.cpu.pprof \
		--mem-prof-file=$(PROFILE_DIR)/precacher.mem.pprof \
		--trace-file=$(PROFILE_DIR)/precacher.trace \
		$(if $(filter y,$(ENABLE_CPU_PROFILE)),--enable-cpu-prof) \
		$(if $(filter y,$(ENABLE_MEM_PROFILE)),--enable-mem-prof) \
		$(if $(filter y,$(ENABLE_TRACE)),--enable-trace) \
		--timestamp-file=$(TIMESTAMP_DIR)/precacher.jsonl && \
```

toolkit/scripts/precache.mk
```diff
@@ -0,0 +1,50 @@
...
+# Contains:
+#	- Tools to fully hydrate the rpm cache from upstream repos with the any versions that match local .spec files
+
+.PHONY: pre-cache clean-precache
+
+precache_state_dir = $(CACHED_RPMS_DIR)/precache
+precache_downloaded_files = $(precache_state_dir)/downloaded_files.txt
+precache_chroot_dir = $(precache_state_dir)/chroot
+
+$(call create_folder,$(precache_state_dir))
```
PawelWMS, line 13: Don't we need to create `precache_chroot_dir` as well?

toolkit/scripts/pkggen.mk
PawelWMS, line 29: The `cache_cache` part of the name is a bit confusing. I think the main problem lies with `CACHED_RPMS_DIR`, since it's not the directory where the RPMs end up. It's more of a top directory for all things cache-related. What do you think about re-naming `rpm_cache_cache_dir` to `remote_rpms_cache_dir`, `rpm_cache_files` to `cached_remote_rpms`, and `CACHED_RPMS_DIR` to `CACHING_TOP_DIR`? Or something along these lines with a focus on clarifying what each variable contains.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	timestamp.BeginTiming("precacher", *timestampFile)
+	defer timestamp.CompleteTiming()
+
+	// Get the state of the current repo from a snapshot
+	rpmSnapshot, err := rpmSnapshotFromFile(*snapshot)
+	if err != nil {
+		logger.PanicOnError(err)
+	}
+
+	// See what is available online from the upstream repos
```
PawelWMS, line 88: General nit-pick: I think we're being a bit too noisy with the comments and that's making it harder to read what's happening in the code. Especially since the function names are already pretty helpful in understanding what's happening.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+		leaveChrootOnDisk = false
+	)
+	timestamp.StartEvent("pull available package data from repos", nil)
+	defer timestamp.StopEvent(nil)
+
+	// Create a chroot to run repoquery in
+	// Create the directory 'buildDir' if it does not exist
+	exists, err := file.DirExists(buildDir)
+	if err != nil {
+		err = fmt.Errorf("failed to check if directory %s exists: %w", buildDir, err)
```
PawelWMS, line 132: Could we surround file/directory names with brackets or single quotes to make debugging easier? This helps a lot, if by mistake the variable ends up being an empty string.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+		return nil, err
+	}
+	if !exists {
+		logger.Log.Infof("Pre-caching: Creating 1st time chroot directory %s", buildDir)
+		err = os.MkdirAll(buildDir, 0755)
+		if err != nil {
+			err = fmt.Errorf("failed to create directory %s: %w", buildDir, err)
+			return nil, err
+		}
+	}
```
PawelWMS, line 142: Why not just run `os.MkdirAll` without the check and remove most of this code?

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+}
+
+func rpmSnapshotFromFile(snapshotFile string) (rpmSnapshot *repocloner.RepoContents, err error) {
+	err = jsonutils.ReadJSONFile(snapshotFile, &rpmSnapshot)
+	return
+}
+
+// getAllRepoData returns a map of package names to URLs for all packages available in the given repos. It uses
+// a chroot to run repoquery.
+func getAllRepoData(repoUrls []string, workerTar, buildDir string) (namesToUrls map[string]string, err error) {
```
PawelWMS, line 121: Nit-pick: let's try using proper abbreviations:
```suggestion
func getAllRepoData(repoURLs []string, workerTar, buildDir string) (namesToURLs map[string]string, err error) {
```

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+		return nil, err
+	}
+	defer queryChroot.Close(leaveChrootOnDisk)
+
+	namesToUrls = make(map[string]string)
+	for _, repoUrl := range repoUrls {
+		// Use the chroot to query each repo for the packages it contains
+		var packages []string
+		err = queryChroot.Run(func() (chrootErr error) {
+			packages, chrootErr = getRepoPackages(repoUrl)
```
PawelWMS, line 155: The name `packages` and `getRepoPackages` is misleading, since we're actually retrieving package URLs. It makes reading the `for _, pkgUrl := range packages` loop a bit confusing. What do you think about `packageURLs` and `getPackageURLs` instead? Or something else, that would make the contents of the variables clearer?

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+		// in the form of "<PARTIAL_URL>/<name>-<version>.<distro>.<arch>.rpm"
+		for _, pkgUrl := range packages {
+			name := path.Base(pkgUrl)
+			name = strings.TrimSuffix(name, ".rpm")
+
+			// We need to prepend the repoUrl to the partial URL to get the full URL
+			pkgUrl = fmt.Sprintf("%s/%s", repoUrl, pkgUrl)
+			namesToUrls[name] = pkgUrl
+		}
+	}
```
PawelWMS, line 172: OK, now I see we don't retrieve full URLs but just the path component of the URL with `getRepoPackage`. In that case, how about:
```suggestion
	for _, repoURL := range repoURLs {
		// Use the chroot to query each repo for the packages it contains
		var packages []string
		err = queryChroot.Run(func() (chrootErr error) {
			packageRepoPaths, chrootErr = getPackageRepoPaths(repoURL)
			return chrootErr
		})
		if err != nil {
			return nil, err
		}

		// We will be searching by the name: "<name>-<version>.<distro>.<arch>", the results from the repoquery will be
		// in the form of "<PARTIAL_URL>/<name>-<version>.<distro>.<arch>.rpm"
		for _, packageRepoPath := range packageRepoPaths {
			packageName := path.Base(packageRepoPath)
			packageName = strings.TrimSuffix(packageName, ".rpm")

			// We need to prepend the repoURL to the partial URL to get the full URL
			pkgURL = fmt.Sprintf("%s/%s", repoURL, packageRepoPath)
			namesToUrls[name] = pkgURL
		}
	}
```

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	err = queryChroot.Initialize(workerTar, nil, nil)
+	if err != nil {
+		err = fmt.Errorf("failed to initialize chroot: %w", err)
+		return
+	}
+
+	// We will need network to install the repoquery package
+	files := []safechroot.FileToCopy{
+		{Src: "/etc/resolv.conf", Dest: "/etc/resolv.conf"},
+	}
```
PawelWMS, line 197: I've seen we're repeating this already multiple times across our code. I'm fine leaving this code the way it is, but if you find a minute and it's going to turn out to be simple, what do you think about adding a `EnableNetwork` method to `*safechroot.Chroot` and cleaning up the repeated code?

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	err = queryChroot.AddFiles(files...)
+	if err != nil {
+		err = fmt.Errorf("failed to add files to chroot: %w", err)
+		queryChroot.Close(false)
```
PawelWMS, line 201: To be on the safe side in case of any future changes adding more early returns, what do you think about conditionally calling `queryChroot.Close(false)` only once from a deferred call? Something like what `srpmpacker.go` does inside its `createChroot` function.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+
+// getRepoPackages returns a list of packages available in the given repoUrl by running repoquery
+func getRepoPackages(repoUrl string) (packages []string, err error) {
+	const (
+		reqoqueryTool    = "repoquery"
+		randomNameLength = 10
+	)
+	var (
+		queryCommonArgList = []string{"-y", "-q", "--disablerepo=*", "-a", "--qf", "%{location}"}
+	)
```
PawelWMS, line 230:
```suggestion
	var queryCommonArgList = []string{"-y", "-q", "--disablerepo=*", "-a", "--qf", "%{location}"}
```

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	repoPathArg := fmt.Sprintf("--repofrompath=mariner-precache-%s,%s", randomName, repoUrl)
+	finalArgList := append(queryCommonArgList, repoPathArg)
+
+	onStdout := func(args ...interface{}) {
+		line := args[0].(string)
+		packages = append(packages, line)
+	}
+
+	// Run the repoquery command
+	err = shell.ExecuteLiveWithCallback(onStdout, logger.Log.Warn, true, reqoqueryTool, finalArgList...)
```
PawelWMS, line 249: Could we move the `true` to a named constant?

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	// We will be downloading packages concurrently, so we need to keep track of when they are all done via a wait group. To
+	// simplify the code just use goroutines with a semaphore channel to limit the number of concurrent network operations. Each
+	// goroutine will handle a single package and will send the result of the download to a results channel. The main thread will
+	// monitor the results channel and update the progress counter accordingly.
+	wg := new(sync.WaitGroup)
+	netOpsSemaphore := make(chan struct{}, concurrentNetOps)
+	results := make(chan downloadResult)
+	doneChannel := make(chan struct{})
+
+	// Spawn a worker for each package, they will all do preliminary checks in parrallel before synchronizing on the semaphore.
```
PawelWMS, line 275:
```suggestion
	// Spawn a worker for each package, they will all do preliminary checks in parallel before synchronizing on the semaphore.
```

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	// Each worker is responsible for adding and removing itself from the wait group.
```
PawelWMS, line 276: Nit-pick: the workers don't add themselves to wait groups, we're doing that still in the main thread.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+
+	if err != nil {
+		logger.Log.Warnf("Pre-caching '%s' failed. Error: %s", fileName, err)
+		result.resultType = downloadResultTypeFailure
+		results <- result
+		return
+	} else {
+		logger.Log.Debugf("Pre-caching '%s' succeeded", fileName)
+		result.resultType = downloadResultTypeSuccess
+		results <- result
```
PawelWMS, line 397: Since this is all happening in multiple threads, I'm not sure, if we shouldn't move all of this logging to `monitorProgress` instead. Otherwise, in case of errors, the debug logs might end up very hard to read.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+
+// monitorProgress will wait for results from the downloadResult channel and update the progress counter accordingly. If
+// no more results are available we expect the done channel to be closed, at which point we will return.
+func monitorProgress(total int, results chan downloadResult, doneChannel chan struct{}) (downloadedPackages []string) {
+	downloaded := 0
+	skipped := 0
+	failed := 0
+	unavailable := 0
+	progressIncrement := 10.0
+	lastProgressUpdate := progressIncrement * -1
```
PawelWMS, line 302: Nit-pick:
```suggestion
	const progressIncrement = 10.0

	downloaded := 0
	skipped := 0
	failed := 0
	unavailable := 0
	lastProgressUpdate := progressIncrement * -1
```

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	// File names are of the form "<name>-<version>.<distro>.<arch>.rpm"
+	pkgName := fmt.Sprintf("%s-%s.%s.%s", pkg.Name, pkg.Version, pkg.Distribution, pkg.Architecture)
+	fileName := fmt.Sprintf("%s.rpm", pkgName)
+	fullFilePath := path.Join(outDir, fileName)
+	result := downloadResult{
+		pkgName: fileName,
+	}
+
+	// Bail out early if the file already exists
+	if exists, _ := file.PathExists(fullFilePath); exists {
```
PawelWMS, line 359: Why ignore the error from `PathExists`?

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	defer timestamp.CompleteTiming()
+
+	// Get the state of the current repo from a snapshot
+	rpmSnapshot, err := rpmSnapshotFromFile(*snapshot)
+	if err != nil {
+		logger.PanicOnError(err)
+	}
+
+	// See what is available online from the upstream repos
+	availablePackages, err := getAllRepoData(*repoUrls, *workerTar, *buildDir)
```
PawelWMS, line 89: Reading "available" later in the code made me thing this is a set of packages we already know are available upstream and it was confusing to later find error logs for when we fail to actually find them in any repo. Could we re-name this variable a bit? Maybe `localPackages` or `packagesToDownload`?

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	if err != nil {
+		logger.Log.Warnf("Pre-caching '%s' failed. Error: %s", fileName, err)
+		result.resultType = downloadResultTypeFailure
+		results <- result
+		return
+	} else {
+		logger.Log.Debugf("Pre-caching '%s' succeeded", fileName)
+		result.resultType = downloadResultTypeSuccess
+		results <- result
+	}
```
PawelWMS, line 398: Nit-pick:

```suggestion
	}

	logger.Log.Debugf("Pre-caching '%s' succeeded", fileName)
	result.resultType = downloadResultTypeSuccess
	results <- result
```

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,409 @@
...
+	}()
+
+	logger.Log.Debugf("Pre-caching '%s' from '%s'", fileName, url)
+	err := retry.Run(func() error {
+		err := network.DownloadFile(url, fullFilePath, nil, nil)
+		if err != nil {
+			logger.Log.Warnf("Pre-caching: Attempt to download (%s) failed. Error: %s", url, err)
+		}
+		return err
+	}, downloadRetryAttempts, downloadRetryDuration)
```
PawelWMS, line 387: While trivial, why not make hold the semaphore even closer to the download logic? Assuming it's not any additional effort. One idea:
```suggestion
	logger.Log.Debugf("Pre-caching '%s' from '%s'", fileName, url)
	err := retry.Run(func() error {
		// Limit the number of concurrent network operations to avoid overloading the network.
		netOpsSemaphore <- struct{}{}
		defer func() {
			<-netOpsSemaphore
		}()

		err := network.DownloadFile(url, fullFilePath, nil, nil)
		if err != nil {
			logger.Log.Warnf("Pre-caching: Attempt to download (%s) failed. Error: %s", url, err)
		}
		return err
	}, downloadRetryAttempts, downloadRetryDuration)
```

Although, I recall @cwize1 saying he's got a better idea to limit the network connections, so I'm guessing this comment might become completely obsolete soon...

PawelWMS: Also, since we're logging errors below as well, I'm curious if we can simplify and reduce the number of repeated error logs in case of multiple download failures:
```suggestion
	logger.Log.Debugf("Pre-caching '%s' from '%s'", fileName, url)
	err := retry.Run(func() error {
		// Limit the number of concurrent network operations to avoid overloading the network.
		netOpsSemaphore <- struct{}{}
		defer func() {
			<-netOpsSemaphore
		}()

		return network.DownloadFile(url, fullFilePath, nil, nil)
	}, downloadRetryAttempts, downloadRetryDuration)
```

toolkit/tools/internal/file/file.go
```diff
@@ -127,6 +127,29 @@ func Write(data string, dst string) (err error) {
...
+		if err != nil {
+			return
+		}
+		_, err = dstFile.WriteString(lineSeparator)
+		if err != nil {
+			return
+		}
+	}
+	return
+}
```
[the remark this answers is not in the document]
dmcilvaney: Was thinking it would be more flexible for dealing with lines that have `\n` already, but I don't think we really have many. I'll swap.

toolkit/scripts/precache.mk
dmcilvaney: The tools actually complain if it already exists, the chroot library will create it (unless its a re-used chroot for docker environments etc.).

toolkit/scripts/precache.mk
dmcilvaney: Its a very niche behavior in Make, but important. A `.PHONY` target will always be 'out of date', ie anything that depends on it will rebuild.

In this case, we always want the tool to run and make sure everything is looking good, but if we run it and decide that nothing interesting happened (ie we didn't download anything) then we shouldn't trigger any follow on steps because of this tool. That is why we check the contents of the summary file, only if the summary file has something in it do we `touch` the flag file.

If Make runs a recipe to make a file, and the file doesn't get changed, it assumes that nothing that depends on it needs to regenerate (assuming all other dependencies are also unchanged).

toolkit/tools/precacher/precacher.go
dmcilvaney: This can be a few thousand passes through the loop, and while the logging package is smart enough to delay running the format command if its unneeded it seemed like enough thrashing to be worth short-circuiting.

toolkit/tools/precacher/precacher.go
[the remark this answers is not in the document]
dmcilvaney: Ah, copy pasted this from elsewhere in the code that predates that switch. I'll update.
(Although, I'm leaning towards mabybe doing `"error msg:\n\t%w"` or something, it can be hard to follow the errors on the console sometimes since they now lack the formatted red header on every line.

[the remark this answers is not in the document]
dmcilvaney: Ah, looks like I went back to the old style elsewhere as well. Fixed those too.

toolkit/tools/precacher/precacher.go
dmcilvaney: Fixed

toolkit/tools/precacher/precacher.go
dmcilvaney: I was thinking something similar. Was also considering setting up the simple chroot tool package to use it as well. I'll add a todo but fix it later.

toolkit/tools/precacher/precacher.go
dmcilvaney: Ah, I like that. Switched.

toolkit/tools/precacher/precacher.go
dmcilvaney: Stale comment, there was a race condition that necessitated this change. Fixed

toolkit/tools/precacher/precacher.go
dmcilvaney: This function is very unlikely to error, and I don't think we care if it does. An error here is effectively the same thing as not having the file. If there is something really wrong then we will hear about it when we try to re-write the file.

toolkit/tools/precacher/precacher.go
dmcilvaney: I think its actually beneficial to have the semaphore outside the retry loop. If its inside, we will just keep hammering the connection as quickly as possible for each worker while what we actually want is the system to globally slow down when errors start occurring.

I'm also personally interested in keeping the logs, we generally shouldn't be seeing these since we won't try to download anything that doesn't exist. So anything we see here is going to be unexpected and transient. I think it would be nice to see how common the errors are, and how long they persist.

toolkit/tools/precacher/precacher.go
dmcilvaney: Done, had most of it there anyways and I think I meant to move these bits as well.

toolkit/scripts/pkggen.mk
dmcilvaney: Agreed, but I don't think I want to change `CACHED_RPMS_DIR` since its a public facing variable... I'll swap the others though.

toolkit/tools/precacher/precacher.go
dmcilvaney: I've updated the variable to be a bit more descriptive.

toolkit/scripts/precache.mk
```diff
@@ -0,0 +1,51 @@
...
+	$(SCRIPTS_DIR)/safeunmount.sh "$(precache_chroot_dir)" && \
+	rm -rf $(precache_state_dir)
+
+# We always want to run the precache script, it will decide if it needs to download anything and update the flag file if
+# it does, so add the phony target as a dependency to the flag file
+.PHONY: precache_always_run_phony
+pre-cache: $(STATUS_FLAGS_DIR)/precache.flag
+$(STATUS_FLAGS_DIR)/precache.flag: $(go-precacher) $(chroot_worker) $(rpms_snapshot) precache_always_run_phony 
+	@if [ "$(DISABLE_UPSTREAM_REPOS)" = "y" ]; then \
+		echo "ERROR: Upstream repos are disabled (DISABLE_UPSTREAM_REPOS=y), cannot precache rpms"; \
```
PawelWMS, line 29: Nit-pick:
```suggestion
		echo "ERROR: Upstream repos are disabled (DISABLE_UPSTREAM_REPOS=y), cannot precache RPMs"; \
```

toolkit/scripts/precache.mk
```diff
@@ -0,0 +1,51 @@
...
+		--cpu-prof-file=$(PROFILE_DIR)/precacher.cpu.pprof \
+		--mem-prof-file=$(PROFILE_DIR)/precacher.mem.pprof \
+		--trace-file=$(PROFILE_DIR)/precacher.trace \
+		$(if $(filter y,$(ENABLE_CPU_PROFILE)),--enable-cpu-prof) \
+		$(if $(filter y,$(ENABLE_MEM_PROFILE)),--enable-mem-prof) \
+		$(if $(filter y,$(ENABLE_TRACE)),--enable-trace) \
+		--timestamp-file=$(TIMESTAMP_DIR)/precacher.jsonl && \
+	if [ -s "$(precache_downloaded_files)" ]; then \
+		touch $@; \
+	fi
```
PawelWMS, line 51: Nit-pick: why do flag file creation in two places?
```suggestion
	$(go-precacher) \
		--snapshot "$(rpms_snapshot)" \
		--output-dir "$(remote_rpms_cache_dir)" \
		--output-summary-file "$(precache_downloaded_files)" \
		$(foreach url,$(PACKAGE_URL_LIST), --repo-url "$(url)") \
		--worker-tar $(chroot_worker) \
		--worker-dir $(precache_chroot_dir) \
		--log-file=$(SRPM_BUILD_LOGS_DIR)/precacher.log \
		--log-level=$(LOG_LEVEL) \
		--cpu-prof-file=$(PROFILE_DIR)/precacher.cpu.pprof \
		--mem-prof-file=$(PROFILE_DIR)/precacher.mem.pprof \
		--trace-file=$(PROFILE_DIR)/precacher.trace \
		$(if $(filter y,$(ENABLE_CPU_PROFILE)),--enable-cpu-prof) \
		$(if $(filter y,$(ENABLE_MEM_PROFILE)),--enable-mem-prof) \
		$(if $(filter y,$(ENABLE_TRACE)),--enable-trace) \
		--timestamp-file=$(TIMESTAMP_DIR)/precacher.jsonl && \
	if [ ! -f $@ || -s "$(precache_downloaded_files)" ]; then \
		touch $@; \
	fi
```

dmcilvaney: Hmm, that should work as well. Will do

dmcilvaney:
Merged recent tooling changes into this branch

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,426 @@
...
+
+// downloadMissingPackages will attemp to download each package listed in rpmSnapshot that is not already present in the
+// outDir. It will return a list of the packages that were downloaded. It will use concurrentNetOps to limit the number of
+// concurrent network operations used to download the missing packages. It will also monitor the results and print periodic
+// progress updates to the console.
+func downloadMissingPackages(rpmSnapshot *repocloner.RepoContents, packagesAvailableFromRepos map[string]string, outDir string, concurrentNetOps uint) (downloadedPackages []string, err error) {
+	timestamp.StartEvent("download missing packages", nil)
+	defer timestamp.StopEvent(nil)
+
+	err = os.MkdirAll(outDir, 0755)
```
PawelWMS, line 270: Similar suggestion about `outDir` to make it a required existing directory.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,426 @@
...
+
+	if err != nil {
+		result.resultType = downloadResultTypeFailure
+		results <- result
+		return
+	}
+
+	result.resultType = downloadResultTypeSuccess
+	results <- result
+}
```
PawelWMS, line 416: Idea to make sure we:
- never forget to send a result,
- educe the number of times we have to do it,
- and make erroring-out require the least amount of changes.

A different default would be also OK, but it's more common to error-out in multiple places, while other statuses are probably more likely to be set only one time.

```suggestion
	const (
		downloadRetryAttempts = 2
		downloadRetryDuration = time.Second
	)

	// File names are of the form "<name>-<version>.<distro>.<arch>.rpm"
	pkgName := fmt.Sprintf("%s-%s.%s.%s", pkg.Name, pkg.Version, pkg.Distribution, pkg.Architecture)
	fileName := fmt.Sprintf("%s.rpm", pkgName)
	fullFilePath := path.Join(outDir, fileName)
	result := downloadResult{
		pkgName: fileName,
		resultType: downloadResultTypeFailure,
	}

	defer defer func() {
		results <- result
		wg.Done()
	}()

	// Bail out early if the file already exists
	if exists, _ := file.PathExists(fullFilePath); exists {
		result.resultType = downloadResultTypeSkipped
		return
	}

	// Get the url for the package, or bail out if it is not available
	url, ok := packagesAvailableFromRepos[pkgName]
	if !ok {
		result.resultType = donwloadResultTypeUnavailable
		return
	}

	// Limit the number of concurrent network operations to avoid overloading the network. All work past this point
	// is either network related, or trivial, so we can safely hold the semaphore until we are done.
	netOpsSemaphore <- struct{}{}
	defer func() {
		<-netOpsSemaphore
	}()

	logger.Log.Debugf("Pre-caching '%s' from '%s'", fileName, url)
	err := retry.Run(func() error {
		err := network.DownloadFile(url, fullFilePath, nil, nil)
		if err != nil {
			logger.Log.Warnf("Attempt to download (%s) failed. Error: %s", url, err)
		}
		return err
	}, downloadRetryAttempts, downloadRetryDuration)
	if err != nil {
		return
	}

	result.resultType = downloadResultTypeSuccess
}
```

PawelWMS (commented):
Just a few minor comments.

toolkit/tools/precacher/precacher.go
```diff
@@ -0,0 +1,426 @@
...
+		leaveChrootOnDisk = false
+	)
+	timestamp.StartEvent("pull available package data from repos", nil)
+	defer timestamp.StopEvent(nil)
+
+	err = os.MkdirAll(buildDir, 0755)
+	if err != nil {
+		err = fmt.Errorf("failed to create directory '%s':\n%w", buildDir, err)
+		return nil, err
+	}
```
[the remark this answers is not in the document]
dmcilvaney: Done

PawelWMS (approved):

PR #5931, opened 2023-08-03
Add better backoff, semaphore to packer

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
During testing of the new parallel build PR gates, it was noticed that the build agents often dropped connections. We can run most of the packing step highly parallelized, but limiting the number of concurrent network operations significantly improves reliability.

Also adding a configurable exponent to the retry backoff logic allows us to be more accommodating by significantly slowing down network requests when things start to go wrong.

###### Change Log  

- Add retry.RunWithExpBackoff() which has a configurable backoff exponent (rather than the implicit `1.0` in `Run()`).
- Add a semaphore to limit concurrent network traffic in the packer

**NO**

###### Test Methodology

- Local builds

### DIFF
--- a/toolkit/tools/internal/retry/retry.go
+++ b/toolkit/tools/internal/retry/retry.go
@@ -35,12 +35,12 @@ func calculateLinearDelay(failCount int, sleep time.Duration) time.Duration {
 	return sleep * time.Duration(failCount)
 }
 
-// backoffSleep sleeps for the given duration, unless the cancelSignal is triggered first. The cancelSignal may be nil
-// in which case the sleep will always complete.
-func backoffSleep(delay time.Duration, cancelSignal <-chan struct{}) (cancelled bool) {
+// backoffSleep sleeps for the given duration, unless the cancel channel is triggered or closed first. The cancel channel
+// may be nil in which case the sleep will always complete.
+func backoffSleep(delay time.Duration, cancel <-chan struct{}) (cancelled bool) {
 	select {
 	case <-time.After(delay):
-	case <-cancelSignal:
+	case <-cancel:
 		cancelled = true
 	}
 	return
@@ -48,10 +48,11 @@ func backoffSleep(delay time.Duration, cancelSignal <-chan struct{}) (cancelled 
 
 // runWithBackoffInternal runs function up to 'attempts' times, waiting delayCalc(failCount) before each i-th attempt.
 // delayCalc(0) is expected to return 0.
-func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int, cancelSignal <-chan struct{}) (err error) {
+// The function will return early if the cancel channel is closed.
+func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int, cancel <-chan struct{}) (err error) {
 	for failures := 0; failures < attempts; failures++ {
 		delayTime := delayCalc(failures)
-		cancelled := backoffSleep(delayTime, cancelSignal)
+		cancelled := backoffSleep(delayTime, cancel)
 		if cancelled {
 			err = fmt.Errorf("%w after %d tries", ErrRetryCancelled, failures)
 			break
@@ -69,17 +70,17 @@ func Run(function func() error, attempts int, sleep time.Duration) (err error) {
 }
 
 // RunWithLinearBackoff runs function up to 'attempts' times, waiting i * sleep duration before each i-th attempt. An
-// optional cancelSignal can be provided to cancel the retry loop immediately.
-func RunWithLinearBackoff(function func() error, attempts int, sleep time.Duration, cancelSignal <-chan struct{}) (err error) {
+// optional cancel channel can be provided to cancel the retry loop immediately by closing the channel.
+func RunWithLinearBackoff(function func() error, attempts int, sleep time.Duration, cancel <-chan struct{}) (err error) {
 	return runWithBackoffInternal(function, func(failCount int) time.Duration {
 		return calculateLinearDelay(failCount, sleep)
-	}, attempts, cancelSignal)
+	}, attempts, cancel)
 }
 
 // RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExponentBase^(i-1) * sleep' duration before
-// each i-th attempt. An optional cancelSignal can be provided to cancel the retry loop immediately.
-func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64, cancelSignal <-chan struct{}) (err error) {
+// each i-th attempt. An optional cancel channel can be provided to cancel the retry loop immediately by closing the channel.
+func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64, cancel <-chan struct{}) (err error) {
 	return runWithBackoffInternal(function, func(failCount int) time.Duration {
 		return calculateExpDelay(failCount, sleep, backoffExponentBase)
-	}, attempts, cancelSignal)
+	}, attempts, cancel)
 }
--- a/toolkit/tools/internal/retry/retry.go
+++ b/toolkit/tools/internal/retry/retry.go
@@ -8,25 +8,35 @@ import (
 	"time"
 )
 
-func calculateDelay(retryCount int, sleep time.Duration, backoffExponent float64) time.Duration {
+// calculateDelay calculates the delay for the given failure count, sleep duration, and backoff exponent base.
+// If the base is negative, it will calculate a linear backoff.
+// If the base is positive, it will calculate an exponential backoff.
+func calculateDelay(failCount int, sleep time.Duration, backoffExponentBase float64) time.Duration {
+	if failCount == 0 {
+		return 0
+	}
+	// Treat negative values of base as linear backoff.
+	if backoffExponentBase < 0 {
+		return sleep * time.Duration(failCount)
+	}
 	// Calculate an exponential backoff.
-	// The formula is: sleep * (backoffExp ^ retryCount)
-	// For example, if sleep = 1 second, backoffExp = 2.0, retryCount = 3
+	// The formula is: sleep * (backoffExpBase ^ failCount)
+	// For example, if sleep = 1 second, backoffExp = 2.0, failCount = 3
 	// then the delay will be 1 * (2 ^ 3) = 8 seconds.
-	expRetry := math.Pow(backoffExp, float64(retryCount))
+	expRetry := math.Pow(backoffExponentBase, float64(failCount-1))
 	return time.Duration(expRetry * float64(sleep))
 }
 
 // Run runs function up to attempts times, waiting i * sleep duration before each i-th attempt.
 func Run(function func() error, attempts int, sleep time.Duration) (err error) {
-	return RunWithExpBackoff(function, attempts, sleep, 1.0)
+	return RunWithExpBackoff(function, attempts, sleep, -1)
 }
 
-// RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExp^i * sleep' duration before each i-th attempt.
-func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExp float64) (err error) {
-	for i := 0; i < attempts; i++ {
+// RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExponentBase^i * sleep' duration before each i-th attempt.
+func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64) (err error) {
+	for failures := 0; failures < attempts; failures++ {
 		// Calculate an exponential backoff. For i=0 we will calculate 0 delay and immediately call the function.
-		time.Sleep(calcDelay(i, sleep, backoffExp))
+		time.Sleep(calculateDelay(failures, sleep, backoffExponentBase))
 		if err = function(); err == nil {
 			break
 		}
--- a/toolkit/tools/internal/retry/retry.go
+++ b/toolkit/tools/internal/retry/retry.go
@@ -4,12 +4,9 @@
 package retry
 
 import (
-	"fmt"
 	"math"
 	"time"
 )
-
-var ErrRetryCancelled = fmt.Errorf("retry cancelled")
 
 // calculateDelay calculates the delay for the given failure count, sleep duration, and backoff exponent base.
 // If the base is positive, it will calculate an exponential backoff.
@@ -49,29 +46,29 @@ func backoffSleep(delay time.Duration, cancel <-chan struct{}) (cancelled bool) 
 // runWithBackoffInternal runs function up to 'attempts' times, waiting delayCalc(failCount) before each i-th attempt.
 // delayCalc(0) is expected to return 0.
 // The function will return early if the cancel channel is closed.
-func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int, cancel <-chan struct{}) (err error) {
+func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int, cancel <-chan struct{}) (wasCancelled bool, err error) {
 	for failures := 0; failures < attempts; failures++ {
 		delayTime := delayCalc(failures)
-		cancelled := backoffSleep(delayTime, cancel)
-		if cancelled {
-			err = fmt.Errorf("%w after %d tries", ErrRetryCancelled, failures)
+		wasCancelled = backoffSleep(delayTime, cancel)
+		if wasCancelled {
 			break
 		}
 		if err = function(); err == nil {
 			break
 		}
 	}
-	return err
+	return wasCancelled, err
 }
 
 // Run runs function up to 'attempts' times, waiting i * sleep duration before each i-th attempt.
 func Run(function func() error, attempts int, sleep time.Duration) (err error) {
-	return RunWithLinearBackoff(function, attempts, sleep, nil)
+	_, err = RunWithLinearBackoff(function, attempts, sleep, nil)
+	return
 }
 
 // RunWithLinearBackoff runs function up to 'attempts' times, waiting i * sleep duration before each i-th attempt. An
 // optional cancel channel can be provided to cancel the retry loop immediately by closing the channel.
-func RunWithLinearBackoff(function func() error, attempts int, sleep time.Duration, cancel <-chan struct{}) (err error) {
+func RunWithLinearBackoff(function func() error, attempts int, sleep time.Duration, cancel <-chan struct{}) (wasCancelled bool, err error) {
 	return runWithBackoffInternal(function, func(failCount int) time.Duration {
 		return calculateLinearDelay(failCount, sleep)
 	}, attempts, cancel)
@@ -79,7 +76,7 @@ func RunWithLinearBackoff(function func() error, attempts int, sleep time.Durati
 
 // RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExponentBase^(i-1) * sleep' duration before
 // each i-th attempt. An optional cancel channel can be provided to cancel the retry loop immediately by closing the channel.
-func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64, cancel <-chan struct{}) (err error) {
+func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64, cancel <-chan struct{}) (wasCancelled bool, err error) {
 	return runWithBackoffInternal(function, func(failCount int) time.Duration {
 		return calculateExpDelay(failCount, sleep, backoffExponentBase)
 	}, attempts, cancel)
--- a/toolkit/tools/internal/retry/retry.go
+++ b/toolkit/tools/internal/retry/retry.go
@@ -4,9 +4,12 @@
 package retry
 
 import (
+	"fmt"
 	"math"
 	"time"
 )
+
+var ErrRetryCancelled = fmt.Errorf("retry cancelled")
 
 // calculateDelay calculates the delay for the given failure count, sleep duration, and backoff exponent base.
 // If the base is positive, it will calculate an exponential backoff.
@@ -32,12 +35,27 @@ func calculateLinearDelay(failCount int, sleep time.Duration) time.Duration {
 	return sleep * time.Duration(failCount)
 }
 
+// backoffSleep sleeps for the given duration, unless the cancelSignal is triggered first. The cancelSignal may be nil
+// in which case the sleep will always complete.
+func backoffSleep(delay time.Duration, cancelSignal <-chan struct{}) (cancelled bool) {
+	select {
+	case <-time.After(delay):
+	case <-cancelSignal:
+		cancelled = true
+	}
+	return
+}
+
 // runWithBackoffInternal runs function up to 'attempts' times, waiting delayCalc(failCount) before each i-th attempt.
 // delayCalc(0) is expected to return 0.
-func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int) (err error) {
+func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int, cancelSignal <-chan struct{}) (err error) {
 	for failures := 0; failures < attempts; failures++ {
 		delayTime := delayCalc(failures)
-		time.Sleep(delayTime)
+		cancelled := backoffSleep(delayTime, cancelSignal)
+		if cancelled {
+			err = fmt.Errorf("%w after %d tries", ErrRetryCancelled, failures)
+			break
+		}
 		if err = function(); err == nil {
 			break
 		}
@@ -47,14 +65,21 @@ func runWithBackoffInternal(function func() error, delayCalc func(failCount int)
 
 // Run runs function up to 'attempts' times, waiting i * sleep duration before each i-th attempt.
 func Run(function func() error, attempts int, sleep time.Duration) (err error) {
-	return runWithBackoffInternal(function, func(failCount int) time.Duration {
-		return calculateLinearDelay(failCount, sleep)
-	}, attempts)
+	return RunWithLinearBackoff(function, attempts, sleep, nil)
 }
 
-// RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExponentBase^(i-1) * sleep' duration before each i-th attempt.
-func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64) (err error) {
+// RunWithLinearBackoff runs function up to 'attempts' times, waiting i * sleep duration before each i-th attempt. An
+// optional cancelSignal can be provided to cancel the retry loop immediately.
+func RunWithLinearBackoff(function func() error, attempts int, sleep time.Duration, cancelSignal <-chan struct{}) (err error) {
+	return runWithBackoffInternal(function, func(failCount int) time.Duration {
+		return calculateLinearDelay(failCount, sleep)
+	}, attempts, cancelSignal)
+}
+
+// RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExponentBase^(i-1) * sleep' duration before
+// each i-th attempt. An optional cancelSignal can be provided to cancel the retry loop immediately.
+func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64, cancelSignal <-chan struct{}) (err error) {
 	return runWithBackoffInternal(function, func(failCount int) time.Duration {
 		return calculateExpDelay(failCount, sleep, backoffExponentBase)
-	}, attempts)
+	}, attempts, cancelSignal)
 }
--- a/toolkit/tools/internal/retry/retry.go
+++ b/toolkit/tools/internal/retry/retry.go
@@ -8,12 +8,12 @@ import (
 	"time"
 )
 
-func calcDelay(retryCount int, sleep time.Duration, backoffExp float64) time.Duration {
+func calculateDelay(retryCount int, sleep time.Duration, backoffExponent float64) time.Duration {
 	// Calculate an exponential backoff.
 	// The formula is: sleep * (backoffExp ^ retryCount)
 	// For example, if sleep = 1 second, backoffExp = 2.0, retryCount = 3
 	// then the delay will be 1 * (2 ^ 3) = 8 seconds.
-	expRetry := math.Pow(float64(retryCount), backoffExp)
+	expRetry := math.Pow(backoffExp, float64(retryCount))
 	return time.Duration(expRetry * float64(sleep))
 }
 
@@ -22,7 +22,7 @@ func Run(function func() error, attempts int, sleep time.Duration) (err error) {
 	return RunWithExpBackoff(function, attempts, sleep, 1.0)
 }
 
-// Run runs function up to attempts times, waiting i * backoffMult * sleep duration before each i-th attempt.
+// RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExp^i * sleep' duration before each i-th attempt.
 func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExp float64) (err error) {
 	for i := 0; i < attempts; i++ {
 		// Calculate an exponential backoff.
--- a/toolkit/tools/internal/retry/retry.go
+++ b/toolkit/tools/internal/retry/retry.go
@@ -25,7 +25,7 @@ func Run(function func() error, attempts int, sleep time.Duration) (err error) {
 // RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExp^i * sleep' duration before each i-th attempt.
 func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExp float64) (err error) {
 	for i := 0; i < attempts; i++ {
-		// Calculate an exponential backoff.
+		// Calculate an exponential backoff. For i=0 we will calculate 0 delay and immediately call the function.
 		time.Sleep(calcDelay(i, sleep, backoffExp))
 		if err = function(); err == nil {
 			break
--- a/toolkit/tools/internal/retry/retry.go
+++ b/toolkit/tools/internal/retry/retry.go
@@ -9,37 +9,52 @@ import (
 )
 
 // calculateDelay calculates the delay for the given failure count, sleep duration, and backoff exponent base.
-// If the base is negative, it will calculate a linear backoff.
 // If the base is positive, it will calculate an exponential backoff.
-func calculateDelay(failCount int, sleep time.Duration, backoffExponentBase float64) time.Duration {
-	if failCount == 0 {
+func calculateExpDelay(failCount int, sleep time.Duration, backoffExponentBase float64) time.Duration {
+	if failCount <= 0 {
 		return 0
 	}
-	// Treat negative values of base as linear backoff.
-	if backoffExponentBase < 0 {
-		return sleep * time.Duration(failCount)
-	}
-	// Calculate an exponential backoff.
-	// The formula is: sleep * (backoffExpBase ^ failCount)
+	// Calculate an exponential backoff. We calculate and sleep BEFORE we try to run the function, so we need to
+	// subtract 1 from the failCount to get the correct exponent.
+
+	// The formula is: sleep * (backoffExpBase ^ (failCount-1))
 	// For example, if sleep = 1 second, backoffExp = 2.0, failCount = 3
-	// then the delay will be 1 * (2 ^ 3) = 8 seconds.
+	// then the delay will be 1 * (2 ^ 2) = 4 seconds.
+	// (0 sec on the first call, 1 sec on the second call, 2 sec on the third call, etc.)
 	expRetry := math.Pow(backoffExponentBase, float64(failCount-1))
 	return time.Duration(expRetry * float64(sleep))
 }
 
-// Run runs function up to attempts times, waiting i * sleep duration before each i-th attempt.
-func Run(function func() error, attempts int, sleep time.Duration) (err error) {
-	return RunWithExpBackoff(function, attempts, sleep, -1)
+func calculateLinearDelay(failCount int, sleep time.Duration) time.Duration {
+	if failCount <= 0 {
+		return 0
+	}
+	return sleep * time.Duration(failCount)
 }
 
-// RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExponentBase^i * sleep' duration before each i-th attempt.
-func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64) (err error) {
+// runWithBackoffInternal runs function up to 'attempts' times, waiting delayCalc(failCount) before each i-th attempt.
+// delayCalc(0) is expected to return 0.
+func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int) (err error) {
 	for failures := 0; failures < attempts; failures++ {
-		// Calculate an exponential backoff. For i=0 we will calculate 0 delay and immediately call the function.
-		time.Sleep(calculateDelay(failures, sleep, backoffExponentBase))
+		delayTime := delayCalc(failures)
+		time.Sleep(delayTime)
 		if err = function(); err == nil {
 			break
 		}
 	}
 	return err
+}
+
+// Run runs function up to 'attempts' times, waiting i * sleep duration before each i-th attempt.
+func Run(function func() error, attempts int, sleep time.Duration) (err error) {
+	return runWithBackoffInternal(function, func(failCount int) time.Duration {
+		return calculateLinearDelay(failCount, sleep)
+	}, attempts)
+}
+
+// RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExponentBase^(i-1) * sleep' duration before each i-th attempt.
+func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64) (err error) {
+	return runWithBackoffInternal(function, func(failCount int) time.Duration {
+		return calculateExpDelay(failCount, sleep, backoffExponentBase)
+	}, attempts)
 }
--- a/toolkit/tools/internal/retry/retry.go
+++ b/toolkit/tools/internal/retry/retry.go
@@ -4,13 +4,29 @@
 package retry
 
 import (
+	"math"
 	"time"
 )
 
+func calcDelay(retryCount int, sleep time.Duration, backoffExp float64) time.Duration {
+	// Calculate an exponential backoff.
+	// The formula is: sleep * (backoffExp ^ retryCount)
+	// For example, if sleep = 1 second, backoffExp = 2.0, retryCount = 3
+	// then the delay will be 1 * (2 ^ 3) = 8 seconds.
+	expRetry := math.Pow(float64(retryCount), backoffExp)
+	return time.Duration((expRetry * float64(sleep)))
+}
+
 // Run runs function up to attempts times, waiting i * sleep duration before each i-th attempt.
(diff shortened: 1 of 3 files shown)


toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,29 @@
...
 	"time"
 )
 
+func calcDelay(retryCount int, sleep time.Duration, backoffExp float64) time.Duration {
+	// Calculate an exponential backoff.
+	// The formula is: sleep * (backoffExp ^ retryCount)
+	// For example, if sleep = 1 second, backoffExp = 2.0, retryCount = 3
+	// then the delay will be 1 * (2 ^ 3) = 8 seconds.
+	expRetry := math.Pow(float64(retryCount), backoffExp)
+	return time.Duration((expRetry * float64(sleep)))
```
cwize1, line 17: nit: Remove extra set of `()`.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -112,6 +113,7 @@ var (
 	runCheck     = app.Flag("run-check", "Whether or not to run the spec file's check section during package build.").Bool()
 
 	workers          = app.Flag("workers", "Number of concurrent goroutines to parse with.").Default(defaultWorkerCount).Int()
+	concurrentNetOps = app.Flag("concurrent-net-ops", "Number of concurrent network operations to perform.").Default(defaultWorkerCount).Int()
```
cwize1, line 116: nit: `defaultNetOpsCount`?

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -572,19 +578,20 @@ func specsToPackWorker(requests <-chan string, results chan<- *specState, cancel
...
-func packSRPMs(specStates []*specState, distTag, buildDir string, templateSrcConfig sourceRetrievalConfiguration, workers int) (err error) {
+func packSRPMs(specStates []*specState, distTag, buildDir string, templateSrcConfig sourceRetrievalConfiguration, workers, concurrentNetOps int) (err error) {
 	tsRoot, _ := timestamp.StartEvent("packing SRPMs", nil)
 	defer timestamp.StopEvent(nil)
 	var wg sync.WaitGroup
 
 	allSpecStates := make(chan *specState, len(specStates))
 	results := make(chan *packResult, len(specStates))
 	cancel := make(chan struct{})
+	netOpsSemaphor := make(chan struct{}, concurrentNetOps)
```
cwize1, line 589: nit: `netOpsSemaphore`

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+	}, 3, time.Second, 1.0)
+	endTime := time.Now()
+	assert.Nil(t, err)
+	// Delays should be 0 seconds, <attempt>
+	// for a total of 0 seconds.
+	maxDelay := time.Millisecond * 100
+	assert.LessOrEqual(t, endTime.Sub(startTime), maxDelay)
+}
+
+func TestZeroMult(t *testing.T) {
```
cwize1, line 80: nit: Move up to be with the other functions that are testing `calcDelay`.

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+	startTime := time.Now()
+	err := RunWithExpBackoff(func() error {
+		return fmt.Errorf("test error")
+	}, 3, time.Millisecond*100, 1.0)
+	endTime := time.Now()
+	assert.NotNil(t, err)
+	// Delays should be 0 seconds, <attempt>, .1 second, <attempt>, .2 seconds, <attempt>
+	// for a total of .3 seconds.
+	minDelay := time.Millisecond * 300
+	maxDelay := time.Millisecond * 500
```
cwize1, line 47: It isn't immediately clear that this represents a 200ms buffer for the timer expiry.

Perhaps, something like:

```golang
expectedDelay := time.Millisecond * 300
delayBuffer := time.Millisecond * 200

assert.GreaterOrEqual(t, endTime.Sub(startTime), expectedDelay)
assert.LessOrEqual(t, endTime.Sub(startTime), expectedDelay + delayBuffer)
```

cwize1 (approved):

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -758,10 +765,17 @@ func packSingleSPEC(specFile, srpmFile, signaturesFile, buildDir, outDir, distTa
...
-	if err != nil {
-		return
+	{
+		// Limit the number of concurrent network operations by pushing a struct{} into the channel. This will block until
+		// another operation completes and removes the struct{} from the channel.
+		netOpsSemaphor <- struct{}{}
+		// Hydrate all sources. Download any missing ones not in `sourceDir`
+		err = hydrateFiles(fileTypeSource, specFile, workingDir, srcConfig, currentSignatures, defines)
+		if err != nil {
+			return
```
trungams, line 775: should also release semaphore if this function fails out

trungams (commented):
I may have misunderstood the problem. This change will help throttle the amount of workers using network connection, but suppose all current workers using network have failed and are waiting to retry (using the exp backoff formula), shouldn't other workers be allowed a chance to use network as well? In that case, I feel like putting the semaphore in the network access layer, e.g by controlling concurrent calls to `hydrateFromRemoteSource` would be better and easier to implement as well.

cwize1:
> I may have misunderstood the problem. This change will help throttle the amount of workers using network connection, but suppose all current workers using network have failed and are waiting to retry (using the exp backoff formula), shouldn't other workers be allowed a chance to use network as well? In that case, I feel like putting the semaphore in the network access layer, e.g by controlling concurrent calls to `hydrateFromRemoteSource` would be better and easier to implement as well.

This would depend on why the download was failing. If the failure was specific to the file, then only applying the exponential back-off to that file makes sense. But if the failure is related to the web server and/or network connection, then having the exponential back-off block other downloads would prevent spamming the server and/or network connection.

dmcilvaney:
> > I may have misunderstood the problem. This change will help throttle the amount of workers using network connection, but suppose all current workers using network have failed and are waiting to retry (using the exp backoff formula), shouldn't other workers be allowed a chance to use network as well? In that case, I feel like putting the semaphore in the network access layer, e.g by controlling concurrent calls to `hydrateFromRemoteSource` would be better and easier to implement as well.
> 
> This would depend on why the download was failing. If the failure was specific to the file, then only applying the exponential back-off to that file makes sense. But if the failure is related to the web server and/or network connection, then having the exponential back-off block other downloads would prevent spamming the server and/or network connection.

The primary failure case looks to be due to network throttling on the build agents. I am not concerned about a missing file or other catastrophic failure blocking builds since that will result in a failed build regardless (since any failed packing job will stop the build). 

So given that, it is intended that we block the other workers from progressing if we start seeing failures. Since the source server 'should' be 100% reliable I don't mind losing out on one of the network slots for a few minutes due to a misisng file, the other slots should be sufficient to still saturate a reasonable network connection.

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: hah, that actually defeats the purpose of this PR... good catch.
```suggestion
	concurrentNetOps = app.Flag("concurrent-net-ops", "Number of concurrent network operations to perform.").Default(defaultNetOpsCount).Int()
```

toolkit/tools/internal/retry/retry.go
dmcilvaney:
```suggestion
	return time.Duration(expRetry * float64(sleep))
```

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney:
```suggestion
			<-netOpsSemaphor
			return
```

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -943,7 +957,7 @@ func hydrateFromLocalSource(fileHydrationState map[string]bool, newSourceDir str
 // Will alter `currentSignatures`.
 func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir string, srcConfig sourceRetrievalConfiguration, skipSignatureHandling bool, currentSignatures map[string]string) {
 	const (
-		downloadRetryAttempts = 3
+		downloadRetryAttempts = 10
```
dmcilvaney, line 960:
```suggestion
		downloadRetryAttempts = 10
		failureBackoffExp = 2.0
```

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -956,14 +970,16 @@ func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir st
...
+		// take at most ~5 minutes before giving up.
+		err := retry.RunWithExpBackoff(func() error {
 			err := network.DownloadFile(url, destinationFile, srcConfig.caCerts, srcConfig.tlsCerts)
 			if err != nil {
 				logger.Log.Warnf("Failed to download (%s). Error: %s", url, err)
 			}
 
 			return err
-		}, downloadRetryAttempts, downloadRetryDuration)
+		}, downloadRetryAttempts, downloadRetryDuration, 2.0)
```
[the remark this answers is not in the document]
dmcilvaney:
```suggestion
		}, downloadRetryAttempts, downloadRetryDuration, failureBackoffExp)
```

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,29 @@
...
+	"math"
 	"time"
 )
 
+func calcDelay(retryCount int, sleep time.Duration, backoffExp float64) time.Duration {
+	// Calculate an exponential backoff.
+	// The formula is: sleep * (backoffExp ^ retryCount)
+	// For example, if sleep = 1 second, backoffExp = 2.0, retryCount = 3
+	// then the delay will be 1 * (2 ^ 3) = 8 seconds.
+	expRetry := math.Pow(float64(retryCount), backoffExp)
```
trungams, line 16: shouldn't this be 
```suggestion
	expRetry := math.Pow(backoffExp, float64(retryCount))
```

trungams (changes requested):

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,29 @@
...
 func Run(function func() error, attempts int, sleep time.Duration) (err error) {
+	return RunWithExpBackoff(function, attempts, sleep, 1.0)
+}
+
+// Run runs function up to attempts times, waiting i * backoffMult * sleep duration before each i-th attempt.
+func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExp float64) (err error) {
 	for i := 0; i < attempts; i++ {
-		time.Sleep(time.Duration(i) * sleep)
+		// Calculate an exponential backoff.
+		time.Sleep(calcDelay(i, sleep, backoffExp))
```
PawelWMS, line 29: For `i = 0` won't this cause us to always sleep for `sleep` amount of time instead of immediately calling `function`?

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+// Delay vs. retry count for mult = 1.0, delay = .1 seconds
+var expectedValuesMillis1 = []int{0, 100, 200, 300, 400, 500}
+
+// Delay vs. retry count for exp = 2.0, delay = 0.1 seconds
+var expectedValuesMillis2 = []int{0, 100, 400, 900, 1600, 2500}
+
+func TestCalcDelay1(t *testing.T) {
```
PawelWMS, line 17: Could we make the test names more descriptive? It's not too clear what `1` is about here.

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// Delay vs. retry count for mult = 1.0, delay = .1 seconds
+var expectedValuesMillis1 = []int{0, 100, 200, 300, 400, 500}
```
PawelWMS, line 12: Let's keep the expected results inside the tests, which use them - that way it's easier to find all of the test's parameters because they are in one place. In this case we're using `expectedValuesMillis1` only once, so we don't even benefit from keeping it a global.

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,29 @@
...
+	expRetry := math.Pow(float64(retryCount), backoffExp)
+	return time.Duration(expRetry * float64(sleep))
+}
+
 // Run runs function up to attempts times, waiting i * sleep duration before each i-th attempt.
 func Run(function func() error, attempts int, sleep time.Duration) (err error) {
+	return RunWithExpBackoff(function, attempts, sleep, 1.0)
+}
+
+// Run runs function up to attempts times, waiting i * backoffMult * sleep duration before each i-th attempt.
```
PawelWMS, line 25:
```suggestion
// RunWithExpBackoff runs function up to 'attempts' times, waiting backoffExp**i * sleep duration before each i-th attempt.
```

toolkit/tools/internal/retry/retry.go
PawelWMS: Also, I think in this case `backoffExponentBase` instead of `backoffExp` would help to avoid confusion.

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,29 @@
 package retry
 
 import (
+	"math"
 	"time"
 )
 
+func calcDelay(retryCount int, sleep time.Duration, backoffExp float64) time.Duration {
```
PawelWMS, line 11: Nit-pick: can we use full words? That's what we're trying to do across the code.
```suggestion
func calculateDelay(retryCount int, sleep time.Duration, backoffExponent float64) time.Duration {
```

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Delay vs. retry count for exp = 2.0, delay = 0.1 seconds
+var expectedValuesMillis2 = []int{0, 100, 400, 900, 1600, 2500}
+
+func TestCalcDelay1(t *testing.T) {
+	mult := 1.0
+	sleep := time.Millisecond * 100
+	for i, expected := range expectedValuesMillis1 {
+		expected := time.Duration(expected) * time.Millisecond
```
PawelWMS, line 21: Why not just put calculate values inside `expectedValuesMillis1`

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+		actual := calcDelay(i, sleep, mult)
+		assert.Equal(t, expected, actual)
+	}
```
PawelWMS, line 24: Could we use full words to make it clearer what the variables mean? Also, let's be consistent with the naming:
```suggestion
	backoffExponentBase := 1.0
	sleep := time.Millisecond * 100
	for i, expected := range expectedValuesMillis1 {
		expected := time.Duration(expected) * time.Millisecond
		actual := calcDelay(i, sleep, backoffExponentBase)
		assert.Equal(t, expected, actual)
	}
```

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,29 @@
...
+	return time.Duration(expRetry * float64(sleep))
+}
+
 // Run runs function up to attempts times, waiting i * sleep duration before each i-th attempt.
 func Run(function func() error, attempts int, sleep time.Duration) (err error) {
+	return RunWithExpBackoff(function, attempts, sleep, 1.0)
+}
+
+// Run runs function up to attempts times, waiting i * backoffMult * sleep duration before each i-th attempt.
+func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExp float64) (err error) {
```
PawelWMS, line 26: Should we add a check for `backoffExp < 0.0`?

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func TestTotalRunTimeWithFailures1(t *testing.T) {
+	startTime := time.Now()
+	err := RunWithExpBackoff(func() error {
+		return fmt.Errorf("test error")
+	}, 3, time.Millisecond*100, 1.0)
```
PawelWMS, line 41: Could we turn the "magic numbers" into variables for the sake of clarity?

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+	assert.NotNil(t, err)
+	// Delays should be 0 seconds, <attempt>, .1 second, <attempt>, .2 seconds, <attempt>
+	// for a total of .3 seconds.
+	minDelay := time.Millisecond * 300
+	maxDelay := time.Millisecond * 500
+	assert.GreaterOrEqual(t, endTime.Sub(startTime), minDelay)
+	assert.LessOrEqual(t, endTime.Sub(startTime), maxDelay)
+}
+
+func TestTotalRunTimeWithFailures2(t *testing.T) {
```
PawelWMS, line 52: Similar request for the test naming. It's unclear what `2` means here. Could we be more descriptive?

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+	endTime := time.Now()
+	assert.Nil(t, err)
+	// Delays should be 0 seconds, <attempt>
+	// for a total of 0 seconds.
+	maxDelay := time.Millisecond * 100
+	assert.LessOrEqual(t, endTime.Sub(startTime), maxDelay)
+}
+
+func TestZeroMult(t *testing.T) {
+	val := calcDelay(2, time.Second, 0.0)
```
PawelWMS, line 81: Can we use named variables/constants?

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -112,6 +113,7 @@ var (
 	runCheck     = app.Flag("run-check", "Whether or not to run the spec file's check section during package build.").Bool()
 
 	workers          = app.Flag("workers", "Number of concurrent goroutines to parse with.").Default(defaultWorkerCount).Int()
+	concurrentNetOps = app.Flag("concurrent-net-ops", "Number of concurrent network operations to perform.").Default(defaultNetOpsCount).Int()
```
PawelWMS, line 116: Should we expose it as a `make` variable? No strong opinions here.

PawelWMS, line 116: Would this allow us drop the check below?
```suggestion
	concurrentNetOps = app.Flag("concurrent-net-ops", "Number of concurrent network operations to perform.").Default(defaultNetOpsCount).UInt()
```

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -758,10 +765,18 @@ func packSingleSPEC(specFile, srpmFile, signaturesFile, buildDir, outDir, distTa
...
+		// another operation completes and removes the struct{} from the channel.
+		netOpsSemaphor <- struct{}{}
+		// Hydrate all sources. Download any missing ones not in `sourceDir`
+		err = hydrateFiles(fileTypeSource, specFile, workingDir, srcConfig, currentSignatures, defines)
+		if err != nil {
+			<-netOpsSemaphor
+			return
+		}
+		// Clear the channel to allow another operation to start
+		<-netOpsSemaphor
```
PawelWMS, line 779: Why the separate scope with `{}`?

PawelWMS, line 779: What do you think about moving acquiring the semaphore and releasing it into `hydrateFiles`? This way we can use `defer` to include the release logic only once.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -943,7 +958,8 @@ func hydrateFromLocalSource(fileHydrationState map[string]bool, newSourceDir str
 // Will alter `currentSignatures`.
 func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir string, srcConfig sourceRetrievalConfiguration, skipSignatureHandling bool, currentSignatures map[string]string) {
 	const (
-		downloadRetryAttempts = 3
+		downloadRetryAttempts = 10
+		failureBackoffExp = 2.0
```
PawelWMS, line 962: Suggestion to use full words and add clarity:
```suggestion
		failureBackoffExponentBase = 2.0
```

Also, I'm not sure we need the `failure` prefix, but I have no strong opinion on this one.

toolkit/tools/srpmpacker/srpmpacker.go
PawelWMS, line 589: Nit-pick: why `chan struct{}` instead of `chan int` or other simple and short type? It would make the code a bit shorter.

toolkit/tools/srpmpacker/srpmpacker.go
PawelWMS: I think we might even move it deeper into `hydrateFromRemoteSource` to block right before we call `network.DownloadFile`.

PawelWMS: Or maybe even move the whole limitation concept into the `network` module, so callers don't even have to think about it. Probably an overkill for now, but something to think about.

toolkit/tools/internal/retry/retry.go
PawelWMS:
```suggestion
		if i > 0 {
			time.Sleep(calcDelay(i, sleep, backoffExp))
		}
```

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: The examples i've seen online, and in our existing code, generally use an empty struct to indicate there is no expected data to pass via the channel.

cwize1: Also `struct{}` has a size of 0. So, in theory, the compiler could optimize around that. (Though I have no idea if it actually does.)

toolkit/tools/internal/retry/retry.go
dmcilvaney: This is the old implementation, the calculation always results in sleep of zero for the 1st pass. I'll add a comment to that affect (and I validated this with a new test case).

toolkit/tools/internal/retry/retry.go
dmcilvaney: Technically negative numbers are not invalid exponents, although maybe not meaningful for normal use (they asymptotically approach zero, ie contant delay with each retry)

toolkit/tools/internal/retry/retry.go
dmcilvaney: I think both work, but looking again that version does seem to be the more commonly used formula.

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: I think we probably won't see much benefit from tuning this, going to leave it default only for now but it is an easy change if we need to adjust it later.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -147,6 +149,10 @@ func main() {
 		logger.Log.Fatalf("Value in --workers must be greater than zero. Found %d", *workers)
 	}
 
+	if *concurrentNetOps <= 0 {
+		logger.Log.Fatalf("Value in --concurrent-net-ops must be greater than zero. Found %d", *concurrentNetOps)
+	}
```
dmcilvaney, line 154:
```suggestion
```

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,29 @@
...
+	"math"
 	"time"
 )
 
+func calculateDelay(retryCount int, sleep time.Duration, backoffExponent float64) time.Duration {
+	// Calculate an exponential backoff.
+	// The formula is: sleep * (backoffExp ^ retryCount)
+	// For example, if sleep = 1 second, backoffExp = 2.0, retryCount = 3
+	// then the delay will be 1 * (2 ^ 3) = 8 seconds.
+	expRetry := math.Pow(backoffExp, float64(retryCount))
```
cwize1, line 16: nit: `backoffExponent` (name wasn't updated)

dmcilvaney: Given the change in fuction, working to rewrite a bit of stuff. Change is included in that (I just accepted the GitHub suggestions for a bunch of these, so only part of it is available online right now)

toolkit/tools/internal/retry/retry.go
dmcilvaney: Reworked formulas, now has special casing for linear and failures=0

toolkit/tools/internal/retry/retry.go
dmcilvaney: Now treat negatives as linear backoff (ie legacy calculations for linear)

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: Left over from previous experiment with scoped variables, cleaned up.

dmcilvaney: Done

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,29 @@
...
 func Run(function func() error, attempts int, sleep time.Duration) (err error) {
+	return RunWithExpBackoff(function, attempts, sleep, 1.0)
+}
+
+// RunWithExpBackoff runs function up to 'attempts' times, waiting 'backoffExp^i * sleep' duration before each i-th attempt.
+func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExp float64) (err error) {
 	for i := 0; i < attempts; i++ {
-		time.Sleep(time.Duration(i) * sleep)
+		// Calculate an exponential backoff. For i=0 we will calculate 0 delay and immediately call the function.
+		time.Sleep(calcDelay(i, sleep, backoffExp))
```
jslobodzian, line 29: Dumb question.  isn't the name of the function "calculateDelay" above?  instead of "calcDelay"?

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -231,7 +233,7 @@ func parsePackListFile(packListFile string) (packList map[string]bool, err error
 
 // createAllSRPMsWrapper wraps createAllSRPMs to conditionally run it inside a chroot.
 // If workerTar is non-empty, packing will occur inside a chroot, otherwise it will run on the host system.
-func createAllSRPMsWrapper(specsDir, distTag, buildDir, outDir, workerTar string, workers int, nestedSourcesDir, repackAll, runCheck bool, packList map[string]bool, templateSrcConfig sourceRetrievalConfiguration) (err error) {
+func createAllSRPMsWrapper(specsDir, distTag, buildDir, outDir, workerTar string, workers, concurrentNetOps int, nestedSourcesDir, repackAll, runCheck bool, packList map[string]bool, templateSrcConfig sourceRetrievalConfiguration) (err error) {
```
PawelWMS, line 236: Why not `uint` now?
```suggestion
func createAllSRPMsWrapper(specsDir, distTag, buildDir, outDir, workerTar string, workers int, concurrentNetOps uint, nestedSourcesDir, repackAll, runCheck bool, packList map[string]bool, templateSrcConfig sourceRetrievalConfiguration) (err error) {
```

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,39 @@
...
+	}
+	// Treat negative values of base as linear backoff.
+	if backoffExponentBase < 0 {
+		return sleep * time.Duration(failCount)
+	}
+	// Calculate an exponential backoff.
+	// The formula is: sleep * (backoffExpBase ^ failCount)
+	// For example, if sleep = 1 second, backoffExp = 2.0, failCount = 3
+	// then the delay will be 1 * (2 ^ 3) = 8 seconds.
+	expRetry := math.Pow(backoffExponentBase, float64(failCount-1))
```
PawelWMS, line 26: The `failCount - 1` is inconsistent with the comment above. Let's get them in sync.

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,13 +4,39 @@
...
 
 import (
+	"math"
 	"time"
 )
 
+// calculateDelay calculates the delay for the given failure count, sleep duration, and backoff exponent base.
+// If the base is negative, it will calculate a linear backoff.
+// If the base is positive, it will calculate an exponential backoff.
+func calculateDelay(failCount int, sleep time.Duration, backoffExponentBase float64) time.Duration {
```
PawelWMS, line 14: After the recent changes, I think we ended up overloading the responsibilities of this function and made `RunWithExpBackoff` a bit inconsistent, since it will now do something else than its name suggests for negative bases. Considering that, I think we should prevent `RunWithExpBackoff` from accepting negative bases or make it work with negative exponent bases (a bit weird but consistent), or at least mention that different behaviour in the comment for `RunWithExpBackoff`.

One more idea that comes to my mind is to use a private
```golang
func runWithBackoff(function func() error, delay func(attempt int), attempts int) (err error) {
	for i := 0; i < attempts; i++ {
		delay(i)
		if err = function(); err == nil {
			break
		}
	}
	return err
}
```
function, which would then be called by the `Run*` ones with different `calculateDelay` implementations. For instance

```golang
func Run(function func() error, attempts int, sleep time.Duration) (err error) {
	return runWithBackoff(function, func(attempt int) {
		time.Sleep(time.Duration(attempt) * sleep)
	}, attempts)
}

func RunWithExpBackoff(function func() error, attempts int, sleep time.Duration, backoffExponentBase float64) (err error) {
	return runWithBackoff(function, func(attempt int) {
		if attempt == 0 {
			return
		}
		expRetry := math.Pow(backoffExponentBase, float64(attempt))
		sleepDuration := time.Duration(expRetry * float64(sleep))
		time.Sleep(sleepDuration)
	}, attempts)
}
```

dmcilvaney: Was thinking the same thing actually. This is getting complicated :P

toolkit/tools/internal/retry/retry.go
dmcilvaney: It was indeed, was halfway through accepting suggested changed then realized I had broken something... took a sec to fix and push it all back up.

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}, tries, time.Second, base)
+	endTime := time.Now()
+	assert.Nil(t, err)
+	// Delays should be 0 seconds, <attempt>
+	// for a total of 0 seconds.
+	maxDelay := timingFudgeFactor
+	assert.LessOrEqual(t, endTime.Sub(startTime), maxDelay)
+}
+
+// func TestFoo(t *testing.T) {
```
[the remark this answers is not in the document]
dmcilvaney: Yes, it should be gone now. Was using that to validate the test periods for the packer tool.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -969,16 +971,29 @@ func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir st
 
 		url := network.JoinURL(srcConfig.sourceURL, fileName)
 
-		// With 10 attempts, initial delay of 1 second, and a backoff factor of 2.0, this will
-		// take at most ~5 minutes before giving up.
+		// Limit the number of concurrent network operations by pushing a struct{} into the channel. This will block until
+		// another operation completes and removes the struct{} from the channel.
+		if netOpsSemaphore != nil {
+			netOpsSemaphore <- struct{}{}
```
cwize1, line 977: You should pass the cancellation signal into this function and then use a `select` statement here so that the function can exit early.

I suspect this will allow you to get rid of the `drainChannel` stuff.

cwize1: You might also want to strongly consider using the [context](https://pkg.go.dev/context) package and its associated coding patterns at some point. (But probably as a separate change.)

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -605,6 +611,8 @@ func packSRPMs(specStates []*specState, distTag, buildDir string, templateSrcCon
 			logger.Log.Errorf("Failed to pack (%s). Error: %s", result.specFile, result.err)
 			err = result.err
 			close(cancel)
+			// Drain the semaphore to ensure all workers can exit quickly
+			drainChannel(netOpsSemaphore)
```
cwize1, line 615: This feels a tad messy. And it isn't immediately clear how or why this works.

cwize1 (changes requested):
Left a comment on a possible better way to have all the works exit early.

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: Huh... that context stuff does look useful.
I wasn't able to get the cancel signal to work well in two spots, but I'll give it another go (since agreed, its a bit cleaner).

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var defaultTestTime = time.Millisecond * 100
+
+// Tests won't run with perfect timing, so we need to allow for some fudge factor when we check maximum duration.
+var timingFudgeFactor = time.Millisecond * 100
```
PawelWMS, line 14: Nit-pick:
```suggestion
const (
	defaultTestTime = time.Millisecond * 100

// Tests won't run with perfect timing, so we need to allow for some fudge factor when we check maximum duration.
	timingFudgeFactor = time.Millisecond * 100
)
```

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+func TestZeroBase(t *testing.T) {
+	// Two failures, 0.0 multiplier, should be 0 second delay
+	base := 0.0
+	fails := 2
+	val := calculateExpDelay(fails, defaultTestTime, base)
+	assert.Equal(t, time.Duration(0), val)
+}
+
+func TestZeroI(t *testing.T) {
```
PawelWMS, line 24: Could we make the test names more descriptive? Names like `ZeroBase` or `NegativeI` don't really say much.

toolkit/tools/internal/retry/retry_test.go
```diff
@@ -0,0 +1,133 @@
...
+	fails := 0
+	base := 1.0
+	val := calculateExpDelay(fails, defaultTestTime, base)
+	assert.Equal(t, time.Second*0, val)
+
+	// Three failures, 1.0 base, should still be 0 second delay
+	fails = 3
+	base = 0.0
+	val = calculateExpDelay(fails, defaultTestTime, base)
+	assert.Equal(t, time.Second*0, val)
```
PawelWMS, line 35: Ideally, I'd try to test only one scenario per test, so that it also can fail only for one reason. Bundling multiple scenarios into one test makes it take longer to find the one, which is failing.

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: Piped the cancel channel all the way through.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -956,14 +958,30 @@ func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir st
...
+		if netOpsSemaphore != nil {
+			select {
+			case netOpsSemaphore <- struct{}{}:
+			case <-cancel:
+				logger.Log.Debug("Cancellation signal received at network operation semaphore")
+				return
+			}
+		}
+
+		err := retry.RunWithExpBackoff(func() error {
```
cwize1, line 972: Ideally the cancellation token should be checked during the retry back-off.

It seems like you can use the `time.After()` function in a `select` statement: https://golangbyexample.com/select-statement-with-timeout-go/

cwize1 (approved):

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: So you are thinking an optional channel into the retry function? Maybe if its `nil` we make a temporary channel inside the function?
```go
func backoffSleep(delay time.Duration, cancelSignal <-chan struct{}) (cancelled bool)) {
	if cancelSignal == nil {
		cancelSignal = make(chan struct{})
	}
	select {
	case <-time.After(delay):
	case <-cancelSignal:
		cancelled = true
	}
	return
}

// runWithBackoffInternal runs function up to 'attempts' times, waiting delayCalc(failCount) before each i-th attempt.
// delayCalc(0) is expected to return 0.
func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int, cancelSignal <-chan struct{}) (err error) {
	for failures := 0; failures < attempts; failures++ {
		delayTime := delayCalc(failures)
		cancelled := backoffSleep(delayTime, cancelSignal)
		if cancelled {
			return fmt.Errorf("cancelled after %d attempts", failures)
		}
		if err = function(); err == nil {
			break
		}
	}
	return err
}
```

cwize1 (approved):

trungams (approved):

dmcilvaney:
FYI, I'd like to rebase and squash into two commits (no modifications). I'll do a non-squash merge with two commits: the retry logic, and the packer. Will make this more clear in the future looking back.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -881,7 +885,12 @@ func hydrateFiles(fileTypeToHydrate fileType, specFile, workingDir string, srcCo
 	}
 
 	if hydrateRemotely && srcConfig.sourceURL != "" {
-		hydrateFromRemoteSource(fileHydrationState, newSourceDir, srcConfig, skipSignatureHandling, currentSignatures, cancel, netOpsSemaphore)
+		err = hydrateFromRemoteSource(fileHydrationState, newSourceDir, srcConfig, skipSignatureHandling, currentSignatures, cancel, netOpsSemaphore)
+		// Most errors are non-fatal and are handled inside the function (encoded inside fileHydrationState), cancellation
+		// however will return and error that should cause an early exit.
```
PawelWMS, line 890: I'm not sure we want such comment outside the function itself. They tend to get stale easily after a function update, since they are hard to keep track of. I'd say we might need it - the main point is we return fast in case of an error without diving into what caused it.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -599,6 +602,7 @@ func packSRPMs(specStates []*specState, distTag, buildDir string, templateSrcCon
 		if result.err != nil {
 			logger.Log.Errorf("Failed to pack (%s). Error: %s", result.specFile, result.err)
 			err = result.err
+			logger.Log.Warn("Cancelling outstanding workers")
```
PawelWMS, line 605: Why not just include this message inside the error above since it's part of the same "event"? For instance:
`logger.Log.Errorf("Failed to pack (%s). Cancelling outstanding workers. Error: %s", result.specFile, result.err)`

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,16 +4,82 @@
...
+}
+
+// runWithBackoffInternal runs function up to 'attempts' times, waiting delayCalc(failCount) before each i-th attempt.
+// delayCalc(0) is expected to return 0.
+func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int, cancelSignal <-chan struct{}) (err error) {
+	for failures := 0; failures < attempts; failures++ {
+		delayTime := delayCalc(failures)
+		cancelled := backoffSleep(delayTime, cancelSignal)
+		if cancelled {
+			err = fmt.Errorf("%w after %d tries", ErrRetryCancelled, failures)
```
PawelWMS, line 56: I don't think we should classify this as an error at this level. The retry logic has no way of telling if the cancellation is due to something going wrong or it happened for any other reason, which might have been an expected part of the logic using the retry functions. If anything, I'd return an additional boolean like `retryCancelled bool`, which the caller then may check, if they need to.

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,16 +4,82 @@
...
+	select {
+	case <-time.After(delay):
+	case <-cancelSignal:
+		cancelled = true
+	}
+	return
+}
+
+// runWithBackoffInternal runs function up to 'attempts' times, waiting delayCalc(failCount) before each i-th attempt.
+// delayCalc(0) is expected to return 0.
```
PawelWMS, line 50:
```suggestion
// delayCalc(0) is expected to return 0.
// The function will return early if a signal has been sent to the "cancelSignal" channel.
```

toolkit/tools/internal/retry/retry.go
```diff
@@ -4,16 +4,82 @@
...
+func runWithBackoffInternal(function func() error, delayCalc func(failCount int) time.Duration, attempts int, cancelSignal <-chan struct{}) (err error) {
```
PawelWMS, line 51: Nit-pick: why `cancelSignal` and not `cancelChannel`?

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -96,6 +98,8 @@ type specState struct {
 	err      error
 }
 
+var errPackerCancelReceived = fmt.Errorf("packer cancel signal received")
```
PawelWMS, line 101: Can we move this inside `hydrateFromRemoteSource`? It seems that's the only place we use it.

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: Was getting lost in all the log spam, but that spam is fixed now so no reason not to unify.

toolkit/tools/internal/retry/retry.go
dmcilvaney: I'd say any result where the requested function couldn't complete is an error. `errors.Is()` makes checking if a specific error type is returned fairly simple (hence the public definition of the error).

toolkit/tools/internal/retry/retry.go
dmcilvaney: Calling it "channel" feels a bit redundant, and I wanted to indicate its expected use. I suppose "cancel" on its own is enough though.

dmcilvaney: Just renamed it to `cancel` which matches other bits of the code.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -989,12 +989,13 @@ func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir st
 			<-netOpsSemaphore
 		}
 
+		if cancelled {
+			err = errPackerCancelReceived
+			return
+		}
+
 		if err != nil {
```
PawelWMS, line 997: Nit-pick: I'd check for this first to be consistent with how we handle errors in other places in the code. The function's output should be considered unreliable when `err != nil`. Probably not a big deal here.

PawelWMS (approved):

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: Generally agree, but I think its critical we clear the semaphore regardless of return state here just to be as correct as possible. While technically we do offer an out via the cancel signal to avoid blocking on the semaphore I'd rather not leave it held unexpectedly.

PR #5939, opened 2023-08-04
Made `python-certifi` stop using its bundled certificates to fix CVE-2023-37920.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Fixing CVE-2023-37920. The CVE is about removing a CA from the bundled list of trusted anchors inside `python-certifi`. Since we don't want to use any other bundles except for the ones provided by `ca-certificates`, the change is about switching `python-certifi` to use the system certificates instead of its own bundle.

###### Change Log  

- Switched `python-certifi` to use the system trusted CAs instead of its own bundle.

No.

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2023-37920

###### Test Methodology

- Local package build and test.

### DIFF
--- a/SPECS/python-certifi/python-certifi.spec
+++ b/SPECS/python-certifi/python-certifi.spec
@@ -1,4 +1,5 @@
 %global certs_bundle_path %{_sysconfdir}/pki/tls/certs/ca-bundle.crt
+
 Summary:        Python package for providing Mozilla's CA Bundle
 Name:           python-certifi
 Version:        2023.05.07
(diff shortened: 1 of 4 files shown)


trungams (approved):

SPECS/python-certifi/python-certifi.spec
```diff
@@ -1,51 +1,67 @@
...
-%package -n     python3-certifi
-Summary:        Python package for providing Mozilla's CA Bundle
 BuildRequires:  python3-devel
-BuildRequires:  python3-setuptools
-%if %{with_check}
 BuildRequires:  python3-pip
+BuildRequires:  python3-wheel
+
+%if %{with_check}
+BuildRequires:  ca-certificates-base
```
jslobodzian, line 19: Consider making this an explicit requires with or without check.

PawelWMS: I was considering it, but I don't want to hard-code a dependency on a specific package `ca-certificates` package, as `ca-certificates` and `ca-certificates-base` (and any bundles we might thing of to provide in the future) are designed to be installable independently of each other. For instance you case save some space, if you install `ca-certificates` and remove `ca-certificates-base` without.

Having said that, I think we can work with a run-time dependency on the path to the certificates bundle itself.

PR #5960, opened 2023-08-08
Add hard check on go version

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add a hard error condition if you try to rebuild the tools with a version of go that is too old. The go compiler itself will report this error, but it is easy to lose that message in the output. Now we will immediately stop parsing the makefile if the version is wrong. We only check if `REBUILD_TOOLS=y` is set to 'y'.

###### Change Log  

- Add check for go compiler version if `REBUILD_TOOLS=y`

**NO**

###### Test Methodology

- Pipeline build id: local test
- Needs full pipeline run

### DIFF
--- a/toolkit/scripts/tools.mk
+++ b/toolkit/scripts/tools.mk
@@ -16,6 +16,7 @@ $(call create_folder,$(BUILD_DIR)/tools)
 # The version as held in the go.mod file (a line like 'go 1.19'). Add "go" to the front of the version number
 # so that it matches the output of 'go version' (e.g. 'go1.19').
 go_min_version = go$(shell grep -E '^go [0-9]+\.[0-9]+' $(TOOLS_DIR)/go.mod | awk '{print $$2}')
+
 # Check if the go version is high enough to build the tools. The 'sort' command is used to compare the versions
 # (with -V which sorts by version number). If the lowest version in the sort is the same as the minimum version, then
 # the installed version must be greater than or equal to the minimum version and we are fine.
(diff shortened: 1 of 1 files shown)


cwize1 (approved):

toolkit/scripts/tools.mk
```diff
@@ -13,6 +13,20 @@ $(call create_folder,$(BUILD_DIR)/tools)
 
 ######## GO TOOLS ########
 
+# The version as reported by 'go version'
+go_min_version = go1.19.0
```
PawelWMS, line 17: Let's take the Golang version from `toolkit/tools/go.mod` since that's our source of truth and this way we won't need to remember to update `tools.mk` every time we update the version of Golang.

PawelWMS (changes requested):

toolkit/scripts/tools.mk
```diff
@@ -13,6 +13,20 @@ $(call create_folder,$(BUILD_DIR)/tools)
 
 ######## GO TOOLS ########
 
+# The version as held in the go.mod file (a line like 'go 1.19'). Add "go" to the front of the version number
+# so that it matches the output of 'go version' (e.g. 'go1.19').
+go_min_version = go$(shell grep -E '^go [0-9]+\.[0-9]+' $(TOOLS_DIR)/go.mod | awk '{print $$2}')
```
PawelWMS, line 18: Small suggestion to prevent this line being lost in the surrounding comments:
```suggestion
go_min_version = go$(shell grep -E '^go [0-9]+\.[0-9]+' $(TOOLS_DIR)/go.mod | awk '{print $$2}')

```

PawelWMS (approved):

jslobodzian (approved):

PR #5982, opened 2023-08-10
Mariner Image Customizer boilerplate

cwize1:
###### Merge Checklist

- [x] Ready to merge

---

###### Summary 

This creates the boilerplate for Mariner Image Customizer (imgcustomizer) tool. In addition, it implements `AdditionalFiles` since that is one of the simplest features to implement.

###### Change Log

- Added feature to `chroot` to allow partitions to be mounted before the default partitions.
- Create imgcustomizer tool boilerplate.
- Implement `AdditionalFiles` in imgcustomizer.
- Added tests for:
  - Unmarshalling of `AdditionalFiles` JSON/YAML.
  - imgcustomizer's top level functions.
  - `AdditionalFiles` logic.

NO

###### Test Methodology

- Tests added.

### DIFF
--- a/toolkit/tools/imagecustomizer/main.go
+++ b/toolkit/tools/imagecustomizer/main.go
@@ -1,0 +1,64 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package main
+
+import (
+	"log"
+	"os"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/exe"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/timestamp"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/pkg/imagecustomizerlib"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/pkg/profile"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+	app = kingpin.New("imagecustomizer", "Customizes a pre-built CBL-Mariner image")
+
+	buildDir          = app.Flag("build-dir", "Directory to run build out of.").Required().String()
+	imageFile         = app.Flag("image-file", "Path of the base CBL-Mariner image which the customization will be applied to.").Required().String()
+	outputImageFile   = app.Flag("output-image-file", "Path to write the customized image to.").Required().String()
+	outputImageFormat = app.Flag("output-image-format", "Format of output image. Supported: vhd, vhdx, qcow2, raw.").Required().Enum("vhd", "vhdx", "qcow2", "raw")
+	configFile        = app.Flag("config-file", "Path of the image customization config file.").Required().String()
+	logFile           = exe.LogFileFlag(app)
+	logLevel          = exe.LogLevelFlag(app)
+	profFlags         = exe.SetupProfileFlags(app)
+	timestampFile     = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
+)
+
+func main() {
+	var err error
+
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+
+	logger.InitBestEffort(*logFile, *logLevel)
+
+	prof, err := profile.StartProfiling(profFlags)
+	if err != nil {
+		logger.Log.Warnf("Could not start profiling: %s", err)
+	}
+	defer prof.StopProfiler()
+
+	timestamp.BeginTiming("imagecustomizer", *timestampFile)
+	defer timestamp.CompleteTiming()
+
+	err = customizeImage()
+	if err != nil {
+		log.Fatalf("image customization failed: %v", err)
+	}
+}
+
+func customizeImage() error {
+	var err error
+
+	err = imagecustomizerlib.CustomizeImageWithConfigFile(*buildDir, *configFile, *imageFile,
+		*outputImageFile, *outputImageFormat)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/toolkit/tools/imagecustomizerapi/README.md
+++ b/toolkit/tools/imagecustomizerapi/README.md
@@ -1,0 +1,16 @@
+# Mariner Image Customizer API
+
+The Mariner image customizer (imgcustomizer) will be released as a standalone tool and
+will provide strong backwards compatibility guarantees (after the first official
+release).
+This is contrast to the Mariner toolkit's new image config, which isn't officially
+released and doesn't provide any backwards compatibility guarantees.
+
+While currently the new image config and imgcustomizer config are very similar, in the
+future there is the possibility they will diverge.
+
+## Known differences
+
+- For the new image config, `AdditionalFiles`' source files are relative to the working
+  directory.
+  Whereas, for imgcustomizer, the source files are relative to the config file.
--- a/toolkit/tools/imagecustomizerapi/fileconfig.go
+++ b/toolkit/tools/imagecustomizerapi/fileconfig.go
@@ -1,0 +1,109 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+// DestinationFileConfigList is a list of destination files where the source file will be copied to in the final image.
+// This type exists to allow a custom marshaller to be attached to it.
+type FileConfigList []FileConfig
+
+// FileConfig specifies options for how a file is copied in the target OS.
+type FileConfig struct {
+	// The file path in the target OS that the file will be copied to.
+	Path string `yaml:"Path"`
+
+	// The file permissions to set on the file.
+	Permissions *FilePermissions `yaml:"Permissions"`
+}
+
+var (
+	DefaultFileConfig = FileConfig{
+		Path:        "",
+		Permissions: nil,
+	}
+)
+
+func (l *FileConfigList) IsValid() (err error) {
+	if len(*l) <= 0 {
+		return fmt.Errorf("list is empty")
+	}
+
+	for i, fileConfig := range *l {
+		err = fileConfig.IsValid()
+		if err != nil {
+			return fmt.Errorf("invalid FileConfig at index %d: %w", i, err)
+		}
+	}
+
+	return nil
+}
+
+func (l *FileConfigList) UnmarshalYAML(value *yaml.Node) error {
+	var err error
+
+	// Try to parse as a single value.
+	var fileConfig FileConfig
+	err = value.Decode(&fileConfig)
+	if err == nil {
+		*l = FileConfigList{fileConfig}
+		return nil
+	}
+
+	// Try to parse as a list.
+	type IntermediateTypeFileConfigList FileConfigList
+	err = value.Decode((*IntermediateTypeFileConfigList)(l))
+	if err != nil {
+		return fmt.Errorf("failed to parse FileConfigList: %w", err)
+	}
+
+	return nil
+}
+
+func (f *FileConfig) IsValid() (err error) {
+	// Path
+	if f.Path == "" {
+		return fmt.Errorf("invalid Path value: empty string")
+	}
+
+	// Permissions
+	if f.Permissions != nil {
+		err = f.Permissions.IsValid()
+		if err != nil {
+			return fmt.Errorf("invalid Permissions value: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (f *FileConfig) UnmarshalYAML(value *yaml.Node) error {
+	var err error
+
+	if value.Kind == yaml.ScalarNode {
+		// Parse as a string.
+		*f = FileConfig{
+			Path:        value.Value,
+			Permissions: nil,
+		}
+		return nil
+	}
+
+	// Parse as a struct.
+	*f = DefaultFileConfig
+
+	type IntermediateTypeFileConfig FileConfig
+	err = value.Decode((*IntermediateTypeFileConfig)(f))
+	if err != nil {
+		return fmt.Errorf("failed to parse FileConfig: %w", err)
+	}
+
+	return nil
+}
--- a/toolkit/tools/imagecustomizerapi/systemconfig.go
+++ b/toolkit/tools/imagecustomizerapi/systemconfig.go
@@ -1,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+)
+
+// SystemConfig defines how each system present on the image is supposed to be configured.
+type SystemConfig struct {
+	AdditionalFiles map[string]FileConfigList `yaml:"AdditionalFiles"`
+}
+
+func (s *SystemConfig) IsValid() error {
+	var err error
+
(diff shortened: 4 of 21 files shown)


toolkit/tools/imagecustomizerapi/README.md
```diff
@@ -0,0 +1,16 @@
+# Mariner Image Customizer API
```
mfrw, line 1: I think, we should probably keep all the libraries in `pkg` directory.
The reason, I request this is because, I have proposed that we have a distinction between `main` packages and `lib` packages.

Reference: 
- https://github.com/microsoft/CBL-Mariner/pull/2983

toolkit/tools/imagecustomizerapi/systemconfig.go
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+)
+
+// SystemConfig defines how each system present on the image is supposed to be configured.
+type SystemConfig struct {
```
jiria, line 11: Is this just the initial iteration and the plan is to add the remaining attributes in subsequent PRs?

toolkit/tools/imagecustomizerlib/imagecustomizer.go
```diff
@@ -0,0 +1,128 @@
...
+	err = doCustomizations(baseConfigPath, config, imageChroot)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func findPartitions(diskDevice string) ([]string, []*safechroot.MountPoint, error) {
+	newMountDirectories := []string{}
```
jiria, line 119: Is the plan to use this later?

toolkit/tools/imagecustomizerlib/customizeutils.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"io/fs"
+	"path/filepath"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safechroot"
+)
+
+func doCustomizations(baseConfigPath string, config *imagecustomizerapi.SystemConfig, imageChroot *safechroot.Chroot) error {
```
jiria, line 14: Would like to open a discussion on a possibility of Trident calling into this function more directly. Trident will be also setting up chroot during image deployment and could take advantage of this function. We can of course create a more suitable wrapper.

toolkit/tools/imagecustomizer/main.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/imagecustomizerlib"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/exe"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+	app = kingpin.New("imagecustomizer", "Customizes a pre-built CBL-Mariner image")
+
+	buildDir          = app.Flag("build-dir", "Directory to run build out of.").Required().String()
```
dmcilvaney, line 19: We haven't been super consistent in our older tools, but kingpin has things like `ExistingFile()` which can simplify stuff later in the tool, depending on how you want the args to be used.

toolkit/tools/imagecustomizer/main.go
```diff
@@ -0,0 +1,49 @@
...
+	imageFile         = app.Flag("image-file", "Path of the base CBL-Mariner image which the customization will be applied to.").Required().String()
+	outputImageFile   = app.Flag("output-image-file", "Path to write the customized image to.").Required().String()
+	outputImageFormat = app.Flag("output-image-format", "Format of output image. Supported: vhd, vhdx, qcow2, raw.").Required().Enum("vhd", "vhdx", "qcow2", "raw")
+	configFile        = app.Flag("config-file", "Path of the image customization config file.").Required().String()
+	logFile           = exe.LogFileFlag(app)
+	logLevel          = exe.LogLevelFlag(app)
+)
+
+func main() {
+	kingpin.MustParse(app.Parse(os.Args[1:]))
```
dmcilvaney, line 29: nit: I think we are hoping to have profiling and timestamping as part of our core set of boilerplate bits. Not critical here but something to keep in mind for later versions. (Actually, now that I look I see `boilerplate.go` doesn't actually have the profiling stuff. I'll see about adding that later)

dmcilvaney: From the new precaher.go tool I wrote:
```golang
	profFlags     = exe.SetupProfileFlags(app)
	timestampFile = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
```
```golang
	prof, err := profile.StartProfiling(profFlags)
	if err != nil {
		logger.Log.Warnf("Could not start profiling: %s", err)
	}
	defer prof.StopProfiler()

	timestamp.BeginTiming("precacher", *timestampFile)
	defer timestamp.CompleteTiming()

```

toolkit/tools/imagecustomizerapi/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+The Mariner image customizer (imgcustomizer) will be released as a standalone tool and
+will provide strong backwards compatibility guarantees (after the first official
+release).
+This is contrast to the Mariner toolkit's new image config, which isn't officially
+released and doesn't provide any backwards compatibility guarantees.
+
+While currently the new image config and imgcustomizer config are very similar, in the
+future there is the possibility they will diverge.
```
dmcilvaney, line 10: Few of us had a idle thought about splitting the existing imager tool in two. What are your thoughts on taking the "customization" parts of the imager tool out, and using this tool as an extra step even for offline buidls? (Or at least the same libraries)
That would couple this tool more closely with the image config.json format, although which direction that coupling happens in isn't clear to me.

toolkit/tools/imagecustomizerapi/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Known differences
+
+- For the new image config, `AdditionalFiles`' source files are relative to the working
```
dmcilvaney, line 14: Are both of these saying the same thing? Do you mean "Wheras, for the old imager tool the source files are realtive to the config file."?

toolkit/tools/imagecustomizerapi/fileconfig.go
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) Microsoft Corporation.
```
dmcilvaney, line 1: In a lot of ways yaml is better than json for this, but I'm leery of splitting our configs like this. (I'm not opposed to looking at moving to yaml for the other configs in the future, maybe with a translator?).

I'm just worried that we will force outselves to have no code-reuse here (maybe a good thing?)

toolkit/tools/imagecustomizerapi/systemconfig.go
```diff
@@ -0,0 +1,26 @@
...
+	AdditionalFiles map[string]FileConfigList `yaml:"AdditionalFiles"`
+}
+
+func (s *SystemConfig) IsValid() error {
+	var err error
+
+	for sourcePath, fileConfigList := range s.AdditionalFiles {
```
dmcilvaney, line 18: No strong oppinions, but does it make sense to validate the source file exists for each map entry? I think it would have to happen here since the `FileConfgList` type won't have access to the `src` here.

toolkit/tools/imagecustomizerlib/customizeutils.go
dmcilvaney: Had another comment above, but maybe in the future we can hook into these functions as part of the normal image build flow as well.

toolkit/tools/imagecustomizerlib/imagecustomizer.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// TODO: Dynamically find partitions instead of hardcoding the mappings.
```
dmcilvaney, line 121: Any idea on how you want to approach this?

toolkit/tools/imagecustomizerapi/systemconfig.go
cwize1: Yes. I am intentionally not adding the APIs until they are implemented. This makes it very clear what APIs are usable and which aren't.

(I have seen other projects that preemptively added all the APIs, only for a ton of them to remain unimplemented indefinitely.)

toolkit/tools/imagecustomizerlib/imagecustomizer.go
cwize1: Yes, I have written a [change](https://github.com/cwize1/CBL-Mariner/blob/1017770c0ff566dbde31eba43be012ba57a6db75/toolkit/tools/imagecustomizerlib/imagecustomizer.go#L125C9-L125C9) for this.

I temporarily mount the EFI system partition, read the `grub.cfg`, and pull out the UUID of the OS partition. (The logic currently doesn't support layouts that have an EFS, boot, and OS partition. But I can add support for that later.)

toolkit/tools/imagecustomizerlib/imagecustomizer.go
cwize1: This is a value needed by `safechroot.Initialize()`, along with `mountPoints`. I don't need to items to this list at the moment. But it is here in case it is ever needed.

toolkit/tools/imagecustomizerapi/systemconfig.go
cwize1: The `IsValid()` function doesn't take any parameters. So, it doesn't know what directory the `sourcePath` is relative to. So, it can't verify if the file exists or not.

I guess I consider `IsValid()` to be more of schema checking than full validity checking. In a future change, I have added an additional config validation to the start of the customization process. I think that would be a more appropriate place for file existence checks.

toolkit/tools/imagecustomizerapi/fileconfig.go
cwize1: Yeah, the split was intentional, regardless of JSON vs. YAML. Image creation and image customization are different procedures. So, there is a good chance they might need to diverge in the future.

toolkit/tools/imagecustomizerapi/README.md
cwize1: My understanding is that for the image creation tool, the paths within the config file are relative to the working directory. (Though this isn't explicitly in the code. So, I could be wrong about this.) That being said, most of the time, when the tool runs, the working directory is set to the config file parent directory.

Whereas, for imgcustomizer, I want the paths to always be relative to the config file's parent directory.

toolkit/tools/imagecustomizerapi/README.md
cwize1: I am not sure I am fond of that code layout. Having Go directories with the same name (e.g. `specreader` and `pkg/specreader`) can make writing Go code kind of annoying as it forces you to rename one of the packages in the imports.

I also want to avoid the case where the package name is something overly generic. Hence, why I have `imagecustomizerapi` and not `imagecustomizer/api`.

toolkit/tools/imagecustomizerlib/imagecustomizer.go
fintelia: Has there been talk of using [discoverable partition types](https://uapi-group.org/specifications/specs/discoverable_partitions_specification/)? That would avoid having to parse grub.cfg or /etc/fstab

cwize1: Does anyone actually use the "Discoverable Partitions Specification". Or is it just another instance of [xkcd 927](https://xkcd.com/927)?

fintelia: A bunch of the systemd utilities integrate with them. I think many of the proposals around unified kernel images (UKIs) also rely on them so that the root partition can be inferred from the disk, rather than passed in as a kernel cmdline argument. Not sure whether there's been much/any adoption by other distros at this point

toolkit/tools/imagecustomizerapi/systemconfig.go
cwize1: I've added additional validity checks in `imagecustomizer.go`.

toolkit/tools/imagecustomizer/main.go
cwize1: Added.

toolkit/tools/imagecustomizer/main.go
cwize1: The current code actually creates the `buildDir` directory if it doesn't exist.

toolkit/tools/imagecustomizerlib/imagecustomizer.go
cwize1: If Mariner ever supports this scheme, we can probably add support for it. Though we would probably still fall back to reading the grub.cfg and fstab files in case the image that doesn't support it.

toolkit/tools/imagecustomizerapi/README.md
cwize1: I moved `imagecustomizerlib` into `pkg` but kept `imagecustomizer` and `imagecustomizerapi` here.

dmcilvaney: Only request is that we track any "internal" code via something like we do for the imager
``` makefile
go_internal_files = $(shell find $(TOOLS_DIR)/internal/ -type f -name '*.go')
go_pkg_files = $(shell find $(TOOLS_DIR)/pkg/ -type f -name '*.go')
go_imagegen_files = $(shell find $(TOOLS_DIR)/imagegen/ -type f -name '*.go')
go_scheduler_files = $(shell find $(TOOLS_DIR)/scheduler -type f -name '*.go')
go_common_files = $(go_module_files) $(go_internal_files) $(go_imagegen_files) $(go_pkg_files) $(go_scheduler_files) $(STATUS_FLAGS_DIR)/got_go_deps.flag $(BUILD_DIR)/tools/internal.test_coverage
```

Its not particularly elegant but it makes sure we rebuild all the go tools if any of the shared libraries change.

toolkit/tools/imagecustomizerlib/imagecustomizer.go
dmcilvaney: @hbeberman FYI

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -0,0 +1,171 @@
...
+	// Validate config.
+	err = validateConfig(baseConfigPath, config)
+	if err != nil {
+		return fmt.Errorf("invalid image config: %w", err)
+	}
+
+	// Convert image file to raw format, so that a kernel loop device can be used to make changes to the image.
+	buildImageFile := filepath.Join(buildDirAbs, "image.raw")
+
+	_, _, err = shell.Execute("qemu-img", "convert", "-O", "raw", imageFile, buildImageFile)
```
romoh, line 71: Do we need to check the current image format to see if this step is necessary? or do we only accept vhd as input?

romoh (approved):
LGTM.

You may want to consider adding a `--version` option to the boilerplate.

dmcilvaney (approved):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: If the file format is the same, then `qemu-img convert` will just copy the file. This is actually what I want to ensure that the original file remains unmodified.

PR #5988, opened 2023-08-11
Added tool name to printed logs.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Adding a `[<tool_name>]` element to our logs to make it easier to tell, which tool generated which log. We're expecting to be able to run multiple Go tools in parallel soon, so this should make it easier to distinguish them from each other.

Sample output:
image

###### Change Log  

- Added a gray `[<tool_name>]` prefix to our logs.

No.

###### Test Methodology

- Local tests.
- PR tooling unit tests.

### DIFF
--- a/toolkit/tools/internal/logger/log.go
+++ b/toolkit/tools/internal/logger/log.go
@@ -8,7 +8,6 @@ package logger
 import (
 	"bufio"
 	"io"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -81,8 +80,7 @@ func InitStderrLog() {
 		log.Panic("Failed to get caller info.")
 	}
 
-	toolName := strings.TrimSuffix(filepath.Base(callerFilePath), ".go")
-	initStderrLogInternal(toolName)
+	initStderrLogInternal(callerFilePath)
 }
 
 // SetFileLogLevel sets the lowest log level for file output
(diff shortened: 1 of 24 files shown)


toolkit/tools/internal/logger/log.go
```diff
@@ -173,7 +181,20 @@ func ReplaceStderrFormatter(newFormatter log.Formatter) (oldFormatter log.Format
...
+	const useColors = true
+
+	Log = log.New()
+	Log.ReportCaller = true
+
+	// By default send all log messages through stderrHook
+	stderrHook = newWriterHook(os.Stderr, defaultStderrLogLevel, useColors, toolName)
+	Log.AddHook(stderrHook)
+	Log.SetLevel(defaultStderrLogLevel)
+	Log.SetOutput(ioutil.Discard)
```
cwize1, line 194: nit: `io.Discard` (`ioutil.Discard` is deprecated.)

toolkit/tools/internal/logger/log.go
```diff
@@ -99,10 +101,16 @@ func InitBestEffort(path string, level string) {
 		level = defaultStderrLogLevel.String()
 	}
 
-	InitStderrLog()
+	_, callerFilePath, _, ok := runtime.Caller(parentCallerLevel)
```
cwize1, line 104: Should this code be moved into `initStderrLogInternal` to avoid duplicating it? (I think if you set `parentCallerLevel` to `2` this should work.)

cwize1 (approved):

dmcilvaney (approved):
Agree with @cwizel1 (i've been cleaning up ioutil as I find it as well), but otherwise looks good.

toolkit/tools/internal/logger/log.go
PawelWMS: Ah, right, I can move the logic for the `toolName` variable.

I intentionally repeat the `runtime.Caller` code, because I want that call to be present only inside the public functions, to have a guarantee we're getting the file path of the actual use of the logger. Moving it deeper makes the code easier to break during refactoring or extension in the future.

PR #5992, opened 2023-08-14
Update ruby default uri to 0.12.2 and bundled uri to 0.10.3 and fix CVE-2023-36617

Redent0r:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Patch ruby-uri version and fix CVE-2023-36617 (vulnerability not present in rub uri 12.2) https://github.com/hsbt/ruby/commit/ab5421547c5546603c2383085005272ba7754fea

###### Change Log  

- CVE-2023-36617

**NO**

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2023-36617

###### Test Methodology

- Unified build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=408103&view=results

### DIFF
--- a/SPECS/ruby/ruby.spec
+++ b/SPECS/ruby/ruby.spec
@@ -71,7 +71,7 @@
 %global tmpdir_version          0.1.2
 %global tsort_version           0.1.0
 %global un_version              0.2.0
-%global uri_version             0.11.1
+%global uri_version             0.12.2
 %global weakref_version         0.1.1
 %global win32ole_version        1.8.8
 %global yaml_version            0.2.0
@@ -83,7 +83,7 @@ Name:           ruby
 # provides should be versioned according to the ruby version.
 # More info: https://stdgems.org/
 Version:        3.1.4
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        (Ruby OR BSD) AND Public Domain AND MIT AND CC0 AND zlib AND UCD
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -97,6 +97,7 @@ Source4:        rubygems.con
 Source5:        rubygems.prov
 Source6:        rubygems.req
 Source7:        macros.rubygems
+Patch0:         CVE-2023-36617.patch
 BuildRequires:  openssl-devel
 BuildRequires:  readline
 BuildRequires:  readline-devel
@@ -399,6 +400,9 @@ sudo -u test make test TESTS="-v"
 %{_rpmconfigdir}/rubygems.con
 
 %changelog
+* Mon Aug 14 2023 Saul Paredes <saulparedes@microsoft.com> - 3.1.4-2
+- Patch CVE-2023-36617
+
 * Wed May 03 2023 Rakshaa Viswanathan <rviswanathan@microsoft.com> - 3.1.4-1
 - Upgrade ruby to 3.1.4
 - Update time_version to v0.2.2 to resolve CVE-2023-28756
--- a/SPECS/ruby/ruby.spec
+++ b/SPECS/ruby/ruby.spec
@@ -97,7 +97,7 @@ Source4:        rubygems.con
 Source5:        rubygems.prov
 Source6:        rubygems.req
 Source7:        macros.rubygems
-# Updates ruby-uri from 0.12.1 to 0.12.2. Remove once ruby gets updated to a version that comes with ruby-uri 0.12.2
+# Updates default ruby-uri from 0.12.1 to 0.12.2 and patches vendored one. Remove once ruby gets updated to a version that comes with ruby-uri 0.12.2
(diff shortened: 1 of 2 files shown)


SPECS/ruby/ruby.spec
```diff
@@ -97,6 +97,7 @@ Source4:        rubygems.con
 Source5:        rubygems.prov
 Source6:        rubygems.req
 Source7:        macros.rubygems
+Patch0:         CVE-2023-36617.patch
```
neha170, line 100: Could you add a note here why this patch was needed, and if needs to be modified when ruby/rubygem-uri is updated in the future, thanks

Redent0r: Added comment to patch and SPEC file. Thanks!

neha170 (commented):
From what I understand, the vulnerability is fixed in uri v0.12.2 and above. We are currently at 0.11.1. So you have updated to 0.12.2.
As per the list of gems for ruby 3.1.4, [Standard Gems 3.1.4 (stdgems.org)](https://stdgems.org/3.1.4/), ruby 3.1.4 comes with uri 0.12.1. This also matches with lib/uri/version.rb file. So, the source code for uri in this ruby is v0.12.1. We can change the version to any number we want, but the source code will still be v0.12.1. As per my understanding, this is not really getting rid of the vulnerability in 0.12.1. It is simply updating the version to another number artificially.
I think the right approach would be to either 1) upgrade ruby to a version that comes with higher source code version of uri (0.12.2 or above), or 2) make sure that the vulnerable code is patched in 0.12.1

neha170 (approved):
Thanks Saul, the new patch looks good. Please merge once it builds cleanly.
For reference, I looked at the changes between 12.1 and 12.2 here https://github.com/ruby/uri/compare/v0-12 They match with your patch, and also resonate with your reasoning above.

Redent0r:
\*sigh\* wait, in 3.1.4 there's https://github.com/ruby/ruby/blob/v3_1_4/lib/bundler/vendor/uri/lib/uri/version.rb#L3 (lib/bundler/vendor/uri/lib/uri/, version 0.10.2) and https://github.com/ruby/ruby/blob/v3_1_4/lib/uri/version.rb#L3 (lib/uri/, version 0.12.1). https://github.com/hsbt/ruby/commit/c7af39ecd8d78d9ffc312dcde2af3466c7bbaf5f patches the bundler version, but based on standard gems for 3.1.4, I think we ought to fix the one under lib/uri, since its version matches the default gem version.
image

To add salt to the wound, there doesn't seem to be a straight commit/patch to 0.12.2 for lib/uri. I *think* what we want it's a mixture of:
* https://github.com/ruby/uri/commit/eaf89cc31619d49e67c64d0b58ea9dc38892d175 (fixes lib/uri/rfc3986_parser.rb)
* https://github.com/ruby/ruby/commit/1eff362492f4a0ca8d6d036c958afa7961e827f0 (fixes lib/uri/rfc2396_parser.rb)
* https://github.com/ruby/ruby/commit/ab5421547c5546603c2383085005272ba7754fea (bumps up the version)

I can make a single patch out of that. These changes also align exactly with https://github.com/ruby/uri/compare/v0-12

Redent0r:
Alrighty, I ended up applying it in both places to be sure. I bumped the lib version from 0.12.1 to 0.12.2, but left the vendored version the same, since [its currently set at 0.10.2 ](https://github.com/ruby/ruby/blob/v3_1_4/lib/bundler/vendor/uri/lib/uri/version.rb#L3). It didn't felt right to change it to 0.12.2 since it may be missing other things. 

BB: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=408552&view=results [Pass]

neha170 (commented):
The lib/ changes look good.
I think we can bump up the version to 10.0.3 in bundler. As per the CVE page, the vulnerability is fixed in 10.0.3. Please verify that the patch to update (and fix vulnerability) to 10.0.3 is the same as the one you've used. Thoughts?

Redent0r:
Yep! If we ignore the test changes, it's exactly the same changes that fix the CVE https://github.com/ruby/uri/compare/v0.10.2...v0.10.3, namely https://github.com/ruby/uri/compare/v0.10.2...v0.10.3#diff-6792cbb51ee9c0c764dc71cdbb424a70fc55849b1d470d80688c352aba088a8d and https://github.com/ruby/uri/compare/v0.10.2...v0.10.3#diff-2fff240f8e962830911219fe98478063f84563f498ac32d327f73e5b357e6725

Excellent! I'll bump the bundler version on the patch to 0.10.3

Redent0r:
New BB: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=409089&view=results

neha170 (approved):
Looks good, thanks for making the changes

PR #5993, opened 2023-08-14
Limit cascading rebuilds in the scheduler

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When the scheduler encounters a package, it must rebuild it will also mark all the downstream packages that depend on it as needing to rebuild. This is the most "correct" approach where we guarantee any possible downstream changes are accounted for.
Unfortunately, this is incredibly time-consuming and gives us very little meaningful confidence beyond the first extra layer of dependencies. 

Here the scheduler is enlightened to track if packages are rebuilt, and also track an attenuating influence on downstream packages. A package is "fresh" if it is rebuilt due to missing files, or user request (i.e., `PACKAGE_REBUILD_LIST=<pkg>`, or `pkg.rpm` isn't present). Each subsequent layer of the build graph is less fresh that this package, and once we reach zero, we stop requiring rebuilds and go back to allowing the cached versions of these packages to be used.

The attenuation happens only on ( * -> run ) edges, i.e., on all edges from any type of node to on a run node. This captures both runtime and build time requirements correctly. A run node and its associated build node will have the same freshness, while meta nodes will not cause the freshness to attenuate.

Where a package has a runtime-only dependency we DO track the freshness, but we will not cause it to rebuild if it would not have rebuilt due to a build dependency (the package should be expected to function as is regardless of the version of the runtime dependency). Any testing of these packages will need to occur after the build stage. The intent of this change is to better validate build success and run the package tests (`%check`).

A cascading rebuild level of `-1` is treated as unbounded rebuilding which is the current behavior. We set the max freshness to `math.MaxInt`. This means even with 32bit int we would still need to build several billion packages before we exhaust the freshness of our initial nodes, effectively making the cascading rebuilds unbounded.

###### Change Log  

- Add `MAX_CASCADING_REBUILDS` flag (default = -1), `QUICK_REBUILD_PACKAGES=y` -> `MAX_CASCADING_REBUILDS=1`
- Scheduler's `GraphBuildState`, `nodeState` now track the maximum allowable "freshness" for any node, and the actual freshness of each node.
- Refactor `preparerequest.go`'s `ConvertNodesToRequests()` to use "freshness" of nodes to decide if a build node must be rebuilt rather than a binary check if a dependency was cached or not
  - The old check is roughly analogous to `if maxFreshness > 0 then useCache = false`
  - We take the freshest dependnecy as our new node's freshness and attenuate by `1` if its a run node edge.
- If the build worker decides it must rebuild, it updates the freshness of the resulting nodes to the maximum allowable freshness, and we restart the attenuation process again

**NO**

###### Test Methodology

- Local builds
- [Pipeline](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=408117&view=results)
- [Pipeline DELTA](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=408118&view=results)


dmcilvaney:
Build with most recent commit before my changes, also with DELTA: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=408608&view=results

toolkit/tools/scheduler/scheduler.go
```diff
@@ -115,6 +117,11 @@ func main() {
 		logger.Log.Fatalf("Value in --build-attempts must be greater than zero. Found %d.", *buildAttempts)
 	}
 
+	if *maxCascadingRebuilds < 0 {
+		logger.Log.Infof("--max-cascading-rebuilds (%d) is negative, setting to unbounded cascading rebuilds.", *maxCascadingRebuilds)
```
PawelWMS, line 121: Since the default behaviour is unbounded, printing this on info level might be confusing for the users. Especially since they don't necessarily are aware of any argument names passed to single Go tools but rely on the `make` arguments instead. I'd suggest moving this to debug level.

PawelWMS, line 121: Since the default behaviour is unbounded, printing this on info level might be confusing for the users. Especially since they don't necessarily are aware of any argument names passed to single Go tools but rely on the `make` arguments instead. I'd suggest moving this to debug level.

```suggestion
		logger.Log.Debugf("--max-cascading-rebuilds (%d) is negative, setting to unbounded cascading rebuilds.", *maxCascadingRebuilds)
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -33,24 +33,32 @@ type BuildChannels struct {
...
+	Node              *pkggraph.PkgNode
+	PkgGraph          *pkggraph.PkgGraph
+	AncillaryNodes    []*pkggraph.PkgNode
+	CanUseCache       bool
+	IsDelta           bool
+	ExpectedFreshness int
 }
 
 // BuildResult represents the results of a build agent trying to build a given node.
 type BuildResult struct {
```
PawelWMS, line 52: Do we want to add comments similar to what you did for `BuildRequest`?

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -99,14 +107,22 @@ func BuildNodeWorker(channels *BuildChannels, agent buildagents.BuildAgent, grap
...
 		}
 
 		switch req.Node.Type {
 		case pkggraph.TypeLocalBuild:
-			res.UsedCache, res.Skipped, res.BuiltFiles, res.LogFile, res.Err = buildBuildNode(req.Node, req.PkgGraph, graphMutex, agent, req.CanUseCache, buildAttempts, checkAttempts, ignoredPackages)
+			var hadMissingFiles bool
+			res.UsedCache, res.Skipped, hadMissingFiles, res.BuiltFiles, res.LogFile, res.Err = buildBuildNode(req.Node, req.PkgGraph, graphMutex, agent, req.CanUseCache, buildAttempts, checkAttempts, ignoredPackages)
+			if hadMissingFiles {
+				res.ActualFreshness = NodeFreshnessRebuildRequired
+				logger.Log.Debugf("Resetting freshness to %d for %s", res.ActualFreshness, res.Node.FriendlyName())
```
PawelWMS, line 122: Since `NodeFreshnessRebuildRequired` is a special-case constant, I think we might get more info about explicitly saying why we're resetting the freshness instead of printing the value.

```suggestion
				logger.Log.Debugf("Resetting freshness for %s due to missing files.", res.ActualFreshness, res.Node.FriendlyName())
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -152,6 +168,7 @@ func buildBuildNode(node *pkggraph.PkgNode, pkgGraph *pkggraph.PkgGraph, graphMu
 	}
 
 	usedCache = false
+	hadMissingFiles = len(missingFiles) > 0
```
PawelWMS, line 171: Small ask: can we move setting `hadMissingFiles` before the `if` above it, so we can replace the code duplication in `if len(missingFiles) > 0` with `if hasMissingFiles`?

toolkit/tools/scheduler/scheduler.go
PawelWMS: I see now that we do very similar logging inside `NewGraphBuildState`. I'm curious if we need this `if` at all in that case.

toolkit/tools/scheduler/schedulerutils/graphbuildstate.go
```diff
@@ -17,31 +18,52 @@ type nodeState struct {
...
+func NewGraphBuildState(reservedFiles []string, maxFreshness int) (g *GraphBuildState) {
 	filesMap := make(map[string]bool)
 	for _, file := range reservedFiles {
 		filesMap[file] = true
 	}
+
+	//Use max int to represent infinity. This will cause unbounded cascading rebuilds so long as we build less that 2^31
+	// or more likely 2^63 packages (i.e., ~2.3 billion packages)
+	if maxFreshness < 0 {
+		logger.Log.Debugf("maxFreshness was set to %d, unbounded cascading rebuilds (maxFreshness = %d)", maxFreshness, NodeFreshnessAbsoluteMax)
```
PawelWMS, line 56: Could you re-word this message a bit? I understand what you mean now, but it seems like something's off about this sentence.

toolkit/tools/scheduler/schedulerutils/graphbuildstate.go
```diff
@@ -17,31 +18,52 @@ type nodeState struct {
...
+//     '0' where the subsequent nodes will no longer be rebuilt. 'maxFreshness < 0' will cause unbounded cascading rebuilds,
+//     while 'maxFreshness = 0' will cause no cascading rebuilds.
+func NewGraphBuildState(reservedFiles []string, maxFreshness int) (g *GraphBuildState) {
 	filesMap := make(map[string]bool)
 	for _, file := range reservedFiles {
 		filesMap[file] = true
 	}
+
+	//Use max int to represent infinity. This will cause unbounded cascading rebuilds so long as we build less that 2^31
+	// or more likely 2^63 packages (i.e., ~2.3 billion packages)
```
PawelWMS, line 54: I think we can be a bit more concise with our comments...;) Also, in this particular place we're checking for negative numbers, so this might be a bit confusing.

toolkit/tools/scheduler/schedulerutils/graphbuildstate.go
```diff
@@ -17,31 +18,52 @@ type nodeState struct {
 	available bool
 	cached    bool
 	usedDelta bool
+	freshness int
 }
 
+const (
+	// Negative values indicate that the node was rebuilt due to missing files and should gain the maximum configured freshness value.
+	NodeFreshnessRebuildRequired int = -1
```
PawelWMS, line 26: I didn't know we can set types for constants. This is very handy - thanks!

toolkit/tools/scheduler/schedulerutils/graphbuildstate.go
```diff
@@ -143,10 +176,22 @@ func (g *GraphBuildState) RecordBuildResult(res *BuildResult, allowToolchainRebu
 		g.failures = append(g.failures, res)
 	}
 
+	// 'NodeFreshnessRebuildRequired' is a special value that indicates that the node was rebuilt due to  missing files
+	// (user requested rebuilds are already at the max freshness). In this case, we want to reset the freshness to the
+	// max, so that subsequent dependant nodes will be rebuilt. Also ensure that the freshness is not greater than the max.
+	freshness := res.ActualFreshness
+	if freshness < 0 || freshness > g.GetMaxFreshness() {
```
PawelWMS, line 183: These unusual values of freshness look like something, that should never happen and is a bug in our code. For such cases I think we should just error-out, since we no longer can say we're in a valid state.

toolkit/tools/scheduler/schedulerutils/preparerequest.go
```diff
@@ -34,17 +34,24 @@ func ConvertNodesToRequests(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMute
 		}
 
 		req := &BuildRequest{
-			Node:           node,
-			PkgGraph:       pkgGraph,
-			AncillaryNodes: []*pkggraph.PkgNode{node},
-			IsDelta:        node.State == pkggraph.StateDelta,
+			Node:              node,
```
PawelWMS, line 37: This part might need a merge from `main`, since it changed quite drastically recently.

toolkit/tools/scheduler/scheduler.go
dmcilvaney: Removed in favor of the internal package version.

toolkit/tools/scheduler/schedulerutils/buildworker.go
dmcilvaney: Will do.

toolkit/tools/scheduler/schedulerutils/graphbuildstate.go
dmcilvaney: Done

toolkit/tools/scheduler/schedulerutils/preparerequest.go
dmcilvaney: Will do, will address feedback first, then give the merge a shot.

toolkit/tools/scheduler/scheduler.go
```diff
@@ -69,6 +70,7 @@ var (
 	rpmmacrosFile              = app.Flag("rpmmacros-file", "Optional file path to an rpmmacros file for rpmbuild to use.").ExistingFile()
 	buildAttempts              = app.Flag("build-attempts", "Sets the number of times to try building a package.").Default(defaultBuildAttempts).Int()
 	checkAttempts              = app.Flag("check-attempts", "Sets the minimum number of times to test a package if the tests fail.").Default(defaultCheckAttempts).Int()
+	maxCascadingRebuilds       = app.Flag("max-cascading-rebuilds", "Sets the maximum number of cascading dependency rebuilds caused by package being rebuilt (< 0 for unbounded).").Default(defaultMaxCascadingRebuilds).Int()
```
PawelWMS, line 73: Came to my mind right now: for consistency across the code, maybe we can just use `NodeFreshnessAbsoluteMax`. This way there's only one source of truth saying what's the maximum. The default value for `MAX_CASCADING_REBUILDS` could be left blank and `--max-cascading-rebuilds` would be set only if it has an actual value.

```suggestion
	maxCascadingRebuilds       = app.Flag("max-cascading-rebuilds", "Sets the maximum number of cascading dependency rebuilds caused by package being rebuilt (< 0 for unbounded).").Default(schedulerutils.NodeFreshnessAbsoluteMax).Int()
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -32,14 +32,15 @@ type BuildChannels struct {
...
-	ExpectedFiles  []string
-	UseCache       bool
-	IsDelta        bool
+	Node           *pkggraph.PkgNode   // The main node being analyzed for the build.
+	PkgGraph       *pkggraph.PkgGraph  // The graph of all packages
+	AncillaryNodes []*pkggraph.PkgNode // For SRPM builds: other nodes stemming from the same SRPM. Empty otherwise.
+	ExpectedFiles  []string            // List of RPMs built by this node.
+	UseCache       bool                // Can we use a cached copy of this package instead of building it
+	IsDelta        bool                // Is this a pre-downloaded rpm (not traditional cache) that we may be able to skip rebuilding
+	Freshness      int                 // The freshness of the node (used to determine if we can skip building future nodes)
```
PawelWMS, line 43: Nit-picks out of lack of more serious things to worry about here:
```suggestion
	PkgGraph       *pkggraph.PkgGraph  // The graph of all packages.
	AncillaryNodes []*pkggraph.PkgNode // For SRPM builds: other nodes stemming from the same SRPM. Empty otherwise.
	ExpectedFiles  []string            // List of RPMs built by this node.
	UseCache       bool                // Can we use a cached copy of this package instead of building it.
	IsDelta        bool                // Is this a pre-downloaded RPM (not traditional cache) that we may be able to skip rebuilding.
	Freshness      int                 // The freshness of the node (used to determine if we can skip building future nodes).
```

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -52,6 +53,7 @@ type BuildResult struct {
 	Ignored        bool                // Indicator if the build was ignored by user request.
 	UsedCache      bool                // Indicator if we used the cached artifacts (external or earlier local build) instead of building the node.
 	WasDelta       bool                // Indicator if we used a pre-built component from an external repository instead of building the node.
+	Freshness      int                 // The freshness of the node (used to determine if we can skip building future nodes)
```
PawelWMS, line 56:
```suggestion
	Freshness      int                 // The freshness of the node (used to determine if we can skip building future nodes).
```

toolkit/tools/scheduler/schedulerutils/graphbuildstate.go
```diff
@@ -17,31 +19,50 @@ type nodeState struct {
...
+//     while 'maxFreshness = 0' will cause no cascading rebuilds.
+func NewGraphBuildState(reservedFiles []string, maxFreshness int) (g *GraphBuildState) {
 	filesMap := make(map[string]bool)
 	for _, file := range reservedFiles {
 		filesMap[file] = true
 	}
+
+	// If we select negative maxFreshness this means we want unbounded cascading. Use max int to represent infinity.
+	// (2^31 or 2^63 packages is much more than the size of our graph, so this has the desired effect)
+	if maxFreshness < 0 {
```
PawelWMS, line 54: What do you think about making `maxFreshness` a `uint` (including the argument for the scheduler and switching the maximum to `math.MaxUint`) to avoid special-casing negative numbers?

toolkit/tools/scheduler/schedulerutils/graphbuildstate.go
```diff
@@ -176,10 +208,20 @@ func (g *GraphBuildState) RecordBuildResult(res *BuildResult, allowToolchainRebu
 		g.failures = append(g.failures, res)
 	}
 
+	// 'NodeFreshnessRebuildRequired' is a special value that indicates that the node was rebuilt due to  missing files
+	// (user requested rebuilds are already at the max freshness). In this case, we want to reset the freshness to the
+	// max, so that subsequent dependant nodes will be rebuilt. Also ensure that the freshness is not greater than the max.
+	freshness := res.Freshness
+	if freshness < 0 || freshness > g.GetMaxFreshness() {
+		err = fmt.Errorf("unexpected freshness value of '%d' for node '%s'. Should be: (0 < %d <= %d)", freshness, res.Node.FriendlyName(), freshness, g.GetMaxFreshness())
```
PawelWMS, line 216: I'm a bit worried that `0 < 15 <= 2` (just an example) might be a bit confusing. What do you think about this:
```suggestion
		err = fmt.Errorf("unexpected freshness value of %d for node '%s'. Should be: (0 < <freshness> <= %d)", freshness, res.Node.FriendlyName(), g.GetMaxFreshness())
```

toolkit/tools/scheduler/schedulerutils/preparerequest.go
```diff
@@ -149,13 +158,20 @@ func partnerTestNodesToRequest(pkgGraph *pkggraph.PkgGraph, buildState *GraphBui
...
-		UseCache:       buildUsesCache && canUseCacheForNode(pkgGraph, testNode, testsToRerun, buildState),
+		Freshness:      buildState.GetMaxFreshness(),
 	}
+
+	requiredRebuild := isRequiredRebuild(request.Node, testsToRerun)
+	if !requiredRebuild && buildUsesCache {
+		// We might be able to use the cache, set the freshness based on node's dependencies.
+		request.UseCache, request.Freshness = canUseCacheForNode(pkgGraph, testNode, buildState)
+	}
+	return
```
PawelWMS, line 174: Looks like I missed that while working on the test nodes change: could I ask you to replace this repeated code with a call to `buildRequest`? It seems like it's identical in terms of what it does.

toolkit/tools/scheduler/schedulerutils/preparerequest.go
```diff
@@ -182,36 +198,65 @@ func testNodesToRequests(pkgGraph *pkggraph.PkgGraph, buildState *GraphBuildStat
...
 		dependency := dependencies.Node().(*pkggraph.PkgNode)
 
-		if !buildState.IsNodeCached(dependency) {
+		inheritedFreshness, shouldRebuild := calculateExpectedFreshness(dependency, buildState)
+		if inheritedFreshness > freshness {
+			freshness = inheritedFreshness
+		}
+
+		if shouldRebuild {
 			logger.Log.Debugf("Can't use cached version of %v because %v is rebuilding", node.FriendlyName(), dependency.FriendlyName())
```
PawelWMS, line 233: To consider: I'm thinking we might want to slightly update this message now. For instance, replace `is rebuilding` with `has been rebuilt and had a freshness of %d`.

toolkit/tools/scheduler/schedulerutils/preparerequest.go
```diff
@@ -182,36 +198,65 @@ func testNodesToRequests(pkgGraph *pkggraph.PkgGraph, buildState *GraphBuildStat
...
+	// Remote nodes are always 'stale' and should never generate a rebuild.
+	if dependencyNode.Type == pkggraph.TypeRemoteRun {
+		return 0, false
+	}
+
+	expectedFreshness = buildState.GetFreshnessOfNode(dependencyNode)
+	shouldRebuild = expectedFreshness > 0
+	// The transition from (* -> run) nodes is sufficient to attenuate the freshness throughout the graph. For BuildRequires,
+	// each build node will always be accompanied by a run node (i.e., no other nodes depend directly on the build
+	// node, and we would like the associated run node to inherit it's build node's freshness). We also want to
```
PawelWMS, line 254:
```suggestion
	// node, and we would like the associated run node to inherit its build node's freshness). We also want to
```

toolkit/tools/scheduler/schedulerutils/graphbuildstate.go
dmcilvaney: Sure, sounds good. It was needed in the older version before you refactored the request building code, but now we can fully calculate the freshness before we even try to build so there is no need to send a special `-1` signal back on a rebuild from the worker. Will update alongside the new default value approach.

toolkit/tools/scheduler/schedulerutils/preparerequest.go
dmcilvaney: Will do

toolkit/docs/building/building.md
```diff
@@ -807,6 +807,7 @@ To reproduce an ISO build, run the same make invocation as before, but set:
 | ALLOW_TOOLCHAIN_REBUILDS      | n                                                                                                      | Do not treat rebuilds of toolchain packages during regular package build phase as errors.
 |  PACKAGE_BUILD_RETRIES        | 1                                                                                                      | Number of build retries for each package
 | CHECK_BUILD_RETRIES           | 1                                                                                                      | Minimum number of check section retries for each package if RUN_CHECK=y and tests fail.
+| MAX_CASCADING_REBUILDS        | -1                                                                                                     | When a package rebuilds, how many additional layers of dependant packages will be forced to rebuild (-1 is unbounded, i.e., all downstream packages will rebuild)
```
PawelWMS, line 810:
```suggestion
| MAX_CASCADING_REBUILDS        |                                                                                                        | When a package rebuilds, how many additional layers of dependant packages will be forced to rebuild (leave unset for unbounded, i.e., all downstream packages will rebuild)
```

PawelWMS (approved):

PR #6010, opened 2023-08-17
[RFC] Add EnableFIPS option to imageconfig schema

christopherco:
This Draft PR is an RFC to get early feedback on adding native tooling support for enabling FIPS in the image generation tooling and adding new image definitions for Gen1, Gen2, and ARM64 Azure VMs with FIPS enabled by default.

To enable FIPS on an image, the following steps are needed:
1. Add `fips=1` to the kernel command line to activate the kernel's FIPS-approved cryptography
2. Install the `dracut-fips` package which includes FIPS self-tests during kernel boot, and will prevent boot if the self-tests fail
3. Add `boot=<bootUUID>` to the kernel command line, which gets consumed by dracut to find the kernel's `.hmac` file

Still needed:
1. Additional unit tests
2. Documentation update for this new option


toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -2038,6 +2045,31 @@ func setGrubCfgSELinux(grubPath string, kernelCommandline configuration.KernelCo
...
+	const (
+		enableFIPSPattern = "{{.FIPS}}"
+		enableFIPS        = "fips=1"
+		bootPrefix        = "boot="
+		uuidPrefix        = "UUID="
+	)
+
+	var fips string
+
+	// If EnableFIPS is set, add "fips=1 boot=UUID=<bootUUID value>" to the kernel cmdline in grub.cfg
```
dmcilvaney, line 2058: I've seen a few places that suggest having `boot=` set when there is NOT a separate boot partition breaks things... I *think* it works fine so long as it does actually point to the partition with boot on it, but I'd like to be sure a system with just a single root partition works ok.

toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
```diff
@@ -157,7 +157,7 @@ func validatePackages(config configuration.Config) (err error) {
 				return fmt.Errorf("%s: [ReadOnlyVerityRoot] and [TmpfsOverlayDebugEnabled] selected, but '%s' package is not included in the package lists", validateError, verityDebugPkgName)
 			}
 		}
-		if strings.Contains(kernelCmdLineString, fipsKernelCmdLine) {
+		if strings.Contains(kernelCmdLineString, fipsKernelCmdLine) || systemConfig.KernelCommandLine.EnableFIPS {
```
PawelWMS, line 160: Is adding `fips=1` to the kernel command line different from what happens when we set `EnableFIPS: true`? If not, I'm curious, if we need the new field.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -2038,6 +2045,31 @@ func setGrubCfgSELinux(grubPath string, kernelCommandline configuration.KernelCo
...
+	} else {
+		fips = ""
+	}
+
+	logger.Log.Debugf("Adding EnableFIPS('%s') to '%s'", fips, grubPath)
+	err = sed(enableFIPSPattern, fips, kernelCommandline.GetSedDelimeter(), grubPath)
+	if err != nil {
+		logger.Log.Warnf("Failed to set grub.cfg's EnableFIPS setting: %v", err)
+	}
+	return
```
PawelWMS, line 2070: Small simplification and clarification of the variable's content:
```suggestion
	// If EnableFIPS is set, add "fips=1 boot=UUID=<bootUUID value>" to the kernel cmdline in grub.cfg
	fipsKernelArgument := ""
	if kernelCommandline.EnableFIPS {
		fipsKernelArgument = fmt.Sprintf("%s %s%s%s", enableFIPS, bootPrefix, uuidPrefix, bootUUID)
	}

	logger.Log.Debugf("Adding EnableFIPS('%s') to '%s'", fipsKernelArgument, grubPath)
	err = sed(enableFIPSPattern, fipsKernelArgument, kernelCommandline.GetSedDelimeter(), grubPath)
	if err != nil {
		logger.Log.Warnf("Failed to set grub.cfg's EnableFIPS setting: %v", err)
	}
	return
```

toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
christopherco: The `EnableFIPS: true` option operates differently in the particular case when the /boot partition is separate from the root partition. When the /boot partition is separate and `fips=1`, the kernel command line must contain a `boot=` parameter, which gets read during the fips self-tests during bootup to find the kernel's .hmac file which resides in the /boot partition.

So when `EnableFIPS: true`, the behavior adds the `boot=<reference to boot partition>` parameter (via UUID `boot=UUID=<boot partition's uuid>`) to the kernel cmdline in addition to `fips=1`.

If `fips=1` and `EnableFIPS: true` were both present, we would have two `fips=1` entries on the kernel cmdline and possibly the `boot=` parameter referenced above. And the kernel would only honor one `fips=1` setting (the last one set in the kernel cmdline).

toolkit/tools/imagegen/installutils/installutils.go
christopherco: Thanks for letting me know. Found some references as well. Will test some behaviors out of curiosity but it is probably safest to add the logic to detect if /boot is a separate partition from the root partition before adding `boot=`.

References:
https://docs.oracle.com/en/operating-systems/oracle-linux/6/security/ol_fips_enable.html
https://www.suse.com/support/kb/doc/?id=000018548

toolkit/tools/imager/imager.go
```diff
@@ -662,8 +662,10 @@ func configureDiskBootloader(systemConfig configuration.SystemConfig, installChr
 	// Add bootloader. Prefer a separate boot partition if one exists.
 	bootDevice, ok := installMap[bootMountPoint]
```
dmcilvaney, line 663: not super important, but you could just rename `ok` -> `isBootSeparatePartition` here, would clarify the follow on conditional as well a bit.

christopherco:
Closing this RFC as the actual patches are now posted

PR #6035, opened 2023-08-23
toolkit: containerized-build: add vim & git in the build env

mfrw:

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

To help improve the developer expeirience w.r.t to the `containerized-build` feature,
add a minimal set of tools, so that we dont have to install them everytime we start.

###### Change Log  

- Add `git` & `vim` to build container

**NO**

###### Associated issues  

- NA

###### Links to CVEs  
- NA

###### Test Methodology

- Local build and test

### DIFF
--- a/toolkit/scripts/containerized-build/resources/mariner.Dockerfile
+++ b/toolkit/scripts/containerized-build/resources/mariner.Dockerfile
@@ -17,3 +17,6 @@ RUN if [[ "${enable_local_repo}" == "true" ]]; then echo "enable_local_repo" >> 
 RUN echo "cat /mariner_setup_dir/splash.txt"                     >> /root/.bashrc && \
     echo "show_help"                                             >> /root/.bashrc && \
     echo "cd /usr/src/mariner/ || { echo \"ERROR: Could not change directory to /usr/src/mariner/ \"; exit 1; }"  >> /root/.bashrc
+
+# Install vim & git in the build env
+RUN tdnf --releasever=$version install -y vim git


ameyapu (approved):
LGTM.  I find this very useful and would suggest that we keep observing if there are other small tools too that can be added.

neha170 (approved):
My only concern with adding packages to the container is that it should not alter behaviour when building packages. The aim is to keep the container as close to the build chroot as possible. When adding packages, the user should be aware that they have installed packages which are not present in typical build chroot (and their effect). These packages (vim, git) look harmless to add. Thoughts @dmcilvaney?

dmcilvaney:
> My only concern with adding packages to the container is that it should not alter behaviour when building packages. The aim is to keep the container as close to the build chroot as possible. When adding packages, the user should be aware that they have installed packages which are not present in typical build chroot (and their effect). These packages (vim, git) look harmless to add. Thoughts @dmcilvaney?

I'm not opposed, but agreed that every change is a trade-off. Each additional .rpm in the build environment will change the behavior of `autoconf` etc. which can result in the built packages having different dependencies/libraries etc than the "real" versions.

My personal workflow for this is to mount the interesting directories out of the container and use my normal editor/tools on them from outside (vscode in this case).

mfrw:
> > My only concern with adding packages to the container is that it should not alter behaviour when building packages. The aim is to keep the container as close to the build chroot as possible. When adding packages, the user should be aware that they have installed packages which are not present in typical build chroot (and their effect). These packages (vim, git) look harmless to add. Thoughts @dmcilvaney?
> 
> I'm not opposed, but agreed that every change is a trade-off. Each additional .rpm in the build environment will change the behavior of `autoconf` etc. which can result in the built packages having different dependencies/libraries etc than the "real" versions.
> 
> My personal workflow for this is to mount the interesting directories out of the container and use my normal editor/tools on them from outside (vscode in this case).

Given this is a `DevEx` feature, aimed at making `our` lives better [IMO].
We might as well, reduce a little bit of repetition. I will probably merge it.
If it causes issues latter-on, we can drop it :)

PR #6051, opened 2023-08-24
make cloud-init-output.log available to serial console

rmhsawyer:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

There is a customer request that asking content of cloud-init-output.log to be available on Azure diagnostics serial log for Mariner VM, same as Ubuntu. In cloud-init.spec file, there is a particular line that makes the StandardOutput logs from cloud-init systemd service only dump to journal but not system console, which causes the content of cloud-init-output.log missing on Azure diagnostics serial log. We need to remove this line to resolve this issue.

###### Change Log  

- Remove the line that replaces "StandardOutput=journal+console" to "StandardOutput=journal" in cloud-init systemd service: "find systemd -name "cloud*.service*" | xargs sed -i s/StandardOutput=journal+console/StandardOutput=journal/g"
- This particular line makes the logs from cloud-init systemd service StandardOutput dump only to journal instead of system console and journal both. Journal+console option will first connect standard output with the journal, which is accessible via [journalctl(1)](https://www.freedesktop.org/software/systemd/man/journalctl.html#), and then copy the output to the system console as well. After removing this line, content of cloud-init-output.log is available on Azure diagnostics serial log.

NO

###### Associated issues  

- [Bug 46024617](https://microsoft.visualstudio.com/OS/_workitems/edit/46024617): [2.0][cloud-init] Azure boot diagnostic logs missing from Mariner's /var/log/cloud-init-output.log

###### Test Methodology

- Pipeline build id: [412772](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=412772&view=results)
- Azure VM deployment and verify that the cloud-init-ouput.log now pops up in serial console
- BuddyBuild: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=412773&view=results

### DIFF
--- a/SPECS/cloud-init/cloud-init.spec
+++ b/SPECS/cloud-init/cloud-init.spec
@@ -1,7 +1,7 @@
 Summary:        Cloud instance init scripts
 Name:           cloud-init
 Version:        23.2
-Release:        2%{?dist}
+Release:        3%{?dist}
 License:        GPLv3
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -74,8 +74,6 @@ Cloud-init configuration for Hyper-V telemetry
 %prep
 %autosetup -p1 -n %{name}-%{version}
 
-find systemd -name "cloud*.service*" | xargs sed -i s/StandardOutput=journal+console/StandardOutput=journal/g
-
 %build
 python3 setup.py build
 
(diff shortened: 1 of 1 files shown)


SPECS/cloud-init/cloud-init.spec
```diff
@@ -74,8 +74,6 @@ Cloud-init configuration for Hyper-V telemetry
 %prep
 %autosetup -p1 -n %{name}-%{version}
 
-find systemd -name "cloud*.service*" | xargs sed -i s/StandardOutput=journal+console/StandardOutput=journal/g
```
christopherco, line 77: Can you update the PR description with why this change fixes the issue with the serial console?

i.e., https://www.freedesktop.org/software/systemd/man/systemd.exec.html#StandardOutput=

christopherco (changes requested):

ddstreetmicrosoft (approved):
Removing the line makes sense, I don't see why it was ever added.

christopherco (approved):

PR #6070, opened 2023-08-25
Strip epoch on packages that use it in pre-cache

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Based on the contents of rpms-snapshot the pre-cache tool was calculating versions of the form `1:2.3.4-5`, of note is the epoch value `1`. The tool uses these values to compute the expected .rpm filepaths, but these paths do not contain the epoch number. Now we try to patch an rpm both with and without the epoch. We try both so we can explicitly warn the user we will not download the version with epoch (since `:` is a special character in `Make` and we use the contents of the rpm cache as an input into the rest of the tools).

###### Change Log  

- String `<epoch>:` from the version of packages being checked by the pre-cache tool.

**NO**

###### Test Methodology

- Local

### DIFF
--- a/toolkit/scripts/precache.mk
+++ b/toolkit/scripts/precache.mk
@@ -8,6 +8,7 @@
 
 precache_state_dir = $(CACHED_RPMS_DIR)/precache
 precache_downloaded_files = $(precache_state_dir)/downloaded_files.txt
+repo_summary_file = $(precache_state_dir)/repo_summary.txt
 precache_chroot_dir = $(precache_state_dir)/chroot
 
 $(call create_folder,$(precache_state_dir))
@@ -34,6 +35,7 @@ $(STATUS_FLAGS_DIR)/precache.flag: $(go-precacher) $(chroot_worker) $(rpms_snaps
 		--snapshot "$(rpms_snapshot)" \
 		--output-dir "$(remote_rpms_cache_dir)" \
 		--output-summary-file "$(precache_downloaded_files)" \
+		--repo-summary-file "$(repo_summary_file)" \
 		$(foreach url,$(PACKAGE_URL_LIST), --repo-url "$(url)") \
 		--worker-tar $(chroot_worker) \
 		--worker-dir $(precache_chroot_dir) \
--- a/toolkit/tools/precacher/precacher.go
+++ b/toolkit/tools/precacher/precacher.go
@@ -59,6 +59,7 @@ var (
 	outDir            = app.Flag("output-dir", "Directory to download packages into.").Required().ExistingDir()
 	snapshot          = app.Flag("snapshot", "Path to the rpm snapshot .json file.").ExistingFile()
 	outputSummaryFile = app.Flag("output-summary-file", "Path to save the summary of packages downloaded").String()
+	repoSummaryFile   = app.Flag("repo-summary-file", "Path to save the list of packages available in the repos").String()
 	repoUrls          = app.Flag("repo-url", "URLs of the repos to download from.").Strings()
 	workerTar         = app.Flag("worker-tar", "Full path to worker_chroot.tar.gz").Required().ExistingFile()
 	buildDir          = app.Flag("worker-dir", "Directory to store chroot while running repo query.").Required().ExistingDir()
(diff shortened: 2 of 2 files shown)


toolkit/tools/precacher/precacher.go
```diff
@@ -160,8 +162,11 @@ func getAllRepoData(repoURLs []string, workerTar, buildDir string) (namesToURLs
 			// We need to prepend the repoURL to the partial URL to get the full URL
 			packageRepoPath = fmt.Sprintf("%s/%s", repoURL, packageRepoPath)
 			namesToURLs[packageName] = packageRepoPath
+			URLList = append(URLList, packageRepoPath)
 		}
 	}
+	file.WriteLines(URLList, listFile)
```
PawelWMS, line 168:
```suggestion
	err = file.WriteLines(URLList, listFile)
```

toolkit/tools/precacher/precacher.go
```diff
@@ -59,6 +59,7 @@ var (
 	outDir            = app.Flag("output-dir", "Directory to download packages into.").Required().ExistingDir()
 	snapshot          = app.Flag("snapshot", "Path to the rpm snapshot .json file.").ExistingFile()
 	outputSummaryFile = app.Flag("output-summary-file", "Path to save the summary of packages downloaded").String()
+	repoSummaryFile   = app.Flag("repo-summary-file", "Path to save the list of packages available in the repos").String()
```
PawelWMS, line 62: Can you rephrase to clarify this is the list of URLs to the packages, not just package names themselves?

toolkit/scripts/precache.mk
```diff
@@ -34,6 +35,7 @@ $(STATUS_FLAGS_DIR)/precache.flag: $(go-precacher) $(chroot_worker) $(rpms_snaps
 		--snapshot "$(rpms_snapshot)" \
 		--output-dir "$(remote_rpms_cache_dir)" \
 		--output-summary-file "$(precache_downloaded_files)" \
+		--repo-summary-file "$(repo_summary_file)" \
```
PawelWMS, line 38: Small request: can we use a different word than "summary-file". We've been using that word for the JSON files containing the contents of the `RepoContents` structs. Best to avoid overloading terms, if possible.

toolkit/tools/precacher/precacher.go
```diff
@@ -350,12 +355,12 @@ func precachePackage(pkg *repocloner.RepoPackage, packagesAvailableFromRepos map
 	const (
 		downloadRetryAttempts = 2
 		downloadRetryDuration = time.Second
+		stripEpoch            = false
+		leaveEpoch            = true
```
PawelWMS, line 359: Nit-pick: alphabetical order.
```suggestion
		leaveEpoch            = true
		stripEpoch            = false
```

PawelWMS, line 359: File names for RPM packages never contain the epoch number - that's only encoded in the RPMs metadata inside these files. We can just always strip the epoch from the version, if it's present and simplify the code a bit.

toolkit/tools/precacher/precacher.go
dmcilvaney: Done

toolkit/tools/precacher/precacher.go
dmcilvaney: removed

dmcilvaney: Good to know, I've simplified this then, and just always strip the epoch if its present in the version

toolkit/tools/precacher/precacher.go
```diff
@@ -377,7 +381,7 @@ func precachePackage(pkg *repocloner.RepoPackage, packagesAvailableFromRepos map
 		return
 	}
 
-	// Get the url for the package, or bail out if it is not available
+	// Get the url for the package, or bail out if it is not available.
```
PawelWMS, line 384: Since we're here...:)
```suggestion
	// Get the URL for the package, or bail out if it is not available.
```

PawelWMS (approved):

PR #6080, opened 2023-08-28
Remove dst file on failed network download

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

When `network.DownloadFile()` it was leaving files on the disk. The pre-cacher tool would create empty `.rpm` files which would confuse the fetcher and scheduler further downstream.

###### Change Log  

- network.DownloadFile() will now clean up its dst file on error
- Added `file.RemoveFileIfExists()`
- Add file_test.go 

**NO**

###### Test Methodology

- Local test

### DIFF
--- a/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
+++ b/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
@@ -423,6 +423,8 @@ func downloadSingleDeltaRPM(realDependencyGraph *pkggraph.PkgGraph, buildNode *p
 			logger.Log.Warnf("Can't find delta RPM to download for %s: %s (local copy may be newer than published version)", fullyQualifiedRpmName, err)
 			return nil
 		}
+	} else {
+		logger.Log.Debugf("Found pre-cached delta RPM for %s, skipping download", fullyQualifiedRpmName)
 	}
 
 	foundCacheRPM, err = file.PathExists(cachedRPMPath)
--- a/toolkit/tools/internal/file/file_test.go
+++ b/toolkit/tools/internal/file/file_test.go
@@ -18,20 +18,28 @@ func TestMain(m *testing.M) {
 	os.Exit(retVal)
 }
 
+// testFileName returns a file name in a temporary directory. This path will
+// be different for EVERY call to this function.
 func testFileName(t *testing.T) string {
 	return filepath.Join(t.TempDir(), t.Name())
 }
 
 func TestRemoveFileIfExistsValid(t *testing.T) {
+	fileName := testFileName(t)
 	// Create a file to remove
-	err := Write("test", testFileName(t))
+	err := Write("test", fileName)
 	assert.NoError(t, err)
 
-	err = RemoveFileIfExists(testFileName(t))
+	err = RemoveFileIfExists(fileName)
 	assert.NoError(t, err)
+
+	exists, err := PathExists(fileName)
+	assert.NoError(t, err)
+	assert.False(t, exists)
 }
 
 func TestRemoveFileDoesNotExist(t *testing.T) {
-	err := RemoveFileIfExists(testFileName(t))
+	fileName := testFileName(t)
+	err := RemoveFileIfExists(fileName)
 	assert.NoError(t, err)
 }
--- a/toolkit/tools/internal/file/file_test.go
+++ b/toolkit/tools/internal/file/file_test.go
@@ -1,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package file
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMain(m *testing.M) {
+	logger.InitStderrLog()
+	retVal := m.Run()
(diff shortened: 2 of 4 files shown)


toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -423,6 +423,8 @@ func downloadSingleDeltaRPM(realDependencyGraph *pkggraph.PkgGraph, buildNode *p
 			logger.Log.Warnf("Can't find delta RPM to download for %s: %s (local copy may be newer than published version)", fullyQualifiedRpmName, err)
 			return nil
 		}
+	} else {
```
neha170, line 426: This seems unrelated to the PR, and probably deserves its own PR. I am thinking from the perspective where we might want to roll-back in some imaginary scenario, and would end up erroneously deleting this edge case in graphpkgfetcher.

toolkit/tools/internal/network/network.go
```diff
@@ -65,6 +75,22 @@ func DownloadFile(url, dst string, caCerts *x509.CertPool, tlsCerts []tls.Certif
...
+	if exists {
+		err = os.Remove(fullPath)
+		if err != nil {
+			err = fmt.Errorf("failed to remove file:\n%w", err)
+			return
+		}
+	}
+	return
+}
+
```
PawelWMS, line 93: I think we might not even need to check if the file exists and rely on `os.Remove` error entirely by ignoring a missing file:
```suggestion
func removeFileIgnoreMissing(fullPath string) (err error) {
	removeErr := os.Remove(fullPath).(*os.PathError)
	if removeErr != nil && removeErr.Err != syscall.ENOENT {
		return fmt.Errorf("failed to remove file (%s):\n%w", fullPath, removeErr)
	}
	return removeErr
}
```

And what do you think about making that a function inside `file.go`?

toolkit/tools/internal/network/network.go
```diff
@@ -30,14 +31,23 @@ func JoinURL(baseURL string, extraPaths ...string) string {
...
+	defer func() {
+		// If there was an error, ensure that the file is removed
+		if err != nil {
+			cleanupErr := removeFileIfExists(dst)
+			if cleanupErr != nil {
+				logger.Log.Errorf("Failed to remove failed network download file '%s': %s", dst, err)
+			}
+		}
+	}()
 	defer dstFile.Close()
```
PawelWMS, line 51: Now the file will get closed before it gets removed, so there's a window where another thread/process may create a new file, which gets then removed by this code. I'd reverse the order of operations. Also, I think there's no need for two separate `defer` if they are called immediately one after another - we can lump these together.

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -423,6 +423,8 @@ func downloadSingleDeltaRPM(realDependencyGraph *pkggraph.PkgGraph, buildNode *p
...
+		logger.Log.Debugf("Found pre-cached delta RPM for %s, skipping download", fullyQualifiedRpmName)
 	}
 
 	foundCacheRPM, err = file.PathExists(cachedRPMPath)
```
PawelWMS, line 430: If we're introducing the `else`, then we should move the `foundCacheRPM, err = file.PathExists(cachedRPMPath)` call and repeated `foundCacheRPM` into the `if !foundCacheRPM {` block above. Then, if the cached RPM is still not there after we try to clone it, we could exit early with the `logger.Log.Warnf("Delta download for '%s' did not generate the correct delta RPM: '%s'", buildNode, cachedRPMPath)` warning we have below.

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
dmcilvaney: This doesn't materially affect the behavior of the tool, so I don't mind bundling it (and it was the debugging information that allowed me to diagnose the issue this is solving so I feel its at least partially related).

toolkit/tools/internal/network/network.go
dmcilvaney: Moved to file.go, and as per comments in the go library `if removeErr != nil && !errors.Is(removeErr, fs.ErrNotExist) {` is the modern way to check for this error.

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
dmcilvaney: I'll make a todo, but I'd like to limit extra changes here.

toolkit/tools/internal/file/file_test.go
```diff
@@ -0,0 +1,37 @@
...
+	logger.InitStderrLog()
+	retVal := m.Run()
+	os.Exit(retVal)
+}
+
+func testFileName(t *testing.T) string {
+	return filepath.Join(t.TempDir(), t.Name())
+}
+
+func TestRemoveFileIfExistsValid(t *testing.T) {
```
PawelWMS, line 25: Thanks for adding the tests! Could we also add a check that the removed file does not exist after calling `RemoveFileIfExists`?

PawelWMS (approved):

toolkit/tools/internal/file/file_test.go
dmcilvaney: Realized I wasn't using t.TempDir right, fixed that as well.

PR #6082, opened 2023-08-28
Add workflow to automatically cherry-pick commits to development branches

trungams:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add a new GH actions workflow to cherry-pick CVE fixes from fast-track branches to dev branches of Mariner, which can be triggered by commenting on CVE fix PRs.

Original idea from this [LinkedIn engineering blog](https://engineering.linkedin.com/blog/2023/how-linkedin-automates-cherry-picking-commits-to-improve-develop)

###### Change Log  

- Add a yaml file for workflow definition and a bash script for the actual cherry-picking work.
- With the addition of fast-track branches for CVE fixes. Once a CVE fix PR is merged to fast-track branch, we can trigger this workflow by adding a comment to the PR in the format
  ```
  /cherry-pick <target-branch-1>, <target-branch-2>, <target-branch-3>
  ```
- This workflow will automatically cherry-pick the merge commit (assuming squash merge is used) to the target branches, and create a new PR for each target branch for final validation/merge. If cherry-picking fails for some branches, the workflow will output logs containing merge conflicts so developers can take action upon.

**NO**

###### Test Methodology

- Sample run of the workflow on my own fork of CBL-Mariner:
  - PR containing dummy CVE fix (just updating a spec file): https://github.com/trungams/CBL-Mariner/pull/3
  - PR created for successful cherry-pick: https://github.com/trungams/CBL-Mariner/pull/6
  - Logs for failed cherry-pick: https://github.com/trungams/CBL-Mariner/actions/runs/6004206617/job/16284283903#step:5:73
 
image

### DIFF
--- a/.github/workflows/cherry-pick.yml
+++ b/.github/workflows/cherry-pick.yml
@@ -1,0 +1,115 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+name: Cherry pick commits from PR
+on:
+  issue_comment:
+    types:
+      - created
+jobs:
+  pr_info:
+    if: |
+      github.event.issue.pull_request &&
+        startsWith(github.event.comment.body, '/cherry-pick ') &&
+        (github.event.comment.author_association == 'MEMBER' ||
+         github.event.comment.author_association == 'OWNER')
+
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    outputs:
+      status: ${{ steps.pr_info.outputs.status }}
+      base_ref: ${{ steps.pr_info.outputs.base_ref }}
+      merge_commit: ${{ steps.pr_info.outputs.merge_commit }}
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Extract PR information
+        id: pr_info
+        run: |
+          echo "status=$(gh pr view ${{ github.event.issue.number }} --repo ${{ github.repository }} --json state --jq '.state')" >> $GITHUB_OUTPUT
+          echo "base_ref=$(gh pr view ${{ github.event.issue.number }} --repo ${{ github.repository }} --json baseRefName --jq '.baseRefName')" >> $GITHUB_OUTPUT
+          echo "merge_commit=$(gh pr view ${{ github.event.issue.number }} --repo ${{ github.repository }} --json mergeCommit --jq '.mergeCommit.oid')" >> $GITHUB_OUTPUT
+
+      - name: Post workflow run to PR
+        run: |
+          if [[ "${{ steps.pr_info.outputs.status }}" != "MERGED" ]] || [[ "${{ steps.pr_info.outputs.base_ref }}" != fasttrack/* ]]; then
+            gh pr comment ${{ github.event.issue.number }} \
+              --repo ${{ github.repository }} \
+              --body 'Cherry pick action is only enabled for merged pull requests to `fasttrack/*` branches'
+          fi
+
+  cherry_pick_commit:
+    needs: pr_info
+    if: |
+      needs.pr_info.outputs.status == 'MERGED' &&
+        startsWith(needs.pr_info.outputs.base_ref, 'fasttrack/')
+
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    defaults:
+      run:
+        shell: bash
+    name: Cherry pick
+    steps:
+      - name: Workflow trigger checkout
+        uses: actions/checkout@v3
+        with:
+          # ref: ${{ needs.pr_info.outputs.base_ref }}
+          fetch-depth: 0
+
+      - name: Validate and extract cherry pick target from comment
+        run: |
+          pr_comment="${{ github.event.comment.body }}"
+          branches_list=${pr_comment#'/cherry-pick '}
+          IFS=", " read -ra target_branches <<< "$branches_list"
+          echo "cherry_pick_target_branches=${target_branches[*]}" >> $GITHUB_ENV
+
+      - name: Configure git
+        run: |
+          git config --local user.email "$GIT_USER_EMAIL"
+          git config --local user.name "$GIT_USER_NAME"
+        env:
+          GIT_USER_EMAIL: "cblmargh@microsoft.com"
+          GIT_USER_NAME: "CBL-Mariner Servicing Account"
+
+      - name: Run cherry-pick action
+        run: |
+          cp toolkit/scripts/cherry_pick.sh ${{ runner.temp }}/cherry_pick.sh
+          for target_branch in ${{ env.cherry_pick_target_branches }}; do
+            echo "Cherry picking commit to $target_branch"
+
+            log_file="${{ runner.temp }}/$target_branch.txt"
+
+            git reset --hard
+            git clean -df
+            git checkout -- .
+
+            ${{ runner.temp }}/cherry_pick.sh \
+              -b "$target_branch" \
+              -c "${{ needs.pr_info.outputs.merge_commit }}" \
+              -o "${{ github.event.issue.pull_request.html_url }}" \
+              -l "$log_file" \
+              -r "${{ github.repository }}" \
+              -t "${{ github.event.issue.title }}" \
+              || rc=$?
+            if [ ${rc:-0} -ne 0 ]; then
+              gh pr comment ${{ github.event.issue.number }} \
+                --repo ${{ github.repository }} \
+                --body "Cherry-pick failed for branch $target_branch. See run log for more details: $RUN_URL"
+            else
+              gh pr comment ${{ github.event.issue.number }} \
+                --repo ${{ github.repository }} \
+                --body "Cherry-pick succeeded for branch $target_branch. See pull request $(cat $log_file)"
+              gh pr edit ${{ github.event.issue.number }} \
+                --repo ${{ github.repository }} \
+                --add-label "cherry_pick-$target_branch"
+            fi
+            echo "=============================================================="
+          done
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
--- a/.github/workflows/cherry-pick.yml
+++ b/.github/workflows/cherry-pick.yml
@@ -1,12 +1,22 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+# This action cherry-picks CVE fix commits from fast-track branches to corresponding
+# development branches for each version of Mariner.
+# To use this action, create a comment in a merged PR to any 'fasttrack/*' branch:
+#   /cherry-pick <target-branch-1>, <target-branch-2>, etc
+# The author of the comment must be a member or owner of the repository for the action
+# to work.
+# After cherry-pick succeeds for a branch, a new PR will be created to merge the commit
+# to that target branch, and a label will be added to the original PR to indicate that
+# the commit has been cherry-picked to that branch.
 name: Cherry pick commits from PR
 on:
   issue_comment:
     types:
       - created
 jobs:
+  # Basic information extraction and validation
   pr_info:
     if: |
       github.event.issue.pull_request &&
@@ -32,7 +42,7 @@ jobs:
           echo "base_ref=$(gh pr view ${{ github.event.issue.number }} --repo ${{ github.repository }} --json baseRefName --jq '.baseRefName')" >> $GITHUB_OUTPUT
           echo "merge_commit=$(gh pr view ${{ github.event.issue.number }} --repo ${{ github.repository }} --json mergeCommit --jq '.mergeCommit.oid')" >> $GITHUB_OUTPUT
 
-      - name: Post workflow run to PR
+      - name: Check that PR satisfies requirements for cherry-pick action and notify if not
         run: |
           if [[ "${{ steps.pr_info.outputs.status }}" != "MERGED" ]] || [[ "${{ steps.pr_info.outputs.base_ref }}" != fasttrack/* ]]; then
             gh pr comment ${{ github.event.issue.number }} \
@@ -40,6 +50,7 @@ jobs:
               --body 'Cherry pick action is only enabled for merged pull requests to `fasttrack/*` branches'
           fi
 
+  # Actual cherry-pick work for each target branch
   cherry_pick_commit:
     needs: pr_info
     if: |
@@ -58,7 +69,6 @@ jobs:
       - name: Workflow trigger checkout
         uses: actions/checkout@v3
         with:
-          # ref: ${{ needs.pr_info.outputs.base_ref }}
           fetch-depth: 0
 
       - name: Validate and extract cherry pick target from comment
@@ -68,7 +78,7 @@ jobs:
           IFS=", " read -ra target_branches <<< "$branches_list"
           echo "cherry_pick_target_branches=${target_branches[*]}" >> $GITHUB_ENV
 
-      - name: Configure git
+      - name: Configure local git repo
         run: |
           git config --local user.email "$GIT_USER_EMAIL"
           git config --local user.name "$GIT_USER_NAME"
@@ -76,9 +86,12 @@ jobs:
           GIT_USER_EMAIL: "cblmargh@microsoft.com"
           GIT_USER_NAME: "CBL-Mariner Servicing Account"
 
+      # Run cherry-pick for each target branch and create new PR for it. If cherry-pick fails,
+      # output conflicts to the action logs and notify in the original PR.
       - name: Run cherry-pick action
         run: |
           cp toolkit/scripts/cherry_pick.sh ${{ runner.temp }}/cherry_pick.sh
+
           for target_branch in ${{ env.cherry_pick_target_branches }}; do
             echo "Cherry picking commit to $target_branch"
 
--- a/toolkit/scripts/cherry_pick.sh
+++ b/toolkit/scripts/cherry_pick.sh
@@ -5,7 +5,14 @@
 set -e
 
 function help {
-    echo "hello, world"
+    echo "Cherry-pick commit to a specific branch and create Github PR"
+    echo "Usage:"
+    echo "-b TARGET_BRANCH -> target branch to cherry pick commit to"
+    echo "-c COMMIT_HASH"
+    echo "-l LOG_FILE -> log file to output conflicts in case cherry-pick fails, or to output URL to new PR"
+    echo "-o ORIGINAL_PR_URL -> original PR that triggers this script"
+    echo "-r REPOSITORY -> name of the repository"
+    echo "-t TITLE -> title of the original PR"
 }
 
 function cherry_pick {
--- a/toolkit/scripts/cherry_pick.sh
+++ b/toolkit/scripts/cherry_pick.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+set -e
+
+function help {
+    echo "hello, world"
+}
+
+function cherry_pick {
+    commit_hash=$1
+    target_branch=$2
+    log_file=$3
+    pr_title=$4
+    original_pr_url=$5
+    repo=$6
+    tmp_branch="cherry-pick-$target_branch-$commit_hash"
+
+    echo "Commit hash = $commit_hash"
+    echo "Target branch = $target_branch"
+
+    git fetch --all
+    git checkout -b "$tmp_branch" origin/"$target_branch"
+
+    git cherry-pick -x "$commit_hash" || rc=$?
+    if [ ${rc:-0} -ne 0 ]; then
+        echo "Cherry pick failed. Displaying conflicts below"
+        git diff --diff-filter=U
+        exit 1
+    else
+        echo "pushing to remote"
+        git push -u origin "$tmp_branch"
+        echo "done pushing to remote"
+        gh pr create \
+            -B "$target_branch" \
+            -H "$tmp_branch" \
+            --repo $repo \
+            --title "[AUTO-CHERRY-PICK] $pr_title - branch $target_branch" \
+            --body "This is an auto-generated pull request to cherry pick commit $commit_hash to $target_branch. Original PR: $original_pr_url" \
+            > $log_file
+    fi
+}
+
+commit_hash=
+target_branch=
+original_pr_url=
+log_file=
+repo=
+pr_title=
+
+while getopts "b:c:l:o:r:t:" opt; do
+    case ${opt} in
+    b ) target_branch="$OPTARG" ;;
+    c ) commit_hash="$OPTARG" ;;
+    l ) log_file="$OPTARG" ;;
+    o ) original_pr_url="$OPTARG" ;;
+    r ) repo="$OPTARG" ;;
+    t ) pr_title="${OPTARG,,}" ;;
+    ? ) echo -e "ERROR: Invalid option.\n\n"; help; exit 1 ;;
+    esac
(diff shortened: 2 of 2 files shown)


.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,128 @@
...
+          echo "status=$(gh pr view ${{ github.event.issue.number }} --repo ${{ github.repository }} --json state --jq '.state')" >> $GITHUB_OUTPUT
+          echo "base_ref=$(gh pr view ${{ github.event.issue.number }} --repo ${{ github.repository }} --json baseRefName --jq '.baseRefName')" >> $GITHUB_OUTPUT
+          echo "merge_commit=$(gh pr view ${{ github.event.issue.number }} --repo ${{ github.repository }} --json mergeCommit --jq '.mergeCommit.oid')" >> $GITHUB_OUTPUT
+
+      - name: Check that PR satisfies requirements for cherry-pick action and notify if not
+        run: |
+          if [[ "${{ steps.pr_info.outputs.status }}" != "MERGED" ]] || [[ "${{ steps.pr_info.outputs.base_ref }}" != fasttrack/* ]]; then
+            gh pr comment ${{ github.event.issue.number }} \
+              --repo ${{ github.repository }} \
+              --body 'Cherry pick action is only enabled for merged pull requests to `fasttrack/*` branches'
```
PawelWMS, line 50: Can we make it so, that we can add the `/cherry-pick X` comment whenever we want (ever before the PR is merged) but the bot will only execute the cherry-pick after the change gets merged? This should make it harder to forget about the comment.

And ideally: could we cherry-pick any PR merged into the `fast-track/*` branch to its corresponding dev branch? That is the default we're interested in. I wouldn't mind adding an optional comment to skip that cherry-picking for a single PR, though, but that's a P2 at the moment.

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,128 @@
...
+          IFS=", " read -ra target_branches <<< "$branches_list"
+          echo "cherry_pick_target_branches=${target_branches[*]}" >> $GITHUB_ENV
+
+      - name: Configure local git repo
+        run: |
+          git config --local user.email "$GIT_USER_EMAIL"
+          git config --local user.name "$GIT_USER_NAME"
+        env:
+          GIT_USER_EMAIL: "cblmargh@microsoft.com"
+          GIT_USER_NAME: "CBL-Mariner Servicing Account"
```
PawelWMS, line 87: Do we need the `env`?
```suggestion
          git config --local user.email "cblmargh@microsoft.com"
          git config --local user.name "CBL-Mariner Servicing Account"
```

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      # Run cherry-pick for each target branch and create new PR for it. If cherry-pick fails,
+      # output conflicts to the action logs and notify in the original PR.
+      - name: Run cherry-pick action
+        run: |
+          cp toolkit/scripts/cherry_pick.sh ${{ runner.temp }}/cherry_pick.sh
```
PawelWMS, line 93: Nit-pick: would it be possible to make the script runnable from wherever and skip copying?

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,128 @@
...
+              -t "${{ github.event.issue.title }}" \
+              || rc=$?
+            if [ ${rc:-0} -ne 0 ]; then
+              gh pr comment ${{ github.event.issue.number }} \
+                --repo ${{ github.repository }} \
+                --body "Cherry-pick failed for branch $target_branch. See run log for more details: $RUN_URL"
+            else
+              gh pr comment ${{ github.event.issue.number }} \
+                --repo ${{ github.repository }} \
+                --body "Cherry-pick succeeded for branch $target_branch. See pull request $(cat $log_file)"
```
PawelWMS, line 119: Curious: why do we print the contents of the log on a successful cherry-pick?

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,77 @@
...
+
+    git cherry-pick -x "$commit_hash" || rc=$?
+    if [ ${rc:-0} -ne 0 ]; then
+        echo "Cherry pick failed. Displaying conflicts below"
+        git diff --diff-filter=U
+        exit 1
+    else
+        echo "pushing to remote"
+        git push -u origin "$tmp_branch"
+        echo "done pushing to remote"
```
PawelWMS, line 41: Nit-pick to make logging more "neat" and consistent:
```suggestion
        echo "Pushing to remote"
        git push -u origin "$tmp_branch"
        echo "Done pushing to remote"
```

.github/workflows/cherry-pick.yml
trungams: I think it's possible as long as the script is present in all branches of Mariner. This is a solution I thought of when I was trying to test this workflow and have the script only available in my default branch but not the target branches for cherry-pick.

.github/workflows/cherry-pick.yml
trungams: I'm using the log file for multiple purposes there. If cherry-pick fails, the log file will store the merge conflicts details. Otherwise, it's used to store the output of the command `gh pr create ...` above, which is a link to the newly created PR

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,128 @@
...
+          cp toolkit/scripts/cherry_pick.sh ${{ runner.temp }}/cherry_pick.sh
+
+          for target_branch in ${{ env.cherry_pick_target_branches }}; do
+            echo "Cherry picking commit to $target_branch"
+
+            log_file="${{ runner.temp }}/$target_branch.txt"
+
+            git reset --hard
+            git clean -df
+            git checkout -- .
```
PawelWMS, line 102: Curious: why we run these commands separately from the `cherry_pick.sh` script? Similar question for the `echo` above.

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,77 @@
...
+    tmp_branch="cherry-pick-$target_branch-$commit_hash"
+
+    echo "Commit hash = $commit_hash"
+    echo "Target branch = $target_branch"
+
+    git fetch --all
+    git checkout -b "$tmp_branch" origin/"$target_branch"
+
+    git cherry-pick -x "$commit_hash" || rc=$?
+    if [ ${rc:-0} -ne 0 ]; then
```
PawelWMS, line 34: Just an alternative idea: we can drop the `rc` variable and use the command directly in the `if`.
```suggestion
    if ! git cherry-pick -x "$commit_hash"; then
```

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,77 @@
...
+    else
+        echo "pushing to remote"
+        git push -u origin "$tmp_branch"
+        echo "done pushing to remote"
+        gh pr create \
+            -B "$target_branch" \
+            -H "$tmp_branch" \
+            --repo $repo \
+            --title "[AUTO-CHERRY-PICK] $pr_title - branch $target_branch" \
+            --body "This is an auto-generated pull request to cherry pick commit $commit_hash to $target_branch. Original PR: $original_pr_url" \
```
PawelWMS, line 47: Nit-pick: instead of printing the full URL, we could either make a link with `[Original PR.]($original_pr_url)` or just use `#<pr_number>` and have GitHub handle linking for us..

PawelWMS (commented):
In general looks good.

Main ask: have the cherry-pick (ideally) happen automatically for PRs against `fast-track/*` and allow people to add the `/cherry-pick` comment even before the PR is merged.

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,77 @@
...
+    repo=$6
+    tmp_branch="cherry-pick-$target_branch-$commit_hash"
+
+    echo "Commit hash = $commit_hash"
+    echo "Target branch = $target_branch"
+
+    git fetch --all
+    git checkout -b "$tmp_branch" origin/"$target_branch"
+
+    git cherry-pick -x "$commit_hash" || rc=$?
```
mfrw, line 33: Although I appreciate the `-x` here to add a ` (cherry picked from commit 86e3e69e2c10e8dd3c6a3beb6196824211f0ab97)`, I think it will only add value, if we are doing the `cherry-pick`s from a protected long-running branch, which is not transient; If we do a force-push to a branch from which we `cherry-pick`ed, the commit sha's will change and the information we captured in our commit will be invalid.

Are we planning to keep the `fasttrack` branches around without any force pushes ?

.github/workflows/cherry-pick.yml
trungams: Those were workarounds for some issues I encountered when testing this workflow, but now that I think about it, those issues wouldn't really appear here. Will update

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,77 @@
...
+function help {
+    echo "Cherry-pick commit to a specific branch and create Github PR"
+    echo "Usage:"
+    echo "-b TARGET_BRANCH -> target branch to cherry pick commit to"
+    echo "-c COMMIT_HASH"
+    echo "-l LOG_FILE -> log file to output conflicts in case cherry-pick fails, or to output URL to new PR"
+    echo "-o ORIGINAL_PR_URL -> original PR that triggers this script"
+    echo "-r REPOSITORY -> name of the repository"
+    echo "-t TITLE -> title of the original PR"
+}
```
neha170, line 16: It would be nice to have an example usage here. Please also highlight which is an optional argument, and which is required.

trungams: Done

.github/workflows/cherry-pick.yml
trungams: Updated. The log file is no longer needed, and all `gh` commands are moved inside the bash script

.github/workflows/cherry-pick.yml
trungams: I updated the workflow. Now it's triggered whenever a PR to fast-track branch is merged. The workflow will cherry-pick to the corresponding dev branch of the fast-track branch by default (specified by the environment variable `CHERRY_PICK_BRANCH_MAPPING`). And we also have an option to specify additional target branches via comments

.github/workflows/cherry-pick.yml
```diff
@@ -2,126 +2,104 @@
...
-        startsWith(github.event.comment.body, '/cherry-pick ') &&
-        (github.event.comment.author_association == 'MEMBER' ||
-         github.event.comment.author_association == 'OWNER')
+      - closed
+    branches:
+      - 'fasttrack/*'
+    paths:
+      - 'SPECS/**'
+      - 'SPECS-EXTENDED/**'
+      - 'SPECS-SIGNED/**'
```
PawelWMS, line 25: No need to filter by paths - we want to catch all PRs going into the `fasttrack/*` branches.
```suggestion
```

.github/workflows/cherry-pick.yml
trungams: one thing though, is that this action is now triggered on PR merge only, so commenting after merging won't work anymore

.github/workflows/cherry-pick.yml
```diff
@@ -2,126 +2,104 @@
...
+
+permissions:
+  contents: read
+  pull-requests: read
+defaults:
+  run:
+    shell: bash
 
+env:
+  CHERRY_PICK_BRANCH_MAPPING: '{"fasttrack/1.0": "1.0-dev", "fasttrack/2.0": "main"}'
```
PawelWMS, line 35: We don't support 1.0, so we can simplify a bit.
```suggestion
  CHERRY_PICK_BRANCH_MAPPING: '{"fasttrack/2.0": "main"}'
```

.github/workflows/cherry-pick.yml
```diff
@@ -2,126 +2,104 @@
...
+                                       --json comments \
+                                       --jq '.comments
+                                              | map(select(.authorAssociation == "MEMBER" or .authorAssociation == "OWNER")
+                                                    | .body
+                                                    | select(startswith("/cherry-pick "))
+                                                    | ltrimstr("/cherry-pick ")
+                                                    | split(" "))
+                                              | flatten
+                                              | . += ["${{ fromJSON(env.CHERRY_PICK_BRANCH_MAPPING)[github.base_ref] }}"]
+                                              | unique')
```
PawelWMS, line 67: This is somewhat complex - could you add a comment explaining what it does? Ideally with a sample input and output.

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,106 @@
...
+          branches_list=$(echo $cherry_pick_branches_json | jq -r @sh | tr -d "'")
+          IFS=" " read -ra target_branches <<< "$branches_list"
```
PawelWMS, line 69: Just curious: what is the format of `branches_list` that we need the additional step to convert it into an array only to expand it in the next line? Nothing important, I was just looking for a way to save a step.

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          git config --local user.email "cblmargh@microsoft.com"
+          git config --local user.name "CBL-Mariner Servicing Account"
+
+      # Run cherry-pick for each target branch and create new PR for it. If cherry-pick fails,
+      # output conflicts to the action logs and notify in the original PR.
+      - name: Run cherry-pick action
+        run: |
+          for target_branch in ${{ needs.collect_target_branches.outputs.cherry_pick_target_branches }}; do
+          toolkit/scripts/cherry_pick.sh \
```
PawelWMS, line 98:
```suggestion
            toolkit/scripts/cherry_pick.sh \
```

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,106 @@
...
+              -r "${{ github.repository }}" \
+              -p "${{ github.event.pull_request.number }}" \
+              -t "$target_branch" || true
```
PawelWMS, line 101: Why the `|| true`? Shouldn't we fail in case of an error?

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,88 @@
...
+    # create a temporary branch from target branch to perform cherry pick
+    git fetch --all
+    git checkout -b "$tmp_branch" origin/"$target_branch"
+
+    if ! git cherry-pick -x "$commit_hash"; then
+        echo "Cherry pick failed. Displaying conflicts below"
+        git diff --diff-filter=U
+        gh pr comment "$pr_number" \
+            --repo "$repo" \
+            --body "Cherry-pick failed for branch \`$target_branch\`. See run logs for more details: $RUN_URL"
```
PawelWMS, line 48: Please pass the value under `RUN_URL` through an argument. This will make it easier to understand where it's coming from.

.github/workflows/cherry-pick.yml
trungams: `branches_list` is a string containing all the target branch names. I need to separate the names in order to iterate over them

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,106 @@
...
+                                              | . += ["${{ fromJSON(env.CHERRY_PICK_BRANCH_MAPPING)[github.base_ref] }}"]
+                                              | unique')
+          branches_list=$(echo $cherry_pick_branches_json | jq -r @sh | tr -d "'")
+          IFS=" " read -ra target_branches <<< "$branches_list"
+          echo "cherry_pick_target_branches=${target_branches[*]}" >> $GITHUB_OUTPUT
+
+  # Actual cherry-pick work for each target branch
+  cherry_pick_commit:
+    needs: collect_target_branches
+    if: github.event.pull_request.merged == true
```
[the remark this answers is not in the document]
trungams: hmm you're right. It's not needed in this case

.github/workflows/cherry-pick.yml
trungams: The error handling is done inside the script. I want to continue the cherry-pick work for other branches even if one fails, regardless of the reason. Probably need to check for the exit code to fail the workflow run later

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,88 @@
...
+    esac
+done
+
+if [[ -z "$repo" ]] || [[ -z "$pr_number" ]] || [[ -z "$target_branch" ]]; then
+    echo -e "Error: missing required arguments"
+    help
+    exit 1
+fi
+
+collect_pr_info "$repo" "$pr_number"
```
PawelWMS, line 87: This is making it a bit hard to understand where the `commit_hash` and `original_pr_title` variables are suddenly coming from. Since you pass `repo` and `pr_number` into `cherry_pick` anyway, why not set `commit_hash` and `original_pr_title` inside that function? Especially since they seem to be a single line each.

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,88 @@
...
+function cherry_pick {
+    commit_hash=$1
+    target_branch=$2
+    repo=$3
+    pr_number=$4
+    original_pr_title=$5
+
+    tmp_branch="cherry-pick-$target_branch-$commit_hash"
+
+    echo "Cherry picking commit ($commit_hash) to target branch $target_branch"
```
PawelWMS, line 32: For the sake of consistency:
```suggestion
    echo "Cherry picking commit ($commit_hash) to target branch ($target_branch)"
```

And in general I'm personally a fan of surrounding string-like variables with brackets or quotes, because that helps during debugging if for example one variable by accident gets set to an empty string.

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,88 @@
...
+    if ! git cherry-pick -x "$commit_hash"; then
+        echo "Cherry pick failed. Displaying conflicts below"
+        git diff --diff-filter=U
+        gh pr comment "$pr_number" \
+            --repo "$repo" \
+            --body "Cherry-pick failed for branch \`$target_branch\`. See run logs for more details: $RUN_URL"
+    else
+        echo "pushing to remote"
+        git push -u origin "$tmp_branch"
+        echo "done pushing to remote"
```
PawelWMS, line 52: Nit-pick: could we unify the logging convention? It looks like capital letters and no period at the end seems to be the most popular in this script.
```suggestion
        echo "Pushing to remote"
        git push -u origin "$tmp_branch"
        echo "Done pushing to remote"
```

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,88 @@
...
+        new_pr=$(gh pr create \
+            -B "$target_branch" \
+            -H "$tmp_branch" \
+            --repo "$repo" \
+            --title "[AUTO-CHERRY-PICK] $original_pr_title - branch $target_branch" \
+            --body "This is an auto-generated pull request to cherry pick commit $commit_hash to $target_branch. Original PR: #$pr_number")
+        gh pr comment "$pr_number" \
+            --repo "$repo" \
+            --body "Cherry-pick succeeded for branch \`$target_branch\`. See pull request #$new_pr"
```
PawelWMS, line 61: Just a suggestion: since we're using `gh pr comment` in two places, we could extract it into a function to not have to repeat the whole call twice and only pass the PR number, repo, and message to the function.

toolkit/scripts/cherry_pick.sh
```diff
@@ -0,0 +1,88 @@
...
+        gh pr edit "$pr_number" \
+            --repo "$repo" \
+            --add-label "cherry_pick-$target_branch"
+    fi
```
PawelWMS, line 65: Let's make sure the script errors out in case of a cherry-pick (or any other) failure. In general, it's good to inform the user the script didn't do what it was supposed to do and leave it up to them to decide how to proceed (stop, ignore, log, etc.). And since this is inside a function, I'd suggest using `return` instead of `exit`, so that the caller of the function has the same possibility to decide what to do with an error. Not a huge difference in case of this script, though.

```suggestion
    if ! git cherry-pick -x "$commit_hash"; then
        echo "Cherry pick failed. Displaying conflicts below"
        git diff --diff-filter=U
        gh pr comment "$pr_number" \
            --repo "$repo" \
            --body "Cherry-pick failed for branch \`$target_branch\`. See run logs for more details: $RUN_URL"
        return 1
    fi

    echo "pushing to remote"
    git push -u origin "$tmp_branch"
    echo "done pushing to remote"
    new_pr=$(gh pr create \
        -B "$target_branch" \
        -H "$tmp_branch" \
        --repo "$repo" \
        --title "[AUTO-CHERRY-PICK] $original_pr_title - branch $target_branch" \
        --body "This is an auto-generated pull request to cherry pick commit $commit_hash to $target_branch. Original PR: #$pr_number")
    gh pr comment "$pr_number" \
        --repo "$repo" \
        --body "Cherry-pick succeeded for branch \`$target_branch\`. See pull request #$new_pr"
    gh pr edit "$pr_number" \
        --repo "$repo" \
        --add-label "cherry_pick-$target_branch"
```

.github/workflows/cherry-pick.yml
PawelWMS: And if we want to continue, please add a comment explaining why this is the preference. This may help a future developer looking at this piece of code.

PawelWMS: Your approach makes sense. And I agree, that we should save the failure information to fail the whole action in the end.

.github/workflows/cherry-pick.yml
PawelWMS: But in what format? As in: separate by spaces, new lines, it's a JSON, etc.?:) What I was getting at, is if it's already a space-separated list, then I think we can pass it directly to the `echo` below and skip the step where we create the `target_branches` array.

trungams: It is a space-separated list, e.g `"1.0-dev main"`. If I try to use a for loop to iterate over the tokens of `branches_list`, it won't work, i.e it treats `"1.0-dev main"` as a single item. I don't think flushing to the GITHUB_OUTPUT file would change that :/

toolkit/scripts/cherry_pick.sh
trungams: it's not a too complex command so I think keeping this way helps with clarity

.github/workflows/cherry-pick.yml
trungams: hmm actually it worked. Bash still confuses me sometimes

toolkit/scripts/cherry_pick.sh
trungams: Updated. I still have the RUN_URL listed in the environment variable for clarity, since it needs to be constructed from multiple variables

.github/workflows/cherry-pick.yml
trungams: Added some explanation in the comment along with examples of how the json object transforms

.github/workflows/cherry-pick.yml
```diff
@@ -45,7 +45,25 @@ jobs:
...
+      # - Remove the "/cherry-pick " prefix from each comment
+      #     ["a b c", "a b"]
+      # - Split the remaining content of each comment by whitespace into a list
+      #     [["a", "b", "c"], ["a", "b"]]
+      # - Flatten the list
+      #     ["a", "b", "c", "a", "b"]
+      # - Add the default target branch to the list
+      #     ["a", "b", "c", "a", "b", "default-target"]
+      # - Remove all duplicate values
+      #     ["a", "b", "c", "default-target"]
```
PawelWMS, line 66: Lovely.:)

toolkit/scripts/cherry_pick.sh
PawelWMS: Yeah, after a second look I agree.

PawelWMS (approved):

PR #6113, opened 2023-08-30
Added ptest results parsing to the `PackageBuild` template

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

In addition to running the package tests, we want to display the results of these tests in the `Tests` tab of each build and publish the results through the pipeline artifacts to be consumed by other tools.

###### Change Log  

- Added the `PackageTestResultsAnalysis.yml` template to handle parsing of the test results.

No.

###### Test Methodology

- [Sample PR check run.](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=415611&view=results)


.pipelines/templates/PackageTestResultsAnalysis.yml
```diff
@@ -0,0 +1,295 @@
...
+
+  - bash: |
+      pip3 install junit_xml
+    displayName: "Install Python dependencies"
+
+  - task: PythonScript@0
+    inputs:
+      pythonInterpreter: "/bin/python3"
+      scriptSource: "inline"
+      script: |
```
mfrw, line 45: Should we probably have a file instead of a large script ?
We could then call that script within this

toolkit/tools/scheduler/schedulerutils/buildworker.go
```diff
@@ -284,7 +284,7 @@ func testSRPMFile(agent buildagents.BuildAgent, checkAttempts int, srpmFile stri
 
 		_, logFile, buildErr = agent.BuildPackage(srpmFile, logBaseName, outArch, runCheck, dependencies)
 		if buildErr != nil {
-			logger.Log.Warnf("Test build for '%s' failed on a non-test build issue. Error: %s", srpmFile, err)
+			logger.Log.Warnf("Test build for '%s' failed on a non-test build issue. Error: %s", srpmFile, buildErr)
```
mfrw, line 287: Nice Catch :)

PawelWMS: Yeah, I created this bug myself, so I guess I should fix it...:)

.pipelines/templates/PackageTestResultsAnalysis.yml
PawelWMS: I'm trying to avoid scripts inside all GitHub templates as they are made to be ran on `steps` level and thus cannot check-out their own repositories and access files in them. I agree, that I'd prefer to have a script in a separate file, but so far I couldn't find any good way to make it not require us to now clone the whole repo in the pipelines just to get access to the script file. Expanding pipeline templates doesn't require cloning the repo, thus inlining a script works.

One more idea that I had on my mind, would be to enclose this script inside its own separate pipeline template to make it more reusable in other pipelines. That still didn't make it possible to run the script manually, though, and added additional complexity.

Having said that, I am open to ideas! I'd love to learn if there's a better way.

mfrw: Ohh, I am not aware if there is a better way to do it.
It felt a little off, which is why the comment :)

mfrw (approved):

PR #6114, opened 2023-08-31
containerized-rpmbuild: use macros from Makefile to set folder definitions

neha170:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Use macros from Makefile to set folder definitions in containerized-rpmbuild

###### Change Log  
- This change was required because if the user passed a build macro like 'TOOL_BINS_DIR' while invoking the make command for containerized-rpmbuild, the tool broke because of hardcoded values. So, we need to remove hardcoded values, use macros as defined in Makefile to set folder default definitions, and be able to override those definitions when provided by the user

Changes:
- Use macros as defined in the Mariner toolkit Makefile to set the defaults
- If the user provides any of these in the make command, they're automatically overridden
- So far, I have kept the number of macros to a minimum set that is used within the script. If the tool gets additional features in the future, we might need to expand the list of folders we're defining

**NO**

###### Test Methodology

- Was able to override toolkit default macros by providing them in the make containerized-rpmbuild command. If not provided, the script constructed a container with the default values.

### DIFF
--- a/toolkit/scripts/containerized-build/create_container_build.sh
+++ b/toolkit/scripts/containerized-build/create_container_build.sh
@@ -42,14 +42,14 @@ To see help, run 'sudo make containerized-rpmbuild-help'
 }
 
 build_chroot() {
-    pushd "${repo_path}/toolkit"
+    pushd $toolkit_root
     echo "Building worker chroot..."
     make graph-cache REBUILD_TOOLS=y > /dev/null
     popd
 }
 
 build_tools() {
-    pushd "${repo_path}/toolkit"
+    pushd $toolkit_root
     echo "Building required tools..."
     make go-srpmpacker REBUILD_TOOLS=y > /dev/null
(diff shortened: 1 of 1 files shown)


toolkit/scripts/containerized-build/create_container_build.sh
```diff
@@ -95,6 +95,19 @@ done
 [[ -z "${mode}" ]] && mode="build"
 [[ -z "${version}" ]] && version="2.0"
 
+# Set relevant folder definitions using Makefile that can be overriden by user
```
dmcilvaney, line 98: This will work well if we are just pointing the tool at a non-standard path, but I'm curious if the user sets `SPECS_DIR=/path/to/my/specs` when calling the container tool if it will inherit here. When the recursive call to make happens I'm curious what environment it sees (since we were called from make, does it inherit it?)

PR #6143, opened 2023-09-04
Cleaned-up invalid edges from duplicate nodes.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Whenever we found multiple packages providing the same capability (for instance `bundled(gnulib)`), we would only add a run-build(-test) set of nodes for it once. Every other package providing the same capability would generate a warning but not a new node. However, when we later generated graph edges to indicate node dependencies, we would read the dependencies for all the packages providing the duplicate capability and attach them to the original node. This caused some packages to be unnecessarily re-built.

Example: `augeas` and `libguestfs` provide `bundled(gnulib)`. We only add a node for the run and build nodes of the `augeas` version of `bundled(gnulib)`. However, later we read the dependencies of the `libguestfs` version of it and connect them with the `augeas` version. If we then built `vim`, we would end up with a rebuild of `libguestfs` (correct, it `BuildRequires` `vim`) and `augeas` (incorrect, the `bundled(gnulib)` node depended on the `BuildRequires` from `libguestfs` despite coming from the SRPM for `augeas`). To add to the confusion, `augeas` would be shown to be built twice: first when nodes belonging to its SRPM were unblocked and later, when all the `BuildRequires` for the remaining `bundled(gnulib)` nodes were satisfied.

###### Change Log  

- Stop adding dependency edges for duplicate package capabilities.
- Removed unnecessary new lines from debug logs.

No.

###### Test Methodology

- Local package builds.

### DIFF
--- a/toolkit/tools/grapher/grapher.go
+++ b/toolkit/tools/grapher/grapher.go
@@ -137,7 +137,7 @@ func addUnresolvedPackage(g *pkggraph.PkgGraph, pkgVer *pkgjson.PackageVer) (new
 		return
 	}
 
-	logger.Log.Infof("Adding unresolved node %s\n", newRunNode.FriendlyName())
+	logger.Log.Infof("Adding unresolved node %s.", newRunNode.FriendlyName())
 
 	return
 }
@@ -145,7 +145,7 @@ func addUnresolvedPackage(g *pkggraph.PkgGraph, pkgVer *pkgjson.PackageVer) (new
 // addNodesForPackage creates a "Run", "Build", and "Test" node for the package described
 // in the Package structure. Returns pointers to the build and run Nodes
 // created, or an error if one of the nodes could not be created.
-func addNodesForPackage(g *pkggraph.PkgGraph, pkg *pkgjson.Package) (err error) {
+func addNodesForPackage(g *pkggraph.PkgGraph, pkg *pkgjson.Package) (foundDuplicate bool, err error) {
 	var (
 		newRunNode   *pkggraph.PkgNode
 		newBuildNode *pkggraph.PkgNode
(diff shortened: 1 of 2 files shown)


toolkit/tools/grapher/grapher.go
```diff
@@ -137,15 +137,15 @@ func addUnresolvedPackage(g *pkggraph.PkgGraph, pkgVer *pkgjson.PackageVer) (new
 		return
 	}
 
-	logger.Log.Infof("Adding unresolved node %s\n", newRunNode.FriendlyName())
+	logger.Log.Infof("Adding unresolved node %s.", newRunNode.FriendlyName())
```
anphel31, line 140: nit: I don't think the `.` is necessary/desired here

toolkit/tools/grapher/grapher.go
```diff
@@ -347,8 +338,7 @@ func populateGraph(graph *pkggraph.PkgGraph, repo *pkgjson.PackageRepo) (err err
 	// Rescan and add all the dependencies
 	logger.Log.Infof("Adding all dependencies from %s", *input)
 	dependenciesAdded := 0
-	for idx := range packages {
-		pkg := packages[idx]
+	for pkg := range uniquePackages {
```
anphel31, line 341: The same variable name `pkg` was used above; I wonder if it would be better to have a unique name here

anphel31 (approved):

toolkit/tools/grapher/grapher.go
PawelWMS: Good idea, I'll update that in a minute!

toolkit/tools/grapher/grapher.go
PawelWMS: I treated this as a regular sentence (although I think I should also surround the node's name in quotes), thus the period. How do you see it?

anphel31: I just thought it was confusing, since it's connected to the node name. Adding quotes would also help

PawelWMS: Discussed offline, added quotes to make it clearer the period is not part of the node's name.

PR #6200, opened 2023-09-08
Add `help` target to toolkit Makefile

reubeno:
###### Merge Checklist

- [x] Ready to merge

---

###### Summary 
This change introduces a toplevel `make help` target to the toolkit Makefile. The documentation in tree has volumes of content on the various targets and variables and various build-related concepts, but there's value in a more terse summary of the syntax such as a standard command-line tool would have (i.e., `--help`).

###### Change Log  
- Introduces a basic pattern for `##help` comments that allow expressing the help content for well-known variables and targets at their respective definition sites.
- Adds `help` Makefile target that extracts the `##help` comments via sed/grep/awk and does a simple rendering of it to stdout.
- Adds `help.mk`, which documents the format of these comments and also includes the toplevel help preamble.
- Populates a set of initial `##help` comments for commonly used targets and variables. The intention is for this to be easy to iterate on in-tree over time.

No

###### Test Methodology
Local build


reubeno:
Here is a screenshot of sample output from `make help`:

image

Feedback is, of course, welcome on any and all of the changes -- I'm most specifically looking for feedback on the format of the `##help` comments as well as the admittedly awkward `grep`/`awk`/`sed` logic I used to render the help output.

toolkit/scripts/help.mk
```diff
@@ -0,0 +1,106 @@
...
+#
+# Examples.
+#
+
+##help:example:Build only my-package.spec and its requirements:
+##help:example:    [3msudo make -j$(nproc) build-packages SRPM_PACK_LIST="my-package"[0m
+##help:example:
+
+##help:example:Force rebuild my-package.spec after having already built it:
+##help:example:    [3msudo make -j$(nproc) build-packages SRPM_PACK_LIST="my-package" PACKAGE_REBUILD_LIST="my-package" REFRESH_WORKER_CHROOT=n[0m
```
PawelWMS, line 101: Regarding the `REFRESH_WORKER_CHROOT` argument: I admit I still use it, but it was added as a workaround until we improve our tooling and I'm not sure, if it's still needed. @dmcilvaney, do you remember if the toolkit is now smart enough to not rebuild the chroot, if it doesn't need to? Or if the `QUICK_REBUILD_*` arguments wouldn't do it.

toolkit/scripts/help.mk
```diff
@@ -0,0 +1,106 @@
...
+
+##help:example-heading:
+##help:example-heading:[1;4;34mExamples:[0m
+
+#
+# Examples.
+#
+
+##help:example:Build only my-package.spec and its requirements:
+##help:example:    [3msudo make -j$(nproc) build-packages SRPM_PACK_LIST="my-package"[0m
```
PawelWMS, line 97: I think we should mention the `QUICK_REBUILD_*` set of arguments, as they were designed to reflect common build cases by changing the defaults of some of the more obscure flags under the hood. There's more details about their purpose inside `building.md` and `incremental_building.mk`, but if that's not enough, @dmcilvaney is an expert on them.

toolkit/scripts/help.mk
reubeno: I'm happy to remove it if y'all think that's appropriate. I took a cue from its documentation in `toolkit/docs/building/building.md`.

toolkit/Makefile
```diff
@@ -12,14 +12,17 @@ toolkit_root := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 SCRIPTS_DIR  ?= $(toolkit_root)/scripts
 
 # Build targets
+##help:var:CONFIG_FILE:<config_path>=Path to image configuration file to use.
 ifeq ($(origin CONFIG_FILE), undefined)
 CONFIG_FILE             = $(toolkit_root)/imageconfigs/core-efi.json
-$(warning CONFIG_FILE is undefined, defaulting to toolkit's core-efi.json.)
+#$(warning CONFIG_FILE is undefined, defaulting to toolkit's core-efi.json.)
```
PawelWMS, line 18: While I also want' to get rid of this warning, I think we shouldn't leave the comment behind. Also, I think @dmcilvaney had a separate change dedicated to this clean-up (I might be misremembering something here, though).

toolkit/scripts/help.mk
reubeno: I'll admit -- I'm not clear when I should and shouldn't use it. If i'm a human being and not an ADO pipeline, should I *always* use `QUICK_REBUILD=y`? If no, then some succinct pointers on when to use it would help. If yes, then is it really `I_AM_A_HUMAN_BEING_AND_WANT_MAKE_TO_GO_FASTER=y`? :) (Since we're using Makefile as a UI to the build, I'd strongly urge us to tune its defaults to the dev case rather than the pipeline case; the pipelines are going to explicitly specify exactly-the-right-set-of-args anyhow.)

toolkit/Makefile
```diff
@@ -12,14 +12,17 @@ toolkit_root := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
...
 endif
 CONFIG_BASE_DIR         ?= $(dir $(CONFIG_FILE))
 PACKAGE_BUILD_LIST      ?=
+##help:var:PACKAGE_REBUILD_LIST:<spec_list>=List of spec basenames to force rebuild.
```
PawelWMS, line 22: A small tweak, so people don't accidentally try to use `kernel.spec` as a valid argument + a mention how to input multiple specs:
```suggestion
##help:var:PACKAGE_REBUILD_LIST:<spec_list>=List of space-separated spec folders to force rebuild. Example: PACKAGE_REBUILD_LIST="kernel go which".
```

toolkit/Makefile
```diff
@@ -12,14 +12,17 @@ toolkit_root := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
...
 PACKAGE_REBUILD_LIST    ?=
 PACKAGE_IGNORE_LIST     ?=
+##help:var:SRPM_PACK_LIST:<spec_list>=List of spec basenames to build into SRPMs; if empty, all found specs will be packed.
```
PawelWMS, line 25: Similar comment as for `PACKAGE_REBUILD_LIST`:
```suggestion
##help:var:SRPM_PACK_LIST:<spec_list>=List of space-separated spec folders inside "SPECS_DIR" to analyze for the build. If empty, all specs inside the "SPECS_DIR" will be analyzed. Example: SRPM_PACK_LIST="kernel go which".
```

toolkit/Makefile
reubeno: Okay, fixed in latest commit.

toolkit/Makefile
```diff
@@ -12,14 +12,17 @@ toolkit_root := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
...
+##help:var:CONFIG_FILE:<config_path>=Path to image configuration file to use.
 ifeq ($(origin CONFIG_FILE), undefined)
 CONFIG_FILE             = $(toolkit_root)/imageconfigs/core-efi.json
 $(warning CONFIG_FILE is undefined, defaulting to toolkit's core-efi.json.)
 endif
 CONFIG_BASE_DIR         ?= $(dir $(CONFIG_FILE))
 PACKAGE_BUILD_LIST      ?=
+##help:var:PACKAGE_REBUILD_LIST:<spec_list>=List of spec basenames to force rebuild.
 PACKAGE_REBUILD_LIST    ?=
 PACKAGE_IGNORE_LIST     ?=
```
PawelWMS, line 24:
```suggestion
##help:var:PACKAGE_IGNORE_LIST:<spec_list>=List of space-separated spec folders to ignore during the build. Must not overlap with "PACKAGE_REBUILD_LIST", may overlap with "PACKAGE_BUILD_LIST". Example: PACKAGE_REBUILD_LIST="kernel go which".
PACKAGE_IGNORE_LIST     ?=
```

toolkit/Makefile
```diff
@@ -12,14 +12,17 @@ toolkit_root := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 SCRIPTS_DIR  ?= $(toolkit_root)/scripts
 
 # Build targets
+##help:var:CONFIG_FILE:<config_path>=Path to image configuration file to use.
 ifeq ($(origin CONFIG_FILE), undefined)
 CONFIG_FILE             = $(toolkit_root)/imageconfigs/core-efi.json
 $(warning CONFIG_FILE is undefined, defaulting to toolkit's core-efi.json.)
 endif
 CONFIG_BASE_DIR         ?= $(dir $(CONFIG_FILE))
 PACKAGE_BUILD_LIST      ?=
```
PawelWMS, line 21:
```suggestion
##help:var:PACKAGE_BUILD_LIST:<spec_list>=List of space-separated spec folders to consider for building. If empty, all packages from "SRPM_PACK_LIST" will be considered. Will not re-build previously built packages.. Example: PACKAGE_BUILD_LIST="kernel go which".
PACKAGE_BUILD_LIST      ?=
```

toolkit/Makefile
```diff
@@ -12,14 +12,17 @@ toolkit_root := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
...
+##help:var:PACKAGE_REBUILD_LIST:<spec_list>=List of spec basenames to force rebuild.
 PACKAGE_REBUILD_LIST    ?=
 PACKAGE_IGNORE_LIST     ?=
+##help:var:SRPM_PACK_LIST:<spec_list>=List of spec basenames to build into SRPMs; if empty, all found specs will be packed.
 SRPM_PACK_LIST          ?=
 TEST_RUN_LIST           ?=
```
PawelWMS, line 27: ##help:var:TEST_RUN_LIST:<spec_list>=List of space-separated spec folders to consider for package tests. Specs from the listed folders MUST contain the "%check" section. If empty, all testable items from "SRPM_PACK_LIST" will be considered. Will not re-test previously built packages. Example: TEST_RUN_LIST="libguestfs zlib".

PawelWMS, line 27:
```suggestion
##help:var:TEST_RUN_LIST:<spec_list>=List of space-separated spec folders to consider for package tests. Specs from the listed folders MUST contain the "%check" section. If empty, all testable items from "SRPM_PACK_LIST" will be considered. Will not re-test previously built packages. Example: TEST_RUN_LIST="libguestfs zlib".
TEST_RUN_LIST           ?=
```

toolkit/scripts/help.mk
dmcilvaney: I think it shouldn't really be needed anymore, if you are still seeing spurious rebuilds let me know, could be a last few bugs.

toolkit/Makefile
```diff
@@ -12,14 +12,17 @@ toolkit_root := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
...
 TEST_RERUN_LIST         ?=
```
PawelWMS, line 28:
```suggestion
##help:var:TEST_RERUN_LIST:<spec_list>=List of space-separated spec folders to force running a package test for. Specs from the listed folders MUST contain the "%check" section. Must not overlap with "TEST_IGNORE_LIST". Example: TEST_RERUN_LIST="libguestfs zlib".
TEST_RERUN_LIST         ?=
##help:var:TEST_IGNORE_LIST:<spec_list>=List of space-separated spec folders to ignore for package tests. Must not overlap with "TEST_RERUN_LIST", may overlap with "TEST_RUN_LIST". Example: TEST_IGNORE_LIST="acl".
```

PawelWMS: I can't add a suggestion over that code, but please also add a line for `RUN_CHECK`.

toolkit/scripts/help.mk
dmcilvaney: I've been tyring to find time to optimize the defaults for users, but features keep taking up my focus.

The big outstanding issue before I turn that on is fixing the toolchain build. Currently if you don't know to always set `CLEAN_TOOLCHAIN_CONTAINERS=n` every time you run `make clean` setting QUICK_REBUILD_TOOLCHAIN=y (which is set by the higher level QUICK_REBUILD=y flag) will hit you with a >1hr build penalty for almost every type of build as it regenerates the bootstrap container (which it might not even need...).

toolkit/Makefile
```diff
@@ -12,14 +12,17 @@ toolkit_root := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 SCRIPTS_DIR  ?= $(toolkit_root)/scripts
 
 # Build targets
+##help:var:CONFIG_FILE:<config_path>=Path to image configuration file to use.
```
dmcilvaney, line 15:
```suggestion
##help:var:CONFIG_FILE:<config_path>=Path to image configuration file to use. Will add package dependencies and define final image.
```

Not sure I like my wording, but would be nice to call out its affects on the package scheduler specifically (the image build part is fairly self explanatory)

toolkit/Makefile
dmcilvaney: I'd also add something like "exclusively analyze" here to indicate we won't look at the other packages.

toolkit/scripts/help.mk
dmcilvaney: (fyi, I have https://github.com/microsoft/CBL-Mariner/pull/5218 but it is now out-of-date. I've got a task to refresh this once we finish the parallel build upgrades).

reubeno: I completely understand the time constraints. It sounds like there's work to be done before certain semantics can be turned on by default... and I can appreciate that. But is there a way to separate that from the human interface? In other words, can we front-load agreement once and for all on what devs with a 100 or 200-level familiarity with the toolkit should put (or should *not* put) on the command line to request "fast local incremental builds"? Once that interface is set, we can presumably change what that maps to underneath the hood, in terms of which individual build features are enabled/disabled.

toolkit/scripts/incremental_building.mk
```diff
@@ -6,6 +6,7 @@
 
 ######## QUICK_REBUILD AUTO CONFIGURE ########
 
+##help:var:QUICK_REBUILD:{y,n}=Use as many packages as possible from upstream repos (for both package building and toolchain creation).
```
dmcilvaney, line 9:
```suggestion
##help:var:QUICK_REBUILD:{y,n}=Optimize the build for speed by using existing published components and optimizing unimpactful package rebuilds (Implies QUICK_REBUILD_PACKAGES=y, QUICK_REBUILD_TOOLCHAIN=y).
##help:var:QUICK_REBUILD_PACKAGES:{y,n}=Use as many packages as possible from upstream repos, limit cascading rebuilds when a dependency is rebuilt.
##help:var:QUICK_REBUILD_TOOLCHAIN:{y,n}=Rebuild the toolchain, but attempt to download components where possible.
```

Just breaking these apart for now.`QUICK_REBUILD_PACKAGES` is probably more useufl for day-to-day use right now.

toolkit/scripts/pkggen.mk
```diff
@@ -231,6 +232,7 @@ srpms_archive  	= $(OUT_DIR)/srpms.tar.gz
 
 .PHONY: build-packages clean-build-packages hydrate-rpms compress-rpms clean-compress-rpms compress-srpms clean-compress-srpms
 
+##help:target:build-packages=Build requested .rpm files.
```
dmcilvaney, line 235:
```suggestion
##help:target:build-packages=Build requested .rpm files from PACKAGE_(RE)BUILD_LIST=, and IMAGE_CONFIG=.
```

Thoughts? Not sure this is needed.

toolkit/scripts/help.mk
reubeno: I'll remove it from the examples in this PR.

toolkit/scripts/srpm_pack.mk
```diff
@@ -40,6 +40,7 @@ $(call create_folder,$(BUILD_SRPMS_DIR))
 $(call create_folder,$(SRPM_BUILD_CHROOT_DIR))
 
 # General targets
+##help:target:input-srpms=Scan the local `*.spec` files, locate sources, and create `*.src.rpm` files.
```
dmcilvaney, line 43:
```suggestion
##help:target:input-srpms=Scan the local `*.spec` files, locate sources, and create `*.src.rpm` files. Limit via SRPM_PACK_LIST.
```

This one is really powerful, good to call out how to use it I think.

toolkit/scripts/pkggen.mk
reubeno: There's probably a balance to be struck here between the tersest reminder of what the targets are called and sufficient context to be able to understand how to use them just from this output. It may help more to include motivating examples and instruct the reader to `(see examples below)`.

toolkit/Makefile
dmcilvaney: Yup, I have a PR that removes this nonsense finally. Probably one of the biggest sources of long builds for users. Will just default to "" going forward.

toolkit/scripts/help.mk
reubeno: Anyhow, I'm happy to chat about this broader discussion separately. For the purposes of this PR, @dmcilvaney , which examples -- if any -- of `QUICK_REBUILD*` would you recommend including in the help examples? I'm also completely okay pending that for a future PR if you'd prefer that we do so.

mfrw (commented):
My request would be add the `containerized-build` also as an output of the `make help`.

Mostly folks who would run this target would be the ones who may not know about all the details and would expect the `make help` to be an exaustive list of all the targets, which in my opinion is not :)

toolkit/scripts/help.mk
dmcilvaney: I'm going to refresh the build docs, why don't I do this at the same time?

dmcilvaney (approved):

toolkit/scripts/help.mk
reubeno: That sounds great. I think you'll be able to compose better verbiage than I would for it anyhow.

neha170:
Do we want to add this target to the documentation?
https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/docs/building/building.md

toolkit/scripts/toolchain.mk
```diff
@@ -49,6 +49,8 @@ $(call create_folder,$(toolchain_downloads_logs_dir))
 $(call create_folder,$(toolchain_from_repos))
 $(call create_folder,$(populated_toolchain_chroot))
 
+##help:target:raw-toolchain=Build the initial toolchain bootstrap stage.
+##help:target:toolchain=Ensure all toolchain RPMs are present.
```
PawelWMS, line 53: This is one comment but applies to other occurrences in the PR:

What do you think about adding these comments directly above the target they refer to? This way it's going to be harder to "misplace" them if we move some targets around. For the sake of making it easier to read, I'm all for adding empty lines between different targets as well.

PawelWMS (approved):
Looks good. Just one more comment but not a blocker.

mfrw (approved):
LGTM :)
Thanks for adding the `containerized-build` target

reubeno:
> Do we want to add this target to the documentation? https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/docs/building/building.md

Added!

reubeno:
@cwize1 -- I'm seeing go test errors like the one below, despite this PR not touching any go code. (I spot some other PRs hitting the same error.) It looks like this is one of the new tests added by your recent image customizer PR. Is this something you've also seen?

```
ERRO[0001][main_test] Mount of (/dev/loop3p2) to (/home/runner/work/CBL-Mariner/CBL-Mariner/toolkit/tools/pkg/imagecustomizerlib/_tmp/TestCustomizeImageCopyFiles/imageroot) failed. Error: no such file or directory 
```

oliviacrain:
> @cwize1 Chris Gunn (MARINER) FTE -- I'm seeing go test errors like the one below, despite this PR not touching any go code.

I believe PR checks run on the theoretical merge commit created by a PR being merged, rather than the latest commit on the branch being pulled from.

reubeno:
> > @cwize1 Chris Gunn (MARINER) FTE -- I'm seeing go test errors like the one below, despite this PR not touching any go code.
> 
> I believe PR checks run on the theoretical merge commit created by a PR being merged, rather than the latest commit on the branch being pulled from.

Understood. I was trying to politely say that I think tests are broken in main :smile:

PR #6216, opened 2023-09-14
Add a fallback lazy mode to chroot unmount

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Sometimes the chroots we use do not unmount correctly. There seem to be many different reasons this can occur (but `/dev/hugepages` has been a recent culprit). Generally, we try to gracefully unmount them when we call `.Close()`, but the `safechroot` package tracks active chroots and calls `cleanupAllChroots()` when the tool exits. Currently we will make an attempt to clean, but if the unmount fails (often due to a busy mount) we give up and leave the mountpoint present. If the user isn't cautious and deletes this folder, it can mess up the host system until a reboot.

Linux offers a lazy unmount option. This will detatch the mountpoint from the fileystem even if it can't fully tear down the mount. This means that at least the user won't be able to damage the host system accidentily. If we use this lazy detach mode in our final cleanup in `cleanupAllChroots()` we have a better chance of leaving the system in a safe state.

###### Change Log  

- Use `unix.MNT_DETACH` on last chroot unmount attempt
- Make failure on final unmount explicilty a failure via `Fatalf(...)`
- Use exponential backoff for unmount so we try longer

**NO**

###### Test Methodology

- Local builds

### DIFF
--- a/toolkit/tools/internal/safechroot/safechroot.go
+++ b/toolkit/tools/internal/safechroot/safechroot.go
@@ -446,6 +446,7 @@ func cleanupAllChroots() {
 	inChrootMutex.Lock()
 
 	// mount is only supported in regular pipeline
+	failedToUnmount := false
 	if buildpipeline.IsRegularBuild() {
 		// Cleanup chroots in LIFO order incase any are interdependent (e.g. nested safe chroots)
 		logger.Log.Info("Cleaning up all active chroots")
@@ -456,11 +457,16 @@ func cleanupAllChroots() {
 			// even if one fails.
 			if err != nil {
 				logger.Log.Errorf("Failed to unmount chroot (%s)", activeChroots[i].rootDir)
+				failedToUnmount = true
 			}
 		}
 	}
 
-	logger.Log.Info("Cleanup finished")
+	if failedToUnmount {
+		logger.Log.Fatalf("Failed to unmount a chroot, manual unmount required")
+	} else {
+		logger.Log.Info("Cleanup finished")
+	}
 }
 
 // unmountAndRemove retries to unmount directories that were mounted into
@@ -507,11 +513,7 @@ func (c *Chroot) unmountAndRemove(leaveOnDisk, isFinalAttempt bool) (err error) 
 		}, totalAttempts, retryDuration, 2.0, nil)
 
 		if err != nil {
-			if isFinalAttempt {
-				logger.Log.Fatalf(errMsg, fullPath, err)
-			} else {
-				logger.Log.Warnf(errMsg, fullPath, err)
-			}
+			logger.Log.Warnf(errMsg, fullPath, err)
 			return
 		}
 	}
--- a/toolkit/tools/internal/safechroot/safechroot.go
+++ b/toolkit/tools/internal/safechroot/safechroot.go
@@ -82,6 +82,11 @@ var defaultChrootEnv = []string{
 	"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
 }
 
+const (
+	doLazyUnmount   = true
+	doNormalUnmount = !doLazyUnmount
+)
+
 // init will always be called if this package is loaded
 func init() {
 	registerSIGTERMCleanup()
@@ -196,7 +201,7 @@ func (c *Chroot) Initialize(tarPath string, extraDirectories []string, extraMoun
 			if buildpipeline.IsRegularBuild() {
 				// mount/unmount is only supported in regular pipeline
 				// Best effort cleanup in case mountpoint creation failed mid-way through. We will not try again so treat as final attempt.
-				cleanupErr := c.unmountAndRemove(leaveChrootOnDisk, true)
+				cleanupErr := c.unmountAndRemove(leaveChrootOnDisk, doLazyUnmount)
 				if cleanupErr != nil {
 					logger.Log.Warnf("Failed to cleanup chroot (%s) during failed initialization. Error: %s", c.rootDir, cleanupErr)
 				}
@@ -371,10 +376,10 @@ func (c *Chroot) Close(leaveOnDisk bool) (err error) {
 
 	if buildpipeline.IsRegularBuild() {
 		// mount is only supported in regular pipeline
-		err = c.unmountAndRemove(leaveOnDisk, false)
+		err = c.unmountAndRemove(leaveOnDisk, doNormalUnmount)
 		if err != nil {
 			logger.Log.Warnf("Chroot cleanup failed, will retry with lazy unmount. Error: %s", err)
-			err = c.unmountAndRemove(leaveOnDisk, true)
+			err = c.unmountAndRemove(leaveOnDisk, doLazyUnmount)
 		}
 		if err == nil {
 			const emptyLen = 0
@@ -452,7 +457,7 @@ func cleanupAllChroots() {
 		logger.Log.Info("Cleaning up all active chroots")
 		for i := len(activeChroots) - 1; i >= 0; i-- {
 			logger.Log.Infof("Cleaning up chroot (%s)", activeChroots[i].rootDir)
-			err := activeChroots[i].unmountAndRemove(leaveChrootOnDisk, true)
+			err := activeChroots[i].unmountAndRemove(leaveChrootOnDisk, doLazyUnmount)
 			// Perform best effort cleanup: unmount as many chroots as possible,
 			// even if one fails.
 			if err != nil {
@@ -473,21 +478,22 @@ func cleanupAllChroots() {
 // the chroot until the unmounts succeed or too many failed attempts.
 // This is to avoid leaving folders like /dev mounted when the chroot folder is forcefully deleted in cleanup.
 // Iff all mounts were successfully unmounted, the chroot's root directory will be removed if requested.
-func (c *Chroot) unmountAndRemove(leaveOnDisk, isFinalAttempt bool) (err error) {
+// If doLazyUnmount is true, use the lazy unmount flag which will allow the unmount to succeed even if the mount point is busy.
+func (c *Chroot) unmountAndRemove(leaveOnDisk, doLazyUnmount bool) (err error) {
 	const (
 		retryDuration      = time.Second
 		totalAttempts      = 3
 		unmountFlagsNormal = 0
-		// Do a lazy unmount on the final attempt. This will allow the unmount to succeed even if the mount point is busy.
+		// Do a lazy unmount as a fallback. This will allow the unmount to succeed even if the mount point is busy.
 		// This is to avoid leaving folders like /dev mounted if the chroot folder is forcefully deleted by the user. Even
 		// if the mount is busy at least it will be detached from the filesystem and will not damage the host.
-		unmountFlagsFinal = unix.MNT_DETACH
-		errMsg            = "Failed to unmount (%s). Error: %s"
+		unmountFlagsLazy = unix.MNT_DETACH
+		errMsg           = "Failed to unmount (%s). Error: %s"
 	)
 	unmountFlags := unmountFlagsNormal
-	if isFinalAttempt {
-		logger.Log.Warnf("Final attempt to unmount chroot (%s)", c.rootDir)
-		unmountFlags = unmountFlagsFinal
+	if doLazyUnmount {
+		logger.Log.Warnf("Final attempt to unmount chroot (%s), using a lazy unmount", c.rootDir)
+		unmountFlags = unmountFlagsLazy
 	}
 
 	for _, mountPoint := range c.mountPoints {
--- a/toolkit/tools/internal/safechroot/safechroot.go
+++ b/toolkit/tools/internal/safechroot/safechroot.go
@@ -19,6 +19,7 @@ import (
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/systemdependency"
 
+	"github.com/moby/sys/mountinfo"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
@@ -376,6 +377,10 @@ func (c *Chroot) Close(leaveOnDisk bool) (err error) {
 	if buildpipeline.IsRegularBuild() {
 		// mount is only supported in regular pipeline
 		err = c.unmountAndRemove(leaveOnDisk, isNotFinalAttempt)
+		if err != nil {
+			logger.Log.Warnf("Chroot cleanup failed, will retry with lazy unmount. Error: %s", err)
+			err = c.unmountAndRemove(leaveOnDisk, isFinalAttempt)
+		}
 		if err == nil {
 			const emptyLen = 0
 			// Remove this chroot from the list of active ones since it has now been cleaned up.
@@ -496,6 +501,17 @@ func (c *Chroot) unmountAndRemove(leaveOnDisk, isFinalAttempt bool) (err error) 
 	for _, mountPoint := range c.mountPoints {
 		fullPath := filepath.Join(c.rootDir, mountPoint.target)
 
+		isMounted := true
+		isMounted, err = mountinfo.Mounted(fullPath)
+		if err != nil {
+			err = fmt.Errorf("failed to check if mount point (%s) is mounted. Error: %s", fullPath, err)
+			return err
+		}
+		if !isMounted {
+			logger.Log.Debugf("Skipping unmount of (%s) because it is not mounted", fullPath)
+			continue
+		}
+
 		logger.Log.Debugf("Unmounting (%s)", fullPath)
 
 		// Skip mount points if they were not successfully created
@@ -507,7 +523,7 @@ func (c *Chroot) unmountAndRemove(leaveOnDisk, isFinalAttempt bool) (err error) 
 			logger.Log.Debugf("Calling unmount (%s, %v)", fullPath, unmountFlags)
 			umountErr := unix.Unmount(fullPath, unmountFlags)
 			if isFinalAttempt && umountErr != nil {
-				logger.Log.Warnf(errMsg, fullPath, err.Error())
+				logger.Log.Warnf(errMsg, fullPath, umountErr.Error())
 			}
 			return umountErr
 		}, totalAttempts, retryDuration, 2.0, nil)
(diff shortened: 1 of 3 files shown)


toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -81,6 +82,11 @@ var defaultChrootEnv = []string{
 	"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
 }
 
+const (
+	isNotFinalAttempt = false
+	isFinalAttempt    = true
```
cwize1, line 87: IMHO, these constants make the code more confusing. `isFinalAttempt` sounds like a variable not a named constant. I would just use `true` and `false` in the code.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -441,50 +451,85 @@ func cleanupAllChroots() {
...
 // Iff all mounts were successfully unmounted, the chroot's root directory will be removed if requested.
-func (c *Chroot) unmountAndRemove(leaveOnDisk bool) (err error) {
+func (c *Chroot) unmountAndRemove(leaveOnDisk, isFinalAttempt bool) (err error) {
 	const (
-		totalAttempts = 3
-		retryDuration = time.Second
-		unmountFlags  = 0
+		retryDuration       = time.Second
+		totalAttemptsNormal = 3
+		totalAttemptsFinal  = 5
```
cwize1, line 485: Why do you need retries on a lazy detach?

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -441,50 +451,85 @@ func cleanupAllChroots() {
...
 			if err != nil {
 				logger.Log.Errorf("Failed to unmount chroot (%s)", activeChroots[i].rootDir)
+				failedToUnmount = true
 			}
 		}
 	}
 
-	logger.Log.Info("Cleanup finished")
+	if failedToUnmount {
+		logger.Log.Fatalf("Failed to unmount a chroot, manual unmount required")
```
cwize1, line 471: Why do we want to hard crash here?

neha170: Is the "manual" unmount something the user has to do? If yes, echoing the dangling mount point and command to unmount it might be helpful for the user.

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: Paranoia mostly, but I've not seen it fail on the 1st attempt... given that maybe I'll just simplify it to use the same retry logic as the normal attempt with the understanding that it is expected to work the 1st time.

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: I'm torn on this. I settled on "stop immediately, something is VERY wrong" here since this should now "never" fail. Any subsequent actions risk damaging the host system here (although in practice the tools will just yell and refuse to continue if we try to re-use the chroot).

Generally the user will want to call `make clean-<sub component>` to clean the chroot, but the tools can't really know this. I guess I could recommend the user call `make clean` but it feels a bit weird to have a library knowing about the makefile several layers below it.

toolkit/tools/internal/safechroot/safechroot.go
PawelWMS: I disagree. We've used to do only `true` and `false` in the code and we ended up not really knowing, what these values were setting without having to jump to the function's definition. Now we try to use well-named constants as convention across the code.

toolkit/tools/internal/safechroot/safechroot.go
PawelWMS: I'd say for errors with an external source, I'd still return `err`, because we don't have control over them and really predict how they may change in the future. For cases where the error can only be caused by a bug in our tool's logic, then I think we can crash (I'm referring to the internal logic of the dependency graph we've discussed a while ago).

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -194,8 +195,8 @@ func (c *Chroot) Initialize(tarPath string, extraDirectories []string, extraMoun
 		if err != nil {
 			if buildpipeline.IsRegularBuild() {
 				// mount/unmount is only supported in regular pipeline
-				// Best effort cleanup in case mountpoint creation failed mid-way through
-				cleanupErr := c.unmountAndRemove(leaveChrootOnDisk)
+				// Best effort cleanup in case mountpoint creation failed mid-way through. We will not try again so treat as final attempt.
+				cleanupErr := c.unmountAndRemove(leaveChrootOnDisk, true)
```
PawelWMS, line 199: Yeah, now this feels more cryptic with just a `true`. Also, I'd rather stick to how we try to do things in the rest of the code.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -441,50 +446,82 @@ func cleanupAllChroots() {
...
+		logger.Log.Info("Cleanup finished")
+	}
 }
 
 // unmountAndRemove retries to unmount directories that were mounted into
 // the chroot until the unmounts succeed or too many failed attempts.
 // This is to avoid leaving folders like /dev mounted when the chroot folder is forcefully deleted in cleanup.
 // Iff all mounts were successfully unmounted, the chroot's root directory will be removed if requested.
-func (c *Chroot) unmountAndRemove(leaveOnDisk bool) (err error) {
+func (c *Chroot) unmountAndRemove(leaveOnDisk, isFinalAttempt bool) (err error) {
```
PawelWMS, line 476: Nit-pick: on the level of this function I'm not sure if the final/non-final distinction makes sense. Seeing what the function's job is, it seems like lazy/non-lazy might be more suitable. Whether someone decided to do a lazy unmount as part of a final unmount attempt or not is outside what this function needs to worry about. What do you think?

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -441,50 +446,82 @@ func cleanupAllChroots() {
...
 		}
 
-		err = retry.Run(func() error {
-			return unix.Unmount(fullPath, unmountFlags)
-		}, totalAttempts, retryDuration)
+		_, err = retry.RunWithExpBackoff(func() error {
+			logger.Log.Debugf("Calling unmount (%s, %v)", fullPath, unmountFlags)
+			umountErr := unix.Unmount(fullPath, unmountFlags)
+			if isFinalAttempt && umountErr != nil {
+				logger.Log.Warnf(errMsg, fullPath, umountErr.Error())
```
PawelWMS, line 518: This warning may be a bit confusing, if we fail the first attempt and succeed with another. I'd suggest either dropping it (we're printing a warning below anyway) or somehow clarifying, that we're not done trying, yet.

toolkit/tools/internal/safechroot/safechroot.go
cwize1: Perhaps change it to `true /*isFinalAttempt*/`?

toolkit/tools/internal/safechroot/safechroot.go
cwize1: Does it matter if the lazy detach succeeds? Isn't it just a mitigation against a bad situation?

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -370,7 +371,11 @@ func (c *Chroot) Close(leaveOnDisk bool) (err error) {
 
 	if buildpipeline.IsRegularBuild() {
 		// mount is only supported in regular pipeline
-		err = c.unmountAndRemove(leaveOnDisk)
+		err = c.unmountAndRemove(leaveOnDisk, false)
+		if err != nil {
+			logger.Log.Warnf("Chroot cleanup failed, will retry with lazy unmount. Error: %s", err)
+			err = c.unmountAndRemove(leaveOnDisk, true)
```
cwize1, line 377: I don't think this function should succeed if the normal unmount fails but the lazy detach succeeds. The lazy detach doesn't guarantee that the mount has been closed. It just tries to ensure that it will close eventually. I think it is fine to use the lazy detach as a mitigation to try to put the user's machine in the least broken state possible.

dmcilvaney: Unfortunately there are cases where the normal path seems to fail 100% of the time (https://github.com/microsoft/CBL-Mariner/pull/6228 seems to leave `/dev/hugepages` active for a long time after the tool is "done"). The lazy unmount does the most important bit, it disconnects the mount from the filesystem. I'm much less concerned about tearing down the actual backing structures. My primary concern is making sure that running `sudo rm -r ./build` doesn't damage the host.

I don't have the time right now to debug the exact failure condition that leaves the chroot stuck, and I have no confidence that would also fix the various failures that leave package build chroots behind as well.

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: Sure, I can do that.

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: I'll just drop it since we capture the last error anyways.

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: As long as the chroot leaves the system in a state where the tools can be run again, I think it counts as "success". I'd rather we be able to explicitly unmount, but practice has shown that just doesn't always happen. Our `make clean` has used lazy unmounts for a long time via `safeunmount.sh`, and I'm not aware of any negative affects from using the lazy flag.

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: I'll be ok with switching this over to an err that bubbles out if we see it failing in practice, but I don't expect it to happen.

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: I think I'm going to go with:
``` golang
const (
	doLazyUnmount   = true
	doNormalUnmount = !doLazyUnmount
)
```

toolkit/tools/internal/safechroot/safechroot.go
cwize1: Can we guarantee the integrity of the image file if unmount fails or is asynchronous? Should we make /dev/hugepages a special case?

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -473,21 +478,22 @@ func cleanupAllChroots() {
 // the chroot until the unmounts succeed or too many failed attempts.
 // This is to avoid leaving folders like /dev mounted when the chroot folder is forcefully deleted in cleanup.
 // Iff all mounts were successfully unmounted, the chroot's root directory will be removed if requested.
-func (c *Chroot) unmountAndRemove(leaveOnDisk, isFinalAttempt bool) (err error) {
+// If doLazyUnmount is true, use the lazy unmount flag which will allow the unmount to succeed even if the mount point is busy.
+func (c *Chroot) unmountAndRemove(leaveOnDisk, doLazyUnmount bool) (err error) {
```
cwize1, line 482: `doLazyUnmount` is both a function parameter and a constant. Please avoid name hiding.

cwize1 (changes requested):

toolkit/tools/internal/safechroot/safechroot.go
PawelWMS: Nice catch!

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: The teardown of the mounted image files used in the imager tool is done through loopbacks, which I believe will fail if the mount is still busy. We also 

In general, `/dev/hugepages` is  just the one case I've looked into but I don't think it captures all cases where the unmounts fail (the package build chroots often flake out when a build is cancelled,  but I don't think it has anything to do with hugepages). My primary goal here is to protect the user's system. Ideally we could debug every chroot failure but given this code path is very uncommon (the precacher tool is the only spot this happens during normal flow, which I agree is a bit odd) I think its sufficient to just add an additional fallback.

(Important to note, in the vast majority of cases a failed unmount generates NO error or other signal that it failed since it is called via defer. The tool will continue as normal with the mount still in-place. So, any spot this would cause issues with correctness would already be failing in the same way.

cwize1: Yeah...the fact that the `Close()` return error isn't being handled is another problem that should be fixed.

e.g.:

```golang
err = installChroot.Initialize(emptyWorkerTar, extraDirectories, extraInstallMountPoints)
if err != nil {
	err = fmt.Errorf("failed to create install chroot: %s", err)
	return
}
defer func() {
	closeErr := installChroot.Close(leaveChrootOnDisk)
	if err == nil {
		err = closeErr
	}
}()
```

cwize1: Yeah...the fact that the `Close()` return error isn't being handled is another problem that should be fixed.

e.g.:

```golang
err = installChroot.Initialize(emptyWorkerTar, extraDirectories, extraInstallMountPoints)
if err != nil {
	err = fmt.Errorf("failed to create install chroot: %s", err)
	return
}
defer func() {
	closeErr := installChroot.Close(leaveChrootOnDisk)
	if err == nil {
		err = closeErr
	}
}()
```

toolkit/tools/internal/safechroot/safechroot.go
cwize1: Is the reason you want to hard crash here related to the fact that the `Close()` error isn't being handled by the caller?

dmcilvaney: Yes, that's a bit part of it. We have an emergency handler to clean up the chroots, but it only runs if the caller uses the logger's exit routines (which... we don't really do...)

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -81,6 +82,11 @@ var defaultChrootEnv = []string{
 	"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
 }
 
+const (
+	doLazyUnmount   = true
```
cwize1, line 86: Perhaps name these `unmountTypeNormal` and `unmountTypeLazy`?

cwize1 (approved):

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: Sure, that seems reasonable.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -441,50 +451,80 @@ func cleanupAllChroots() {
...
+		unmountFlagsNormal = 0
+		// Do a lazy unmount as a fallback. This will allow the unmount to succeed even if the mount point is busy.
+		// This is to avoid leaving folders like /dev mounted if the chroot folder is forcefully deleted by the user. Even
+		// if the mount is busy at least it will be detached from the filesystem and will not damage the host.
+		unmountFlagsLazy = unix.MNT_DETACH
+		errMsg           = "Failed to unmount (%s). Error: %s"
 	)
+	unmountFlags := unmountFlagsNormal
+	if lazyUnmount {
+		logger.Log.Warnf("Final attempt to unmount chroot (%s), using a lazy unmount", c.rootDir)
```
PawelWMS, line 495: I wouldn't print a warning on this level for the same reason why I mentioned the variable re-name - it's unnecessary for this function to know if this is a final attempt of anything or someone just running a lazy unmount as part of some regular flow.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -441,50 +451,80 @@ func cleanupAllChroots() {
...
+		unmountFlags = unmountFlagsLazy
+	}
 
 	for _, mountPoint := range c.mountPoints {
 		fullPath := filepath.Join(c.rootDir, mountPoint.target)
 
+		isMounted := true
+		isMounted, err = mountinfo.Mounted(fullPath)
```
PawelWMS, line 503: Nit-pick: separate declaration and init of `isMounted` unnecessary. And since we're here, I'd unify how we return from this function (either `return` or `return err`). If you take the below suggestion, then we'd need `return err`.
```suggestion
		isMounted, err := mountinfo.Mounted(fullPath)
```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -441,50 +451,80 @@ func cleanupAllChroots() {
...
 		// Skip mount points if they were not successfully created
 		if !mountPoint.isMounted {
 			continue
 		}
 
-		err = retry.Run(func() error {
-			return unix.Unmount(fullPath, unmountFlags)
-		}, totalAttempts, retryDuration)
+		_, err = retry.RunWithExpBackoff(func() error {
+			logger.Log.Debugf("Calling unmount (%s, %v)", fullPath, unmountFlags)
```
PawelWMS, line 521: Could you mention in the logs what `%s` and `%v` is? The path might be clear, but the flags might be harder to guess.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -441,50 +451,80 @@ func cleanupAllChroots() {
...
-		retryDuration = time.Second
-		unmountFlags  = 0
+		retryDuration      = time.Second
+		totalAttempts      = 3
+		unmountFlagsNormal = 0
+		// Do a lazy unmount as a fallback. This will allow the unmount to succeed even if the mount point is busy.
+		// This is to avoid leaving folders like /dev mounted if the chroot folder is forcefully deleted by the user. Even
+		// if the mount is busy at least it will be detached from the filesystem and will not damage the host.
+		unmountFlagsLazy = unix.MNT_DETACH
+		errMsg           = "Failed to unmount (%s). Error: %s"
```
PawelWMS, line 491: Doesn't need to be separately declared anymore - it's only used once.

toolkit/tools/internal/safechroot/safechroot.go
dmcilvaney: This will mask `err` and generates a compiler error unfortunately, so either `isMounted := true` or add a new `mountErr`.

dmcilvaney: I'll move to `var isMounted bool` so its more clear this is just a declaration.

dmcilvaney: Ah, I get what you mean about the `return err` now. I'll just leave it with `var isMounted bool` and clean up the `return err`.

PawelWMS (approved):

PR #6243, opened 2023-09-20
Get go tools unit tests to pass in VSCode

reubeno:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary
The go tools' unit tests presently don't run cleanly in VSCode. The main blockers are tests that require running as `root` / under `sudo`. While triaging some of those failures, I also noticed a unit test that is writing to `/etc` on the host system--that should be remedied.

###### Change Log
- Fix the network config unit test so that it arranges to write to a test dir, not to real `/etc`. Actually makes the test simpler since we can leverage `t.TempDir()` and rely on it being existent (and empty) and cleaned up for us afterwards.
- Identify unit tests that require running in a chroot and arrange for them to be skipped when the effective UID isn't 0 (i.e., when we're running as non-root, and not under `sudo`).

No.

###### Test Methodology
Local test run.

### DIFF
--- a/toolkit/tools/imagegen/configuration/networkconfig_test.go
+++ b/toolkit/tools/imagegen/configuration/networkconfig_test.go
@@ -5,6 +5,7 @@ package configuration
 
 import (
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/file"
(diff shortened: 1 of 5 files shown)


reubeno:
VScode test result:

image

toolkit/tools/imagegen/configuration/networkconfig_test.go
```diff
@@ -134,28 +135,12 @@ func TestShouldFailParsingInvalidDevice_Network(t *testing.T) {
...
+	testNetworkFileDir := t.TempDir()
+	testNetworkFile := filepath.Join(testNetworkFileDir, "10-static-eth1.network")
 
 	testNetwork := validNetworks[0]
 
-	err = createNetworkConfigFile(nil, testNetwork, "eth1")
+	err := createNetworkConfigFile(nil, testNetwork, "eth1", testNetworkFileDir)
 	assert.NoError(t, err)
 	t.Cleanup(func() {
 		os.Remove(testNetworkFile)
```
neha170, line 146: Do we also need to remove the tempDir (testNetworkFileDir) we created?

reubeno: No, not as far as I can tell but I will double-check. The official [docs](https://pkg.go.dev/testing#T.TempDir) say this:

>TempDir returns a temporary directory for the test to use. The directory is automatically removed by Cleanup when the test and all its subtests complete. Each subsequent call to t.TempDir returns a unique directory; if the directory creation fails, TempDir terminates the test by calling Fatal.

dmcilvaney: I've used it before, it should clear up after the tests are done. 

Fun little tidbit that caught me off guard, I missed the bit "Each subsequent call to t.TempDir returns a unique directory" and from the top part of the description assumed it was unique to the test, but it can create multiple unique directories even inside a single test.

toolkit/tools/internal/safechroot/safechroot_test.go
```diff
@@ -42,7 +42,13 @@ func TestMain(m *testing.M) {
 	testDir = filepath.Join(workingDir, "testdata")
 	tmpDir = filepath.Join(workingDir, "_tmp")
 
-	retVal := m.Run()
+	var retVal int
+	if os.Geteuid() != 0 {
```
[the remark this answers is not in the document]
reubeno: The `m` in `TestMain` is of a different type (`*testing.M`) from the `t` in the individual test cases (`*testing.T`). The latter has a `Skip` method but the former does not. The appropriate way to skip execution in `TestMain` is not to call `m.Run()`.

As for why the code isn't doing it at the very beginning... in order to log that we're skipping execution, we need to at least bring up logging. Looking back through this, I think we can clean it up a bit more. I've pushed some additional changes to this end.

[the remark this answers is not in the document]
neha170: Thanks for the explanation

neha170 (approved):

PR #6248, opened 2023-09-20
Unify behavior of `USE_PREVIEW_REPO` on url and repo lists

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

There is currently an inconsistency in how `USE_PREVIEW_REPO=y` treats `PACKAGE_URL_LIST`, `SRPM_URL_LIST`, and `REPO_LIST`. Currently the preview repos are appended to `REPO_LIST` regardless of how the initial value was set.

> In make, variables may come from several places like env, cmdline, or inside the makefile. `?=` will set a variable if it is undefined. `+=` will append to a variable. However, `+=` will ONLY append to variables that are set inside the makefile (generalizing slightly here). This means that if a user passed `make my-target MY_VAR=value`, and inside the makefile we see `MY_VAR ?= default` and `MY_VAR += new_value`, the final result will still be `"value"`. If the user calls `make my-target` however we will get `"default new_value"`. This behaviour can be changed by using the `override` keyword.

`REPO_LIST` is appended to by using `override`, but the other values are not. Unify the handling here so that we use `override` on all input lists when we turn on preview repos.

###### Change Log  

- `USE_PREVIEW_REPO=y` appends to user set `PACKAGE_URL_LIST` now
- `USE_PREVIEW_REPO=y` appends to user set `SRPM_URL_LIST` now

**NO**
(The toolchain preferentially uses the 1st entry in `PACKAGE_URL_LIST` so appending an extra value should not affect the toolchain builds)

###### Associated issues  

- https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/45998385/

###### Test Methodology

- Local testing

### DIFF
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -116,13 +116,14 @@ SOURCE_URL         ?= https://cblmarinerstorage.blob.core.windows.net/sources/co
 # Note on order of precedence: When a variable is passed from the commandline (i.e., make PACKAGE_URL_LIST="my list"), append
 # assignments do not take affect without using 'override'. This means that all of the following PACKAGE_URL_LIST values will
 # be ignored if the user sets any value.
+##help:var:PACKAGE_URL_LIST:<urls_list>=Space-separated list of URLs to download toolchain RPM packages from, used to populate the toolchain packages if `REBUILD_TOOLCHAIN=y'. The URLs are appended to the default set of URLs. Print default list with 'make -s printvar-PACKAGE_URL_LIST'.
 PACKAGE_URL_LIST   ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/$(build_arch)
 PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/debuginfo/$(build_arch)
 PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/Microsoft/$(build_arch)
 REPO_LIST          ?=
 SRPM_URL_LIST      ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/srpms
 
-# Preview urls/repos will be appended even if the user sets custom values for the lists (via override).
+# Preview URLs/repos will be appended (via override) even if the user sets custom values for the lists.
 ifeq ($(USE_PREVIEW_REPO),y)
    override PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/preview/base/$(build_arch)
    override PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/preview/base/debuginfo/$(build_arch)
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -113,17 +113,21 @@ IMAGES_DIR      ?= $(OUT_DIR)/images
 # External source server
 SOURCE_URL         ?= https://cblmarinerstorage.blob.core.windows.net/sources/core
 
+# Note on order of precedence: When a variable is passed from the commandline (i.e., make PACKAGE_URL_LIST="my list"), append
+# assignments do not take affect without using 'override'. This means that all of the following PACKAGE_URL_LIST values will
+# be ignored if the user sets any value.
 PACKAGE_URL_LIST   ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/$(build_arch)
 PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/debuginfo/$(build_arch)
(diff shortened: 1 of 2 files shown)


neha170 (approved):

toolkit/Makefile
```diff
@@ -113,17 +113,21 @@ IMAGES_DIR      ?= $(OUT_DIR)/images
...
+# Note on order of precedence: When a variable is passed from the commandline (i.e., make PACKAGE_URL_LIST="my list"), append
+# assignments do not take affect without using 'override'. This means that all of the following PACKAGE_URL_LIST values will
+# be ignored if the user sets any value.
 PACKAGE_URL_LIST   ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/$(build_arch)
 PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/debuginfo/$(build_arch)
 PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/Microsoft/$(build_arch)
 REPO_LIST          ?=
 SRPM_URL_LIST      ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/srpms
 
+# Preview urls/repos will be appended even if the user sets custom values for the lists (via override).
```
PawelWMS, line 125: I moved the `(via override)` part, because otherwise it sounded as if the user had an option to use `override` somehow. Plus a small nit-pick regarding using capital letter for acronyms.;)
```suggestion
# Preview URLs/repos will be appended (via override) even if the user sets custom values for the lists.
```

toolkit/Makefile
```diff
@@ -113,17 +113,21 @@ IMAGES_DIR      ?= $(OUT_DIR)/images
...
 SOURCE_URL         ?= https://cblmarinerstorage.blob.core.windows.net/sources/core
 
+# Note on order of precedence: When a variable is passed from the commandline (i.e., make PACKAGE_URL_LIST="my list"), append
+# assignments do not take affect without using 'override'. This means that all of the following PACKAGE_URL_LIST values will
+# be ignored if the user sets any value.
 PACKAGE_URL_LIST   ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/$(build_arch)
 PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/debuginfo/$(build_arch)
 PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/Microsoft/$(build_arch)
 REPO_LIST          ?=
 SRPM_URL_LIST      ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/srpms
```
PawelWMS, line 123: Could you also update `building.md` to explicitly explain this appending behaviour/the default set of URLs?

toolkit/Makefile
```diff
@@ -113,17 +113,21 @@ IMAGES_DIR      ?= $(OUT_DIR)/images
 # External source server
 SOURCE_URL         ?= https://cblmarinerstorage.blob.core.windows.net/sources/core
 
+# Note on order of precedence: When a variable is passed from the commandline (i.e., make PACKAGE_URL_LIST="my list"), append
+# assignments do not take affect without using 'override'. This means that all of the following PACKAGE_URL_LIST values will
+# be ignored if the user sets any value.
```
PawelWMS, line 118: The first sentence makes sense, but the second one is a big confusing to me. It sounds like we **WILL** ignore the default values, if the user sets anything themselves, but I think that's the opposite of what's happening.

PawelWMS, line 118: Since we're here:

```suggestion
# Note on order of precedence: When a variable is passed from the commandline (i.e., make PACKAGE_URL_LIST="my list"), append
# assignments do not take affect without using 'override'. This means that all of the following PACKAGE_URL_LIST values will
# be ignored if the user sets any value.
##help:var:PACKAGE_URL_LIST:<urls_list>=Space-separated list of URLs to download toolchain RPM packages from, used to populate the toolchain packages if `REBUILD_TOOLCHAIN=y'.
```

anphel31 (approved):

toolkit/Makefile
dmcilvaney: No, that is correct. We would need to use `override PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/prod/base/debuginfo/$(build_arch)` to append. Instead we completely ignore all of these appends if the user does `PACKAGE_URL_LIST="my list"`. The goal here is to make `USE_PREVIEW_REPO=y` have the same affect on the URL lists as it does on the repo files (i.e., uncondinally append).

PawelWMS: Ah, got it, I missed the fact that the default set of URLs doesn't use `override`. Thanks for the clarification!

In that case, my suggestion for the `help:var` comment was incorrect. And judging by your docs changes, the `REBUILD_TOOLCHAIN=y` part of that comment should have been `REBUILD_TOOLCHAIN=n` as well. Could you update that comment as well?

PawelWMS (approved):
One more comment to tweak but already approving.

PR #6264, opened 2023-09-21
Allow skipping broken packages in fetcher

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

There are occasionally situations where the upstream repo contains packages that are not installable (perhaps a dependency was broken/missing etc.). The fetcher currently follows several steps to pick a package:

1) Find all options that provide a dependency
1) Download each of them (and their dependencies)
1) Determine which is the best choice to use
1) Update the remote node with that cached RPM

If one of the candidate packages has a broken dependnecy, it will show  up in the list from 1), however the clone operation 2) will fail. 

Allow for some of the candidates to fail to download, but ensure we have at least one successful clone before proceeding to 3).

###### Change Log  

- Allow a fetch candidate to fail so long as we have at least one success.

**NO**

###### Test Methodology

- Local

### DIFF
--- a/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
+++ b/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
@@ -511,6 +511,7 @@ func resolveSingleNode(cloner *rpmrepocloner.RpmRepoCloner, node *pkggraph.PkgNo
 		return
 	} else {
 		// Reset the error since we found at least one source and can continue.
+		logger.Log.Warn("Some candidate downloads failed, but at least one package was successfully cloned, continuing.")
 		err = nil
 	}
 
--- a/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
+++ b/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
@@ -474,25 +474,47 @@ func resolveSingleNode(cloner *rpmrepocloner.RpmRepoCloner, node *pkggraph.PkgNo
 	}
 
 	preBuilt := false
+	// We will try to clone each package. We need at least one to succeed.
+	foundAtLeastOneSource := false
+	cloneErr := error(nil)
+	successfullyFetchedPackages := make([]string, 0, len(resolvedPackages))
 	for _, resolvedPackage := range resolvedPackages {
 		if !fetchedPackages[resolvedPackage] {
 			desiredPackage := &pkgjson.PackageVer{
 				Name: resolvedPackage,
 			}
 
-			preBuilt, err = cloner.Clone(cloneDeps, desiredPackage)
-			if err != nil {
-				err = fmt.Errorf("failed to clone '%s' from RPM repo:\n%w", resolvedPackage, err)
-				return
+			preBuilt, cloneErr = cloner.Clone(cloneDeps, desiredPackage)
+			if cloneErr != nil {
+				// There may be cases were certain packages are broken and can't be cloned. In this case we should
+				// continue to try to clone other packages. (e.g. a package is in the repo but a dependency is missing)
+				logger.Log.Warnf("failed to clone a possible candidate package '%s' from RPM repo: %s", resolvedPackage, cloneErr.Error())
+				if err == nil {
+					// Capture the first error we see, but continue to try to clone other packages.
+					err = cloneErr
+				}
+				continue
+			} else {
+				foundAtLeastOneSource = true
 			}
 			fetchedPackages[resolvedPackage] = true
 			prebuiltPackages[resolvedPackage] = preBuilt
+			successfullyFetchedPackages = append(successfullyFetchedPackages, resolvedPackage)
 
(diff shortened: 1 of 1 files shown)


toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -474,25 +474,47 @@ func resolveSingleNode(cloner *rpmrepocloner.RpmRepoCloner, node *pkggraph.PkgNo
...
 		}
 	}
 
-	err = assignRPMPath(node, outDir, resolvedPackages)
+	// If we failed to clone any packages, return the first error we saw.
+	if err != nil && !foundAtLeastOneSource {
+		err = fmt.Errorf("failed to clone any packages providing '%v':\n%w", node.VersionedPkg, err)
+		return
+	} else {
+		// Reset the error since we found at least one source and can continue.
```
anphel31, line 513: can we add a message indicating that at least one package could be cloned (at the end)?

anphel31 (approved):

PawelWMS (changes requested):
Thanks for the change, Daniel! I'm freezing this change for now, until we figure out if we even want the toolkit to behave this way (referring to an offline e-mail thread).

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
dmcilvaney:
```suggestion
		// Reset the error since we found at least one source and can continue.
		logger.Log.Warn("Some candidate downloads failed, but at least one package was successfully cloned, continuing.")
```

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -474,25 +474,48 @@ func resolveSingleNode(cloner *rpmrepocloner.RpmRepoCloner, node *pkggraph.PkgNo
...
+			if cloneErr != nil {
+				// There may be cases were certain packages are broken and can't be cloned. In this case we should
+				// continue to try to clone other packages. (e.g. a package is in the repo but a dependency is missing)
+				logger.Log.Warnf("failed to clone a possible candidate package '%s' from RPM repo: %s", resolvedPackage, cloneErr.Error())
+				if err == nil {
+					// Capture the first error we see, but continue to try to clone other packages.
+					err = cloneErr
+				}
+				continue
+			} else {
```
cwize1, line 497: Because of the `continue` in the block above, the `else` isn't needed.

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -474,25 +474,48 @@ func resolveSingleNode(cloner *rpmrepocloner.RpmRepoCloner, node *pkggraph.PkgNo
 	}
 
 	preBuilt := false
+	// We will try to clone each package. We need at least one to succeed.
+	foundAtLeastOneSource := false
+	cloneErr := error(nil)
```
cwize1, line 479: Why are you defining `cloneErr` here? It doesn't seem to be used outside of the loop.

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -474,25 +474,48 @@ func resolveSingleNode(cloner *rpmrepocloner.RpmRepoCloner, node *pkggraph.PkgNo
...
 			prebuiltPackages[resolvedPackage] = preBuilt
+			successfullyFetchedPackages = append(successfullyFetchedPackages, resolvedPackage)
 
 			logger.Log.Debugf("Fetched '%s' as potential candidate (is pre-built: %v).", resolvedPackage, prebuiltPackages[resolvedPackage])
 		}
 	}
 
-	err = assignRPMPath(node, outDir, resolvedPackages)
+	// If we failed to clone any packages, return the first error we saw.
+	if err != nil && !foundAtLeastOneSource {
```
cwize1, line 509: Would it make sense to replace `!foundAtLeastOneSource` with `len(successfullyFetchedPackages) == 0`?

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -474,25 +474,48 @@ func resolveSingleNode(cloner *rpmrepocloner.RpmRepoCloner, node *pkggraph.PkgNo
...
+		err = fmt.Errorf("failed to clone any packages providing '%v':\n%w", node.VersionedPkg, err)
+		return
+	} else {
+		// Reset the error since we found at least one source and can continue.
+		logger.Log.Warn("Some candidate downloads failed, but at least one package was successfully cloned, continuing.")
```
cwize1, line 514: Should this warning only be logged when `err != nil` and `foundAtLeastOneSource` is `true`?

cwize1 (changes requested):

dmcilvaney:
Fix was made to upstream packages, abandoning this fix since we want to detect these failures rather than skipping them.

PR #6287, opened 2023-09-26
Allow scheduler to update cached implicit nodes

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

There is an edge case that may break builds using implicit provides. The symptoms are: `pkgA` queues for build, the scheduler calculates its build-time dependencies: `pkgB-1, pkgB-2, pkgC, ...`. When the packages are installed into the build environment, `pkgB-1` conflicts with `pkgB-2` and the build fails.

The scheduler calculates the build-time dependencies by doing a walk of the build graph starting from the node to be built (`pkgA_build`). In this example, lets say we have a newly updated `pkgB` with `version:2`. Our package also relies on `pkgC`. Both `pkgB` and `pkgC` have a dependency on the implicit package `/usr/bin/tool_b` (which by observation we know will come from `pkgB`).
```mermaid
flowchart TB
    classDef a fill:#F7DC6F,color:#000000
    classDef b fill:#85C1E9,color:#000000
    classDef b_imp fill:#A569BD,color:#000000
    classDef c fill:#F5B7B1,color:#000000

    pkgA_run:::a --> pkgA_build:::a --> pkgB-2_run:::b --> pkgB-2_build:::b
    pkgB-2_run:::b --> /usr/bin/tool_b:::b_imp
    pkgA_build:::a --> pkgC_run:::c --> /usr/bin/tool_b:::b_imp
    
```
During the fetch stage, we defensively query our repo servers to see where we might expect to find `/usr/bin/tool_b` incase we don't have a package that will provide it locally. In this case we find out that it comes from `pkgB`, but the version in the repo is `version:1`, so the fetcher grabs version 1.
```mermaid
flowchart TB
    classDef a fill:#F7DC6F,color:#000000
    classDef b fill:#85C1E9,color:#000000
    classDef b_imp fill:#A569BD,color:#000000
    classDef c fill:#F5B7B1,color:#000000
    
    pkgA_run:::a --> pkgA_build:::a --> pkgB-2_run:::b --> pkgB-2_build:::b
    pkgB-2_run:::b --> /usr/bin/tool_b:::b_imp
    pkgA_build:::a --> pkgC_run:::c --> /usr/bin/tool_b(/usr/bin/tool_b - *version 1*):::b_imp
```
The scheduler will attempt to avoid using "implicit" nodes that have no local backing as long as possible (e.g., `/usr/bin/tool_b`), but eventually the build can be stuck. Lets assume the build has reached that point, and enables implicit cached nodes (e.g., `/usr/bin/tool_b`). We now unblock the graph and eventually build `pkgB-2_build`. Currently we don't update implicit cached nodes once we enter the global implicit cache allow state (more on why later).

Packages `b`, and `c` will build fine, resulting in something that looks like:
```mermaid
flowchart TB
    classDef a fill:#F7DC6F,color:#000000
    classDef b fill:#85C1E9,color:#000000,stroke-dasharray: 5 5,stroke:#00AF00,stroke-width:4px
    classDef b_imp fill:#A569BD,color:#000000,stroke-dasharray: 5 5,stroke:#00AF00,stroke-width:4px
     classDef c fill:#F5B7B1,color:#000000,stroke-dasharray: 5 5,stroke:#00AF00,stroke-width:4px
    
    pkgA_run:::a --> pkgA_build:::a --> pkgB-2_run:::b --> pkgB-2_build:::b
    pkgB-2_run:::b --> /usr/bin/tool_b:::b_imp
    pkgB-2_run:::b ~~~ /usr/bin/tool_b2(/usr/bin/tool_b - *version 2*):::b -.-> pkgB-2_build
    pkgA_build:::a --> pkgC_run:::c --> /usr/bin/tool_b(/usr/bin/tool_b - *version 1*):::b_imp
```

Notice we found a new copy of `/usr/bin/tool_b`, but didn't update any nodes to  use it. The scheduler will now build `pkgA_build`, using a walk of the graph to determine the packages to install into the worker (Yellow).
```mermaid
flowchart TB
    classDef a fill:#F7DC6F,color:#000000
    classDef b fill:#85C1E9,color:#000000,stroke-dasharray: 5 5,stroke:#00AF00,stroke-width:4px
    classDef b_imp fill:#A569BD,color:#000000,stroke-dasharray: 5 5,stroke:#00AF00,stroke-width:4px
     classDef c fill:#F5B7B1,color:#000000,stroke-dasharray: 5 5,stroke:#00AF00,stroke-width:4px
    classDef install fill:#Bda233,color:#000000
    
    pkgA_run:::a --> pkgA_build:::a --> pkgB-2_run:::install --> pkgB-2_build:::install
    pkgB-2_run:::install --> /usr/bin/tool_b:::install
    pkgB-2_run:::install ~~~ /usr/bin/tool_b2(/usr/bin/tool_b - *version 2*):::b -.-> pkgB-2_build
    pkgA_build:::a --> pkgC_run:::install --> /usr/bin/tool_b(/usr/bin/tool_b - *version 1*):::b_imp
```

Notice we install both version 1 and version 2 of `pkgB`. Generally this will not work and the build will fail.

This change alters the behavior of the scheduler. We will now always allow the implicit cached nodes to be updated, even after the global implicit cache mode is enabled. In this world, we would see:

```mermaid
flowchart TB
    classDef a fill:#F7DC6F,color:#000000
    classDef b fill:#85C1E9,color:#000000
    classDef b_imp fill:#A569BD,color:#000000
    classDef c fill:#F5B7B1,color:#000000
    
    pkgA_run:::a --> pkgA_build:::a --> pkgB-2_run:::b --> pkgB-2_build:::b
    pkgB-2_run:::b ~~~ /usr/bin/tool_b:::b_imp
    pkgB-2_run:::b --> /usr/bin/tool_b2(/usr/bin/tool_b - *version 2*):::b -.-> pkgB-2_build
    pkgA_build:::a --> pkgC_run:::c ~~~ /usr/bin/tool_b(/usr/bin/tool_b - *version 1*):::b_imp
    pkgC_run:::c --> /usr/bin/tool_b2(/usr/bin/tool_b - *version 2*):::b
```

### Implications:
Unfortunately, there is a downside to this change. Previously the scheduler's output could generally be considered deterministic, even if the internal flow to reach the end state was not. With this change, we can no longer guarantee that if we enable the global cached implicit mode.

Consider some package `pkgX`, that has two build dependencies: `/usr/bin/tool_b`, `pkgY`. It is otherwise not influenced by the other packages, it just needs `tool_b` and `pkgY` to build. As soon as we enable the cached implicit mode `pkgX` will start to build since the `/usr/bin/tool_b` (version 1) we have in the cache is sufficient (assume he `pkgY` built already). Jump forward a while, and we now rebuild `pkgB`. What should happen to `pkgX`?

In the old system it would always build with the older version of `pkgB`, no matter what. But now we have a new version. Depending on how long `pkgY` takes to build we might see either the new or old version of `/usr/bin/tool_b`.

Depending on how serious this is we may need to add a meachanism to invalidate an already building/built package and tell the scheduler to re-queue it (and anything downstream of it).

###### Change Log  

- Update cached implicit packages in the scheduler ever after we enter global implicit mode

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/45502828/

###### Test Methodology

- Local testing

### DIFF
--- a/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
+++ b/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
@@ -111,7 +111,7 @@ func implicitPackagesToUnresolvedNodesInGraph(pkgGraph *pkggraph.PkgGraph, useCa
 		// Only consider these as resolved if useCachedImplicit is set.
 		if n.State == pkggraph.StateCached {
 			if useCachedImplicit {
-				logger.Log.Warnf("Implicit node (%s) was already cached and may have been used to satisfy another node (useCachedImplicit=%v). Updating implicit path regardless!", n.FriendlyName(), useCachedImplicit)
+				logger.Log.Warnf("Implicit node (%s) was already cached and may have been used to satisfy another node (useCachedImplicit=true). Updating implicit path regardless!", n.FriendlyName())
 			}
 		} else if n.State != pkggraph.StateUnresolved {
 			continue
--- a/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
+++ b/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
@@ -111,7 +111,7 @@ func implicitPackagesToUnresolvedNodesInGraph(pkgGraph *pkggraph.PkgGraph, useCa
 		// Only consider these as resolved if useCachedImplicit is set.
 		if n.State == pkggraph.StateCached {
 			if useCachedImplicit {
-				continue
+				logger.Log.Warnf("Implicit node (%s) was already cached and may have been used to satisfy another node (useCachedImplicit=%v). Updating regardless implicit path anyways!", n.FriendlyName(), useCachedImplicit)
 			}
 		} else if n.State != pkggraph.StateUnresolved {
 			continue
--- a/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
+++ b/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
@@ -79,8 +79,9 @@ func replaceNodesWithProvides(res *BuildResult, pkgGraph *pkggraph.PkgGraph, pro
 	return
 }
 
-// implicitPackagesToUnresolvedNodesInGraph returns a map of package names to implicit nodes.
-func implicitPackagesToUnresolvedNodesInGraph(pkgGraph *pkggraph.PkgGraph, useCachedImplicit bool) (nameToNodes map[string][]*pkggraph.PkgNode) {
+// implicitPackageNamesToNodesInGraph returns a map of package names to implicit nodes. These nodes will either be unresolved,
+// or in the cache.
+func implicitPackageNamesToNodesInGraph(pkgGraph *pkggraph.PkgGraph, useCachedImplicit bool) (nameToNodes map[string][]*pkggraph.PkgNode) {
 	nameToNodes = make(map[string][]*pkggraph.PkgNode)
 
 	// Depending on the node order that the graph was created, there may be multiple unresolved nodes for a single package.
@@ -128,7 +129,7 @@ func implicitPackagesToUnresolvedNodesInGraph(pkgGraph *pkggraph.PkgGraph, useCa
 // matchProvidesToUnresolvedNodes matches a list of provides to unresolved nodes that they satisfy in the graph.
 func matchProvidesToUnresolvedNodes(provides []*pkgjson.PackageVer, pkgGraph *pkggraph.PkgGraph, useCachedImplicit bool) (matches map[*pkgjson.PackageVer][]*pkggraph.PkgNode, err error) {
 	matches = make(map[*pkgjson.PackageVer][]*pkggraph.PkgNode)
-	implicitPackagesToUnresolvedNodes := implicitPackagesToUnresolvedNodesInGraph(pkgGraph, useCachedImplicit)
+	implicitPackagesToUnresolvedNodes := implicitPackageNamesToNodesInGraph(pkgGraph, useCachedImplicit)
 
 	// An unresolved node can only be satisfied by a single provide, prevent duplicate matching
 	nodeToSatisfier := make(map[*pkggraph.PkgNode]*pkgjson.PackageVer)
--- a/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
+++ b/toolkit/tools/scheduler/schedulerutils/implicitprovides.go
@@ -79,7 +79,7 @@ func replaceNodesWithProvides(res *BuildResult, pkgGraph *pkggraph.PkgGraph, pro
 	return
 }
 
-// implicitPackagesToUnresolvedNodesInGraph returns a map of package names to unresolved implicit nodes.
+// implicitPackagesToUnresolvedNodesInGraph returns a map of package names to implicit nodes.
 func implicitPackagesToUnresolvedNodesInGraph(pkgGraph *pkggraph.PkgGraph, useCachedImplicit bool) (nameToNodes map[string][]*pkggraph.PkgNode) {
 	nameToNodes = make(map[string][]*pkggraph.PkgNode)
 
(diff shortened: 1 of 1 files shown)


toolkit/tools/scheduler/schedulerutils/implicitprovides.go
```diff
@@ -111,7 +111,7 @@ func implicitPackagesToUnresolvedNodesInGraph(pkgGraph *pkggraph.PkgGraph, useCa
 		// Only consider these as resolved if useCachedImplicit is set.
 		if n.State == pkggraph.StateCached {
 			if useCachedImplicit {
-				continue
+				logger.Log.Warnf("Implicit node (%s) was already cached and may have been used to satisfy another node (useCachedImplicit=%v). Updating regardless implicit path anyways!", n.FriendlyName(), useCachedImplicit)
```
dmcilvaney, line 114:
```suggestion
				logger.Log.Warnf("Implicit node (%s) was already cached and may have been used to satisfy another node (useCachedImplicit=%v). Updating implicit path regardless!", n.FriendlyName(), useCachedImplicit)
```

trungams:
> Notice we install both version 1 and version 2 of pkgB. Generally this will not work and the build will fail.

Could you explain this bit with more details. Is this a limitation on the scheduler side or the package installer side? If it's on the scheduler side, i.e something we can control, is having this fix (and therefore a non-deterministic scheduler output) better than other possible approaches? Would it work if we tried to delay using implicit nodes even more, since we can scan the graph and detect whether the node that provides the implicit dependencies will eventually be queued.

toolkit/tools/scheduler/schedulerutils/implicitprovides.go
```diff
@@ -111,7 +111,7 @@ func implicitPackagesToUnresolvedNodesInGraph(pkgGraph *pkggraph.PkgGraph, useCa
 		// Only consider these as resolved if useCachedImplicit is set.
 		if n.State == pkggraph.StateCached {
 			if useCachedImplicit {
-				continue
+				logger.Log.Warnf("Implicit node (%s) was already cached and may have been used to satisfy another node (useCachedImplicit=%v). Updating implicit path regardless!", n.FriendlyName(), useCachedImplicit)
```
anphel31, line 114: won't useCachedImplicit always be true in this log output?

anphel31 (approved):

dmcilvaney:
> > Notice we install both version 1 and version 2 of pkgB. Generally this will not work and the build will fail.
> 
> Could you explain this bit with more details. Is this a limitation on the scheduler side or the package installer side? If it's on the scheduler side, i.e something we can control, is having this fix (and therefore a non-deterministic scheduler output) better than other possible approaches? Would it work if we tried to delay using implicit nodes even more, since we can scan the graph and detect whether the node that provides the implicit dependencies will eventually be queued.

The scheduler doesn't trust tdnf to pick the right packages, so it does a BFS down from the current build node, getting a set of all build dependencies, and recursively the run-time dependencies of those packages. It then puts them all together into an install line and tries to install them as a single transaction.

Agreed we might be able to be smarter here, but it would require a fairly in-depth rework of some of the scheduler. This is an urgent fix for a build issue and we can return here to restore the determinism (in practice the non-deterministic behavior should not be a big problem)

toolkit/tools/scheduler/schedulerutils/implicitprovides.go
dmcilvaney: hah, true...
```suggestion
				logger.Log.Warnf("Implicit node (%s) was already cached and may have been used to satisfy another node (useCachedImplicit=true). Updating implicit path regardless!", n.FriendlyName())
```

toolkit/tools/scheduler/schedulerutils/implicitprovides.go
```diff
@@ -111,7 +111,7 @@ func implicitPackagesToUnresolvedNodesInGraph(pkgGraph *pkggraph.PkgGraph, useCa
 		// Only consider these as resolved if useCachedImplicit is set.
 		if n.State == pkggraph.StateCached {
 			if useCachedImplicit {
-				continue
+				logger.Log.Warnf("Implicit node (%s) was already cached and may have been used to satisfy another node (useCachedImplicit=true). Updating implicit path regardless!", n.FriendlyName())
```
PawelWMS, line 114: Can't comment on the right line: I think we should now also update the comment for this function, since now return a map of package names to implicit nodes, regardless if they were resolved byt a cached package or not.

PawelWMS: Actually, the name of the function itself might be a bit confusing.

PawelWMS (approved):
Just a request to update a comment and function name but functionally looks good.

trungams (approved):

PR #6290, opened 2023-09-26
Damcilva/2.0/tools/add scheduler stuck debug code

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

We are seeing some builds getting stuck while still reporting all packages were built. This should only happen if there is node stuck that isn't getting queued to build somehow. If the build fails, dump the state of the available nodes via `IsNodeAvailable()` recursively starting at the goal node. This is a similar mechanism as is used to queue nodes so we should see why the goal node isn't available (which is the "success" state for the scheduler).

###### Change Log  

- Add  temporary `debugStuckNode()` function to help diagnose scheduler issues.

**NO**

###### Test Methodology

- Local

### DIFF
--- a/toolkit/tools/scheduler/scheduler.go
+++ b/toolkit/tools/scheduler/scheduler.go
@@ -279,21 +279,18 @@ func startWorkerPool(agent buildagents.BuildAgent, workers, buildAttempts, check
 
 // debugStuckNode is a debugging function that will print out the stuck node and all nodes that are blocking it.
 func debugStuckNode(buildState *schedulerutils.GraphBuildState, pkgGraph *pkggraph.PkgGraph, stuckNode *pkggraph.PkgNode, indent int) {
-	isStuck := !buildState.IsNodeAvailable(stuckNode)
-	indentSpaces := ""
-	for i := 0; i < indent; i++ {
-		indentSpaces += "  "
+	if buildState.IsNodeAvailable(stuckNode) {
+		return
 	}
-	if isStuck {
-		logger.Log.Debugf("DEBUG: %s**STUCK** (%s)", indentSpaces, stuckNode.FriendlyName())
 
-		// Iterate over all the nodes that are blocking the stuck node.
-		dependency := pkgGraph.From(stuckNode.ID())
-		for dependency.Next() {
-			dependent := dependency.Node().(*pkggraph.PkgNode)
-			debugStuckNode(buildState, pkgGraph, dependent, indent+1)
-		}
+	nodeName := fmt.Sprintf("(%s)", stuckNode.FriendlyName())
+	logger.Log.Debugf("%*s", indent, nodeName)
 
+	// Iterate over all the nodes that are blocking the stuck node.
+	dependency := pkgGraph.From(stuckNode.ID())
+	for dependency.Next() {
+		dependent := dependency.Node().(*pkggraph.PkgNode)
+		debugStuckNode(buildState, pkgGraph, dependent, indent+1)
 	}
 }
 
--- a/toolkit/tools/scheduler/scheduler.go
+++ b/toolkit/tools/scheduler/scheduler.go
@@ -277,6 +277,25 @@ func startWorkerPool(agent buildagents.BuildAgent, workers, buildAttempts, check
 	return
 }
 
+func debugStuckNode(buildState *schedulerutils.GraphBuildState, pkgGraph *pkggraph.PkgGraph, stuckNode *pkggraph.PkgNode, indent int) {
+	isStuck := !buildState.IsNodeAvailable(stuckNode)
+	indentSpaces := ""
+	for i := 0; i < indent; i++ {
+		indentSpaces += "  "
+	}
+	if isStuck {
+		logger.Log.Errorf("DEBUG: %s**STUCK** (%s)", indentSpaces, stuckNode.FriendlyName())
+
+		// Iterate over all the nodes that are blocking the stuck node.
+		dependency := pkgGraph.From(stuckNode.ID())
+		for dependency.Next() {
+			dependent := dependency.Node().(*pkggraph.PkgNode)
+			debugStuckNode(buildState, pkgGraph, dependent, indent+1)
+		}
+
+	}
+}
+
 // buildAllNodes will build all nodes in a given dependency graph.
 // This routine only contains control flow logic for build scheduling.
 // It iteratively:
@@ -346,6 +365,7 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
 		if len(buildState.ActiveBuilds()) == 0 && len(channels.Results) == 0 {
 			if useCachedImplicit {
 				err = fmt.Errorf("could not build all packages")
+				debugStuckNode(buildState, pkgGraph, goalNode, 0)
 				break
 			} else {
 				logger.Log.Warn("Enabling cached packages to satisfy unresolved dynamic dependencies.")
@@ -400,6 +420,7 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
 						// When querying their edges, the graph library will return an empty iterator (graph.Empty).
 						pkgGraph = newGraph
 						goalNode = newGoalNode
+						logger.Log.Errorf("DEBUG: Updated goal node!")
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


toolkit/tools/scheduler/scheduler.go
```diff
@@ -346,6 +366,8 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
 		if len(buildState.ActiveBuilds()) == 0 && len(channels.Results) == 0 {
 			if useCachedImplicit {
 				err = fmt.Errorf("could not build all packages")
+				// Temporarily print debug information about the stuck node.
```
PawelWMS, line 369: I don't think it's temporary - it's useful info.

toolkit/tools/scheduler/scheduler.go
```diff
@@ -400,6 +422,8 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
 						// When querying their edges, the graph library will return an empty iterator (graph.Empty).
 						pkgGraph = newGraph
 						goalNode = newGoalNode
+						// Temporarily print debug information about the stuck nodes.
+						logger.Log.Debugf("DEBUG: Updated goal node!")
```
PawelWMS, line 426: Probably not needed considering the warning above.

toolkit/tools/scheduler/scheduler.go
```diff
@@ -277,6 +277,26 @@ func startWorkerPool(agent buildagents.BuildAgent, workers, buildAttempts, check
...
 
+// debugStuckNode is a debugging function that will print out the stuck node and all nodes that are blocking it.
+func debugStuckNode(buildState *schedulerutils.GraphBuildState, pkgGraph *pkggraph.PkgGraph, stuckNode *pkggraph.PkgNode, indent int) {
+	isStuck := !buildState.IsNodeAvailable(stuckNode)
+	indentSpaces := ""
+	for i := 0; i < indent; i++ {
+		indentSpaces += "  "
+	}
+	if isStuck {
+		logger.Log.Debugf("DEBUG: %s**STUCK** (%s)", indentSpaces, stuckNode.FriendlyName())
```
anphel31, line 288: wondering if this should be a Warning message so it shows up easier in  the pipeline builds?

toolkit/tools/scheduler/scheduler.go
PawelWMS, line 426: Probably not needed considering the warning above.
```suggestion
```

toolkit/tools/scheduler/scheduler.go
```diff
@@ -277,6 +277,26 @@ func startWorkerPool(agent buildagents.BuildAgent, workers, buildAttempts, check
 	return
 }
 
+// debugStuckNode is a debugging function that will print out the stuck node and all nodes that are blocking it.
+func debugStuckNode(buildState *schedulerutils.GraphBuildState, pkgGraph *pkggraph.PkgGraph, stuckNode *pkggraph.PkgNode, indent int) {
```
PawelWMS, line 281: I've noticed this right now: at the end of each successful package build we run `graphanalytics`, which already prints a nice summary of what got blocked. I'm curious, if instead of adding this function, we could run the logic from that one but regardless if a build failure or succeeded. Either directly from the scheduler or through modifying our make files.
image

toolkit/tools/scheduler/scheduler.go
```diff
@@ -277,6 +277,26 @@ func startWorkerPool(agent buildagents.BuildAgent, workers, buildAttempts, check
...
+	isStuck := !buildState.IsNodeAvailable(stuckNode)
+	indentSpaces := ""
+	for i := 0; i < indent; i++ {
+		indentSpaces += "  "
+	}
```
PawelWMS, line 286:
```suggestion
	indentSpaces := strings.Repeat(" ", utf8.RuneCountInString(indent))
```

toolkit/tools/scheduler/scheduler.go
```diff
@@ -277,6 +277,26 @@ func startWorkerPool(agent buildagents.BuildAgent, workers, buildAttempts, check
 	return
 }
 
+// debugStuckNode is a debugging function that will print out the stuck node and all nodes that are blocking it.
+func debugStuckNode(buildState *schedulerutils.GraphBuildState, pkgGraph *pkggraph.PkgGraph, stuckNode *pkggraph.PkgNode, indent int) {
+	isStuck := !buildState.IsNodeAvailable(stuckNode)
```
PawelWMS, line 282: Since these chains may be long/deep, I'd focus only on the stuck nodes. I think we don't care much about the ones, which were fine and we can make reading this output simpler on the developer:
```suggestion
	if buildState.IsNodeAvailable(stuckNode) {
		return
	}
```

toolkit/tools/scheduler/scheduler.go
PawelWMS: Alternatively, see my slightly bigger suggestion below with string padding.

toolkit/tools/scheduler/scheduler.go
```diff
@@ -277,6 +277,26 @@ func startWorkerPool(agent buildagents.BuildAgent, workers, buildAttempts, check
...
+		logger.Log.Debugf("DEBUG: %s**STUCK** (%s)", indentSpaces, stuckNode.FriendlyName())
+
+		// Iterate over all the nodes that are blocking the stuck node.
+		dependency := pkgGraph.From(stuckNode.ID())
+		for dependency.Next() {
+			dependent := dependency.Node().(*pkggraph.PkgNode)
+			debugStuckNode(buildState, pkgGraph, dependent, indent+1)
+		}
+
+	}
```
PawelWMS, line 297: Since these chains may be long/deep, I'd focus only on the stuck nodes. I think we don't care much about the ones, which were fine and we can make reading this output simpler on the developer.

I've also included one more suggestion about using [string padding](https://golang.cafe/blog/golang-string-padding-example.html).
```suggestion
	const indentWidth = 4

	if buildState.IsNodeAvailable(stuckNode) {
		return
	}

	nodeName := fmt.Sprintf("(%s)", stuckNode.FriendlyName())
	logger.Log.Debugf("%*s", indent * indentWidth, nodeName)

	// Iterate over all the nodes that are blocking the stuck node.
	dependency := pkgGraph.From(stuckNode.ID())
	for dependency.Next() {
		dependent := dependency.Node().(*pkggraph.PkgNode)
		debugStuckNode(buildState, pkgGraph, dependent, indent+1)
	}
```

toolkit/tools/scheduler/scheduler.go
```diff
@@ -346,6 +366,8 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
 		if len(buildState.ActiveBuilds()) == 0 && len(channels.Results) == 0 {
 			if useCachedImplicit {
 				err = fmt.Errorf("could not build all packages")
+				// Temporarily print debug information about the stuck node.
+				debugStuckNode(buildState, pkgGraph, goalNode, 0)
```
PawelWMS, line 370: I'd wrap this function in another to print an initial message saying we're printing a stack of nodes, on which node `goalNode` was stuck on. This way we won't have to keep printing "STUCK" on each line, because it's going to be clear from that initial message.

toolkit/tools/scheduler/scheduler.go
PawelWMS: Good call. I'm thinking we might even want to do an error message here, since this is something that's not supposed to happen and it breaks the build.

toolkit/tools/scheduler/scheduler.go
dmcilvaney: This will generate a HUGE amount of output for a "normal" failed build (I could see thousands of lines easily enough) We already print out details about the stuck RPMs if everything is working as intended.

toolkit/tools/scheduler/scheduler.go
dmcilvaney: I can see if I can enlighten `PrintBuildSummary()` to do something similar, but the code will likely be more complex than this. I don't think we know before then if this is a "normal" failure where we get everything we need in the summary, or an abnormal failure where the blocking node isn't showing up.

dmcilvaney:
If we are going to keep this long-term, I'm going to redo it completely. This was written as a stopgap to help diagnose the build issues but its not suitable to keep as is.

toolkit/tools/scheduler/scheduler.go
dmcilvaney: At this point I think it will be hard to reconstruct the failure state. The analytics tool only considers the node state/type (ie unresolved remote nodes). As best I can tell the failures we are seeing are in the actual scheduler state rather than the graph.

toolkit/tools/scheduler/scheduler.go
dmcilvaney: Used arbitrary length padding below

PawelWMS (approved):

PR #6314, opened 2023-09-28
Performance Improvement: Add support for downloading/uploading ccache archives

gmileka:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Add support for downloading/uploading ccache archives.
- User can provide a configuration file or use the default included one.
- The configuration file specifies:
  -  what packages can share the ccache during their build.
  - what azure blob storage/container to use for downloading/uploading.
 - The build will now try to download an archived ccache from the remote store right before the package starts building.
 - The build will upload the created ccache artifacts after the package build if the ccache is not shared.
 - After the first build, subsequent builds will take less time to complete.

###### Change Log  

- Change
- Change
- Change

**YES**
With this change, building the tool chain may use compiled units from previous build if ccache is enabled in the build pipelines.

###### Associated issues  
- n/a

###### Links to CVEs  
- n/a

###### Test Methodology

- In progress...


toolkit/tools/internal/ccachemanager/ccachemanager.go
```diff
@@ -0,0 +1,552 @@
...
+
+		// Read the text contents...
+		latestBuildTag, err := ioutil.ReadFile(m.PkgTagFile.LocalSourcePath)
+		if err != nil {
+			logger.Log.Warnf("Unable to read ccache tag file contents. Error: %v", err)
+			return err
+		}
+
+		// Adjust the download folder from 'latest' to the tag loaded from the file...
+		remoteStoreConfig.DownloadFolder = string(latestBuildTag)
```
gmileka, line 384: This logic needs to be update. Updating the DownloadFolder has no effect on the download path at this point (it used to in a previous code iteration). Working on a fix now and will update the PR.

toolkit/resources/manifests/package/ccache-configuration.json
```diff
@@ -0,0 +1,5674 @@
...
+    "storageAccount": "marinerccache",
+    "containerName": "20-stable",
+    "tagsFolder": "tags",
+    "downloadEnabled": true,
+    "downloadFolder": "latest",
+    "uploadEnabled": false,
+    "uploadFolder": "",
+    "updateLatest": false
+  },
+  "groups": [
```
anphel31, line 16: Does this file cover every core package spec? Will we need to update it as new specs are added or removed?

toolkit/tools/scheduler/scheduler.go
```diff
@@ -189,6 +192,20 @@ func main() {
...
 	}
+
+	if *useCcache {
+		logger.Log.Infof("  ccache is enabled. processing created artifacts under (%s)...", *ccacheDir)
+		var ccacheManager ccachemanagerpkg.CCacheManager
+		err = ccacheManager.Initialize(*ccacheDir, *ccacheConfig)
+		if err == nil {
+			err = ccacheManager.UploadAllPkgGroupCCaches()
+			if err != nil {
+				logger.Log.Warnf("Failed to archive CCache artifacts. Error: %s.", err)
```
anphel31, line 203: Most other errors are printed as %v - should this be as well?

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -188,6 +212,15 @@ func buildSRPMInChroot(chrootDir, rpmDirPath, toolchainDirPath, workerTar, srpmF
 		builtRPMs, err = moveBuiltRPMs(chroot.RootDir(), rpmDirPath)
 	}
 
+	// Only if the groupSize is 1 we can archive since no other packages will
+	// re-update this cache.
+	if useCcache && ccacheManager.CurrentPkgGroup.Size == 1 {
+		err = ccacheManager.UploadPkgGroupCCache()
+		if err != nil {
+			logger.Log.Infof("  unable to upload ccache archive. Error: %v", err)
```
anphel31, line 220: should this be a warning (or error?)

toolkit/resources/manifests/package/ccache-configuration.json
dmcilvaney: Would it be better to have a list of exceptions? Curious how many packages aren't 1:1 mappings. (ie how many are omitted, or have multiple together).

gmileka: The packages in this file are the ones where the tools will download/upload. We'd need to maintain the list every time a new package is added/removed - which is not ideal.
I agree a list of exceptions is a better option. Originally, I thought we will have a small set of packages that we want ccache download/upload enabled for to avoid size getting out of control.
After doing a full build, the entire ccache is about 30GB - which is manageable. So, we can keep the long list of packages.
I'll convert it into a list of exceptions - where:
- by default, all packages have ccache enable and they belong to their own group.
- groups that contain multiple packages will need to be defined in this file.
- groups that are disabled will need to be defined in this file.

That way, we don't need to maintain the file as much - addressing Andrew's point as well.

gmileka: @anphel31 @dmcilvaney  - I've updated to logic as follows:
- the default for any package is that ccache is enabled. It will build in its own folder by default as well.
- the config is used to:
  - disable ccache for certain packages.
  - create package groups.

toolkit/tools/pkgworker/pkgworker.go
gmileka: Fixed.

toolkit/tools/scheduler/scheduler.go
gmileka: Fixed.

anphel31:
Do these changes affect how local builds with Ccache work? Suppose a developer wants to iterate locally by calling `build-packages` - previously they only needed to also set `USE_CCACHE=y` and a local ccache directory would be created and used. Just wondering if that is still the case, or are there any extra steps requried?

gmileka:
> Do these changes affect how local builds with Ccache work? Suppose a developer wants to iterate locally by calling `build-packages` - previously they only needed to also set `USE_CCACHE=y` and a local ccache directory would be created and used. Just wondering if that is still the case, or are there any extra steps requried?

Yes, that is still the case. Just setting `USE_CCACHE=y` will allow the packages build to use the default ccache configuration which include: (1) downloading an archived ccache directory for the packages to build and (2) enable ccache as before.

If the user does not want the default behavior of downloading archives, the user can disable it in the default configuration.

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
```diff
@@ -0,0 +1,148 @@
...
+	localFileName string) (err error) {
+
+	downloadStartTime := time.Now()
+
+	localFile, err := os.Create(localFileName)
+	if err != nil {
+		logger.Log.Infof("  failed to create local file for download. Error: %v", err)
+		return err
+	}
+	defer localFile.Close()
```
dmcilvaney, line 67: We ran into an issue with our network function were we would create an empty file, fail the download, and then leave the file. Might want a similar cleanup here. (see the `defer func()...` in `network.go:DownloadFile()`.

toolkit/tools/internal/ccachemanager/ccachemanager.go
```diff
@@ -0,0 +1,697 @@
...
+
+	logger.Log.Infof("  checking if (%s) already exists...", g.TagFile.LocalSourcePath)
+	_, err := os.Stat(g.TagFile.LocalSourcePath)
+	if err != nil {
+		// If file is not available locally, try downloading it...
+		logger.Log.Infof("  downloading (%s) to (%s)...", g.TagFile.RemoteSourcePath, g.TagFile.LocalSourcePath)
+		err = azureBlobStorage.Download(context.Background(), containerName, g.TagFile.RemoteSourcePath, g.TagFile.LocalSourcePath)
+		if err != nil {
+			logger.Log.Warnf("  unable to download ccache tag file.")
+			return "", err
```
dmcilvaney, line 143: We are trying to transition to wrapped errors while minimizing logging as we update code. This would be:
```suggestion
			err = fmt.Errorf("unable to download ccache tag file:\n%w",err)
			return "", err
```

dmcilvaney: If the warning logs still make sense, feel free to ignore this.

toolkit/tools/internal/ccachemanager/ccachemanager.go
```diff
@@ -0,0 +1,697 @@
...
+		}
+	}
+
+	latestBuildTagData, err := ioutil.ReadFile(g.TagFile.LocalSourcePath)
+	if err != nil {
+		logger.Log.Warnf("Unable to read ccache tag file contents. Error: %v", err)
+		return "", err
```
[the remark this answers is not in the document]
dmcilvaney: See above, will just omit for the rest.

toolkit/tools/internal/ccachemanager/ccachemanager.go
```diff
@@ -0,0 +1,697 @@
...
+	logger.Log.Infof("    DownloadFolder : %s", configuration.RemoteStoreConfig.DownloadFolder)
+	logger.Log.Infof("    UploadEnabled  : %v", configuration.RemoteStoreConfig.UploadEnabled)
+	logger.Log.Infof("    UploadFolder   : %s", configuration.RemoteStoreConfig.UploadFolder)
+	logger.Log.Infof("    UpdateLatest   : %v", configuration.RemoteStoreConfig.UpdateLatest)
+	logger.Log.Infof("    KeepLatestOnly : %v", configuration.RemoteStoreConfig.KeepLatestOnly)
+
+	return configuration, err	
+}
+
+func ensureDirExists(dirName string) (err error) {
```
dmcilvaney, line 247: Possible candidate for `directory.go`.

toolkit/tools/internal/ccachemanager/ccachemanager.go
```diff
@@ -0,0 +1,697 @@
...
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/azureblobstorage"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/jsonutils"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
+)
+
```
dmcilvaney, line 20: Could you add a description either here, or spread out through the fuctions, to help people ramping up? Just high level flow stuff.

toolkit/resources/manifests/package/ccache-configuration.json
```diff
@@ -0,0 +1,74 @@
+{
+  "remoteStore": {
+    "type": "azure-blob-storage",
+    "tenantId": "",
+    "userName": "",
+    "password": "",
```
dmcilvaney, line 6: Is the idea that a local user doesn't need a password to download, and a pipeline will already be authenticated and won't need to login again?

.gitignore
```diff
@@ -64,6 +64,8 @@ out/
 
 # Builds may generate a CCache directory.
 ccache/
+ccache-downloads/
```
dmcilvaney, line 67: Are these still used? I couldn't find anything outide `ccache/`.

toolkit/tools/internal/ccachemanager/ccachemanager.go
gmileka: None of ccache failures should indicate a problem with the build or the generated bits. They all just means that an optimization for build time failed to take place. So, I think using a warning for all of them will make them less distracting for people looking for actual blocking failures.
The reason I have remove the printing of the error is because the error has already been printed out by then. I initially printed the error ever where - and that results in seeing the same error 4 times in the logs (as it travels the stack). This made it very hard to know how many errors of that particular type occured (not readable).

toolkit/resources/manifests/package/ccache-configuration.json
gmileka: Correct re. a local user won't need a password to download. However, it is because download is allowed to be anonymous. So, a dev building on a local box will automatically get the benefit of ccache remote archive without haivng to log-in.

.gitignore
gmileka: Not anymore - thx. I moved them inside the ccache folder itself. I'll remove.

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
gmileka: I indeed ran into this problem while testing on a non-clean machine (dev scenario). I've solved it by having the Download() call downloadInternal() and then clean-up. I'll merge and use the `defer func()` pattern as it's more contained.

gmileka: Updated to match network.go:DownloadFile().

toolkit/tools/internal/ccachemanager/ccachemanager.go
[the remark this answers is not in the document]
gmileka: Changed the lower level object (azureblobstorage and ccachemanager) to not print any warnings/errors and instead wrap their errors and return them.

toolkit/tools/internal/ccachemanager/ccachemanager.go
gmileka: Moved.

toolkit/tools/internal/ccachemanager/ccachemanager.go
gmileka: Added.

.gitignore
gmileka: Removed.

dmcilvaney (approved):

toolkit/tools/internal/ccachemanager/ccachemanager.go
```diff
@@ -0,0 +1,769 @@
...
+	tarArgs := []string{
+		"cf",
+		archiveName,
+		"-C",
+		sourceDir,
+		"."}
+
+	_, stderr, err := shell.Execute("tar", tarArgs...)
+	if err != nil {
+		return fmt.Errorf("Unable compress ccache files itno archive:\n%s", stderr)
```
anphel31, line 412: nit: typo on `into`. also should the `stderr` be printed with `$w` like the others?

toolkit/tools/internal/ccachemanager/ccachemanager.go
```diff
@@ -0,0 +1,769 @@
...
+	tarArgs := []string{
+		"xf",
+		archiveName,
+		"-C",
+		targetDir,
+		"."}
+
+	_, stderr, err := shell.Execute("tar", tarArgs...)
+	if err != nil {
+		return fmt.Errorf("Unable extract ccache files from archive:\n%s", stderr)
```
anphel31, line 431: print error with `%w` for consistency?

toolkit/tools/internal/ccachemanager/ccachemanager.go
```diff
@@ -0,0 +1,769 @@
...
+
+	remoteStoreConfig := m.Configuration.RemoteStoreConfig
+	if !remoteStoreConfig.UploadEnabled {
+		logger.Log.Infof("  ccache update is disabled for this build.")
+		return nil
+	}
+
+	err = compressDir(m.CurrentPkgGroup.CCacheDir, m.CurrentPkgGroup.TarFile.LocalTargetPath)
+	if err != nil {
+		return fmt.Errorf("Unable compress ccache files itno archive:\n%w", err)
```
anphel31, line 621: nit: typo on `into`

toolkit/tools/internal/ccachemanager/ccachemanager.go
```diff
@@ -0,0 +1,769 @@
...
+	architectures, err := directory.GetChildDirs(m.RootWorkDir)
+	errorsOccured := false
+	if err != nil {
+		return fmt.Errorf("failed to enumerate ccache child folders under (%s):\n%w", m.RootWorkDir, err)
+	}
+
+	for _, architecture := range architectures {
+		groupNames, err := directory.GetChildDirs(filepath.Join(m.RootWorkDir, architecture))
+		if err != nil {
+			logger.Log.Warnf("failed to enumerate child folders under (%s):\n%v", m.RootWorkDir, err)
```
anphel31, line 722: is `%v` intentional for printing err or should it be `%w`?

anphel31: same comment for a few other Warnf() calls in this file

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -85,14 +89,36 @@ func main() {
...
+		if *useCcache {
+			buildArch, err := rpm.GetRpmArch(runtime.GOARCH)
+			if err == nil {
+				err = ccacheManager.SetCurrentPkgGroup(*basePackageName, buildArch)
+				if err == nil {
+					if ccacheManager.CurrentPkgGroup.Enabled {
+						defines[rpm.MarinerCCacheDefine] = "true"
+					}
+				} else {
+					logger.Log.Warnf("Failed to set package ccache configuration:\n%v", err)
```
anphel31, line 104: use `%w` instead of `%v` in this file?

anphel31 (approved):

toolkit/tools/internal/ccachemanager/ccachemanager.go
gmileka: It led to a type mismatch compilation error when I tried %w. I'll fix the typo.

gmileka: Here's the error I'm seeing: `fmt.Errorf format %w has arg stderr of wrong type string`

toolkit/tools/internal/ccachemanager/ccachemanager.go
gmileka: Herre's the error I'm seeing `fmt.Errorf format %w has arg stderr of wrong type string`.

toolkit/tools/internal/ccachemanager/ccachemanager.go
gmileka: I get an error: `(*github.com/sirupsen/logrus.Logger).Warnf does not support error-wrapping directive %w`

toolkit/tools/pkgworker/pkgworker.go
gmileka: I get an error `(*github.com/sirupsen/logrus.Logger).Warnf does not support error-wrapping directive %w`.

toolkit/tools/internal/ccachemanager/ccachemanager.go
gmileka: Fixed.

PR #6323, opened 2023-09-29
Image customizer: Implement `PostInstallScripts` and `FinalizeImageScripts`.

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add and implement the `PostInstallScripts` and `FinalizeImageScripts` config values in the image customizer.

###### Change Log  

- Image customizer: Implement `PostInstallScripts` and `FinalizeImageScripts`.

NO

###### Test Methodology

- Ran toolkit UTs.
- Manually run image customizer.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
@@ -22,6 +22,9 @@ const (
 func doCustomizations(baseConfigPath string, config *imagecustomizerapi.SystemConfig, imageChroot *safechroot.Chroot) error {
 	var err error
 
+	// Note: The ordering of the customization steps here should try to mirror the order of the equivalent steps in imager
+	// tool as closely as possible.
+
 	err = copyAdditionalFiles(baseConfigPath, config.AdditionalFiles, imageChroot)
 	if err != nil {
 		return err
--- a/toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
@@ -4,17 +4,35 @@
 package imagecustomizerlib
 
 import (
+	"fmt"
 	"io/fs"
 	"path/filepath"
 
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/imagecustomizerapi"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safemount.go"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	configDirMountPathInChroot = "/_imageconfigs"
 )
 
 func doCustomizations(baseConfigPath string, config *imagecustomizerapi.SystemConfig, imageChroot *safechroot.Chroot) error {
 	var err error
 
 	err = copyAdditionalFiles(baseConfigPath, config.AdditionalFiles, imageChroot)
+	if err != nil {
+		return err
+	}
+
+	err = runScripts(baseConfigPath, config.PostInstallScripts, imageChroot)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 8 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -42,3 +60,39 @@ func copyAdditionalFiles(baseConfigPath string, additionalFiles map[string]image
...
+		return err
+	}
+	defer mount.Close()
+
+	for _, script := range scripts {
+		scriptPathInChroot := filepath.Join(configDirMountPathInChroot, script.Path)
+		command := fmt.Sprintf("%s %s", scriptPathInChroot, script.Args)
+
+		// Run the script.
+		err = imageChroot.UnsafeRun(func() error {
```
dmcilvaney, line 79: Do we need `UnsafeRun()` here? I think the intent of Unsafe is to allow two chroot operations to happen at the same time, might not be needed here since I don't think there is a nested chroot here (at least I didn't notice one lookup up the call stack).

toolkit/tools/pkg/imagecustomizerlib/testdata/runscripts-config.yaml
```diff
@@ -0,0 +1,5 @@
+PostInstallScripts:
```
dmcilvaney, line 1: Is this missing from `customizeutils_test.go`? Looks like `addfiles-config.yaml` is getting used.

cwize1: No. The install scripts need a full image to be able to run. So, it is not really practical to add tests for them to the current set of UTs. This file exists just to assist with manual testing.

At some point, we should probably add a set of tier 2 tests that can have more requirements and allows to run for longer.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: The imager/installutils code also uses `UnsafeRun` for running these scripts. So, that logic is mirrored here.

I think this might be because the imager tool runs most of the image build logic under the setup chroot. So, the commands run against the install chroot are "nested" chroot commands. Hence using `Run` won't work.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -124,6 +126,30 @@ func validateConfig(baseConfigPath string, config *imagecustomizerapi.SystemConf
...
+		err = validateScript(baseConfigPath, &script)
+		if err != nil {
+			return fmt.Errorf("invalid FinalizeImageScripts item at index %d: %w", i, err)
+		}
+	}
+
+	return nil
+}
+
+func validateScript(baseConfigPath string, script *imagecustomizerapi.Script) error {
```
romoh, line 146: maybe also validate it has execution permission?

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -19,6 +27,16 @@ func doCustomizations(baseConfigPath string, config *imagecustomizerapi.SystemCo
 		return err
 	}
 
+	err = runScripts(baseConfigPath, config.PostInstallScripts, imageChroot)
+	if err != nil {
+		return err
+	}
+
+	err = runScripts(baseConfigPath, config.FinalizeImageScripts, imageChroot)
```
romoh, line 35: Might be a good idea to add a comment here that this should be kept as the final step.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: Done.

romoh (approved):

vinceaperri (approved):

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
dmcilvaney: That would make sense.

dmcilvaney (approved):

PR #6329, opened 2023-10-02
Update go-test-coverage.yml to verify go version.

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

We were not correctly detecting build breaks in our go tools due to tooling version changes. The Github Action worker was always using a very recent version of the Go tools and this could mask breaking changes.

We use `sudo` to run out tools, which clears the environment of the calling user by default. The action used to install Go does so by modifying the user's PATH. We need to explicitly pass this updated PATH to the tools if we want them to use the correct go.

Also add a check that validates the workflow against the `go.mod` file and alert if they get out of sync.

###### Change Log  

- Add a step to run `sudo make go-tools REBUILD_TOOLS=y` to the `go-test-coverage.yml` tests.
- Use correct version of go tools
- Validate go.mod against workflow.

**NO**

### DIFF
--- a/.github/workflows/go-test-coverage.yml
+++ b/.github/workflows/go-test-coverage.yml
@@ -9,21 +9,38 @@ on:
   pull_request:
     branches: [main, dev, 1.0*, 2.0*, fasttrack/*]
 
-jobs:
+env:
+  EXPECTED_GO_VERSION: "1.20"
 
+jobs:
   build:
     name: Go Test Coverage
     runs-on: ubuntu-latest
     steps:
 
     - name: Set up Go 1.x
-      uses: actions/setup-go@v2
+      uses: actions/setup-go@v4
       with:
-        go-version: 1.19
+        go-version: '${{ env.EXPECTED_GO_VERSION }}'
       id: go
+
+    - name: Check active go version
+      run: |
+        go version && which go
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


cwize1 (approved):

dmcilvaney:
This didn't capture the expected build failure... pipeline may be using a more recent version of go.

dmcilvaney:
This didn't capture the expected build failure... pipeline may be using a more recent version of go.

.github/workflows/go-test-coverage.yml
```diff
@@ -9,22 +9,39 @@ on:
...
     - name: Check out code into the Go module directory
       uses: actions/checkout@v4
 
+    - name: Check go.mod
+      run: |
+        if grep -q "go $EXPECTED_GO_VERSION" ./toolkit/tools/go.mod; then
+          echo "go.mod has correct version ($EXPECTED_GO_VERSION)"
+        else
+          actual_version="$(grep -E '^go [0-9]+\.[0-9]+' ./toolkit/tools/go.mod)"
+          echo "go.mod has bad version expected:$EXPECTED_GO_VERSION, found: $actual_version"
```
PawelWMS, line 40: I'd maybe reverse the message - if someone changed `go.mod`, then it's probably intentional and they just didn't realize they have to update this check as well.

cwize1 (approved):

PR #6333, opened 2023-10-02
Add explicit timeout to package builds

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Sometimes package builds (or more often tests) get stuck. Generally, pipeline builds will timeout and kill the task, but we don't get our log output etc. in this case. We should instead gracefully monitor package builds and fail them if they take too long (say `4h`).

If the call to `rpmbuild` in the worker does not complete within that timeframe the worker will kill the `rpmbuild` process and return an error to the scheduler. The scheduler will be able to continue running as normal and will report this failure as normal at the end.

This can be configured by setting `PACKAGE_BUILD_TIMEOUT=10h` for 10 hours, `30m` for 30 minutes, etc.

###### Change Log  

- Add `--timeout` to scheduler and workers which kills a build if it runs too long
- Add `PACKAGE_BUILD_TIMEOUT=4h` to Makefile to control timeout

**NO**

###### Test Methodology

- Local

### DIFF
--- a/toolkit/scripts/incremental_building.mk
+++ b/toolkit/scripts/incremental_building.mk
@@ -78,6 +78,7 @@ REBUILD_TOOLS                   ?= n
 DELTA_BUILD                     ?= n
 CLEAN_TOOLCHAIN_CONTAINERS      ?= y
 MAX_CPU                         ?=
+PACKAGE_BUILD_TIMEOUT           ?= 4h
 DELTA_FETCH                     ?= n
 PRECACHE                        ?= n
 MAX_CASCADING_REBUILDS          ?=
--- a/toolkit/tools/pkgworker/pkgworker.go
+++ b/toolkit/tools/pkgworker/pkgworker.go
@@ -22,6 +22,7 @@ import (
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/sliceutils"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/tdnf"
+	"golang.org/x/sys/unix"
 
 	"gopkg.in/alecthomas/kingpin.v2"
 )
@@ -54,6 +55,7 @@ var (
 	outArch              = app.Flag("out-arch", "Architecture of resulting package").String()
 	useCcache            = app.Flag("use-ccache", "Automatically install and use ccache during package builds").Bool()
 	maxCPU               = app.Flag("max-cpu", "Max number of CPUs used for package building").Default("").String()
+	timeout              = app.Flag("timeout", "Timeout for package building").Required().Duration()
 
 	logFile  = exe.LogFileFlag(app)
 	logLevel = exe.LogLevelFlag(app)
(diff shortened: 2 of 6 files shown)


dmcilvaney:
Open question: Should we set a restrictive default here (`4h`)? Or should we only set a timeout when running builds in the pipeline? I'm mostly worried about breaking niche scenarios like emulated cross-compiling via qemu etc. Those builds can be VERY slow.

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -177,9 +179,23 @@ func buildSRPMInChroot(chrootDir, rpmDirPath, toolchainDirPath, workerTar, srpmF
...
+		})
+		results <- buildErr
+	}()
+
+	select {
+	case err = <-results:
+	case <-time.After(timeout):
+		shell.PermanentlyStopAllProcesses(unix.SIGKILL)
+		logger.Log.Errorf("Timeout after %v: killing all processes in chroot...", timeout)
+		err = fmt.Errorf("build timed out after %s", timeout)
```
PawelWMS, line 196: Can we either depend on the top-level caller to print the error or print it only here? It seems like we can end up with a duplicate error message.

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -177,9 +179,23 @@ func buildSRPMInChroot(chrootDir, rpmDirPath, toolchainDirPath, workerTar, srpmF
...
+		buildErr := chroot.Run(func() (err error) {
+			return buildRPMFromSRPMInChroot(srpmFileInChroot, outArch, runCheck, defines, packagesToInstall, useCcache)
+		})
+		results <- buildErr
+	}()
+
+	select {
+	case err = <-results:
+	case <-time.After(timeout):
+		shell.PermanentlyStopAllProcesses(unix.SIGKILL)
```
PawelWMS, line 194: Do you know which processed this kills? I'm guessing it's only the children, but in that case, could we re-name it to something like `PermanentlyStopAllChildrenProcesses` just to be more clear?

toolkit/scripts/incremental_building.mk
```diff
@@ -78,6 +78,7 @@ REBUILD_TOOLS                   ?= n
 DELTA_BUILD                     ?= n
 CLEAN_TOOLCHAIN_CONTAINERS      ?= y
 MAX_CPU                         ?=
+PACKAGE_BUILD_TIMEOUT           ?= 4h
```
PawelWMS, line 81: I'd start with something even longer (8 hours?) for the first attempt. Our pipelines build agents are recently faster, so `rust` doesn't take forever, but we don't know, what other people/teams are using.

PawelWMS (approved):

toolkit/tools/pkgworker/pkgworker.go
dmcilvaney: Specifically, it sends the signal to everything tracked by `shell.activeCommands` which is a map of running commands spawned by the `shell` package.

I'm fine with updating the name.

toolkit/tools/pkgworker/pkgworker.go
dmcilvaney: The top-level error gets delayed a lot by the cleanup flows. (~30 lines too late). Having this hear makes it clear why we suddenly stopped doing normal build stuff and started running `kill` and `unmount` etc. I'd like to leave it.

toolkit/scripts/incremental_building.mk
dmcilvaney: Done

toolkit/tools/pkgworker/pkgworker.go
dmcilvaney: Done

gmileka (approved):

PR #6336, opened 2023-10-02
Image Customizer: Add/update users.

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add the ability to add or update users to the Image Customizer tool.

###### Change Log  

- Image Customizer: Add/update users.
- Moved a bunch of user utilities to their own module.

NO

###### Test Methodology

### DIFF
--- a/toolkit/tools/imagecustomizerapi/user.go
+++ b/toolkit/tools/imagecustomizerapi/user.go
@@ -5,24 +5,45 @@ package imagecustomizerapi
 
 import (
 	"fmt"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/userutils"
 )
 
 type User struct {
-	Name                string   `json:"Name"`
-	UID                 string   `json:"UID"`
-	PasswordHashed      bool     `json:"PasswordHashed"`
-	Password            string   `json:"Password"`
-	PasswordPath        string   `json:"PasswordPath"`
-	PasswordExpiresDays int64    `json:"PasswordExpiresDays"`
-	SSHPubKeyPaths      []string `json:"SSHPubKeyPaths"`
-	PrimaryGroup        string   `json:"PrimaryGroup"`
-	SecondaryGroups     []string `json:"SecondaryGroups"`
-	StartupCommand      string   `json:"StartupCommand"`
+	Name                string   `yaml:"Name"`
+	UID                 *int     `yaml:"UID"`
+	PasswordHashed      bool     `yaml:"PasswordHashed"`
+	Password            string   `yaml:"Password"`
+	PasswordPath        string   `yaml:"PasswordPath"`
+	PasswordExpiresDays *int64   `yaml:"PasswordExpiresDays"`
+	SSHPubKeyPaths      []string `yaml:"SSHPubKeyPaths"`
+	PrimaryGroup        string   `yaml:"PrimaryGroup"`
+	SecondaryGroups     []string `yaml:"SecondaryGroups"`
+	StartupCommand      string   `yaml:"StartupCommand"`
 }
 
 func (u *User) IsValid() error {
+	err := userutils.NameIsValid(u.Name)
+	if err != nil {
+		return fmt.Errorf("user (%s) is invalid:\n%w", u.Name, err)
+	}
+
+	if u.UID != nil {
+		err := userutils.UIDIsValid(*u.UID)
+		if err != nil {
+			return fmt.Errorf("user (%s) is invalid:\n%w", u.Name, err)
+		}
+	}
+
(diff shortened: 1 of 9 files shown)


toolkit/tools/imagecustomizerapi/user.go
```diff
@@ -0,0 +1,29 @@
...
+	Password            string   `json:"Password"`
+	PasswordPath        string   `json:"PasswordPath"`
+	PasswordExpiresDays int64    `json:"PasswordExpiresDays"`
+	SSHPubKeyPaths      []string `json:"SSHPubKeyPaths"`
+	PrimaryGroup        string   `json:"PrimaryGroup"`
+	SecondaryGroups     []string `json:"SecondaryGroups"`
+	StartupCommand      string   `json:"StartupCommand"`
+}
+
+func (u *User) IsValid() error {
```
romoh, line 23: The toolkit doesn't allow the usage of "mariner" as a user. We should probably align to that here too.

toolkit/tools/internal/userutils/userutils.go
```diff
@@ -0,0 +1,362 @@
...
+		}
+
+		logger.Log.Infof("Adding ssh key (%s) to user (%s) .ssh/authorized_users", filepath.Base(pubKey), username)
+		pubKeyData, err = file.ReadLines(pubKey)
+		if err != nil {
+			logger.Log.Warnf("Failed to read from SSHPubKey : %v", err)
+			return
+		}
+
+		// Append to the tmp/authorized_users file
```
romoh, line 278: Should we check if the user is `root` then update `PermitRootLogin=yes`?

toolkit/tools/imagecustomizerapi/user.go
cwize1: Really? I can't find any logic that enforces this requirement.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1121,40 +1120,35 @@ func addGroups(installChroot *safechroot.Chroot, groups []configuration.Group) (
...
+		isRoot, err = createUserWithPassword(installChroot, user)
 		if err != nil {
 			return
 		}
 		if isRoot {
 			rootUserAdded = true
 		}
 
-		err = configureUserGroupMembership(installChroot, user)
+		err = ConfigureUserGroupMembership(installChroot, user.Name, user.PrimaryGroup, user.SecondaryGroups)
```
dmcilvaney, line 1141: I'm assuming the changes to the parameters is to support both the new and old `user` here? Ideally I'd like us to move more to passing structs around (just look at some of the imager tool... yuck) but I get that makes life difficult here.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1273,23 +1234,21 @@ func createUserWithPassword(installChroot *safechroot.Chroot, user configuration
 			return
 		}
 
-		err = installChroot.UnsafeRun(func() error {
-			return chage(user.PasswordExpiresDays, user.Name)
-		})
+		err = Chage(user.PasswordExpiresDays, user.Name)
```
dmcilvaney, line 1237: Will this still correctly find the correct /etc/shadow for offline builds? I'd think it would get the setupChroot's copy without the nested chroot call? Do we need to pass `installChrootShadowFile` instead of getting it via `shadow, err = file.ReadLines(ShadowFile)` in `Chage()`?

cwize1: Good catch.

toolkit/tools/imagegen/installutils/installutils.go
cwize1: Yep. I figure params are easier than having yet another "user" type.

toolkit/tools/imagegen/installutils/installutils.go
cwize1: Fixed.

toolkit/tools/internal/userutils/userutils.go
cwize1: The image builder toolkit doesn't do this.

dmcilvaney (approved):

toolkit/tools/imagecustomizerapi/user.go
romoh: It may be related to https://github.com/microsoft/CBL-Mariner/blob/ff133d6c92b57b88edeb2789f47f3daefceb0f89/toolkit/imageconfigs/additionalconfigs/cloud-init.cfg#L79 but I guess it is fair to not account for that since it is not a toolkit default user.

romoh (approved):

PR #6343, opened 2023-10-04
feat: Introduce kernel-mshv's custom menuentry for the grub2-mkconfig flow

Camelron:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Introduce kernel-mshv's custom menuentry for the grub2-mkconfig flow

###### Change Log  

- Introduce kernel-mshv's custom menuentry for the grub2-mkconfig flow

NO

###### Test Methodology

- Local build, with other related mkconfig changes cherry-picked into cameronbaird/3.0-mkconfig-integration
- See that images containing kernel-mshv don't boot to the mshv kernel until the 50_mariner_mshv.cfg detects all proprietary components are present.


christopherco (changes requested):
Requesting to retarget to `3.0-dev`

Camelron:
Rebased to 3.0-dev

Camelron:
This ended up a duplicate PR. Closing.

dmcilvaney (approved):

ddstreetmicrosoft (changes requested):
i think my comments in #6391 apply here as well

SPECS/kernel-mshv/kernel-mshv.spec
```diff
@@ -120,7 +121,15 @@ make INSTALL_MOD_PATH=%{buildroot} modules_install
 # This configuration contains additional boot parameters required in our
 # Linux-Dom0-based images. 
 mkdir -p %{buildroot}%{_sysconfdir}/default/grub.d
-install -m 750 %{SOURCE3} %{buildroot}%{_sysconfdir}/default/grub.d/50_mariner_mshv.cfg
+mkdir -p %{buildroot}%{_sysconfdir}/grub.d
+
+# Initially install this configuration script WITHOUT execution bits.
```
Camelron, line 126: @ddstreetmicrosoft I understand your comment about using the execution bits to control which bash scripts are sourced is fundamentally wrong practice.

Given that, what is a good way to enforce this behavior: install the /etc/default/grub.d configuration for dom0 but leave it in a dormant state.

We could simply not worry about this, assuming that any system which installs kernel-mshv has the proprietary components (lxhvloader, hvix, etc). The downside to that is that anyone who accidentally installs kernel-mshv will be implicitly using the kernel-mshv kernel.

Camelron: Note that the dom0 menuentry still boots successfully without the proprietary components. Grub will be confused for an extra few seconds during startup however and we will boot into the kernel-mshv version of the kernel.

dmcilvaney: Could the script check if it is "valid" on its own? What information do we have to decide if we should  use dom0 or not?

SPECS/kernel-mshv/kernel-mshv.spec
```diff
@@ -120,7 +121,10 @@ make INSTALL_MOD_PATH=%{buildroot} modules_install
 # This configuration contains additional boot parameters required in our
 # Linux-Dom0-based images. 
 mkdir -p %{buildroot}%{_sysconfdir}/default/grub.d
-install -m 750 %{SOURCE3} %{buildroot}%{_sysconfdir}/default/grub.d/50_mariner_mshv.cfg
+mkdir -p %{buildroot}%{_sysconfdir}/grub.d
```
Camelron, line 124: Need to snip these out

christopherco (changes requested):

SPECS/kernel-mshv/50_mariner_mshv.cfg
```diff
@@ -1,5 +1,15 @@
-#!/bin/bash
-        
+requirements="Windows/System32/hvix64.exe Windows/System32/hvax64.exe lxhvloader.dll HvLoader.efi"
+for file in $requirements; do
+  if [ ! -f /boot/efi/$file ]; then
+    echo "Proprietary mshv component $file does not exist in /boot/efi"
+    echo "Not booting kernel-mshv..."
+    exit 0
```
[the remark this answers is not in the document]
Camelron: There is currently no error handling in our /etc/default/grub script (the thing that sources this script). I think I'll add some handling to at least log a warning. 

I think what we want to do here is exit early from 50_mariner_mshv.cfg and continue sourcing the next cfgs. If we ensure that the %grub2_post macro is included in packages that provide the needed proprietary bits, the last one will cause a call to mkconfig which succeeds in this check.

SPECS/initramfs/initramfs.spec
```diff
@@ -1,12 +1,13 @@
 Summary:        initramfs
 Name:           initramfs
 Version:        2.0
-Release:        13%{?dist}
+Release:        14%{?dist}
```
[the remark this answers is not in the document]
Camelron: Sure.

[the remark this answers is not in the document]
Camelron: https://github.com/microsoft/CBL-Mariner/pull/6943

SPECS/kernel-mshv/50_mariner_mshv.cfg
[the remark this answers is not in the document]
Camelron: See latest changes, we now create a comment documenting this decision in the produced grub.cfg, as well as logging a warning to the user during the mkconfig call. 

We also return 1 to /etc/default/grub, which logs another warning to users specifying the scriptlet (50_mariner_mshv.cfg in this case) which failed.

SPECS/kernel-mshv/50_mariner_mshv.cfg
```diff
@@ -1,5 +1,15 @@
-#!/bin/bash
-        
+requirements="Windows/System32/hvix64.exe Windows/System32/hvax64.exe lxhvloader.dll HvLoader.efi"
+for file in $requirements; do
+  if [ ! -f /boot/efi/$file ]; then
+    echo "Proprietary mshv component $file does not exist in /boot/efi"
```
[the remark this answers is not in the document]
Camelron: See latest wording.

Camelron:
Tested a manually tweaked core-efi image with kernel-mshv and kernel installed side-by-side. The tools call grub2-mkconfig during image build due to EnableGrubMkconfig=true.
- Resulting image boots the standard kernel, inspection of /boot/grub2/grub.cfg shows that the tool-called mkconfig noticed hvix64.exe missing, avoided executing the rest of 50_mariner_mshv.cfg
- Next, I installed mshv, mshv-bootloader-lx, hvloader and called grub2-mkconfig again. After rebooting, we are using the Dom0 menuentry.

SPECS/kernel-mshv/50_mariner_mshv.cfg
```diff
@@ -1,9 +1,10 @@
 requirements="Windows/System32/hvix64.exe Windows/System32/hvax64.exe lxhvloader.dll HvLoader.efi"
 for file in $requirements; do
   if [ ! -f /boot/efi/$file ]; then
-    echo "Proprietary mshv component $file does not exist in /boot/efi"
-    echo "Not booting kernel-mshv..."
-    exit 0
+    warning_msg="MSHV component ${file} does not exist in /boot/efi. Not enabling kernel-mshv menuentry."
+    echo "# ${warning_msg}"
+    grub_warn "${warning_msg}"
+    return 1
```
Camelron, line 7: This was a bug; exit causes the entire "call stack" of bash scripts to exit; we wanted return.

SPECS/kernel-mshv/50_mariner_mshv.cfg
```diff
@@ -1,9 +1,10 @@
 requirements="Windows/System32/hvix64.exe Windows/System32/hvax64.exe lxhvloader.dll HvLoader.efi"
 for file in $requirements; do
   if [ ! -f /boot/efi/$file ]; then
-    echo "Proprietary mshv component $file does not exist in /boot/efi"
-    echo "Not booting kernel-mshv..."
-    exit 0
+    warning_msg="MSHV component ${file} does not exist in /boot/efi. Not enabling kernel-mshv menuentry."
+    echo "# ${warning_msg}"
```
Camelron, line 5: stdout gets printed into the generated grub.cfg, hence comment.

SPECS/kernel-mshv/50_mariner_mshv.cfg
```diff
@@ -1,9 +1,10 @@
...
+    grub_warn "${warning_msg}"
```
Camelron, line 6: grub_warn gets shown to the user calling grub2-mkconfig.

toolkit/tools/internal/resources/assets/grub2/grub
```diff
@@ -17,6 +17,6 @@ GRUB_CMDLINE_LINUX_DEFAULT="{{.ExtraCommandLine}} $kernelopts"
 # =============================notice===============================
 for x in /etc/default/grub.d/*.cfg ; do
     if [ -e "${x}" ]; then
-        . "${x}"
+        . "${x}" || grub_warn "Received non-zero exit code from ${x}."
```
ddstreetmicrosoft, line 20: do we really want to ignore errors here? I mean technically this should never happen, these *.cfg files should always only contain env var settings, so if there is anything else that actually fails it probably should let the error flow up, no?

With that said I don't have any particular problem with this since, as mentioned, the *.cfg files shouldn't ever contain anything other than setting env vars. I'll leave it up to you.

ddstreetmicrosoft (approved):
I hesitate to provide much feedback for this highly specialized mshv setup; it seems like it LGTM (notwithstanding my comment about ignoring .../grub.d/*.cfg sourcing errors), but I'm relying a lot on your familiarity with the mshv specifics which I don't have. Probably would be good to get someone else's review as well.

toolkit/tools/internal/resources/assets/grub2/grub
Camelron: This is the edge case where we have kernel-mshv installed but no mshv. Normally we would enforce this via the package manager (kernel-mshv Requires: mshv...). mshv is a proprietary package of close-source hypervisor binaries however; we can't have this package depend on it directly.  

Instead, I have the 50_mariner_mshv.cfg determine whether or not the mshv components are installed at mkconfig-time and either set GRUB_DEFAULT=dom0 or return 1.

SPECS/kernel-mshv/50_mariner_mshv_menuentry
```diff
@@ -0,0 +1,66 @@
...
+else
+    if [ -z "${prepare_boot_cache}" ]; then
+        prepare_boot_cache="$(prepare_grub_to_access_device ${GRUB_DEVICE_BOOT} | grub_add_tab)"
+    fi
+    printf '%s\n' "${prepare_boot_cache}" | sed "s/^/$submenu_indentation/"
+fi
+
+printf '\t%s\n' "linux ${rel_dirname}/${basename} root=${linux_root_device_thisversion} ro ${GRUB_CMDLINE_LINUX} ${GRUB_CMDLINE_LINUX_DEFAULT}"
+printf '\t%s\n' "initrd ${rel_dirname}/${initrd_basename}"
+printf '}\n'
```
dmcilvaney, line 66:
```suggestion
printf '}\n'

```

SPECS/kernel-mshv/kernel-mshv.signatures.json
```diff
@@ -1,8 +1,9 @@
...
  "Signatures": {
-  "50_mariner_mshv.cfg": "0a5fcad1efb1fd37f910f675c5303210a2aeeef9e089d804510ce40ff9b26369",
+  "50_mariner_mshv.cfg": "12bf23f8857f893549933a20062ad2a69c3c654b87893e8dd3ce42f9329801c7",
+  "50_mariner_mshv_menuentry": "aebb8a39f7f34563c52403e2666ceab4d487b34a39b2fbd4d7affb48b4f90d0d",
   "cbl-mariner-ca-20211013.pem": "5ef124b0924cb1047c111a0ecff1ae11e6ad7cac8d1d9b40f98f99334121f0b0",
   "config": "2227744a0089cceae344dec57d6ff0540e265056f3831e24805d9cad4571e825",
   "kernel-mshv-5.15.126.mshv3.tar.gz": "a4a19caadbcb6c367bbc8d92338bbf6843f5e0fbd411f0ff0ba7650d44505e87"
  }
-}
+}
```
dmcilvaney, line 9:
```suggestion
}

```

SPECS/kernel-mshv/50_mariner_mshv_menuentry
```diff
@@ -0,0 +1,66 @@
...
+EOF
+
+# btrfs may reside on multiple devices. We cannot pass them as value of root= parameter
+# and mounting btrfs requires user space scanning, so force UUID in this case.
+if ( [ "x${GRUB_DEVICE_UUID}" = "x" ] && [ "x${GRUB_DEVICE_PARTUUID}" = "x" ] ) \
+    || ( [ "x${GRUB_DISABLE_LINUX_UUID}" = "xtrue" ] \
+	&& [ "x${GRUB_DISABLE_LINUX_PARTUUID}" = "xtrue" ] ) \
+    || ( ! test -e "/dev/disk/by-uuid/${GRUB_DEVICE_UUID}" \
+	&& ! test -e "/dev/disk/by-partuuid/${GRUB_DEVICE_PARTUUID}" ) \
+    || ( test -e "${GRUB_DEVICE}" && uses_abstraction "${GRUB_DEVICE}" lvm ); then
```
dmcilvaney, line 20: out of curiosity, how does this interact with verity images?

christopherco (approved):

PR #6344, opened 2023-10-04
Image Customizer: Add/remove packages

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add the ability to add and remove packages to the image customizer.

###### Change Log  

- Image Customizer: Add/remove packages

NO

###### Test Methodology

- Ran toolkit tools UTs.
- Add toolkit tools UTs.

### DIFF
--- a/toolkit/tools/imagecustomizer/main.go
+++ b/toolkit/tools/imagecustomizer/main.go
@@ -18,15 +18,17 @@ import (
 var (
 	app = kingpin.New("imagecustomizer", "Customizes a pre-built CBL-Mariner image")
 
-	buildDir          = app.Flag("build-dir", "Directory to run build out of.").Required().String()
-	imageFile         = app.Flag("image-file", "Path of the base CBL-Mariner image which the customization will be applied to.").Required().String()
-	outputImageFile   = app.Flag("output-image-file", "Path to write the customized image to.").Required().String()
-	outputImageFormat = app.Flag("output-image-format", "Format of output image. Supported: vhd, vhdx, qcow2, raw.").Required().Enum("vhd", "vhdx", "qcow2", "raw")
-	configFile        = app.Flag("config-file", "Path of the image customization config file.").Required().String()
-	logFile           = exe.LogFileFlag(app)
-	logLevel          = exe.LogLevelFlag(app)
-	profFlags         = exe.SetupProfileFlags(app)
-	timestampFile     = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
+	buildDir                 = app.Flag("build-dir", "Directory to run build out of.").Required().String()
+	imageFile                = app.Flag("image-file", "Path of the base CBL-Mariner image which the customization will be applied to.").Required().String()
+	outputImageFile          = app.Flag("output-image-file", "Path to write the customized image to.").Required().String()
+	outputImageFormat        = app.Flag("output-image-format", "Format of output image. Supported: vhd, vhdx, qcow2, raw.").Required().Enum("vhd", "vhdx", "qcow2", "raw")
+	configFile               = app.Flag("config-file", "Path of the image customization config file.").Required().String()
+	rpmSources               = app.Flag("rpm-source", "Path to a RPMs tarball or a directory containing RPMs.").Strings()
+	disableBaseImageRpmRepos = app.Flag("disable-base-image-rpm-repos", "Disable the base image's RPM repos as an RPM source").Bool()
+	logFile                  = exe.LogFileFlag(app)
+	logLevel                 = exe.LogLevelFlag(app)
+	profFlags                = exe.SetupProfileFlags(app)
+	timestampFile            = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
 )
 
 func main() {
@@ -55,7 +57,7 @@ func customizeImage() error {
 	var err error
 
 	err = imagecustomizerlib.CustomizeImageWithConfigFile(*buildDir, *configFile, *imageFile,
-		*outputImageFile, *outputImageFormat)
+		*rpmSources, *outputImageFile, *outputImageFormat, !*disableBaseImageRpmRepos)
 	if err != nil {
 		return err
 	}
--- a/toolkit/tools/imagecustomizer/main.go
+++ b/toolkit/tools/imagecustomizer/main.go
@@ -23,7 +23,7 @@ var (
 	outputImageFile          = app.Flag("output-image-file", "Path to write the customized image to.").Required().String()
 	outputImageFormat        = app.Flag("output-image-format", "Format of output image. Supported: vhd, vhdx, qcow2, raw.").Required().Enum("vhd", "vhdx", "qcow2", "raw")
 	configFile               = app.Flag("config-file", "Path of the image customization config file.").Required().String()
-	rpmSources               = app.Flag("rpm-source", "Path to a RPMs tarball or a directory containing RPMs.").Strings()
+	rpmSources               = app.Flag("rpm-source", "Path to a RPM repo config file or a directory containing RPMs.").Strings()
 	disableBaseImageRpmRepos = app.Flag("disable-base-image-rpm-repos", "Disable the base image's RPM repos as an RPM source").Bool()
 	logFile                  = exe.LogFileFlag(app)
 	logLevel                 = exe.LogLevelFlag(app)
--- a/toolkit/tools/imagecustomizerapi/packagelist.go
+++ b/toolkit/tools/imagecustomizerapi/packagelist.go
@@ -1,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+type PackageList struct {
+	Packages []string `yaml:"Packages"`
+}
+
+func (s *PackageList) IsValid() error {
+	return nil
+}
--- a/toolkit/tools/imagecustomizerapi/systemconfig.go
+++ b/toolkit/tools/imagecustomizerapi/systemconfig.go
@@ -14,12 +14,12 @@ import (
 type SystemConfig struct {
 	Hostname                string                    `yaml:"Hostname"`
 	UpdateBaseImagePackages bool                      `yaml:"UpdateBaseImagePackages"`
-	PackagesToAddLists      []string                  `yaml:"PackagesToAddLists"`
-	PackagesToAdd           []string                  `yaml:"PackagesToAdd"`
-	PackagesToRemoveLists   []string                  `yaml:"PackagesToRemoveLists"`
-	PackagesToRemove        []string                  `yaml:"PackagesToRemove"`
-	PackagesToUpdateLists   []string                  `yaml:"PackagesToUpdateLists"`
-	PackagesToUpdate        []string                  `yaml:"PackagesToUpdate"`
+	PackageListsInstall     []string                  `yaml:"PackageListsInstall"`
+	PackagesInstall         []string                  `yaml:"PackagesInstall"`
+	PackageListsRemove      []string                  `yaml:"PackageListsRemove"`
+	PackagesRemove          []string                  `yaml:"PackagesRemove"`
+	PackageListsUpdate      []string                  `yaml:"PackageListsUpdate"`
+	PackagesUpdate          []string                  `yaml:"PackagesUpdate"`
 	AdditionalFiles         map[string]FileConfigList `yaml:"AdditionalFiles"`
 	PostInstallScripts      []Script                  `yaml:"PostInstallScripts"`
 	FinalizeImageScripts    []Script                  `yaml:"FinalizeImageScripts"`
--- a/toolkit/tools/imagecustomizerapi/systemconfig.go
+++ b/toolkit/tools/imagecustomizerapi/systemconfig.go
@@ -12,15 +12,18 @@ import (
 
 // SystemConfig defines how each system present on the image is supposed to be configured.
 type SystemConfig struct {
-	Hostname              string                    `yaml:"Hostname"`
-	PackagesToAddLists    []string                  `yaml:"PackagesToAddLists"`
-	PackagesToAdd         []string                  `yaml:"PackagesToAdd"`
-	PackagesToRemoveLists []string                  `yaml:"PackagesToRemoveLists"`
-	PackagesToRemove      []string                  `yaml:"PackagesToRemove"`
-	AdditionalFiles       map[string]FileConfigList `yaml:"AdditionalFiles"`
-	PostInstallScripts    []Script                  `yaml:"PostInstallScripts"`
-	FinalizeImageScripts  []Script                  `yaml:"FinalizeImageScripts"`
-	Users                 []User                    `yaml:"Users"`
+	Hostname                string                    `yaml:"Hostname"`
+	UpdateBaseImagePackages bool                      `yaml:"UpdateBaseImagePackages"`
+	PackagesToAddLists      []string                  `yaml:"PackagesToAddLists"`
+	PackagesToAdd           []string                  `yaml:"PackagesToAdd"`
+	PackagesToRemoveLists   []string                  `yaml:"PackagesToRemoveLists"`
+	PackagesToRemove        []string                  `yaml:"PackagesToRemove"`
+	PackagesToUpdateLists   []string                  `yaml:"PackagesToUpdateLists"`
+	PackagesToUpdate        []string                  `yaml:"PackagesToUpdate"`
+	AdditionalFiles         map[string]FileConfigList `yaml:"AdditionalFiles"`
+	PostInstallScripts      []Script                  `yaml:"PostInstallScripts"`
+	FinalizeImageScripts    []Script                  `yaml:"FinalizeImageScripts"`
+	Users                   []User                    `yaml:"Users"`
 }
 
 func (s *SystemConfig) IsValid() error {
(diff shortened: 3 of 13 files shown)


toolkit/tools/imagecustomizerapi/systemconfig.go
```diff
@@ -12,10 +12,14 @@ import (
 
 // SystemConfig defines how each system present on the image is supposed to be configured.
 type SystemConfig struct {
-	Hostname             string                    `yaml:"Hostname"`
-	AdditionalFiles      map[string]FileConfigList `yaml:"AdditionalFiles"`
-	PostInstallScripts   []Script                  `yaml:"PostInstallScripts"`
-	FinalizeImageScripts []Script                  `yaml:"FinalizeImageScripts"`
+	Hostname              string                    `yaml:"Hostname"`
+	PackagesToAddLists    []string                  `yaml:"PackagesToAddLists"`
```
romoh, line 16: Should that be PackageListsToAdd instead?

cwize1: Probably not. It is not "package lists" that are being added to the image but rather "packages".

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -0,0 +1,178 @@
...
+	for _, packageName := range allPackagesToRemove {
+		tnfRemoveArgs[len(tnfRemoveArgs)-1] = packageName
+
+		err = imageChroot.Run(func() error {
+			err := shell.ExecuteLiveWithCallback(tdnfRemoveStdoutFilter, logger.Log.Warn, false, "tdnf",
+				tnfRemoveArgs...)
+			return err
+		})
+		if err != nil {
+			return fmt.Errorf("failed to install package (%s):\n%w", packageName, err)
```
romoh, line 87: uninstall/remove

toolkit/tools/imagecustomizerapi/packagelist.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
```
dmcilvaney, line 4: The existing tools support two inputs, `PackageLists` (path to .txt files) and `Packages` (list of package names), thoughts on continuing to support both here?

toolkit/tools/imagecustomizerapi/systemconfig.go
```diff
@@ -12,10 +12,14 @@ import (
...
+	PackagesToAdd         []string                  `yaml:"PackagesToAdd"`
```
dmcilvaney, line 17: Is this meant to be `[]string`, or use the new `PackageList` type?

toolkit/tools/pkg/imagecustomizerlib/rpmsourcesmounts.go
```diff
@@ -0,0 +1,338 @@
...
+	case ".conf":
+		return "conf", nil
+
+	default:
+		return "", nil
+	}
+}
+
+// Add a local directory containing RPMs to the allrepos.repo file.
+func appendLocalRepo(iniFile *ini.File, mountTargetDirectoryInChroot string) error {
```
romoh, line 296: side question - are we planning on supporting external repos through an API?

toolkit/tools/imagecustomizer/main.go
```diff
@@ -18,15 +18,17 @@ import (
...
-	logFile           = exe.LogFileFlag(app)
-	logLevel          = exe.LogLevelFlag(app)
-	profFlags         = exe.SetupProfileFlags(app)
-	timestampFile     = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
+	buildDir                 = app.Flag("build-dir", "Directory to run build out of.").Required().String()
+	imageFile                = app.Flag("image-file", "Path of the base CBL-Mariner image which the customization will be applied to.").Required().String()
+	outputImageFile          = app.Flag("output-image-file", "Path to write the customized image to.").Required().String()
+	outputImageFormat        = app.Flag("output-image-format", "Format of output image. Supported: vhd, vhdx, qcow2, raw.").Required().Enum("vhd", "vhdx", "qcow2", "raw")
+	configFile               = app.Flag("config-file", "Path of the image customization config file.").Required().String()
+	rpmSources               = app.Flag("rpm-source", "Path to a RPMs tarball or a directory containing RPMs.").Strings()
```
dmcilvaney, line 26: Our other tools support `--repo-file` to allow a set of custom repos to be passed (via `REPO_LIST="a.repo b.repo c.repo"`). Any interest in doing the same here? (The caveat being they rely on `tdnf` to install them. Maybe we can use the existing image fetcher tool to acquire the desired packages?)

toolkit/tools/imagecustomizer/main.go
```diff
@@ -18,15 +18,17 @@ import (
...
+	disableBaseImageRpmRepos = app.Flag("disable-base-image-rpm-repos", "Disable the base image's RPM repos as an RPM source").Bool()
```
dmcilvaney, line 27: Is this different from the existing `--disable-default-repos` and `--disable-upstream-repos`? Might want to align on a name for these if its doing the same thing.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -0,0 +1,178 @@
+// Copyright (c) Microsoft Corporation.
```
dmcilvaney, line 1: There is a bit of overlap here with `TdnfInstall()` and `TdnfInstallWithProgress()`. They aren't really the same, but just wanted to make sure there wasn't anything we can reuse.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -0,0 +1,178 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func collectPackagesList(baseConfigPath string, packageLists []string, packages []string) ([]string, error) {
+	var err error
+
+	// Read in the packages from the package list files.
```
dmcilvaney, line 49: In our main tools, we also consider the kernel package as a special case. Does there need to be any special handling for the kernel for post-customization? (I'm not actually sure there is a reason for special handling in the main tools anymore tbh).

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -0,0 +1,178 @@
...
+
+	allPackages = append(allPackages, packages...)
+	return allPackages, nil
+}
+
+func removePackages(buildDir string, allPackagesToRemove []string, imageChroot *safechroot.Chroot) error {
+	var err error
+
+	tnfRemoveArgs := []string{
+		"-v", "remove", "--assumeyes", "--disablerepo", "*",
```
dmcilvaney, line 71: Maybe of interest is `autoremove`. It will pull any dependencies as well as the main package.
Warning: I've seen it pull packages unexpectedly sometimes that shouldn't get yanked... like `rpm` and `tdnf` itself...

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to install package (%s):\n%w", packageName, err)
+		}
+	}
+
+	return nil
+}
+
+// Process the stdout of a `tdnf install -v` call and send the list of installed packages to the debug log.
+func tdnfRemoveStdoutFilter(args ...interface{}) {
```
dmcilvaney, line 95: I don't have a strong opinion on this one, but we generally define the callbacks just before we call `ExecuteLiveWithCallback()` in the same function. 

I don't care either way though.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -46,9 +62,53 @@ func doCustomizations(baseConfigPath string, config *imagecustomizerapi.SystemCo
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Override the resolv.conf file, so that in-chroot processes can access the network.
+// For example, to install packages from packages.microsoft.com.
+func overrideResolvConf(imageChroot *safechroot.Chroot) error {
```
dmcilvaney, line 75: We have generally used `SafeChroot.AddFiles()` to add the network configs, but this works fine as well. (see RpmRepoCloner.addNetworkFiles()`

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -0,0 +1,178 @@
...
+	"path"
+	"strings"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
+)
+
+func updatePackages(buildDir string, baseConfigPath string, packagesToAddLists []string, packagesToAdd []string,
```
dmcilvaney, line 17: Should we have an actual (maybe optional) "Update" function here that pulls the latest copies of packages?

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return err
+	}
+	defer mounts.close()
+
+	// Create tdnf command args.
+	// Note: When using `--repofromdir`, tdnf will not use any default repos and will only use the last
+	// `--repofromdir` specified.
+	tnfInstallArgs := []string{
+		"-v", "install", "--nogpgcheck", "--assumeyes",
```
dmcilvaney, line 134: Do we want to support replacing an already installed package? Not sure exactly what that would look like, but say the user has a set of customized versions locally they want to use, I expect tdnf will just skip over them (or if they are lower version will refuse to downgrade).

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -0,0 +1,178 @@
...
+		"--setopt", fmt.Sprintf("reposdir=%s", rpmsMountParentDirInChroot),
+		// Placeholder for package name.
+		"",
+	}
+
+	// Install packages.
+	// Do this one at a time, to avoid running out of memory.
```
dmcilvaney, line 141: Did you see this actually EOM? I know old `tdnf` had issues with this, but I've been curious if the more recent versions are less problematic.

toolkit/tools/pkg/imagecustomizerlib/rpmsourcesmounts.go
```diff
@@ -0,0 +1,338 @@
...
+	allReposConfigFilePath    string
+}
+
+func mountRpmSources(buildDir string, imageChroot *safechroot.Chroot, rpmsSources []string,
+	useBaseImageRpmRepos bool,
+) (*rpmSourcesMounts, error) {
+	var err error
+
+	var mounts rpmSourcesMounts
+	err = mounts.mountRpmSourcesHelper(buildDir, imageChroot, rpmsSources, useBaseImageRpmRepos)
```
dmcilvaney, line 42: Any reason to not integrate this with the `safechroot.MountPoint` type? afaik that will handle cleanup of the mounts etc.

toolkit/tools/imagecustomizerapi/systemconfig.go
cwize1: PackageLists contains file paths to a separate YAML file that contains the `PackageList` type.

toolkit/tools/pkg/imagecustomizerlib/rpmsourcesmounts.go
cwize1: If you pass in an RPM repo config file as an RPM source, you can point that to an external repo.

toolkit/tools/imagecustomizer/main.go
cwize1: This flag can take any of the following:

- An RPM repo config file.
- A directory containing RPMs files.

So, it includes the `--repo-file` functionality. But is also more generic.

The image customizer tool is using `tdnf` to do the installation.

What would be the advantage of using the fetcher instead of just using `tdnf` directly?

toolkit/tools/imagecustomizer/main.go
cwize1: Yes and no. For image customizer, the default repo files are coming from the image file itself, not the toolkit.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: Nope. I'm just following the existing code.

toolkit/tools/pkg/imagecustomizerlib/rpmsourcesmounts.go
cwize1: The cleanup of logic of `safechroot.MountPoint` lives in the `safechroot.Chroot` type. And the chroot is created well before the point where we are looking to mount the RPMs directories.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: You raise an interesting point.

For package replacement the user can specify the package in both the `PackagesToRemove` and `PackagesToAdd`.

Package update is less straightforward. Perhaps add a `PackagesToUpdate` config property?

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: Do you know what the original reason for the special handling?

That being said, I do have another task to deal with the kernel package installation/replacement.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: Hmm, I think `autoremove` could result in unexpected behavior for users since the packages that get removed could change when the package dependencies are updated. So, I think sticking with `remove` will be better. (Though perhaps we can mention in the docs that `autoremove` can be used to find additional packages that can removed.)

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: I've added support for package updates.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: Done.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: Ended up reusing one of the functions for package updates anyway.

toolkit/tools/imagecustomizerapi/systemconfig.go
cwize1: Hmm. Daniel was also confused by the name. So, these will probably need to change.

cwize1: Change the names.

toolkit/tools/imagecustomizerapi/systemconfig.go
cwize1: Changed the names.

romoh (commented):
Should we add any validation to the repo config provided?

romoh (approved):

cwize1:
> Should we add any validation to the repo config provided?

I have at least ensured that they are valid INI files. Beyond that, I am not sure what validation should be done.

toolkit/tools/imagecustomizer/main.go
dmcilvaney: Major benefit is you don't need to use/modify the real image to deal with networking, but that isn't a huge deal.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
dmcilvaney: No, I think that is probably lost to the mists of time. I'll ask around and see if anyone recalls.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
dmcilvaney: At your own risk, but this may be unnecessary and doing everything in a single transaction will generally work better if there are conflicts.

dmcilvaney (approved):

toolkit/tools/imagecustomizer/main.go
cwize1: I figure, giving the custom scripts access to the network in case the users wants/needs it, is probably goodness anyway.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: Yeah, I needed some subtle changes for image customization. Hence, the code duplication.

cwize1:
> > > Should we add any validation to the repo config provided?
> > 
> > 
> > I have at least ensured that they are valid INI files. Beyond that, I am not sure what validation should be done.
> 
> A trivial one that still trips me up on occasion is that repo files MUST end with `.repo` or tdnf etc. will ignore them.

Oh, yeah. I should be checking for the `.repo` extension, not `.conf`.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: Leaving this for now. But I've opened a task to investigate further.

romoh (approved):

PR #6350, opened 2023-10-05
Upgrade python-gevent to 23.9.1 to fix CVE-2023-41419

0xba1a:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Upgrade python-gevent to 23.9.1 to fix CVE-2023-41419

###### Change Log  

- Upgrade python-gevent to 23.9.1 to fix CVE-2023-41419

**NO**

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2023-41419

###### Test Methodology

- Local build: successful
- Buddy Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=431566&view=results
- Full Build: 431567


SPECS/python-gevent/python-gevent.spec
```diff
@@ -1,15 +1,15 @@
 Summary:        Coroutine-based network library
 Name:           python-gevent
-Version:        1.3.6
-Release:        8%{?dist}
+Version:        23.9.1
+Release:        1%{?dist}
```
mfrw, line 4: Can we change the target branch to `main` instead of `2.0`

jslobodzian (changes requested):
I think you and Mandeep both worked on this at the same time.  This is a huge jump in version.  Mandeep has applied the patch but the ptest needs to be fixed.  The ptest has always been broken, so it needs to be fixed too.  If the patch works and the ptest passed, then we should take his change over this one.

0xba1a:
> I think you and Mandeep both worked on this at the same time. This is a huge jump in version. Mandeep has applied the patch but the ptest needs to be fixed. The ptest has always been broken, so it needs to be fixed too. If the patch works and the ptest passed, then we should take his change over this one.

It's fine. The patching approach is much better than a version bump.

PR #6365, opened 2023-10-07
Kata-CC: UVM - Enable encfs sidecar container

ms-mahuber:
These configuration changes enable the encfs [sidecar container](https://github.com/microsoft/confidential-sidecar-containers/tree/main/examples/encfs) use case on Kata-CC. The encfs sidecar r/w mounts an encrypted Azure blob storage as transparent device mapper target (LUKS) into the container filesystem (after secure key release and attestation). For doing this, relevant kernel support is required.

This PR enables the minimum set of relevant kernel options for kernel-uvm and kernel-uvm-cvm.


Redent0r (approved):
LGTM. Would wait for https://dev.azure.com/mariner-org/mariner/_build/results?buildId=438235&view=results to pass

Camelron (approved):

PR #6373, opened 2023-10-09
Add priority queuing to optimize implicit nodes

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Add a priority based build queue such that implicit provides are resolved without wasted work.

The goal is to avoid building packages that aren't required for a given goal. Currently if the subgraph rooted at our goal (say the list of all packages that need to be built for an image config) contains a dynamic dependency (looks like `something(name)` or `/path/to/file/`) the scheduler can't optimize the build.

The scheduler will attempt to take the full graph that contains all packages defined locally, and prune to just the required subgraph. It does this every time it gets a build result back, until the graph is successfully optimized. The optimization will fail if we have a dynamic/implicit dependency. We can't optimize since we don't konw what package will actually end up supplying that dependncy, they generally won't have an explicit `Provides: /path/to/file` in the `.spec` which would allow the graph to encode that information. Instead, we have to wait until a package build is complete to scan the resulting `.rpm` to see if it provides anything extra beyond what the `.spec` claims.

In the current version, if the graph cannot be optimized the scheduler will affectively a random walk through the graph, operating on each leaf node it finds.  When a package is built it is "removed" from the graph (actually just marked as done) and new "leaf" nodes may be exposed. Each time a package is built, the rpms are scanned and the graph updated to remove any implicit nodes that are now satisfied. Once all implicit nodes are removed from the subgraph under the goal, the graph is pruned of all the non-essential nodes and the build is optimized.

Instead, we should try to prioritize reaching the nodes that provide the implicit dependencies. As above, it isn't possible to know with 100% confidence that a given package will still provide a given implicit dependency, but there is some information that can provide a decent guess.

When the package fetcher runs to populate the graph, it assumes that each implicit dependency may not be found in the local packages and queries the repo for a provider (99% of the time, we get the same package back as we have locally +- some version differences). It places this package into the cache as a backup incase the local packages can't provide the dependency. The scheduler has a global mode that tries to build as much as possible without using these cached copies. Only if the scheduler gets completely stuck will it enable the cached implicit nodes.

We can use this information to make an educated guess which nodes will end up providing the implicit dependency. Say we have `my-pkg-1.1.1-1` published on PMC, and `my-pkg-1.1.1-2` as a `.spec` file locally. We have another package `goal-pkg-1-1` that has the line `BuildRequires: /path/to/my-pkg/some-file`.  If we look at the `%files` section of `my-pkg` we might see:
```spec
%files
/path/to/%{name}/*
```
Clearly we won't know what files are available in `/path/to/my-pkg` until after we build the `.rpm`... But if we look at the cache we will see an implicit remote node pointing to `my-pkg-1.1.1-1` from PMC. So, if we assume nothing has changed, we can look at the remote node and guess that since we can get that file from `my-pkg-1.1.1-1`, we can probably expect it from `1.1.1-2` as well.

To this end, we can have three levels of priority for the scheduler:
- High: put `my-pkg` here. Check it as fast as possible so we can see if the implicit dependnecy is resolved or not. If it is, we can optimize the graph.
- Medium: put `goal-pkg` here. We know anything else under this subgraph will be required for the end goal and we always want to build it.
- Low: Everything else. We ideally don't want to build anything here, but in the case where the implicit dependencies aren't found based on our guess we need to fall back to the old mechanism (random builds)

An added complexity: The scheduler will queue all packages into the build channels in a greedy fashion. We need a mechanism to ignore the low priority channels, or avoid queuing low priority packages, until such time as we get stuck and failed to optimize.

Some test results:
```
# Prep build to remove variability from network
sudo make clean-build-packages &&  sudo make graph-preprocessed input-srpms go-tools  PRECACHE=y CONFIG_FILE=./imageconfigs/core-efi.json
# Build images needed for core-efi.json
time sudo make build-packages PRECACHE=y CONFIG_FILE=./imageconfigs/core-efi.json 2>&1 | tee log.txt
```
| Mode | Wall Time | Total packages built | Wasted builds |
|--------|------------|-----------------------|-----------------|
| New    | 57m (1hr) | 109                          |0                       |
| Old     | 144m (2.3hr) | 477                      | 368                  |

**tldr: We wasted over an hour building packages we ended up not using in the old flow.**

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/toolkit/tools/scheduler/scheduler.go
+++ b/toolkit/tools/scheduler/scheduler.go
@@ -386,7 +386,7 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
 				allowLowPriorityNodes = true
 				nodesToBuild = schedulerutils.LeafNodes(pkgGraph, graphMutex, goalNode, buildState, useCachedImplicit)
 				// Check if any of the low priority nodes are still in the graph. If not, we can skip building them.
-				lowPriorityNodes = filterLowPriorityNodesAgainstGraph(pkgGraph, graphMutex, buildState, lowPriorityNodes)
+				lowPriorityNodes = filterNodesAgainstGraphState(pkgGraph, graphMutex, buildState, lowPriorityNodes)
 				nodesToBuild = append(nodesToBuild, lowPriorityNodes...)
 				// We should generate no more low priority nodes going forward.
 				lowPriorityNodes = nil
@@ -450,7 +450,7 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
 
 						// If we are shrinking the graph we need to filter out any low priority nodes that are no longer in the graph.
 						// We may still have some low priority nodes, those will be handled as normal when we switch allowLowPriorityNodes=true
-						lowPriorityNodes = filterLowPriorityNodesAgainstGraph(pkgGraph, graphMutex, buildState, lowPriorityNodes)
+						lowPriorityNodes = filterNodesAgainstGraphState(pkgGraph, graphMutex, buildState, lowPriorityNodes)
 					}
 				}
 
--- a/toolkit/tools/scheduler/scheduler.go
+++ b/toolkit/tools/scheduler/scheduler.go
@@ -364,7 +364,7 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
 			//    resolve implicit provides if the scheduler is unable to find a provider for them.
 			switch buildState.GetNodePriority(req.Node) {
 			case schedulerutils.HighNodePriority:
-				logger.Log.Tracef("Priority priority request (%d) for node '%s'", buildState.GetNodePriority(req.Node), req.Node.FriendlyName())
+				logger.Log.Tracef("High priority request (%d) for node '%s'", buildState.GetNodePriority(req.Node), req.Node.FriendlyName())
 				channels.HighPriorityRequests <- req
 			case schedulerutils.MediumNodePriority:
 				logger.Log.Tracef("Medium priority request (%d) for node '%s'", buildState.GetNodePriority(req.Node), req.Node.FriendlyName())
--- a/toolkit/tools/scheduler/scheduler.go
+++ b/toolkit/tools/scheduler/scheduler.go
@@ -43,11 +43,12 @@ var (
 // Unlike BuildChannels, schedulerChannels holds bidirectional channels that
 // only the top-level scheduler should have. BuildChannels contains directional channels.
 type schedulerChannels struct {
-	Requests         chan *schedulerutils.BuildRequest
-	PriorityRequests chan *schedulerutils.BuildRequest
-	Results          chan *schedulerutils.BuildResult
-	Cancel           chan struct{}
-	Done             chan struct{}
+	LowPriorityRequests    chan *schedulerutils.BuildRequest
+	MediumPriorityRequests chan *schedulerutils.BuildRequest
+	HighPriorityRequests   chan *schedulerutils.BuildRequest
+	Results                chan *schedulerutils.BuildResult
+	Cancel                 chan struct{}
+	Done                   chan struct{}
 }
 
 var (
@@ -273,20 +274,22 @@ func buildGraph(inputFile, outputFile string, agent buildagents.BuildAgent, work
 // channelBufferSize controls how many entries in the channels can be buffered before blocking writes to them.
 func startWorkerPool(agent buildagents.BuildAgent, workers, buildAttempts, checkAttempts, channelBufferSize int, graphMutex *sync.RWMutex, ignoredPackages, ignoredTests []*pkgjson.PackageVer) (channels *schedulerChannels) {
 	channels = &schedulerChannels{
-		Requests:         make(chan *schedulerutils.BuildRequest, channelBufferSize),
-		PriorityRequests: make(chan *schedulerutils.BuildRequest, channelBufferSize),
-		Results:          make(chan *schedulerutils.BuildResult, channelBufferSize),
-		Cancel:           make(chan struct{}),
-		Done:             make(chan struct{}),
+		LowPriorityRequests:    make(chan *schedulerutils.BuildRequest, channelBufferSize),
+		MediumPriorityRequests: make(chan *schedulerutils.BuildRequest, channelBufferSize),
+		HighPriorityRequests:   make(chan *schedulerutils.BuildRequest, channelBufferSize),
+		Results:                make(chan *schedulerutils.BuildResult, channelBufferSize),
+		Cancel:                 make(chan struct{}),
+		Done:                   make(chan struct{}),
 	}
 
 	// Downcast the bidirectional scheduler channels into directional channels for the build workers.
 	directionalChannels := &schedulerutils.BuildChannels{
-		Requests:         channels.Requests,
-		PriorityRequests: channels.PriorityRequests,
-		Results:          channels.Results,
-		Cancel:           channels.Cancel,
-		Done:             channels.Done,
+		LowPriorityRequests:    channels.LowPriorityRequests,
+		MediumPriorityRequests: channels.MediumPriorityRequests,
+		HighPriorityRequests:   channels.HighPriorityRequests,
+		Results:                channels.Results,
+		Cancel:                 channels.Cancel,
+		Done:                   channels.Done,
 	}
 
 	// Start the workers now so they begin working as soon as a new job is queued.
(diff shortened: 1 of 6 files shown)


toolkit/tools/scheduler/schedulerutils/priority.go
```diff
@@ -0,0 +1,240 @@
...
+	matches, err = findBuildNodesByGraph(implicitNode, pkgGraph, matchVersion)
+	if err != nil {
+		err = fmt.Errorf("failed to find best package node for implicit node (%s):\n%w", implicitNode.FriendlyName(), err)
+		return
+	}
+	if len(matches) > 0 {
+		return
+	}
+
+	matches = findBuildNodesByPath(implicitNode.RpmPath, pkgGraph, matchVersion)
```
Camelron, line 167: This part confuses me; I thought that an implicit node, by definition, does not map to a specific RPM, rpmpath, etc. 

An implicit node in my mind is /bin/sh, and we don't know which RPM(s) end up providing that file.

toolkit/tools/scheduler/schedulerutils/priority.go
```diff
@@ -0,0 +1,240 @@
...
+	search.Walk(pkgGraph, node, func(n graph.Node, d int) (stopSearch bool) {
+		pkgNode := n.(*pkggraph.PkgNode)
+		logger.Log.Warnf("Setting %s to priority %d", pkgNode.FriendlyName(), priority)
+		s.IncreaseNodePriority(pkgNode, priority)
+		return
+	})
+}
+
+// findMatchesForImplicit will try and find real packages that satisfies the implicit node. This is done through four methods:
+// 1. Find the best package node for the implicit node
```
Camelron, line 148: Just to make sure I grok this (please correct my understanding):

For a given implicit node (say, /bin/bash),
1. Look for an existing package (run node?) which provides said file 
2. if not found, search by rpmpath for  an exact match. How do we know an rpmpath for an implicit though? 
3. If both 1&2 fail, stop caring about finding a matching version for the provider nodes (the cached prebuilts?)

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -223,20 +231,38 @@ func buildResultsSetToNodesSet(statesSet map[string]*BuildResult) (result map[st
...
+			optimizedSrpmPaths[node.SrpmPath] = true
+		}
+	}
+
+	for _, node := range fullGraph.AllBuildNodes() {
+		// check if path was used in the optimized graph
+		nodeInOptimizedGraph := optimizedGraph.HasNode(node)
+		srpmUsed := optimizedSrpmPaths[node.SrpmPath]
+		if !nodeInOptimizedGraph {
+			if !srpmUsed && buildState.IsNodeAvailable(node) {
```
Camelron, line 261: Can a node be cached and available? In that case, we didn't really waste any time building it...

Is that why we check srpmUsed? Can we not instead check IsNodeCached(node) so that there is not some undocumented invariant (e.g. cached nodes did not mark their SrpmPath used)

toolkit/tools/scheduler/scheduler.go
```diff
@@ -323,57 +326,76 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
...
-		// If there are no active builds running or results waiting to check try enabling cached packages for unresolved
-		// dynamic dependencies to unblock more nodes. Otherwise, there is nothing left that can be built.
+		// If there are no active builds running or results waiting to check try enabling low priority nodes  to unblock
+		// more nodes, then cached packages for unresolved dynamic dependencies. Otherwise, there is nothing left that can be built.
 		if len(buildState.ActiveBuilds()) == 0 && len(channels.Results) == 0 {
-			if useCachedImplicit {
-				err = fmt.Errorf("could not build all packages")
-				break
-			} else {
+			if !allowLowPriorityNodes {
```
Camelron, line 381: We might want logic in the loop to disable allowLowPriorityNodes once we get the implicit node we wanted.

toolkit/tools/scheduler/schedulerutils/priority.go
dmcilvaney: At the end of the day every dependency does need to be provided by some rpm. The fetcher asks PMC for all the options that provide an implicit dependency and chooses one.

This path is based on that data we received from PMC when we ran the fetcher. It may or may not be correct, we don't know. But it's probably a decent guess.

toolkit/tools/scheduler/schedulerutils/priority.go
dmcilvaney: 2: See above, PMC will have given us a package when we ran the fetcher.

toolkit/tools/scheduler/schedulerutils/printresults.go
dmcilvaney: I'd consider the wasted build output a debugging tool, I'm probably not going to include it in the final PR. Left it in for now since its useful though to get a rough idea of improvements. My local builds I do for data will have zero cached nodes used.

toolkit/tools/scheduler/scheduler.go
dmcilvaney: I did have that originally, but we could end up with cases were we DID find a node, but it has run-time dependencies that we didn't plan on building before. Assuming we find all our implicits we will optimize the graph. That graph may include those low-priority nodes as runtime deps, so no reason to force that mode back off again. Once we hit the optimized stage we have high confidence we won't be wasting time regardless.

toolkit/tools/scheduler/scheduler.go
```diff
@@ -323,57 +326,76 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
...
-				fallthrough
-			case pkggraph.TypeLocalBuild:
+			// Queue nodes based on how likely they are to help optimize the build.
+			// - High priority nodes are expected to resolve implicit provides that are needed to build other packages.
+			// - Medium priority nodes are our desired build goal.
+			// - Low priority nodes are packages that are not needed to build the desired goal, but may be needed to
+			//    resolve implicit provides if the scheduler is unable to find a provider for them.
+			switch buildState.GetNodePriority(req.Node) {
+			case schedulerutils.HighNodePriority:
+				logger.Log.Tracef("Priority priority request (%d) for node '%s'", buildState.GetNodePriority(req.Node), req.Node.FriendlyName())
```
trungams, line 364:
```suggestion
				logger.Log.Tracef("High priority request (%d) for node '%s'", buildState.GetNodePriority(req.Node), req.Node.FriendlyName())
```

toolkit/tools/scheduler/scheduler.go
```diff
@@ -323,57 +326,76 @@ func buildAllNodes(stopOnFailure, canUseCache bool, packagesToRebuild, testsToRe
...
+				channels.HighPriorityRequests <- req
+			case schedulerutils.MediumNodePriority:
+				logger.Log.Tracef("Medium priority request (%d) for node '%s'", buildState.GetNodePriority(req.Node), req.Node.FriendlyName())
+				channels.MediumPriorityRequests <- req
+			case schedulerutils.LowNodePriority:
 				fallthrough
 			default:
```
trungams, line 371: what is the default case here, haven't all requests been assigned a priority at this point?

toolkit/tools/scheduler/schedulerutils/priority.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+}
+
+func (s NodePriorityMap) GetPriority(node *pkggraph.PkgNode) (priority int) {
+	return s.priorities[node]
+}
+
+// SetSubgraphPriority will increase the priority of the given node and all of its dependencies to the given priority.
+func (s NodePriorityMap) SetSubgraphPriority(pkgGraph *pkggraph.PkgGraph, node *pkggraph.PkgNode, priority int) {
+	search := traverse.BreadthFirst{}
```
trungams, line 136: this struct has a `Visit` field that can be used over the `until` param in `seach.Walk`, it's only preference but we can get rid of the `stopSearch` value that we don't ever use

toolkit/tools/scheduler/scheduler.go
dmcilvaney: Fair point, probably better to throw an error here instead. Will fix.

toolkit/tools/scheduler/schedulerutils/priority.go
dmcilvaney: Does read more cleanly to me as well, updated.

dmcilvaney:
This PR is now re-targeted at 3.0-dev.

Camelron (approved):

PR #6376, opened 2023-10-09
patch CVE-2022-1941 for protobuf

mbykhovtsev-ms:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

patches CVE-2022-1941

###### Change Log  

- Added a patch to fix CVE-2022-1941

**NO**

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2022-1941

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=433656&view=results


miz060 (approved):

SPECS/protobuf/protobuf.spec
```diff
@@ -1,6 +1,6 @@
 Summary:        Google's data interchange format
 Name:           protobuf
-Version:        3.17.3
+Version:        3.18.3
```
PawelWMS, line 3: We try to avoid version changes to something higher than the patch version (the third number in the version). This is to avoid breaking changes as much as possible. The changes in the `%files` section indicate that we'd need to re-compile all binaries, which built against the older version and that's tricky, since we don't know where all of these binaries might have built (we're aware about our packages, but not external projects).

With that in mind, can you try to port the CVE patch to the 3.17.X version?

PawelWMS (changes requested):
Upgrading the minor version from 17 to 18 is risky considering the changes to the versions of the shared libraries. We'd need to backport the CVE patch to 3.17.X.

SPECS/protobuf/protobuf.spec
mbykhovtsev-ms: I have looked into getting a patch at first. But unfortunately the only patch is to upgrade to certain *versions*, they did not specifically produce the patch for it as it is parsing and memory management issue that affects specific versions of protobuf

mbykhovtsev-ms:
> Upgrading the minor version from 17 to 18 is risky considering the changes to the versions of the shared libraries. We'd need to backport the CVE patch to 3.17.X.

I have managed to finally find the patch. It was buried deep and had no description in change explicitly stating that it fixes the CVE. However, I did find some other place noting the change as fix https://security-tracker.debian.org/tracker/CVE-2022-1941. So I have reverted the upgraded and added the patch, looks like it successfully builds on fast trak.

SPECS/protobuf/protobuf.spec
```diff
@@ -108,6 +109,9 @@ popd
 %{python3_sitelib}/*
 
 %changelog
+* Mon Oct 09 2023 Mykhailo Bykhovtsev <mbykhovtsev@microsft.com> - 3.18.3-1
```
PawelWMS, line 112:
```suggestion
* Mon Oct 09 2023 Mykhailo Bykhovtsev <mbykhovtsev@microsft.com> - 3.17.3-3
```

Please also update the `Release` tag.

PawelWMS (approved):

PawelWMS:
Looks good, let's see if this unblocks the PR check.

PawelWMS (changes requested):
The package builds (yay!) but the ptests for it started failing and that looks like a regression when compared to earlier releases - could you check what that is about?

PawelWMS (approved):

mbykhovtsev-ms:
Patch fails existing unit tests, risky to take it. The only solution otherwise is the upgrade minor version which also breaks other packages. Abandoning PR

PR #6379, opened 2023-10-10
fix: properly track updates to libcrypt

christopherco:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Issue discovered where glibc rpm installs and owns the libcrypt.so file
from rpm's perspective, however there is post transaction script logic
in libxcrypt that overwrites the libcrypt.so file with a symlink to
libxcrypt. Since glibc rpm still officially owns the libcrypt.so file,
on a glibc upgrade operation, the package upgrade will overwrite the
current libcrypt.so file with the updated glibc libcrypt.so file,
effectively destroying the symlink to libxcrypt.so if libxcrypt package
is installed.

The way we are addressing this is by first moving glibc's libcrypt file
into a new subpackage called glibc-libcrypt. This package will also
conflict against the libxcrypt package and will provide "libcrypt". Then
in libxcrypt package, we remove references to the problematic
override_glibc macro and symlink scriptlets, and instead allow the
package to naturally create and supply libcrypt.so. We place a conflicts
on glibc-libcrypt. Finally in the glibc package, we place a general
requires on any package which supplies "libcrypt".

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/46702977

###### Test Methodology

- Pipeline build id:


SPECS/glibc/glibc.spec
```diff
@@ -36,6 +36,7 @@ BuildRequires:  gettext
 BuildRequires:  kernel-headers
 BuildRequires:  texinfo
 Requires:       filesystem
+Requires:       pkgconfig(libcrypt)
```
ddstreet, line 39: I think for this specific situation, we probably don't need to use pkgconfig? Just Requires: libcrypt should be enough I think? That will require the 'virtual' libcrypt package that only glibc-libcrypt and libxcrypt provide (via the Provides: we're adding to them).

Also it looks like we can set a preference: https://docs.fedoraproject.org/en-US/packaging-guidelines/WeakDependencies/#_package_preference

So probably we want to also include here:
Suggests: glibc-libcrypt

SPECS/glibc/glibc.spec
```diff
@@ -257,6 +264,7 @@ grep "^FAIL: nptl/tst-eintr1" tests.sum >/dev/null && n=$((n+1)) ||:
 %config(noreplace) %{_sysconfdir}/rpc
 %config(missingok,noreplace) %{_sysconfdir}/ld.so.cache
 %config %{_sysconfdir}/locale-gen.conf
+%exclude /%{_lib64}/libcrypt.so.1
```
ddstreet, line 270: if the line below hardcodes /lib64 in the glob, this exclude should probably hardcode /lib64 as well right? Or, change the line below to /%{_lib64}, but I don't know if that's a good idea for a stable release change

SPECS/glibc/glibc.spec
```diff
@@ -298,6 +306,9 @@ grep "^FAIL: nptl/tst-eintr1" tests.sum >/dev/null && n=$((n+1)) ||:
 %{_sbindir}/nscd
 %dir %{_localstatedir}/cache/nscd
 
+%files libcrypt
+/%{_lib64}/libcrypt.so.1
```
ddstreet, line 313: and this should probably hardcode /lib64/ too as above?

SPECS/libxcrypt/libxcrypt.spec
```diff
@@ -46,9 +46,7 @@
 # Fail linking if there are undefined symbols.
 # Required for proper ELF symbol versioning support.
 %global _ld_strict_symbol_defs 1
-# override_glibc and glibcversion are temporary to make libxcrypt install on top of glibc
 %define glibcversion 2.35
```
ddstreet, line 49: should remove glibcversion define completely I think

SPECS/libxcrypt/libxcrypt.spec
```diff
@@ -138,14 +136,10 @@ Obsoletes:      %{name}-compat < %{version}-%{release}
...
 #Requires:       glibc%{?_isa}         >= %{glibc_minver}
-%if %{with override_glibc}
-# Require a specific glibc version so the post macro is compatible.
-BuildRequires:  glibc-devel = %{glibcversion}
-Requires:       glibc = %{glibcversion}
-%endif
 %if 0%{?fedora} >= 30
 Recommends:     mkpasswd
 %endif
+Conflicts:      glibc-libcrypt
```
ddstreet, line 142: probably should also add a Requires: glibc >= $VERSION so we make sure the libcrypt file this package installs doesn't conflict with the libcrypt file an older glibc installs (where VERSION is whatever glibc version first includes the glibc-libcrypt split)

SPECS/libxcrypt/libxcrypt.spec
```diff
@@ -408,13 +377,8 @@ ln -s %{_libdir}/libcrypt-%{glibcversion}.so %{_libdir}/libcrypt.so.1
 %files devel
 %doc %{_pkgdocdir}/ChangeLog
 %doc %{_pkgdocdir}/TODO
-%if %{with override_glibc}
-%exclude %{_libdir}/libcrypt.so
-%exclude %{_includedir}/crypt.h
-%else
 %{_libdir}/libcrypt.so
```
ddstreet, line 380: this will mean that we'll also need a glibc-libcrypt-devel package that conflicts with the libxcrypt-devel package, right?

ddstreet (changes requested):
Overall looks good, I added a few comments, and specifically note the comment around adding Suggests: to force preference for glibc-libcrypt. Also we might need to add a glibc-libcrypt-devel so we don't conflict with libxcrypt-devel files.

PR #6385, opened 2023-10-10
Image Customizer: Add `Config` struct.

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Currently, the image customizer API uses the `SystemConfig` struct directly for its config. However, there is supposed to be a parent `Config` struct on top. This is so the API better mirrors the new image config.

###### Change Log  

- Image Customizer: Add `Config` struct to API.

NO

###### Test Methodology

- Ran UTs

### DIFF
--- a/toolkit/tools/imagecustomizerapi/config.go
+++ b/toolkit/tools/imagecustomizerapi/config.go
@@ -1,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+type Config struct {
+	SystemConfig SystemConfig `yaml:"SystemConfig"`
+}
+
+func (c *Config) IsValid() error {
+	err := c.SystemConfig.IsValid()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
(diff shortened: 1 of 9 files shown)


toolkit/tools/imagecustomizerapi/config.go
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+type Config struct {
+	SystemConfig SystemConfig `yaml:"SystemConfig"`
```
dmcilvaney, line 7: Not sure if it makes sense here, but we support a list of configs (mostly as a feature of the ISO installer which can offer a selection of different configs to the user to install).

cwize1: For the image customizer, having a list didn't make much sense and might have caused user confusion. So, I removed it.

reubeno: Is it a goal or non-goal to keep the schemas in sync? Naively, I'd assume that compatibility of interface would be a good thing. If there's a desire to make this more of a singleton and less of a list... is there a way to adapt the existing logic in the core toolkit to accept either a single object *or* a list? That would permit greater compatibility of the format.

cwize1: Not perfectly in sync, no. But the intention is to avoid having differences without good cause.

I could probably extend the image creation schema to also support both `SystemConfig` and `SystemConfigs`.

cwize1: @reubeno, PR for adding `SystemConfig` property to the new image config as well: #6393

PR #6390, opened 2023-10-10
Add the repoquerywrapper tool.

gmileka:
The repoquerywrapper tools creates a chroot, installs dnf packages, and runs specific repoquery commands within that environment.

This tools allows us to run repoquery on systems that may not have repoquery installed/available.

It also allows us to run queries that are not directly supported by repoquery - for example, in this check-in, it allows us to check which packages in a list are available on the target repros.

Such queries are very specific - and can be added incrementally to the tool as the need arises.

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -92,6 +92,7 @@ CCACHE_DIR       ?= $(PROJECT_ROOT)/ccache
 LOGS_DIR           ?= $(BUILD_DIR)/logs
 PKGBUILD_DIR       ?= $(BUILD_DIR)/pkg_artifacts
 CACHED_RPMS_DIR    ?= $(BUILD_DIR)/rpm_cache
+REPO_QUERY_DIR     ?= $(BUILD_DIR)/repo_query
 BUILD_SRPMS_DIR    ?= $(BUILD_DIR)/INTERMEDIATE_SRPMS
 MACRO_DIR          ?= $(BUILD_DIR)/macros
 BUILD_SPECS_DIR    ?= $(BUILD_DIR)/INTERMEDIATE_SPECS
@@ -208,6 +209,10 @@ include $(SCRIPTS_DIR)/toolkit.mk
 # Fill the cache with rpms from the package server without using package manager or chroot with:
 #   pre-cache
 include $(SCRIPTS_DIR)/precache.mk
+
+# Run repo query under chroot with:
+# repo-query
+include $(SCRIPTS_DIR)/repoquerywrapper.mk
 
 # Create a package build workplan with:
 #   workplan, clean-workplan clean-cache
--- a/toolkit/scripts/repoquerywrapper.mk
+++ b/toolkit/scripts/repoquerywrapper.mk
@@ -1,0 +1,57 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# Contains:
+#	- Tools to query rpms repos in bulk.
+
+.PHONY: repo-query clean-repo-query
+
+repoquerywrapper_state_dir = $(REPO_QUERY_DIR)/repoquerywrapper
+repo_urls_file = $(repoquerywrapper_state_dir)/repo_urls.txt
+repoquerywrapper_chroot_dir = $(repoquerywrapper_state_dir)/chroot
+repoquerywrapper_logs_path = $(LOGS_DIR)/repoquerywrapper/repoquerywrapper.log
+
+$(call create_folder,$(repoquerywrapper_state_dir))
+
+clean: clean-repo-query
+clean-repo-query:
+	@echo Verifying no mountpoints present in $(repoquerywrapper_chroot_dir)
+	$(SCRIPTS_DIR)/safeunmount.sh "$(repoquerywrapper_chroot_dir)" && \
+	rm -rf $(repoquerywrapper_state_dir)
+
+#
+# We always want to run the repoquerywrapper tool.
+# However, we do not want targets that depend on repo-query to be re-run if the
+# output of the repoquerywrapper has not changed. Such dependents can reference
+# the $(STATUS_FLAGS_DIR)/repoquerywrapper.flag instead of the repo-query
+# target.
+#
+.PHONY: repoquerywrapper_always_run_phony
+repo-query: $(STATUS_FLAGS_DIR)/repoquerywrapper.flag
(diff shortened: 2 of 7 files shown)


toolkit/tools/internal/packagerepo/repoutils/repoquery.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/sliceutils"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/timestamp"
+)
+
+const (
+	chrootRepoDir      = "/etc/yum.repos.d/"
+)
+
+// GetAllRepoData returns a map of package names to URLs for all packages
```
gmileka, line 26: This code is moved as-is from precacher.go.

toolkit/scripts/repoquerywrapper.mk
```diff
@@ -0,0 +1,48 @@
...
+$(call create_folder,$(repoquerywrapper_state_dir))
+
+clean-repoquerywrapper: clean-repoquerywrapper
+clean: clean-repoquerywrapper
+clean-repoquerywrapper:
+	@echo Verifying no mountpoints present in $(repoquerywrapper_chroot_dir)
+	$(SCRIPTS_DIR)/safeunmount.sh "$(repoquerywrapper_chroot_dir)" && \
+	rm -rf $(repoquerywrapper_state_dir)
+
+# We always want to run the repoquerywrapper script, it will decide if it needs to download anything and update the flag file if
```
dmcilvaney, line 23: Need to update this comment, and consider adjusting the make file a bit here if we want to keep the same sort of behaviour.

This is some weird make magic but here goes:
- Consider dep chain A -> B -> C
- Normally, if C file changes, B will rebuild meaning B timestamp is new, which means A rebuilds, etc...
- If C is a `.PHONY:` target, B will ALWAYS rebuild no matter what
- So, if B rebuilds and updates it's timestamps, A will rebuild

So in the naive case here we have `<some dep>` -> `$(STATUS_FLAGS_DIR)/repoquerywrapper.flag`. What if we only want A to rebuild if things actually change? We can't tell without querying the internet.. though, and we need to run the tool to do that. So, add that .PHONY target in `<some dep>` -> `$(STATUS_FLAGS_DIR)/repoquerywrapper.flag` -> `repoquerywrapper_always_run_phony`. We always run "B" now, but what about A? If nothing changed (ie the repo is the same), there is not point in re-running "A".

So, enter the `*.flag` files... The "output" of the tool is `$(QUERY_OUTPUT_FILE)` but we don't use that as our target, instead we use the `.flag` file. Generally the pattern I've used is:
```make
$(STATUS_FLAGS_DIR)/repoquerywrapper.flag:
    my_tool --output "some_file.txt" && \
    if [ ! -f $@ ] || <some_file.txt changed from last time>; then
        touch $@
    fi
```
This means that "A" only updates on the 1st run, or when the results actually change.

toolkit/scripts/repoquerywrapper.mk
```diff
@@ -0,0 +1,48 @@
...
+		--worker-dir $(repoquerywrapper_chroot_dir) \
+		--log-file=$(repoquerywrapper_logs_path) \
+		--log-level=$(LOG_LEVEL) \
+		--cpu-prof-file=$(PROFILE_DIR)/repoquerywrapper.cpu.pprof \
+		--mem-prof-file=$(PROFILE_DIR)/repoquerywrapper.mem.pprof \
+		--trace-file=$(PROFILE_DIR)/repoquerywrapper.trace \
+		$(if $(filter y,$(ENABLE_CPU_PROFILE)),--enable-cpu-prof) \
+		$(if $(filter y,$(ENABLE_MEM_PROFILE)),--enable-mem-prof) \
+		$(if $(filter y,$(ENABLE_TRACE)),--enable-trace) \
+		--timestamp-file=$(TIMESTAMP_DIR)/repoquerywrapper.jsonl
```
dmcilvaney, line 48: See above, but if we want to use the `.flag` files we should to at least:
```suggestion
		--timestamp-file=$(TIMESTAMP_DIR)/repoquerywrapper.jsonl
	touch $@
```

Alternatively, change the target to:
$(QUERY_OUTPUT_FILE): $(go-repoquerywrapper) $(chroot_worker) repoquerywrapper_always_run_phony

toolkit/Makefile
```diff
@@ -92,6 +92,7 @@ CCACHE_DIR       ?= $(PROJECT_ROOT)/ccache
 LOGS_DIR           ?= $(BUILD_DIR)/logs
 PKGBUILD_DIR       ?= $(BUILD_DIR)/pkg_artifacts
 CACHED_RPMS_DIR    ?= $(BUILD_DIR)/rpm_cache
+REPO_QUERY_DIR     ?= $(BUILD_DIR)/repo-query
```
anphel31, line 95: I'd suggest `repo_query` to match the other directories which use underscore instead of a dash

toolkit/scripts/repoquerywrapper.mk
gmileka: thx Daniel - I've updated the make file to implement the logic you've described.

toolkit/scripts/repoquerywrapper.mk
gmileka: Yep - fixed as well.

toolkit/Makefile
gmileka: thx - fixed.

dmcilvaney (approved):

toolkit/scripts/update-target-if-output-changed.sh
```diff
@@ -0,0 +1,36 @@
...
+# $1 path to flag file that will be created if needed.
+# $2 path to the old output file.
+# $3 path to the new output file.
+flagFile=${1}
+oldOutputFile=${2}
+newOutputFile=${3}
+
+usage() {
+	echo
+	echo "./update-target-if-output-changed.sh.sh <flag-file> <old-output-file> <new-output-file>"
```
anphel31, line 14: typo: `.sh.sh`

anphel31 (approved):

toolkit/scripts/update-target-if-output-changed.sh
gmileka: Fixed.

PR #6391, opened 2023-10-10
fix: Add grub2-mkconfig macros to initramfs postrans generation

Camelron:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

I discovered another race condition on mkconfig-based images: 
1. kernel upgrade happens during rpm transaction
2. grub2_post triggers, causing the grub.cfg to be generated
3. posttrans script is called to regenerate initramfs for the new kernel bin
4. !!! since the initramfs was not present at the time we last called mkconfig, it is not part of the menuentry. Thus, we can fail to boot or boot with different behavior.

This change makes it so that grub2-mkconfig is always called after the posttrans script for initramfs, ensuring that the most final version of the image is used to configure the final menuentries.  

Furthermore, it backports several other fixes discovered when integrating mkconfig changes in 3.0-dev:
- cp initrd-mshv images in /boot --> /boot/efi 
- check /etc/default/grub.d/* execution bits before sourcing them
- set $prefix variable in UEFI-based image rump grub.cfg (fixes grubenv, grub2-editenv, related utilities)
 
###### Change Log  

- Add grub2_post to initramfs posttrans script
- cp initrd-mshv images in /boot --> /boot/efi 
- check /etc/default/grub.d/* execution bits before sourcing them
- set $prefix variable in UEFI-based image rump grub.cfg (fixes grubenv, grub2-editenv, related utilities)

NO

###### Test Methodology

- Tested all the way through our aks pipelines, deployed test cluster with az aks create..


SPECS/initramfs/initramfs.spec
```diff
@@ -101,6 +102,7 @@ elif [ -d %{_localstatedir}/lib/rpm-state/initramfs/pending ]; then \
         fi \
     done; \
 fi \
+%grub2_post
```
dmcilvaney, line 105: Do we need a similar entry for `%posttrans`? I don't 100% recall when each flow runs, but maybe better to be defensive and do both.

Camelron: Yes, I think that is a good idea. Added.

Camelron (commented):
Note: this should probably be a 'cp', not a 'mv'

Should we roll it into this PR or make another PR/version?

dmcilvaney (approved):

toolkit/tools/internal/resources/assets/efi/grub/grub.cfg
```diff
@@ -1,6 +1,11 @@
 search -n -u {{.BootUUID}} -s
+search.fs_uuid {{.BootUUID}} root
```
dmcilvaney, line 2: This is equivalent isn't it? We should use one or the other I think.

SPECS/initramfs/initramfs.spec
```diff
@@ -101,14 +102,16 @@ elif [ -d %{_localstatedir}/lib/rpm-state/initramfs/pending ]; then \
...
 %removal_action
 
 %posttrans
 echo "initramfs" %{version}-%{release} "posttrans" >&2
 %removal_action
 mkinitrd -q
-# Move initrd generated for kernel-mshv to /boot/efi, where linuxloader expects to find it
-mv /boot/initrd.img-*mshv* /boot/efi/ >/dev/null 2>&1 || :
+# Copy initrd generated for kernel-mshv to /boot/efi, where linuxloader expects to find it
+cp /boot/initrd.img-*mshv* /boot/efi/ >/dev/null 2>&1 || :
```
ddstreetmicrosoft, line 113: It's probably a bad idea to ignore failure here, as the ESP could easily be full...

toolkit/tools/internal/resources/assets/efi/grub/grub.cfg
```diff
@@ -1,6 +1,9 @@
 search -n -u {{.BootUUID}} -s
-
 # If '/boot' is a seperate partition, BootUUID will point directly to '/boot'.
 # In this case we should omit the '/boot' prefix from all paths.
 set bootprefix={{.BootPrefix}}
+# For images using grub2-mkconfig, $prefix is the variable
+# We need to populate with the proper path to menuentry grub.cfg,
+# grubenv
+set prefix=($root)'/boot/grub2'
```
ddstreetmicrosoft, line 8: why do we need this? grub detects the prefix and we really shouldn't be hardcoding it, right?

toolkit/tools/internal/resources/assets/grub2/grub
```diff
@@ -7,7 +7,7 @@ GRUB_CMDLINE_LINUX_DEFAULT="{{.ExtraCommandLine}} $kernelopts"
     
 # =============================notice===============================
 # IMPORTANT: package and feature-specific behaviors are defined in
-#   /etc/default/grub.d/*.cfg. The cfg files are sourced last
+#   /etc/default/grub.d/*.cfg. The executable cfg files are sourced last
```
ddstreetmicrosoft, line 10: they don't need to be (and shouldn't) be executable...they're intended to be sourced, not run

toolkit/tools/internal/resources/assets/grub2/grub
```diff
@@ -16,7 +16,7 @@ GRUB_CMDLINE_LINUX_DEFAULT="{{.ExtraCommandLine}} $kernelopts"
 #     - and finally GRUB_CMDLINE_LINUX_DEFAULT 
 # =============================notice===============================
 for x in /etc/default/grub.d/*.cfg ; do
-    if [ -e "${x}" ]; then
+    if [ -x "${x}" ]; then
```
ddstreetmicrosoft, line 19: no this is wrong

ddstreetmicrosoft (changes requested):
left some comments

SPECS/initramfs/initramfs.spec
Camelron: Good point, I'll make that fix.

toolkit/tools/internal/resources/assets/efi/grub/grub.cfg
Camelron: In practice, on systems booting from an ESP partition, it thinks $prefix is /boot/efi/boot/grub2 which causes /boot/grub2/grubenv to not be sourced. 

This breaks things like grub2-editenv on Mariner.

toolkit/tools/internal/resources/assets/efi/grub/grub.cfg
```diff
@@ -1,6 +1,9 @@
...
-
 # If '/boot' is a seperate partition, BootUUID will point directly to '/boot'.
 # In this case we should omit the '/boot' prefix from all paths.
 set bootprefix={{.BootPrefix}}
-configfile $bootprefix/grub2/grub.cfg
\ No newline at end of file
+# For images using grub2-mkconfig, $prefix is the variable
+# We need to populate with the proper path to menuentry grub.cfg,
+# grubenv
+set prefix=($root)"$bootprefix/grub2"
```
Camelron, line 8: prefix:
the path to /boot/grub2/ relative to the bootUUID
bootprefix:
the path to /boot/ relative to the bootUUID

Both variables must be defined in order to support our legacy grub.cfg as well as mkconfig-generated ones.

ddstreetmicrosoft: it looks like `bootprefix` is a simple path, while `prefix` includes the leading drive identifier; does that affect use of `configfile`? It looks like before `configfile` was called without the leading drive identifier, while now it will be called with the drive identifier. Is that change intentional and correct?

Also, using `$root` for the drive identifier probably isn't correct is it? Won't `$root` point to the partition to use as the rootfs? But the grub drive identifier format is like `(hd0,1)`, and `$root` won't be set to `hd0,1` will it?

toolkit/tools/internal/resources/assets/efi/grub/grub.cfg
```diff
@@ -1,6 +1,11 @@
...
 set bootprefix={{.BootPrefix}}
-configfile $bootprefix/grub2/grub.cfg
\ No newline at end of file
+# For images using grub2-mkconfig, $prefix is the variable
+# grub expects to be populated with the proper path to the grub.cfg, grubenv.
+# Thus, there are two variables:
+#    - $bootprefix: the path to /boot/ relative to the {{.BootUUID}}
+#    - $prefix: the path to /boot/grub2/ relative to the {{.BootUUID}}
+set prefix=($root)"$bootprefix/grub2"
+configfile $prefix/grub.cfg
```
ddstreetmicrosoft, line 11: can you please make sure to fix your Windows text editor to stop eliding the EOF newline?

SPECS/initramfs/initramfs.spec
```diff
@@ -101,14 +102,20 @@ elif [ -d %{_localstatedir}/lib/rpm-state/initramfs/pending ]; then \
...
 %posttrans
 echo "initramfs" %{version}-%{release} "posttrans" >&2
 %removal_action
 mkinitrd -q
-# Move initrd generated for kernel-mshv to /boot/efi, where linuxloader expects to find it
-mv /boot/initrd.img-*mshv* /boot/efi/ >/dev/null 2>&1 || :
+# Copy initrd generated for kernel-mshv to /boot/efi, where linuxloader expects to find it
+cp /boot/initrd.img-*mshv* /boot/efi/
+if [ $? -ne 0 ]; then
+    echo "WARNING: ESP partition (/boot/efi) is likely full!"
```
ddstreetmicrosoft, line 115: I think this should be a fatal error, if we allow rpm installation to continue the system could wind up in an unbootable state...

ddstreetmicrosoft (changes requested):

toolkit/tools/internal/resources/assets/efi/grub/grub.cfg
Camelron: It comes down to the behavior of the search command we use on ln1 (https://www.gnu.org/software/grub/manual/grub/html_node/search.html).
It is implicitly setting $root to the drive identified by "{{.BootUUID}}". In Mariner, {{.BootUUID}} can be either the rootfs (in which case bootprefix="/boot") or it can be a separate boot partition (like when we need a verity-protected root partition, bootprefix="")

Note that grub interprets paths that do not specify a device (a relative path) to be $(root)/"relpath" https://www.gnu.org/software/grub/manual/grub/html_node/root.html#root

So, we are just being more explicit here in providing the device name. This ensures that even if we change root later (by running another search command...), prefix is still correct.

toolkit/tools/internal/resources/assets/efi/grub/grub.cfg
Camelron: Ack, followed the advice in the email thread you started.

SPECS/initramfs/initramfs.spec
Camelron: But this is posttrans, right? Isn't the damage already done?

Camelron: Hold on, we just removed support for the old linuxloader https://dev.azure.com/mariner-org/mariner/_git/CBL-Mariner-Proprietary/pullrequest/17113. We now can rely upon the initrd present in /boot and remove this posttrans logic to duplicate it.

Camelron: Removing this block of logic entirely, no longer needed.

toolkit/tools/internal/resources/assets/efi/grub/grub.cfg
ddstreetmicrosoft: sounds good to me, thanks

ddstreetmicrosoft (approved):
lgtm

christopherco (approved):

PR #6430, opened 2023-10-13
Image Customizer: Add support to load and unload modules

elainezhao96:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary
add support to load or unload modules to image customizer tool

###### Change Log
Image customizer: add support to load or unload modules

NO

###### Test Methodology
did `go run  . --build-dir ./build --config-file ../pkg/imagecustomizerlib/testdata/modules-config.yaml --image-file core-2.0.20231004.vhdx --o
utput-image-file ./out/image1.vhd --output-image-format vhd --log-file=trace` and test with output image , the modules specified in the testdata file are loaded and unloaded.
image

note: This feature only supports loading existing modules or those with the required environment, and unloading modules without dependencies. For example, it can't load 'ptp_kvm' due to a 'No such device' error, nor unload 'nfit' because of its dependencies.

### DIFF
--- a/toolkit/tools/imagecustomizerapi/module.go
+++ b/toolkit/tools/imagecustomizerapi/module.go
@@ -20,7 +20,7 @@ func (m *Module) IsValid() error {
 }
 
 type Modules struct {
-	Load   []Module `yaml:"Load"`
+	Load    []Module `yaml:"Load"`
 	Disable []Module `yaml:"Disable"`
 }
 
--- a/toolkit/tools/imagecustomizerapi/module.go
+++ b/toolkit/tools/imagecustomizerapi/module.go
@@ -1,0 +1,41 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+)
+
+type Modules struct {
+	Load   []Module `yaml:"Load"`
+	Unload []Module `yaml:"Unload"`
+}
+
+type Module struct {
+	Name string `yaml:"Name"`
+}
+
+func (m *Module) IsValid() error {
+	if m.Name == "" {
+		return fmt.Errorf("name of module may not be empty")
+	}
+
+	return nil
+}
+
+func (m *Modules) IsValid() error {
+	for i, module := range m.Load {
+		if err := module.IsValid(); err != nil {
+			return fmt.Errorf("invalid Modules.Load item at index %d: %w", i, err)
+		}
+	}
+
+	for i, module := range m.Unload {
+		if err := module.IsValid(); err != nil {
+			return fmt.Errorf("invalid Modules.Unload item at index %d: %w", i, err)
+		}
+	}
+
+	return nil
+}
--- a/toolkit/tools/imagecustomizerapi/module.go
+++ b/toolkit/tools/imagecustomizerapi/module.go
@@ -7,11 +7,6 @@ import (
 	"fmt"
 )
 
-type Modules struct {
-	Load   []Module `yaml:"Load"`
-	Unload []Module `yaml:"Unload"`
-}
-
 type Module struct {
 	Name string `yaml:"Name"`
 }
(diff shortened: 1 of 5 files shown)


toolkit/tools/imagecustomizerapi/module.go
```diff
@@ -0,0 +1,41 @@
...
+type Modules struct {
+	Load   []Module `yaml:"Load"`
+	Unload []Module `yaml:"Unload"`
+}
+
+type Module struct {
+	Name string `yaml:"Name"`
+}
+
+func (m *Module) IsValid() error {
```
cwize1, line 19: nit: You should probably have the order of the `IsValid()` functions match the order of the structs.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -325,3 +330,30 @@ func enableOrDisableServices(services imagecustomizerapi.Services, baseConfigPat
...
+		logger.Log.Infof("Loading module (%s)", module.Name)
+		moduleFileName := module.Name + ".conf"
+		moduleFilePath := filepath.Join(imageChroot.RootDir(), "/etc/modules-load.d/", moduleFileName)
+		err = file.Write(module.Name, moduleFilePath)
+		if err != nil {
+			return fmt.Errorf("failed to write module load configuration: %w", err)
+		}
+	}
+
+	for _, module := range modules.Unload {
```
cwize1, line 347: Should `Unload` be renamed to `Disable`?

The module isn't really being "unloaded", since it is never loaded in the first place. And neither are you removing it from the "load" list but instead blocking it from being loaded at all.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -325,3 +330,30 @@ func enableOrDisableServices(services imagecustomizerapi.Services, baseConfigPat
 
 	return nil
 }
+
+func loadOrUnloadModules(modules imagecustomizerapi.Modules, imageChroot *safechroot.Chroot) error {
+	var err error
+
+	for _, module := range modules.Load {
+		logger.Log.Infof("Loading module (%s)", module.Name)
```
cwize1, line 338: nit: `"Loading kernel module (%s)"` just to be more specific.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -325,3 +330,30 @@ func enableOrDisableServices(services imagecustomizerapi.Services, baseConfigPat
...
+		moduleFileName := module.Name + ".conf"
+		moduleFilePath := filepath.Join(imageChroot.RootDir(), "/etc/modules-load.d/", moduleFileName)
+		err = file.Write(module.Name, moduleFilePath)
+		if err != nil {
+			return fmt.Errorf("failed to write module load configuration: %w", err)
```
cwize1, line 343: Might be worth adding the `module.Name` to the error string.

Same thing below.

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/module.go
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+)
+
+type Modules struct {
+	Load   []Module `yaml:"Load"`
```
romoh, line 11: Should we explicitly call it "Blacklist", so it sets the right expectations?

cwize1: The word "blacklist" should generally be avoided. Substitutions include "block list", "deny list", and "disallow list".

(Though there isn't much that can be done when using an existing API, such as `/etc/modprobe.d/`.)

romoh: I hear you.. but then I thought that will be the best representation of what we do under the hoods and all related implications that comes with modprobe implementation. 
Maybe that's something we can clarify instead in our documentation

cwize1 (approved):

PR #6446, opened 2023-10-16
Add tdnf remove cache script and run it for marketplace images

mbykhovtsev-ms:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Adds a script that cleans and removes tdnf cache. Runs the script as post-install for marketplace generated images. Needed to fix bug https://microsoft.visualstudio.com/OS/_workitems/edit/45977096

###### Change Log  

- Adds a script that cleans and removes tdnf cache
- Runs tdnf remove cache as post-install step for marketplace image generation

**YES**

###### Test Methodology

- Built new ISO images locally, installed Mariner from ISO images and booted up. Confirmed that the tdnf cache was clear inside of installed Mariner VM


toolkit/imageconfigs/postinstallscripts/remove-tdnf-cache.sh
```diff
@@ -0,0 +1,3 @@
+echo removing tdnf cache
```
anphel31, line 1: Can you add the copyright notice to the top of this script?
Ex: https://github.com/microsoft/CBL-Mariner/blob/3d3fcf7ecb1300e436d07a59ed78279ac89f5985/.pipelines/prchecks/PackageBuildPRCheck.yml#L1C1-L2C34

mbykhovtsev-ms:
> Maybe a future todo, but is there a reason we don't want to integrate this into our tools rather than use a customization script? This seems like a fairly standard thing we might want to do across all of our images.

This can be added to our tools for all images. I was just not sure if we needed them for all images or just the marketplace images, the bug I was trying to solve only talks about marketplace images, so I wanted to consult with the team whether it is needed for all of them? If it does, I can just integrated it as part of our `cleanup.sh` script that is run for all images

dmcilvaney:
> > Maybe a future todo, but is there a reason we don't want to integrate this into our tools rather than use a customization script? This seems like a fairly standard thing we might want to do across all of our images.
> 
> This can be added to our tools for all images. I was just not sure if we needed them for all images or just the marketplace images, the bug I was trying to solve only talks about marketplace images, so I wanted to consult with the team whether it is needed for all of them? If it does, I can just integrate it as part of our `cleanup.sh` script that is run for all images

I think that is unique to just the container image. I was thinking more towards the imager tool itself, with a flag in the config schema. I'm not aware of any good reason to keep the cache in any image.

tobiasb-ms (approved):

toolkit/imageconfigs/marketplace-gen1-fips.json
```diff
@@ -67,6 +67,9 @@
             "PostInstallScripts": [
                 {
                     "Path": "additionalconfigs/configure-image.sh"
+                },
+                {
+                    "Path": "postinstallscripts/remove-tdnf-cache.sh"
```
PawelWMS, line 72: The change looks fine but I'd consider these two modifications to either:
- add the contents of the new script into `configure-image.sh` or
- re-name `configure-image.sh` to something more precise like `configure-systemd-networkd.sh`.

Currently, I'm leaning more towards the second option, so the configuration scripts are more granular and their responsibility is clearer.

mbykhovtsev-ms: Added the renaming

PR #6479, opened 2023-10-19
fix: introduce mariner_2_initrd_use_suffix kdump.conf option

### LINKED ISSUE #6478: kdump minimal initrd rebuild overwrites host's initrd
kexec-tools kdump.conf default config sets `force_no_rebuild 1` which has kdump rely on the default initrd as the crashkernel environment. Typically `force_no_rebuild` is not set, allowing kdumpctl to build a minimal host-specific initrd for the crashkernel environment. However the regenerated initrd actually overwrites the host's default initrd, thus leading to broken behavior after rebooting out of the crashkernel environment.

christopherco:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

There are two issues. First, our default kdump configuration causes the
kdump service to use the host's default initrd as the crashkernel
environment. This can lead to issues when the default initrd size is
larger than the reserved memory for the crash kernel, which is set via
kernel command line at boot time.

It is common to have the kdump service instead rebuild a minimal
host-specific initrd specifically for the crashkernel environment.

To change this behavior, comment out force_no_rebuild 1 from kdump.conf.

After doing this, a second issue was discovered where the system would
enter the crashkernel environment when a kernel panic occurred, and
successfully collect the crash dump, but upon reboot back to the normal
host OS, the system would no longer boot.

This behavior was root caused to an issue in our kdumpctl command where
the TARGET_INITRD was pointing to our default host initrd, thus when the
kdump service would regenerate the minimal host-specific initrd, this
new initrd would overwrite the host's normal initrd, therefore leading
to failed normal boot.

Since Mariner 2.0 already has a precident to use the host system's initrd,
we need to still preserve this behavior by default, but allow users to opt
in to the better option.

Therefore, this change introduces a new "mariner_2_initrd_use_suffix"
option. When set, the option appends "kdump" suffix to TARGET_INITRD path,
which means the host system's initrd is no longer being targeted for
kdump.

This change also adds a guard-rail in kdumpctl to ensure both
"force_no_rebuild" and "mariner_2_initrd_use_suffix" are not set, otherwise
the kdump will fail to arm correctly since kdump will not be able to locate
the host's initrd.

When compressed kdump collection is enabled, vmcore data is now being
stored in /var/crash/<IP>-<DATE>-<TIME> directory using the
kdump-lib-initramfs.sh script. Specifically
   /var/crash/<IP>-<date +%Y-%m-%d-%T>

Previously, the vmcore data was being saved from kdumpctl, which was
storing vmcores in /var/crash/<DATE>-<TIME>. Specifically

   /var/crash/<date +%Y-%m-%d-%H:%M>

Since there could be automation already in place that expects the older
format, adjust the newer compressed kdump version to align with the older
/var/crash/<DATE>-<TIME> directory naming.

**NO**

###### Associated issues  

- fixes #6478 
- https://microsoft.visualstudio.com/OS/_workitems/edit/43839616

###### Test Methodology

- Pipeline build id: [510074](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=510074&view=results)
- option permutation:
   - force_no_reload 1 && mariner_2_initrd_add_suffix 0 => observe original default behavior
   - force_no_reload 0 && mariner_2_initrd_add_suffix 1 => observe compressed kdump behavor
   - force_no_reload 1 && mariner_2_initrd_add_suffix 1 => observe kdump service failure


christopherco:
Hmmm I am concerned about the cleanup process for new TARGET_INITRDs. As the kernel upgrades and downgrades, new minimal kdump initrds will be created and persist in /boot

christopherco:
Another side-effect of this change is the actual /var/crash directory naming schema is different after using the minimal host-specific initrd

``` console
# tree /var/crash
/var/crash
|-- 127.0.0.1-2023-10-20-03:03:50
|   |-- vmcore
|   `-- vmcore-dmesg.txt
|-- 2023-10-20-02:35
|   `-- vmcore
```

On the plus side, core dumps collection time on larger memory systems improved significantly due to specialized crashkernel initrd environment (i.e., on a system with 128GB memory, vmcore creation time went from ~2 mins to ~20 sec). Additionally the vmcore file is lzo compressed so size on disk is much smaller. We also get the vmcore-dmesg as a text file

ddstreetmicrosoft:
> Hmmm I am concerned about the cleanup process for new TARGET_INITRDs. As the kernel upgrades and downgrades, new minimal kdump initrds will be created and persist in /boot if there isn't one present already

yep, it's probably ok since we don't use a small separate /boot (do we?), though it's true we would be 'leaking' the kdump initrds created for older kernels, which isn't good.

I'm sure we could come up with a hack to address this, like some additional script to try to find and remove 'old' initrd, or maybe set up a tmpfiles.d config to remove them by age, or maybe just use a single non-versioned kdump initrd path and force recreation each time kdump.service is started (i.e. use force_rebuild 1). I don't think any hack is really approriate for a 'stable' release though, and I think it's likely better to just leave the 'leaked' kdump initrds until we can do it better in mariner 3.

ddstreetmicrosoft:
> Another side-effect of this change is the actual /var/crash directory naming schema is different after using the minimal host-specific initrd
> 
> ```
> # tree /var/crash
> /var/crash
> |-- 127.0.0.1-2023-10-20-03:03:50
> |   |-- vmcore
> |   `-- vmcore-dmesg.txt
> |-- 2023-10-20-02:35
> |   `-- vmcore
> ```

while the specific path format doesn't really matter, if anyone out there is actually (programatically) checking for kernel crashes, the path change could break them. So it's probably not good for mariner 2. However the path is probably set by something that we could change so the kdump-specific initrd still uses the previous path format...

SPECS/kexec-tools/kdump.conf
```diff
@@ -179,7 +179,7 @@ core_collector makedumpfile -l --message-level 1 -d 31
 #extra_modules gfs2
 #failure_action shell
 #force_rebuild 1
-force_no_rebuild 1
+#force_no_rebuild 1
```
ddstreetmicrosoft, line 182: since this is changing a config(noreplace) file, when existing instances upgrade to this version, if they've modified this config file themselves then this change won't get applied. When that is combined with the kdumpctl change to the initrd path, this could result in the kdump.service failing to start, because no kdump initrd exists (i.e. an explicit kdump rebuild would be needed to create it, or manually editing this config file on each instance).

ddstreetmicrosoft:
generally it looks fine as a way to get it working for mariner 2 without too much difficulty, however i think the potential issues probably need to be addressed, or some other method used...

christopherco:
Encountering an issue on ARM64 where, when booting into the new crashkernel environment yields the following error

``` console
[  406.251915] Memory Limit: none
[  406.359332] Starting crashdump kernel...
[  406.362558] Bye!
/bin/sh: error while loading shared libraries: libreadline.so.8: cannot open shared object file: No such file or directory
[    0.286994] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00007f00
[    0.292531] CPU: 0 PID: 1 Comm: init Not tainted 5.15.133.1-1.cm2 #1
[    0.297194] Hardware name: Microsoft Corporation Virtual Machine/Virtual Machine, BIOS Hyper-V UEFI Release v4.1 05/16/2022
[    0.305485] Call trace:
[    0.307289]  dump_backtrace+0x0/0x1f4
[    0.310063]  show_stack+0x20/0x2c
[    0.312547]  dump_stack_lvl+0x8c/0xb8
[    0.315269]  dump_stack+0x18/0x34
[    0.317719]  panic+0x194/0x384
[    0.320022]  do_exit+0xb04/0xb30
[    0.322367]  do_group_exit+0x40/0xb0
[    0.325083]  __wake_up_parent+0x0/0x40
[    0.327796]  invoke_syscall+0x78/0x100
[    0.330630]  el0_svc_common.constprop.0+0x6c/0x1a0
[    0.334303]  do_el0_svc+0x30/0xb0
[    0.336784]  el0_svc+0x2c/0xc0
[    0.339083]  el0t_64_sync_handler+0xa4/0x12c
[    0.342234]  el0t_64_sync+0x1a4/0x1a8
```

The minimal initrd looks to be sane and has libreadline.so.8 present:
``` console
# tree .
.
|-- bin -> /usr/bin
|-- init
|-- lib -> usr/lib
|-- sbin -> /usr/sbin
|-- squash
|-- squash-root.img
`-- usr
    |-- bin
    |   |-- bash
    |   |-- grep
    |   |-- kmod
    |   |-- libreadline.so.8
    |   |-- mkdir
    |   |-- mount
    |   `-- sh -> bash
    |-- lib
    |   |-- dracut
    |   |   |-- build-parameter.txt
    |   |   |-- dracut-055-5.cm2
    |   |   |-- hostonly-files
    |   |   |-- hostonly-kernel-modules.txt
    |   |   |-- modules.txt
    |   |   |-- need-initqueue
    |   |   `-- no-switch-root
    |   |-- ld-linux-aarch64.so.1
    |   |-- libblkid.so -> ../../lib/../../lib/libblkid.so.1.1.0
    |   |-- libblkid.so.1 -> ../../lib/libblkid.so.1.1.0
    |   |-- libblkid.so.1.1.0
    |   |-- libc.so
    |   |-- libc.so.6
    |   |-- libgcc_s.so
    |   |-- libgcc_s.so.1
    |   |-- liblzma.so -> ../../lib/../../usr/lib/liblzma.so.5.2.5
    |   |-- liblzma.so.5 -> ../../lib/liblzma.so.5.2.5
    |   |-- liblzma.so.5.2.5
    |   |-- libmount.so -> ../../lib/../../lib/libmount.so.1.1.0
    |   |-- libmount.so.1 -> ../../lib/libmount.so.1.1.0
    |   |-- libmount.so.1.1.0
    |   |-- libncursesw.so.6 -> libncursesw.so.6.4
    |   |-- libncursesw.so.6.4
    |   |-- libpcre.so.1 -> ../../lib/libpcre.so.1.2.13
    |   |-- libpcre.so.1.2.13
    |   |-- libreadline.so.8 -> ../../lib/libreadline.so.8.1
    |   |-- libreadline.so.8.1
    |   |-- libselinux.so.1
    |   |-- libtinfo.so.6 -> libtinfo.so.6.4
    |   |-- libtinfo.so.6.4
    |   |-- libz.so -> ../../lib/../../lib/libz.so.1.2.13
    |   |-- libz.so.1 -> ../../lib/libz.so.1.2.13
    |   |-- libz.so.1.2.13
    |   `-- modules
    |       `-- 5.15.133.1-1.cm2
    |           |-- kernel
    |           |   `-- fs
    |           |       `-- overlayfs
    |           |           `-- overlay.ko.xz
    |           |-- modules.alias
    |           |-- modules.alias.bin
    |           |-- modules.builtin
    |           |-- modules.builtin.alias.bin
    |           |-- modules.builtin.bin
    |           |-- modules.builtin.modinfo
    |           |-- modules.dep
    |           |-- modules.dep.bin
    |           |-- modules.devname
    |           |-- modules.order
    |           |-- modules.softdep
    |           |-- modules.symbols
    |           `-- modules.symbols.bin
    `-- sbin
        |-- modprobe -> ../../bin/kmod
        `-- switch_root
```

As a test, I attempted to chroot into this initrd environment and encountered the same error. Setting LD_LIBRARY_PATH="/lib" does resolve the error and I was able to chroot into a valid shell environment.

Additionally, when comparing this ARM64 kdump initrd with a working x86_64 kdump initrd, there is a difference where the x86_64 kdump initrd had a symlink for `lib64` to `usr/lib` that was missing from the ARM64 kdump initrd. Experimentally with the same chroot environment and unsetting the LD_LIBRARY_PATH env var and having this `lib64` symlink, I was able to chroot into a valid shell environment.

However recreating the initrd with the `lib64` symlink to `usr/lib` still yielded the same errors in the crash kernel environment.

So I am currently stumped...

christopherco:
> yep, it's probably ok since we don't use a small separate /boot (do we?), though it's true we would be 'leaking' the kdump initrds created for older kernels, which isn't good.

Our default Azure VM images actually do have a separate /boot partition, which is about 500MB in size.

> while the specific path format doesn't really matter, if anyone out there is actually (programatically) checking for kernel crashes, the path change could break them. So it's probably not good for mariner 2. However the path is probably set by something that we could change so the kdump-specific initrd still uses the previous path format...

Agreed. We don't want the path change to break current scripts expecting this pattern in Mariner 2.

SPECS/kexec-tools/kdump.conf
christopherco: Good catch. I'm considering a change into kdumpctl to adjust the TARGET_INITRD path based on the force_no_rebuild value, so the proper TARGET_INITRD is used:
- if `force_no_rebuild 1`, `TARGET_INITRD="${KDUMP_BOOTDIR}/initrd.img-${kdump_kver}"`
- if `force_no_rebuild 0` -> `TARGET_INITRD="${KDUMP_BOOTDIR}/initrd.img-${kdump_kver}kdump`
 
And remove this change in Mariner 3.0 wholesale.
This change may be better served as a patch file to kexec-tools, rather than a direct edit into kdumpctl.

ddstreetmicrosoft:
> > yep, it's probably ok since we don't use a small separate /boot (do we?), though it's true we would be 'leaking' the kdump initrds created for older kernels, which isn't good.
> 
> Our default Azure VM images actually do have a separate /boot partition, which is about 500MB in size.

ah well in that case it certainly could become an issue.

ddstreetmicrosoft (approved):
The conf file change is part of a noreplace config file, so it shouldn't break any existing users who do have manual changes. Additionally, the default config (with force_no_rebuild 1) will not break end users by replacing their initrd. So this change should allow a backwards-compatible way for end users to opt-in to rebuilding a kdump-specific initrd.

The concern about filling up the /boot filesystem is still a possibility, but I think it's a low probability for most end users, and the result of not addressing this bug (i.e. kudmp overwriting the main initrd and leaving the system unbootable) is far worse, so I think this LGTM.

The only suggestion I have is that it may be useful to open a new issue to explain the possibility of /boot filling due to old kdump initrd, and track any discussion and/or fixes for that in the new issue.

PR #6487, opened 2023-10-19
Image Customizer: Extend modules load API to include flags

elainezhao1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary
Extend modules load API to include flags

###### Change Log  
Image customizer: Extend modules load API to include flags

NO

###### Test Methodology
test with sample config for vfio, the modules is loaded and the parameter is set based on the options in the config file
image


toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -346,6 +347,21 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
...
 			return fmt.Errorf("failed to write module load configuration: %w", err)
 		}
+
+		if module.Options != nil {
+			var options []string
+			for key, value := range module.Options {
+				options = append(options, fmt.Sprintf("%s=%s", key, value))
+			}
+
+			moduleOptionsFileName := module.Name + "-options.conf"
```
romoh, line 357: Looks like we are creating an -options file for each module. Should we instead aggregate them all to one config?

toolkit/tools/imagecustomizerapi/module.go
```diff
@@ -8,14 +8,23 @@ import (
 )
 
 type Module struct {
```
cwize1, line 10: This design is a little odd. If you want to set kernel module options, you have to set it to be auto-enabled. Also, the `Options` field does nothing when attached to the `Disable` list.

Perhaps the `Enable` and `Disable` list should just be string lists and then have a separate list for configuring options?

toolkit/tools/imagecustomizerapi/module.go
```diff
@@ -8,14 +8,23 @@ import (
...
+	Options map[string]string `yaml:"Options,omitempty"`
 }
 
 func (m *Module) IsValid() error {
 	if m.Name == "" {
 		return fmt.Errorf("name of module may not be empty")
 	}
 
+	if m.Options != nil {
+		for key, value := range m.Options {
```
cwize1, line 21: Should be check that `key` and `value` don’t contain any spaces or newline characters?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -338,21 +339,47 @@ func enableOrDisableServices(services imagecustomizerapi.Services, imageChroot *
...
 			return fmt.Errorf("failed to write module load configuration: %w", err)
 		}
+		loadModuleMap[module] = true
+	}
+
+	var modulesWithOptions []string
+	for module, moduleOptions := range modules.Options {
+		if !loadModuleMap[module] {
+			// load the module in options list if it is not specified in load list
+			if err := writeModuleLoadConfiguration(module, imageChroot); err != nil {
```
cwize1, line 354: Why do we want to force the module to load if options are specified?

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -338,21 +339,47 @@ func enableOrDisableServices(services imagecustomizerapi.Services, imageChroot *
...
+			options = append(options, fmt.Sprintf("%s=%s", key, value))
+		}
+
+		data := fmt.Sprintf("options %s %s", module, strings.Join(options, " "))
+		modulesWithOptions = append(modulesWithOptions, data)
+	}
+
+	if len(modulesWithOptions) > 0 {
+		modulesWithOptions = append(modulesWithOptions, "")
+		moduleOptionsFilePath := filepath.Join(imageChroot.RootDir(), "/etc/modprobe.d/module-options.conf")
```
cwize1, line 371: It seems a little odd that for load and disable you give each modules its own config file. But for options, you combine them all into a single file.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
elainezhao96: if understand previous comment correctly "If you want to set kernel module options, you have to set it to be auto-enabled", so I thought in a yaml file, if a module is in options list but not in load list, we will also need to add a .conf in /etc/modules-load.d to load it, in combination with an options config in /etc/modprobe.d/module-options.conf

toolkit/tools/imagecustomizerapi/module.go
```diff
@@ -5,37 +5,54 @@ package imagecustomizerapi
...
+		}
+	}
+
+	for i, moduleName := range m.Disable {
+		if err := validateModuleName(moduleName, "Disable", i); err != nil {
+			return err
 		}
 	}
+	for moduleName, moduleOptions := range m.Options {
+		if moduleName == "" {
```
cwize1, line 32: Should you also check for spaces and newlines against this `moduleName`?

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: It's possible that a user might want to set module options but continue to let the module load only if the correct hardware is present, instead of forcing the module to load on boot.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: I think it might be better to unify on 1 file per module instead of a 1 file per operation.

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/module.go
```diff
@@ -5,37 +5,54 @@ package imagecustomizerapi
...
-func (m *Module) IsValid() error {
-	if m.Name == "" {
-		return fmt.Errorf("name of module may not be empty")
-	}
-
-	return nil
-}
+type ModuleOptions map[string]string
 
 type Modules struct {
```
cwize1, line 13: Please add these changes to `toolkit/tools/imagecustomizer/docs/configuration.md`.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
elainezhao96: make sense, then I will not put it in the list for boot time loading.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -338,26 +339,62 @@ func enableOrDisableServices(services imagecustomizerapi.Services, imageChroot *
...
+		logger.Log.Infof("Loading kernel module %s", module)
+		modulesToLoad = append(modulesToLoad, module)
+		modulesToLoadMap[module] = true
+	}
+
+	var modulesWithOptions []string
+	for module, moduleOptions := range modules.Options {
+		if !modulesToLoadMap[module] {
+			// Load module in options list if it is not specified in load list
+			logger.Log.Infof("Loading kernel module %s", module)
```
romoh, line 355: I find this part of the design a bit confusing. 
so specifying `Options` can substitute for `Load`? Shouldn't we merge the 2 concepts then?

romoh: Based on offline discussion, let's not auto-load if Options are specified.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -348,33 +350,289 @@ func enableOrDisableServices(services imagecustomizerapi.Services, imageChroot *
 	return nil
 }
 
-func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safechroot.Chroot) error {
+func loadOrDisableModules(modules []imagecustomizerapi.Module, imageChroot *safechroot.Chroot) error {
```
cwize1, line 353: There is now a lot of code for handling kernel modules. It might be worth splitting this off into its own file.

elainezhao96: good point. Will do

toolkit/tools/imagecustomizerapi/systemconfig.go
```diff
@@ -97,8 +97,11 @@ func (s *SystemConfig) IsValid() error {
 		return err
 	}
 
-	if err := s.Modules.IsValid(); err != nil {
-		return err
+	for i, module := range s.Modules {
```
romoh, line 100: should we check that there are no duplicates? or we don't care?

romoh (changes requested):
Update the markdown docs too

toolkit/tools/imagecustomizerapi/module.go
```diff
@@ -5,37 +5,61 @@ package imagecustomizerapi
...
+	}
+
+	if err := validateModuleLoadMode(m.LoadMode); err != nil {
+		return err
 	}
 
+	if len(m.Options) > 0 {
+		for optionKey, optionValue := range m.Options {
+			if optionKey == "" || optionValue == "" {
+				return fmt.Errorf("option key or value cannot be empty for module %s", m.Name)
```
cwize1, line 31: This error message could be misinterpreted as saying that either the key or the value must not be empty. That is, they can't both be empty at the same time. It might be better to give the key and the value their own error message.

Same thing for the spaces/newlines error message below.

toolkit/tools/imagecustomizerapi/module.go
```diff
@@ -5,37 +5,61 @@ package imagecustomizerapi
...
 	"fmt"
+	"strings"
 )
 
+type ModuleOptions map[string]string
+
 type Module struct {
-	Name string `yaml:"Name"`
+	Name     string            `yaml:"Name"`
+	LoadMode string            `yaml:"LoadMode"`
```
cwize1, line 15: Give `LoadMode` a string enum. (e.g. [IdType](https://github.com/microsoft/azurelinux/blob/3.0-dev/toolkit/tools/imagecustomizerapi/idtype.go))

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -348,33 +350,289 @@ func enableOrDisableServices(services imagecustomizerapi.Services, imageChroot *
...
+		// Append the content to the file.
+		if _, err := file.WriteString(content); err != nil {
+			return fmt.Errorf("failed to write module options to %s: %w", moduleOptionsFilePath, err)
+		}
+	}
+
+	return nil
+}
+
+func ensureModulesLoaded(moduleNames []string, moduleLoadFilePath string) error {
```
cwize1, line 458: You could probably write unit-tests for some of these helper functions.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -348,33 +350,289 @@ func enableOrDisableServices(services imagecustomizerapi.Services, imageChroot *
...
+
+	file, err := os.Open(moduleDisableFilePath)
+	if err != nil {
+		return false, err
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		if strings.TrimSpace(scanner.Text()) == moduleName {
```
cwize1, line 541: Would the line have the `blacklist ` prefix before the module name?

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -348,33 +350,289 @@ func enableOrDisableServices(services imagecustomizerapi.Services, imageChroot *
...
+		// If the file does not exist, initialize content as empty
+		content = []byte{}
+	}
+
+	contentString := string(content)
+	needUpdate := false
+
+	for _, moduleName := range moduleNames {
+		blacklistEntry := "blacklist " + moduleName
+		if !strings.Contains(string(content), blacklistEntry+"\n") {
```
cwize1, line 504: It seems odd that you have 2 different ways of checking if a module is disabled: this `if` check, and the code in `isModuleDisabled`.

Also, this code seems simpler than the code in `isModuleDisabled`.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -348,33 +350,289 @@ func enableOrDisableServices(services imagecustomizerapi.Services, imageChroot *
...
+
+	// File exists, if the option exists in the file, directly update the file, otherwise, append it and return the options
+	file, err := os.Open(moduleOptionsFilePath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
```
cwize1, line 601: The files these functions are dealing with will be pretty small. So, I don't think it is necessary to use a scanner pattern here. I would just read in the entire file as a string, make the changes you want, and then write out the whole file.

Similar comment for `isModuleDisabled()`.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -348,33 +350,289 @@ func enableOrDisableServices(services imagecustomizerapi.Services, imageChroot *
...
+					// Keep other options as they are
+					updatedLine = append(updatedLine, field)
+				}
+			}
+			line = strings.Join(updatedLine, " ")
+			found = true
+		}
+		updatedLines = append(updatedLines, line)
+		// Directly write back to file as this is updating existing option
+		os.WriteFile(moduleOptionsFilePath, []byte(strings.Join(updatedLines, "\n")), 0644)
```
cwize1, line 621: Check error code.

cwize1, line 621: Why are you writing the file out on each iteration of the `for` loop?

cwize1 (changes requested):

PR #6299, opened 2023-09-27
feat: Allow new builds of Mariner to generate grub.cfg using grub2-mkconfig using imageconfig EnableGrubMkconfig

Camelron:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 

What does the PR accomplish, why was it needed?
This change allows new builds of Mariner to generate grub.cfg using grub2-mkconfig using EnableGrubMkconfig
- Introduce new image config which enables mkconfig by default.

###### Change Log  

- Allow new builds of Mariner to generate grub.cfg using grub2-mkconfig using KernelCommandLine.EnableGrubMkconfig
- Introduce new image config which enables mkconfig by default.
- Add imagecfg validation testing to ensure that images that enable mkconfig-boot by default include the needed utilities from grub2.rpm
- set $prefix in ESP partition grub.cfg which is analogous to $bootprefix in legacy boot cases; ensures grubenv is properly sourced by mkconfig-based grub.cfgs

###### Does this affect the toolchain?  

NO

###### Test Methodology

- locally; booted newly built image, observe that it boots/works the same, that grub.cfg is generated by mkconfig. 
[...]

PR #6497, opened 2023-10-19
fix: Package both grub2 configuration dirs in grub2-configuration.rpm

Camelron:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

###### Change Log  

- Package both /etc/grub.d and /etc/default/grub.d in grub2-configuration

NO

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Local build, with other related mkconfig changes cherry-picked into cameronbaird/3.0-mkconfig-integration


SPECS/grub2/grub2.spec
```diff
@@ -148,6 +148,7 @@ the grub2-mkconfig flow on AzureLinux
 %package configuration
 Summary:        Location for local grub configurations
 Group:          System Environment/Base
+Requires:       %{name} >= 2.06-12
```
dmcilvaney, line 151: Is this backwards? This means that if we install the conf files we must have full grub right?

christopherco (changes requested):
Requesting to retarget to `3.0-dev`

SPECS/grub2/grub2.spec
Camelron: This becomes a non-issue now that we are targeting 3.0. All 3.0 images will ship with >= grub-2.06-12

Camelron: Actually, I think we still want the reverse of this (requires from grub -> grub-config) to block users from downgrading to a version of grub which packages the config dirs the old way.  That way, 3.0 is resilient to the unexpected case where they downgrade grub.

Camelron:
Now targetting 3.0-dev.

dmcilvaney (approved):

SPECS/grub2/grub2.spec
```diff
@@ -60,6 +60,7 @@ BuildRequires:  systemd-devel
 BuildRequires:  xz-devel
 Requires:       device-mapper
 Requires:       xz
+Requires:       %{name}-configuration >= 2.06-12
```
ddstreetmicrosoft, line 63: when did this -configuration package get added? i must have missed it...i don't see the point of having a separate -configuration package for the config files?

i guess we should keep it now for m2 (let's please drop it in azl3?), but for same-package deps like this you generally want to do a 1:1 requires, i.e. `Requires: %{name}-configuration = %{version}-%{release}`

ddstreetmicrosoft (changes requested):

SPECS/grub2/grub2.spec
Camelron: Because the grub2 package is "large" from M2's perspective. If you want to be able to install grub-configuring packages (kexec-tools) on a non-mkconfig system and then later have grub2-mkconfig work out of the box, you need these directories but not the tools. That is why we split grub2-configuration out.

Camelron: We could remove grub2-configuration for azl3 and enforce that images that are not mkconfig-based out-of-the-box will stay that way. Otherwise, I think we still need configuration and this change is relevant.

Camelron: https://github.com/microsoft/CBL-Mariner/pull/6299#discussion_r1340401013

SPECS/grub2/grub2.spec
dmcilvaney: Sorry for the confusion, I meant he entire requires clause seems backwards, not just the version. Shouldn't `grub2` have the requires on `grub2-configuration`?

Camelron: Yes, see current change: https://github.com/microsoft/CBL-Mariner/pull/6497/files#diff-9dd7fef1e850ad265fea94cbe5a86e163172ee833f8ec954d946d57b0ca44ce4R63

SPECS/grub2/grub2.spec
Camelron: == %{version}

ddstreetmicrosoft (approved):
needs rebase, otherwise LGTM

dmcilvaney (approved):

PR #6586, opened 2023-10-24
Disable TestReferenceDOTFile() until fix is found

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

On some systems `TestReferenceDOTFile()` has been failing. The graph serializes into a different set of base64 encodings. I believe the `gob` library has started behaving differently on some systems. The `gob` library does not appear to guarantee identical outputs as it is a stateful protocol. The resulting serialized file is still valid, and encodes the same graph. 

Until we better understand the issue we should disable the test.

###### Change Log  

- Skip `TestReferenceDOTFile` on short tests.

**NO**

### DIFF
--- a/toolkit/tools/internal/pkggraph/pkggraph_test.go
+++ b/toolkit/tools/internal/pkggraph/pkggraph_test.go
@@ -949,6 +949,10 @@ func TestReadWriteGraph(t *testing.T) {
 
 // Validate the reference graph is valid, and that it matches the output of the test graph.
 func TestReferenceDOTFile(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Short mode enabled")
+	}
+
 	gIn, err := ReadDOTGraphFile("test_graph_reference.dot")
 	assert.NoError(t, err)
 


toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -949,6 +949,13 @@ func TestReadWriteGraph(t *testing.T) {
 
 // Validate the reference graph is valid, and that it matches the output of the test graph.
 func TestReferenceDOTFile(t *testing.T) {
+	if testing.Short() {
+		// Encoding is unreliable on some systems. The final output is still a valid
+		// graph but the base64 encoding of the nodes is different. The final graph once
+		// decoded is the same however (probably gob encoding is different since its stateful?)
```
reubeno, line 955: If the point of this serialization is to produce an opaque blob we can deserialize back and get semantically equivalent results -- then shouldn't the test just round-trip the graphs and compare results instead of doing anything with binary representation? Sounds like an overly constrained test contract.

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -949,6 +949,13 @@ func TestReadWriteGraph(t *testing.T) {
...
+		t.Skip("Short mode enabled")
```
reubeno, line 956: If this test is plain-old unreliable, then is there a reason to skip it _only_ in "short mode"?

anphel31 (approved):

toolkit/tools/internal/pkggraph/pkggraph_test.go
dmcilvaney: I want to turn it back on ASAP as soon as I understand why its not working. The build targets use the short tests so this stops breaking stuff without removing the tests.

toolkit/tools/internal/pkggraph/pkggraph_test.go
dmcilvaney: I'm leaning towards this as well, but this has been reliable for ~4 years so I'm curious why it has changed all of a sudden.

PR #6594, opened 2023-10-24
 Image Customizer: Add MIC release file

elainezhao96:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
Add MIC release file

NO

###### Test Methodology
This is to collect early feedback, will do a test once comments are resolved.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"time"
 
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/imagecustomizerapi"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/imagegen/installutils"
@@ -65,6 +66,11 @@ func doCustomizations(buildDir string, baseConfigPath string, config *imagecusto
 	}
 
 	err = loadOrDisableModules(config.SystemConfig.Modules, imageChroot)
+	if err != nil {
+		return err
+	}
+
+	err = addCustomizerRelease(imageChroot)
 	if err != nil {
 		return err
 	}
@@ -357,6 +363,27 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
 		if err != nil {
 			return fmt.Errorf("failed to write module disable configuration: %w", err)
 		}
+	}
+
+	return nil
+}
+
+func addCustomizerRelease(imageChroot *safechroot.Chroot) error {
+	var err error
+
+	logger.Log.Infof("Creating image customizer release file")
+	micVersion := "0.0.0"
+	currentTime := time.Now().Format("2023-10-24T23:30:45Z")
+
+	customizerreleaseFilePath := filepath.Join(imageChroot.RootDir(), "/etc/mariner-customizer-release")
+	err = file.Write(fmt.Sprintf("%s=\"%s\"", "VERSION", micVersion), customizerreleaseFilePath)
+	if err != nil {
+		return fmt.Errorf("error writing version to '%s': %w", customizerreleaseFilePath, err)
+	}
+
+	err = file.Write(fmt.Sprintf("%s=\"%s\"", "BUILD_DATE", currentTime), customizerreleaseFilePath)
+	if err != nil {
+		return fmt.Errorf("error writing build date to '%s': %w", customizerreleaseFilePath, err)
 	}
 
 	return nil


toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -361,3 +367,24 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
...
 }
+
+func addCustomizerRelease(imageChroot *safechroot.Chroot) error {
+	var err error
+
+	logger.Log.Infof("Creating image customizer release file")
+	micVersion := "0.0.0"
+	currentTime := time.Now().Format("2023-10-24T23:30:45Z")
+
+	customizerreleaseFilePath := filepath.Join(imageChroot.RootDir(), "/etc/mariner-customizer-release")
```
cwize1, line 378: nit: `customizerReleaseFilePath`

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -361,3 +367,24 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
...
+	err = file.Write(fmt.Sprintf("%s=\"%s\"", "VERSION", micVersion), customizerreleaseFilePath)
+	if err != nil {
+		return fmt.Errorf("error writing version to '%s': %w", customizerreleaseFilePath, err)
+	}
+
+	err = file.Write(fmt.Sprintf("%s=\"%s\"", "BUILD_DATE", currentTime), customizerreleaseFilePath)
```
cwize1, line 384: I believe calling `file.Write` overwrites any existing file. So, calling `file.Write` twice won't work.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -361,3 +367,24 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
 
 	return nil
 }
+
+func addCustomizerRelease(imageChroot *safechroot.Chroot) error {
+	var err error
+
+	logger.Log.Infof("Creating image customizer release file")
+	micVersion := "0.0.0"
+	currentTime := time.Now().Format("2023-10-24T23:30:45Z")
```
cwize1, line 376: I don't think you can use any date for the format string. You have a specific date/time that the `Format` function is expecting.

From the Golang [docs](https://pkg.go.dev/time#Time.Format):

```golang
// The layout string used by the Parse function and Format method
// shows by example how the reference time should be represented.
// We stress that one must show how the reference time is formatted,
// not a time of the user's choosing. Thus each layout string is a
// representation of the time stamp,
//	Jan 2 15:04:05 2006 MST
// An easy way to remember this value is that it holds, when presented
// in this order, the values (lined up with the elements above):
//	  1 2  3  4  5    6  -7
```

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -361,3 +367,24 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
 
 	return nil
 }
+
+func addCustomizerRelease(imageChroot *safechroot.Chroot) error {
+	var err error
+
+	logger.Log.Infof("Creating image customizer release file")
+	micVersion := "0.0.0"
```
romoh, line 375: nit - maybe add a comment or TODO

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
romoh: It makes sense to aggregate and dump things in file once.

elainezhao96: sorry overlooked this, will append them

romoh (approved):

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -361,3 +367,27 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
...
+	customizerReleaseFilePath := filepath.Join(imageChroot.RootDir(), "/etc/mariner-customizer-release")
+	lines := []string{
+		fmt.Sprintf("%s=\"%s\"", "VERSION", micVersion),
+		fmt.Sprintf("%s=\"%s\"", "BUILD_DATE", currentTime),
+		"",
+	}
+
+	err = file.WriteLines(lines, customizerReleaseFilePath)
+	if err != nil {
+		return fmt.Errorf("error writing data to '%s': %w", customizerReleaseFilePath, err)
```
cwize1, line 389: nit: `"error writing customizer release file (%s): %w"`

In the rest of the toolkit code, it seems error strings typically put values in parenthesis instead of quotes.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -361,3 +367,27 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
...
+
+	logger.Log.Infof("Creating image customizer release file")
+
+	// TODO: update micVersion once the mic release process is in place
+	micVersion := "0.0.0"
+	currentTime := time.Now().Format("2006-01-02T15:04:05Z")
+
+	customizerReleaseFilePath := filepath.Join(imageChroot.RootDir(), "/etc/mariner-customizer-release")
+	lines := []string{
+		fmt.Sprintf("%s=\"%s\"", "VERSION", micVersion),
```
cwize1, line 382: Probably should call this `TOOL_VERSION` to avoid it being confused for an image version.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -361,3 +367,27 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
 
 	return nil
 }
+
+func addCustomizerRelease(imageChroot *safechroot.Chroot) error {
```
cwize1, line 371: This function seems simple enough that we can add UTs for it.

Check out `TestUpdateHostname` in `customizeutils_test.go`.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -361,3 +367,27 @@ func loadOrDisableModules(modules imagecustomizerapi.Modules, imageChroot *safec
...
+	var err error
+
+	logger.Log.Infof("Creating image customizer release file")
+
+	// TODO: update micVersion once the mic release process is in place
+	micVersion := "0.0.0"
+	currentTime := time.Now().Format("2006-01-02T15:04:05Z")
```
cwize1, line 378: It might be goodness to retrieve the date/time at the start of the customization process and pass it down to the functions that need it. That way, if another component needs this value, then the values won't disagree because `time.Now()` was called by multiple functions.

It will also make it easier to write the UT (as per my other comment) since you can pass a fixed time value into this function.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
elainezhao96: moved time.Now() to the beginning of the doCustomizations function, let me know if it doesnt make sense

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -34,6 +35,8 @@ func doCustomizations(buildDir string, baseConfigPath string, config *imagecusto
 	// Note: The ordering of the customization steps here should try to mirror the order of the equivalent steps in imager
 	// tool as closely as possible.
 
+	buildTime := time.Now().Format("2006-01-02T15:04:05Z")
```
cwize1, line 38: nit: Only retrieve `time.Now()` here. Let `addCustomizerRelease` do the string formatting.

cwize1 (approved):

PR #6623, opened 2023-10-27
Add single transaction for image package cloner

dmcilvaney:
###### Merge Checklist  

- [x ] Any updated packages successfully build (or no packages were changed)
- [ ] Ready to merge

---

###### Summary 

There can be situations where we end up with conflicting package versions. Consider a config file with two packages listed (either directly or in a package list). `pkgA` which depends on a specific version of `dep`, and `pkgB` which also depends on a specific version of `dep`. Lets say that `dep-v1-1` is available in the toolchain, but we published `dep-v1-2` via a fasttrack publish. `pkgA` is also available in the toolchain.

In the current implementation we clone each package one at a time, prioritizing chroot, then toolchain, then local packages, then PMC. If we clone `pkgA`, we will try the toolchain before PMC, and successfully clone `pkgA` and `dep-v1-1`. If we then clone `pkgB` we will resort to PMC since it's not in the toolchain and get the newer `dep-v1-2`.

Fast forward to install time and we install each package in order. `pkgA` installs, and since it has a specific version requirement it will install `dev-v1-1`. We then install `pkgB`, but it will try to install `dev-v1-2` which conflicts. 

In hindsight we should have either picked an older version of `pkgB`, or downloaded a newer version of `pkgA`, so they could both use the same `dev` dependency.

Attempt to clone all packages for an image in a single transaction with a new function `CloneTransaction()`, falling back to the old flow if it fails for any reason (OOM for example).

###### Change Log  

- Add CloneTransaction() to the cloner interface which attempts to clone all packages as a single transaction.

**NO**

###### Test Methodology

- Pipeline build id: local test only

### DIFF
--- a/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
+++ b/toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
@@ -418,7 +418,7 @@ func downloadSingleDeltaRPM(realDependencyGraph *pkggraph.PkgGraph, buildNode *p
 	// already have it in the cache.
 	if !foundCacheRPM {
 		// Avoid any processing since we know the exact RPM we want to download
-		_, err = cloner.CloneRawPackageNames(downloadDependencies, false /*no transaction*/, fullyQualifiedRpmName)
+		_, err = cloner.CloneByName(downloadDependencies, fullyQualifiedRpmName)
 		if err != nil {
 			logger.Log.Warnf("Can't find delta RPM to download for %s: %s (local copy may be newer than published version)", fullyQualifiedRpmName, err)
 			return nil
@@ -480,7 +480,7 @@ func resolveSingleNode(cloner *rpmrepocloner.RpmRepoCloner, node *pkggraph.PkgNo
 				Name: resolvedPackage,
 			}
 
-			preBuilt, err = cloner.Clone(cloneDeps, desiredPackage)
+			preBuilt, err = cloner.CloneByPackageVer(cloneDeps, desiredPackage)
 			if err != nil {
 				err = fmt.Errorf("failed to clone '%s' from RPM repo:\n%w", resolvedPackage, err)
 				return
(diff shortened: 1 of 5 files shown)


jslobodzian (approved):

anphel31 (approved):

toolkit/tools/graphpkgfetcher/graphpkgfetcher.go
```diff
@@ -418,7 +418,7 @@ func downloadSingleDeltaRPM(realDependencyGraph *pkggraph.PkgGraph, buildNode *p
 	// already have it in the cache.
 	if !foundCacheRPM {
 		// Avoid any processing since we know the exact RPM we want to download
-		_, err = cloner.CloneRawPackageNames(downloadDependencies, fullyQualifiedRpmName)
+		_, err = cloner.CloneRawPackageNames(downloadDependencies, false /*no transaction*/, fullyQualifiedRpmName)
```
PawelWMS, line 421: Could we use the well-named constants approach we have elsewhere instead of adding a new approach of in-line comments? I'd prefer we stick to one convention.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -363,11 +377,22 @@ func (r *RpmRepoCloner) CloneRawPackageNames(cloneDeps bool, rawPackageNames ...
...
+		for _, packageName := range rawPackageNames {
+			transactions = append(transactions, []string{packageName})
+		}
+	}
+
 	allPackagesPrebuilt = true
-	for _, packageNameToClone := range rawPackageNames {
-		logger.Log.Debugf("Cloning raw name (%s).", packageNameToClone)
+	for _, packageNamesToClone := range transactions {
+		logger.Log.Debugf("Cloning raw names (%v).", packageNamesToClone)
```
PawelWMS, line 393: Since we always trickle down to this function, I'm curious if we could replace "raw names" with a simple "packages". The "raw" part may not be super clear to everyone.

PawelWMS (approved):

PR #6640, opened 2023-10-31
Pass toolchain archive to make  in pipeline template

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

The toolkit should generally be called with a static set of inputs, we are using `TOOLCHAIN_ARCHIVE=./path/to/toolchain.tar.gz` when we call `make toolchain`. We should continue to use the same input for future commands. Generally we won't re-create the toolchain .rpms since nothing will cause a rebuild, but we should be defensive. If something DOES change, without setting `TOOLCHAIN_ARCHIVE` the toolkit will assume it should download the .rpms. This may fail if one of the toolchain packages has been bumped and was rebuild in the previous stage.

If we pass `TOOLCHAIN_ARCHIVE` to all commnads that use the toolchain the toolkit will know to extract antying that is missing from the archive instead of trying to download a missing file from PMC.

###### Change Log  

- Pass `TOOLCHAIN_ARCHIVE` to the `"Build packages"` stage of `PackageBuild.yml`.

**NO**

###### Associated issues  

- https://dev.azure.com/mariner-org/mariner/_git/CBL-Mariner-Pipelines/pullrequest/16818

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=443426&view=results

### DIFF
--- a/.pipelines/templates/PackageBuild.yml
+++ b/.pipelines/templates/PackageBuild.yml
@@ -160,6 +160,17 @@ steps:
         displayName: "Populate RPMs cache"
 
   - script: |
+      # Make sure toolchain_archive is empty
+      if [[ -n "$toolchain_archive" ]]; then
+        echo "ERROR: toolchain_archive is not empty!" >&2
+        exit 1
+      fi
+      toolchain_archive="$(find "$(Pipeline.Workspace)" -name "${{ parameters.customToolchainTarballName }}" -print -quit)"
+      if [[ ! -f "$toolchain_archive" ]]; then
+        echo "ERROR: toolchain archive not found!" >&2
+        exit 1
+      fi
+
       if [[ ${{ parameters.isDeltaBuild }} == "true" ]]; then
         delta_fetch_arg="DELTA_FETCH=y"
       elif [[ ${{ parameters.isDeltaBuild }} == "false" ]]; then
@@ -185,6 +196,7 @@ steps:
       fi
 
       sudo make -C "${{ parameters.buildRepoRoot }}/toolkit" build-packages -j$(nproc) \
+        TOOLCHAIN_ARCHIVE="$toolchain_archive" \
         CONCURRENT_PACKAGE_BUILDS=${{ parameters.concurrentPackageBuilds }} \
         CONFIG_FILE="" \
         REBUILD_TOOLS=y \
--- a/.pipelines/templates/PackageBuild.yml
+++ b/.pipelines/templates/PackageBuild.yml
@@ -164,8 +164,11 @@ steps:
       if [[ -n "$toolchain_archive" ]]; then
         echo "ERROR: toolchain_archive is not empty!" >&2
         exit 1
+      else
+        echo "toolchain_archive is empty"
       fi
       toolchain_archive="$(find "$(Pipeline.Workspace)" -name "${{ parameters.customToolchainTarballName }}" -print -quit)"
+      echo "toolchain_archive: $toolchain_archive"
       if [[ ! -f "$toolchain_archive" ]]; then
         echo "ERROR: toolchain archive not found!" >&2
         exit 1
--- a/.pipelines/templates/PackageBuild.yml
+++ b/.pipelines/templates/PackageBuild.yml
@@ -160,15 +160,7 @@ steps:
         displayName: "Populate RPMs cache"
 
   - script: |
-      # Make sure toolchain_archive is empty
-      if [[ -n "$toolchain_archive" ]]; then
-        echo "ERROR: toolchain_archive is not empty!" >&2
-        exit 1
-      else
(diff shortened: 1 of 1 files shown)


.pipelines/templates/PackageBuild.yml
```diff
@@ -160,6 +160,12 @@ steps:
         displayName: "Populate RPMs cache"
 
   - script: |
+      toolchain_archive="$(find "$(Pipeline.Workspace)" -name "${{ parameters.customToolchainTarballName }}" -print -quit)"
```
anphel31, line 163: What if the toolchain archive is outside of the workspace, but the provided TOOLCHAIN_ARCHIVE still points to it?

dmcilvaney: We use the same logic above in "Populate toolchain" (line 128). I'm assuming that one requirement for this template is a valid toolchain, either built or provided via artifact. The `TOOLCHAIN_ARCHIVE` value itself doesn't care where the file is, it just reads the .tar.gz and pulls the files it needs from it.

.pipelines/templates/PackageBuild.yml
```diff
@@ -160,6 +160,12 @@ steps:
...
+      if [[ ! -f "$toolchain_archive" ]]; then
```
PawelWMS, line 164: There is no guarantee the user will pass any custom toolchain archives, so we can't expect it to be present.

.pipelines/templates/PackageBuild.yml
```diff
@@ -185,6 +187,7 @@ steps:
       fi
 
       sudo make -C "${{ parameters.buildRepoRoot }}/toolkit" build-packages -j$(nproc) \
+        TOOLCHAIN_ARCHIVE="$(toolchainArchive)" \
```
cwize1, line 190: If `parameters.customToolchainArtifactName` is false, then `toolchainArchive` won't be set to anything. If the pipeline can't find a variable, then it doesn't modify the string. So, `TOOLCHAIN_ARCHIVE` would be set to literally `$(toolchainArchive)`, which would probably confuse `make`.

PawelWMS: You're right, I completely missed that and though non-set pipeline variables are evaluated to an empty string. @dmcilvaney, in that case adding a `- ${{ else }}:` part, where we set `toolchainArchive` to an empty string should help.

PawelWMS: Alternatively, I'd also support doing something similar we do for a few other arguments:
```
if [[ -n ${{ parameters.customToolchainArtifactName }} == "true" ]]; then
        toolchain_archive_arg="TOOLCHAIN_ARCHIVE=$(toolchainArchive)"
fi
(...)
sudo make -C "${{ parameters.buildRepoRoot }}/toolkit" build-packages -j$(nproc) \
(...)
$run_check_arg \
$toolchain_archive_arg \
```

dmcilvaney:
Will look into suggestions, hadn't had a chance to run a test yet so I'll update and then run a build.

.pipelines/templates/PackageBuild.yml
dmcilvaney: Ok, will try the arg variable since that is the pattern we are already using.

.pipelines/templates/PackageBuild.yml
```diff
@@ -194,7 +199,8 @@ steps:
         $delta_fetch_arg \
         $quick_rebuild_packages_arg \
         $run_check_arg \
-        $use_ccache_arg
+        $use_ccache_arg \
+        $toolchain_archive_arg
```
PawelWMS, line 203: Nit-pick request to keep these in alphabetical order.

```suggestion
        $toolchain_archive_arg \
        $use_ccache_arg

```

PawelWMS (approved):

.pipelines/templates/PackageBuild.yml
```diff
@@ -184,6 +185,10 @@ steps:
         use_ccache_arg="USE_CCACHE=n"
       fi
 
+      if [[ -n ${{ parameters.customToolchainArtifactName }} ]]; then
```
dmcilvaney, line 188:
```suggestion
      if [[ -n "${{ parameters.customToolchainArtifactName }}" ]]; then
```

.pipelines/templates/PackageBuild.yml
PawelWMS: The workspace is the download location for the pipeline artifact and the `customToolchainTarballName` parameter is only telling the template what is the name of the toolchain tarball in that artifact. In other words: the template is hard-coded to expect the toolchain to be passed to it through a pipeline artifact and the build will fail of it's not there.

PR #6649, opened 2023-11-01
Sort, reorder and color build summary output

neha170:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Sort summary, reorder summary, add option to color output on terminal

###### Change Log  

- Add new make option LOG_COLOR with options auto (default), always, never
- LOG_COLOR = always: Color both terminal output and logs
- LOG_COLOR = auto: Color only terminal output but not the logs
- LOG_COLOR = never: Disable color in both log and terminal output
- Remove color codes from logs when LOG_COLOR is not always
- Print SRPMs in build summary in alphabetical order
- Reorder summary (print failed, built and prebuilt ones at the end)
- Add new color package import for golang

**NO**

###### Associated issues  

- Some initial discussion can be found here [#46098944](https://microsoft.visualstudio.com/OS/_workitems/edit/46098944)

###### Test Methodology

- Locally built and confirmed the changes in build summary

Summary( before without sort/color):
image

Sorted and colored summary:
image

Logs are non-colored when option is not LOG_COLOR=always:
image

On dark background
image


toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -129,89 +130,104 @@ func PrintBuildSummary(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMutex, bu
...
+		for _, testedSRPM := range keys {
+			logger.Log.Infof("--> %s", filepath.Base(testedSRPM))
 		}
 	}
 
 	if len(prebuiltDeltaSRPMs) != 0 {
 		logger.Log.Info("Skipped SRPMs (i.e., delta mode is on, packages are already available in a repo):")
-		for srpm := range prebuiltDeltaSRPMs {
-			logger.Log.Infof("--> %s", filepath.Base(srpm))
+		keys := getSortedKeys(prebuiltDeltaSRPMs)
```
dmcilvaney, line 143: For order, I'd say delta and skipped are the least important (I'm thinking least important goes at the top?), tested RPMs are probably of similar importance to built rpms...

To avoid duplication, would we want to merge built/tested? Maybe have `built only`, `tested only`, built and tested`?

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -129,89 +130,104 @@ func PrintBuildSummary(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMutex, bu
...
+	if len(failedSRPMsTests) != 0 {
+		logger.Log.Info("Failed SRPMs tests:")
+		keys := getSortedKeys(failedSRPMsTests)
+		for _, key := range keys {
+			failure := failedSRPMsTests[key]
+			logger.Log.Infof("--> %s , error: %s, for details see: %s", failure.Node.SRPMFileName(), failure.Err, failure.LogFile)
+		}
+	}
+
+	if len(prebuiltSRPMs) != 0 {
```
dmcilvaney, line 215: As above, I'd leave prebuilt near the top, generally not super important to see (and there are often a lot of them for delta builds)

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -129,89 +130,104 @@ func PrintBuildSummary(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMutex, bu
...
+	}
+
+	if len(builtSRPMs) != 0 {
+		logger.Log.Info("Built SRPMs:")
+		keys := getSortedKeys(builtSRPMs)
+		for _, builtSRPM := range keys {
+			logger.Log.Infof("--> %s", filepath.Base(builtSRPM))
+		}
+	}
+
```
dmcilvaney, line 230: Could we split the summary, (or maybe duplicate it?) here? It can be a pain to scroll all the way up to see if there is anything unexpected when you can just see "build x#, failed 0" and be happy.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -109,108 +110,142 @@ func PrintBuildSummary(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMutex, bu
...
+	logger.Log.Info("---------------------------")
+	logger.Log.Info("--------- Summary ---------")
+	logger.Log.Info("---------------------------")
 
-	if len(srpmConflicts) != 0 {
-		conflictsLogger("SRPM conflicts with toolchain: ")
-		for _, conflict := range srpmConflicts {
-			conflictsLogger("--> %s", conflict)
-		}
+	logger.Log.Infof("\033[32mNumber of built SRPMs:             %d\033[33m", len(builtSRPMs))
```
dmcilvaney, line 235: We should order these the same way we order the other lists. Maybe we want a dedicated function to print these since we run it twice?

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -334,3 +369,14 @@ func unbuiltPackagesCSVRows(pkgGraph *pkggraph.PkgGraph, unbuiltPackages, failed
 
 	return
 }
+
+// Helper function to return a sorted slice of a map's keys
+// Supports *pkggraph.PkgNode, bool and *BuildResult as map values
+func getSortedKeys[V *pkggraph.PkgNode | bool | *BuildResult](inputMap map[string]V) (keys []string) {
```
dmcilvaney, line 375: We have some functions in `sliceutils`, we already have a function to convert a map to a slice at least. If we want a sort, maybe add it there?
Or, might be enough to just call `sort.Strings(sliceutils.SetToSlice(<set>))`

neha170: The current function in sliceutils was returning a slice of keys, where the value of the key was True (bool). Added another generic function in sliceutils, which returns a slice of all keys.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -334,3 +362,30 @@ func unbuiltPackagesCSVRows(pkgGraph *pkggraph.PkgGraph, unbuiltPackages, failed
 
 	return
 }
+
+// Helper function to print summarized numbers of the build to the logger.
+func PrintSummary(failedSRPMs, failedSRPMsTests map[string]*BuildResult, prebuiltSRPMs, prebuiltDeltaSRPMs, builtSRPMs, testedSRPMs, skippedSRPMsTests, unresolvedDependencies map[string]bool, blockedSRPMs, blockedSRPMsTests map[string]*pkggraph.PkgNode, rpmConflicts, srpmConflicts []string, allowToolchainRebuilds bool, conflictsLogger func(format string, args ...interface{})) {
```
dmcilvaney, line 367:
```suggestion
func printSummary(failedSRPMs, failedSRPMsTests map[string]*BuildResult, prebuiltSRPMs, prebuiltDeltaSRPMs, builtSRPMs, testedSRPMs, skippedSRPMsTests, unresolvedDependencies map[string]bool, blockedSRPMs, blockedSRPMsTests map[string]*pkggraph.PkgNode, rpmConflicts, srpmConflicts []string, allowToolchainRebuilds bool, conflictsLogger func(format string, args ...interface{})) {
```

dmcilvaney (approved):

corvus-callidus:
Is there a way to make the color optional for accessibility purposes?

neha170:
> Is there a way to make the color optional for accessibility purposes?

@https://github.com/dmcilvaney Are we ok adding another make option for choosing to keep color on or off?
I am thinking something like LOG_COLOR=on/off like we have LOG_LEVEL...

toolkit/tools/internal/sliceutils/sliceutils.go
```diff
@@ -72,7 +73,16 @@ func SetToSlice[T comparable](inputSet map[T]bool) []T {
 	return outputSlice[:index]
 }
 
-// SliceToSet converts a slice of K to a map[K]bool.
+// SetToSliceAll converts a map[K]V to a slice containing the map's keys.
+func SetToSliceAll[K comparable, V any](inputSet map[K]V) []K {
```
cwize1, line 77: You could call this function `MapKeys`.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -5,14 +5,36 @@ package schedulerutils
...
 )
 
+const (
+	colorReset = "\033[33m"
+	colorRed   = "\033[31m"
+	colorGreen = "\033[32m"
+	colorBlue  = "\033[34m"
+)
+
+func printInRed(inputString string) string {
```
cwize1, line 26: It would be goodness to have format versions of these functions.

reubeno: Are there any mature golang OSS packages that already provide good support for colorized output? That would be even better!

toolkit/tools/internal/sliceutils/sliceutils.go
```diff
@@ -59,11 +59,12 @@ func PackageVerMatch(expected, given interface{}) bool {
...
 
-// SetToSlice converts a map[T]bool to a slice containing the map's keys.
-func SetToSlice[T comparable](inputSet map[T]bool) []T {
+// SetToSlice converts a map[K]bool to a slice containing the map's keys, iff the key's value is true.
+func SetToSlice[K comparable](inputSet map[K]bool) []K {
 	index := 0
-	outputSlice := make([]T, len(inputSet))
+	outputSlice := make([]K, len(inputSet))
 	for element, elementInSet := range inputSet {
+		// Add key to slice if value is true
```
PawelWMS, line 67: Seems unnecessary - the variable's name is already explicit enough. Too many comments may muddy the code instead of adding clarity. I'd strongly suggest saving them only for cases where the logic is too complex to be explained through clear code or well-named variables.
```suggestion
```

toolkit/tools/internal/sliceutils/sliceutils.go
PawelWMS: I think `SetToSliceAll` or `SetToSliceWithRepetition` is nice, as it makes it clearer that this is an alternative to the plain `SetToSlice`. `MapKeys` makes feels a bit unclear to me. How are we mapping the keys? According to what method?

toolkit/tools/internal/sliceutils/sliceutils_test.go
```diff
@@ -154,3 +154,38 @@ func TestShouldRemoveDuplicates(t *testing.T) {
...
+	}
+	outputSlice := SetToSliceAll(inputSet)
+
+	assert.NotNil(t, outputSlice)
+	assert.Len(t, inputSet, 4)
+	assert.Len(t, outputSlice, 4)
+	assert.Contains(t, outputSlice, "A")
+	assert.Contains(t, outputSlice, "B")
+	assert.Contains(t, outputSlice, "X")
+	assert.Contains(t, outputSlice, "Y")
```
PawelWMS, line 190: You could leave the robust output checking to a single `assert` function:
```suggestion
	expectedOutput := []string{"A", "B", "X", "Y"}
	outputSlice := SetToSliceAll(inputSet)

	assert.ElementsMatch(t, expectedOutput, outputSlice)
```

toolkit/tools/internal/sliceutils/sliceutils_test.go
```diff
@@ -154,3 +154,38 @@ func TestShouldRemoveDuplicates(t *testing.T) {
 	assert.Contains(t, outputSlice, "C")
 	assert.NotContains(t, outputSlice, "X")
 }
+
+// SetToSliceAll() should return empty slice for nil map
+func TestSetToSliceAllShouldCreateEmptySliceFromNil(t *testing.T) {
```
PawelWMS, line 159: Nice test names!

PawelWMS (changes requested):
Let's update the `SetToSliceAll` function's name to make it clear we're allowing duplicates in this case (see: comments near that function).

And let's try to strip the colouring from file logs to keep the easy to read, parse, and search through them. You can try modifying `newWriterHook` inside `toolkit/tools/internal/logger/writerhook.go` so that it cuts out the colouring when `useColors` is `false`.

Ideally, let's also try to find and use existing libraries to help with coloring the output.

Minor comments aside from that.

Thank you for cleaning this up!

toolkit/tools/internal/sliceutils/sliceutils.go
neha170: Yes, I have kept called it `SetToSlice` only to keep it consistent with the rest of sliceutils. We're using `set` instead of `map` throughout the codebase. I'm planning to make another PR to fix all `set(s)` to `map(s)` wherever applicable in the toolkit codebase.

dmcilvaney:
> > Is there a way to make the color optional for accessibility purposes?
> 
> @https://github.com/dmcilvaney Are we ok adding another make option for choosing to keep color on or off? I am thinking something like LOG_COLOR=on/off like we have LOG_LEVEL...

I think `always`, `never`, `auto` is the tuple I've always seen with most linux tools, but the flag name is fine with me.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -352,18 +352,41 @@ func printSummary(failedSRPMs, failedSRPMsTests map[string]*BuildResult, prebuil
...
+		logger.Log.Infof(color.RedString("Number of failed SRPMs:             " + fmt.Sprint(len(failedSRPMs))))
+	} else {
+		logger.Log.Infof(color.GreenString("Number of failed SRPMs:             " + fmt.Sprint(len(failedSRPMs))))
+	}
+
+	if len(failedSRPMsTests) > 0 {
+		logger.Log.Infof(color.RedString("Number of failed SRPMs tests:       " + fmt.Sprint(len(failedSRPMsTests))))
+	} else {
+		logger.Log.Infof(color.GreenString("Number of failed SRPMs tests:       " + fmt.Sprint(len(failedSRPMsTests))))
+	}
```
PawelWMS, line 389: Since it's repeated multiple times, could we extract this logic to a helper function?

toolkit/tools/internal/logger/writerhook.go
```diff
@@ -56,6 +58,13 @@ func (h *writerHook) Fire(entry *logrus.Entry) (err error) {
 		return
 	}
 
+	if h.useColors == false {
```
PawelWMS, line 61: Nit-pick to be consistent with what we do in the rest of our code. At least in most cases:
```suggestion
	if !h.useColors {
```

toolkit/tools/internal/logger/writerhook.go
```diff
@@ -56,6 +58,13 @@ func (h *writerHook) Fire(entry *logrus.Entry) (err error) {
...
+		entry.Message = strings.ReplaceAll(entry.Message, "\x1b[31m", "")
+		entry.Message = strings.ReplaceAll(entry.Message, "\x1b[34m", "")
+		entry.Message = strings.ReplaceAll(entry.Message, "\x1b[0m", "")
+		entry.Message = strings.ReplaceAll(entry.Message, "\x1b[32m", "")
```
PawelWMS, line 65: Let's not hard-code every single colour we use. If we ever add another one, we'll need to keep expanding this section (not to mention someone will even need to know this is here). I'd suggest changing it to a regex instead, which will catch all `\x1b[<digits>m` strings. I think `\\x1b\[[0-9]+m` should work, but please double-check in https://regex101.com/ for instance.

toolkit/tools/internal/logger/writerhook.go
```diff
@@ -31,11 +34,9 @@ func newWriterHook(writer io.Writer, level logrus.Level, useColors bool, toolNam
 
 	if toolName != "" {
 		formatter.CallerPrettyfier = func(frame *runtime.Frame) (function string, file string) {
-			const gray = 90
```
PawelWMS, line 34: Thanks for updating that!

toolkit/tools/internal/logger/writerhook.go
```diff
@@ -31,11 +34,9 @@ func newWriterHook(writer io.Writer, level logrus.Level, useColors bool, toolNam
...
-
 			toolNameField := fmt.Sprintf("[%s]", toolName)
 			if useColors {
-				toolNameField = fmt.Sprintf("[\x1b[%dm%s\x1b[0m]", gray, toolName)
+				toolNameField = fmt.Sprintf(color.BlackString("[%s]"), toolName)
```
PawelWMS, line 39: Won't this blend with a black console background? If you haven't already, can you double-check before we merge. This part is to print the gray tools' names in the brackets btw. Example:
image

toolkit/tools/internal/logger/writerhook.go
neha170: Yes, sorry, should have mentioned, this is still work in progress. I'm planning to make this into a regex

toolkit/tools/internal/logger/writerhook.go
neha170: Oh, yes thanks for pointing this out. I'll try it on a black background, and update color accordingly

toolkit/docs/building/building.md
```diff
@@ -809,6 +809,7 @@ To reproduce an ISO build, run the same make invocation as before, but set:
 | Variable                      | Default                                                                                                | Description
 |:------------------------------|:-------------------------------------------------------------------------------------------------------|:---
 | LOG_LEVEL                     | info                                                                                                   | Console log level for go tools (`panic, fatal, error, warn, info, debug, trace`)
+| LOG_COLOR                     | auto                                                                                                   | Console log color for go tools (`always, auto, never`)
```
PawelWMS, line 812: Please also mention what each option means.

toolkit/tools/internal/exe/exe.go
```diff
@@ -40,6 +40,11 @@ func OutputDirFlag(k *kingpin.Application, doc string) *string {
 	return k.Flag("output-dir", doc).Required().String()
 }
 
+// LogColorFlag registers a log color flag for k and returns the passed value
+func LogColorFlag(k *kingpin.Application) *string {
+	return k.Flag(logger.ColorFlag, logger.ColorFlagHelp).String()
```
PawelWMS, line 45: Let's also enum the 3 possible values. See how `LogLevelFlag` does it for instance.

toolkit/tools/internal/logger/log.go
```diff
@@ -179,8 +185,15 @@ func ReplaceStderrFormatter(newFormatter logrus.Formatter) (oldFormatter logrus.
...
-	const useColors = true
+func initStderrLogInternal(callerFilePath string, color string) {
+	var useColors bool
+	if color == "auto" || color == "always" {
+		useColors = true
+	} else if color == "never" {
+		useColors = false
+	} else {
+		useColors = true
+	}
```
PawelWMS, line 196: Could we move `auto`, `always`, and `never` to well-named constants? This way they raw string will be defined in only one place and then referenced with the constant. That makes it harder to create a hard to spot bug with a typo and also any modifications to the level (although unlikely in this case) would be only necessary in one place in the code.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -334,3 +345,39 @@ func unbuiltPackagesCSVRows(pkgGraph *pkggraph.PkgGraph, unbuiltPackages, failed
...
+	logger.Log.Infof(color.GreenString("Number of prebuilt SRPMs:           " + fmt.Sprint(len(prebuiltSRPMs))))
+	logger.Log.Infof(color.BlueString("Number of prebuilt delta SRPMs:     " + fmt.Sprint(len(prebuiltDeltaSRPMs))))
+	logger.Log.Infof(color.BlueString("Number of skipped SRPMs tests:      " + fmt.Sprint(len(skippedSRPMsTests))))
+	logger.Log.Infof(color.GreenString("Number of built SRPMs:              " + fmt.Sprint(len(builtSRPMs))))
+	logger.Log.Infof(color.GreenString("Number of tested SRPMs:             " + fmt.Sprint(len(testedSRPMs))))
+	printSummaryByLen(len(unresolvedDependencies), "Number of unresolved dependencies:  ")
+	printSummaryByLen(len(blockedSRPMs), "Number of blocked SRPMs:            ")
+	printSummaryByLen(len(blockedSRPMsTests), "Number of blocked SRPMs tests:      ")
+	printSummaryByLen(len(failedSRPMs), "Number of failed SRPMs:             ")
+	printSummaryByLen(len(failedSRPMsTests), "Number of failed SRPMs tests:       ")
```
PawelWMS, line 364: We could simplify further by passing the slice itself and calling `len` inside the function.
```suggestion
	printSummaryByLen(unresolvedDependencies, "Number of unresolved dependencies:  ")
	printSummaryByLen(blockedSRPMs, "Number of blocked SRPMs:            ")
	printSummaryByLen(blockedSRPMsTests, "Number of blocked SRPMs tests:      ")
	printSummaryByLen(failedSRPMs, "Number of failed SRPMs:             ")
	printSummaryByLen(failedSRPMsTests, "Number of failed SRPMs tests:       ")
```

toolkit/tools/internal/exe/exe.go
```diff
@@ -40,6 +40,11 @@ func OutputDirFlag(k *kingpin.Application, doc string) *string {
 	return k.Flag("output-dir", doc).Required().String()
 }
 
+// LogColorFlag registers a log color flag for k and returns the passed value
+func LogColorFlag(k *kingpin.Application) *string {
```
dmcilvaney, line 44: Since we are adding this to every tool, do we want to consider doing something like `SetupProfileFlags()` which just handles logging as a single entity? We need to do 3x calls in every tool now. I'm fine leaving the separate flags if we want,  but maybe we want a single convenience function that generates the struct, and update `logger.InitBestEffort()` to just take that struct as a parameter? Or add a new version.

dmcilvaney: A second function would also allow the changes to all the tools to be bundled into a separate PR.

toolkit/tools/internal/logger/log.go
dmcilvaney: They could also be a type, something like:
```go
type LoggerColourMode string
const (
	ColourModeAuto LoggerColourMode "auto"
	ColourModeAlways LoggerColourMode "always"
	ColourModeNever LoggerColourMode "never"
)
```

toolkit/tools/internal/logger/writerhook.go
dmcilvaney: It looks like both logrus and fatith/color have enable/disable flags for colours, could we just use those somehow? Maybe save some pain here? (I'm not 100% sure, since we have two parallel flows of log data...)

toolkit/tools/internal/logger/writerhook.go
neha170: Changed it to hi-intensity yellow. This is how it looks on white and black background terminals
Also changed blue text to green/red depending on whether they could be treated as success/failure, as blue was difficult to see on black background
image
image

neha170:
Keeping this PR focused on summary (color, sort, reorder) and making a separate PR for changes requested in https://microsoft.visualstudio.com/OS/_workitems/edit/46098944 
PR: https://github.com/microsoft/CBL-Mariner/pull/6768

toolkit/tools/scheduler/schedulerutils/printresults.go
neha170: The type of unresolvedDependencies, etc. are all different. I thought of keeping the helper function generic, so it could be extended in the future and used elsewhere also. I don't see much benefit in passing the SRPM variable, and finding length inside helper

toolkit/tools/internal/logger/writerhook.go
neha170: The disabling of color in both logrus and fatih/color render the output without color, but do not remove color code that has already been applied to a string.
We are experiencing something similar to this, https://github.com/fatih/color/issues/150, and looks like the fatih/color library is not providing a resolution. So, resorting to regex to clean up the string written to logs.
The coloring of log itself is handled by logger.initLogFile() which sets `useColors  = false`
image
v/s
image

toolkit/tools/internal/logger/writerhook.go
```diff
@@ -59,10 +59,8 @@ func (h *writerHook) Fire(entry *logrus.Entry) (err error) {
 	}
 
 	if !h.useColors {
-		entry.Message = strings.ReplaceAll(entry.Message, "\x1b[31m", "")
-		entry.Message = strings.ReplaceAll(entry.Message, "\x1b[34m", "")
-		entry.Message = strings.ReplaceAll(entry.Message, "\x1b[0m", "")
-		entry.Message = strings.ReplaceAll(entry.Message, "\x1b[32m", "")
+		re := regexp.MustCompile(`\x1b\[[0-9]+m`)
```
PawelWMS, line 62: Please move this to a global private variable (see: examples elsewhere where we use `MustCompile`). There's no need to re-compile the regular expression every time we enter this function. Which may be often, since it's fired for every log we take. At least that's my understanding.

toolkit/docs/building/building.md
```diff
@@ -809,6 +809,7 @@ To reproduce an ISO build, run the same make invocation as before, but set:
 | Variable                      | Default                                                                                                | Description
 |:------------------------------|:-------------------------------------------------------------------------------------------------------|:---
 | LOG_LEVEL                     | info                                                                                                   | Console log level for go tools (`panic, fatal, error, warn, info, debug, trace`)
+| LOG_COLOR                     | auto                                                                                                   | Console log color for go tools (`always, auto, never`). `always` and `auto` enable color, `never` disables color in terminal output.
```
PawelWMS, line 812: Please explain what's the difference between `auto` and `always`. Right now it looks like they do the same thing. If that's the case, I think we can drop one and go with a simple `y` and `n`. In other words: let's clarify this option.:)

toolkit/tools/internal/logger/log.go
```diff
@@ -179,8 +209,15 @@ func ReplaceStderrFormatter(newFormatter logrus.Formatter) (oldFormatter logrus.
...
-	const useColors = true
+func initStderrLogInternal(callerFilePath string, color string) {
+	var useColors bool
+	if color == colourModeAuto || color == colourModeAlways {
+		useColors = true
+	} else if color == colourModeNever {
+		useColors = false
+	} else {
+		useColors = true
+	}
```
PawelWMS, line 220: I think we can simplify this logic a bit:
```suggestion
	useColors := true
	if color == colourModeNever {
		useColors = false
	}
```

Also, please pick either "colour" or "colour" - let's be consistent across the code.:)

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -105,113 +110,119 @@ func PrintBuildSummary(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMutex, bu
...
+		for _, prebuiltDeltaSRPM := range keys {
+			logger.Log.Infof("--> %s", filepath.Base(prebuiltDeltaSRPM))
 		}
 	}
 
 	if len(skippedSRPMsTests) != 0 {
-		logger.Log.Info("Skipped SRPMs tests:")
-		for srpm := range skippedSRPMsTests {
-			logger.Log.Infof("--> %s", filepath.Base(srpm))
+		logger.Log.Info(color.RedString("Skipped SRPMs tests:"))
```
PawelWMS, line 135: Why is this red? This is more of a neutral thing - tests are skipped when they didn't need to be run because none of their dependencies changed for instance.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -105,113 +110,119 @@ func PrintBuildSummary(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMutex, bu
...
+		keys := sliceutils.MapToSlice(skippedSRPMsTests)
+		sort.Strings(keys)
```
PawelWMS, line 137: Seeing how often you call `MapToSlice` followed by `sort.Strings`, I'd suggest adding another function to avoid this repetition. For instance `sliceutils.MapToSortedSlice`.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -105,113 +110,119 @@ func PrintBuildSummary(pkgGraph *pkggraph.PkgGraph, graphMutex *sync.RWMutex, bu
...
+		conflictsLogger(color.RedString("RPM conflicts with toolchain:"))
+		sort.Strings(rpmConflicts)
 		for _, conflict := range rpmConflicts {
 			conflictsLogger("--> %s", conflict)
 		}
 	}
 
 	if len(srpmConflicts) != 0 {
-		conflictsLogger("SRPM conflicts with toolchain: ")
+		conflictsLogger(color.RedString("SRPM conflicts with toolchain:"))
```
PawelWMS, line 197: The `conflictsLogger` is already set-up to print the output as an error or as an info log depending on the situation. For cases, where it's only an info (for instance where we allow toolchain rebuilds and then they are an expected occurrence), I think we shouldn't print the output in red. It would be sending the wrong message.

toolkit/tools/scheduler/schedulerutils/printresults.go
PawelWMS: They are all maps and with generics the types of the keys and values shouldn't matter. Having said that, this is more a matter of preference, so no push here.:)

toolkit/tools/internal/logger/log.go
neha170: Thank you for catching that inconsistency

toolkit/tools/scheduler/schedulerutils/printresults.go
neha170: Ok, I changed from blue, because blue was not visible on dark backgrounds. I see, in that case, I'll keep it green always

toolkit/docs/building/building.md
neha170: I chose 'auto, always, never' to keep it consistent with the linux standard.
I can change 'always' to enable color in logs also. Are we ok with coloring the logs? 'auto' and 'never' would remain the same.
The other option is to use y/n.

toolkit/tools/internal/logger/log.go
neha170: I was trying to be intentional about the cases, so any future changes would require attention here. But I can see how it looks messy, will change as you suggested

toolkit/tools/scheduler/schedulerutils/printresults.go
neha170: Thanks for highlighting this. Let me color it green (if allowToolchainRebuilds=y) and red otherwise

toolkit/docs/building/building.md
neha170: Chatted with Daniel offline, going ahead with these options:
'never' to never use colorized output (neither logs nor terminal)
'always' to always use colors (both logs and terminal)
'auto' to color terminal output (not logs, only terminal)

toolkit/docs/building/building.md
```diff
@@ -809,6 +809,7 @@ To reproduce an ISO build, run the same make invocation as before, but set:
 | Variable                      | Default                                                                                                | Description
 |:------------------------------|:-------------------------------------------------------------------------------------------------------|:---
 | LOG_LEVEL                     | info                                                                                                   | Console log level for go tools (`panic, fatal, error, warn, info, debug, trace`)
+| LOG_COLOR                     | auto                                                                                                   | Console log color for go tools (`always, auto, never`). `always` enables color in both logs and terminal output, `auto`(`default` option) enables color in terminal output, and `never` disables color in all.
```
PawelWMS, line 812: Just a few nit-picks after viewing the formatter version:
```suggestion
| LOG_COLOR                     | auto                                                                                                   | Console log color for go tools (`always`, `auto`, `never`). `always` enables color in both logs and terminal output, `auto`(default option) enables color in terminal output, and `never` disables color in all.
```

PawelWMS, line 812: Question: does the code now indeed enables the colouring in logs for the `always` option?

toolkit/tools/internal/logger/writerhook.go
```diff
@@ -18,8 +20,15 @@ type writerHook struct {
 	level     logrus.Level
 	writer    io.Writer
 	formatter logrus.Formatter
+	useColors bool
 }
 
+var (
+
+	// colorCodeRegex is of type '\x1b0m' or '\x1b31m', etc.
```
PawelWMS, line 28: Nit-picks:
```suggestion
	// colorCodeRegex is of type '\x1b[0m' or '\x1b[31m', etc.
```

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -334,3 +335,44 @@ func unbuiltPackagesCSVRows(pkgGraph *pkggraph.PkgGraph, unbuiltPackages, failed
...
+}
+
+// Helper function to print summary in specific color, based on val and cond.
+func printSummaryByVal(cond bool, val int, msg string) {
+	if cond && (val > 0) {
+		logger.Log.Errorf(color.RedString(msg + fmt.Sprint(val)))
+	} else {
+		logger.Log.Infof(color.GreenString(msg + fmt.Sprint(val)))
+	}
+}
```
PawelWMS, line 371: The name of the function and the logic is getting a bit cryptic. My one suggestion to clean it up a little is to turn its signature to `(<condition>, <var_args_for_printing>)`, where:
- `<condition>` is a `bool` detrmining if we print an error or an info. For instance `len(failedSRPMs)` or `!allowToolchainRebuilds && len(rpmConflicts)`.
- `<var_args_for_printing>` is a variable number of arguments, which can be passed to `fmt.Sprintf()`.

If you're curious how to use variable argument lists, `shell.go` has many examples. You can also read more about it online ([this article](https://golangdocs.com/variadic-functions-in-golang) looks interesting for instance).

I'd also then re-name the function to something a bit more descriptive. Maybe `printErrorInfoOnCondition`? I'm not good with names here, but the goal is to give the reader a greater chance at understanding what the function will do from the name.:)

PawelWMS: Actually, I see that `RedString()` already accepts `<format>, <var_args>`, so you can probably skip `fmt.Sprintf`.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -334,3 +335,44 @@ func unbuiltPackagesCSVRows(pkgGraph *pkggraph.PkgGraph, unbuiltPackages, failed
...
 	return
 }
+
+// Helper function to print summarized numbers of the build to the logger.
+func printSummary(failedSRPMs, failedSRPMsTests map[string]*BuildResult, prebuiltSRPMs, prebuiltDeltaSRPMs, builtSRPMs, testedSRPMs, skippedSRPMsTests, unresolvedDependencies map[string]bool, blockedSRPMs, blockedSRPMsTests map[string]*pkggraph.PkgNode, rpmConflicts, srpmConflicts []string, allowToolchainRebuilds bool, conflictsLogger func(format string, args ...interface{})) {
+	logger.Log.Info("---------------------------")
+	logger.Log.Info("--------- Summary ---------")
+	logger.Log.Info("---------------------------")
+
+	logger.Log.Infof(color.GreenString("Number of prebuilt SRPMs:           " + fmt.Sprint(len(prebuiltSRPMs))))
```
PawelWMS, line 345: Can we use the previous string formatting? It looks like `GreenString` (and other colours) should already support something like this:
```suggestion
	logger.Log.Infof(color.GreenString("Number of prebuilt SRPMs:           %d", (len(prebuiltSRPMs)))
```

toolkit/docs/building/building.md
neha170: Yes, this is how the logs look in vscode and in terminal when we 'cat' them, with LOG_COLOR=always
image

toolkit/tools/scheduler/schedulerutils/printresults.go
PawelWMS: Reopening. The current version with an additional `boolToInt` helper function seems like a bit of an overkill. Please re-consider my suggestion above and let me know, if you have questions or objections.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -341,22 +342,22 @@ func printSummary(failedSRPMs, failedSRPMsTests map[string]*BuildResult, prebuil
...
+	logger.Log.Infof(color.GreenString(summaryLine("Number of prebuilt SRPMs:", len(prebuiltSRPMs))))
+	logger.Log.Infof(color.GreenString(summaryLine("Number of prebuilt delta SRPMs:", len(prebuiltDeltaSRPMs))))
+	logger.Log.Infof(color.GreenString(summaryLine("Number of skipped SRPMs tests:", len(skippedSRPMsTests))))
+	logger.Log.Infof(color.GreenString(summaryLine("Number of built SRPMs:", len(builtSRPMs))))
+	logger.Log.Infof(color.GreenString(summaryLine("Number of tested SRPMs:", len(testedSRPMs))))
+	printErrorInfoByCondition(len(unresolvedDependencies)>0, "%s", summaryLine("Number of unresolved dependencies:", len(unresolvedDependencies)))
+	printErrorInfoByCondition(len(blockedSRPMs)>0, "%s", summaryLine("Number of blocked SRPMs:", len(blockedSRPMs)))
+	printErrorInfoByCondition(len(blockedSRPMsTests)>0, "%s", summaryLine("Number of blocked SRPMs tests:", len(blockedSRPMsTests)))
+	printErrorInfoByCondition(len(failedSRPMs)>0, "%s", summaryLine("Number of failed SRPMs:", len(failedSRPMs)))
+	printErrorInfoByCondition(len(failedSRPMsTests)>0, "%s", summaryLine("Number of failed SRPMs tests:", len(failedSRPMsTests)))
```
PawelWMS, line 354: Format string argument `"%s"` should no longer be needed. See how `fmt.Printf` is written for instance, in case things break.

toolkit/tools/scheduler/schedulerutils/printresults.go
```diff
@@ -334,3 +335,48 @@ func unbuiltPackagesCSVRows(pkgGraph *pkggraph.PkgGraph, unbuiltPackages, failed
...
+	if allowToolchainRebuilds && (len(rpmConflicts) > 0 || len(srpmConflicts) > 0) {
+		logger.Log.Infof("Toolchain RPMs conflicts are ignored since ALLOW_TOOLCHAIN_REBUILDS=y")
+	}
+
+	printErrorInfoByCondition(!allowToolchainRebuilds && len(rpmConflicts)>0, "%s", summaryLine("Number of toolchain RPM conflicts:", len(rpmConflicts)))
+	printErrorInfoByCondition(!allowToolchainRebuilds && len(srpmConflicts)>0, "%s", summaryLine("Number of toolchain SRPM conflicts:", len(srpmConflicts)))
+}
+
+// Helper function to print error or info based on condition.
```
PawelWMS, line 363: One thing I noticed right now: could you reformat the comments to follow the [Golang convention of commenting functions](https://go.dev/blog/godoc)? We try to follow that in our code. In here it boils down to starting the comment with the name of the commented function. Example:
```suggestion
// printErrorInfoByCondition prints error or info based on condition.
// If the condition is true, it prints an error level log and an info level one otherwise.
```

Also, try to provide more details for the function. In the example above it was unclear what "based on condition" meant.

PawelWMS (approved):

PR #6651, opened 2023-11-01
Nopatch kernel CVE-2023-2430, CVE-2023-3338, CVE-2023-39191, CVE-2023-42752 ...

Camelron:
…for kernel

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

###### Change Log  

- CVE-2023-2430.nopatch
- CVE-2023-3338.nopatch
- CVE-2023-39191.nopatch
- CVE-2023-42752.nopatch

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2023-2430
- https://nvd.nist.gov/vuln/detail/CVE-2023-3338
- https://nvd.nist.gov/vuln/detail/CVE-2023-39191
- https://nvd.nist.gov/vuln/detail/CVE-2023-42752

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/SPECS/kernel/CVE-2023-2430.nopatch
+++ b/SPECS/kernel/CVE-2023-2430.nopatch
@@ -1,4 +1,4 @@
 CVE-2023-2430 - Introducing commit not present
 Introducing patch not present. Note that 5.15 doesn't even have the vulnerable file https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/io_uring/msg_ring.c?h=v5.15.137
 upstream fix: e12d7a46f65ae4b7d58a5e0c1cbfa825cf8
-But io_uring.c isn't introduced until 36404b09aa609e00f8f0108356830c22b99b3cbf; not backported to 5.15.*
+But msg_ring.c isn't introduced until 36404b09aa609e00f8f0108356830c22b99b3cbf; not backported to 5.15.*
--- a/SPECS/kernel/CVE-2023-2430.nopatch
+++ b/SPECS/kernel/CVE-2023-2430.nopatch
@@ -1,0 +1,4 @@
+CVE-2023-2430 - Introducing commit not present
+Introducing patch not present. Note that 5.15 doesn't even have the vulnerable file https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/io_uring/msg_ring.c?h=v5.15.137
+upstream fix: e12d7a46f65ae4b7d58a5e0c1cbfa825cf8
+But io_uring.c isn't introduced until 36404b09aa609e00f8f0108356830c22b99b3cbf; not backported to 5.15.*
--- a/SPECS/kernel/CVE-2023-39191.nopatch
+++ b/SPECS/kernel/CVE-2023-39191.nopatch
@@ -1,5 +1,4 @@
-CVE-2023-39191 - Not vulnerable;
+CVE-2023-39191 - Not vulnerable; vulnerable code not yet backported to 5.15.*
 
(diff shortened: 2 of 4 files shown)


SPECS/kernel/CVE-2023-39191.nopatch
```diff
@@ -0,0 +1,5 @@
+CVE-2023-39191 - Not vulnerable;
+
+Since we set sysctl.kernel.unprivileged_bpf_disabled=1 for all of our images,
```
Camelron, line 3: @ddstreetmicrosoft Do you mind vetting the logic here? https://nvd.nist.gov/vuln/detail/CVE-2023-39191

1. the CVE may allow an attacker with CAP_BPF privileges to escalate privileges and execute arbitrary code in the context of the kernel
2. In all cases, we ship unprivileged_bpf_disabled=1 meaning users without CAP_SYS_ADMIN can't load ebpf programs
3. Thus, there is no danger of privilege escalation, one needs to be admin already to abuse this. Admin is not part of our thread model. 

Are any of my assumptions wrong?

SPECS/kernel/CVE-2023-2430.nopatch
```diff
@@ -0,0 +1,4 @@
+CVE-2023-2430 - Introducing commit not present
+Introducing patch not present. Note that 5.15 doesn't even have the vulnerable file https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/io_uring/msg_ring.c?h=v5.15.137
+upstream fix: e12d7a46f65ae4b7d58a5e0c1cbfa825cf8
+But io_uring.c isn't introduced until 36404b09aa609e00f8f0108356830c22b99b3cbf; not backported to 5.15.*
```
rlmenge, line 4: Nit: I think you mean msg_ring.c isn't introduced until

Camelron: absolutely, thanks!

SPECS/kernel/CVE-2023-39191.nopatch
christopherco: Are you sure we are actually affected by this CVE? From my investigation, it looks like this vulnerability is in the dynptr feature of BPF which is a newer feature and I do not see in the 5.15 stable trees.

christopherco (changes requested):

SPECS/kernel/CVE-2023-39191.nopatch
ddstreetmicrosoft: > In all cases, we ship unprivileged_bpf_disabled=1

This is only our default boot cmdline though; in general, we should still fix bugs even if our default cmdline settings would mitigate the issue, as end users could remove this cmdline setting.

But for this particular case, I think @christopherco is correct that it doesn't affect our older kernel because the dynptr feature that introduced the bug was added later (it looks like around 5.19 I think)

christopherco (approved):

SPECS/kernel/CVE-2023-39191.nopatch
Camelron: This reasoning makes sense, thanks! 

I have independently verified that the introducing commit is not present in the 5.15 tree https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v5.15.137&id=97e03f521050c092919591e668107b3d69c5f426

The changes to add dynptrs are indeed not present, for example, in the header file: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/bpf.h?h=v5.15.137

PR #6682, opened 2023-11-06
Use embedded binary resources for grub templates.

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Move the grub template files from files that sit beside the imager binary to instead be embedded in the image binary.

This is being done for 2 reasons:

1. For Image Customizer, I don't want these template files to be modifable by the user, as this will make supporting back-compat more difficult.

2. It makes it easier to test the Go tools using `go run`, since the location of these files doesn't have to be provided.

###### Change Log  

- Use embedded binary resources for grub templates.

NO

###### Test Methodology

- Built a Mariner vhdx file and ensured it booted.
- Built a Mariner ISO, installed an OS with it and verified the newly installed OS booted.
- Ran toolkit UTs.


elainezhao96 (approved):

toolkit/tools/internal/file/file.go
```diff
@@ -282,15 +301,38 @@ func copyWithPermissions(src, dst string, dirmode os.FileMode, changeMode bool,
 		}
 	}
 
-	err = shell.ExecuteLive(squashErrors, "cp", "--preserve=mode", src, dst)
+	return
+}
+
+func CopyFileFromFS(srcFS fs.FS, srcFile, dst string, dirmode os.FileMode, filemode os.FileMode) error {
```
dmcilvaney, line 307: Could we reference asset, resource, etc. in the function name here, just to make it easier to figure out what its doing for a new reader?

dmcilvaney (approved):

toolkit/tools/internal/file/file.go
cwize1: Technically `fs.FS` is a generic file system interface, with implementations for both embedded files and OS files. Though realistically this function will only ever be used for embedded files. So, it is probably fine to change the name.

PR #6698, opened 2023-11-08
Filter out debuginfo packages when running sodiff

sameluch:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

sodiff was not functioning as expected for 2.0 given the split of debuginfo repos from the base packages when compared to its implementation for 1.0. Added repositories for debug info for both core and extended repo generation commands. Additionally added the functionality for use with extended and a modular choice on whether the script produces an error.

###### Change Log  

- Add debuginfo repo to sodiff checks
- Add commands for use with extended repo
- fix error when no rpms are built
- Add error variable for when we want the script to cause an error
- Add parameter validation

**NO**

###### Test Methodology

- Local testing
- Pipeline build id: Extended Build with Run: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=449079&view=results
- build RPMS: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=447198&view=results
- Fast Track: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=445738&view=results (Failed result, but good sodiff for core and extended of ARM64)


toolkit/scripts/analysis.mk
```diff
@@ -71,6 +75,17 @@ sodiff-repo: $(SODIFF_REPO_FILE)
 
 $(SODIFF_REPO_FILE):
 	echo $(SODIFF_REPO_SOURCES) | sed -E 's:([^ ]+[.]repo):$(SPECS_DIR)/mariner-repos/\1:g' | xargs cat > $(SODIFF_REPO_FILE)
+	awk '{if ($$0 ~ /enabled=0/) $$0="enabled=1"; print}' $(SODIFF_REPO_FILE) > $(SODIFF_REPO_FILE)_fixed
```
anphel31, line 78: What is `$$0` used for?

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -2,11 +2,36 @@
...
+    echo "INVALID ARBUMENT: REPO_FILE_PATH is empty"
+    exit 1
+fi
 
-rpms_folder="$1"
-repo_file_path="$2"
-mariner_version="$3"
-sodiff_out_dir="$4"
+if [[ -z "$mariner_version" ]]; then
+    echo "INVALID ARBUMENT: MARINER_VERSION is empty"
```
anphel31, line 27: typo  - ARBUMENT

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -2,11 +2,36 @@
...
+    exit 1
+fi
+
+if [[ -z "$sodiff_out_dir" ]]; then
+    echo "INVALID ARBUMENT: SODIFF_OUT_DIR is empty"
```
anphel31, line 32: typo - ARBUMENT

anphel31 (approved):

toolkit/scripts/analysis.mk
sameluch: in this case the $$0 refers to the line that awk is reading in from the file. When concatenating the repo files together I want to ensure that each repo being added is enabled

toolkit/scripts/sodiff/mariner-sodiff.sh
sameluch: Fixed

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -2,11 +2,36 @@
 
 # Required binaries:
```
anphel31, line 3: Can you take a look at the grep errors that I think are generated from this file?
```
grep: invalid option -- '['
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=455863&view=logs&j=2ccbf1db-6c73-5af8-5ba2-cf05469dcf51&t=62118ca4-5f53-54b3-7ec9-8ef4ab0e9237

jslobodzian (approved):

toolkit/scripts/analysis.mk
```diff
@@ -71,6 +75,17 @@ sodiff-repo: $(SODIFF_REPO_FILE)
...
+	mv $(SODIFF_REPO_FILE)_fixed $(SODIFF_REPO_FILE)
```
PawelWMS, line 79: I may be misunderstanding the logic, but if we only want to make sure `enabled=0` is turned into `enabled=1` everywhere, then would this work to keep things simpler:
```suggestion
	sed -iE 's/enabled=0/enabled=1/g' $(SODIFF_REPO_FILE)
```

PawelWMS: Having said that, it might be even cleaner and future-proof, if we enabled all repos when they are actually used. If we're using a package manager, then a single `--enablerepos=*` should do the trick. Unless there are some, you don't want to enable.

toolkit/scripts/analysis.mk
```diff
@@ -71,6 +75,17 @@ sodiff-repo: $(SODIFF_REPO_FILE)
...
+
+# sodiff-repo-extended: Generate just the sodiff.repo file
+.PHONY: sodiff-repo-extended
+sodiff-repo-extended: $(SODIFF_REPO_FILE_EXTENDED)
+
+$(SODIFF_REPO_FILE_EXTENDED):
+	echo $(SODIFF_REPO_SOURCES_EXTENDED) | sed -E 's:([^ ]+[.]repo):$(SPECS_DIR)/mariner-repos/\1:g' | xargs cat > $(SODIFF_REPO_FILE_EXTENDED)
+	awk '{if ($$0 ~ /enabled=0/) $$0="enabled=1"; print}' $(SODIFF_REPO_FILE_EXTENDED) > $(SODIFF_REPO_FILE_EXTENDED)_fixed
+	mv $(SODIFF_REPO_FILE_EXTENDED)_fixed $(SODIFF_REPO_FILE_EXTENDED)
```
PawelWMS, line 88: Similar idea for simplification here:
```suggestion
	sed -iE 's/enabled=0/enabled=1/g' $(SODIFF_REPO_FILE_EXTENDED)
```

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -2,11 +2,36 @@
...
+
+if [[ -z "$rpms_folder" ]]; then
+    echo "INVALID ARGUMENT: RPMS_FOLDER is empty"
+    exit 1
+fi
+
+if [[ -z "$repo_file_path" ]]; then
+    echo "INVALID ARGUMENT: REPO_FILE_PATH is empty"
+    exit 1
+fi
```
PawelWMS, line 24: Might be worth mentioning the flag, which is necessary to pass this value. Right now it's hard to say that `RPMS_FOLDER` is set by `-r`.

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -2,11 +2,36 @@
```
PawelWMS, line 2: I think we should add a Microsoft copyright notice, since this is a public repository.

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -2,11 +2,36 @@
 
 # Required binaries:
 # rpm and dnf
```
PawelWMS, line 4: Does this comment still make sense after recent changes? Not sure what it's about.

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -86,6 +113,9 @@ echo "######################"
 if [[ $pkgsFound -gt 0 ]]; then
     echo "The Following Packages Are in Need of an Update:"
     cat "$sodiff_out_dir"/sodiff-summary.txt
+    if [[ "$sodiff_script_error" -eq "true" ]]; then
+        exit 1
+    fi
```
PawelWMS, line 118: I'd suggest to always error out and leave it to the caller to deal with the error as they see fit.

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -2,11 +2,36 @@
 
 # Required binaries:
 # rpm and dnf
+sodiff_script_error=false
```
PawelWMS, line 5: Might be worth to also add `set -eo pipefail` to always error out on unexpected issues.

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -18,7 +43,7 @@ common_options="-c $repo_file_path --releasever $mariner_version"
 
 DNF_COMMAND=dnf
 # Cache RPM metadata
->/dev/null dnf $common_options -y makecache
+>/dev/null $DNF_COMMAND $common_options -y makecache
```
PawelWMS, line 46: Is `DNF_COMMAND` needed, if it never changes and doesn't add any extra arguments to `dnf`?

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -58,7 +83,9 @@ done
 # Obtain a list of unique packages to be updated
 2>/dev/null cat "$sodiff_out_dir"/require* | sort -u > "$sodiff_out_dir"/sodiff-intermediate-summary.txt
 
-rm "$sodiff_out_dir"/require*
+if [ ls "$sodiff_out_dir"/require* 1> /dev/null 2>&1 ]; then
+    rm "$sodiff_out_dir"/require*
+fi
```
PawelWMS, line 88: What is the purpose of this check? If you're trying to ignore missing files without an error, then `rm` has the `-f` flag.
```suggestion
rm -f "$sodiff_out_dir"/require*
```

toolkit/scripts/analysis.mk
```diff
@@ -20,11 +20,15 @@ ifneq ($(build_arch),x86_64)
...
 SODIFF_REPO_SOURCES="mariner-official-base.repo"
+SODIFF_REPO_SOURCES_EXTENDED="mariner-official-base.repo mariner-extended.repo"
 else
 SODIFF_REPO_SOURCES="mariner-official-base.repo mariner-microsoft.repo"
+SODIFF_REPO_SOURCES_EXTENDED="mariner-official-base.repo mariner-microsoft.repo mariner-extended.repo"
 endif
 
+
 SODIFF_REPO_FILE=$(SCRIPTS_DIR)/sodiff/sodiff.repo
+SODIFF_REPO_FILE_EXTENDED=$(SCRIPTS_DIR)/sodiff/sodiff-extended.repo
```
PawelWMS, line 31: Could you move these files inside the `$(BUILD_DIR)` directory? That's where we try to keep all files generated during build time, which are meant for internal consumption by the tooling. If it's supposed to be an output for the user, then please use the `$(OUT_DIR)` directory.

toolkit/scripts/sodiff/mariner-sodiff.sh
sameluch: This optionality was introduced for use in the pipelines.  In case there is something wrong where we need to ignore the hard error in the check, this is a backstop.

toolkit/scripts/sodiff/mariner-sodiff.sh
sameluch: The DNF_COMMAND variable is mainly there for easy refactoring in the future (if/when we change over to dnf5 or other lightweight manager in the future).

If there is no true objection, I would like to leave it in place but changed the var to all lower case for consistency in naming.

toolkit/scripts/sodiff/mariner-sodiff.sh
sameluch: Added

toolkit/scripts/sodiff/mariner-sodiff.sh
sameluch: Good call, Added

toolkit/scripts/analysis.mk
sameluch: replaced the awk with sed for simplicity/constistency.

toolkit/scripts/sodiff/mariner-sodiff.sh
sameluch: Added error messages for each command option.

toolkit/scripts/sodiff/mariner-sodiff.sh
sameluch: changed to use -f

toolkit/scripts/analysis.mk
sameluch: These were in the scripts_dir from back when this script was made in 1.0. Moved to BUILD_DIR since the placement is pretty arbitrary, and is better there.

toolkit/scripts/sodiff/mariner-sodiff.sh
sameluch: scratch that, adding `set -e` is causing the script to fail due to grep failures on 54 becoming hard errors, which are normally suppressed.

jslobodzian (approved):

toolkit/scripts/sodiff/mariner-sodiff.sh
```diff
@@ -1,12 +1,37 @@
 #!/bin/bash
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+sodiff_script_error=false
+while getopts "r:f:v:o:e:" opt; do
```
PawelWMS, line 6: Since we're here, I'm curious if we couldn't also re-name this to `azurelinux-sodiff.sh` or something along those lines. Just a P2 suggestion, though.

PR #6711, opened 2023-11-09
Switch ccache to using compiler content instead of its modified time.

gmileka:
This is necessary because the compiler gets rebuilt with each new toolchain build.

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
CCache hashes a number of attributes of each compilation operation including attributes of the compiler binary, the compiler switches, and the source code.

CCache is by default configured to record the modified time of the compiler binary as one of those attributes. This causes a problem because with each of our full builds, we recompiler the compiler as part of the toolchain and then use it to build the different packages. This means that with every full build, the entire ccache cache will get invalidated. Partial builds that just re-builds the rpms are okay since they use the same toolchain packages.

The fix is configure ccache to use the content of the compiler binary instead of its modified time. This will eliminate the confusion from the rebuild of the compiler. The downside is that it's a tiny bit slower since the compiler content has to be hashed as opposed to just reading the modified time stamp.

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


anphel31:
Please update the toolchain manifests for the new mariner-rpm-macros release number. This should resolve the "Check Manifests" PR issue.
From the toolkit directory, running the following should fix this:
`sed -i "s/-2.0-23.cm2/-2.0-24.cm2/g" resources/manifests/package/*.txt`

SPECS/mariner-rpm-macros/macros
```diff
@@ -93,6 +94,7 @@
     %{__os_install_post_leave_signatures} \
     %{?__brp_strip} \
     %{?mariner_ccache_enabled:CCACHE_DIR="/ccache-dir" ; export CCACHE_DIR ;} \
+    %{?mariner_ccache_enabled:CCACHE_COMPILERCHECK=content ; export CCACHE_COMPILERCHECK ;}\
```
anphel31, line 97: I'm not sure this change is needed. All we're doing here is printing the ccache stats. In fact I've considered removing the call to `ccache --show-stats` as it can clutter the logs and the numbers printed are cumulative, so perhaps not much value to the current build. While we're making a change to this file, i think we should either: (1) remove this call or (2) also call `ccache -z` to clear the stats at the beginning. What do you think?

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -201,12 +201,13 @@ func buildSRPMInChroot(chrootDir, rpmDirPath, toolchainDirPath, workerTar, srpmF
 	toolchainRpmsOverlayMount, toolchainRpmsOverlayExtraDirs := safechroot.NewOverlayMountPoint(chroot.RootDir(), overlaySource, chrootLocalToolchainDir, toolchainDirPath, chrootLocalToolchainDir, overlayWorkDirToolchain)
 	rpmCacheMount := safechroot.NewMountPoint(*cacheDir, chrootLocalRpmsCacheDir, "", safechroot.BindMountPointFlags, "")
 	mountPoints := []*safechroot.MountPoint{outRpmsOverlayMount, toolchainRpmsOverlayMount, rpmCacheMount}
+	extraDirs := append(outRpmsOverlayExtraDirs, chrootLocalRpmsCacheDir)
+	extraDirs = append(extraDirs, toolchainRpmsOverlayExtraDirs...)
 	if isCCacheEnabled(ccacheManager) {
 		ccacheMount := safechroot.NewMountPoint(ccacheManager.CurrentPkgGroup.CCacheDir, chrootCcacheDir, "", safechroot.BindMountPointFlags, "")
 		mountPoints = append(mountPoints, ccacheMount)
+		extraDirs = append(extraDirs, chrootCcacheDir)
```
anphel31, line 209: What is this needed for? (and why wasn't it needed before?) Could you please add a comment here explaining?

anphel31 (approved):

SPECS/mariner-rpm-macros/macros
gmileka: Makes sense, I will remove CCACHE_COMPILERCHECK from here.
For printing the stats, I think it's really helpful for debugging. The numbers in the official build are easy to read since it's always the first run on the cache and there's no aggregation. I see the problem for dev builds which are harder to read due to their cumulative ... Let me switch to -z...

toolkit/tools/pkgworker/pkgworker.go
gmileka: It was actually needed before, I just moved where it's being done.
`extraDirs` are the directories to be created. Before this change, while I moved ccache logic under isCCacheEnabled(), I have left the directory creation outside. This change moves the creation of the ccache folder to also be under isCCacheEnabled().

gmileka: I'll add a comment explaining what extraDirs are...

SPECS/mariner-rpm-macros/macros
anphel31: Sounds good, calling the `ccache -z` in the `__spec_build_pre` should fix this then

PR #6737, opened 2023-11-11
Fixed signature checking for local sources to break the build on a mismatch.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Our tooling is expected to fail the build by default in case of a spec source signature mismatch. Right now we were ignoring the errors returned by out validation logic for local files and only displayed a warning.

###### Change Log  

- Made sources' signatures mismatches fatal by default.

No.

###### Test Methodology

- Local tests.
- Buddy build for a successful case.

### DIFF
--- a/toolkit/tools/srpmpacker/srpmpacker.go
+++ b/toolkit/tools/srpmpacker/srpmpacker.go
@@ -917,19 +917,16 @@ func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir st
 		failureBackoffBase    = 2.0
 		downloadRetryDuration = time.Second
 	)
-	errOther := fmt.Errorf("hydrating from remote source failed. See earlier errors for details")
 	errPackerCancelReceived := fmt.Errorf("packer cancel signal received")
 
 	// Making sure we remove the hydrated file in case of signature validation failure.
 	filesToRemove := []string{}
 	defer func() {
-		if err != nil {
-			for _, fileToRemove := range filesToRemove {
-				logger.Log.Debugf("Removing file (%s) after signature validation failure.", fileToRemove)
-				removeErr := os.Remove(fileToRemove)
-				if removeErr != nil {
-					logger.Log.Errorf("Failed to delete file (%s) after signature validation failure. Error: %s", fileToRemove, removeErr)
-				}
+		for _, fileToRemove := range filesToRemove {
+			logger.Log.Debugf("Removing file (%s) after signature validation failure.", fileToRemove)
+			removeErr := os.Remove(fileToRemove)
+			if removeErr != nil {
+				logger.Log.Errorf("Failed to delete file (%s) after signature validation failure. Error: %s", fileToRemove, removeErr)
 			}
 		}
 	}()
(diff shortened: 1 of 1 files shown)


toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -941,11 +955,10 @@ func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir st
 			}
 		}
 
-		cancelled := false
-		cancelled, err = retry.RunWithExpBackoff(func() error {
+		cancelled, internalErr := retry.RunWithExpBackoff(func() error {
```
dmcilvaney, line 958: While here, should we switch this to use an explicit `err` to avoid naming collisions?
```go
netErr := network.DownloadFile(url, destinationFile, srcConfig.caCerts, srcConfig.tlsCerts)
if netErr != nil {
	logger.Log.Debugf("Failed an attempt to download (%s). Error: %s", url, netErr )
}

return netErr
```

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -956,28 +969,24 @@ func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir st
...
 		if cancelled {
 			err = errPackerCancelReceived
 			return
 		}
 
-		if err != nil {
-			// We may intentionally fail early due to a cancellation signal, stop immediately if that is the case.
+		if internalErr != nil {
+			logger.Log.Errorf("Failed to download (%s). Error: %s", url, internalErr)
+			err = errOther
```
dmcilvaney, line 980: I think this error just gets swallowed on the next pass through the loop. If I recall correctly the error state for each file is tracked by `fileHydrationState[fileName] = true` towards the end of the loop.

toolkit/tools/srpmpacker/srpmpacker.go
PawelWMS: I don't see why not.:)

toolkit/tools/srpmpacker/srpmpacker.go
PawelWMS: You're right, I wanted to preserve the error but a cancel (or anything else we might do here in the future) would override it. I haven't noticed that `fileHydrationState` already does what I needed.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -843,7 +843,7 @@ func hydrateFiles(fileTypeToHydrate fileType, specFile, workingDir string, srcCo
 		err = hydrateFromLocalSource(fileHydrationState, newSourceDir, srcConfig, skipSignatureHandling, currentSignatures)
 		// On error warn and default to hydrating from an external server.
```
dmcilvaney, line 844: Comment isn't accurate anymore, we now error out.

dmcilvaney (approved):

toolkit/tools/srpmpacker/srpmpacker.go
PawelWMS: I think it was never entirely accurate. I've updated it and the `hydrateFromLocalSource` function + its comment as well.

dmcilvaney (approved):

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -854,29 +853,40 @@ func hydrateFiles(fileTypeToHydrate fileType, specFile, workingDir string, srcCo
 		}
 	}
 
+	missingFiles := []string{}
 	for fileNeeded, alreadyHydrated := range fileHydrationState {
 		if !alreadyHydrated {
-			err = fmt.Errorf("unable to hydrate file: %s", fileNeeded)
-			logger.Log.Error(err)
+			missingFiles = append(missingFiles, fileNeeded)
+			logger.Log.Errorf("Unable to hydrate file: %s.", fileNeeded)
```
anphel31, line 860: nit: could you remove the period from this message? otherwise the filename might appear to end with it

anphel31 (approved):

toolkit/tools/srpmpacker/srpmpacker.go
PawelWMS: Ah, I have not though about that. Will do!

PR #6747, opened 2023-11-13
Image Customizer: Implement fallback partition customization.

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This change implements the default/fallback logic for partition customization. This will be used when the partition layout changes are too complex for modifying the partitions directly.

The fallback logic creates the desired partition layout and then does a file copy from the base image to the new image. After that, some minor fix-ups are done (e.g. configuring the boot-loader).

The Image Customizer partition config schema mostly mirrors that of the new-image config. Some features were not included (e.g. dm-verity, encryption, overlays) to avoid making a large change even larger. In addition, a lot of additional config validity checks were added to try to limit it to only things that are known to work.

###### Change Log  

- Image Customizer: Implement fallback partition customization.

NO

###### Test Methodology

- Added UTs.

### DIFF
--- a/toolkit/tools/imagecustomizerapi/config.go
+++ b/toolkit/tools/imagecustomizerapi/config.go
@@ -3,14 +3,81 @@
 
 package imagecustomizerapi
 
+import (
+	"fmt"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/sliceutils"
+)
+
 type Config struct {
+	Disks        *[]Disk      `yaml:"Disks"`
 	SystemConfig SystemConfig `yaml:"SystemConfig"`
 }
 
 func (c *Config) IsValid() error {
+	if c.Disks != nil {
+		disks := *c.Disks
+		if len(disks) < 1 {
+			return fmt.Errorf("at least 1 disk must be specified (or the Disks field should be ommited)")
+		}
+		if len(disks) > 1 {
+			return fmt.Errorf("multiple disks is not currently supported")
+		}
+
+		for i, disk := range disks {
+			err := disk.IsValid()
+			if err != nil {
+				return fmt.Errorf("invalid disk at index %d:\n%w", i, err)
+			}
+		}
+	}
+
 	err := c.SystemConfig.IsValid()
 	if err != nil {
 		return err
+	}
+
+	hasDisks := c.Disks != nil
+	hasBootType := c.SystemConfig.BootType != BootTypeUnset
+
+	if hasDisks != hasBootType {
+		return fmt.Errorf("SystemConfig.BootType and Disks must be specified together")
+	}
+
+	// Ensure the correct partitions exist to support the specified the boot type.
+	switch c.SystemConfig.BootType {
+	case BootTypeEfi:
+		hasEsp := sliceutils.ContainsFunc(*c.Disks, func(disk Disk) bool {
+			return sliceutils.ContainsFunc(disk.Partitions, func(partition Partition) bool {
+				return sliceutils.ContainsValue(partition.Flags, PartitionFlagESP)
+			})
+		})
+		if !hasEsp {
+			return fmt.Errorf("'esp' partition must be provided for 'efi' boot type")
+		}
+
+	case BootTypeLegacy:
+		hasBiosBoot := sliceutils.ContainsFunc(*c.Disks, func(disk Disk) bool {
+			return sliceutils.ContainsFunc(disk.Partitions, func(partition Partition) bool {
+				return sliceutils.ContainsValue(partition.Flags, PartitionFlagBiosGrub)
+			})
+		})
+		if !hasBiosBoot {
+			return fmt.Errorf("'bios_grub' partition must be provided for 'legacy' boot type")
+		}
+	}
+
+	// Ensure all the partition settings object have an equivalent partition object.
+	for i, partitionSetting := range c.SystemConfig.PartitionSettings {
+		diskExists := sliceutils.ContainsFunc(*c.Disks, func(disk Disk) bool {
+			return sliceutils.ContainsFunc(disk.Partitions, func(partition Partition) bool {
+				return partition.ID == partitionSetting.ID
+			})
+		})
+		if !diskExists {
+			return fmt.Errorf("invalid PartitionSetting at index %d:\nno partition with matching ID (%s)", i,
+				partitionSetting.ID)
+		}
 	}
 
 	return nil
--- a/toolkit/tools/imagecustomizerapi/disk.go
+++ b/toolkit/tools/imagecustomizerapi/disk.go
@@ -1,0 +1,98 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/sliceutils"
+)
+
+type Disk struct {
+	// The type of partition table to use (e.g. mbr, gpt)
+	PartitionTableType PartitionTableType `yaml:"PartitionTableType"`
+
+	// The virtual size of the disk.
+	MaxSize uint64 `yaml:"MaxSize"`
+
+	// The partitions to allocate on the disk.
+	Partitions []Partition `yaml:"Partitions"`
+}
+
+func (d *Disk) IsValid() error {
+	err := d.PartitionTableType.IsValid()
+	if err != nil {
+		return err
+	}
+
+	if d.MaxSize <= 0 {
+		return fmt.Errorf("a disk's MaxSize value (%d) must be a positive non-zero number", d.MaxSize)
+	}
+
+	for i, partition := range d.Partitions {
+		err := partition.IsValid()
+		if err != nil {
+			return fmt.Errorf("invalid partition at index %d:\n%w", i, err)
+		}
+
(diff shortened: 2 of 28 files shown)


toolkit/tools/imagecustomizerapi/config.go
```diff
@@ -3,15 +3,82 @@
...
+	hasBootType := c.SystemConfig.BootType != BootTypeUnset
+
+	if hasDisks != hasBootType {
+		return fmt.Errorf("SystemConfig.BootType and Disks must be specified together")
+	}
+
+	// Ensure the correct partitions exist to support the specified the boot type.
+	switch c.SystemConfig.BootType {
+	case BootTypeEfi:
+		hasEsp := sliceutils.ContainsFunc(*c.Disks, func(disk Disk) bool {
```
gmileka, line 50: just curious, is the use of ContainsFunction/Value here equivalent to nexted loops? Is there an advantage of using lambdas here? (in C++ this would typically be a bit more expensive than just doing 3 nested loops)

toolkit/tools/imagecustomizerapi/disk.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/sliceutils"
+)
+
+type Disk struct {
+	// The type of partition table to use (e.g. mbr, gpt)
+	PartitionTableType PartitionTableType `yaml:"PartitionTableType"`
+
+	// The virtual size of the disk.
+	MaxSize uint64 `yaml:"MaxSize"`
```
gmileka, line 19: curious, why `MaxSize` and not just `Size`?

toolkit/tools/imagecustomizerapi/mountidentifier.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import "fmt"
+
+// MountIdentifierType indicates how a partition should be identified in the fstab file
+type MountIdentifierType string
```
gmileka, line 9: Are any of these types already defined in the Mariner toolkit since they have some overlap with these?

toolkit/tools/imagecustomizerapi/config.go
cwize1: Yes, this could also be implemented with nested loops. It is just a matter of style preference.

(This code os only run once per program execution and works on very short lists. So, it doesn't need to be optimized.)

toolkit/tools/imagecustomizerapi/disk.go
cwize1: Not really sure. But it's the name the new-image config uses. So, we mirror it here.

toolkit/tools/imagecustomizerapi/mountidentifier.go
cwize1: Yeah, a lot of this has direct equivalents in the new-image config. But the image customizer guarantees back-compat. Whereas imager does not. So, the types are kept separate.

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
```diff
@@ -0,0 +1,195 @@
...
+	partitionSettings []imagecustomizerapi.PartitionSetting, bootType imagecustomizerapi.BootType, buildDir string,
+	chrootDirName string, installOS installOSFunc,
+) (*ImageConnection, error) {
+	imageConnection := &ImageConnection{}
+
+	err := createNewImageHelper(imageConnection, filename, diskConfig, partitionSettings, bootType, buildDir,
+		chrootDirName, installOS,
+	)
+	if err != nil {
+		imageConnection.Close()
```
elainezhao96, line 70: why cleanclose is not needed here?

toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
```diff
@@ -0,0 +1,86 @@
...
+	copyArgs := []string{"--verbose", "--no-clobber", "-a", "--no-dereference", "--sparse", "always", "-t", newImageChroot.RootDir()}
+
+	files, err := os.ReadDir(existingImageChroot.RootDir())
+	if err != nil {
+		return fmt.Errorf("failed to read base image root directory:\n%w", err)
+	}
+
+	for _, file := range files {
+		switch file.Name() {
+		case "dev", "proc", "sys", "run", "tmp":
```
elainezhao96, line 70: afaik some dirs under /var/ might not be meaningful to be copied over to the new disk? do we need to list them all here?

toolkit/tools/imagecustomizerapi/disk_test.go
```diff
@@ -0,0 +1,263 @@
...
+func TestDiskIsValidOverlaps(t *testing.T) {
+	disk := &Disk{
+		PartitionTableType: PartitionTableTypeGpt,
+		MaxSize:            4,
+		Partitions: []Partition{
+			{
+				ID:     "a",
+				FsType: "ext4",
+				Start:  1,
+				End:    ptrutils.PtrTo(uint64(3)),
```
romoh, line 118: curious, why not use `End: 3` ?

toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
```diff
@@ -0,0 +1,86 @@
...
+func copyFilesIntoNewDisk(existingImageChroot *safechroot.Chroot, newImageChroot *safechroot.Chroot) error {
+	err := copyFilesIntoNewDiskHelper(existingImageChroot, newImageChroot)
+	if err != nil {
+		return fmt.Errorf("failed to copy files into new partition layout:\n%w", err)
+	}
+	return nil
+}
+
+func copyFilesIntoNewDiskHelper(existingImageChroot *safechroot.Chroot, newImageChroot *safechroot.Chroot) error {
+	copyArgs := []string{"--verbose", "--no-clobber", "-a", "--no-dereference", "--sparse", "always", "-t", newImageChroot.RootDir()}
```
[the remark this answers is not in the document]
cwize1: `-a` includes `-r` and `--preserve`.

toolkit/tools/pkg/imagecustomizerlib/imageConnection.go
```diff
@@ -0,0 +1,84 @@
...
+	if c.chroot != nil {
+		c.chroot.Close(c.chrootIsExistingDir)
+	}
+
+	if c.loopback != nil {
+		c.loopback.Close()
+	}
+}
+
+func (c *ImageConnection) CleanClose() error {
```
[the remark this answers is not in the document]
cwize1: Hmm. You are unlikely to call `CleanClose()` if you didn't successfully call both `ConnectLoopback` and `ConnectChroot`. So, it isn't really necessary to check for `nil` here.

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
cwize1: `CleanClose()` is used when you want to make sure the resource fully closes before continuing.

`Close()` is used when you just want the resource to close, and you don't care if it succeeds or is done asynchronously. This is typically used when an error occurs.

toolkit/tools/imagecustomizerapi/partition.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type Partition struct {
+	// ID is used to correlate `Partition` objects with `PartitionSetting` objects.
+	ID string `yaml:"ID"`
+	// FsType is the type of file system to use on the partition.
+	FsType FileSystemType `yaml:"FsType"`
+	// Name is the label to assign to the partition.
+	Name string `yaml:"Name"`
+	// Start is the offset where the partition begins (inclusive), in MiBs.
```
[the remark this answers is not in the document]
cwize1: `End` is what the imager's API uses. But I guess I could also support `Size`.

toolkit/tools/imagecustomizerapi/config.go
```diff
@@ -3,15 +3,82 @@
...
+			return sliceutils.ContainsFunc(disk.Partitions, func(partition Partition) bool {
+				return sliceutils.ContainsValue(partition.Flags, PartitionFlagBiosGrub)
+			})
+		})
+		if !hasBiosBoot {
+			return fmt.Errorf("'bios_grub' partition must be provided for 'legacy' boot type")
+		}
+	}
+
+	// Ensure all the partition settings object have an equivalent partition object.
```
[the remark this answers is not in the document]
cwize1: Yes, good catch.

toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
cwize1: Those are a couple of symlinks under `/var`, including `/var/lock` and `/var/run`. Since `cp` is called with `--no-dereference`, these symlinks will be copied as symlinks.

Are there any other directories you are concerned about?

romoh (approved):

toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
elainezhao96: got it. no other concerns

elainezhao96 (approved):

PR #6762, opened 2023-11-14
Toolkit: Add retry to safemount.Close().

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Adds a retry loop when safemount attempts to unmount the directory. This should hopefully prevent the resource busy errors that very occasionally appear.

###### Change Log  

- Toolkit: Add retry to safemount.Close().

NO

###### Test Methodology

- Add UT.

### DIFF
--- a/toolkit/tools/internal/safemount/main_test.go
+++ b/toolkit/tools/internal/safemount/main_test.go
@@ -1,0 +1,44 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package safemount
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+)
+
+var (
+	tmpDir     string
+	workingDir string
+)
+
+func TestMain(m *testing.M) {
+	var err error
+
+	logger.InitStderrLog()
+
+	workingDir, err = os.Getwd()
+	if err != nil {
+		logger.Log.Panicf("Failed to get working directory, error: %s", err)
+	}
+
+	tmpDir = filepath.Join(workingDir, "_tmp")
+
+	err = os.MkdirAll(tmpDir, os.ModePerm)
+	if err != nil {
+		logger.Log.Panicf("Failed to create tmp directory, error: %s", err)
+	}
+
+	retVal := m.Run()
+
+	err = os.RemoveAll(tmpDir)
+	if err != nil {
+		logger.Log.Warnf("Failed to cleanup tmp dir (%s). Error: %s", tmpDir, err)
+	}
+
+	os.Exit(retVal)
+}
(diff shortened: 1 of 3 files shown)


toolkit/tools/internal/safemount/safemount_test.go
```diff
@@ -0,0 +1,124 @@
...
+	err = mount.CleanClose()
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, "busy")
+
+	// Close the file.
+	fileOnPartition.Close()
+
+	// Try to close the mount again.
+	err = mount.CleanClose()
+	assert.NoError(t, err, "failed to close the mount")
```
dmcilvaney, line 116: To better exercise the retry, maybe something like this?
I did some similar patterns in the retry unit tests

```suggestion
	// Close the file.
	go func() {
		time.Sleep(time.Duration(2)  * time.Second)
		fileOnPartition.Close()
	}

	// Try to close the mount again.
	time.Sleep(time.Duration(1) * time.Second)
	err = mount.CleanClose()
	assert.NoError(t, err, "failed to close the mount")
```

toolkit/tools/internal/safemount/safemount_test.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// Open a file.
+	fileOnPartitionPath := filepath.Join(partitionMountPath, "test")
+
+	fileOnPartition, err := os.OpenFile(fileOnPartitionPath, os.O_RDWR|os.O_CREATE, 0)
+	if !assert.NoErrorf(t, err, "failed to open file", fileOnPartitionPath) {
+		return
+	}
+	defer fileOnPartition.Close()
```
romoh, line 104: nice idea to test this

romoh (approved):

toolkit/tools/internal/safemount/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+)
+
+var (
+	tmpDir     string
+	workingDir string
+)
+
+func TestMain(m *testing.M) {
```
dmcilvaney, line 19: I'm not super familiar with `main_test.go`, we have generally dropped `TestMain()` into the same `<pkg>_test.go` files as the other tests. Is there a benefit to splitting this out into its own file?

cwize1: The `TestMain()` function is for the entire package. So, it makes sense to keep it in a separate file. Granted, it doesn't matter much in this package since there is only a single file. But this is the way I've done it elsewhere. So, I am copying it here.

toolkit/tools/internal/safemount/safemount_test.go
cwize1: That might be a little prone to race conditions. However, I did add a duration check for the first close() to sanity check that the retries are actually occurring.

PR #6763, opened 2023-11-15
update to systemd 254

ddstreetmicrosoft:
###### Merge Checklist  

- [x] Ready to merge

---

udpate to systemd v254


reubeno:
What's your thinking about version of systemd-bootstrap with respect to systemd? Are these changes an implicit decision to keep the bootstrap package on an older version, or is it going to be upgraded in a subsequent PR?

Also just an FYI that v255 rc1 was released last week (https://github.com/systemd/systemd/releases/tag/v255-rc1). There appear to be a number of interesting changes in it.

ddstreetmicrosoft:
> What's your thinking about version of systemd-bootstrap with respect to systemd? Are these changes an implicit decision to keep the bootstrap package on an older version, or is it going to be upgraded in a subsequent PR?

systemd-bootstrap should go away completely. I chatted with @dmcilvaney in teams and it sounds like the toolkit currently needs it so i stashed it off in systemd-bootstrap; but nobody should ever use it for anything other than the toolkit's internal bootstrap process. I hope the toolkit can be updated to no longer need it at some point. I don't plan to ever touch it again.

> 
> Also just an FYI that v255 rc1 was released last week (https://github.com/systemd/systemd/releases/tag/v255-rc1). There appear to be a number of interesting changes in it.

Yeah I'll update to 255 once released, but I don't think we should be using RCs.

SPECS/systemd/systemd.spec
```diff
@@ -209,7 +175,6 @@ fi
 %dir %{_sysconfdir}/modules-load.d
 %dir %{_sysconfdir}/binfmt.d
 %{_sysconfdir}/X11/xinit/xinitrc.d/50-systemd-user.sh
-%{_sysconfdir}/sysctl.d/50-security-hardening.conf
```
christopherco, line 212: Where is the recommended place to set the default hardening sysctls? I'd like to make sure we keep many of these settings

toolkit/resources/manifests/package/toolchain_aarch64.txt
```diff
@@ -594,4 +594,4 @@ zstd-1.5.5-1.cm2.aarch64.rpm
 zstd-debuginfo-1.5.5-1.cm2.aarch64.rpm
 zstd-devel-1.5.5-1.cm2.aarch64.rpm
 zstd-doc-1.5.5-1.cm2.aarch64.rpm
-zstd-libs-1.5.5-1.cm2.aarch64.rpm
\ No newline at end of file
+zstd-libs-1.5.5-1.cm2.aarch64.rpm
```
christopherco, line 597: just curious what this zstd-libs diff is - missing EOF or invisible character?

cgmanifest.json
```diff
@@ -28187,8 +28187,8 @@
         "type": "other",
         "other": {
           "name": "systemd",
-          "version": "250.3",
-          "downloadUrl": "https://github.com/systemd/systemd-stable/archive/v250.3.tar.gz"
+          "version": "254",
+          "downloadUrl": "https://github.com/systemd/systemd/archive/v254.tar.gz"
```
christopherco, line 28191: When we get to AzLinux 3.0 release, is the thinking to stay with systemd upstream releases or go back to the systemd-stable releases?

SPECS/systemd/systemd.spec
```diff
@@ -1,35 +1,14 @@
...
-Patch4:         CVE-2022-45873.patch
-Patch5:         backport-helper-util-macros.patch
-Patch6:         CVE-2022-4415.patch
-Patch7:         serve-stale-0001-resolved-added-serve-stale-feature-implementation-of.patch
-Patch8:         serve-stale-0002-resolved-Initialize-until_valid-while-storing-negati.patch
-# Patch9 should be dropped for mariner 3
-Patch9:         mariner-2-do-not-default-zstd-journal-files-for-backwards-compatibility.patch
-Patch10:        update-cifs-for-kernel-headers-6.1.patch
+Source0:        https://github.com/%{name}/%{name}/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
+Source1:        95-disable-systemd-oomd.preset
```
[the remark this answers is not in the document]
ddstreetmicrosoft: right, systemd 'preset' files only set the default enablement for services, so here we're shipping a preset file that defaults systemd-oomd (service and socket) to disabled, which means in new instance builds they'll be disabled, and end users will need to manually enable them (i.e. systemctl enable systemd-oomd.socket or systemctl enable systemd-oomd.service).

systemd-oomd is a much more configurable (and generally, much more aggressive) oom killer than the standard kernel oom killer, and so really isn't appropriate to be enabled by default; it would result in too many end user complaints about inappropriately killed processes. To use it, end users really need to evaluate and customize the config to their specific use case and then enable it.

and yeah, the removal of the systemd.cfg boot params means we'll default to cgroupsv2.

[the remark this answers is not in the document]
christopherco: Makes sense - thanks!

SPECS/systemd/systemd.spec
ddstreetmicrosoft: these are already the default and so aren't actually 'hardening' anything, and don't need to be explicitly specified:
```
kernel.randomize_va_space=2
kernel.dmesg_restrict = 1
kernel.unprivileged_bpf_disabled = 1
net.ipv4.ip_forward = 0
net.ipv6.conf.all.forwarding = 0
```

it's debatable if this 'hardens' anything, since root can just set it to whatever else is desired, and it definitely reduces ease of debugging kernel issues (by disabling all sysrq operations and key sequences):
`kernel.sysrq=0`

this has nothing to do with 'hardening' and simply disables ipv6 autoconfiguration, which *at most* should be something that end users opt-in to (instead of end users having to opt-in to ipv6 autoconfiguration actually working):
`net.ipv6.conf.all.accept_ra = 0`

this actually '**de-hardens**' things by violating [rfc 5961](https://datatracker.ietf.org/doc/html/rfc5961):
`net.ipv4.tcp_challenge_ack_limit = 1073741823`

this is the only 'hardening' setting:
`kernel.kptr_restrict=1`

the `kernel.kptr_restrict=1` setting should go (along with any other 'identity-specific' config) into a `azure-linux-release-identity-*` package, similar to the `fedora-release-identity-*` packages. If it's a config that should go into *all* our images and/or install media, it could go into a `azure-linux-release-common` package, like the `fedora-release-common` package (we also should probably add a `99-default-disable.preset` into `azure-linux-release-common`, like the `fedora-release-common` package has).

For hardening-specific stuff we might want to have a `azure-linux-release-identity-hardened` or something that other more-specific release identity packages (that should have hardening) can require.

toolkit/resources/manifests/package/toolchain_aarch64.txt
ddstreetmicrosoft: yeah this is yet other text file without a EOF newline; since all unix editors add a EOF newline because all unix text files are required to have one, it's actually quite hard for me to edit a text file and *not* have the missing EOF newline automatically fixed (i.e. I didn't add this EOF newline manually; simply editing the release number earlier in the file and then saving it performed the EOF newline correction automatically)

ddstreetmicrosoft: it's easier to see with git cli:

```
$ git show toolkit/resources/manifests/package/toolchain_aarch64.txt
commit 2f6893a608a1b2bcce88cc732ab83e803498c4fd (HEAD -> ddstreet/systemd-254, origin/ddstreet/systemd-254)
Author: Dan Streetman <ddstreet@ieee.org>
Date:   Tue Nov 14 18:35:30 2023 -0500

    update to systemd 254

diff --git a/toolkit/resources/manifests/package/toolchain_aarch64.txt b/toolkit/resources/manifests/package/toolchain_aarch64.txt
index e871137fd..81650c388 100644
--- a/toolkit/resources/manifests/package/toolchain_aarch64.txt
+++ b/toolkit/resources/manifests/package/toolchain_aarch64.txt
@@ -555,10 +555,10 @@ sqlite-devel-3.44.0-1.cm2.aarch64.rpm
 sqlite-libs-3.44.0-1.cm2.aarch64.rpm
 swig-4.0.2-3.cm2.aarch64.rpm
 swig-debuginfo-4.0.2-3.cm2.aarch64.rpm
-systemd-bootstrap-250.3-13.cm2.aarch64.rpm
-systemd-bootstrap-debuginfo-250.3-13.cm2.aarch64.rpm
-systemd-bootstrap-devel-250.3-13.cm2.aarch64.rpm
-systemd-bootstrap-rpm-macros-250.3-13.cm2.noarch.rpm
+systemd-bootstrap-250.3-14.cm2.aarch64.rpm
+systemd-bootstrap-debuginfo-250.3-14.cm2.aarch64.rpm
+systemd-bootstrap-devel-250.3-14.cm2.aarch64.rpm
+systemd-bootstrap-rpm-macros-250.3-14.cm2.noarch.rpm
 tar-1.34-2.cm2.aarch64.rpm
 tar-debuginfo-1.34-2.cm2.aarch64.rpm
 tdnf-3.5.2-2.cm2.aarch64.rpm
@@ -594,4 +594,4 @@ zstd-1.5.5-1.cm2.aarch64.rpm
 zstd-debuginfo-1.5.5-1.cm2.aarch64.rpm
 zstd-devel-1.5.5-1.cm2.aarch64.rpm
 zstd-doc-1.5.5-1.cm2.aarch64.rpm
-zstd-libs-1.5.5-1.cm2.aarch64.rpm
\ No newline at end of file
+zstd-libs-1.5.5-1.cm2.aarch64.rpm

```

cgmanifest.json
ddstreetmicrosoft: in m2, we weren't actually following the systemd stable releases (the latest stable-250 is 250.13), so if we continue that it doesn't really matter much if we pull from stable or non-stable.

But yes, I do think it would be a good idea for azl3 to actually follow systemd-stable releases; I'll change this to switch over to 254.6 and we can move over to stable-255 once that's out

ddstreetmicrosoft: well, unfortunately the systemd-254.6.tar.gz in our 'blobstore' was added with the wrong upstream URL (which doesn't work), so there is no way for me to get the PR check in this to work; if the spec file has the blobstore-registered URL then the PR check can't download the tarball (because the URL is wrong), and if the spec file has the correct URL then it doesn't match the registered URL and the PR check fails.

Hopefully someone can delete the broken blobstore tarball and i'll re-publish with the right url

SPECS/systemd-bootstrap/systemd-bootstrap.spec
```diff
@@ -1,7 +1,7 @@
 Summary:        Bootstrap version of systemd. Workaround for systemd circular dependency.
 Name:           systemd-bootstrap
 Version:        250.3
```
dmcilvaney, line 3: What is the rationale for leaving the bootstrap version behind? Ideally both versions should be the same shouldn't they? Can we guarantee that every package in the toolchain looking for systemd will choose configuration flags compatible with 254 if they see only 250 available during the toolchain build?

dmcilvaney (changes requested):

SPECS/systemd-bootstrap/systemd-bootstrap.spec
ddstreetmicrosoft: > What is the rationale for leaving the bootstrap version behind? Ideally both versions should be the same shouldn't they? Can we guarantee that every package in the toolchain looking for systemd will choose configuration flags compatible with 254 if they see only 250 available during the toolchain build?

we should be building packages with the systemd we actually ship - if we aren't then that's a major problem and needs to be fixed.

dmcilvaney: Unfortunately circular dependencies make that impossible. We need some mechanism to break those cycles and in the rpm world that is `%bootstrap`. To simplify the tools we don't use the macro and instead maintain separate .spec files, but the concept is the same. Its a value judgement from a developer where and how to inject the bootstrap packages, and they are done very sparingly. ie, the bootsrapped package should never statically link, ideally not even provide dynamic libraries and should only supply tools as a build-requires, header files, etc. 

The boostrap packages ARE real packages, we publish them, and a user may install them if they really want, but the "real" package conflicts with, and obsoletes, the bootstrap versions. They should only ever show up as a BuildRequires: in a spec file, and ideally a VERY select set. Looks like the systemd one has a few BRs (`audit`, `dbus`, `lvm2`, `p11-kit`). I can't speak to how valid the bootstrap requirement is for each of those, but generally if it shows up in the toolchain we need to either use bootstrap version, or move a package out of the toolchain.

cgmanifest.json
ddstreetmicrosoft: ok working now, not sure if it was my mistake or what, but i used 254.5 instead and it's working ok

SPECS/systemd-bootstrap/systemd-bootstrap.spec
ddstreetmicrosoft: I'm happy to adopt the fedora bootstrapping macro in systemd, i.e. https://docs.fedoraproject.org/en-US/packaging-guidelines/#bootstrapping

However an entire duplication of the systemd package, including publishing the resulting built binaries, is a nonstarter. it makes no sense, doubles the maintenance work, and introduces the possibility of end users actually *using* the systemd-bootstrap package which would be obviously completely wrong.

As long as the systemd-bootstrap package is its own separate entity, i'm happy to leave it untouched (and hopefully unused) but i certainly won't be maintaining 2 copies of systemd.

ddstreetmicrosoft: just to clarify - the 'bootstrap' process is intended for bringup of *new arches*, where it is possible to have 'circular' build dependencies, because there are literally no packages built. The bootstrap process works in stages and the output of the early stages (where packages are built using bootstrap) *should not be published or used outside the bootstrap process*.

For example:
https://fedoraproject.org/wiki/Architectures/AArch64/Bootstrap
https://github.com/rwmjones/fedora-riscv-bootstrap/blob/master/README

We now *have* a complete package repository for Mariner 2. We no longer need to bootstrap and we should no longer do so.

SPECS/systemd/systemd.spec
christopherco: These all make sense. I've booked some work to sanitize these settings, and I like the idea to what settings we really need to keep available in `azure-linux-release-identity-*` variants for making different variants & also having a `-common` package for settings we always want. Thanks for the very thorough explanation!

SPECS/systemd-bootstrap/systemd-bootstrap.spec
dmcilvaney: One of the core tenants of Mariner is the ability for anyone to bootstrap from source, and time and anywhere. We have held that since before M1, and it underpins some of the core design decisions of the build system. If we want to abandon that tenant that will have to be an entire separate discussion.

My understanding is that Fedora also uses the bootstrap process https://docs.fedoraproject.org/en-US/packaging-guidelines/#bootstrapping during the bring up of a new release as well, but I could be wrong there.

christopherco (commented):
I am less familiar with the systemd-bootstrap side of things but sounds like there's a bit of discussion going on. Everything else LGTM.

SPECS/systemd-bootstrap/systemd-bootstrap.spec
christopherco: > One of the core tenants of Mariner is the ability for anyone to bootstrap from source, and time and anywhere. We have held that since before M1, and it underpins some of the core design decisions of the build system

This is true and will require quite a bit of a redesign to our tools to change. 

Some questions to help my understanding of the toolchain problem @dmcilvaney:
- Thoughts on the bootstrap macro method? There will likely be tooling changes for supporting the build for bootstrap (thinking grapher & scheduler here), and we'd also need the changes to the systemd spec with the bootstrap macro, but I do want to see if we can avoid hard duplication of the spec files this way.

- Are there any known issues we have actually encountered with using a different systemd version in the `-bootstrap` package / toolchain stage (either in the distro itself or in the tooling)?

reubeno: >Thoughts on the bootstrap macro method?

I see need for a repeatable pattern to allow use of macro-conditional variation among multiple variants of a package. We've already got duplication in the multiple kernel packages and I've seen need in a Mariner derivative for similar. I've managed to skate by in a few cases using `%include` carefully but I'm quite supportive of seeing us use conditional macro logic as the RPM community intended. In other words, if we decide we need to maintain both a bootstrap package and a product one, then I'd vote for the "macro" method. My request is that we try to avoid hard-coded special support for *just* a special bootstrap macro and explore options that would work more generally.

As for which version to use in general, do we know what the backporting policy is for `systemd-stable`? I see, for example, `v250` getting far fewer backported patches than `v252`.

SPECS/systemd/systemd.spec
```diff
@@ -1,35 +1,14 @@
...
-Patch4:         CVE-2022-45873.patch
-Patch5:         backport-helper-util-macros.patch
-Patch6:         CVE-2022-4415.patch
-Patch7:         serve-stale-0001-resolved-added-serve-stale-feature-implementation-of.patch
-Patch8:         serve-stale-0002-resolved-Initialize-until_valid-while-storing-negati.patch
-# Patch9 should be dropped for mariner 3
-Patch9:         mariner-2-do-not-default-zstd-journal-files-for-backwards-compatibility.patch
-Patch10:        update-cifs-for-kernel-headers-6.1.patch
+Source0:        https://github.com/%{name}/%{name}-stable/archive/refs/tags/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
+Source1:        95-disable-systemd-oomd.preset
```
tobiasb-ms, line 11: Does this need to be in `systemd.signatures.json` too?

SPECS/systemd-bootstrap/systemd-bootstrap.spec
ddstreetmicrosoft: > My understanding is that Fedora also uses the bootstrap process https://docs.fedoraproject.org/en-US/packaging-guidelines/#bootstrapping during the bring up of a new release as well, but I could be wrong there.

AFAIK Fedora does not re-bootstrap each new stable release; it's branched off rawhide. You might be thinking of [mass rebuilds](https://docs.fedoraproject.org/en-US/infra/release_guide/mass_rebuild/), which are different from bootstrapping.

> As for which version to use in general, do we know what the backporting policy is for systemd-stable? 

Last time we checked I believe the answer was that there are no LTS releases, so it is probably fine to just pick whatever specific release we want (e.g. the latest before we get to a freeze point in our dev cycle).

> if we decide we need to maintain both a bootstrap package and a product one

The bootstrap package should never be used by anyone, and we shouldn't be publishing it. It's a intermediate package that should be used only to build other packages that require it. The non-bootstrap version of a package is the only one anyone should use and is the only one we should publish.

> use conditional macro logic as the RPM community intended

While I don't understand what the technical benefit is to the toolkit performing a full 'bootstrap' rebuild of all packages from nothing, if the toolkit is designed around that then I don't think there is any particular urgency to change it (as long as building normal package updates don't use the bootstrap process). But yes, it should do the bootstrapping right, which means using spec macros so there are no package duplications (e.g. systemd.spec and systemd-bootstrap.spec). I don't know what needs to change in the toolkit to do that, but from just a package building perspective, it should be as simple as e.g. `rpmbuild --with bootstrap -bb systemd.spec` instead of `rpmbuild -bb systemd-bootstrap.spec`.

This also does extend into actual packages that have BuildRequires: set to bootstrap packages (e.g. the `p11-kit` package has `BuildRequires: systemd-bootstrap-devel` which should change to `systemd-devel`).

SPECS/systemd/systemd.spec
ddstreetmicrosoft: I am not actually clear on why the signatures json is needed at all, but it seems like the toolkit only needs the source tarball included in it.

SPECS/systemd-bootstrap/systemd-bootstrap.spec
reubeno: > ...it should be as simple as e.g. rpmbuild --with bootstrap -bb systemd.spec instead of rpmbuild -bb systemd-bootstrap.spec..

Agreed. FWIW, from recent experience, I've found that the number of call sites to `rpmspec` and `rpmbuild` (which both need to specify all `--with` args correctly) is more than I'd expect, and also quite distributed. (The number goes higher when you trace back all callers of some of the helper functions generically invoking them as well.) Perhaps that's less of an issue for bootstrapping which may not require enlightening all such call sites that are using `rpmspec` to parse specs.

SPECS/systemd/systemd.spec
anphel31: All source files need to be in the signatures.json file. When the toolkit attempts to pack the SRPM files it verifies all of the sha256 hashes of each source. some notes here https://github.com/microsoft/CBL-Mariner/blob/42bc30b1ad4745468911e6749b460947c5e3e9bc/toolkit/docs/building/building.md?plain=1#L295

SPECS/systemd/systemd.signatures.json
```diff
@@ -1,9 +1,5 @@
 {
  "Signatures": {
-  "50-security-hardening.conf": "c9fc6624a3178cc0980ec3495ac2d5f02a8a1f3a4cdadb8770628d7394b92e2c",
-  "99-dhcp-en.network": "158c0003826cef9d780799c6edb2814800d4518f85f884710550f4c46d14be67",
-  "99-mariner.preset": "84326fcef828f893827f3d5acdb7a6fe722a8c8e593b19dcc68b082a1c597c79",
-  "systemd-250.3.tar.gz": "87b0eee7b6e5aaab2ab56d158f9536daa6bfd5de011f2a5fc6ccdd81ee1e7a24",
-  "systemd.cfg": "1fd673c1da90560d9395a18df31f2ce6ef23972a6bf8c910ba6f3fb07818afa9"
+  "systemd-254.5.tar.gz": "41873783aa1c680e10d2f2626797a1c2fef8018d69b68c8c77639e140ee7846d"
```
anphel31, line 3: can you add back all of the rest of the source files? these are needed in order for "make input-srpms" to pass

SPECS/systemd/systemd.spec
ddstreetmicrosoft: What I'm not really clear on is what benefit the signature file actually brings...it doesn't add any security to anything, it just seems to add extra work? For example, a malicious actor who is able to change one of the files included in the srpm build can just as easily update the hash in the signatures file...there is no added security benefit.

SPECS/systemd/systemd.signatures.json
ddstreetmicrosoft: done

tobiasb-ms (approved):

SPECS/systemd-bootstrap/systemd-bootstrap.spec
christopherco: Based on the conversations on this, I'm in favor of going with bootstrapping macros for this, with evaluation on getting an n-1 support in the toolkit so we can evaluate.

I'm also good with approving this current set of `systemd-bootstrap` v250 changes to allow the current toolkit to operate, while additional toolkit support is built up for supporting bootstrapping macros and n-1 poc.

christopherco (approved):
From the bootstrapping conversations, I'm in favor of going with bootstrapping macros for this, with evaluation on getting an n-1 support in the toolkit so we can evaluate.

I'm also good with approving this current change set with systemd-bootstrap v250 separately to allow the current toolkit to operate and get us standard systemd v254.

Meanwhile added some work items for us to prioritize - 1.) toolkit support for bootstrapping macros; and 2.) n-1 support in toolkit so we can evaluate it.

SPECS/systemd/systemd.spec
christopherco: https://microsoft.visualstudio.com/OS/_workitems/edit/47697769

ddstreetmicrosoft:
@dmcilvaney i think this PR is blocked on your review requesting changes

dmcilvaney (approved):

PR #6766, opened 2023-11-15
fix tmux crashing bug

### LINKED ISSUE #6598: tmux crashes when mouse is enabled and used for text selection
### Repro
1. With a fully updated version of Mariner 2.0, enable mouse input on `tmux` by adding `set -g mouse on` to `~/.tmux.conf`.
2. Run `tmux` and highlight text with the mouse.

Result: `tmux` crashes when the mouse button is released.
image

### Investigation
The [crash was introduced](https://github.com/microsoft/CBL-Mariner/commit/a1e39c361a57b84a77a0fd132fb67b5471fcb13b) by upgrading `ncurses` to [6.4-20230408](https://invisible-island.net/ncurses/NEWS.html#index-t20230408) to fix [CVE-2023-29491](https://nvd.nist.gov/vuln/detail/CVE-2023-29491).

This change introduced an incompatibility with `tmux` where `tmux` calls an `ncurses` function and it now returns `null`, but released versions of `tmux` don't handle that case.

`ncurses` subsequently released [6.4-20230423](https://invisible-island.net/ncurses/NEWS.html#index-t20230423) that introduces a new function to call that will fix the issue `tmux` has. `tmux` [added code to use that](https://github.com/tmux/tmux/commit/39d41d0810d4e8ae6ce8d27776dfbb96722d9319), but it is not yet in a released build and it does not apply cleanly to `tmux 3.2a`.

I've created a [patch file](https://github.com/microsoft/CBL-Mariner/files/13161336/0001-Manual-patch-to-fix-crash-due-to-change-to-ncurses.patch) that will get us most of the way to fixing this issue. Note that this introduces a build-time (and run-time) requirement for `tmux` on `ncurses-6.4-20230423`.

tobiasb-ms:
###### Merge Checklist  
- [x] Ready to merge

---

###### Summary 
Fixes issue #6598 which is a crash on selection in tmux. The fix requires an update to ncurses and a patch to tmux. From the patch comments:

```
ncurses-6.4-20230408 change tparm to require cur_term, which broke tmux usage of it.

ncurses-6.4-20230423 then added tiparm_s that allows usage without cur_term.

tmux change https://github.com/tmux/tmux/commit/39d41d0810d4e8ae6ce8d27776dfbb96722d9319 uses tiparm_s if it exists, but cannot be cleanly applied to tmux tag 3.2a.

That change uses a config setting to created #defines to determine which version of tparm it should use, and only conditionally uses tiparm_s, because it needs to be backwards compatible with previous versions of ncurses.

But to use that, we would need to get the actual source as it appears in github, rather than the released version (they are different downloads: see https://github.com/tmux/tmux/releases).

Fortunately, we have the luxery of forcing tmux to use a version of ncurses that has the function we want (see above).

Given all this, this patch takes the change to use tiparm_s, removes the conditional compilation portion so it always uses tiparm_s and applies it to the code as it exists in 3.2a.

It has both a build-time and run-time dependency on ncurses-6.4-20230423 or later.
```

###### Change Log  
- Update ncurses to a version that includes a new version of the tiparm API that tmux uses to fix the crash.
- Patch tmux to use new API.

YES

###### Associated issues  

- Fixes #6598

###### Test Methodology
- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=453180&view=results
- Manual testing
- Buddy-build testing with internal customer who saw this bug


SPECS/tmux/manual-patch-to-fix-crash-due-to-change-to-ncurses.patch
```diff
@@ -0,0 +1,90 @@
...
+
+That change uses a config setting to created #defines to determine which version of tparm it should use, and only conditionally uses tiparm_s, because it needs to be backwards compatible with previous versions of ncurses.
+
+But to use that, we would need to get the actual source as it appears in github, rather than the released version (they are different downloads: see https://github.com/tmux/tmux/releases).
+
+Fortunately, we have the luxery of forcing tmux to use a version of ncurses that has the function we want (see above).
+
+Given all this, this patch takes the change to use tiparm_s, removes the conditional compilation portion so it always uses tiparm_s and applies it to the code as it exists in 3.2a.
+
+It has both a build-time and run-time dependency on ncurses-6.4-20230423 or later.
```
rlmenge, line 20: Nit: Can you add this blurb about the patch to the PR description?

neha170 (approved):
LGTM. Did we try rebuilding the toolchain with the updated ncurses package?

tobiasb-ms:
> LGTM. Did we try rebuilding the toolchain with the updated ncurses package?

Yes. The build I linked to did full toolchain, package and image builds for AMD and ARM.

rlmenge (approved):

SPECS/tmux/tmux.spec
```diff
@@ -1,16 +1,17 @@
...
 Group:          Applications/System
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 Source0:        https://github.com/tmux/tmux/releases/download/%{version}/%{name}-%{version}.tar.gz
 Patch0:         CVE-2022-47016.patch
-Requires:       libevent ncurses
-BuildRequires:  libevent-devel ncurses-devel
+Patch1:         manual-patch-to-fix-crash-due-to-change-to-ncurses.patch
+Requires:       libevent ncurses >= 6.4-2
+BuildRequires:  libevent-devel ncurses-devel >= 6.4-2
```
hbeberman, line 14:
```suggestion
Requires:       libevent
Requires:       ncurses >= 6.4-2
BuildRequires:  libevent-devel
BuildRequires:  ncurses-devel >= 6.4-2
```

SPECS/tmux/manual-patch-to-fix-crash-due-to-change-to-ncurses.patch
```diff
@@ -0,0 +1,90 @@
...
+
+ncurses-6.4-20230423 then added tiparm_s that allows usage without cur_term.
+
+tmux change https://github.com/tmux/tmux/commit/39d41d0810d4e8ae6ce8d27776dfbb96722d9319 uses tiparm_s if it exists, but cannot be cleanly applied to tmux tag 3.2a.
+
+That change uses a config setting to created #defines to determine which version of tparm it should use, and only conditionally uses tiparm_s, because it needs to be backwards compatible with previous versions of ncurses.
+
+But to use that, we would need to get the actual source as it appears in github, rather than the released version (they are different downloads: see https://github.com/tmux/tmux/releases).
+
+Fortunately, we have the luxery of forcing tmux to use a version of ncurses that has the function we want (see above).
```
neha170, line 16: nit: luxury

tobiasb-ms: ugh

rlmenge (approved):

PR #6769, opened 2023-11-15
Add Mariner Image Modifier

elainezhao96:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
Add initial Mariner Image modifier(IM2) files 

###### Change Log  
- Extracted the core logic from user related functions `AddUser`, `UserExists`, `ConfigureUserGroupMembership` to allow null chroot, and invoke them from pkg/imagemodifierlib/modifierutils.go with `nil` passed into chroot 
- Add the SSHPubKeys to user struct as requested by Trident crew, add `ProvisionUserSSHCertsWithPubKeys` function accordingly
- Add some initial doc. Will complete the docs for this IM2

TODO: Complete the TODOs for other user related functions

NO

###### Test Methodology
Will do an end-to-end test in trident to validate we can set up users correctly.


toolkit/tools/pkg/imagemodifierlib/imagemodifier.go
```diff
@@ -0,0 +1,44 @@
...
+	"path/filepath"
+
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/imagecustomizerapi"
+)
+
+func ModifyImageWithConfigFile(configFile string) error {
+	var err error
+
+	var config imagecustomizerapi.Config
+	err = imagecustomizerapi.UnmarshalYamlFile(configFile, &config.SystemConfig)
```
elainezhao96, line 17: @cwize1  Here I pass &config.SystemConfig as opposed to &config,  so from customer side they don't need to prefix `SystemConfig:` at the top of the yaml. (e.g. Trident doesnt have this SystemConfig instead they have osconfig:, the entities under osconfig will be written to a yaml). let me know if this makes sense or not.

cwize1: Yeah, that fine. Though, I would use `var systemConfig imagecustomizerapi.SystemConfig`. No point using the `Config` type if you are not using it.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1422,35 +1422,37 @@ func Chage(installChroot *safechroot.Chroot, passwordExpirationInDays int64, use
 	return fmt.Errorf(`user "%s" not found when trying to change the password expiration date`, username)
 }
 
-func ConfigureUserGroupMembership(installChroot *safechroot.Chroot, username string, primaryGroup string,
-	secondaryGroups []string,
-) (err error) {
+func ConfigureUserGroupMembership(installChroot *safechroot.Chroot, username string, primaryGroup string, secondaryGroups []string) (err error) {
+	if installChroot == nil {
```
cwize1, line 1426: You should probably just make `updateUserGroupMembershipCore` public and let callers use that one if they want to run without providing a chroot context.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1422,35 +1422,37 @@ func Chage(installChroot *safechroot.Chroot, passwordExpirationInDays int64, use
...
+	}
+
+	err = installChroot.UnsafeRun(func() error {
+		return updateUserGroupMembershipCore(username, primaryGroup, secondaryGroups)
+	})
+
+	return err
+}
+
+func updateUserGroupMembershipCore(username string, primaryGroup string, secondaryGroups []string) error {
```
cwize1, line 1437: Perhaps `updateUserGroupMembershipInOS`?

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1561,32 +1615,43 @@ func ProvisionUserSSHCerts(installChroot *safechroot.Chroot, username string, ss
...
 
-	return
+	// Set ownership of the SSH directory
+	ownership := fmt.Sprintf("%s:%s", username, primaryGroup)
+	if err := shell.ExecuteLive(squashErrors, "chown", "-R", ownership, userSSHKeyDir); err != nil {
+		return err
+	}
+
+	// Set permissions of the SSH directory
+	if err := shell.ExecuteLive(squashErrors, "chmod", "-R", sshDirectoryPermission, userSSHKeyDir); err != nil {
```
cwize1, line 1650: Changing the error flow-style is making the diff larger than it needs to be.

Also, the Mariner toolkit generally doesn't place operations within `if` statements. So, it would probably be good to remain consistent with the rest of the codebase.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1477,45 +1479,97 @@ func ConfigureUserStartupCommand(installChroot *safechroot.Chroot, username stri
...
-func ProvisionUserSSHCerts(installChroot *safechroot.Chroot, username string, sshPubKeyPaths []string) (err error) {
-	var (
-		pubKeyData []string
-		exists     bool
-	)
-	const squashErrors = false
-	const authorizedKeysTempFilePerms = 0644
-	const authorizedKeysTempFile = "/tmp/authorized_keys"
-	const sshDirectoryPermission = "0700"
+func setupSSHEnvironment(authorizedKeysTempFilePerms os.FileMode, authorizedKeysTempFile string) (err error) {
```
cwize1, line 1482: If you place `setupSSHEnvironment` between `ProvisionUserSSHCertsWithPubKeys` and `ProvisionUserSSHCerts`, then the diff would be smaller.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1477,45 +1479,97 @@ func ConfigureUserStartupCommand(installChroot *safechroot.Chroot, username stri
...
+		err = os.Truncate(authorizedKeysTempFile, 0)
+		if err != nil {
+			return fmt.Errorf("failed to truncate %s file: %v", authorizedKeysTempFile, err)
+		}
+	}
+
+	return nil
+}
 
+func ProvisionUserSSHCertsWithPubKeys(username string, sshPubKeys []string) (err error) {
```
cwize1, line 1507: You should allow the user to specify both `SSHPubKeyPaths` and `SSHPubKeys` and then concatenate them together. This will give the user extra flexibility. It would probably make the code in this file a little simpler, since `ProvisionUserSSHCertsWithPubKeys` and `ProvisionUserSSHCerts` will be combined into a single function and won't need to duplicate logic.

toolkit/tools/imagecustomizerapi/user.go
```diff
@@ -17,6 +17,7 @@ type User struct {
 	PasswordPath        string   `yaml:"PasswordPath"`
 	PasswordExpiresDays *int64   `yaml:"PasswordExpiresDays"`
 	SSHPubKeyPaths      []string `yaml:"SSHPubKeyPaths"`
+	SSHPubKeys          []string `yaml:"SSHPubKeys"`
```
cwize1, line 20: This is image customizer's public interface. The image modifier is free to reuse these types if it makes sense to do so. But this API can't be changed without also applying the same functionality to the image customizer.

toolkit/tools/imagemodifier/README.md
```diff
@@ -0,0 +1,36 @@
+# Mariner Image Modifier
```
cwize1, line 1: We probably don't need to maintain documentation for the image modifier.

toolkit/tools/pkg/imagemodifierlib/modifierutils.go
```diff
@@ -0,0 +1,102 @@
...
+		err := addOrUpdateUser(user, baseConfigPath)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func addOrUpdateUser(user imagecustomizerapi.User, baseConfigPath string) error {
```
cwize1, line 35: It would probably be goodness to figure out a way to avoid needing to duplicate this code.

cwize1, line 35: For basic functions, splitting the functions into a chroot variant and in-OS variant is fairly straightforward. But once we start dealing with these higher-level functions, it starts to become a lot messier.

It is possible that my original idea of having a chroot interface and a dummy implementation for in-OS scenarios, might be a little cleaner.

elainezhao96: make sense

elainezhao96: I have a commit of this change, could you take a look at https://github.com/microsoft/CBL-Mariner/compare/main...elaine/add-initial-im2?diff=split and let me know if this is the right path?

cwize1: Yeah, that’s looks far cleaner to me.

(By the way, the RootDir() should probably be “/“.)

toolkit/tools/imagecustomizerapi/user.go
elainezhao96: got it. then I will also add same functionality to the image customizer to allow this in the API

elainezhao96:
created another pr with above comments addressed: https://github.com/microsoft/CBL-Mariner/pull/6824

toolkit/tools/imagegen/installutils/installutils.go
elainezhao96: Decided to keep it as is in the new pr, as we are making sshkeys as a param in the original function

toolkit/tools/imagemodifier/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+The Mariner Image Modifier (IM2) is a tool derived from Mariner Image Customizer (MIC).
+This is invoked inside a chroot envirtonment where the desired filesystems are mounted,
```
Britel, line 4: (sp) environment

PR #6785, opened 2023-11-17
[3.0] Update kernel configs and dwarves to 1.25 to match new toolchain packages

rlmenge:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Binutils 2.41 introduced a new type for DW_TAG. Dwarves (pahole)
1.21 was returning "Unsupported DW_TAG_unspecified_type(0x3b)"
because it could not recognize this new type. Therefore update to
1.25 which adds support. Discussion can be seen at
https://lore.kernel.org/all/YzwkazNc6wNCpQTN@kernel.org/t/

The kernel required updating the configs to match the new versions of
gcc and pahole.

The kernel also required a patch for tools/perf/tests/btf.c to address some
build errors seen with the new gcc.
See https://lore.kernel.org/all/ZMj8+bvN86D0ZKiB%40kernel.org/

Note there are many warnings with:
"egrep: warning: egrep is obsolescent; using grep -E"
These should resolve with the next LTS of 6.6.x.
(For example https://github.com/torvalds/linux/commit/2a8ec611ac2feef30716066e031a32004e5664d0)
###### Change Log  

- Update kernel configs and dwarves to 1.25 to match new toolchain packages

**NO**

###### Test Methodology

- Pipeline build id AMD: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=454570&view=logs&j=75cba72d-09cb-503a-3927-4ab61ac4a179&t=460b7289-1832-5507-1175-ad5f30576086&l=5611
- Pipeline build id ARM; https://dev.azure.com/mariner-org/mariner/_build/results?buildId=454688&view=logs&j=75cba72d-09cb-503a-3927-4ab61ac4a179&t=460b7289-1832-5507-1175-ad5f30576086&l=5626


anphel31 (approved):

SPECS/kernel/perf_bpf_test_add_nonnull_argument.patch
```diff
@@ -0,0 +1,35 @@
+From 81807bcba96f3d98824e9fa2c597c22eb619dd7a Mon Sep 17 00:00:00 2001
+From: Rachel Menge <rachelmenge@microsoft.com>
+Date: Fri, 17 Nov 2023 05:21:58 +0000
+Subject: [PATCH] Fix perf bpf tests
+
+Taken from: 
+https://lore.kernel.org/all/ZMj8+bvN86D0ZKiB%40kernel.org/
```
ddstreetmicrosoft, line 7: It's generally not a good idea to pull patches from mailing list archives; that skips the very, very important steps of peer review and upstream acceptance for the patch. In other words, just because a patch is in a mailing list archive doesn't mean it's upstream.

Also, the best way to generate a patch file from an upstream commit (for any project, not just the kernel) is to create the patch file with git format-patch; since this particular commit upstream is ed847e30f001b207013b6136c264454d7560557f, the right way to do this is `git format-patch -1 ed847e30f001b207013b6136c264454d7560557f`. That brings along all the commit description as well as retaining the patch authorship and, most importantly, it retains the actual upstream commit id (on the first line, e.g. `From ed847e30f001b207013b6136c264454d7560557f Mon Sep 17 00:00:00 2001`)

SPECS/kernel/kernel.spec
```diff
@@ -155,7 +156,7 @@ manipulation of eBPF programs and maps.
 
 %prep
 %setup -q -n CBL-Mariner-Linux-Kernel-rolling-lts-mariner-%{mariner_version}-%{version}
-
+%patch1 -p1
```
ddstreetmicrosoft, line 159: wait, does this mean we weren't actually applying Patch0 before? This should probably be changed over to %autosetup so we don't miss future patches, right?

SPECS/dwarves/dwarves.spec
```diff
@@ -117,7 +115,6 @@ rm -Rf %{buildroot}
 %doc MANIFEST README
 %{_includedir}/dwarves/btf_encoder.h
 %{_includedir}/dwarves/config.h
-%{_includedir}/dwarves/ctf_encoder.h
```
ddstreetmicrosoft, line 120: This should be ok for 3.0 but definitely not for 2.0 (since it would break existing users who might be doing e.g. `#include <dwarves/ctf_encoder.h>`.

Also, removing header files usually isn't done very often, were all these header files actually removed? Might be good to at least double-check that...if they were removed upstream, that's fine, just would be good to take a sec to verify.

Also note this may break other packages we have that depend on this package (which is ok, assuming it's fixed when we update those packages, assuming there are any that break)

ddstreetmicrosoft (changes requested):
minor stuff, otherwise LGTM

SPECS/kernel/kernel.spec
rlmenge: I intentionally turned off the nvme patch for 3.0 but have a task to see whether it should be kept. I agree this is confusing though as is. Let me yank out the patch

SPECS/dwarves/dwarves.spec
rlmenge: Yup! I checked because it also seemed fishy. Arnaldo Carvalho de Melo is the upstream maintainer and is from redhat. Here is his commit https://src.fedoraproject.org/rpms/dwarves/c/550742fafde6bd74864491d12410404529ba7d4f?branch=rawhide

rlmenge: Also this is only for 3.0 to match latest binutils :)

SPECS/kernel/perf_bpf_test_add_nonnull_argument.patch
rlmenge: Makes sense! Updated patch to contain upstream info and will look to use git format-patch -1 in future :)

ddstreetmicrosoft (approved):
LGTM

PR #6809, opened 2023-11-20
imagecustomizer: initial dm-verity enablement by nbd.

liulanze:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Initialize dm-verity enablement using MIC.

###### Change Log  

- Change
- Change
- Change

**NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,222 @@
...
+
+	if len(kernelFiles) == 0 {
+		return fmt.Errorf("no kernels found in chroot environment")
+	}
+
+	// Extract the version from the kernel filename (e.g., vmlinuz-5.15.131.1-2.cm2 -> 5.15.131.1-2.cm2)
+	kernelVersion := strings.TrimPrefix(kernelFiles[0], "vmlinuz-")
+
+	buildDracutModuleArgs := []string{
+		"dracut", "-f", "--kver", kernelVersion, "-a",
```
liulanze, line 76: Will address [comment](https://github.com/microsoft/CBL-Mariner/pull/6679) from [cwize1](https://github.com/cwize1).

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -84,6 +84,11 @@ func doCustomizations(buildDir string, baseConfigPath string, config *imagecusto
 		return err
 	}
 
+	err = enableVerityPartition(config.SystemConfig.Verity, imageChroot)
```
liulanze, line 87: Will address https://github.com/microsoft/CBL-Mariner/pull/6679 from [cwize1](https://github.com/cwize1).

vinceaperri:
Would be good to prefix this PR title so it's obvious what component this PR is for.

toolkit/tools/imagecustomizerapi/verity.go
```diff
@@ -0,0 +1,79 @@
...
+	DefaultVerity = Verity{
+		VerityTab:                "",
+		VerityDevice:             "",
+		HashDevice:               "",
+		BootDevice:               "",
+		VerityCorruptionResponse: "",
+	}
+)
+
+func (v *Verity) IsValid() (err error) {
```
troy0820, line 30: nit: 
```suggestion
func (v *Verity) IsValid() error {
```

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+		if strings.HasPrefix(trimmedLine, "set rootdevice=PARTUUID=") {
+			// Replace the root device line with the new root device. TODO: add supported type 'user'
+			line = "set rootdevice=/dev/mapper/root"
+		}
+		updatedLines = append(updatedLines, line)
+	}
+
+	// Write the updated content back to grub.cfg
+	err = ioutil.WriteFile(grubConfigPath, []byte(strings.Join(updatedLines, "\n")), 0644)
```
troy0820, line 165:
```suggestion
	err = os.WriteFile(grubConfigPath, []byte(strings.Join(updatedLines, "\n")), 0644)
```
`iotutil` is deprecated

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,222 @@
...
+
+// findFreeNBDDevice finds the first available NBD device.
+func findFreeNBDDevice() (string, error) {
+	files, err := filepath.Glob("/sys/class/block/nbd*")
+	if err != nil {
+		return "", err
+	}
+
+	for _, file := range files {
+		size, err := ioutil.ReadFile(filepath.Join(file, "size"))
```
troy0820, line 181:
```suggestion
		size, err := os.ReadFile(filepath.Join(file, "size"))
```
`ioutil` is deprecated

troy0820:
Some initial nits but I was curious around the testing of this.  Are we going to add unit tests around this?
Also, do we have to use the os/exec package to shell out to the host?

troy0820 (changes requested):
Initial review.  Can you please update the nits for the deprecated items.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+		}
+		output := strings.Join(lines, "\n")
+		return ioutil.WriteFile(cfgPath, []byte(output), 0644)
+	})
+
+	return nil
+}
+
+func updateGrubConfig(resolvedVerityDevice string, resolvedHashDevice string, salt string, rootHash string, verityCorruptionResponse string) error {
```
vinceaperri, line 137: When updating the linux command line in grub.cfg, I suggest to follow the pattern seen in the 2.0 grub.cfg for systemd:

- https://github.com/microsoft/cbl-mariner/blob/main/toolkit/tools/internal/resources/assets/grub2/grub.cfg#L6-L8
- https://github.com/microsoft/cbl-mariner/blob/main/SPECS/systemd/systemd.cfg#L2

Here's my suggestion. Add this if-else block to the base grub.cfg.

```
if [ -f  $bootprefix/verity.cfg ]; then
	load_env -f $bootprefix/verity.cfg
else
	set verity_cmdline=
fi
```

Add $verity_cmdline to the end of the linux command in the base grub.cfg too.

Then, define a template and have this function (over)write it to $bootprefix/verity.cfg:

```
rd.systemd.verity=1 roothash={{.RootHash}} systemd.verity_root_data={{.SystemdVerityRootData}} systemd.verity_root_hash={{.SystemdVerityRootHash}} systemd.verity_root_options={{.SystemdVerityRootOptions}},salt={{.Salt}}
```

This makes would make this function simpler and idempotent, but requires a small change to the grub.cfg of the base image.

toolkit/tools/imagecustomizerapi/verity.go
```diff
@@ -0,0 +1,79 @@
...
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+type Verity struct {
```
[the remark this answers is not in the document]
cwize1: Also, should we use the existing imager utils for creating dm-verity disks? Or do we need to do something different in the image customizer?

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,222 @@
...
+		return "", err
+	}
+
+	for _, file := range files {
+		size, err := ioutil.ReadFile(filepath.Join(file, "size"))
+		if err != nil {
+			continue
+		}
+
+		if strings.TrimSpace(string(size)) == "0" {
```
vinceaperri, line 186: This doesn't work if there is a backing file with length = 0. I suggest checking if /sys/block/nbd*/pid is present, and if not, the nbd* device is free/unconnected because there is no client for it.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -105,6 +107,14 @@ func CustomizeImage(buildDir string, baseConfigPath string, config *imagecustomi
 		return fmt.Errorf("failed to convert image file to format: %s:\n%w", outputImageFormat, err)
 	}
 
+	if config.SystemConfig.Verity.VerityTab != "" {
```
vinceaperri, line 110: Should do this before the final conversion to the output image format.

toolkit/tools/imagecustomizerapi/verity.go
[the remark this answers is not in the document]
romoh: I wouldn't lock our implementation to root only as systemd also support /usr readonly. 
That said, It is not clear to me how we can extend the current API for more than just root.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	return fmt.Sprintf("%sp%s", nbdDevice, partitionNumber), nil
+}
+
+// findDeviceByUUIDOrLabel attempts to resolve a PARTUUID, PARTLABEL, UUID, or LABEL to a device file.
+func findDeviceByUUIDOrLabel(uuidOrLabel string) (string, error) {
+	// Check if the input is already a device path
+	if strings.HasPrefix(uuidOrLabel, "/dev/") {
+		return uuidOrLabel, nil
```
vinceaperri, line 211: Should throw an error if this is the case.

vinceaperri: It would be a logical error if this function is called with an already-resolved /dev path.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,222 @@
...
+
+	if partitionNumber == "" {
+		return "", fmt.Errorf("failed to extract partition number from %s", systemDevice)
+	}
+
+	return fmt.Sprintf("%sp%s", nbdDevice, partitionNumber), nil
+}
+
+// findDeviceByUUIDOrLabel attempts to resolve a PARTUUID, PARTLABEL, UUID, or LABEL to a device file.
+func findDeviceByUUIDOrLabel(uuidOrLabel string) (string, error) {
```
vinceaperri, line 208: I don't think the config should include UUIDs or PARTUUIDs since they are different for different images and aren't human readable. I think you only need to support the partition name.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	// Integrate systemd veritysetup dracut module into initramfs img.
+	systemdVerityDracutModule := "systemd-veritysetup"
+	err = buildDracutModule(systemdVerityDracutModule, imageChroot) 
+	if err != nil {
+		return err
+	}
+
+	// Update mariner config file with the new generated initramfs file.
```
vinceaperri, line 33: Why would the path to the initrd.img change?

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -233,3 +243,110 @@ func customizeImageHelper(buildDir string, baseConfigPath string, config *imagec
...
+	var err error
+
+	// Connect the disk image to an NBD device using qemu-nbd
+	// Find a free NBD device
+	nbdDevice, err := findFreeNBDDevice()
+	if err != nil {
+		return fmt.Errorf("failed to find a free nbd device: %v", err)
+	}
+
+	_, _, err = shell.Execute("sudo", "qemu-nbd", "-c", nbdDevice, buildImageFile)
```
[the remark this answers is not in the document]
liulanze: The reason for using `qemu-nbd` as an additional tool here is that when I calculate the root hash from a loop device, it consistently results in hash corruption upon booting the image. However, this error does not occur when using nbd. Since the POC is successfully demonstrated with nbd, I am considering using nbd for assistance. But I think perhaps the nbd logic can be moved before final convert yeah.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -233,3 +243,110 @@ func customizeImageHelper(buildDir string, baseConfigPath string, config *imagec
...
+	if salt == "" || rootHash == "" {
+		return fmt.Errorf("failed to parse salt or root hash from veritysetup output")
+	}
+
+	resolvedBootDevice, err := findDeviceByUUIDOrLabel(config.SystemConfig.Verity.BootDevice)
+	if err != nil {
+		return fmt.Errorf("failed to resolve boot device %s: %v", config.SystemConfig.Verity.BootDevice, err)
+	}
+
+	nbdBootDevice, err := convertToNbdDevicePath(nbdDevice, resolvedBootDevice)
```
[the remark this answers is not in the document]
liulanze: Thanks, I will add additional check statement.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
liulanze: The `systemd-veritysetup` module exists by default in `/etc/dracut.conf.d/`. However, it will not be included automatically in the dracut install/build process because no dependencies trigger the inclusion of this module. Therefore, I need to include it manually by using the `-a` flag and specifying the name of this module.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -301,15 +301,15 @@ func customizeVerityImageHelper(buildDir string, baseConfigPath string, config *
 	saltMatches := saltRegex.FindStringSubmatch(verityOutput)
 	if len(saltMatches) > 1 {
```
cwize1, line 302: Write it like this:

```golang
if len(saltMatches) <= 1 {
    return fmt.Errorf("failed to parse salt from veritysetup output")
}

salt = saltMatches[1]
```

See, https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

toolkit/tools/imagecustomizerapi/veritycorruptionresponse.go
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+)
+
+type VerityCorruptionResponse string
```
cwize1, line 10: Rename to `VerityErrorBehavior` to be consistent with the imager's config.

toolkit/tools/imagecustomizerapi/veritycorruptionresponse.go
```diff
@@ -0,0 +1,28 @@
...
+
+const (
+	VerityCorruptionResponseIgnore  VerityCorruptionResponse = "ignore-corruption"
```
cwize1, line 13: Change to `"ignore"`, `"restart"`, and `"panic"` to be consistent with the imager's config.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -206,10 +206,10 @@ func convertToNbdDevicePath(nbdDevice, systemDevice string) (string, error) {
 
 // findDeviceByUUIDOrLabel attempts to resolve a PARTUUID, PARTLABEL, UUID, or LABEL to a device file.
 func findDeviceByUUIDOrLabel(uuidOrLabel string) (string, error) {
-	// Check if the input is already a device path
+	// Error if the input is already a device path
 	if strings.HasPrefix(uuidOrLabel, "/dev/") {
-		return uuidOrLabel, nil
-	}
+        return "", fmt.Errorf("logical error: function called with resolved device path '%s'", uuidOrLabel)
+    }
```
cwize1, line 212: Remember to run `go fmt`.

toolkit/tools/imagecustomizerapi/verity.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+)
+
+type Verity struct {
```
cwize1, line 10: The user shouldn't need to know the UUIDs or labels of the image's partitions to be able to setup dm-verity.

The partition UUIDs of the standard Mariner image will change every time it is built. And the user should be able to update the base image without updating their config. Also, if the config contains partition customizations, then all the UUID could change during partition customization. 

The tool should automatically detect the correct partitions to use. The [findPartitions()](https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/tools/pkg/imagecustomizerlib/partitionutils.go#L24C6-L24C20) function could be useful for this.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
cwize1: In the [dracut config](https://man7.org/linux/man-pages/man5/dracut.conf.5.html), there is the `add_dracutmodules` command.

cwize1 (changes requested):

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/systemconfig.go
```diff
@@ -28,6 +28,7 @@ type SystemConfig struct {
 	Users                   []User                    `yaml:"Users"`
 	Services                Services                  `yaml:"Services"`
 	Modules                 Modules                   `yaml:"Modules"`
+	Verity                  Verity                    `yaml:"Verity"`
```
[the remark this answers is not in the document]
liulanze: Please review @cwize1

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,249 @@
...
+	newInitramfs := filepath.Base(initramfsFiles[0])
+
+	cfgPath := filepath.Join(imageChroot.RootDir(), "boot/mariner.cfg")
+	// Update mariner.cfg to reference the new initramfs
+	input, err := os.ReadFile(cfgPath)
+	if err != nil {
+		return fmt.Errorf("failed to read mariner.cfg: %w", err)
+	}
+
+	lines := strings.Split(string(input), "\n")
```
cwize1, line 108: You could use `file.ReadLines` and `file.WriteLines` here.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,249 @@
...
+			lines[i] = "mariner_initrd=" + newInitramfs
+		}
+	}
+	output := strings.Join(lines, "\n")
+	os.WriteFile(cfgPath, []byte(output), 0644)
+
+	return nil
+}
+
+func updateGrubConfig(dataPartitionIdType imagecustomizerapi.IdType, dataPartitionId string, hashPartitionIdType imagecustomizerapi.IdType, hashPartitionId string, rootHash string, grubCfgFullPath string) error {
```
cwize1, line 120: Line is too long.

```golang
func updateGrubConfig(dataPartitionIdType imagecustomizerapi.IdType, dataPartitionId string,
	hashPartitionIdType imagecustomizerapi.IdType, hashPartitionId string, rootHash string, grubCfgFullPath string,
) error {
```

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,249 @@
...
+	var err error
+
+	const cmdlineTemplate = "rd.systemd.verity=1 roothash=%s systemd.verity_root_data=%s systemd.verity_root_hash=%s systemd.verity_root_options=panic-on-corruption"
```
cwize1, line 123: nit: Place the string literal directly in the call to `fmt.Sprintf`. Splitting the format string from its args makes the code harder to read.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,249 @@
...
+	// Format the dataPartitionId and hashPartitionId using the helper function.
+	formattedDataPartitionPlaceHolder, err := systemdFormatPartitionId(dataPartitionIdType, dataPartitionId)
```
cwize1, line 125: Out of interest, why are you calling this variable a "placeholder"?

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,249 @@
...
+
+// systemdFormatPartitionId formats the partition ID based on the ID type following systemd dm-verity style.
+func systemdFormatPartitionId(idType imagecustomizerapi.IdType, id string) (string, error) {
+	imagerIdType, err := IdTypeToImager(idType)
+	if err != nil {
+		return "", err
+	}
+
+	switch imagerIdType {
+	case "Partition":
```
cwize1, line 223: nit: `IdTypePartition`

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,249 @@
...
+
+// idToPartitionBlockDevicePath returns the block device path for a given idType and id.
+func idToPartitionBlockDevicePath(idType imagecustomizerapi.IdType, id string, nbdDevice string, diskPartitions []diskutils.PartitionInfo) (string, error) {
+	imagerIdType, err := IdTypeToImager(idType)
+	if err != nil {
+		return "", err
+	}
+
+	// Create a regular expression to find the last number in the id string.
+	partitionRegex := regexp.MustCompile(`\d+$`)
```
cwize1, line 175: Delete this regex. Instead, the user should only specify the partition number.

toolkit/tools/imagecustomizerapi/verity.go
```diff
@@ -0,0 +1,35 @@
...
+
+package imagecustomizerapi
+
+import (
+	"fmt"
+)
+
+type VerityPartition struct {
+	IdType IdType `yaml:"IdType"`
+	Id     string `yaml:"Id"`
```
cwize1, line 12: You should verify that the value of `Id` has the correct format given the value of `IdType`.

The [isGPTNameValid](https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/tools/imagecustomizerapi/partition.go#L90C19-L90C20) function may be useful when checking the `IdTypePartlabel` type.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if len(kernelFiles) == 0 {
+		return fmt.Errorf("no kernels found in chroot environment")
+	}
+
+	// Extract the version from the kernel filename (e.g., vmlinuz-5.15.131.1-2.cm2 -> 5.15.131.1-2.cm2)
+	kernelVersion := strings.TrimPrefix(kernelFiles[0], "vmlinuz-")
+
+	err = imageChroot.Run(func() error {
+		// TODO: Config Dracut module systemd-veritysetup.
```
cwize1, line 72: Add task number.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,249 @@
...
+func systemdFormatPartitionId(idType imagecustomizerapi.IdType, id string) (string, error) {
+	imagerIdType, err := IdTypeToImager(idType)
+	if err != nil {
+		return "", err
+	}
+
+	switch imagerIdType {
+	case "Partition":
+		return id, nil
+	// case imagecustomizerapi.ID: // Ignored for now.
```
cwize1, line 225: nit: Delete line

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -222,3 +233,93 @@ func customizeImageHelper(buildDir string, baseConfigPath string, config *imagec
...
+			// Skips file system types that aren't known to support the boot loader partition.
+			continue
+		}
+
+		// Temporarily mount the partition.
+		partitionMount, err := safemount.NewMount(diskPartition.Path, bootPartitionTmpDir, diskPartition.FileSystemType, 0, "", true)
+		if err != nil {
+			return fmt.Errorf("failed to mount partition (%s):\n%w", diskPartition.Path, err)
+		}
+		defer partitionMount.Close()
```
cwize1, line 306: You need to also call `partitionMount.CleanClose()` and check its error result.

`.Close()` runs asynchronously and doesn't guarantee that the mount closed successfully. This is fine when there is already an error. But you don't want the tool to report a success if there were problems unmounting a partition.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -222,3 +233,93 @@ func customizeImageHelper(buildDir string, baseConfigPath string, config *imagec
...
+
+		grubCfgFullPath := filepath.Join(bootPartitionTmpDir, "grub2/grub.cfg")
+		grubCfgExists, err := file.PathExists(grubCfgFullPath)
+		if err != nil {
+			return fmt.Errorf("failed to stat file (%s):\n%w", grubCfgFullPath, err)
+		}
+
+		if grubCfgExists {
+			// Update grub configuration
+			err = updateGrubConfig(config.SystemConfig.Verity.DataPartition.IdType, config.SystemConfig.Verity.DataPartition.Id, config.SystemConfig.Verity.HashPartition.IdType, config.SystemConfig.Verity.HashPartition.Id, rootHash, grubCfgFullPath)
```
cwize1, line 316: Line is too long.

cwize1 (changes requested):

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/verity.go
```diff
@@ -0,0 +1,30 @@
...
+	HashPartition VerityPartition `yaml:"HashPartition"`
+}
+
+func (v *Verity) IsSet() bool {
+	return v.DataPartition != VerityPartition{} ||
+		v.HashPartition != VerityPartition{}
+}
+
+func (v *Verity) IsValid() error {
+	if err := v.DataPartition.IdType.IsValid(); err != nil || v.DataPartition.Id == "" {
```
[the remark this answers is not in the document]
liulanze: This is for checking if users have provided id. I thought we still need it?

toolkit/tools/imagecustomizerapi/systemconfig.go
```diff
@@ -94,5 +95,9 @@ func (s *SystemConfig) IsValid() error {
 		return err
 	}
 
+	if err := s.Verity.IsValid(); err != nil {
```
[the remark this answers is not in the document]
liulanze: How did you run the UTs? `go test <test_package_path>`?

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
liulanze: We need to add directly into the Kernel cmdline argument since Trident needs to consume in the later workflow.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,224 @@
...
+		return fmt.Errorf("failed to build dracut module - (%s):\n%w", dracutModuleName, err)
+	}
+
+	return nil
+}
+
+func updateMarinerCfgWithInitramfs(imageChroot *safechroot.Chroot) error {
+	var err error
+
+	initramfsPath := filepath.Join(imageChroot.RootDir(), "boot/initramfs-*")
```
vinceaperri, line 87: Shouldn't hard-code directory separator "/" in combination with filePath.Join usage.

Does this work?

filepath.Join(imageChroot.RootDir(), "boot", "initramfs-*")

vinceaperri, line 87: I'd call this a initramfsPattern not path

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+		return kernels, nil
+	}
+
+	kernelFiles, err := listKernels()
+	if err != nil {
+		return fmt.Errorf("failed to list kernels: %w", err)
+	}
+
+	if len(kernelFiles) == 0 {
```
vinceaperri, line 60: Just do 

if len(kernelFiles) != 1 {
  return fmt.Errorf("expected one kernel file, but found %d", len(kernelFiles))
}

Like you do in updateMarinerCfgWithInitramfs

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
vinceaperri: Also below

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,224 @@
...
+	var updatedLines []string
+	for _, line := range lines {
+		trimmedLine := strings.TrimSpace(line)
+		if strings.HasPrefix(trimmedLine, "linux ") {
+			// Append new arguments to the line that starts with "linux"
+			line += " " + newArgs
+		}
+		if strings.HasPrefix(trimmedLine, "set rootdevice=PARTUUID=") {
+			// Replace the root device line with the new root device. TODO: add supported type 'user'
+			line = "set rootdevice=/dev/mapper/root"
```
vinceaperri, line 159: Hardcoding the name of this device to "root" doesn't seem right to me. Should this be configurable?

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read grub config: %v", err)
+	}
+
+	var updatedLines []string
+	for _, line := range lines {
+		trimmedLine := strings.TrimSpace(line)
+		if strings.HasPrefix(trimmedLine, "linux ") {
+			// Append new arguments to the line that starts with "linux"
+			line += " " + newArgs
```
vinceaperri, line 155: This doesn't seem to be safe for multiple runs of imagecustomizer, in that it would blindly append newArgs even though the args already existed. I would find a regex search and replace more resilient. That way you can be sure there aren't unintentionally two values of, say, roothash, on the command line.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
liulanze: Thanks, fixed.

vinceaperri (approved):

toolkit/tools/imagecustomizerapi/verity.go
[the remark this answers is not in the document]
liulanze: I see, fixed.

PR #6813, opened 2023-11-20
Add support for squashfs image format

ellie-di:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

squahsfs is a compressed and read-only file system that is commonly used in embedded systems. This PR adds the support for composing a squashfs image type to Mariner toolkit.

**NO**

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/toolkit/tools/roast/formats/squashfs.go
+++ b/toolkit/tools/roast/formats/squashfs.go
@@ -4,7 +4,6 @@
 package formats
 
 import (
-	"fmt"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
 )
 
@@ -17,15 +16,8 @@ type SquashFS struct {
 
 // Convert converts the image in the squashfs format
 func (t *SquashFS) Convert(input, output string, isInputFile bool) (err error) {
-	const (
-		squashErrors = false
-	)
-
-	if !isInputFile {
-		return fmt.Errorf("squashfs conversion requires an input")
-	}
-
-	err = shell.ExecuteLive(squashErrors, "mksquashfs", input, output, "-all-root", "-wildcards", "-e", "tmp/*", "-noappend")
+	const squashErrors = false
+	err = shell.ExecuteLive(squashErrors, "mksquashfs", input, output, "-noappend")
 	return
 }
 
--- a/toolkit/tools/roast/formats/squashfs.go
+++ b/toolkit/tools/roast/formats/squashfs.go
@@ -1,0 +1,40 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package formats
+
+import (
+	"fmt"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
+)
+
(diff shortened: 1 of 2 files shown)


toolkit/tools/roast/formats/squashfs.go
```diff
@@ -0,0 +1,40 @@
...
+func (t *SquashFS) Convert(input, output string, isInputFile bool) (err error) {
+	const (
+		squashErrors = false
+	)
+
+	if !isInputFile {
+		return fmt.Errorf("squashfs conversion requires an input")
+	}
+
+	err = shell.ExecuteLive(squashErrors, "mksquashfs", input, output, "-all-root", "-wildcards", "-e", "tmp/*", "-noappend")
```
reubeno, line 28: For general image usage in the toolkit are `-all-root` and `-e tmp/*` appropriate? The former seems to clobber ownership of files/dirs in the image being converted, which may not be the right call for all images. As for excluding `tmp/*`, I'm wondering if the image format converter may not be the best place in the flow to specify that it should be removed.

toolkit/tools/roast/formats/squashfs.go
```diff
@@ -0,0 +1,40 @@
...
+type SquashFS struct {
+}
+
+// Convert converts the image in the squashfs format
+func (t *SquashFS) Convert(input, output string, isInputFile bool) (err error) {
+	const (
+		squashErrors = false
+	)
+
+	if !isInputFile {
```
reubeno, line 24: If I recall correctly, when `isInputFile` is `false`, then the input is a directory -- and capturing that into `squashfs` format should be doable (and to my understanding, required for the derivative we're working on, right?).

When it's `true`, I've forgotten; is the input a disk image (or is it a tarball)?

toolkit/tools/roast/formats/squashfs.go
ellie-di: totally agree. I've removed those options.

toolkit/tools/roast/formats/squashfs.go
```diff
@@ -0,0 +1,32 @@
...
+const SquashFSType = "squashfs"
+
+// SquashFS implements Converter interface to convert a RAW image into a squashfs file
+type SquashFS struct {
+}
+
+// Convert converts the image in the squashfs format
+func (t *SquashFS) Convert(input, output string, isInputFile bool) (err error) {
+	const squashErrors = false
+	err = shell.ExecuteLive(squashErrors, "mksquashfs", input, output, "-noappend")
```
reubeno, line 20: Can you remind me -- are there cases in which this will convert a .tar directly into a .squashfs? If so, then do we need to specify something appropriate for `-root-mode` since .tar files don't actually capture the root directory's properties? Or are we now only ever capturing from an expanded out directory?

ellie-di: It's not converting a tar to squashfs, and yes it captures an expanded directory.

reubeno (approved):

PR #6815, opened 2023-11-20
Added `cross-binutils` and `kernel-cross-headers` for AMD64.

PawelWMS:
###### Merge Checklist  

- [X] [The toolchain has been rebuilt successfully](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=458755&view=logs&j=e42c274b-c79e-5dfe-e218-6a609d4ebb40&t=d384e9fc-22bd-5f26-1bfa-60e01076455e&l=271) (or no changes were made to it)
- [X] Ready to merge

---

###### Summary 

First part of migrating #5569 to the latest `main`. We're adding packages to enable cross-compilation for the ARM64 architecture on ADM64 machines.

###### Change Log  

- Added cross-compilation packages for `binutils` and `kernel-headers`.

Yes.

###### Associated issues  

- #5569

###### Test Methodology

- Local builds.
- [Buddy build.](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=457177&view=results)
- Toolchain build in [this PR check](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=461902&view=results).


SPECS/binutils/binutils.spec
```diff
@@ -16,8 +37,29 @@ Patch2:         thin_archive_descriptor.patch
...
 Patch5:         CVE-2022-4285.patch
+# The gold linker doesn't understand the 'module_info.ld' script passed to all linkers and the tests fail to correctly link.
+Patch6:         disable_gold_test.patch
 Provides:       bundled(libiberty)
 
+Requires:       %{name}-common = %{version}-%{release}
+
+# Moving macro before the "SourceX" tags breaks PR checks parsing the specs.
+%global do_package() \
+%if %2 \
```
dallasd1, line 48: How are the %1 and %2 variables set? Is there a new build option to enabel cross building?

SPECS/kernel-headers/kernel-headers.spec
```diff
@@ -1,41 +1,80 @@
...
+Source0:        https://github.com/microsoft/CBL-Mariner-Linux-Kernel/archive/rolling-lts/mariner-2/%{version}.tar.gz#/kernel-%{version}.tar.gz
 # Historical name shipped by other distros
 Provides:       glibc-kernheaders = %{version}-%{release}
 BuildArch:      noarch
 
 %description
 The Linux API Headers expose the kernel's API for use by Glibc.
 
+%if "%{_arch}" == "x86_64"
+%package -n kernel-cross-headers
```
dallasd1, line 28: Will the cross package always get built and will we want to make that consistent with all other cross packages? It could add a significant amount of build time if so, especially for glibc and gcc.

And does it make sense to keep this package definition consistent with the cross-binutils design of moving it near the top of the file with macros?

SPECS/binutils/binutils.spec
PawelWMS: No new option, they have different values for AMD64 and ARM64 builds. Check the top of the file - that's where the arguments passed to `do_package` are defined for each architecture.

SPECS/kernel-headers/kernel-headers.spec
dallasd1: This package's build time is pretty fast, but glibc and gcc are quite long. I'm wondering if it's possible for the cross packages to be conditionally built (besides relying on the host arch). From what I understand here it looks like both the native and cross version of kernel-headers get built for AMD64. Will glibc and gcc follow this pattern and also build the native and cross packages? This isn't meant to be a blocking comment, but might be a question for the next iteration.

SPECS/binutils/binutils.spec
PawelWMS: These are macro arguments and the macro is used slightly lower to conditionally define a subpackage in this spec. Right now it's only to build `binutils-aarch64-linux-gnu` for the AMD64 architecture but I decided to follow Fedora's idea with these macros to make it easier to add new subpackages in the future and control, which architecture builds which cross-compilation components.

SPECS/kernel-headers/kernel-headers.spec
PawelWMS: We don't have any mechanism right now to conditionally disabled building subpackages within one architecture, so yes, we will always build the cross-compiling subpackages on AMD64. But that is also the way all other package builds work. The additional overhead here or in other packages is currently not a big worry.

Having said that, could you say scenario are you worried about in this case? I might be missing a crucial concern.

SPECS/binutils/binutils.spec
```diff
@@ -16,8 +37,20 @@ Patch2:         thin_archive_descriptor.patch
 Patch3:         CVE-2021-45078.patch
 Patch4:         CVE-2022-38533.patch
 Patch5:         CVE-2022-4285.patch
+# The gold linker doesn't understand the 'module_info.ld' script passed to all linkers and the tests fail to correctly link.
+Patch6:         disable_gold_test.patch
```
PawelWMS, line 41: I'm more than open for better solutions here, so we can still run these tests.

SPECS/binutils/binutils.spec
```diff
@@ -43,20 +127,75 @@ for handling compiled objects.
...
+    # Remove cross man files and ldscripts.
+    rm -rf %{buildroot}%{_mandir}/man1/$target-*
+    rm -rf %{buildroot}%{auxbin_prefix}/*/lib
+done < cross.list
+
+rm -rf %{buildroot}%{_infodir}
+find %{buildroot} -type f -name "*.la" -delete -print
+
+%if %{build_cross}
+    echo "=== INSTALL po targets ==="
```
Camelron, line 181: What is special about po?

SPECS/binutils/binutils.spec
```diff
@@ -30,10 +63,61 @@ Requires:       %{name} = %{version}
...
+%if %{build_cross}
+%package -n cross-%{name}-common
+Summary: Cross-compilation binutils documentation
+BuildArch: noarch
+
+%description -n cross-%{name}-common
+Documentation for the cross-compilation binutils package.
+%endif
+
+%do_package aarch64-linux-gnu %{build_aarch64}
```
Camelron, line 75: Just making sure I understand this; 

This invocation of %do_package creates a binutils-aarch64-linux-gnu package.

toolkit/resources/manifests/package/toolchain_x86_64.txt
```diff
@@ -136,7 +138,8 @@ intltool-0.51.0-7.cm2.noarch.rpm
 itstool-2.0.6-4.cm2.noarch.rpm
 kbd-2.2.0-1.cm2.x86_64.rpm
 kbd-debuginfo-2.2.0-1.cm2.x86_64.rpm
-kernel-headers-5.15.138.1-4.cm2.noarch.rpm
+kernel-cross-headers-5.15.138.1-5.cm2.noarch.rpm
+kernel-headers-5.15.138.1-5.cm2.noarch.rpm
```
Camelron, line 142: I think that the kernel auto-upgrade pipeline logic is somewhere in this Repo. We should update it to increment this new package.

Camelron: I think I found the file we would need to change; 

https://dev.azure.com/mariner-org/mariner/_search?action=contents&text=KernelCVE.json&type=code&lp=code-Project&filters=ProjectFilters%7Bmariner%7DRepositoryFilters%7BCBL-Mariner-Pipelines%7D&pageSize=25&result=DefaultCollection/mariner/CBL-Mariner-Pipelines/GBmaster//spec-auto-patch-update/config/KernelCVE.json

We would need to make the configuration aware of kernel-cross-headers-*. so that UpdateKernel.py catches it.

https://dev.azure.com/mariner-org/mariner/_search?action=contents&text=CblMarinerKernelRepo&type=code&lp=code-Project&filters=ProjectFilters%7Bmariner%7DRepositoryFilters%7BCBL-Mariner-Pipelines%7D&pageSize=25&result=DefaultCollection/mariner/CBL-Mariner-Pipelines/GBmaster//spec-auto-patch-update/tools/UpdateKernel.py

PawelWMS: Amazing, thank you Cameron! I'll look into this.

SPECS/binutils/binutils.spec
PawelWMS: Yes, but only if `%{build_aarch64}` is `1`.

More context: it's an idea I migrated from Fedora, which in our case is supposed to make it easier to add other architectures in the future.

SPECS/kernel-headers/kernel-headers.spec
dallasd1: As long as the additional build time isn't an issue this is a good approach.

dallasd1 (approved):

toolkit/resources/manifests/package/toolchain_x86_64.txt
PawelWMS: Here's the [update to the auto-updater](https://dev.azure.com/mariner-org/mariner/_git/CBL-Mariner-Pipelines/pullrequest/17200).

SPECS/binutils/binutils.spec
PawelWMS: We have only one set of these in the `cross-binutils-common` package, which is shared across all cross-compilation packages in this spec (only one for now). We also don't build and install it for architectures, where we won't be providing cross-compilation (ARM64 to be more precise). And that's why this section is special-cased.

SPECS/kernel-headers/kernel-headers.spec
```diff
@@ -1,41 +1,82 @@
...
 
 %description
 The Linux API Headers expose the kernel's API for use by Glibc.
 
+%if %{build_cross}
+%package -n kernel-cross-headers
+Summary: Header files for the Linux kernel for use by cross-glibc.
+
+%description -n kernel-cross-headers
+Kernel-cross-headers includes the C header files that specify the interface
```
Camelron, line 34: Can you specify what is different about these headers relative to the standard headers? I can't really tell what this accomplishes for us currently.

Camelron: And why one "cross" package is sufficient when there should really be one for each value in %cross_archs

Camelron: Ah, looks like cross will contain the set of all non-x86 headers.

toolkit/resources/manifests/package/toolchain_x86_64.txt
```diff
@@ -136,7 +138,8 @@ intltool-0.51.0-7.cm2.noarch.rpm
 itstool-2.0.6-4.cm2.noarch.rpm
 kbd-2.2.0-1.cm2.x86_64.rpm
 kbd-debuginfo-2.2.0-1.cm2.x86_64.rpm
-kernel-headers-5.15.138.1-4.cm2.noarch.rpm
+kernel-cross-headers-5.15.138.1-5.cm2.noarch.rpm
```
Camelron, line 141: Why do we need cross-headers exclusively in toolchain_x86?

Camelron (approved):
Looks good to me, left some minor comments/questions.

SPECS/kernel-headers/kernel-headers.spec
PawelWMS: Yes, that's correct. Here I also followed Fedora's approach, which seems reasonable as headers aren't that big and having to use separate ones for each architecture would probably end up being a headache down the road.

toolkit/resources/manifests/package/toolchain_x86_64.txt
PawelWMS: They are not built for ARM64 because for now we want to keep it simple and start with cross-compiling on the AMD64 architecture.

SPECS/binutils/binutils.spec
```diff
@@ -30,10 +63,61 @@ Requires:       %{name} = %{version}
...
+    local target=$1
+    local condition=$2
+
+    if [ $condition != 0 ]
+    then
+        echo $1 >> cross.list
+    fi
+}
+
+touch cross.list
```
neha170, line 90: Just wondering if this file, cross.list, is deleted after build. We don't want it packaged into the rpm right?

PR #6821, opened 2023-11-21
toolkit image build: Fix make error for config files outside imageconfig directory

neha170:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

If config file lies outside the imageconfig/ directory, makefile is not able to find it, and errors out with '*** multiple target patterns. Stop'. This change modifies makefile to extract all .json and .sh filenames from config_name, fixes their paths relative to the makefile, and watches them to set imageconfigvalidator flag.

###### Change Log  

- Extract config directory for CONFIG_FILE into CONFIG_BASE_DIR
- Extract all filenames with .sh or .json type in CONFIG_FILE
- Fix paths to these files wrt the makefile using CONFIG_BASE_DIR
- Pass this list of paths to imageconfigvalidator flag to watch for changes

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/44648678

###### Test Methodology

- Built image successfully with `sudo make image CONFIG_FILE=../core-efi.json` and `sudo make image CONFIG_FILE=imagecondigs/core-efi.json` and `sudo make image CONFIG_FILE=complete/path/to/core-efi.json`
- Changes in any config related files (CONFIG_FILE, or any .json or .sh file referenced in CONFIG_FILE are recognized by imageconfigvalidator flag, and imageconfigvalidator is therefore run when any of these file(s) are changed.

### DIFF
--- a/toolkit/scripts/get_config_deps.sh
+++ b/toolkit/scripts/get_config_deps.sh
@@ -2,26 +2,32 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-# $1 - CONFIG_FILE
-# $2 - CONFIG_BASE_DIR
+# $1 - config_file
 
-if [[ $# -ne 2 ]]
+if [[ $# -ne 1 ]]
 then
-    echo "ERROR: Must provide CONFIG_FILE and CONFIG_BASE_DIR to get_config_deps.sh" >&2
+    echo "ERROR: Must provide config file to get_config_deps.sh" >&2
     exit 1
 fi
 
-CONFIG_FILE="$1"
-CONFIG_BASE_DIR="$2"
+config_file="$1"
 
-config_other_files=$(grep -E '.json|.sh' $CONFIG_FILE| sed 's/\"Path\"\://' | tr ", \n" " " | tr "\"" " " | xargs)
+if [[ ! -f "$config_file" ]]
+then
+    echo "ERROR: Config file '$config_file' does not exist" >&2
+    exit 1
+fi
+
+config_base_dir=$(dirname "$config_file")
+
+config_other_files=$(grep -E '.json|.sh' $config_file| sed 's/\"Path\"\://' | tr ", \n" " " | tr "\"" " " | xargs)
 for filename in $config_other_files
 do
-    # fix path if it's relative to CONFIG_FILE
+    # fix path if it's relative to config_file
 	if [ "${filename:0:1}" == "/" ]
 	then
-		echo "$filename"
+		echo $(realpath "$filename")
 	else
-		echo "$CONFIG_BASE_DIR$filename"
+		echo $(realpath "$config_base_dir/$filename")
 	fi
 done
--- a/toolkit/scripts/get_config_deps.sh
+++ b/toolkit/scripts/get_config_deps.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# $1 - CONFIG_FILE
+# $2 - CONFIG_BASE_DIR
+
+if [[ $# -ne 2 ]]
+then
+    echo "ERROR: Must provide CONFIG_FILE and CONFIG_BASE_DIR to get_config_deps.sh" >&2
+    exit 1
+fi
+
+CONFIG_FILE="$1"
+CONFIG_BASE_DIR="$2"
+
+config_other_files=$(grep -E '.json|.sh' $CONFIG_FILE| sed 's/\"Path\"\://' | tr ", \n" " " | tr "\"" " " | xargs)
+for filename in $config_other_files
+do
+    # fix path if it's relative to CONFIG_FILE
+	if [ "${filename:0:1}" == "/" ]
+	then
+		echo "$filename"
+	else
+		echo "$CONFIG_BASE_DIR$filename"
+	fi
+done
--- a/toolkit/scripts/imggen.mk
+++ b/toolkit/scripts/imggen.mk
@@ -5,8 +5,7 @@ $(call create_folder,$(IMAGEGEN_DIR))
 
 # Resources
 config_name              = $(notdir $(CONFIG_FILE:%.json=%))
-CONFIG_BASE_DIR          = $(dir $(CONFIG_FILE:%.json=%))
-config_other_files       = $(if $(CONFIG_FILE),$(shell $(SCRIPTS_DIR)/get_config_deps.sh $(CONFIG_FILE) $(CONFIG_BASE_DIR)),)
+config_other_files       = $(if $(CONFIG_FILE),$(shell $(SCRIPTS_DIR)/get_config_deps.sh $(CONFIG_FILE)),)
 assets_dir               = $(RESOURCES_DIR)/assets/
 assets_files             = $(call shell_real_build_only, find $(assets_dir))
 imggen_local_repo        = $(MANIFESTS_DIR)/image/local.repo
--- a/toolkit/scripts/imggen.mk
+++ b/toolkit/scripts/imggen.mk
@@ -6,12 +6,7 @@ $(call create_folder,$(IMAGEGEN_DIR))
 # Resources
 config_name              = $(notdir $(CONFIG_FILE:%.json=%))
 CONFIG_BASE_DIR          = $(dir $(CONFIG_FILE:%.json=%))
-config_other_files      := $(shell grep -E '.json|.sh' $(CONFIG_FILE)| sed 's/\"Path\"\://' | tr ", \n" " " | tr "\"" " " | xargs)
-define fix_filenames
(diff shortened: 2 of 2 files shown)


toolkit/scripts/imggen.mk
```diff
@@ -5,7 +5,8 @@ $(call create_folder,$(IMAGEGEN_DIR))
 
 # Resources
 config_name              = $(notdir $(CONFIG_FILE:%.json=%))
-config_other_files       = $(if $(CONFIG_FILE),$(call shell_real_build_only, find $(CONFIG_BASE_DIR)))
+config_dir_name          = $(dir $(CONFIG_FILE:%.json=%))
+config_other_files       = $(if $(CONFIG_FILE),$(call shell_real_build_only, find $(config_dir_name) -type f -name $(config_name)))
```
dmcilvaney, line 9: This will not capture things like the `my_postinstall.sh` which might be next to the config.json. I think the more robust solution here is to use `jq` or a similar tool to parse the config, and gather a list of files to monitor. (install scripts, additional files, package lists, etc.).

toolkit/scripts/imggen.mk
```diff
@@ -5,7 +5,13 @@ $(call create_folder,$(IMAGEGEN_DIR))
 
 # Resources
 config_name              = $(notdir $(CONFIG_FILE:%.json=%))
-config_other_files       = $(if $(CONFIG_FILE),$(call shell_real_build_only, find $(CONFIG_BASE_DIR)))
+CONFIG_BASE_DIR          = $(dir $(CONFIG_FILE:%.json=%))
+config_other_files      := $(shell grep -E '.json|.sh' $(CONFIG_FILE)| sed 's/\"Path\"\://' | tr ", \n" " " | tr "\"" " " | xargs)
```
dmcilvaney, line 9: I'm still leaning towards a dedicated script for this, say `./toolkit/scripts/get_config_deps.sh` or something? I feel like makefile functions are something we want to do sparingly if possible.

Also, I tried to poke at `jq` quickly, its honestly not too bad for this use case.
```bash
config_other_files_relative="$(jq -r '.SystemConfigs[]?.PackageLists[]' "${config_file}")"
config_other_files_relative="${config_other_files_relative} $(jq -r '.SystemConfigs[]?.PostInstallScripts[]?.Path' "${config_file}")"
```

The one caveat is that we haven't traditionally required `jq` as a build prerequisite. I'm thinking, especially if this lives in a shell script, that we could take both approaches. If `jq` is available we use it, otherwise fall back to the grep/sed approach.

toolkit/scripts/imggen.mk
```diff
@@ -5,7 +5,13 @@ $(call create_folder,$(IMAGEGEN_DIR))
...
+define fix_filenames
+$(call shell, if [ "$1" = "/" ]; then echo $2; else echo $(CONFIG_BASE_DIR)$2 | tr -d '[:blank:]'; fi)
+endef
+config_other_files_fixed := $(foreach filename,$(config_other_files),$(call fix_filenames,$(shell echo $(filename) | head -c 1), $(filename)))
```
dmcilvaney, line 13: Iff this moves to a script, ``$(call shell_real_build_only, ...)`` might be a good fit. That means we don't execute it for dry-runs (like tab-completing the makefile options). If you keep the current `define` function approach I its probably simpler to just leave it as a normal shell

toolkit/scripts/imggen.mk
dmcilvaney, line 9: Is this going to behave ok with `CONFIG_FILE=""`? Need to make sure that we don't end up with an error message or something in `config_other_files` which Make will then try to find real files for.

dmcilvaney: (I'm aware I've added more than my fair share of indecipherable makefile functions... I'm trying to convert those to scripts where it makes sense as I update them).

toolkit/scripts/imggen.mk
```diff
@@ -5,7 +5,8 @@ $(call create_folder,$(IMAGEGEN_DIR))
 
 # Resources
 config_name              = $(notdir $(CONFIG_FILE:%.json=%))
-config_other_files       = $(if $(CONFIG_FILE),$(call shell_real_build_only, find $(CONFIG_BASE_DIR)))
+CONFIG_BASE_DIR          = $(dir $(CONFIG_FILE:%.json=%))
```
dmcilvaney, line 8: This could probably be moved into the script, do a `config_file_base_dir=$(dirname "$config_file")`

toolkit/scripts/get_config_deps.sh
```diff
@@ -0,0 +1,27 @@
...
+if [[ $# -ne 2 ]]
+then
+    echo "ERROR: Must provide CONFIG_FILE and CONFIG_BASE_DIR to get_config_deps.sh" >&2
+    exit 1
+fi
+
+CONFIG_FILE="$1"
+CONFIG_BASE_DIR="$2"
+
+config_other_files=$(grep -E '.json|.sh' $CONFIG_FILE| sed 's/\"Path\"\://' | tr ", \n" " " | tr "\"" " " | xargs)
```
dmcilvaney, line 17: If you want to go the optional `jq` route, I tried this again and as far as I can tell it does work in a script (wonder if it was just an issue with make..., wouldn't surprise me).

I like using `jq` since its more resilient than grep/sed, but I think making it optional for now isn't a bad idea.

```suggestion
if jq --version &> /dev/null
then
    config_other_files="$(jq -r '.SystemConfigs[]?.PackageLists[]' "${config_file}")"
    config_other_files="${config_other_files_relative} $(jq -r '.SystemConfigs[]?.PostInstallScripts[]?.Path' "${config_file}")"
else
    config_other_files=$(grep -E '.json|.sh' $CONFIG_FILE| sed 's/\"Path\"\://' | tr ", \n" " " | tr "\"" " " | xargs)
fi
```

toolkit/scripts/get_config_deps.sh
```diff
@@ -0,0 +1,27 @@
...
+# $1 - CONFIG_FILE
+# $2 - CONFIG_BASE_DIR
+
+if [[ $# -ne 2 ]]
+then
+    echo "ERROR: Must provide CONFIG_FILE and CONFIG_BASE_DIR to get_config_deps.sh" >&2
+    exit 1
+fi
+
+CONFIG_FILE="$1"
```
dmcilvaney, line 14: Should also check that the file does in fact exist.
```suggestion
CONFIG_FILE="$1"
if [[ ! -f "$CONFIG_FILE" ]]
then
    echo "ERROR: Config file '$CONFIG_FILE' does not exist" >&2
    exit 1
fi
```

toolkit/scripts/get_config_deps.sh
```diff
@@ -0,0 +1,27 @@
...
+
+config_other_files=$(grep -E '.json|.sh' $CONFIG_FILE| sed 's/\"Path\"\://' | tr ", \n" " " | tr "\"" " " | xargs)
+for filename in $config_other_files
+do
+    # fix path if it's relative to CONFIG_FILE
+	if [ "${filename:0:1}" == "/" ]
+	then
+		echo "$filename"
+	else
+		echo "$CONFIG_BASE_DIR$filename"
```
dmcilvaney, line 25: Personal preference, but I like to convert to absolute paths when generating paths like this.
```suggestion
		echo $(realpath "$filename")
	else
		echo $(realpath "$CONFIG_BASE_DIR$filename")
```

toolkit/scripts/imggen.mk
```diff
@@ -5,8 +5,7 @@ $(call create_folder,$(IMAGEGEN_DIR))
 
 # Resources
 config_name              = $(notdir $(CONFIG_FILE:%.json=%))
-CONFIG_BASE_DIR          = $(dir $(CONFIG_FILE:%.json=%))
-config_other_files       = $(if $(CONFIG_FILE),$(shell $(SCRIPTS_DIR)/get_config_deps.sh $(CONFIG_FILE) $(CONFIG_BASE_DIR)),)
+config_other_files       = $(if $(CONFIG_FILE),$(shell $(SCRIPTS_DIR)/get_config_deps.sh $(CONFIG_FILE)),)
```
dmcilvaney, line 8:
```suggestion
config_other_files       = $(if $(CONFIG_FILE),$(call shell_real_build_only, $(SCRIPTS_DIR)/get_config_deps.sh $(CONFIG_FILE)),)
```

dmcilvaney: This function will only execute if its a "real" build, I don't think we care about any of these outputs when its a dry-run (ie tab completing).

toolkit/scripts/get_config_deps.sh
dmcilvaney: Offline: won't bother with jq, not worth the complexity

dmcilvaney (approved):

PR #6824, opened 2023-11-22
Add Initial Mariner Image Modifier Files

elainezhao96:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 
Add initial Mariner Image modifier(EMU) files 

###### Change Log  
- Created chroot interface and a dummy implementation for in-OS scenarios
- Add the SSHPubKeys to user struct as requested by Trident crew, added it to `ProvisionUserSSHCerts` function accordingly

NO

###### Test Methodology
Inside imagemodifier dir, run `go build -o imagemodifier .`
Will do an end-to-end test in trident to validate we can set up users correctly.


toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1489,11 +1489,12 @@ func ProvisionUserSSHCerts(installChroot *safechroot.Chroot, username string, ss
 
 	// Skip user SSH directory generation when not provided with public keys
 	// Let SSH handle the creation of this folder on its first use
-	if len(sshPubKeyPaths) == 0 {
-		return
-	}
+	if len(sshPubKeyPaths) == 0 && len(sshPubKeys) == 0 {
+        return nil
```
cwize1, line 1493: Be careful with changing formatting/whitespace. It isn’t a bad idea to use `go fmt` before committing. If you use VSCode, it will auto format whenever you save a file.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1582,13 +1596,10 @@ func ProvisionUserSSHCerts(installChroot *safechroot.Chroot, username string, ss
 		return
 	})
 
-	if err != nil {
-		return
-	}
-
-	return
+    return err
```
cwize1, line 1599: Probably goodness to avoid changing code when it isn’t neccessary.

toolkit/tools/imagecustomizerapi/user.go
```diff
@@ -17,6 +17,7 @@ type User struct {
 	PasswordPath        string   `yaml:"PasswordPath"`
 	PasswordExpiresDays *int64   `yaml:"PasswordExpiresDays"`
 	SSHPubKeyPaths      []string `yaml:"SSHPubKeyPaths"`
+	SSHPubKeys          []string `yaml:"SSHPubKeys"`
```
cwize1, line 20: Adding this should probably be done in a separate change.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -661,3 +660,37 @@ func extractWorkerTar(chroot string, workerTar string) (err error) {
 	_, _, err = shell.Execute("tar", "-I", gzipTool, "-xf", workerTar, "-C", chroot)
 	return
 }
+
+type ChrootInterface interface {
+    RootDir() string
+	Initialize(tarPath string, extraDirectories []string, extraMountPoints []*MountPoint) error
```
cwize1, line 666: `Initialize()` probably shouldn’t be a part of this interface.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -661,3 +660,37 @@ func extractWorkerTar(chroot string, workerTar string) (err error) {
...
+	// Only execute the function, no chroot operations
+    return toRun()
+}
+
+func (d DummyChroot) UnsafeRun(toRun func() error) (err error) {
+    return toRun()
+}
+
+func (d DummyChroot) AddFiles(filesToCopy ...FileToCopy) (err error) {
+    return nil
```
cwize1, line 694: Why does this not have an implementation?

cwize1 (changes requested):
(See other comments.)

toolkit/tools/imagecustomizerapi/user.go
elainezhao96: ok will include it in a separate pr

romoh (commented):
LGTM

One comment thou, what's our logging strategy here? how do we ensure we have enough logging to investigate customization failures?

reubeno:
Naive question -- what's the difference between an image modifier and an image customizer? Without additional context, they sound like quite the same thing.

romoh:
> Naive question -- what's the difference between an image modifier and an image customizer? Without additional context, they sound like quite the same thing.

Image customizer is an officially supported CLI tool for customizing images (official releases, guaranteed backward compatibility, public documentation... etc.), while image modifier is more of a utility tool/wrapper on the customization lib that will be leveraged for ad-hoc image customizations scenarios and other efforts under development. Hence the separation to allow for a more fluid changing interface without breaking the official tool.

elainezhao96:
> LGTM
> 
> One comment thou, what's our logging strategy here? how do we ensure we have enough logging to investigate customization failures?

The logging is the same with MIC, e.g. when passing a yam into it, it verifies the content and throw error when there is any step wrong while creating a new user.  the logging of the invocation of this binary resides in Trident repo.

elainezhao96:
> Naive question -- what's the difference between an image modifier and an image customizer? Without additional context, they sound like quite the same thing.

image modifier does not create chroot env and it will do in-OS change without an input(base) image or output an image.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -661,3 +661,51 @@ func extractWorkerTar(chroot string, workerTar string) (err error) {
...
+
+// DummyChroot is a placeholder that implements ChrootInterface.
+type DummyChroot struct {
+}
+
+func (d DummyChroot) RootDir() string {
+	return "/"
+}
+
+func (d DummyChroot) Initialize(tarPath string, extraDirectories []string, extraMountPoints []*MountPoint) error {
```
cwize1, line 680: This function doesn't appear to be used. Delete?

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -661,3 +661,51 @@ func extractWorkerTar(chroot string, workerTar string) (err error) {
...
+	Run(toRun func() error) error
+	UnsafeRun(toRun func() error) error
+	AddFiles(filesToCopy ...FileToCopy) error
+}
+
+// DummyChroot is a placeholder that implements ChrootInterface.
+type DummyChroot struct {
+}
+
+func (d DummyChroot) RootDir() string {
```
cwize1, line 676: nit: `d *DummyChroot`

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -661,3 +661,51 @@ func extractWorkerTar(chroot string, workerTar string) (err error) {
...
+func (d DummyChroot) Run(toRun func() error) (err error) {
+	// Only execute the function, no chroot operations
+	return toRun()
+}
+
+func (d DummyChroot) UnsafeRun(toRun func() error) (err error) {
+	return toRun()
+}
+
+func (d DummyChroot) AddFiles(filesToCopy ...FileToCopy) (err error) {
```
cwize1, line 694: This code seems to be identical to the `Chroot` variant. It might be goodness to figure out a way for them to share code.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -661,3 +661,51 @@ func extractWorkerTar(chroot string, workerTar string) (err error) {
 	_, _, err = shell.Execute("tar", "-I", gzipTool, "-xf", workerTar, "-C", chroot)
 	return
 }
+
+type ChrootInterface interface {
```
cwize1, line 665: I would place `ChrootInterface` and `DummyChroot` in their own files (but still within this package/directory).

cwize1 (changes requested):

cwize1:
> > Naive question -- what's the difference between an image modifier and an image customizer? Without additional context, they sound like quite the same thing.
> 
> image modifier does not create chroot env and it will do in-OS change without an input(base) image or output an image.

Perhaps we should name it "OS modifier" instead of "image modifier", since it isn't modifying an image file.

elainezhao96:
> > > Naive question -- what's the difference between an image modifier and an image customizer? Without additional context, they sound like quite the same thing.
> > 
> > 
> > image modifier does not create chroot env and it will do in-OS change without an input(base) image or output an image.
> 
> Perhaps we should name it "OS modifier" instead of "image modifier", since it isn't modifying an image file.

make sense, as it modifies makes live changes to an active OS, I will update all the references to os modifier

toolkit/tools/imagemodifier/main.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package main
```
dmcilvaney, line 4: Generally we use `./toolkit/tools/<toolname>/<toolname>.go` four our tools.
Also, please add this too to `go_tool_list` in `tools.mk`. This will let the makefile build the tool if needed (you can do `make go-imagemodifier REBUILD_TOOLS=y` to build it, and if needed refer to it elsewhere in the makefiles with `$(go-imagemodifier)`.

elainezhao96: resolved the second comment, for the first comment, the imagecustomizer.go resides in ./toolkit/tools/pkg/imagecustomizerlib, and the main.go is in ./toolkit/tools/imagecustomizer, I thought this is by design, so the osmodifier follows suit.

romoh (approved):

cwize1 (approved):

toolkit/tools/imagemodifier/main.go
dmcilvaney: ok, not a big deal either way.

dmcilvaney (approved):

PR #6833, opened 2023-11-22
[3.0] Add cpupower.service to kernel-tools

rlmenge:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

The 'cpupower' systemd startup script is useful for users to change the behavior of cpufreq at runtime rather than relying on the config default set at compile time. Specifically, it allows users to change the scaling governor. By default Azure Linux sets the scaling governor to performance mode via `CONFIG_CPU_FREQ_DEFAULT_GOV_PERFORMANCE`. This was confirmed by flashing the latest CBL-Mariner 2.0 iso onto an Intel NUC. It by default used the performance scaling governor and the cpufreq driver was intel_pstates. The powersave governor was available but not used. More on CPU frequency scaling can be seen at [kernel.org
](https://www.kernel.org/doc/html/v5.15/admin-guide/pm/cpufreq.html)

While useful, introducing the cpupower.service script in Mariner 2.0 would break the assumption for how users can affect the cpu frenquency. Adding the service file would change the cpu frequency for users who have taken action to change from the default "performance" governor to some other governor (powersave, ondemand, etc.). Set in 3.0 to encourage users to have the most performant state.

Just as a note that there are many places the cpu frequency setting can be altered
- bios
- governor in kernel configs ([CONFIG_CPU_FREQ_DEFAULT_GOV_PERFORMANCE=y](https://github.com/microsoft/CBL-Mariner/blob/main/SPECS/kernel/config#L605) in Mariner)
- governor in kernel command line ([cpufreq.default_governor](https://www.kernel.org/doc/html/v5.15/admin-guide/kernel-parameters.html#:~:text=cpufreq.default_governor%3D) not set in Mariner)
- cpufreq direct call in userspace
- [tuned ](https://github.com/microsoft/CBL-Mariner/blob/main/SPECS/tuned/tuned.spec)(which we don't ship with by default)

###### Change Log  

- Add cpupower.service to kernel-tools

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/47395198/

###### Test Methodology

Prep: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=469294&view=results
---Old---
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=467857&view=results
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=467929&view=results
- RPMS: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=459543&view=logs&j=75cba72d-09cb-503a-3927-4ab61ac4a179&t=460b7289-1832-5507-1175-ad5f30576086&l=5553
- rebase full: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=465381&view=results


SPECS/kernel/kernel.spec
```diff
@@ -220,6 +223,12 @@ install -vdm 700 %{buildroot}/boot
 install -vdm 755 %{buildroot}%{_defaultdocdir}/linux-%{uname_r}
 install -vdm 755 %{buildroot}%{_prefix}/src/linux-headers-%{uname_r}
 install -vdm 755 %{buildroot}%{_libdir}/debug/lib/modules/%{uname_r}
+
+install -d -m 755 %{buildroot}%{_sysconfdir}/sysconfig
+install -c -m 644 %{SOURCE5} %{buildroot}/%{_sysconfdir}/sysconfig/cpupower
+install -d -m 755 %{buildroot}%{_unitdir}
```
Camelron, line 229: nit: I think these two lines to make the dir, install in said dir can be combined by using the -D option.

install -D -c -m 644 %{SOURCE6} %{buildroot}%{_unitdir}/cpupower.service

SPECS/kernel/kernel.spec
```diff
@@ -321,6 +333,7 @@ then
      fi
 fi
 %grub2_postun
+%systemd_postun cpupower.service
```
ddstreetmicrosoft, line 336: this should be in a `%postun tools` scriptlet, not the base `%postun` scriptlet, right?

SPECS/kernel/kernel.spec
```diff
@@ -220,6 +223,12 @@ install -vdm 700 %{buildroot}/boot
 install -vdm 755 %{buildroot}%{_defaultdocdir}/linux-%{uname_r}
 install -vdm 755 %{buildroot}%{_prefix}/src/linux-headers-%{uname_r}
 install -vdm 755 %{buildroot}%{_libdir}/debug/lib/modules/%{uname_r}
+
+install -d -m 755 %{buildroot}%{_sysconfdir}/sysconfig
+install -c -m 644 %{SOURCE5} %{buildroot}/%{_sysconfdir}/sysconfig/cpupower
+install -d -m 755 %{buildroot}%{_unitdir}
+install -c -m 644 %{SOURCE6} %{buildroot}%{_unitdir}/cpupower.service
```
ddstreetmicrosoft, line 230: as @Camelron pointed out you can leave out the `install` lines for the dirs and just use `-D` when installing the files, unless you are specifically creating the dirs separate for a reason?

SPECS/kernel/kernel.spec
```diff
@@ -404,6 +420,8 @@ ln -sf linux-%{uname_r}.cfg /boot/mariner.cfg
 %{_libdir}/perf/examples/bpf/*
 %{_libdir}/perf/include/bpf/*
 %{_includedir}/perf/perf_dlfilter.h
+%{_unitdir}/cpupower.service
+%{_sysconfdir}/sysconfig/cpupower
```
ddstreetmicrosoft, line 424: this should be `config(noreplace)` shouldn't it?

ddstreetmicrosoft (changes requested):
mostly looks good, just a few comments

SPECS/kernel/kernel.spec
rlmenge: I did consider this initially but not looking to not change the permissions of the directories.

SPECS/kernel/kernel.spec
rlmenge: 100%. changing

ddstreetmicrosoft (approved):

Camelron (approved):
lgtm

PR #6864, opened 2023-11-29
Anphel/add cloud native repo

anphel31:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


SPECS/mariner-repos/mariner-cloud-native.repo
```diff
@@ -0,0 +1,29 @@
+[mariner-official-cloud-native]
+name=CBL-Mariner Official Cloud Native $releasever $basearch
+baseurl=https://packages.microsoft.com/cbl-mariner/$releasever/prod/cloud-native/$basearch
+gpgkey=file:///etc/pki/rpm-gpg/MICROSOFT-RPM-GPG-KEY file:///etc/pki/rpm-gpg/MICROSOFT-METADATA-GPG-KEY
+gpgcheck=1
+repo_gpgcheck=1
+enabled=0
```
PawelWMS, line 7: One concern here: so far we've been advertising that the way to enable access to a specific Mariner repo was to `tdnf install <repo_package>` and that was it. All other ones are enabled by default (I mean the ones with "regular" packages, not debuginfo ones or sources). This deviation from that approach might introduce some confusion and I'm curious, if we can avoid it making this repo enabled by default as well.

Also, right now what is the point of installing the new package inside the `core-packages`, if none of repos it provides are enabled by default? We're making all base images slightly bigger and the user still has to perform some manual work to enable the new repo, but this time they can't use the known `tdnf install` way.

christopherco: I share @PawelWMS's concerns, especially since this is for 2.0 which is our stable release. Adding this new package repo should be user opt-in via installing the repo package, rather than a forced default change to every 2.0 production instance.

A change in defaults like this is more suitable for our 3.0 development branch

christopherco (changes requested):

SPECS/mariner-repos/mariner-cloud-native.repo
anphel31: That makes sense. I've updated the PR to remove cloud-native repos from the core image.

PawelWMS (approved):

christopherco (approved):

PR #6874, opened 2023-11-30
Blocksize 4096

ddaney-fungible:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Enable building of disk images with 4096 byte block size.

Some of the systems we want to run Mariner on are either more efficient when using 4096 byte block size storage devices, or in some cases, 4096 is the only supported block size.

###### Change Log  

- Add optional "BlockSize" field to Disk parameters and propagate it to losetup invocations.
- Add some documentation on the fields in the Disk section of the config file.

YES

###### Test Methodology

- Local builds on 2.0 branch.

### DIFF
--- a/toolkit/tools/imagegen/configuration/disk.go
+++ b/toolkit/tools/imagegen/configuration/disk.go
@@ -19,6 +19,7 @@ import (
 type Disk struct {
 	PartitionTableType PartitionTableType `json:"PartitionTableType"`
 	MaxSize            uint64             `json:"MaxSize"`
+	BlockSize          uint32             `json:"BlockSize"`
 	TargetDisk         TargetDisk         `json:"TargetDisk"`
 	Artifacts          []Artifact         `json:"Artifacts"`
 	Partitions         []Partition        `json:"Partitions"`
@@ -81,6 +82,9 @@ func checkMaxSizeCorrectness(disk *Disk) (err error) {
 
 // IsValid returns an error if the PartitionTableType is not valid
 func (d *Disk) IsValid() (err error) {
+	if d.BlockSize != 0 && d.BlockSize != 512 && d.BlockSize != 4096 {
+		return fmt.Errorf("invalid [BlockSize]: %d. Must be 0, 512, or 4096", d.BlockSize)
+	}
 	if err = d.PartitionTableType.IsValid(); err != nil {
 		return fmt.Errorf("invalid [PartitionTableType]: %w", err)
 	}
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -239,10 +239,56 @@ func CreateSparseDisk(diskPath string, size uint64, perm os.FileMode) (err error
 	return
 }
 
+// devineBlockSize tries to determine the block size used by the file system image
+func devineBlockSize(diskFilePath string) uint32 {
+	var blockSize uint32 = 512 // Default to 512
+	diskFile, err := os.Open(diskFilePath)
+	if err != nil {
+		logger.Log.Warnf("Failed to open %v to devine block size, assuming %d.", diskFilePath, blockSize)
+		return blockSize
+	}
+	defer diskFile.Close()
+
+	var probePoint int64 = 4096
+	signature := make([]byte, 8)
+	_, err = diskFile.ReadAt(signature, probePoint)
+	if err != nil {
+		logger.Log.Warnf("Failed to read from %v to devine block size, assuming %d.", diskFilePath, blockSize)
+		return blockSize
+	}
+	// GPT header starts with 8 bytes of ascii "EFI PART"
+	if signature[0] == 0x45 &&
+		signature[1] == 0x46 &&
+		signature[2] == 0x49 &&
+		signature[3] == 0x20 &&
+		signature[4] == 0x50 &&
+		signature[5] == 0x41 &&
+		signature[6] == 0x52 &&
+		signature[7] == 0x54 {
+		logger.Log.Debugf("Found GPT signature at %d in: %v, assuming that block size", probePoint, diskFilePath)
+		blockSize = uint32(probePoint)
+	}
+	return blockSize
+}
+
 // SetupLoopbackDevice creates a /dev/loop device for the given disk file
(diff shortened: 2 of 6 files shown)


dmcilvaney:
I'll have a look through this when I have a chance, but fyi I had a half-baked version of this here: https://github.com/microsoft/CBL-Mariner/pull/5818/files

I didn't bother with trying to read the underlying block size when creating the loopbacks and just took the default.

Priorities changed so I never finished it up, happy to just go forward with this PR.

ddaney-fungible:
> I'll have a look through this when I have a chance, but fyi I had a half-baked version of this here: https://github.com/microsoft/CBL-Mariner/pull/5818/files
> 
> I didn't bother with trying to read the underlying block size when creating the loopbacks and just took the default.
> 
> Priorities changed so I never finished it up, happy to just go forward with this PR.

Interesting.

Setting the block size used by a file system is somewhat orthogonal to the block size of the underlying storage device, as the file system block size might be any of several different multiples of the device block size.  Also, all partitions on a single device will be on a device block boundary, which would probably be a common minimum block size for any file systems resident on the partitions.

The system partitioning and file system creation tools will all choose usable (although perhaps not optimal) file system block sizes based on the underlying device block size.

Summary: I think both a version of your patch and this one would be useful.

The main motivation for this patch is so that I can create multi-partition Disk images that can be copied into an NVMe namespace that cannot be changed to 512 byte block size.  I think it may be of more general interest as many modern storage systems work better, or are more efficient, when configured for 4096 byte access.

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -239,10 +239,56 @@ func CreateSparseDisk(diskPath string, size uint64, perm os.FileMode) (err error
...
+
+	var probePoint int64 = 4096
+	signature := make([]byte, 8)
+	_, err = diskFile.ReadAt(signature, probePoint)
+	if err != nil {
+		logger.Log.Warnf("Failed to read from %v to devine block size, assuming %d.", diskFilePath, blockSize)
+		return blockSize
+	}
+	// GPT header starts with 8 bytes of ascii "EFI PART"
+	if signature[0] == 0x45 &&
```
cwize1, line 260: Use `bytes.Equal`

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -239,10 +239,56 @@ func CreateSparseDisk(diskPath string, size uint64, perm os.FileMode) (err error
...
+		blockSize = uint32(probePoint)
+	}
+	return blockSize
+}
+
 // SetupLoopbackDevice creates a /dev/loop device for the given disk file
-func SetupLoopbackDevice(diskFilePath string) (devicePath string, err error) {
-	logger.Log.Debugf("Attaching Loopback: %v", diskFilePath)
-	stdout, stderr, err := shell.Execute("losetup", "--show", "-f", "-P", diskFilePath)
+func SetupLoopbackDevice(diskFilePath string, disk *configuration.Disk) (devicePath string, err error) {
```
cwize1, line 275: Pass in the `BlockSize` as a `uint64` (or `*uint64`) instead of taking in the entire `Disk` object. This will make the function easier to use in other contexts. e.g. the imager customizer tool.

toolkit/tools/imagegen/configuration/disk.go
```diff
@@ -81,6 +82,9 @@ func checkMaxSizeCorrectness(disk *Disk) (err error) {
 
 // IsValid returns an error if the PartitionTableType is not valid
 func (d *Disk) IsValid() (err error) {
+	if d.BlockSize != 0 && d.BlockSize != 512 && d.BlockSize != 4096 {
```
cwize1, line 85: nit: Use a `switch` statement.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -156,7 +156,7 @@ func createOverlayPartition(partitionSetting configuration.PartitionSetting, mou
 	//Mount the base image
 	//Create a temp upper dir
 	//Add to the mount args
-	devicePath, err := diskutils.SetupLoopbackDevice(partitionSetting.OverlayBaseImage)
+	devicePath, err := diskutils.SetupLoopbackDevice(partitionSetting.OverlayBaseImage, nil)
```
cwize1, line 159: You probably want to make sure that the overlay image is using the same block size as the disk it is overlaying.

toolkit/tools/imagegen/diskutils/diskutils.go
ddaney-fungible: My idea here was to not have to de-encapsulate magic parameters out of the disk configuration before we get to the point where the parameters are consumed. If it is very objectionable, I guess we can pull out the block size farther up the call chain and pass it instead of (a pointer to) the configuration object.

Any manipulation of the disk image after it is partitioned, should use the same configuration as the disk image was created with.

In the case of the image customizer, I think the original configuration.Disk may not be available, so passing nil is really no worse than a hypothetical value for the block size that would represent do-the-right-thing-because-we-don't-know-the-block-size

toolkit/tools/imagegen/diskutils/diskutils.go
ddaney-fungible: OK. Lack of knowledge of go is surely biting me here.

toolkit/tools/imagegen/installutils/installutils.go
ddaney-fungible: I don't think that is a requirement.

Disk block size is independent of file system block size. But it shouldn't matter as overlay mounts don't even have to use the same file system let alone the file system block size.

toolkit/tools/imagegen/diskutils/diskutils.go
cwize1: There is a good chance I'll need to add support for specifying the block size to the image customizer at some point. So, even if you don't fix this now, I may end up doing this in the future.

ddaney-fungible: I made the suggested change.

jslobodzian:
PR has been opened for over 2 years against Mariner 2.0.  Mariner 2.0 support has ended.  Closing as won't fix.

PR #6876, opened 2023-11-30
Support setting home directory and primary group when creating users

ellie-di:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

- Ability to specify the user's home directory when creating the user. We needed this to be over-ridden for some of the service users that are predefined in the derivative image that we are working on.
Related PR https://github.com/microsoft/CBL-Mariner/pull/6957.

###### Change Log  
- Changed toolkit to support setting the home directory and group when creating users

**NO**

###### Test Methodology

Manually tested with a derivative image and verified it works as expected.
- Pipeline build id: 482218

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -433,11 +433,20 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 	// Keep a running total of how many packages have been installed through all the `TdnfInstallWithProgress` invocations
 	packagesInstalled := 0
 
+	var (
+		corePkgs = []string{"filesystem"}
+	)
+	if len(config.Users) > 0 || len(config.Groups) > 0 {
+		corePkgs = append(corePkgs, "shadow-utils")
+	}
+
 	timestamp.StartEvent("installing packages", nil)
 	// Install filesystem package first
-	packagesInstalled, err = TdnfInstallWithProgress(filesystemPkg, installRoot, packagesInstalled, totalPackages, true)
-	if err != nil {
-		return
+	for _, pkg := range corePkgs {
+		packagesInstalled, err = TdnfInstallWithProgress(pkg, installRoot, packagesInstalled, totalPackages, true)
+		if err != nil {
+			return
+		}
 	}
 
 	hostname := config.Hostname
@@ -447,6 +456,18 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		if err != nil {
 			return
 		}
+	}
+
+	// Add groups
+	err = addGroups(installChroot, config.Groups)
+	if err != nil {
+		return
+	}
+
+	// Add users
+	err = addUsers(installChroot, config.Users)
+	if err != nil {
+		return
 	}
 
 	// Install packages one-by-one to avoid exhausting memory
@@ -473,18 +494,6 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		if err != nil {
 			return
 		}
-
-		// Add groups
-		err = addGroups(installChroot, config.Groups)
-		if err != nil {
-			return
-		}
-	}
-
-	// Add users
-	err = addUsers(installChroot, config.Users)
-	if err != nil {
-		return
 	}
 
 	// Add machine-id
@@ -1310,7 +1319,7 @@ func createUserWithPassword(installChroot *safechroot.Chroot, user configuration
 		}
 		isRoot = true
 	} else {
-		err = userutils.AddUser(user.Name, hashedPassword, user.UID, installChroot)
+		err = userutils.AddUser(user.Name, user.HomeDirectory, user.PrimaryGroup, hashedPassword, user.UID, installChroot)
 		if err != nil {
 			return
 		}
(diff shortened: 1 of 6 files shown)


toolkit/tools/internal/userutils/userutils.go
```diff
@@ -70,14 +70,21 @@ func UserExists(username string, installChroot *safechroot.Chroot) (bool, error)
...
+func AddUser(username string, homeDir string, primaryGroup string, hashedPassword string, uid string, installChroot *safechroot.Chroot) error {
 	var args = []string{username, "-m"}
 	if hashedPassword != "" {
 		args = append(args, "-p", hashedPassword)
 	}
 	if uid != "" {
 		args = append(args, "-u", uid)
 	}
+	if len(homeDir) == 0 {
+		homeDir = filepath.Join(UserHomeDirPrefix, username)
```
reubeno, line 82: Previously the code was not passing `-d` at all into `useradd`. Is there a risk that the default behavior != expliciting setting the home dir to `filepath.Join(UserHomeDirPrefix, username)`?

reubeno: _And as a minor code reading question, comparing a string against `""` seems to be idiomatic golang, or at least common to this code base; if we do end up needing this conditional, do you think it would be more consistent to change from `len(homeDir) == 0` to `homeDir == ""`?_

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -449,6 +458,18 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		}
 	}
 
+	// Add groups
+	err = addGroups(installChroot, config.Groups)
```
reubeno, line 462: It looks like this PR is making 2 sets of changes at once:

1) Allowing specification of custom home dir + primary group for user accounts created.
2) Shift user/group creation *earlier* in the image build process, such that it happens right after installing `filesystem` (and now `shadow-utils`, which is required for `useradd`).

From a PR perspective, it's probably cleanest to split these into separate commits/PRs. (That also would make it easier to identify problems down the road / revert logic if one half of the changes becomes an issue.)

From a code change perspective, the second change described above seems more impactful than the first. Can we think of what unintended behavioral deltas could result? For example:

* With the explicit installation of `shadow-utils`, is there a risk that someone who had pinned `shadow-utils` to a specific package version in their `.json` package list file now may end up with odd behavior during image build? 
* I see the previous logic didn't always invoke `addGroups()`. I'm not sure I understand why at first glance, but with this change, we'd now always call it -- is that okay?

toolkit/tools/internal/userutils/userutils.go
ellie-di: right, I don't think home dir needs to be initialized with a default. I've updated this part to only set the home dir if it's set.

toolkit/tools/imagegen/installutils/installutils.go
ellie-di: Makes sense, thanks for the suggestion!
I've separated those and posted the PR for the second part here: https://github.com/microsoft/CBL-Mariner/pull/6957

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1310,7 +1310,7 @@ func createUserWithPassword(installChroot *safechroot.Chroot, user configuration
 		}
 		isRoot = true
 	} else {
-		err = userutils.AddUser(user.Name, hashedPassword, user.UID, installChroot)
+		err = userutils.AddUser(user.Name, user.HomeDirectory, user.PrimaryGroup, hashedPassword, user.UID, installChroot)
```
reubeno, line 1313: Thanks for splitting up this change from the other one!

I looked back and saw that `PrimaryGroup` was already being consumed on or around line 1240 of this file, with the call to `ConfigureUserGroupMembership()`. Is there a potential conflict with this logic? Or alternatively, is that preexisting logic sufficient or does it end up creating a primary group for the user that gets "leaked" if the user's `PrimaryGroup` isn't the same as `Name`?

ellie-di: I investigated this a bit, and my conclusion is that no there is no conflict, and the preexisting logic is not sufficient because `ConfigureUserGroupMembership` is getting called after `createUserWithPassword` (that calls `AddUser`). Given that groups get created before users, the `AddUser` fails if we want to add a user to an already existing group, this forces us to us `-g` option with useradd tool.

toolkit/tools/internal/userutils/userutils.go
```diff
@@ -70,14 +70,20 @@ func UserExists(username string, installChroot safechroot.ChrootInterface) (bool
...
 	if hashedPassword != "" {
 		args = append(args, "-p", hashedPassword)
 	}
 	if uid != "" {
 		args = append(args, "-u", uid)
 	}
+	if homeDir != "" {
+		args = append(args, "-d", homeDir)
+	}
+	if primaryGroup != "" {
```
dmcilvaney, line 84: I think for non-rootfs we won't have created this group yet right? I'm thinking maybe `addGroups()` should take both the `config.Groups` and each users' custom group if set. It would need to be run for rootfs images as well though.

ellie-di: No, considering [this PR](https://github.com/microsoft/CBL-Mariner/pull/6957) that is related to the changes in current PR, the groups will be created before adding the users for both rootfs and non-rootfs.

dmcilvaney: Sorry, wasn't clear. I'm mostly worried about the case where a user might have a different group than the ones defined in `config.Groups`. Is this a possibility? My reading of useradd's docs suggest that `-g` only works on already existing groups. We can handle this two ways, either update `addGroups()`, or add a validation step to the config that enforces each user can only belong to groups that are defined.

PR #6881, opened 2023-11-30
Image Customizer: Add support for kernel command-line

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add support for specifying additional kernel command line args during image customization.

###### Change Log  

- Image Customizer: Add support for kernel command-line

NO

###### Test Methodology

- Added UTs.


toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -179,6 +179,22 @@ SystemConfig:
...
 
+## KernelCommandLine type
+
+Options for configuring the kernel.
+
+### ExtraCommandLine
+
+Additional Linux kernel command line options to add during partition customization.
+
+This value may only be used when when the [Disks](#disks-disk) and
```
jiria, line 190: Would the same limitation apply to EMU? We thought we can leverage the same functionality from EMU for Trident.

cwize1: Yes, this field won't be usable with the OS-modifier. But `CommandLineAdd` below should be usable.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -0,0 +1,57 @@
...
+
+func handleKernelCommandLine(commandLineAdd string, imageChroot *safechroot.Chroot) error {
+	var err error
+
+	if commandLineAdd == "" {
+		// Nothing to do.
+		return nil
+	}
+
+	grub2ConfigFilePath := filepath.Join(imageChroot.RootDir(), "/boot/grub2/grub.cfg")
```
dmcilvaney, line 27: @camelron has been working on using grub2 mkconfig to dynamically generate `grub.cfg`. If a user is using this new flow `/boot/grab2/grub.cfg` will get overwritten. We would need to update both the template input and this file (or just trigger mkconfig again to update).

This is mostly for M3, but we should probably handle it here so we don't have any bad surprises in the future. I'll get him to comment on what we should check for.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -0,0 +1,57 @@
...
+	// Read the existing grub.cfg file.
+	grub2ConfigFileBytes, err := os.ReadFile(grub2ConfigFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to read existing grub2 config file: %w", err)
+	}
+
+	grub2ConfigFile := string(grub2ConfigFileBytes)
+
+	// Find the point where the new command line arguments should be added.
+	match := linuxCommandLineRegex.FindStringSubmatchIndex(grub2ConfigFile)
```
dmcilvaney, line 38: More brainstorming, is there value in moving to a dedicated variable for `root=` etc. so we don't stomp the disks? Then we can just replace/add after that well-known variable.

Not useful for M2, but maybe M3 feature?

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
dmcilvaney: These are the two files of interest: https://github.com/microsoft/CBL-Mariner/blob/eebc0beedc34649f3f10bb9e1763afbef531080c/toolkit/tools/imagegen/installutils/installutils.go#L1088

dmcilvaney: per offline, change will come in future PR.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer_test.go
```diff
@@ -195,6 +208,193 @@ func TestValidateConfigScriptNonExecutable(t *testing.T) {
...
+	grub2ConfigFilePath := filepath.Join(imageConnection.Chroot().RootDir(), "/boot/grub2/grub.cfg")
+
+	grub2ConfigFile, err := os.ReadFile(grub2ConfigFilePath)
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	t.Logf("%s", grub2ConfigFile)
+
+	linuxCommandLineRegex, err := regexp.Compile(`linux .* console=tty0 console=ttyS0 `)
```
dmcilvaney, line 390: Thoughts on expanding this so we can double check it doesn't mangle the rest of the command line we want to keep? Bit hacky but if the image config uses PARTLABEL I think the whole line should be deterministic.

dmcilvaney (approved):

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
Camelron: This isn't grub-enlightened; changes to /boot/grub2/grub.cfg get stomped next time grub2-mkconfig is called. 

You could drop all imagecustomizer settings in an /etc/default/grub.d/10_customizer_cfg.cfg 

See example: https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS/kernel-mshv/50_mariner_mshv.cfg#L4C2-L4C2 

This will cause any additional cmdline params to stream into the generated /boot/grub2/grub.cfg every time.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
Camelron: At least in the grub2-mkconfig case, root is supplied separately. 

Example in kernel-mshv, which uses the same logic as the standard 10_linux template: https://github.com/microsoft/CBL-Mariner/pull/6343/files#diff-6d30214896e767042064c1c3dafc8beb64ce2e096b1f5245fce15a57a687f0e0R64

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -0,0 +1,57 @@
...
+	if match == nil {
+		return fmt.Errorf("failed to find Linux kernel command line params in grub2 config file")
+	}
+
+	// Get the location of "$kernelopts".
```
Camelron, line 43: Splicing directly into /boot/grub2/grub.cfg will not be persistent. We need to find a way to edit /etc/default/grub, /etc/default/grub.d/*.cfg in order for mkconfig to be supported.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -179,6 +179,22 @@ SystemConfig:
...
+[SystemConfig.PartitionSettings](#partitionsettings-partitionsetting) fields are
+specified.
+
+### CommandLineAdd
+
+Additional Linux kernel command line options to add during image customization.
```
jiria, line 196: Should there be a capability to override the command line as well/instead of?

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Yeah, I have a separate task for handling `grub-mkconfig` enabled images.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer_test.go
cwize1: I added the two surrounding values.

toolkit/tools/imagecustomizerapi/config.go
```diff
@@ -39,11 +39,16 @@ func (c *Config) IsValid() error {
 
 	hasDisks := c.Disks != nil
 	hasBootType := c.SystemConfig.BootType != BootTypeUnset
+	hasKernelCommandLineArgs := c.SystemConfig.KernelCommandLine.ExtraCommandLine != ""
 
 	if hasDisks != hasBootType {
 		return fmt.Errorf("SystemConfig.BootType and Disks must be specified together")
 	}
 
+	if hasKernelCommandLineArgs && !hasDisks {
```
romoh, line 48: why are we adding that limitation?

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -179,6 +179,22 @@ SystemConfig:
       Permissions: "664"
 ```
 
+## KernelCommandLine type
+
+Options for configuring the kernel.
+
+### ExtraCommandLine
```
romoh, line 186: should we rename that to something more obvious. Also It seems to me that this should move under the partition API

cwize1: I am trying to remain consistent with the imager's config.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -179,6 +179,26 @@ SystemConfig:
...
+If the partitions are customized, then the `grub.cfg` file will be reset to handle the
+new partition layout.
+When the new `grub.cfg` file is created, the `ExtraCommandLine` value will be added to
+the Linux kernel command line options.
+
+This field may only be used when when the [Disks](#disks-disk) and
+[SystemConfig.PartitionSettings](#partitionsettings-partitionsetting) fields are
+specified.
+
+### ExtraCommandLineAdd
```
jiria, line 197: PR description is out of date.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -179,6 +179,26 @@ SystemConfig:
...
+
+Additional Linux kernel command line options that are added to the existing `grub.cfg`
```
jiria, line 199: I dont see a reason, from user perspective, to have these two varied attributes. Why should user care and differentiate between the ExtraCommandLine and ExtraCommandLineAdd?

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -179,6 +179,26 @@ SystemConfig:
       Permissions: "664"
 ```
 
+## KernelCommandLine type
+
+Options for configuring the kernel.
```
jiria, line 184: Are there no plans to set/override the kernel command line?

cwize1: Nope. Too difficult.

The complete set of args is split across multiple files:

- /boot/grub2/grub.cfg
- /boot/systemd.cfg
- /boot/mariner.cfg
- /boot/mariner-mshv.cfg
- /boot//grub2/grubenv

Trying to come up with an API to add/remove args from any of these files seems intractable. And worse, it is all subject to change (particularly in relation to introducing grub-mkconfig in Mariner 3.0).

So, I am just going to let the user use post-install scripts if they want to do more complex modifications.

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: Fair enough. I merged them.

romoh (approved):

PR #6886, opened 2023-12-01
Fix instances of deadname and old alias in changelogs

oliviacrain:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When I changed my name in 2022, I elected to not change instances of my [deadname](https://en.wikipedia.org/wiki/Deadnaming) for reasons I won't get into here.

Now it's 2023 and my feelings have changed- it's jarring to see that name, and I'd like it gone from the changelogs. I think we're at a point in the 3.0-dev lifecycle where this change would be least disruptive.

It doesn't get rid of the name from git metadata, but [the only way to do that](https://blog.amy-k.net/git-mail-map/) would be to add another persistent file with my deadname in it ([`.mailmap`](https://git-scm.com/docs/gitmailmap)) or rewrite git history. I don't think either of those are good options, so we won't do that.

###### Change Log  

- Fix deadname occurrences (and typos of those occurrences) in changelogs

**NO** (not in a way that affects any packages/manifests)

###### Test Methodology

- Went through every single line changed to verify I only changed my own information.


oliviacrain:
Spec files check is failing on unverified licenses in a few specs- this is not an appropriate PR to do that in, so I'm going to not address that.

jslobodzian:
Will not be taking mass changes to the Changelog history.  Closing this PR.

PR #6889, opened 2023-12-01
timestamp: overwrite timestamp logs on different builds

trungams:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

- When we re-run build commands that has timestamp feature on, the timestamp log file for that specific program should be overwritten. This is not an issue most of the time, but it could error out if we try to read the timestamp file content and some unexpected data are present. Adding os.O_TRUNC flag when opening log file for the first time to make sure existing content is erased before we proceed with the program.

###### Change Log  

- Add os.O_TRUNC flag when beginning timestamp for a program.

**NO**

###### Test Methodology

- Local build

### DIFF
--- a/toolkit/tools/internal/timestamp/timestamp_mgr.go
+++ b/toolkit/tools/internal/timestamp/timestamp_mgr.go
@@ -144,7 +144,7 @@ func BeginTiming(toolName, outputFile string) (*TimeStamp, error) {
 
 	initTimeStampManager()
 
-	outputFileDescriptor, err := os.OpenFile(outputFile, os.O_CREATE|os.O_WRONLY, 0664)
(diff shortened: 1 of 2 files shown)


trungams:
> Was this one of the errors that was showing up as a build break? 

Yeah I believe this is the one. It only happens when we're using bldtracker because that one reads data from an existing file. 

> Ideally I'd like to see if we can completely isolate the rest of the tools from esoteric failures in the timing/profiling code. Best if it just falls over and we can keep going.

Agree. Actually most of the errors with timestamp code for example are ignored, this one just breaks the build because it's invalid data being accessed and nothing is done to prevent it, so the program just crashes. We'll probably have to figure out a mechanism to support these things running in parallel with the build tool.

Issue #6788, opened 2023-11-17
Hypothetical new toolchain build flow


dmcilvaney:
I'm investigating a new wrapper around the toolchain bootstrap scripts, I'd like some input on the various modes the toolchain can operate in.

The "build" mode can be one of `Auto`, `Fast`, `Force`, `Never`. Generally auto will try and get a toolchain as fast as possible and ideally should be the default going forward.

The archive is a path to a `toolchain.tar.gz`, and will take precedence over locally built/downloaded packages.

A new feature will allow automatic updating of the manifest files. This is future looking and likely won't be in V1.

There is also a global cache flag. It doesn't affect the overall behavior of the tools, just short-circuits certain build steps using pre-built parts available locally in the dev environment.

** Does the table make sense? Any suggestions for changing the behavior? ** 

| Mode:Auto 	| Mode:Fast 	| Mode:Force 	| Mode:Never 	| Archive 	| Update Manifest 	|   	| Behavior                                                                                                                         	|
|-----------	|-----------	|------------	|------------	|---------	|-----------------	|---	|----------------------------------------------------------------------------------------------------------------------------------	|
| ✅         	| ⛔         	| ⛔          	| ⛔          	| ✅       	| ✅               	|   	| Manifest: Update from Archive<br>Download: NO<br>Rebuild: NO                                                                     	|
| ⛔         	| ✅         	| ⛔          	| ⛔          	| ✅       	| ✅               	|   	| INVALID (can't build & use archive)                                                                                              	|
| ⛔         	| ⛔         	| ✅          	| ⛔          	| ✅       	| ✅               	|   	| INVALID (can't build & use archive)                                                                                              	|
| ⛔         	| ⛔         	| ⛔          	| ✅          	| ✅       	| ✅               	|   	| Manifest: Update from Archive<br>Download: NO<br>Rebuild: NO                                                                     	|
| ✅         	| ⛔         	| ⛔          	| ⛔          	| ✅       	| ⛔               	|   	| Manifest: Toolkit<br>Download: NO<br>Rebuild: NO                                                                                 	|
| ⛔         	| ✅         	| ⛔          	| ⛔          	| ✅       	| ⛔               	|   	| INVALID (can't build & use archive)                                                                                              	|
| ⛔         	| ⛔         	| ✅          	| ⛔          	| ✅       	| ⛔               	|   	| INVALID (can't build & use archive)                                                                                              	|
[...]

PR #6892, opened 2023-12-02
RFC: New go tool to wrap the existing toolchain

dmcilvaney:
This is a large set of changes, but the end goal is to support https://github.com/microsoft/CBL-Mariner/issues/6788.
Each individual change will likely be its own PR.

Ideally this can be back-ported as an optional feature to 2.0 as well, but at least we should aim for 3.0.


dmcilvaney:
Note: currently targeted at main for dev purposes, will likely point to 3.0

dmcilvaney:
Individual PRs:
https://github.com/microsoft/CBL-Mariner/pull/6956
...

PR #6705, opened 2023-11-08
kernel-mshv: Expand crashkernel reservation based on system RAM

Camelron:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [ ] Documentation has been updated to match any changes to the build system
- [ ] Ready to merge

---

###### Summary 

What does the PR accomplish, why was it needed?
Note: AB vhd build has been sufficiently cleaned up, to the point where the cloud-init provisioning bug is no longer present.
This change sets the crashkernel param for kernel-mshv to 256M for systems with less than 32G of ram, otherwise 512M. 

###### Change Log  

- Change crashkernel param to be dynamic based on amount of system RAM. 

###### Does this affect the toolchain?  

NO

###### Test Methodology

- pipeline run **https://dev.azure.com/mariner-org/mariner/_build/results?buildId=456533&view=results**

PR #6896, opened 2023-12-04
Mask kdump.service; remove crashkernel when kernel-mshv is installed

Camelron:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

The smallest SKU we currently support for kata scenarios has 16G of ram. This can go all the way up to 384G of ram in the more expensive skus. Since we don't currently support vmcore compression nor omission of zero/unallocated pages, this can very quickly fill out our rootfs. 

Given this info, I think it is rather risky to keep kdump enabled by default. The larger skus may naively fill up their rootfs after a single crash. We really need to get that compression behavior configuration working, and, per /etc/kdump.conf man page:

```kdump.conf provides post-kexec instructions to the kdump kernel. It is stored in the initrd file managed by the kdump service. If you change this file and do not want to restart before it takes effect, restart the kdump service to rebuild to initrd.```

Basically we need to make kdump-generated initrd work in order to configure compression, see earlier comments in this thread as well as https://github.com/microsoft/CBL-Mariner/pull/6479.

###### Change Log  

- kernel-mshv masks kdump.service in posttrans script
- remove crashkernel= param from kernel-mshv legacy boot
- remove crashkernel= param from kernel-mshv mkconfig boot

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=463004&view=results

### DIFF
--- a/SPECS/kernel-mshv/kernel-mshv.spec
+++ b/SPECS/kernel-mshv/kernel-mshv.spec
@@ -11,7 +11,7 @@
 Summary:        Mariner kernel that has MSHV Host support
 Name:           kernel-mshv
 Version:        5.15.126.mshv9
-Release:        2%{?dist}
+Release:        3%{?dist}
 License:        GPLv2
 Group:          Development/Tools
 Vendor:         Microsoft Corporation
@@ -139,7 +139,7 @@ ln -s vmlinux-%{uname_r} %{buildroot}%{_libdir}/debug/lib/modules/%{uname_r}/vml
 
 cat > %{buildroot}/boot/linux-%{uname_r}.cfg << "EOF"
 # GRUB Environment Block
-mariner_cmdline_mshv=rd.auto=1 lockdown=integrity sysctl.kernel.unprivileged_bpf_disabled=1 init=/lib/systemd/systemd ro no-vmw-sta crashkernel=128M audit=0 console=ttyS0,115200n8 earlyprintk
+mariner_cmdline_mshv=rd.auto=1 lockdown=integrity sysctl.kernel.unprivileged_bpf_disabled=1 init=/lib/systemd/systemd ro no-vmw-sta audit=0 console=ttyS0,115200n8 earlyprintk
 mariner_linux_mshv=vmlinuz-%{uname_r}
 mariner_initrd_mshv=initrd.img-%{uname_r}
 EOF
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by christopherco — CHANGES_REQUESTED** [review]
Given your description, just mask the service.

Don't change the reservation properties of the VM, especially if the plan is to re-introduce the memory reservation later.

File: SPECS/kernel-mshv/kernel-mshv.spec [hunk 1]
```diff
@@ -208,6 +208,9 @@ fi
 ln -sf linux-%{uname_r}.cfg /boot/mariner-mshv.cfg
 %grub2_post
 
+%posttrans
+systemctl mask kdump.service >/dev/null 2>&1 || :
```
christopherco, line 212: Should this really be masked at the kernel-mshv RPM transaction level? Maybe this masking is better done at the image composition level?

Camelron:
> Given your description, just mask the service.
> 
> Don't change the reservation properties of the VM, especially if the plan is to re-introduce the memory reservation later.

Is it worth carrying the crashkernel reservation on all systems even if we won't benefit from it? 

This is an argument along the lines of maintaining consistent behavior across 2.0 releases. In that case, I think it makes sense to merge this adjustment instead (ASAP) https://github.com/microsoft/CBL-Mariner/pull/6705 so that we are always testing/iterating on images with the same crashkernel= param.

SPECS/kernel-mshv/kernel-mshv.spec
Camelron: Given your above comment about not changing crashkernel= reservation and this, I think I will close this PR and instead do the masking in AgentBaker.

christopherco:
> > Given your description, just mask the service.
> > Don't change the reservation properties of the VM, especially if the plan is to re-introduce the memory reservation later.
> 
> Is it worth carrying the crashkernel reservation on all systems even if we won't benefit from it?
> 
> This is an argument along the lines of maintaining consistent behavior across 2.0 releases. In that case, I think it makes sense to merge this adjustment instead (ASAP) #6705 so that we are always testing/iterating on images with the same crashkernel= param.

Rationale is - if you're planning to re-introduce `crashkernel=` memory reservation later on for 2.0 `kernel-mshv`, then keeping the reservation makes sense because consumers will stay used to operating with the memory constraints. Removing it now is probably fine from the consumer perspective (though also a breaking change technically), but reintroducing memory limitations on customer nodes later on is a clear breaking change.

If you are NOT going to introduce `crashkernel=` memory reservation for the lifetime of 2.0 `kernel-mshv`, then it makes sense to remove `crashkernel=` altogether and give that memory back to the consumer.

I am in agreement with merging the memory reservation adjustment in #6705 if this is representative of the final reservation we're expecting at this time

PR #6954, opened 2023-12-08
Made only non-test package builds produce SRPMs.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

With running regular and test package builds separately we ended up copying the SRPM for the same package twice. There's no need to do that and in scenarios where we only run tests and skip building the RPMs, copying the SRPM to the output would give a false indication, that the RPMs were actually built. With that in mind, I update the logic to skip copying of SRPMs for test package builds.

###### Change Log  

- Updated the logic inside `pkgworker.go` to copy SRPMs only during regular (non-test) package builds.

No.

###### Test Methodology

- PR check.
- Local builds.

### DIFF
--- a/toolkit/tools/pkgworker/pkgworker.go
+++ b/toolkit/tools/pkgworker/pkgworker.go
@@ -14,7 +14,7 @@ import (
 	"strings"
 	"time"
 
-	ccachemanagerpkg "github.com/microsoft/CBL-Mariner/toolkit/tools/internal/ccachemanager"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/ccachemanager"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/exe"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/file"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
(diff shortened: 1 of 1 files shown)


toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -14,7 +14,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/ccachemanager"
+	ccachemanagerpkg "github.com/microsoft/CBL-Mariner/toolkit/tools/internal/ccachemanager"
```
anphel31, line 17: is this change intentional?

anphel31 (approved):

toolkit/tools/pkgworker/pkgworker.go
dmcilvaney: I've seen the same, `go fmt` prefers this. Was going to do the same next time I changed the file.

dmcilvaney (approved):

toolkit/tools/pkgworker/pkgworker.go
PawelWMS: Yes and no. The Go plug-in for VS Code adds these for cases where the "basename" of the package from the string doesn't match what's actually the package name where you go to the source. I can revert this line. I'm thinking about updating the package contents themselves to follow what we've been doing with other Go packages but that's a change for `main`.

PawelWMS: Created a separate PR here: #6958.

PR #6957, opened 2023-12-08
Shift user/group creation earlier in image build for rootfs image types

ellie-di:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Mariner toolkit didn't have a capability to set a predefined list of users and groups when building a rootFS image type.
Given some packages also define some users or groups, there is a chance that the user/group is already defined if we do the use/group add after installing the packages in chroot.
This PR is related to https://github.com/microsoft/CBL-Mariner/pull/6876 change.
 
###### Change Log  
- Change toolkit to shift user/group creation earlier in image build for rootfs image types

**NO**

###### Test Methodology

Manually tested with a derivative image and verified it works as expected.
- Pipeline build id: 482218

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -387,7 +387,7 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 	defer timestamp.StopEvent(nil)
 
 	const (
-		filesystemPkg = "filesystem"
+		filesystemPkg  = "filesystem"
 		shadowUtilsPkg = "shadow-utils"
 	)
 
@@ -441,7 +441,7 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		return
 	}
 
-	if isRootFS && (len(config.Users) > 0 || len(config.Groups) > 0)  {
+	if isRootFS && (len(config.Users) > 0 || len(config.Groups) > 0) {
 		// Install shadow-utils package
 		packagesInstalled, err = TdnfInstallWithProgress(shadowUtilsPkg, installRoot, packagesInstalled, totalPackages, true)
 		if err != nil {
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -388,6 +388,7 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 
 	const (
 		filesystemPkg = "filesystem"
+		shadowUtilsPkg = "shadow-utils"
 	)
 
 	defer stopGPGAgent(installChroot)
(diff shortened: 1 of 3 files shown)


toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -440,6 +441,26 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		return
 	}
 
+	if isRootFS && (len(config.Users) > 0 || len(config.Groups) > 0) {
+		// Install shadow-utils package
```
dmcilvaney, line 445: I'd prefer to see this check moved to the validator stage. I'm a bit leery of arbitrarily installing a package that isn't listed in the config. (see `imageconfigvalidator.go:validatePackages()` on how to enforce this). I see the need to prioritize installing the package early though, before the other packages have a chance to pollute the users.

Do we require shadow-utils for all user operations? I feel like there are cases were we don't want to install shadow-utils, but we still need to add users.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -440,6 +441,26 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		return
 	}
 
+	if isRootFS && (len(config.Users) > 0 || len(config.Groups) > 0) {
```
dmcilvaney, line 444: Could we refactor this to only call addUser etc. once, it feels very odd to have two flows here (for rootfs and non-rootfs). I'd assume that non-rootfs images can show the same type of issues with pre-defining users.

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: nvm, `useradd` comes form `shadow-utils`. Still recommend adding this to the validator rather than arbitrarily installing it.

ellie-di: Done. Thanks for your review!

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -449,6 +449,18 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		}
 	}
 
+	// Add groups
```
dmcilvaney, line 452: I think we will still need to add an install of shadow-utils early here, maybe include it with filesystem on line 438.
Something like
```go
if len(config.Users) > 0 || len(config.Groups) > 0 {
    	shadowInstalled, err := TdnfInstallWithProgress(filesystemPkg, installRoot, packagesInstalled, totalPackages, true)
	if err != nil {
		return
	}
        packagesInstalled += shadowInstalled
}

dmcilvaney (commented):
Oops, forgot to hit submit on this...

dmcilvaney (approved):

ellie-di:
> Oops, forgot to hit submit on this...

Thanks so much @dmcilvaney for your feedback and review!

PR #6963, opened 2023-12-09
NewToolchain 003: Define a user to own directories we create in make

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Part of https://github.com/microsoft/CBL-Mariner/pull/6892, related to https://github.com/microsoft/CBL-Mariner/issues/6788

This is part of a series of PRs for the new toolchain prototype. Some PRs are fundamental bug fixes for 2.0, some are new features better targeted at 3.0, and some are early prototypes/RFCs.

##### Destination
2.0 - Maybe
3.0 - Yes
RFC - 

##### Specific PR summary:
The calling user should probably own at least some of the files the tools generate. Lets start by defining the ideal owner, and making the directories we create be owned by that user.

###### Change Log  

- Environment variable `MARINER_BUILDER_USER` will be set to either `$SUDO_USER` if defined, or the current `$USER` otherwise.
- Create `makedirs.sh` to replace the existing `create_folder` define in Make. The script will be called instead and may now report an error back if it fails.

**YES/NO**

###### Associated issues  

- https://github.com/microsoft/CBL-Mariner/issues/6788

###### Test Methodology

- Local builds
- Full tests TBD

### DIFF
--- a/toolkit/scripts/makedirs.sh
+++ b/toolkit/scripts/makedirs.sh
@@ -11,28 +11,6 @@
 DIR="${1}"
 USER="${2}"
 
-function make_dir_recursive {
-    local dir="${1}"
-    local user="${2}"
-    local parent_dir
-    parent_dir="$(dirname "${dir}")"
-
-    # If the parent directory also needs to be created, recurse
-    if [[ ! -d "${parent_dir}" ]]; then
-        make_dir_recursive "${parent_dir}" "${user}" || { echo "Failed to recursively create '${parent_dir}'" ; return 1 ; }
-    fi
-
-    if [[ -d "${dir}" ]]; then
-        return 0
-    fi
-
-    mkdir -p "${dir}" || { echo "Failed to create '${dir}'" ; return 1 ; }
-    chown "${user}" "${dir}" || { echo "Failed to chown '${dir}' to user '${user}'" ; return 1 ; }
-    touch -d @0 "${dir}" || { echo "Failed to reset timestamp on '${dir}'" ; return 1 ; }
-
-    return 0
-}
-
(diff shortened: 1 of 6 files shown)


toolkit/scripts/makedirs.sh
```diff
@@ -0,0 +1,46 @@
...
+# Usage: /path/to/dirs/to/make owner_user_name
+
+# Will recursively create all directories in the path, and chown them to the specified user
+# If the directory already exists, it will be skipped and no changes will be made.
+# New directories will be created with timestamp 0.
+
+DIR="${1}"
+USER="${2}"
+
+function make_dir_recursive {
```
cwize1, line 14: Could this be replaced with `runuser -u "$SUDO_USER" -- mkdir -p "<dir>"`?

dmcilvaney: I'll give it a shot, would certainly simplify the code a bit.

dmcilvaney: We lose the touch -d @0, but I'm not sure that was ever a critical featuer of this. Generally if a directory is missing then we probably need to trigger a rebuild of stuff anyways. And once a directory is created we should never need to remove it unless we run a clean.

dmcilvaney: Seems to work.

We lose the touch -d @0, but I'm not sure that was ever a critical featuer of this. Generally if a directory is missing then we probably need to trigger a rebuild of stuff anyways. And once a directory is created we should never need to remove it unless we run a clean.

dmcilvaney:
Had to add a bit more handling for non-sudo calls. runuser can only be used as sudo, even if you are just asking to run as yourself.

PR #6358, opened 2023-10-06
Preserve yum backend on tdnf package upgrade

AZaugg:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [X] The toolchain has been rebuilt successfully (or no changes were made to it)
- [X] The toolchain/worker package manifests are up-to-date
- [X] Any updated packages successfully build (or no packages were changed)
- [X] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [X] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [X] All package sources are available
- [X] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [X] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [X] All source files have up-to-date hashes in the `*.signatures.json` files
- [X] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [X] Documentation has been updated to match any changes to the build system
- [X] Ready to merge

---

###### Summary 
If a user updates the yum backend, a tdnf package update should not reset the backend. Move /bin/yum ownership from package and move it into the post install and post uninstall scripts

###### Change Log  

- Preserve yum backend on tdnf package upgrade

###### Does this affect the toolchain?  

**YES**

###### Test Methodology

- Build and updated package

1. Built package
```
azaugg@azaugg-ld77 [ ~/CBL-Mariner/toolkit ]$ sudo make build-packages -j$(nproc) REBUILD_TOOLS=y SRPM_PACK_LIST="tdnf" PACKAGE_REBUILD_LIST="tdnf" REFRESH_WORKER_CHROOT=y RUN_CHECK=y
[sudo] password for azaugg:
/home/azaugg/CBL-Mariner/toolkit/out/tools/specreader \
	--dir /home/azaugg/CBL-Mariner/SPECS \
	--spec-list=/home/azaugg/CBL-Mariner/build/INTERMEDIATE_SRPMS/pack_list.txt \
[...]

PR #6968, opened 2023-12-10
fix: preserve yum backend on tdnf package upgrade

### LINKED ISSUE #6609: User updates to yum symlink are not preserved after tdnf update
The yum symlink is tracked in the tdnf package. So when tdnf package is upgraded, the yum symlink is replaced. This means if a user reconfigured the yum symlink to a different application (i.e., dnf), then a tdnf package update will overwrite the user's symlink

christopherco:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Today, if a user updates the yum backend by editing the /usr/bin/yum symlink, a tdnf package update will reset the symlink back to pointing to /usr/bin/tdnf, thus resetting the user's desired configuration. This is because tdnf RPM packaging indicates /usr/bin/yum is owned by tdnf and so package update operation will always stomp /usr/bin/yum with the symlink pointing to /usr/bin/tdnf.

To address this, remove /usr/bin/yum ownership from tdnf package. To maintain backwards compatibility with users who rely on this behavior, generate the default yum -> tdnf symlink in the final stage of the transaction (%posttrans scriptlet). Also for clean package uninstallation, update the %preun scriptlet to remove the default yum -> tdnf symlink if this specific link is present.

Note: upgrade operation from a prior tdnf package version to this one (or beyond) will overwrite any custom yum symlink to point to tdnf, which is the same behavior as before. However future upgrade/reinstall operations will preserve the user's custom yum symlink

###### Change Log  

- Preserve yum backend on tdnf package upgrade
- Remove tdnf package ownership of the yum symlink
- Generate yum symlink to tdnf if no symlink exists in the %posttrans scriptlet

On package upgrade transaction, %post always runs before %preun, so
if a symlink is created in %post, it gets deleted on %preun, and the
result after the RPM transaction is no symlink.

Additionally, when upgrading from a tdnf package version before this,
the first %post attempt will not generate a new symlink since one is
already present, then the old tdnf package will be uninstalled which
also removes the symlink.

To address both scenarios, move the action to generate the yum
symlink to tdnf if no symlink exists from %post to the %posttrans
scriptlet which is the final step run in the entire RPM transaction

The %posttrans step is not run on dnf remove so uninstalling the package
will remove the yum symlink if it is pointing to tdnf

https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#ordering

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/46895684
- Fixes #6609
- Rebase and continuation of #6358

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=467307&view=results

Scenarios:
- Package transactions with default yum symlink to tdnf: upgrade, downgrade, reinstall, removal
- Package transactions with custom yum symlink to dnf: upgrade, downgrade, reinstall, removal
   - Note: upgrade operation to this tdnf will overwrite the custom yum symlink to point to tdnf (which is the same behavior as before), but all future upgrade/reinstall operations will preserve the user's custom symlink


SPECS/tdnf/tdnf.spec
```diff
@@ -167,6 +166,19 @@ if [[ ! -f %{_tdnf_history_db_dir}/history.db ]]; then
     %{_libdir}/tdnf/tdnf-history-util init
 fi
 
+# Make the decision to update yum sym link during post script
+# not package build. Some workloads want to disable tdnf, so
+# decision should not be made during build but during install
+# of package.
+if [ ! -h $(rpm --eval %{_bindir})/yum ]; then
```
christopherco, line 173: There are 2 problems on the upgrade operation from previous behavior to this new behavior.
1. Since the older tdnf with the `yum` symlink is still around at time of installation of new `tdnf` and this `%post` script runs, the conditional fails and the body of this script never runs so the `yum` symlink is never created. This may be acceptable behavior.
2. Then later in the transaction, the old `tdnf` package gets deleted, which takes away the `tdnf` symlink. Therefore at the end of the transaction, there is no symlink at all

Thinking about moving this logic to `%posttrans` which runs at the very end of the transaction to ensure we always have a symlink.

mfrw (approved):

SPECS/tdnf/tdnf.spec
```diff
@@ -167,9 +166,23 @@ if [[ ! -f %{_tdnf_history_db_dir}/history.db ]]; then
     %{_libdir}/tdnf/tdnf-history-util init
 fi
 
+%preun
+if  [[ $(readlink $(rpm --eval %{_bindir})/yum) == $(rpm --eval %{_bindir})/tdnf ]]; then
+  rm $(rpm --eval %{_bindir})/yum
+fi
+
```
ddstreetmicrosoft, line 173: This is *only* needed for the uninstall (not upgrade) action, right? In that case I think this also needs to check `$1 == 0` to make sure we're executing uninstall instead of upgrade.

Also if the symlink is changed to a relative symlink, then the right half of the symlink target check would just be `== tdnf`.

SPECS/tdnf/tdnf.spec
```diff
@@ -167,9 +166,23 @@ if [[ ! -f %{_tdnf_history_db_dir}/history.db ]]; then
...
 %postun
 /sbin/ldconfig
 
+%posttrans
+# Make the decision to update yum sym link during post script
+# not package build. Some workloads want to disable tdnf, so
+# decision should not be made during build but during install
+# of package.
+if [ ! -h $(rpm --eval %{_bindir})/yum ]; then
```
ddstreetmicrosoft, line 182: i don't think this is the right thing to check; i believe we want to create the symlink iff there is no 'yum' file/link at all, otherwise we don't want to do anything. so probably this should use `-e` not `-h`.

SPECS/tdnf/tdnf.spec
```diff
@@ -167,9 +166,23 @@ if [[ ! -f %{_tdnf_history_db_dir}/history.db ]]; then
...
+ln -sf $(rpm --eval %{_bindir})/tdnf $(rpm --eval %{_bindir})/yum
```
ddstreetmicrosoft, line 183: I guess creating an absolute symlink is ok but in general symlinks to same-directory files should always be relative

SPECS/tdnf/tdnf.spec
christopherco: Yes, correct, this `rm` action is only needed on uninstall. Will add the check for uninstall only.
Also will add the relative symlink change since I agree it will be good to avoid absolute symlinks for files in the same directory

SPECS/tdnf/tdnf.spec
christopherco: Ah yup, we don't want to clobber `yum` if it is a file only. So we can change the test to be for file presence (`-e`).

ddstreetmicrosoft (approved):
lgtm

sameluch (approved):

hbeberman (approved):

Issue #6922, opened 2023-12-06
Remove hidepid=2 default setting from /proc mount option


christopherco:
**Describe the bug**
Currently our images set `/proc` mount option with `hidepid=2` in `/etc/fstab`. The `hidepid` option prevents process IDs from being visible to all users. We introduced this as a security hardening setting for early targeted use-cases. However, as our distro has expanded in scope and usage, this setting has become a common compat issue for applications and system services, which heavily outweigh its proposed "security benefit".

Any user or service who wishes to utilize this setting for further PID protection can set the value as a post image configuration step.

**Expected behavior**
For Azure Linux 3.0, remove `hidepid=2` setting from /etc/fstab and remove the ability for our tools to configure `hidepid=2` by default.

christopherco:
Additional references for moving away from `hidepid=2`:
- https://access.redhat.com/solutions/6704531
- https://security.stackexchange.com/questions/259134/why-is-the-mount-option-hidepid-2-not-used-by-default-is-there-a-danger-in-us

rlmenge:
For further context, here is the original PR which added hidepid: https://github.com/microsoft/CBL-Mariner/pull/797 . This was added in relation to a bug [here](https://microsoft.visualstudio.com/OS/_workitems/edit/31858389). As noted above, there is an ability to turn this off in imageconfigs today by setting `HidepidDisabled` to `false`. However, this is not the default. 

After reading through the above references, I agree that the security benefits of hidepid (outlined in this [linux-audit article](https://linux-audit.com/linux-system-hardening-adding-hidepid-to-proc/) are not worth the broken functionality, and can be potentially avoided "via systemd's D-Bus interface"([RedHat solution 6704531](https://access.redhat.com/solutions/6704531)) although I have not tested this myself.

christopherco:
Merged into 3.0 development branch. Closing

PR #6969, opened 2023-12-10
feat: change default hidepid setting to be opt-in

christopherco:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Change the user-facing option HidepidDisabled to EnableHidepid to allow user to now opt-in to add hidepid settings.

This change inverts the default setting when no option is supplied to configure hidepid=0, which is more generally desired for better software compat for applications and system services. User can now opt-in to configure the tools to set up hidepid=2.
 
This change also simplifies the hidepid code by only accessing the option from the config right before use. This allows us to remove unnecessary function arguments passed through the various call layers.

**NO**

###### Associated issues  

- #6922


toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -176,8 +176,6 @@ func (s *SystemConfig) IsValid() (err error) {
 
 	//Validate Encryption
 
-	//Validate HidepidDisabled
```
dmcilvaney, line 179: More of a high level discussion, do we want to start adding logic to validate migrations? I'm not sure we can do much here, but at a high level I'm thinking it might be interesting to detect if a user is setting `HidepidDisabled` and mark it as deprecated somehow...

Outside the scope of this PR I think, but might be an interesting topic as we start to make larger changes to the config formats.

dmcilvaney (approved):

rlmenge (approved):
I am not regularly working with the go tools but in regards to switching the default of the inital Hidepid PR, these changes look good to me

toolkit/tools/imagegen/configuration/systemconfig.go
christopherco: Interesting point - this can be a tool feature we can discuss further outside the PR. Feature issue created - #6973

PR #6985, opened 2023-12-11
Image Customizer: Refresh initrd when partitions are customized

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When the image's partitions are customized, then initrd file needs to refreshed so that the hardcoded partition UUIDs within the initrd file can be updated. The initrd file is refreshed by reinstalling the `initramfs` package. It is done this way because only the RPM post- install scripts within the `initramfs` package know how to properly refresh the initrd file.

###### Change Log  

- Image Customizer:
  - Refresh initrd when partitions are customized.
  - Check that we have RPM sources (when needed) during config validation.

NO

###### Test Methodology

- Ran toolkit UTs.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
@@ -15,32 +15,17 @@ import (
 )
 
 func addRemoveAndUpdatePackages(buildDir string, baseConfigPath string, config *imagecustomizerapi.SystemConfig,
-	imageChroot *safechroot.Chroot, rpmsSources []string, useBaseImageRpmRepos bool,
+	imageChroot *safechroot.Chroot, rpmsSources []string, useBaseImageRpmRepos bool, partitionsCustomized bool,
 ) error {
-	allPackagesRemove, err := collectPackagesList(baseConfigPath, config.PackageListsRemove, config.PackagesRemove)
-	if err != nil {
-		return err
-	}
+	var err error
 
-	allPackagesInstall, err := collectPackagesList(baseConfigPath, config.PackageListsInstall, config.PackagesInstall)
-	if err != nil {
-		return err
-	}
-
-	allPackagesUpdate, err := collectPackagesList(baseConfigPath, config.PackageListsUpdate, config.PackagesUpdate)
-	if err != nil {
-		return err
-	}
-
-	needRpmsSources := len(allPackagesInstall) > 0 || len(allPackagesUpdate) > 0 || config.UpdateBaseImagePackages
+	// Note: The 'validatePackageLists' function read the PackageLists files and merged them into the inline package lists.
+	needRpmsSources := len(config.PackagesInstall) > 0 || len(config.PackagesUpdate) > 0 ||
+		config.UpdateBaseImagePackages || partitionsCustomized
 
 	// Mount RPM sources.
 	var mounts *rpmSourcesMounts
 	if needRpmsSources {
-		if len(rpmsSources) <= 0 && !useBaseImageRpmRepos {
-			return fmt.Errorf("have packages to install or update but no RPM sources were specified")
-		}
-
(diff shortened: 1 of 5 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizepartitions.go
```diff
@@ -11,11 +11,11 @@ import (
 
 func customizePartitions(buildDir string, baseConfigPath string, config *imagecustomizerapi.Config,
 	buildImageFile string,
-) (string, error) {
-	if config.Disks == nil && config.SystemConfig.BootType == imagecustomizerapi.BootTypeUnset {
+) (bool, string, error) {
+	if !hasPartitionCustomizations(config) {
```
dmcilvaney, line 15: No strong opinions, but curious why not a member function of the `Config` struct? I've gone both ways, mostly just curious.

amritakohli (approved):

toolkit/tools/pkg/imagecustomizerlib/customizepartitions.go
cwize1: I generally like to keep my API types as free of code as possible.

vinceaperri (approved):

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -15,40 +15,34 @@ import (
...
 			return err
 		}
 		defer mounts.close()
 	}
 
-	err = removePackages(allPackagesRemove, imageChroot)
+	if partitionsCustomized {
+		logger.Log.Infof("Updating initrd file")
+
+		err = installOrUpdatePackages("reinstall", []string{"initramfs"}, imageChroot)
```
jiria, line 39: Can we have an option to not use the hostonly dracut flag used here?

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -200,8 +215,49 @@ func validateScript(baseConfigPath string, script *imagecustomizerapi.Script) er
...
+	if err != nil {
+		return err
+	}
+
+	allPackagesUpdate, err := collectPackagesList(baseConfigPath, config.PackageListsUpdate, config.PackagesUpdate)
+	if err != nil {
+		return err
+	}
+
+	hasRpmSources := len(rpmsSources) > 0 || useBaseImageRpmRepos
```
romoh, line 236: is `rpmsSources` inferred or passed from user?

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: Nope.

Anytime a package is installed that changes a relevant file (e.g. adding a file under `/etc/dracut.conf.d/), the initrd file will be regenerated using the default config. Ideally, the user would be able to add the `hostonly` option to the dracut config files. But the `mkinitrd` script force sets the `hostonly` value through the CLI. And the regeneration scripts in the `initramfs` package calls `mkinitrd` with default options.

This is a mess and can't be fixed without making breaking changes to Mariner. (It is probably something we should do in Mariner 3.0.)

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
romoh: Then I'm confused by the if condition `if !hasRpmSources`. 
If you have custom partitions in addition to packages to install and relying on pmc. Then this condition will be `true`, but then the following errors will be hit. Am i missing something?

cwize1: If `--disable-base-image-rpm-repos` isn't set, then `useBaseImageRpmRepos` will be `true` and so `hasRpmSources` will be `true` and so `!hasRpmSources` will be false and so the errors won't trigger.

PR #6945, opened 2023-12-07
Added cross-compilation `binutils` and `kernel-headers`.

PawelWMS:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [X] The toolchain has been rebuilt successfully (or no changes were made to it)
- [X] The toolchain/worker package manifests are up-to-date
- [X] Any updated packages successfully build (or no packages were changed)
- [X] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [X] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [X] All package sources are available
- [X] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [X] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [X] All source files have up-to-date hashes in the `*.signatures.json` files
- [X] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [X] Documentation has been updated to match any changes to the build system
- [X] Ready to merge

---

###### Summary 

This is a fix for a breaking change I introduced in #6815. The original change caused other package to start failing with the `configure: error: C compiler cannot create executables` error and the native `binutils` package contained `cross-<binary_name>.mo` files meant for `cross-binutils-common` instead of the plain `<binary_name>.mo`. The original change was reverted and this is an updated version.

The main changes are:
- Instead of using a single directory with the sources for `binutils` and have multiple `configure` and `make` calls point to that directory, I've added copies of the source for each pair of `configure` and `make` calls. For reasons still unknown to me, the original approach caused the `configure: error: C compiler cannot create executables` error. I'd greatly appreciate an explanation from someone more experienced, what was wrong and how to avoid copying the original sources. For now, however, that I've implemented the said workaround and created a separate bug to address it in the future.
- Updated the build steps to produce the `cross-<binary_name>.mo` only for the `cross-binutils-common` subpackage.

###### Change Log  

- Fixed the build issues with packages depending on `binutils`.
[...]

PR #6996, opened 2023-12-13
Added a cross-compilation subpackage for aarch64 into `gcc`.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

#### WARNING: requires #6945 for full build or the fast-track PR check run.

Continuation of #6945 and the remainder of the changes from #5569. This change adds a cross-compilation version of `gcc`.

###### Change Log  

- Added cross-compilation build bits into `gcc.spec.`
- Small spec clean-up.

Yes.

###### Associated issues  

- #5569
- #6945

###### Test Methodology

- Local package build and test.
- [Full build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=485545&view=results) after merging #6945 inside the `pawelwi/dadelan/cross_full_test` branch.
- [Fast-track style PR check](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=487368&view=results) from the same `pawelwi/dadelan/cross_full_test` branch.
- Verified I was able to build ARM64 glibc binaries by following an online article starting from the "[Install glibc headers and bootstrap libraries](https://blog.jgosmann.de/posts/2021/02/07/a-guide-to-crosscompiling-applications/#install-glibc-headers-and-bootstrap-libraries)" step (earlier steps are handled by installing the new cross-compilation packages) all the way to the "[Build and install glibc for the target system](https://blog.jgosmann.de/posts/2021/02/07/a-guide-to-crosscompiling-applications/#build-and-install-glibc-for-the-target-system)". I **skipped** the "[Build and install libgcc](https://blog.jgosmann.de/posts/2021/02/07/a-guide-to-crosscompiling-applications/#build-and-install-libgcc)" step as libgcc is already provided inside the new gcc subpackage.

    **Notes for the "Install glibc headers and bootstrap libraries" instructions**:
    - Set `PREFIX=/usr` and `TARGET=aarch64-linux-gnu`.
    - Used `--with-headers=$PREFIX/arm64-linux-gnu/include` instead of the path from the article, since our path is a bit different.


dmcilvaney:
Haven't had a chance to look at the binutils pr, is there a set of pro/cons around for having a single .spec versus a dedicated cross version of the spec?

PawelWMS:
> Haven't had a chance to look at the binutils pr, is there a set of pro/cons around for having a single .spec versus a dedicated cross version of the spec?

Having it all in one file makes it definitely more robust, but splitting into separate spec files make it harder to track version upgrades and CVE fixes, as they will have to be duplicated among these separate specs.

My main worry with a single spec is that we're adding the package to the toolchain without a good reason, but I decided to postpone the solution until we enable custom per-spec macros - the idea you mentioned to me in December, which would allow us to build the same spec but with a slightly different configuration.

SPECS/gcc/gcc.spec
```diff
@@ -288,7 +496,18 @@ $tests_ok
 %{_lib64dir}/libgomp.so
 %{_lib64dir}/libgomp.spec
 
+%if %{build_cross}
+%files -n cross-gcc-common
+%license COPYING
+%endif
+
+%do_files aarch64-linux-gnu %{build_aarch64}
```
dmcilvaney, line 504: This doesn't really make sense to me, why do we need to hard-code `aarch4-linux-gnu` here? Could you add a comment either here, or with `do_files` explaining the logic?

SPECS/gcc/gcc.spec
```diff
@@ -230,6 +436,8 @@ $tests_ok
 %exclude %{_lib64dir}/libstdc++*
 %exclude %{_lib64dir}/libsupc++*
 
+%do_exclude aarch64-linux-gnu %{build_aarch64}
```
dmcilvaney, line 439: Same comment as `do_files`, need some explanation.

SPECS/gcc/gcc.spec
PawelWMS: I've added a comment explaining `do_files` and `do_exclude` - good idea. The macros are adding files for each cross-compilation architecture. Right now there's only one, so we could drop the macros but this way it's easier to add any other architectures in the future. I also wanted to keep it similar to Fedora's way of doing thing, since we often rely on their specs to update ours, so that future devs may have it slightly easier.

SPECS/gcc/gcc.spec
PawelWMS: Added a comment. Thank you, Daniel.

SPECS/gcc/gcc.spec
```diff
@@ -1,16 +1,74 @@
...
+%files -n gcc-c++-%1 \
+%{_bindir}/%{1}*-c++ \
+%{_bindir}/%{1}*-g++ \
+%endif
+
 Summary:        Contains the GNU compiler collection
 Name:           gcc
 Version:        11.2.0
-Release:        7%{?dist}
+Release:        8%{?dist}
```
mfrw, line 59: Taking a look :)
[given its a big PR .. would take sometime .. hope it is fine]

jslobodzian (approved):

SPECS/gcc/gcc.spec
```diff
@@ -136,35 +219,139 @@ Requires:       libgomp = %{version}-%{release}
...
+    --enable-default-pie \
     --enable-languages=c,c++,fortran \
-    --disable-multilib \
-    --disable-bootstrap \
     --enable-linker-build-id \
     --enable-plugin \
-    --enable-default-pie \
+    --enable-shared \
+    --enable-threads=posix \
     --with-system-zlib
```
PawelWMS, line 330: No changes to the default configuration - just putting the arguments in alphabetical order whenever possible.:)

SPECS/gcc/gcc.spec
PawelWMS: Sorry, can't really wait any longer and will merge now. If you see something worrying, let me know, and we can address it in a separate PR.

PR #7057, opened 2023-12-19
Upgrade glusterfs from 7.9 to 11.1

bfjelds:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  
Update glusterfs from 7.9 to 11.1
* remove patch that is no longer relevant
* update all spec to align with glusterfs spec.in
* create patch to fix build break
* strip all rhel/fedora
* replace hardcoded paths with variables
* reorder/alphabetize dependencies
* remove ganesha package

**NO**

###### Test Methodology
- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=493233&view=results


mfrw (commented):
Seems to need a rebase

bfjelds:
> Seems to need a rebase

rebased :)

SPECS/glusterfs/glusterfs.spec
```diff
@@ -811,196 +1129,245 @@ exit 0
...
+%endif
 
 ##-----------------------------------------------------------------------------
 ## All %%postun should be placed here and keep them sorted
 ##
 %postun
-%{?ldconfig}
 %if ( 0%{!?_without_syslog:1} )
-
+%if ( 0%{?fedora} ) || ( 0%{?rhel} && 0%{?rhel} >= 6 )
```
neha170, line 1163: Please remove all references to fedora, rhel, etc. thanks

SPECS/glusterfs/glusterfs.spec
```diff
@@ -753,16 +1071,16 @@ fi
 # add marker translator
 # but first make certain that there are no old libs around to bite us
 # BZ 834847
-if [ -e %{_sysconfdir}/ld.so.conf.d/glusterfs.conf ]; then
-    rm -f %{_sysconfdir}/ld.so.conf.d/glusterfs.conf
+if [ -e /etc/ld.so.conf.d/glusterfs.conf ]; then
```
neha170, line 1074: Please use rpm macros wherever possible. E.g. %{_sysconfdir} in place of /etc

SPECS/glusterfs/glusterfs.spec
bfjelds: i think it should stay.

the more we diverge from the upstream spec.in (by removing these things), the harder it is to update it each time.

SPECS/glusterfs/glusterfs.spec
bfjelds: this is from the upstream spec.in ... see previous comment about aligning to that file.

SPECS/glusterfs/glusterfs.spec
bfjelds: @neha170 , is this an acceptable answer?

bfjelds: out of curiosity, why does it matter that rhel/fedora references exist?

kanikanema: I think, generally for aesthetics and reduction of unnecessary code in the spec but maybe there are other reasons that I am not aware of.

bfjelds: seems to introduce complication for future upgrades

bfjelds:
validated RPM install using local VM:

``` shell
root@ecp-machine [ /tmp/RPMS/x86_64 ]# ls gluster*
glusterfs-11.1-1.azl3.x86_64.rpm                    glusterfs-debuginfo-11.1-1.azl3.x86_64.rpm
glusterfs-cli-11.1-1.azl3.x86_64.rpm                glusterfs-extra-xlators-11.1-1.azl3.x86_64.rpm
glusterfs-client-xlators-11.1-1.azl3.x86_64.rpm     glusterfs-fuse-11.1-1.azl3.x86_64.rpm
glusterfs-cloudsync-plugins-11.1-1.azl3.x86_64.rpm
root@ecp-machine [ /tmp/RPMS/x86_64 ]# rpm -iv --nodeps $(ls gluster*)
Verifying packages...
Preparing packages...
glusterfs-client-xlators-11.1-1.azl3.x86_64
glusterfs-11.1-1.azl3.x86_64
glusterfs-fuse-11.1-1.azl3.x86_64
glusterfs-extra-xlators-11.1-1.azl3.x86_64
glusterfs-debuginfo-11.1-1.azl3.x86_64
glusterfs-cloudsync-plugins-11.1-1.azl3.x86_64
glusterfs-cli-11.1-1.azl3.x86_64
```

SPECS/glusterfs/glusterfs.spec
```diff
@@ -34,463 +45,666 @@
...
-This package provides the api include files.
-
 %package cli
-Summary:        GlusterFS CLI
-License:        GPLv2 OR LGPLv3+
-
-Requires:       %{name}-libs = %{version}-%{release}
+Summary:          GlusterFS CLI
+%if ( 0%{?mariner} && 0%{?mariner} >= 3 )
+BuildRequires:    bash-completion
```
bfjelds, line 321: use bash-completion-devel

anphel31 (approved):

SPECS/glusterfs/glusterfs.spec
```diff
@@ -1,16 +1,27 @@
+#
+# Use mariner=<version> for Mariner-specific settings
+#
+%global mariner 3
```
PawelWMS, line 4: Not needed, just remove the `fedora` and `rhel` specs as if `fedora == 32`. If you got this update from a newer version of Fedora than 32, take that number (and mentione the Fedora version in the changelog).

SPECS/glusterfs/glusterfs.spec
```diff
@@ -34,463 +45,666 @@
...
 %global _pythonver 3
-%global service_start()   /bin/systemctl --quiet start %{1}.service || : \
+%else
+%global _usepython3 0
+%global _pythonver 2
+%endif
+# From https://fedoraproject.org/wiki/Packaging:Python#Macros
+%if ( 0%{?rhel} && 0%{?rhel} <= 6 ) && ( 0%{!?mariner} )
+%{!?python2_sitelib: %global python2_sitelib %(python2 -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")}
+%{!?python2_sitearch: %global python2_sitearch %(python2 -c "from distutils.sysconfig import get_python_lib; print(get_python_lib(1))")}
```
PawelWMS, line 152: Remove **ALL** references to Python2. We've had a major clean-up for that in 2.0 and don't want to bring that back in.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -34,463 +45,666 @@
...
+%global glusterd_svcfile   %{_sysconfdir}/init.d/glusterd
+%global glusterfsd_svcfile %{_sysconfdir}/init.d/glusterfsd
+%global glustereventsd_svcfile %{_sysconfdir}/init.d/glustereventsd
+%endif
 %{!?_pkgdocdir: %global _pkgdocdir %{_docdir}/%{name}-%{version}}
+# We do not want to generate useless provides and requires for xlator
+# .so files to be set for glusterfs packages.
+# Filter all generated:
+#
+# TODO: RHEL5 does not have a convenient solution
```
PawelWMS, line 191: Please also remove comments referring to other distros, since they don't make much sense here. Ideally, also remove the ones, which are still here from the 2.0 version, but I'm not going to push for that.:)

SPECS/glusterfs/glusterfs.spec
```diff
@@ -34,463 +45,666 @@
...
+Requires:         libgfrpc0%{?_isa} = %{version}-%{release}
+Requires:         libgfxdr0%{?_isa} = %{version}-%{release}
+%if ( 0%{?_with_systemd:1} )
+%{?systemd_requires}
+%endif
+%if 0%{?_with_asan:1} && !( 0%{?rhel} && 0%{?rhel} < 7 )
+BuildRequires:    libasan
+%endif
+%if 0%{?_with_tsan:1} && !( 0%{?rhel} && 0%{?rhel} < 7 )
+BuildRequires:    libtsan
```
PawelWMS, line 257: Please keep the order of `BuildRequires` coming first in alhpabetical order, then `Requires`, `Provides`, and `Obsoletes`. If `Provides` and `Obsoletes` form "pairs", it's also fine to keep these pairs together.

PawelWMS: Actually, since this is a new set of packages for 3.0, all `Provides`-`Obsolete` "pairs" like `%{name}-common` for instance can be removed completely. There's no older package to obsolete.

`Provides` without pairs should probably still remain, though.

bfjelds: this is the upstream spec.  if we change it, the next upgrade will be harder.

SPECS/glusterfs/glusterfs.spec
bfjelds: i don't understand why we are making changes to upstream bits.  it will just make life harder for the next upgrader.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -574,55 +850,63 @@ sed -i -e 's/--quiet//' configure.ac
...
+%if ( 0%{?fedora} && 0%{?fedora} > 17 ) || ( 0%{?rhel} && 0%{?rhel} > 6 ) || ( 0%{?mariner} && 0%{?mariner} >= 3 )
 sed -i 's| \\\$compiler_flags |&\\\$LDFLAGS |' libtool
-
+%endif
 sed -i 's|^hardcode_libdir_flag_spec=.*|hardcode_libdir_flag_spec=""|' libtool
 sed -i 's|^runpath_var=LD_RUN_PATH|runpath_var=DIE_RPATH_DIE|' libtool
 
-gcc -v
-%make_build
+make %{?_smp_mflags}
```
PawelWMS, line 863: Why revert to the plain `make`? Wherever possible please keep `%make_build` as it includes any build flags we may wish to add or modify globally.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -574,55 +850,63 @@ sed -i -e 's/--quiet//' configure.ac
...
 
 %check
 make check
 
 %install
-%make_install
-
+rm -rf %{buildroot}
```
PawelWMS, line 869: Not needed.
```suggestion
```

PawelWMS: Same comments as for `%make_build`.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -678,47 +973,77 @@ touch %{buildroot}%{_sharedstatedir}/glusterd/nfs/nfs-server.vol
...
-## Install bash completion for cli
-install -p -m 0755 -D extras/command-completion/gluster.bash \
-    %{buildroot}%{_sysconfdir}/bash_completion.d/gluster
+find ./tests ./run-tests.sh -type f | cpio -pd %{buildroot}%{_prefix}/share/glusterfs
+
+%if 0%{!?mariner}
+# Mariner does not require a clean stage
+%clean
+rm -rf %{buildroot}
+%endif
```
PawelWMS, line 978: No need to comment, just remove.
```suggestion
```

SPECS/glusterfs/glusterfs.spec
```diff
@@ -791,7 +1116,7 @@ exit 0
 ##
 %pre
 getent group gluster > /dev/null || groupadd -r gluster
-getent passwd gluster > /dev/null || useradd -r -g gluster -d %{_rundir}/gluster -s %{_sbindir}/nologin -c "GlusterFS daemons" gluster
+getent passwd gluster > /dev/null || useradd -r -g gluster -d %{_rundir}/gluster -s /sbin/nologin -c "GlusterFS daemons" gluster
```
PawelWMS, line 1115: Please leave the path macros instead of the hard-coded strings.

SPECS/glusterfs/glusterfs.spec
bfjelds: all the changes here are upstream changes.  my goal was to align to upstream except when the mariner variable is used to define behavior specific to mariner.  i can do that here, but didn't know any better.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -811,196 +1136,245 @@ exit 0
...
-%exclude %{_libdir}/glusterfs/%{version}%{?dev}/rpc-transport/rdma*
-%endif
-%if ( 0%{!?_without_server:1} )
+%if 0%{?!_without_server:1}
 %dir %{_datadir}/glusterfs
 %dir %{_datadir}/glusterfs/scripts
-%{_datadir}/glusterfs/scripts/post-upgrade-script-for-quota.sh
-%{_datadir}/glusterfs/scripts/pre-upgrade-script-for-quota.sh
+     %{_datadir}/glusterfs/scripts/post-upgrade-script-for-quota.sh
+     %{_datadir}/glusterfs/scripts/pre-upgrade-script-for-quota.sh
```
PawelWMS, line 1226: Why the indent?

SPECS/glusterfs/glusterfs.spec
bfjelds: all the changes here are upstream changes.  my goal was to align to upstream except when the mariner variable is used to define behavior specific to mariner.  i can do that here if you think it is important.

SPECS/glusterfs/glusterfs.spec
bfjelds: all the changes here are upstream changes.  my goal was to align to upstream except when the mariner variable is used to define behavior specific to mariner.  my preference would be to make the choice overt so the next upgrader can understand.

SPECS/glusterfs/glusterfs.spec
bfjelds: all the changes here are upstream changes.  my goal was to align to upstream except when the mariner variable is used to define behavior specific to mariner.  i can make this change, but it is going to make the next upgrade harder.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -1189,23 +1607,23 @@ exit 0
...
-%dir %{_libexecdir}/glusterfs
-%{_libexecdir}/glusterfs/glusterfind
-%{_libexecdir}/glusterfs/peer_add_secret_pub
+     %{_libexecdir}/glusterfs/glusterfind
 %{_bindir}/glusterfind
+     %{_libexecdir}/glusterfs/peer_add_secret_pub
 
 %if ( 0%{?_with_firewalld:1} )
-%{_libdir}/firewalld/services/glusterfs.xml
+%{_prefix}/lib/firewalld/services/glusterfs.xml
```
PawelWMS, line 1621: Previous macro was more accurate/less error-prone in case we ever touch these macros.

I think it might be worth to run the spec linter for this spec once other changes are done to update it to use our path macros.

SPECS/glusterfs/glusterfs.spec
bfjelds: upstream indented the lines

SPECS/glusterfs/glusterfs.spec
bfjelds: i started a thread to come to some understanding about modifying upstream bits (like this spec file).

SPECS/glusterfs/glusterfs.spec
bfjelds: i find this variable an easy-to-understand way of communicating choices explicitly with the next upgrader.

SPECS/glusterfs/include-eventtypes-always.patch
```diff
@@ -0,0 +1,14 @@
...
+index 68b7996519..8874698bb2 100644
+--- a/libglusterfs/src/glusterfs/events.h
++++ b/libglusterfs/src/glusterfs/events.h
+@@ -11,8 +11,8 @@
+ #ifndef __EVENTS_H__
+ #define __EVENTS_H__
+ 
+-#ifdef USE_EVENTS
+ #include "eventtypes.h"
++#ifdef USE_EVENTS
```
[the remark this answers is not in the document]
bfjelds: i dont think it can be handled that way.  the code has a #if and a #else, both of which require eventtypes.h.

PawelWMS (changes requested):
A few comments and some clean-up.

SPECS/glusterfs/glusterfs.spec
bfjelds: removed

SPECS/glusterfs/glusterfs.spec
bfjelds: removed (debate can continue, but want to get this merged)

SPECS/glusterfs/glusterfs.spec
bfjelds: removed

SPECS/glusterfs/glusterfs.spec
bfjelds: modified per suggestion

SPECS/glusterfs/glusterfs.spec
bfjelds: modified per suggestion

SPECS/glusterfs/glusterfs.spec
bfjelds: modified per suggestion

SPECS/glusterfs/glusterfs.spec
bfjelds: modified per suggestion

SPECS/glusterfs/glusterfs.spec
bfjelds: modified per suggestion

SPECS/glusterfs/include-eventtypes-always.patch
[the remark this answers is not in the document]
bfjelds: glusterfs in mariner 2.0 built with disable-events, when doing that for 3.0, the header file is not included (despite being needed) without this patch.

SPECS/glusterfs/glusterfs.spec
bfjelds: i'm going to leave this alone unless it is important ... reopen if you feel it is vital.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -733,11 +890,11 @@ fi
 # to /var/lib. (N.B. Starting with 3.3.0 all gluster files are in /var/lib
 # in gluster.org RPMs.) Be careful to copy them on the off chance that
 # /etc and /var/lib are on separate file systems
-if [ -d %{_sysconfdir}/glusterd -a ! -h %{_sharedstatedir}/glusterd ]; then
+if [ -d /etc/glusterd -a ! -h %{_sharedstatedir}/glusterd ]; then
```
bfjelds, line 893: more source spec file paths that need replacing with variables

SPECS/glusterfs/glusterfs.spec
```diff
@@ -811,196 +968,222 @@ exit 0
...
 %if ( 0%{!?_without_server:1} )
 %postun server
-%{?ldconfig}
 %if (0%{?_with_firewalld:1})
     %firewalld_reload
 %endif
 exit 0
 %endif
 
+%if ( 0%{!?_without_server:1} )
```
bfjelds, line 1012: can probably remove empty %if blocks

SPECS/glusterfs/glusterfs.spec
bfjelds: have grouped like things and alphabetized them.  left provides-obselete pairs because they existed previously.

SPECS/glusterfs/include-eventtypes-always.patch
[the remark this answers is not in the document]
PawelWMS: Well I'll be, indeed it looks like a bug in the upstream ([this change](https://github.com/gluster/glusterfs/commit/3b00855b0f69e4cc36160d56b36f6b011bdd1177#diff-aead3630bb5df6aed67b31bea331412c2c900f295e76a8f369e4a84096783880) might be the culprit). I was not expecting that. In that case I'd only ask to make the comment above this patch in the spec more robust, explaining in more detail what was wrong (I understand it's actually impossible to compile correctly with `--disable-events`), maybe with the above commit link if you also believe this may be the source of the issue. Ideally, please also open an issue in upstream's GitHub to link it here.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -1216,16 +1438,23 @@ exit 0
...
+%if ( 0%{?_with_systemd:1} )
 %{_unitdir}/glustereventsd.service
+%else
+%{_sysconfdir}/init.d/glustereventsd
+%endif
 %endif
 
 %changelog
+* Thu Dec 14 2023  Brian Fjeldstad <bfjelds@microsoft.com> - 11.1-1
+- Upgrade to 11.1
```
PawelWMS, line 1455: In case of updates where we use other distro's specs to guide us, please also include a small note about the version to both help future us to understand where the upgrade came from and to give credit to the other authors. For example (taken from other specs):
```suggestion
- Upgrade to 11.1 using Fedora 39 spec for guidance (license: MIT).
```

SPECS/glusterfs/glusterfs.spec
```diff
@@ -34,35 +34,42 @@
...
 # rpmbuild -ta @PACKAGE_NAME@-@PACKAGE_VERSION@.tar.gz --without libtirpc
 %{?_without_libtirpc:%global _without_libtirpc --without-libtirpc}
+# libtcmalloc
+# if you wish to compile an rpm without tcmalloc (i.e. use gluster mempool)
+# rpmbuild -ta @PACKAGE_NAME@-@PACKAGE_VERSION@.tar.gz --without tcmalloc
+%{?_without_tcmalloc:%global _without_tcmalloc --without-tcmalloc}
+%ifnarch x86_64
+%global _without_tcmalloc --without-tcmalloc
+%endif
+# Use same malloc that previous CBL-Mariner used
```
PawelWMS, line 57: Please expand mentioning the number of the previous version. Makes potential future lookup a bit easier. And since we're switching names, we might as well:
```suggestion
# Use same malloc that previous (<version>) Azure Linux used
```

SPECS/glusterfs/glusterfs.spec
```diff
@@ -34,35 +34,42 @@
...
-# No RDMA Support on armv7hl
-%ifarch armv7hl
-%global _without_rdma --disable-ibverbs
-%endif
 # server
-# Disabling 'server' subpackage for CBL-Mariner as it's currently unnecessary.
+# if you wish to build rpms without server components, compile like this
+# rpmbuild -ta @PACKAGE_NAME@-@PACKAGE_VERSION@.tar.gz --without server
+%{?_without_server:%global _without_server --without-server}
+# disable server components forcefully as rhel <= 6
```
PawelWMS, line 67: Since for this one we already had our custom configuration with a comment, I'd leave it as it was:
```suggestion
Disabling 'server' subpackage for CBL-Mariner as it's currently unnecessary.
```

SPECS/glusterfs/glusterfs.spec
```diff
@@ -34,35 +34,42 @@
...
 %global _without_server --without-server
-# disable syslog forcefully as rhel <= 6 doesn't have rsyslog or rsyslog-mmcount
-# Fedora deprecated syslog, see
-#  https://fedoraproject.org/wiki/Changes/NoDefaultSyslog
-# (And what about RHEL7?)
-%global _without_syslog --disable-syslog
+# syslog
+# if you wish to build rpms without syslog logging, compile like this
+# rpmbuild -ta @PACKAGE_NAME@-@PACKAGE_VERSION@.tar.gz --without syslog
+%{?_without_syslog:%global _without_syslog --disable-syslog}
```
PawelWMS, line 72: I think we are now accidentally enabling syslog, since we don't have the `_without_syslog` macro anywhere.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -74,74 +81,116 @@
...
 # without server should also disable some server-only components
-%if ( 0%{?_without_server:1} )
+%if 0%{?_without_server:1}
 %global _without_events --disable-events
 %global _without_georeplication --disable-georeplication
+%global _without_linux_io_uring --disable-linux-io_uring
 %global _with_gnfs %{nil}
 %global _without_ocf --without-ocf
 %endif
+%global _usepython3 1
```
PawelWMS, line 101: We don't really have other options in our case. In such cases we can just remove the macro and wherever it's used assume it was set to `1` to clean-up any `%if` macros as well.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -74,74 +81,116 @@
 ##-----------------------------------------------------------------------------
 ## All %%global definitions should be placed here and keep them sorted
 ##
+# selinux booleans whose defalut value needs modification
+# these booleans will be consumed by "%%selinux_set_booleans" macro.
+%global _with_systemd true
```
PawelWMS, line 86: What is this part about? I couldn't find `_with_systemd` inside the macros file defining `selinux_set_booleans`. Can you explain how this is used in this spec?

SPECS/glusterfs/glusterfs.spec
```diff
@@ -74,74 +81,116 @@
...
 %{!?_pkgdocdir: %global _pkgdocdir %{_docdir}/%{name}-%{version}}
 # modern rpm and current Fedora do not generate requires when the
 # provides are filtered
 %global __provides_exclude_from ^%{_libdir}/glusterfs/%{version}/.*$
+%global bashcompdir %(pkg-config --variable=completionsdir bash-completion 2>/dev/null)
+%if "%{bashcompdir}" == ""
+%global bashcompdir ${sysconfdir}/bash_completion.d
+%endif
+# Skip regression-tests because they previously were excluded in
+# CBL-Mariner glusterfs and the new version introduces a dependency
```
PawelWMS, line 144: In this case please also mention specific version number just to make it easier to follow in the future.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -74,74 +81,116 @@
...
+# (dbench)
+%global _without_regression_tests true
+# Explicitly require rpcgen
+%global _require_rpcgen true
```
PawelWMS, line 148: Why do we need this macro? I think we can just leave the parts of the spec it's guarding as enabled and drop it. Especially since we already had `BuildRequires: rpcgen` for a while.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -74,74 +81,116 @@
...
 Source2:        glusterfsd.sysconfig
 Source7:        glusterfsd.service
 Source8:        glusterfsd.init
-Patch0001:      0001-rpc-rpc-lib-src-Makefile.am.patch
+
+# Patch created for Mariner to fix build break related to
+# including eventtypes.h
+Patch001:       include-eventtypes-always.patch
+
+BuildRoot:        %(mktemp -ud %{_tmppath}/%{name}-%{version}-%{release}-XXXXXX)
```
PawelWMS, line 170: Why do we need to define this explicitly for this spec?

SPECS/glusterfs/glusterfs.spec
```diff
@@ -74,74 +81,116 @@
...
 BuildRequires:  libxml2-devel
 BuildRequires:  make
 BuildRequires:  ncurses-devel
 BuildRequires:  openssl-devel
+BuildRequires:  openssl
 BuildRequires:  python%{_pythonver}-devel
 BuildRequires:  readline-devel
-BuildRequires:  rpcgen
-BuildRequires:  systemd
 BuildRequires:  userspace-rcu-devel >= 0.7
```
PawelWMS, line 186: Nit-pick: can we keep the dependencies in alphabetical order as much as possible?

SPECS/glusterfs/glusterfs.spec
bfjelds: because without it the build fails.

SPECS/glusterfs/glusterfs.spec
bfjelds: ok ... just makes even more departure from the source spec.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -151,346 +200,411 @@ BuildRequires:  libattr-devel
...
+
+%if ( 0%{!?_without_regression_tests:1} )
+%package regression-tests
+Summary:          Development Tools
+Requires:         %{name}%{?_isa} = %{version}-%{release}
+Requires:         %{name}-fuse%{?_isa} = %{version}-%{release}
+Requires:         %{name}-server%{?_isa} = %{version}-%{release}
+## thin provisioning support
+Requires:         lvm2 >= 2.02.89
+Requires:         perl(App::Prove) perl(Test::Harness) gcc util-linux-ng
```
PawelWMS, line 580: Please use one dependency per line.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -151,346 +200,411 @@ BuildRequires:  libattr-devel
...
+into one large parallel network filesystem. GlusterFS is one of the
+most sophisticated file systems in terms of features and extensibility.
+It borrows a powerful concept called Translators from GNU Hurd kernel.
+Much of the code in GlusterFS is in user space and easily manageable.
+
+This package provides libglusterfs.so and the gluster C header files.
+
+%package -n libgfapi0
+Summary:          GlusterFS api library
+Requires:         libglusterfs0%{?_isa} = %{version}-%{release}
```
PawelWMS, line 430: There's quite a bunch of package additions and reshuffling happening in this spec and I can't get through all of the dependencies they have. Can you tell me, if in case of new dependencies (assuming there are any), they will be satisfied by our packages? I'm a bit on the fence when it comes to introducing new packages (I think there are some although most only look like a re-name of existing ones) but if we're going to do so, we must make sure we have their run-time dependencies.

SPECS/glusterfs/glusterfs.spec
```diff
@@ -574,55 +718,56 @@ sed -i -e 's/--quiet//' configure.ac
...
-gcc -v
 %make_build
 
 %check
 make check
 
 %install
-%make_install
-
+make install DESTDIR=%{buildroot}
```
PawelWMS, line 739:
```suggestion
%make_install
```

SPECS/glusterfs/glusterfs.spec
```diff
@@ -574,55 +718,56 @@ sed -i -e 's/--quiet//' configure.ac
...
 %if ( 0%{!?_without_server:1} )
+%if ( 0%{_for_fedora_koji_builds} )
```
PawelWMS, line 741: I'd remove all references to `_for_fedora_koji_builds`. Right now I think we're not erroring-out, because these `%if` statements are never analyzed by RPM due to the earlier check for `_without_server` or other macros, which evaluate to a false statement. But these Fedora macros will never be present in our system and with the way they are used now, they are prone to unnecessarily generate failures. I'd just assume the macro is not defined and update the spec accordingly.

SPECS/glusterfs/glusterfs.spec
PawelWMS: Not vital and at a second glance look actually pretty helpful.

SPECS/glusterfs/glusterfs.spec
PawelWMS: I think we could use more details here. What is the build complaining about? Earlier we were building fine without it and in general, if we'll end up needing it, we could use a good explanation why we used it here, so in the future it's easier to understand what happened.

SPECS/glusterfs/glusterfs.spec
bfjelds: i think this is a misplaced comment.  systemd has nothing to do with selinux.

SPECS/glusterfs/glusterfs.spec
bfjelds: previously, the sections were just deleted.  so i guess i'll do the same instead of having a conditional variable.

SPECS/glusterfs/glusterfs.spec
bfjelds: regression test code is deleted now

SPECS/glusterfs/glusterfs.spec
bfjelds: i'm not sure how to do what you're asking or what it refers to.  can you provide more info?

bfjelds: from a quick scan of the Requires

```
$ cat ./SPECS/glusterfs/glusterfs.spec | grep -v '^\-' | grep -v BuildRequir | grep Requires | cut -c19- | awk '{print $1}' | sort --unique
attr
dbus
firewalld-filesystem
gperftools-libs%{?_isa}
libacl-devel
%{_libdir}/ocf/resource.d
libgfapi0%{?_isa}
libgfchangelog0%{?_isa}
libgfrpc0%{?_isa}
libgfrpc-devel%{?_isa}
libgfxdr0%{?_isa}
libgfxdr-devel%{?_isa}
libglusterfs0%{?_isa}
libglusterfs-devel%{?_isa}
lvm2
%{name}-client-xlators%{?_isa}
%{name}-cli%{?_isa}
%{name}-fuse%{?_isa}
%{name}%{?_isa}
%{name}-server
%{name}-server%{?_isa}
nfs-ganesha-gluster
nfs-ganesha-selinux
nfs-utils
pcs
portmap
psmisc
python%{_pythonver}
python%{_pythonver}-gluster
python%{_pythonver}-prettytable
python%{_pythonver}-requests
pyxattr
resource-agents
rsync
shadow-utils
tar
util-linux
valgrind
```

to me, these seem to be the only ones that are unfulfilled:

nfs-ganesha-gluster [package ganesha]
nfs-ganesha-selinux [package ganesha]
pcs [package ganesha]
pyxattr (in EXTENDED) [package server]

bfjelds: server is disabled.  maybe it should be deleted from the spec rather than disabled?

i'm not sure what it is, but maybe we disable/delete the ganesha package too?

bfjelds: removing ganesha package

PawelWMS (approved):

PR #7215, opened 2024-01-10
Introduce Rust virtiofsd package

sprt:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This introduces the new virtiofsd package written in Rust. This will be needed for kata-containers as upstream recently made the switch from C to the Rust version of virtiofsd.

* New standalone Rust package virtiofsd
* Installs only the `virtiofsd-rs` binary, not the `vhostuser-backend(fs)` QEMU-related file to not break existing users
* No coupling/conflict/dependency with QEMU
* ⚠️ The Kata config will have to be changed to point to this new binary

See https://github.com/microsoft/CBL-Mariner/pull/7215#issuecomment-1896168734 for a summary of the discussion.

**NO**

###### Test Methodology

- Local build, install and running `/usr/libexec/virtiofsd-rs -h` ✅
- Pipeline build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=485293&view=results


SPECS/virtiofsd/virtiofsd.spec
```diff
@@ -0,0 +1,61 @@
...
+Source1:        %{name}-v%{version}-cargo.tar.gz
+Source2:        config.toml
+
+ExclusiveArch:  x86_64
+
+BuildRequires:  cargo
+BuildRequires:  libcap-ng-devel
+BuildRequires:  libseccomp-devel
+
+Requires:       qemu-common
```
sprt, line 25: This is required because of where we install `50-qemu-virtiofsd.json`.

reubeno: Should that QEMU integration conf file be in a separate optional sub-package, then, and have this requires on that subpackage? Are there valid scenarios for using an alternative to QEMU (say, cloud-hypervisor) that would benefit from virtiofsd without needing to pull in QEMU components?

sprt: Yes I had the same thought. Our offering is based on Cloud Hypervisor, though we definitely want to maintain compat with QEMU. With that said, I'm not sure what to call that subpackage - `virtiofsd-qemu` seems awkward given the existing `qemu-virtiofsd`.

reubeno: Hmm. Looking at [Fedora's `virtiofsd` spec](https://src.fedoraproject.org/rpms/virtiofsd/blob/rawhide/f/virtiofsd.spec) for inspiration, I see they took the same approach of just making the whole `virtiofsd` package take a dependency on `qemu-common`. Interestingly, I also see they mark an explicit `Conflicts` with their `qemu-virtiofsd` sub-package.

Stepping back a bit, do we need the qemu conf file from the virtiofsd package? (Does cloud-hypervisor require it, or just if someone wanted to use qemu with this?) Thinking ahead toward Mariner 3.0, we'll presumably upgrade to QEMU 8.0+, which according to [this](https://www.qemu.org/docs/master/about/removed-features.html) has outright dropped its own internal virtiofs daemon in favor of this Rust one so we could actually have `qemu` or one of its subpackages depend on this new `virtiofsd` package, and in so doing arrange to pull in the right conf files? Just brainstorming here...

sprt: > Hmm. Looking at [Fedora's `virtiofsd` spec](https://src.fedoraproject.org/rpms/virtiofsd/blob/rawhide/f/virtiofsd.spec) for inspiration, I see they took the same approach of just making the whole `virtiofsd` package take a dependency on `qemu-common`. Interestingly, I also see they mark an explicit `Conflicts` with their `qemu-virtiofsd` sub-package.

Yeah I also looked at Fedora for inspiration. The `Conflicts` makes sense as you probably don't want to have both the C and Rust versions installed simultaneously, plus they install into the same location anyway. The `Obsoletes/Provides` weren't directly obvious to me but your comment about QEMU 8.0 clarify that a bit. The `Obsoletes` comparison should probably use `<` though (see also [the Fedora docs](https://docs.fedoraproject.org/en-US/packaging-guidelines/#renaming-or-replacing-existing-packages)). Ultimately I decided not to include `Obsoletes` as there's minor API incompatibilities between the C and Rust versions - I think Fedora doesn't care because for them QEMU is the only consumer of virtiofsd, and they updated it.

> Stepping back a bit, do we need the qemu conf file from the virtiofsd package? (Does cloud-hypervisor require it, or just if someone wanted to use qemu with this?) Thinking ahead toward Mariner 3.0, we'll presumably upgrade to QEMU 8.0+, which according to [this](https://www.qemu.org/docs/master/about/removed-features.html) has outright dropped its own internal virtiofs daemon in favor of this Rust one so we could actually have `qemu` or one of its subpackages depend on this new `virtiofsd` package, and in so doing arrange to pull in the right conf files? Just brainstorming here...

I like that idea! Looking at [Fedora's QEMU 8 spec](https://src.fedoraproject.org/rpms/qemu/blob/rawhide/f/qemu.spec#_228), `qemu-virtiofsd` is indeed gone and `qemu` requires `vhostuser-backend(fs)` instead of `virtiofsd` explicitely.

Unfortunately there's still tight coupling between `vhostuser-backend` and QEMU by design. Fedora works with that coupling (their `virtiofsd` is coupled with `vhostuser-backend`), but if we want to work around it we'd have to deviate from Fedora.

Knowing all this, here's what we could do for **3.0**:

  * Assuming QEMU 8
    * Requires `vhostuser-backend(fs)`
  * Introduce standalone Rust package `virtiofsd` (provides **only** the `virtiofsd` binary)
  * Keep `qemu-virtiofsd`
    * Requires `virtiofsd`
    * Provides `vhostuser-backend(fs)` (by installing the JSON file)
    * ⚠️ Caveat is that the versioning scheme will have to change, and AFAIK Mariner doesn't do epochs?

And actually since the new virtiofsd is a breaking change, we probably want to install it in a different location for **2.0**:

  * With current QEMU < 8 requiring `vhostuser-backend(fs)`
  * Introduce standalone Rust package `virtiofsd` (provides **only** the `virtiofsd` binary at e.g. `/usr/libexec/virtiofsd-rs`)
    * Do **not** provide `vhostuser-backend(fs)`, as I believe our current QEMU version is incompatible
  * Keep `qemu-virtiofsd` unchanged

Thoughts?

sprt: > as I believe our current QEMU version is incompatible

Correction here as we didn't find evidence that QEMU is incompatible, rather it was a new version of Kata making calls to the C binary using options introduced by the Rust virtiofsd. Nevertheless, it could be that the Rust package breaks other use cases, in which case we still wouldn't want to provide `vhostuser-backend(fs)`.

dallasd1: Do we need to change the installation path if virtiofsd conflicts with qemu-virtiofsd?

sprt:
Summary of our call for 2.0:
 * Keep qemu-virtiofsd unchanged to not break existing users
 * Introduce standalone Rust virtiofsd
   * ONLY the virtiofsd binary, in a different location from the C one
   * NO Conflicts: qemu-virtiofsd or Requires: qemu-common: so that both qemu and virtiofsd can coexist (note that qemu-common depends on qemu-virtiofsd through vhostuser-backend(fs))
   * NO Provides: vhostuser-backend(fs): so that existing users of the vhostuser-backend(fs) capability are not broken
     * Also NOT installing the vhostuser-backend(fs) JSON file, as that could cause unexpected conflicts with the one from qemu-virtiofsd

SPECS/virtiofsd/config.toml
```diff
@@ -0,0 +1,5 @@
+[source.crates-io]
```
dallasd1, line 1: Is this a standard way of adding a config for vendored sources? Comparing with the way we vendor kata-cc, the directory .config/ and file config.toml are included in the "cargo.tar.gz". So for those cases there isn't a need to include this config

dallasd1 (approved):

SPECS/virtiofsd/config.toml
sprt: Discussed offline but this is following cloud-hypervisor: https://github.com/microsoft/CBL-Mariner/blob/3798a92d85758462fd12f7fa8abce93802de6234/SPECS/cloud-hypervisor/cloud-hypervisor.spec#L16-L23

SPECS/virtiofsd/virtiofsd.spec
```diff
@@ -0,0 +1,55 @@
...
+# To update the vendor tarball and config.toml:
+#   wget %{SOURCE0}
+#   tar -xf %{name}-v%{version}.tar.gz
+#   cd %{name}-v%{version}
+#   cargo vendor > ../config.toml
+#   tar -czf ../%{name}-v%{version}-cargo.tar.gz vendor/
+Source1:        %{name}-v%{version}-cargo.tar.gz
+Source2:        config.toml
+
+ExclusiveArch:  x86_64
```
mfrw, line 20: Is there a reason for only building for `x86_64` ?

PR #7294, opened 2024-01-17
fix python urllib3 test

mandeepsplaha:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR fixes the failing test called `test_recent_date` by updating the source file containing the RECENT_DATE defined in file `src/urllib3/connection.py` which is used in the test.

###### Change Log  

Package: python-urllib3
Change: Fix test_recent_date test by updating the hard-coded date in file `src/urllib3/connection.py`.

**NO**

###### Test Methodology

- Failing Test build: [479264](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=479264&view=ms.vss-test-web.build-test-results-tab&runId=153941&resultId=100190&paneView=attachments)
- Passing Test build: [485241](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=485241&view=ms.vss-test-web.build-test-results-tab)


SPECS/python-urllib3/urllib3_test_recent_date.patch
```diff
@@ -0,0 +1,12 @@
+diff -ur a/src/urllib3/connection.py b/src/urllib3/connection.py
+--- a/src/urllib3/connection.py	2023-10-17 10:26:30
++++ b/src/urllib3/connection.py	2024-01-17 08:03:33
+@@ -68,7 +68,7 @@
+ 
+ # When it comes time to update this value as a part of regular maintenance
+ # (ie test_recent_date is failing) update it to ~6 months before the current date.
+-RECENT_DATE = datetime.date(2022, 1, 1)
++RECENT_DATE = datetime.date(2023, 8, 1)
```
PawelWMS, line 9: Since this is a temporary solution, could we make it so, that we always get a date that's 6 months before current date? Example from [here](https://stackoverflow.com/questions/546321/how-do-i-calculate-the-date-six-months-from-the-current-date-using-the-datetime):

```python
RECENT_DATE = datetime.date.today() + dateutil.relativedelta(months=-6)
```

Might need adding some imports to simplify/work correctly.

PawelWMS (approved):
Approving, but please consider making the fix more long-lasting.

SPECS/python-urllib3/urllib3_test_recent_date.patch
mandeepsplaha: I don't think this is a temporary change; it's in line with what the suggestion in the source file is. I thought this test was set up to fail every two years to prompt the maintainers to regularly update the package. As in, a failing test draws attention to this package. Making the suggested change makes the test ineffective because it will always pass. It's kind of similar to removing it altogether. If you really want to make this change, I can make it because I do understand that we are not the ones maintaining this package and that it will make our lives easier in terms of p-tests.

PawelWMS: I see. In that case, please double-check what `RECENT_DATE` is used for, since it's inside regular code, not only part of the test. If you're certain it's not used for anything else than to signal the code is getting old, then I'm fine removing the test along with the variable itself, since we're not going to update `urllib3` to a newer version in Mariner 2.0.

mandeepsplaha: RECENT_DATE is used in another place in regular code:

```
is_time_off = datetime.date.today() < RECENT_DATE
if is_time_off:
    warnings.warn(
        (
            "System time is way off (before {0}). This will probably "
            "lead to SSL verification errors"
        ).format(RECENT_DATE),
        SystemTimeWarning,
    )
```
And there's another test called `test_ssl_wrong_system_time` that tests this warning message if the current date falls before RECENT_DATE.

mandeepsplaha: As discussed in PR https://github.com/microsoft/CBL-Mariner/pull/7295 and offline, we are good to merge this PR as is.

PR #7315, opened 2024-01-18
update systemd to 255

ddstreetmicrosoft:
This contains many changes; please see the individual commits for specific details.

- fixes PR checks that were breaking due to our PR check runner using Ubuntu with an older rpm that doesn't understand %bcond
- move libmicrohttpd from extended to base
- update libmicrohttpd to 0.9.77
- update libxkbcommon to 1.6.0
- convert zstd to fedora packaging
- update rpm, crash, squashfs-tools, systemd dep statements to use the standard pkgconfig(libzstd) instead of zstd-devel, and standard libzstd instead of zstd-libs
- add python-zstd
- change polkit to require 'bootstrap' systemd packages
- update systemd to 255


ddstreetmicrosoft:
pr check failure is due to #7414 and should be ignored

SPECS/systemd/systemd.spec
```diff
@@ -1,62 +1,342 @@
...
+%endif
+BuildRequires:  libidn2-devel
+BuildRequires:  libcurl-devel
+BuildRequires:  kmod-devel
+BuildRequires:  elfutils-devel
+BuildRequires:  openssl-devel
+%if %{with gnutls}
+BuildRequires:  gnutls-devel
+%endif
+%if %{undefined azure}
```
PawelWMS, line 161: If we hard-code this macro to a specific value, what's the point of the `%if %{undefined azure}`? Can't we just remove these lines? Also, why do we need the `azure` macro at all in this spec?

SPECS/systemd/systemd.spec
```diff
@@ -1,62 +1,342 @@
...
+Source15:       10-oomd-per-slice-defaults.conf
+Source16:       10-timeout-abort.conf
+Source17:       10-map-count.conf
+
+Source21:       macros.sysusers
+Source22:       sysusers.attr
+Source23:       sysusers.prov
+Source24:       sysusers.generate-pre.sh
+
+Source25:       98-default-mac-none.link
```
PawelWMS, line 88: Why all the new config files? Was the 254->255 version bump a significant change? Or did you just want to make our configuration explicit while you were updating this package (+1 to that)?

SPECS/systemd/systemd.spec
```diff
@@ -1,62 +1,342 @@
...
+Summary:        System and Service Manager
+
+# download tarballs with "spectool -g systemd.spec"
+%if %{defined commit}
+Source0:        https://github.com/systemd/systemd%{?stable:-stable}/archive/%{commit}/%{name}-%{shortcommit}.tar.gz
+%else
+%if 0%{?stable}
+Source0:        https://github.com/systemd/systemd-stable/archive/v%{version_no_tilde}/%{name}-%{version_no_tilde}.tar.gz
+%else
+Source0:        https://github.com/systemd/systemd/archive/v%{version_no_tilde}/%{name}-%{version_no_tilde}.tar.gz
```
PawelWMS, line 61: I don't understand. If we don't use the `commit` macro and `stable` is hard-coded (and I believe we don't really want to use non-stable versions), then why the `if-else`?

SPECS/systemd/systemd.spec
```diff
@@ -1,62 +1,342 @@
...
-BuildRequires:  docbook-style-xsl
-BuildRequires:  gettext
-BuildRequires:  glib-devel
-BuildRequires:  gperf
-BuildRequires:  intltool
-BuildRequires:  kbd
-BuildRequires:  kmod-devel
+Url:            https://systemd.io
+%if %{without inplace}
+Version:        255
```
PawelWMS, line 39: This is an enormous change. Was there an issue with a simple version bump and getting the new sources? If this change is about more than just updating the version of `systemd`, please also include these details in the PR's summary.

SPECS/systemd/systemd.spec
ddstreetmicrosoft: the 'azure' macro won't be hardcoded, it'll come from the standard location in `macros.dist` from the `azure-release-common` package (just like the `fedora-release-common` package); my priority is currently systemd so i'm updating this before changing that package, and then i'll remove the hardcoded macro from here.

SPECS/systemd/systemd.spec
ddstreetmicrosoft: direct from fedora

SPECS/systemd/systemd.spec
ddstreetmicrosoft: direct from fedora

SPECS/systemd/systemd.spec
ddstreetmicrosoft: yes, we need to shift to fedora's packaging

SPECS/kernel/kernel.spec
```diff
@@ -29,7 +29,7 @@
 Summary:        Linux Kernel
 Name:           kernel
 Version:        6.6.2.1
-Release:        3%{?dist}
+Release:        4%{?dist}
```
christopherco, line 32: Due to some current entangled kernel-related specs, `kernel-signed`, `kernel-headers`, and corresponding manifest .txt files need to also be updated (i.e., https://github.com/microsoft/CBL-Mariner/commit/b6646353c8ddda6768cfa5399eaf1865bf67c86e)

christopherco (changes requested):
systemd update itself LGTM. kernel-related specs need to match kernel version

SPECS/kernel/kernel.spec
ddstreetmicrosoft: ack, updated all of them to match, thanks

ddstreetmicrosoft:
@christopherco if you have time can you re-review please

@dmcilvaney @PawelWMS also please take a look as this modifies toolkit and pr check stuff

SPECS/python-zstd/python-zstd.spec
```diff
@@ -0,0 +1,108 @@
...
+
+Name:           python-%{pypi_name}
+Version:        1.5.5.1
+Release:        2%{?dist}
+Summary:        Zstd Bindings for Python
+
+# original zstd bits are GPL-2.0-or-later OR BSD-2-Clause
+License:        BSD-2-Clause AND (GPL-2.0-or-later OR BSD-2-Clause)
+URL:            https://github.com/sergey-dryabzhinsky/python-zstd
+Source:         %{pypi_source}
```
bfjelds, line 12: where is pypi_source defined?

ddstreetmicrosoft: python-srpm-macros

dmcilvaney (approved):

SPECS/zstd/zstd.spec
```diff
@@ -1,96 +1,323 @@
...
-%files doc
-%{_mandir}/man1/*
+%ldconfig_scriptlets -n lib%{name}
 
 %changelog
-* Tue Jan 16 2024 Andrew Phelps <anphel@microsoft.com> - 1.5.5-2
-- Add pzstd binary
+* Wed Jan 31 14:14:49 EST 2024 Dan Streetman <ddstreet@ieee.org> - 1.5.5-5
+- Update to version from Fedora 39.
+- Next line is present only to avoid tooling failures, and does not indicate the actual package license.
```
anphel31, line 156: Just a note; I believe the (MIT) license on the next line is for the Fedora SPEC, not for the zstd source

anphel31: side note; we might need to update the following to use `Initial Azure Linux import` https://github.com/microsoft/CBL-Mariner/blob/3.0-dev/toolkit/scripts/spec_source_attributions.py

.github/workflows/check-manifests.yml
```diff
@@ -18,6 +18,10 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v4
 
+    - name: Define missing rpm macros
```
dmcilvaney, line 21: Would this be better in `rpmops.sh`? We use that in a few spots. Maybe something like
```bash
# Mariner macro files used during spec parsing.
MACROS=()
for macro_file in "$SPECS_DIR"/mariner-rpm-macros/macros* "$SPECS_DIR"/pyproject-rpm-macros/macros.pyproject "$SPECS_DIR"/perl/macros.perl
do
  MACROS+=("--load=$macro_file")
done

if [[ -n $(rpm --eval '%bcond test 1') ]]; then
    temp_macros="$(mktemp)"
    echo '%bcond() %[ (%{2}) ? "%{expand:%%bcond_without %{1}}" : "%{expand:%%bcond_with %{1}}" ]' > "$temp_macros"
    MACROS+=("--load=$temp_macros")
```

toolkit/resources/manifests/package/pkggen_core_x86_64.txt
```diff
@@ -74,9 +74,9 @@ xz-5.4.4-1.azl3.x86_64.rpm
 xz-devel-5.4.4-1.azl3.x86_64.rpm
 xz-lang-5.4.4-1.azl3.x86_64.rpm
 xz-libs-5.4.4-1.azl3.x86_64.rpm
-zstd-1.5.5-2.azl3.x86_64.rpm
-zstd-devel-1.5.5-2.azl3.x86_64.rpm
-zstd-libs-1.5.5-2.azl3.x86_64.rpm
+zstd-1.5.5-5.azl3.x86_64.rpm
+libzstd-1.5.5-5.azl3.x86_64.rpm
```
anphel31, line 78: This might not be a big deal, but just noting: When building the worker chroot, the packages are installed in the order listed in this file. With this change the zstd libs and devel package install order is swapped. Probably not a big deal but it would be safest to preserve the order (`zstd` -> `libzstd-devel` -> `libzstd`)

toolkit/resources/manifests/package/toolchain_x86_64.txt
```diff
@@ -606,8 +606,8 @@ zip-debuginfo-3.0-5.azl3.x86_64.rpm
...
 zlib-debuginfo-1.3-1.azl3.x86_64.rpm
 zlib-devel-1.3-1.azl3.x86_64.rpm
-zstd-1.5.5-2.azl3.x86_64.rpm
-zstd-debuginfo-1.5.5-2.azl3.x86_64.rpm
-zstd-devel-1.5.5-2.azl3.x86_64.rpm
-zstd-doc-1.5.5-2.azl3.x86_64.rpm
-zstd-libs-1.5.5-2.azl3.x86_64.rpm
+zstd-1.5.5-5.azl3.x86_64.rpm
+zstd-debuginfo-1.5.5-5.azl3.x86_64.rpm
+libzstd-1.5.5-5.azl3.x86_64.rpm
```
anphel31, line 611: the `toolchain_*.txt` files are listed in alphabetical order. Can you please alphabetize this file

anphel31 (approved):
approving specifically for `.github/workflows/check-manifests.yml` changes

SPECS/openssh/openssh.spec
```diff
@@ -261,6 +261,9 @@ fi
 %{_mandir}/man8/ssh-sk-helper.8.gz
 
 %changelog
+* Fri Feb  2 15:55:52 EST 2024 Dan Streetman <ddstreet@ieee.org> - 9.5p1-2
```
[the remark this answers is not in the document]
ddstreetmicrosoft: I use rpmdev-bumpspec, I don't ever hand-edit changelog timestamps.

SPECS/zstd/zstd.spec
ddstreetmicrosoft: I'm not sure why our toolkit thinks all fedora spec files are licensed MIT, as that is definitely not correct; but it does *require* the line to match exactly that, so I just noted that the MIT license isn't necessarily the license actually in use (either for the spec file or the package).

toolkit/resources/manifests/package/pkggen_core_x86_64.txt
ddstreetmicrosoft: oh wow this lists packages to install entirely separately/individually and in order? I thought it was mostly just alphabetical. Why is each rpm installed separately instead of just passing them all to rpm (or dnf) and letting rpm work out the dependency details?

toolkit/resources/manifests/package/toolchain_x86_64.txt
ddstreetmicrosoft: wait the toolchain_*.txt packages are alphabetized but the pkggen_*.txt packages are individually installed in order? that is confusing on many levels...

ddstreetmicrosoft: ok i sorted the file correctly alphabetically, there was quite a bit out of order.

.github/workflows/check-manifests.yml
PawelWMS: Probably a good idea to add it there in addition to these two YAML files. The scripts ran with the `check-x86_64-manifests` target, however, currently use the regular `rpmspec` command, so I think a more thorough change would need to happen. I think a proper update is out of scope for this PR but I support the idea.

.github/workflows/check-manifests.yml
```diff
@@ -18,6 +18,10 @@ jobs:
...
+      run: |
+        [[ -n $(rpm --eval '%bcond test 1') ]] && echo '%bcond() %[ (%{2}) ? "%{expand:%%bcond_without %{1}}" : "%{expand:%%bcond_with %{1}}" ]' > ~/.rpmmacros
```
PawelWMS, line 23: Thanks for adding this workaround, @ddstreetmicrosoft! Could you only add a short comment explaining that this is because we run these checks on a Ubuntu image with RPM version < 4.18.0, where `%bcond` was introduced?

toolkit/resources/manifests/package/pkggen_core_x86_64.txt
ddstreetmicrosoft: ok sorted them in order or dependence. i'm still surprised that the user of this file requires that.

SPECS/iptables/iptables.spec
```diff
@@ -18,7 +18,8 @@ BuildRequires:  libmnl-devel
 BuildRequires:  libnftnl-devel
 BuildRequires:  systemd-bootstrap-rpm-macros
 Requires:       iana-etc
-Requires:       systemd
+# Our build tooling cannot handle this
+#Requires:       systemd
```
PawelWMS, line 22: This looks worrying. What is exactly the issue here? If `systemd` is a real run-time dependency of this package, then we must still keep it.

PawelWMS (commented):
Adding a first few comments and will continue reviewing.

.github/workflows/check-manifests.yml
ddstreetmicrosoft: to state my opinion more clearly and directly - the *problem* here is our PR runner, which IMHO is where we should put the fix/workaround. Since the problem is *not* in our toolkit code, we should not bloat/kludge our toolkit code.

SPECS/libmicrohttpd/libmicrohttpd.spec
```diff
@@ -1,21 +1,29 @@
...
+Distribution:   Microsoft Azure Linux
+Version:        0.9.77
+Release:        3%{?dist}
+Summary:        Lightweight library for embedding a webserver in applications
+
+# * COPYING says that some main sources are only under LGPL-2.1-or-later
+#   and the rest is dual licensed under LGPL-2.1-or-later OR GPL-2.0-or-later WITH eCos-exception-2.0.
+# * Some docs are under GFDL-1.3-no-invariants-or-later.
+# * Tests and some parts of the build system are under other licenses but they are NOT shipped.
+#License:        LGPL-2.1-or-later AND (LGPL-2.1-or-later OR GPL-2.0-or-later WITH eCos-exception-2.0) AND GFDL-1.3-no-invariants-or-later
```
PawelWMS, line 12: Can you say what was the issue with the tooling? We should log a bug for this.

SPECS/libmicrohttpd/libmicrohttpd.spec
```diff
@@ -1,21 +1,29 @@
...
-Version:        0.9.76
-Release:        1%{?dist}
-License:        LGPLv2+
 Vendor:         Microsoft Corporation
-Distribution:   Mariner
-URL:            https://www.gnu.org/software/libmicrohttpd/
+Distribution:   Microsoft Azure Linux
+Version:        0.9.77
+Release:        3%{?dist}
+Summary:        Lightweight library for embedding a webserver in applications
```
PawelWMS, line 6: Please keep the original order of the tags.

SPECS/systemd/systemd.spec
```diff
@@ -1,62 +1,342 @@
...
+%bcond zstd   1
+
+# Bootstrap may be needed to break circular dependencies with cryptsetup,
+# e.g. when re-building cryptsetup on a json-c SONAME-bump.
+%bcond bootstrap 0
+%bcond tests     1
+%bcond lto       1
+
+# Support for quick builds with rpmbuild --build-in-place.
+# See README.build-in-place.
```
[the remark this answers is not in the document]
ddstreetmicrosoft: readme was removed but i can add it back, though i am pretty sure our setup will not actually work with build-in-place without significant other changes. I left this in place to reduce delta with the fedora spec; i am assuming nobody reading this spec file wants to attempt build-in-place and so will likely just ignore this section, but it can be removed (at least for now, as i would very much like build-in-place working eventually). I didn't see its removal as important.

SPECS/libmicrohttpd/libmicrohttpd.spec
```diff
@@ -52,24 +60,19 @@ Doxygen documentation for libmicrohttpd and some example source code
...
 %check
-%ifnarch s390x
 %make_build check
-%endif
 
 %install
 %make_install
 
-find %{buildroot} -type f -name "*.la" -delete -print
+rm -f %{buildroot}%{_libdir}/*.la
```
PawelWMS, line 75: This is an outdated way of removing the `.la` files and we try to stick to the one from the previous version. It makes updates less prone to break during updates. Having said that, it's a minor detail, no strong pressure to revert.

SPECS/libxkbcommon/libxkbcommon.spec
```diff
@@ -1,30 +1,20 @@
...
-Summary:        X.Org X11 XKB parsing library
 Name:           libxkbcommon
-Version:        1.5.0
-Release:        1%{?dist}
-License:        MIT
+Version:        1.6.0
+Release:        2%{?dist}
+Summary:        X.Org X11 XKB parsing library
+License:        MIT AND X11 AND MIT-CMU
+URL:            http://www.x.org
```
PawelWMS, line 6: Similar ask here to keep the original order of tags.

.github/workflows/check-manifests.yml
ddstreetmicrosoft: ack

SPECS/libmicrohttpd/libmicrohttpd.spec
```diff
@@ -1,21 +1,29 @@
-Summary:        Lightweight library for embedding a webserver in applications
 Name:           libmicrohttpd
-Version:        0.9.76
-Release:        1%{?dist}
-License:        LGPLv2+
 Vendor:         Microsoft Corporation
-Distribution:   Mariner
-URL:            https://www.gnu.org/software/libmicrohttpd/
+Distribution:   Microsoft Azure Linux
```
PawelWMS, line 3: Let's use `Azure Linux` instead. There's [a bulk update](https://github.com/microsoft/CBL-Mariner/pull/7606/) underway to modify all specs.
```suggestion
Distribution:   Azure Linux
```

SPECS/iptables/iptables.spec
ddstreetmicrosoft: this is because of our "circular dependencies". I actually don't understand myself why our toolkit fails around runtime dependencies, as it's simply *building* the package, but it does indeed fail when attempting to build package A if there is a "circular dependency" - even *runtime* dependency - on package B. I'm open to alternate suggestions, although I do plan to look into a more comprehensive actual fix separately from this pr.

ddstreetmicrosoft: also, i'm not aware of *any* situation where our distro runs without systemd installed, are you?

SPECS/libmicrohttpd/libmicrohttpd.spec
ddstreetmicrosoft: this is the original order of tags in the spec file we inherit from.

SPECS/libmicrohttpd/libmicrohttpd.spec
ddstreetmicrosoft: this is from our inherited spec file and it's better for us to use this format (i.e. the result is identical) so we reduce future delta.

.github/workflows/check-manifests.yml
PawelWMS: I agree that we shouldn't modify the toolkit to address issues limited to the PR runner. The story behind the `mariner_rpmspec` function from that script is just related to the issue we have with the runners. The function is a workaround to allow querying our spec files with a similar RPM configuration as what is happening inside the Mariner build chroot. But without having to build or use it (as it may not always be feasible). The functions from `rpmops.sh` are meant to be ran under Ubuntu, so there's a similar issue of an outdated RPM. At least on the latest LTS version, which is the officially supported one for the time being.

SPECS/libxkbcommon/libxkbcommon.spec
ddstreetmicrosoft: same response; this is exactly the order of the inherited spec file.

SPECS/libmicrohttpd/libmicrohttpd.spec
ddstreetmicrosoft: ack

.github/workflows/check-manifests.yml
ddstreetmicrosoft: i don't really touch the tooling at all, so you guys feel free to update it as you like, obviously; but this won't be the last new rpm macro and it certainly wouldn't be what i would do if i had to maintain the toolkit.

SPECS/libxkbcommon/libxkbcommon.spec
```diff
@@ -1,30 +1,20 @@
...
-BuildRequires:  git
-BuildRequires:  libX11-devel
-BuildRequires:  libxml2-devel
-BuildRequires:  meson
-BuildRequires:  pkg-config
+Source0:        https://xkbcommon.org/download/%{name}-%{version}.tar.xz
+BuildRequires:  gcc
+BuildRequires:  git meson
+BuildRequires:  byacc flex bison
+BuildRequires:  xorg-x11-proto-devel libX11-devel
```
PawelWMS, line 14: Please use one dependency per line. Ideally also in alphabetical order but that's a P2.

.github/workflows/check-manifests.yml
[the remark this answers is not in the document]
PawelWMS: +1 here. This PR has gotten quite bulky.

SPECS/openssh/openssh.spec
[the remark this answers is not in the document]
PawelWMS: +1 to @anphel31's ask. Everyone uses slightly different tooling but we try to maintain the same format whenever possible.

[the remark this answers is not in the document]
PawelWMS:
```suggestion
* Fri Feb 02 15:55:52 EST 2024 Dan Streetman <ddstreet@ieee.org> - 9.5p1-2
```

SPECS/polkit/polkit.spec
```diff
@@ -15,12 +15,13 @@ BuildRequires:     gobject-introspection-devel
...
 BuildRequires:     pam-devel
-BuildRequires:     systemd-devel
+BuildRequires:     systemd-bootstrap-devel
 Requires:          duktape
 Requires:          expat
 Requires:          glib
 Requires:          pam
-Requires:          systemd
+# Our build tooling cannot handle this
+#Requires:          systemd
```
PawelWMS, line 24: Same question here as in `iptables.spec`.

SPECS/libmicrohttpd/libmicrohttpd.spec
ddstreetmicrosoft: i dont actually remember exactly (there have been quite a lot of toolkit/prcheck failures i had to work around); i updated the code to go back to the complex license and pushed it, so if it passes pr checks and test runs then i guess it's ok :-)

SPECS/polkit/polkit.spec
ddstreetmicrosoft: same circular *runtime* dep problem. I've no idea why the toolkit fails due to *runtime* "circular" deps when building a package.

SPECS/libxkbcommon/libxkbcommon.spec
ddstreetmicrosoft: no.

this is how our inherited spec file lists it; it is counterproductive to change that for zero technical reason.

SPECS/zstd/zstd.spec
```diff
@@ -1,96 +1,323 @@
...
+%{_bindir}/%{name}less
+%{_mandir}/man1/%{name}.1*
+%{_mandir}/man1/un%{name}.1*
+%{_mandir}/man1/%{name}cat.1*
+%{_mandir}/man1/%{name}grep.1*
+%{_mandir}/man1/%{name}less.1*
+%license COPYING LICENSE
 
-%ldconfig_scriptlets
+%files -n lib%{name}
```
PawelWMS, line 137: Can we stick to the original `zstd-libs` name of the package? It causes unnecessary re-names in other spec but more importantly this may be breaking for people already using the original version. Alternatively, please use the `Provides` tag.

SPECS/python-zstd/python-zstd.spec
```diff
@@ -0,0 +1,108 @@
...
+%doc README.rst
+%{python3_sitearch}/%{pypi_name}-%{version}-py%{python3_version}.egg-info
+%{python3_sitearch}/%{pypi_name}*.so
+
+%changelog
+* Wed Jan 31 13:55:52 EST 2024 Dan Streetman <ddstreet@ieee.org> - 1.5.5.1-2
+- Update to version from Fedora 39.
+- Next line is present only to avoid tooling failures, and does not indicate the actual package license.
+- Initial CBL-Mariner import from Fedora 39 (license: MIT).
+- license verified
```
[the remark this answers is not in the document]
ddstreetmicrosoft: are you seriously asking me to capitalize the changelog statement?

SPECS/zstd/zstd.spec
```diff
@@ -1,96 +1,323 @@
...
 %description
-The zstd package contains programs for compressing and decompressing files
+Zstd, short for Zstandard, is a fast lossless compression algorithm,
+targeting real-time compression scenarios at zlib-level compression ratio.
 
-%package        devel
-Summary:        Header and development files for zstd
-Requires:       %{name} = %{version}-%{release}
-Provides:       lib%{name}-devel = %{version}-%{release}
+%package -n lib%{name}
```
PawelWMS, line 53: Why was the re-name needed? With the `Provides` below it seems harmless but seems unnecessary.

christopherco (approved):
My previous concerns were addressed and the current changes LGTM

SPECS/zstd/zstd.spec
ddstreetmicrosoft: reduce delta from inherited spec file

SPECS/zstd/zstd.spec
```diff
@@ -1,96 +1,323 @@
...
+
+%description -n lib%{name}-devel
+Header files for Zstd library.
+
+%description -n lib%{name}-static
+Static variant of the Zstd library.
 
 %prep
-%autosetup
+%setup -q
```
PawelWMS, line 82: Why the revert? We try to stick to the new `%autosetup` wherever possible.

SPECS/zstd/zstd.spec
```diff
@@ -1,96 +1,323 @@
...
-Requires:       %{name} = %{version}-%{release}
+# For backwards compatibility
+Provides:       zstd-devel = %{version}-%{release}
+Obsoletes:      zstd-devel <= 1.5.5-5
 
-%description    doc
-Documentation files for zstd
+%package -n lib%{name}-static
+Summary:        Static variant of the Zstd library
+Requires:       lib%{name}-devel = %{version}-%{release}
```
PawelWMS, line 73: Are the static libraries required for anything at the moment? If not, we try to avoid them wherever possible to not encourage people to use them in other specs (they make package updates more tricky). In this case we also bloat the toolchain.

SPECS/zstd/zstd.spec
```diff
@@ -1,96 +1,323 @@
...
+%check
+execstack lib/libzstd.so.1
+
+export CFLAGS="$RPM_OPT_FLAGS"
+export LDFLAGS="$RPM_LD_FLAGS"
+make -C tests test-zstd
+%if %{with pzstd} && %{with gtest}
+export CXXFLAGS="$RPM_OPT_FLAGS"
+make -C contrib/pzstd test
+%endif
```
PawelWMS, line 109: Thanks for expanding the tests. This section may need a bit of rework, so we don't accidentally skip over failing tests.

More context:

The way a test pass/failure is currently figured out is from the `====== CHECK DONE rubygem-bindata. EXIT STATUS X` line RPM puts in the logs. The `X` is determined from the exit code of the last run command, so with mult-command `%check` sections we need to keep track of the exit code of at least the commands, which run the tests. One example would be inside `edk2.spec` where we set `tests_ok=true` at the beginning of the `%check` section and then update it to `false` whenever any of the tests fail. Then we make sure running `$tests_ok` is the last command of the `%check` section to capture the test failure if any of the subtests failed.

There are other ways to do it as well, but the general ask is make sure the `%check` section returns with a non-zero exit code when at least one of its tests fail.

toolkit/resources/manifests/package/pkggen_core_x86_64.txt
PawelWMS: > oh wow this lists packages to install entirely separately/individually and in order? I thought it was mostly just alphabetical. Why is each rpm installed separately instead of just passing them all to rpm (or dnf) and letting rpm work out the dependency details?

I share your concern here (and I'm also personally annoyed by this in addition to being concerned). This is one of the things we want to fix as part of either completely dropping the manifest files or at least making them "versionless"/easier to use. We have that work item scheduled for later.

SPECS/systemd/systemd.spec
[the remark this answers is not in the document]
PawelWMS: I'd suggest removing this particular macro completely to just make the spec a bit easier to read, since, like you noticed, we may not have much use of it in predictable future and it doesn't ease enabling/disabling specific features like the macros set with `%bcond`s above. Not a huge concern for me right now, though.

SPECS/systemd/systemd.spec
PawelWMS: I'm not sure this is our current goal or approach right now. Having said that, I believe you mentioned that the subpackage splitting you've done here allowed to reduce the size of some of our images, so the change is definitely valuable. Please only mention that in the PR's summary to make it easier in the future to understand the reasoning which lead us to this update.

SPECS/systemd/systemd.spec
PawelWMS: If it's not used, we try to clean-up the specs of any bits which only make sense for other distros (to the best of our abilities and time constraints), so that it's easier to understand the spec. If we're not using these bits, it would help to remove them.

SPECS/systemd/systemd.spec
PawelWMS: Could you elaborate? We don't need to keep other distro-specific bits in our specs.

SPECS/systemd/systemd.spec
```diff
@@ -1,62 +1,345 @@
...
+Source23:       sysusers.prov
+Source24:       sysusers.generate-pre.sh
+
+Source25:       98-default-mac-none.link
+
+%if 0
+GIT_DIR=../../src/systemd/.git git format-patch-ab --no-signature -M -N v235..v235-stable
+i=1; for j in 00*patch; do printf "Patch%04d:      %s\n" $i $j; i=$((i+1));done|xclip
+GIT_DIR=../../src/systemd/.git git diffab -M v233..master@{2017-06-15} -- hwdb/[67]* hwdb/parse_hwdb.py >hwdb.patch
+%endif
```
PawelWMS, line 103: If it's not used, let's remove it. Or please add a comment explaining why we need to keep it.
```suggestion
```

SPECS/systemd/systemd.spec
```diff
@@ -1,62 +1,345 @@
...
+BuildRequires:  perl
+BuildRequires:  perl(IPC::SysV)
+
+%ifnarch %ix86
+# bpftool is not built for i368
+BuildRequires:  bpftool
+%global have_bpf 1
+%endif
+
+%if %{undefined azure}
```
PawelWMS, line 234: I think we can remove all sections with `%{undefined azure}`, since they are never going to be used in our specs, create noise, and bloat the spec. Keeping the spec as complex as this one clean will make it easier for future devs to understand it and make updates.

PawelWMS (changes requested):
A few things to figure out before we merge:
- The removed run-time dependency in `iptables` and `polkit`. If it's needed for that package, we should figure out what's wrong and keep it. If it's not needed, we can just remove the `Requires` without a comment.
- Re-ordering the top spec tags like `Summary`. `Name`, `Version`, etc. to stay consistant with the rest of the specs.
- Using `Azure Linux` for the `Distribution` tag.
- Specs clean-up: one dependency per line.
- Re-name of `zstd-libs` to `libzstd` without providing backward compatibility.
- Do we need the new subpackage with static libs for `zstd`? They currently seem unused and bloat the toolchain.
- Tweak the `%check` section in `zstd.spec`.
- Still working on `systemd.spec` - it would help to remove any sections of the spec, that are only relevant in other distros to help with the review and future maintenance.

SPECS/zstd/zstd.spec
ddstreetmicrosoft: I did not expand the tests; this is directly from our inherited spec file. 

I do not follow what you are talking about in the %check section; if a test fails, it will return nonzero and fail the rpm build. There is no need to keep track of individual tests because the build is immediately failed on the first failure. There is nothing unusual with this %check section which, again, is directly from our inherited spec file and, as far as i can tell, is perfectly fine.

SPECS/systemd/systemd.spec
ddstreetmicrosoft: reducing delta from our inherited spec file is far more important that removing this just for the sake of its removal without any actual technical change.

SPECS/systemd/systemd.spec
ddstreetmicrosoft: i understand that would be your preference, but i disagree with you, and the change would not actually result in any real change to the built package.

ddstreetmicrosoft:
> A few things to figure out before we merge:
> 
> * The removed run-time dependency in `iptables` and `polkit`. If it's needed for that package, we should figure out what's wrong and keep it. If it's not needed, we can just remove the `Requires` without a comment.
> * Re-ordering the top spec tags like `Summary`. `Name`, `Version`, etc. to stay consistant with the rest of the specs.
> * Using `Azure Linux` for the `Distribution` tag.
> * Specs clean-up: one dependency per line.
> * Do we need the new subpackage with static libs for `zstd`? They currently seem unused and bloat the toolchain.
> * Tweak the `%check` section in `zstd.spec`.
> * Still working on `systemd.spec` - it would help to remove any sections of the spec, that are only relevant in other distros to help with the review and future maintenance.

i responded to all comments above, and much of this is simple disagreements between us on non-technical and non-functional aspects, which I hope aren't essential to you.

SPECS/zstd/zstd.spec
ddstreetmicrosoft: reduce delta from inherited spec

SPECS/zstd/zstd.spec
ddstreetmicrosoft: ack

SPECS/systemd/systemd.spec
ddstreetmicrosoft: please take a look at the entire patch series

SPECS/git/git.spec
```diff
@@ -1,7 +1,13 @@
+
+%bcond daemon 1
+%bcond subtree 1
+%bcond svn 0
```
PawelWMS, line 4: Why are we dropping the `svn` subpackage in this PR? I doubt anyone's using it, but seems unrelated to updating `systemd`.

SPECS/ipset/ipset.spec
```diff
@@ -34,12 +34,13 @@ Libraries and header files for ipset.
...
+BuildRequires:  systemd-bootstrap-rpm-macros
 Requires:       %{name} = %{version}-%{release}
 Requires:       iptables-services
-Requires(post): systemd
-Requires(postun): systemd
-Requires(preun): systemd
+# Our build tooling cannot handle this
+#Requires(post): systemd
+#Requires(postun): systemd
+#Requires(preun): systemd
```
PawelWMS, line 43: What error were you getting with these dependencies enabled? With the `BuildRequires` change above, I think there should be no issues with circular dependencies for this spec.

Having said, that I think I am leaning towards treating such `systemd` run-time dependencies as a special case. Like you noticed, I also can't image a running VM image without `systemd`. Containers on the other hand and our build environment can't make use of `systemd` services and this is what these run-time dependencies are about. In other words, dropping a dependency on `systemd` for cases where it's only there because a package wants to run a service should not have any negative side-effects but would simplify building our packages.

Anyway, I'm fine with commenting out these dependencies for this PR. We may need to think this through, but I'm curious, if you agree with this assessment?

SPECS/python-zstd/python-zstd.spec
```diff
@@ -0,0 +1,109 @@
...
+%{__python3} setup.py test
+
+%files -n python3-%{pypi_name}
+%license LICENSE
+%doc README.rst
+%{python3_sitearch}/%{pypi_name}-%{version}-py%{python3_version}.egg-info
+%{python3_sitearch}/%{pypi_name}*.so
+
+%changelog
+* Wed Jan 31 13:55:52 EST 2024 Dan Streetman <ddstreet@ieee.org> - 1.5.5.1-2
```
PawelWMS, line 66: Please let's keep the changelog format consistent with the rest of the spec files.
```suggestion
* Wed Jan 31 2024 Dan Streetman <ddstreet@ieee.org> - 1.5.5.1-2
```

SPECS/zstd/zstd.spec
PawelWMS: Failing `%check` sections don't fail a package build as long as all other sections ran successfully. A shorter version of what I had in mind is: package test's status is determined by the exit code of the last command.

PawelWMS: Wait, nevermind, I now saw the latest change. Sorry for the noise.

PawelWMS:
> > A few things to figure out before we merge:
> > 
> > * The removed run-time dependency in `iptables` and `polkit`. If it's needed for that package, we should figure out what's wrong and keep it. If it's not needed, we can just remove the `Requires` without a comment.
> 
> as stated above this is due to the "circular dependencies" resulting from the rather unique design of our toolkit, which other distros don't need to worry about. You can ignore this for now as all our installs will have systemd.
> 
> > * Re-ordering the top spec tags like `Summary`. `Name`, `Version`, etc. to stay consistant with the rest of the specs.
> 
> I'm going do decline any non-technical stuff like this, as I mentioned above that reducing future delta is far more important than the order of tags.
> 
> > * Using `Azure Linux` for the `Distribution` tag.
> 
> sure, but you should not forget to change the `mariner-release` package since that's what it uses.
> 
> > * Specs clean-up: one dependency per line.
> 
> Another point where I will politely decline for the same reason as before.
> 
> > * Do we need the new subpackage with static libs for `zstd`? They currently seem unused and bloat the toolchain.
> 
> I ripped out all the upgrades to zstd as you did not seem to like the changes and I _think_ python-zstd can work with our current zstd packaging, so this point is no longer relevant.
> 
> > * Tweak the `%check` section in `zstd.spec`.
> 
> Not relevant anymore.
> 
> > * Still working on `systemd.spec` - it would help to remove any sections of the spec, that are only relevant in other distros to help with the review and future maintenance.
> 
> Another point I will politely decline, thanks.
> 
> This is still running through a full build and I probably will have to go a few more rounds to "fix" all the "circular dependencies", but other than that it should be good to review again, @PawelWMS Pawel Winogrodzki FTE it's just waiting on your approval.

Thanks for addressing my comments!

From this list I think there's only one thing I'd still want to get back to, which is spec clean-up. Over the years we tried to stick to specific ways of writing the specs (some encoded in our [internal guidelines](https://dev.azure.com/mariner-org/mariner/_wiki/wikis/mariner.wiki/125/Introduction-to-SPEC-Files)). We also included specs from multiple distributions or written our own and so there's no one specific "upstream" to talk about. Not to mention that we never had any specific reason to assume package upgrades will be done by constantly relying on any specific distribution performing the upgrade ahead of us, so that is why I would prefer a clean-up. Keeping a spec as unchanged as possible from its source would in some cases be a regression in our specs at this point.

Having said that, I think we would benefit from more automation performing the clean-up whenever we decide to take a spec from another distribution, so that devs don't have to worry about it and still have the benefit of the new spec being similar to the old (assuming the source of the new spec is the same distro), which you mentioned. But until we have that, I don't want to revert the work we done so far.

SPECS/git/git.spec
ddstreetmicrosoft: "circular dependencies" problems.

SPECS/ipset/ipset.spec
ddstreetmicrosoft: I am not an expert in our tooling, so I can't answer specifically why our tooling fails due to this, but it does.

> Containers on the other hand and our build environment can't make use of systemd services and this is what these run-time dependencies are about.

Er, I don't think that's actually correct, but in any case it seems we're in agreement that we can comment these out, at least for now.

PR #7316, opened 2024-01-18
glibc locale en_US.UTF-8 should be en_US.utf8

ddstreetmicrosoft:
glibc ignores case and removes '-' from locale requests so it will reduce a request for en_US.UTF-8 down to en_US.utf8, but it will not convert a request for en_US.utf8 to en_US.UTF-8. So the actual locale must be en_US.utf8, not en_US.UTF-8.


SPECS/glibc/glibc.spec
```diff
@@ -323,6 +322,9 @@ grep "^FAIL: nptl/tst-eintr1" tests.sum >/dev/null && n=$((n+1)) ||:
 %defattr(-,root,root)
 
 %changelog
+* Wed Jan 17 13:56:40 EST 2024 Dan Streetman <ddstreet@ieee.org> - 2.38-2
```
anphel31, line 325: can you remove "13:56:40 EST"; most of our changelogs do not have these details

anphel31 (commented):
I ran a full build to help validate this change. https://dev.azure.com/mariner-org/mariner/_build/results?buildId=486909&view=results

SPECS/glibc/glibc.spec
ddstreetmicrosoft: ack

ddstreetmicrosoft:
The cgmanifest pr check is complaining that the busybox tarball can't be downloaded, but that's just because the upstream busybox website appears to be offline and/or very slow right now.
The url is https://www.busybox.net/downloads/busybox-1.36.1.tar.bz2

anphel31 (approved):

ddstreetmicrosoft:
started a new full build https://dev.azure.com/mariner-org/mariner/_build/results?buildId=515189&view=results

PR #7318, opened 2024-01-18
Allow custom %clean section

ddstreetmicrosoft:
###### Summary 

Allows custom %clean section, for 3.0 and later.

###### Change Log  

Remove 3.0 branch from the check-clean-stage github action, so 3.0 (and later) can once again use custom %clean sections.

NO

### DIFF
--- a/.github/workflows/check-clean-stage.yml
+++ b/.github/workflows/check-clean-stage.yml
@@ -3,11 +3,12 @@
 
 name: Spec %clean stage check
 
+# NOTE - allow custom %clean section for 3.0 and later (remove this entire check after 2.0 is EOL)
 on:
   push:
-    branches: [main, dev, 1.0*, 2.0*, 3.0*, fasttrack/*]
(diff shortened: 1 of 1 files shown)


ddstreetmicrosoft:
I have no idea if this will actually affect the github action, maybe this needs to be checked into `main` or `2.0` to actually take effect for PRs

anphel31 (approved):
Why do we need custom %clean sections in 3.0? Should they be discouraged? If so, perhaps this check should just generate a warning.

Since this was never set for the `3.0-dev` branch I guess it's never been used for 3.0 so far

ddstreetmicrosoft:
> Why do we need custom %clean sections in 3.0? Should they be discouraged? If so, perhaps this check should just generate a warning.
> 
> Since this was never set for the `3.0-dev` branch I guess it's never been used for 3.0 so far

custom clean sections are used to clean up stuff that the build generates that isn't otherwise automatically cleaned up. There is no good reason I know of for *not* allowing specs to do this.

PR #7319, opened 2024-01-18
Upgrade qemu to v8.2.0

kanikanema:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
Upgrade to v8.2.0

###### Change Log  
Use the fedora spec for v8.2.0-rc2 as reference
Disable packages that were disabled in CBLM 2.0, most notably for ppc/sparc, brltty, libssh, pulseaudio, sdl
Disable all UI related packages to minimize dependencies 
Disable manpages to remove python-sphinx and python-sphinx-rtd dependency
Disabld qemu-keymap to remove xkbcommon dependency
Disabld new audio drivers like pipewire
Disable rbd package to drop dependency on librbd
All enable/disable config values are based on `azl` global variable or derived from it. This is to make the azlinux specific changes obvious.

NOTE: installation will fail due to a runtime dependency on virtiofsd - PR in-progress 
**NO**

###### Test Methodology
local build:
image


SPECS/qemu/qemu.spec
```diff
@@ -1,64 +1,257 @@
-# Disable stripping - strip is trying to strip binaries (.img, .elf, etc.) built for different architectures.
+Vendor:         Microsoft Corporation
+Distribution:   Azure Linux
+# Provide a way to skip tests via rpmbuild `--without`
+# This makes it easier to skip tests in copr repos, where
+# the qemu test suite is historically flakey
+%bcond_without check
+
+# This spec is for AzLinux
```
bfjelds, line 8: i agree with what you're doing here with the macros (and preserving the other distro macros), but i dislike having our specs follow a different set of rules than all of the other specs.  i've been following the mariner recommendations:
* removing all distro macros (fedora, rhel, suse, azl)
* alphabetizing and grouping the buildrequires, requires, etc

SPECS/qemu/qemu.spec
```diff
@@ -1,64 +1,257 @@
...
+%endif
+
+%global have_kvm 0
+%if 0%{?kvm_package:1}
 %global have_kvm 1
+%endif
+
+# Matches numactl ExcludeArch
 %global have_numactl 1
+%ifarch %{arm}
```
bfjelds, line 85: it looks like previously there was no conditional for arm and have_numactl.  are we changing how arm works here?

SPECS/qemu/qemu.spec
```diff
@@ -1,64 +1,257 @@
...
+%if 0%{?azl}
+%global have_liburing 1
+%endif
+%if 0%{?fedora}
+%ifnarch %{arm}
 %global have_liburing 1
+%endif
+%endif
+
+# 2.0 has this set
```
bfjelds, line 119: why are we changing from 2.0?

SPECS/qemu/qemu.spec
```diff
@@ -1,64 +1,257 @@
...
+    %global modprobe_kvm_conf %{_sourcedir}/kvm-s390x.conf
+%endif
+%ifarch %{ix86} x86_64
     %global modprobe_kvm_conf %{_sourcedir}/kvm-x86.conf
 %endif
+
 %global tools_only 0
+
+
+%global user_static 1
```
reubeno, line 69: Random comment -- I know it's not turning this feature on, but I'm glad to see having at least latent support in the spec for user_static. This is one of the features required for Azure Linux to support running cross-architecture Docker containers (e.g., running an x86_64 container on aarch64, or vice versa).

SPECS/qemu/qemu.spec
```diff
@@ -1,64 +1,257 @@
...
+%if 0%{?azl}
+%global have_edk2 0
+%endif
+
 # All modules should be listed here.
 %define have_block_rbd 1
+%ifarch %{ix86} %{arm}
+%define have_block_rbd 0
+%endif
+# disable for azl till dependency is available
```
bfjelds, line 183: what depenedency is this waiting on?  is this mergable without the dep?  it seems like a change from 2.0 to have this set to 0.

SPECS/qemu/qemu.spec
```diff
@@ -1,64 +1,257 @@
...
+%define have_block_nfs 0
+%if 0%{?azl}
 %define have_block_nfs 1
+%endif
+%if 0%{?fedora}
+%define have_block_nfs 1
+%endif
+
 %define have_librdma 1
+%ifarch %{arm}
```
bfjelds, line 206: 2.0 didn't set this to 0 for arm, why the change?

SPECS/qemu/qemu.spec
```diff
@@ -1,64 +1,257 @@
...
+%define have_librdma 0
+%endif
+
 %define have_libcacard 1
+%if 0%{?azl}
+%define have_libcacard 0
```
bfjelds, line 212: 2.0 set have_libacard  to 1, why 0 now?

SPECS/qemu/qemu.spec
kanikanema: Yes it is but without support of ceph-rbd. This will be deviation from 2.0. From what I know ceph-rbd is not used by baremetal as of now. Once librbd is available we can enable it back. I will file ADO work items for disabled packages so that we can revisit them as more pkgs are made available and/or we need a disabled package. Does that make sense?
My aim is to get this merged and then handle additional subpackages separately. Waiting for every subpackage to be available is very long process. Atleast if the pkg is merged we can start getting some testing done with it.

SPECS/qemu/qemu.spec
kanikanema: I agree that the cleanup is needed since that is the current recommended strategy. I will do it towards the end of the review. Ideally, I would like to do it as a separate commit so the git history has the unmodified (fedora-like) version as well.

SPECS/qemu/qemu.spec
kanikanema: Good catch. For azl this should not be required. I am still to test the arm build. Have removed the condition and triggered the build.

SPECS/qemu/qemu.spec
kanikanema: Do reduce the package dependencies. This is related to  enabling support in qemu for virgl - a virtual 3D capable graphics card for virtual machines running inside QEMU.

SPECS/qemu/qemu.spec
kanikanema: This adds support for 'emulation of smart cards to a virtual card reader running in a guest virtual machine'. Do we really need it? I have intentionally turned of some subpackages that I consider not useful to us in the near future. We can easily turn on the support when required.

SPECS/qemu/qemu.spec
bfjelds: i'm not sure i understand exactly.  if librbd existed in 2.0, it should be available as a dependency for 3.0 ... unless qemu has a min version of librdb that mariner2.0 doesn't satisfy?

SPECS/qemu/qemu.spec
bfjelds: i have no idea if we need it, but removing it is a breaking change we should make purposefully.  i suspect we are fine to remove, but i'm not sure who knows best here.

SPECS/qemu/qemu.spec
bfjelds: i have no idea if we need it, but removing it is a breaking change we should make purposefully.

SPECS/qemu/qemu.spec
kanikanema: Ah, sorry for the confusion. I was waiting for the updated librbd package to be available in 3.0 (I believe that is WIP). So till that package is available, I have configured qemu to build with --disable-librbd.

SPECS/qemu/qemu.spec
kanikanema: I believe BMP is the only real user of qemu on mariner so far. We should be able to make a call on it.

SPECS/qemu/qemu.spec
kanikanema: Removed all references to fedora and rhel. For the BR the main set has been put in alphabetic order. There are a few misc conditional BRs that are separate. It is similar to how 2.0 is layout.

SPECS/qemu/qemu.spec
kanikanema: I agree. But I doubt it will break any real usecase immediately. We can work on enabling it post 3/1.

SPECS/qemu/qemu.spec
```diff
@@ -1,64 +1,214 @@
...
+
+%global have_liburing 0
+%if 0%{?azl}
 %global have_liburing 1
+%endif
+
 %global have_virgl 1
+# 2.0 has this set, disabled till dep is available in 3.0
+%if 0%{?azl}
+%global have_virgl 0
```
bfjelds, line 102: add a comment why this isn't needed for merge

SPECS/qemu/qemu.spec
bfjelds: i'm still not following.  librbd exists in the 3.0 spec tree, it just isn't being built.  you should still be able to build it, right?  so you should be able to leverage that 2.0 spec code here?

SPECS/qemu/qemu.spec
bfjelds: i think anyone can create a mariner VM in azure, right?  there is a marketplace image.  if we make a breaking change to qemu, i think we should consider more than just BMP scenario.

bfjelds: talked to britel, you are totally right.  if we get subsequent requests to reenable, we can revisit.

SPECS/qemu/qemu.spec
bfjelds: if rbd is not used by BMP, can disregard this comment from me.

bfjelds (approved):

PR #7211, opened 2024-01-10
fix: upgrade cloud-init to v23.4.1 and add patch to retain exit code for recoverable errors

christopherco:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 

Reverts the revert of cloud-init v23.4.1 and adds a patch backport to gracefully handle the new status exit code behavior which caused a regression in our extended testing apparatus, prompting the initial revert.

From https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2048522

`cloud-init status` introduced a new exit code of 2 with the meaning of recoverable errors / warnings.
Without this fix, status exited with 2 in cases where previously did it with 0, potentially breaking consumers (scripts or other programs) of `cloud-init status`.

The fix is to include a quilt patch retaining the exit code of 0 for recoverable errors / warnings for stable releases.

Fixes: 1a57d91875d208808582f483361b866fe0310453 ("Revert "fix: upgrade cloud-init to v23.4.1"")

Signed-off-by: Chris Co <chrco@microsoft.com>

###### Does this affect the toolchain?  

**NO**

###### Associated issues  

[...]

PR #7330, opened 2024-01-18
Fixed `cloud-init` tests.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Recently ptests for `cloud-init` started to fail. The culprit seems to be an update to `jsonschema` - one of the test dependencies. I'm switching to using our `python3-jsonschema` package.

###### Change Log  

- Replaced upstream `jsonschema` module with ours.
- Fixed the source URL.

No.

###### Test Methodology

- Local test run.
- [Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=486690&view=results).

### DIFF
--- a/SPECS/cloud-init/cloud-init.spec
+++ b/SPECS/cloud-init/cloud-init.spec
@@ -1,20 +1,23 @@
+%define full_version 23.3.3
+%define short_version %(echo %{full_version} | cut -d. -f1-2)
+
 Summary:        Cloud instance init scripts
 Name:           cloud-init
-Version:        23.3
-Release:        1%{?dist}
+Version:        %{short_version}
+Release:        2%{?dist}
 License:        GPLv3
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 Group:          System Environment/Base
 URL:            https://launchpad.net/cloud-init
-Source0:        https://launchpad.net/cloud-init/trunk/%{version}/+download/%{name}-%{version}.tar.gz
+Source0:        https://launchpad.net/cloud-init/trunk/%{full_version}/+download/%{name}-%{version}.tar.gz
 Source1:        10-azure-kvp.cfg
 Patch0:         overrideDatasourceDetection.patch
 %define cl_services cloud-config.service cloud-config.target cloud-final.service cloud-init.service cloud-init.target cloud-init-local.service
 BuildRequires:  automake
 BuildRequires:  dbus
 BuildRequires:  iproute
-BuildRequires:  mariner-release 
+BuildRequires:  mariner-release
 BuildRequires:  python3
 BuildRequires:  python3-PyYAML
 BuildRequires:  python3-certifi
@@ -107,6 +110,8 @@ echo -e 'line1\nline2\nline3\ncloud-init-ca-certs.crt\n' > "${conf_file}"
 %define test_pkgs pytest-metadata unittest2 mock attrs iniconfig netifaces pyserial
 
 pip3 install --upgrade %{test_pkgs}
+# Higher versions of 'jsonschema' break the tests.
+sed -E -i 's/jsonschema/jsonschema==4.20.0/' test-requirements.txt
 pip3 install -r test-requirements.txt
 
 make check %{?_smp_mflags}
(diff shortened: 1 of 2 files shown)


ddstreetmicrosoft:
oh man, uh, could we not do this but instead fix it so pip doesn't pull it at all please? e.g. #7328 

freezing the version still results in pip getting it which really we shouldn't be doing at all.

but if you like this way better it will work too i suppose.

PawelWMS:
> oh man, uh, could we not do this but instead fix it so pip doesn't pull it at all please? e.g. #7328
> 
> freezing the version still results in pip getting it which really we shouldn't be doing at all.
> 
> but if you like this way better it will work too i suppose.

I'd also prefer to use our own packages wherever possible, but we're way behind with `jsonschema`, so I'd prefer to avoid introducing other breaking changes and spending more time on this than necessary. This is only meant to be a quick fix for a failing test. Feel free to continue on with your work in the other PR, however.

SPECS/cloud-init/cloud-init.spec
```diff
@@ -1,20 +1,23 @@
+%define full_version 23.3.3
+%define short_version %(echo %{full_version} | cut -d. -f1-2)
+
 Summary:        Cloud instance init scripts
 Name:           cloud-init
-Version:        23.3
-Release:        1%{?dist}
+Version:        %{short_version}
```
christopherco, line 6: 23.3 and 23.3.3 are two separate versions. We should keep to `23.3` for this change.

christopherco (changes requested):

SPECS/cloud-init/cloud-init.spec
```diff
@@ -107,6 +110,8 @@ echo -e 'line1\nline2\nline3\ncloud-init-ca-certs.crt\n' > "${conf_file}"
 %define test_pkgs pytest-metadata unittest2 mock attrs iniconfig netifaces pyserial
 
 pip3 install --upgrade %{test_pkgs}
+# Higher versions of 'jsonschema' break the tests.
+sed -E -i 's/jsonschema/jsonschema==4.20.0/' test-requirements.txt
```
christopherco, line 114: Not supportive of this workaround. Why can't we use our distribution's jsonschema for this? Would using our distro's jsonschema incur even more changes beyond just adding a BR?

PawelWMS: It would, yes. We're on version 2.6.0, while the tests have been always running on 4.xx. I don't want to rock the boat with this change, just bring it back to the previous state where the tests were stable.

SPECS/cloud-init/cloud-init.spec
PawelWMS: I'm not changing the version, we're still on 23.3. This is just making sure the source URL is correct.

SPECS/cloud-init/cloud-init.spec
ddstreetmicrosoft: FWIW my PR that uses our existing packaged python3-jsonschema passes tests just fine.

PawelWMS: Ah, that's great! Please check that in and I'll close this PR in a minute. Thanks, Dan!

PawelWMS: Wait, I see you closed it. I'll just move that change here, so we fix the source URL in one go.

SPECS/cloud-init/cloud-init.spec
christopherco: Oh....I see what you mean now... After comparing the tarball hashes, the real version we build is actually `23.3.3` but our spec version here is actually incorrect. Thoughts of fixing the SPEC file's version number here to be `23.3.3`, or prefer to wait for the next cloud-init version bump which is in the works already (#7211)

christopherco (approved):
Approving since I am generally good with this change. @PawelWMS Please do look at the comment on `23.3` vs `23.3.3` versioning in the SPEC itself.

anphel31 (approved):

SPECS/cloud-init/cloud-init.spec
PawelWMS: The 23.3 is the correct version. The URL path from the upstream is just weird. From [upstream](https://launchpad.net/cloud-init/trunk/23.3.3):
image

PR #7350, opened 2024-01-19
Fix SELinux labeling of special dirs

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Building a Mariner image occurs within a chroot environment. To make the chroot environment more useful, a bunch of special directories (e.g. /dev, /proc) are mounted into the OS image. However, this causes problems when applying the SELinux labels to the filesystem since mount-point directories are hidden by the mounts and therefore don't receive their SELinux labels.

This change fixes this problem by using bind mounts to get an unobstructed view of each filesystem.

###### Change Log  

- Fix SELinux labeling of special dirs

NO

###### Test Methodology

- Manually built and verified SELinux image.

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -25,10 +25,12 @@ import (
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/resources"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/retry"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safemount"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/tdnf"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/timestamp"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/userutils"
+	"golang.org/x/sys/unix"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


romoh (approved):

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1693,29 +1695,56 @@ func selinuxRelabelFiles(systemConfig configuration.SystemConfig, installChroot
...
-		onStdout := func(args ...interface{}) {
-			if len(args) > 0 {
-				files++
-				lastFile = fmt.Sprintf("%v", args)
+	targetRootPath := "/mnt/_bindmountroot"
+
+	for _, mountToLabel := range listOfMountsToLabel {
+		logger.Log.Debugf("Running setfiles to apply SELinux labels on mount points: %v", mountToLabel)
+
+		// Create a bind mount so that the filesystem's files can be labelled without interference from other mounts.
```
dmcilvaney, line 1703: Logic looks ok, but could you add a bit more context here? This looks a bit unnatural at first glance and this might help in the future.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1693,29 +1695,56 @@ func selinuxRelabelFiles(systemConfig configuration.SystemConfig, installChroot
...
+		err = bindMount.CleanClose()
 		if err != nil {
-			return fmt.Errorf("failed while labeling files (last file: %s) %w", lastFile, err)
+			return err
 		}
-		return err
-	})
+	}
+
+	err = os.RemoveAll(targetRootPath)
```
dmcilvaney, line 1744: I'm 50/50 on this, but part of me things this could be a deferred function... but on the other hand if this fails the build is dead anyways so... 🤷

dmcilvaney (approved):
Left two comments, not required though.

toolkit/tools/imagegen/installutils/installutils.go
cwize1: Considering that this is within the image and not the host, I think it is fine to leave it as is. Also, using `defer` has a tendency to encourage people to ignore errors.

toolkit/tools/imagegen/installutils/installutils.go
cwize1: I've expanded the comment.

dmcilvaney (approved):

PR #7373, opened 2024-01-22
Retry RPM cloning on http/502 error (bad gateway)

reubeno:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
We continue to intermittently see HTTP 502 errors (Bad Gateway) when tdnf download'ing packages from packages.microsoft.com, usually in build pipelines. This PR introduces granular retry logic for that specific failure case. (We can't retry on any HTTP error, as 404 is an expected error that shows up in normal execution.)

###### Change Log  
Update the rpmrepcloner to inspect stderr on tdnf download failure. If the specific known error string indicating a 502 error is found, then we retry with exponential backoff.

No

###### Test Methodology
(Testing in progress)

### DIFF
--- a/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
+++ b/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
@@ -10,12 +10,14 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/buildpipeline"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/packagerepo/repocloner"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/packagerepo/repomanager/rpmrepomanager"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/pkgjson"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/retry"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safechroot"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/shell"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/tdnf"
@@ -598,9 +600,14 @@ func (r *RpmRepoCloner) Close() error {
 // It will gradually enable more repos to consider until the package is found.
 func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err error) {
 	const (
-		unresolvedOutputPrefix  = "No package"
-		toyboxConflictsPrefix   = "toybox conflicts"
-		unresolvedOutputPostfix = "available"
+		// With 6 attempts, initial delay of 1 second, and a backoff factor of 3.0 the total time spent retrying will be
+		// 1 + 3 + 9 + 27 + 81 = 121 seconds.
+		//
+		// *NOTE* These values are copied from downloader/downloader.go; they need not be the same but seemed like a
+		// good enough starting point.
+		downloadRetryAttempts = 6
+		failureBackoffBase    = 3.0
+		downloadRetryDuration = time.Second
 	)
 
 	releaseverCliArg, err := tdnf.GetReleaseverCliArg()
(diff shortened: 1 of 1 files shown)


toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -615,47 +622,92 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
...
+	// cases in which the upstream repo doesn't contain the package and a 404 error is to be
+	// expected. This involves scraping through stderr, but it's better than not doing so.
+	//
+	if err != nil {
+		for _, line := range strings.Split(stderr, "\n") {
+			if strings.Contains(line, "Error: 502 when downloading") {
+				logger.Log.Warn("Encountered possibly intermittent HTTP 502 error.")
+				retriable = true
+			}
+		}
```
PawelWMS, line 708: Two things:
1. Do we only want to deal with 502s or all 5XX errors? Since we're here, it might be worth handling all server-side issues.
2. What do you think about dropping the `for` loop in favour of a regular expression?

```suggestion
		serverErrorMatch := serverErrorsRegex.FindStringSubmatch(stderr)
		if len(serverErrorMatch) > errorCodeIndex {
			logger.Log.Warn("Encountered possibly intermittent HTTP %d error.", serverErrorMatch[errorCodeIndex])
			retriable = true
		}
```

The regex and the index could be defined as global, private variables (making it global forces a one-time compilation of the regex during program's start-up):
```golang
// In global context:
var (
(...)
serverErrorsRegex = regexp.MustCompile(`(?m)Error: (5\d{2}) when downloading`)
errorCodeIndex = 1
...
```

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -615,47 +622,92 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
...
+	// on services that are prone to intermittent errors (e.g., HTTP 502 errors). We
+	// specifically look for such known cases and apply some retry logic in hopes of getting
+	// a better result; note that we don't indiscriminately retry because there are legitimate
+	// cases in which the upstream repo doesn't contain the package and a 404 error is to be
+	// expected. This involves scraping through stderr, but it's better than not doing so.
+	//
+	if err != nil {
+		for _, line := range strings.Split(stderr, "\n") {
+			if strings.Contains(line, "Error: 502 when downloading") {
+				logger.Log.Warn("Encountered possibly intermittent HTTP 502 error.")
```
PawelWMS, line 705: I'd turn that into a debug message - these tend to lead to a lot of confusion for the users. And may be a bit noisy, while in reality we retried and succeeded later on.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -598,9 +600,14 @@ func (r *RpmRepoCloner) Close() error {
...
 func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err error) {
 	const (
-		unresolvedOutputPrefix  = "No package"
-		toyboxConflictsPrefix   = "toybox conflicts"
-		unresolvedOutputPostfix = "available"
+		// With 6 attempts, initial delay of 1 second, and a backoff factor of 3.0 the total time spent retrying will be
+		// 1 + 3 + 9 + 27 + 81 = 121 seconds.
+		//
+		// *NOTE* These values are copied from downloader/downloader.go; they need not be the same but seemed like a
+		// good enough starting point.
```
PawelWMS, line 607: Is this comment here just for the sake of the draft? It looks like something that we may not want in the final version.

reubeno: I'd actually appreciate suggestions of how best to share these -- or perspectives on whether they should be? (Should we create a wrapper with these defaults?)

Perhaps you or @dmcilvaney have thoughts on that?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
reubeno: If I make it a debug message will it still end up int he logs by default? What I want to make sure we can do is extract them out in pipelines so we can provide feedback to the hosting services if we start seeing significantly more retries needed.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -615,47 +622,92 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
...
-		)
-		stdout, stderr, err = shell.Execute("tdnf", finalArgs...)
+		// We run in a retry loop on errors deemed retriable.
+		cancel := make(chan struct{})
+		retryNum := 1
+		_, err = retry.RunWithExpBackoff(func() error {
+			downloadErr, retriable := tdnfDownload(finalArgs...)
+			if downloadErr != nil {
+				if retriable {
+					logger.Log.Warnf("Attempt %d/%d: Failed to clone packages", retryNum, downloadRetryAttempts)
```
PawelWMS, line 632: I'd change that to a debug entry. These messages tend to generate noise, confuse people who don't know why they are generated (which is most users) and are red herrings when we debug issues.

```suggestion
					logger.Log.Debugf("Attempt %d/%d: Failed to clone packages", retryNum, downloadRetryAttempts)
```

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -615,47 +622,92 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
...
 		finalArgs := append(baseArgs, reposArgs...)
 
-		var (
-			stdout string
-			stderr string
-		)
-		stdout, stderr, err = shell.Execute("tdnf", finalArgs...)
+		// We run in a retry loop on errors deemed retriable.
+		cancel := make(chan struct{})
+		retryNum := 1
```
PawelWMS, line 627: I'm fine leaving `retryNum` here but if you have time, I think it might be worth considering changing how `Run*` functions work, so that they pass they keep track of the attempt number (print it on the debug level for instance) and the callback can then assume the attempt counting is done for them in the background.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -615,47 +622,92 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
...
+	// it will still download N-1 packages while also printing the message.
+	splitStdout := strings.Split(stdout, "\n")
+	for _, line := range splitStdout {
+		trimmedLine := strings.TrimSpace(line)
+		// Toybox conflicts are a known issue, reset the err value if encountered
+		if strings.HasPrefix(trimmedLine, toyboxConflictsPrefix) {
+			logger.Log.Warn("Ignoring known toybox conflict")
+			err = nil
+			continue
+		}
```
PawelWMS, line 686: I think I've seen a question about `toybox` and if we still need this check. It looks like `toybox` is not present in 2.0, so I think we can remove this logic completely.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: Yes, the logs gather everything on the debug level and up (down?).

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: Nevermind, I gave it a shot as well and it looks this change would be too intrusive for this PR.

PawelWMS:
Moving the PR to #7445.

PR #7374, opened 2024-01-22
virtiofsd: mention version dependency coming from Kata

sprt:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Adds a comment to the spec to ensure we keep the package version in sync with upstream Kata. When upstream Kata is built, they pull the virtiofsd version specified in that `versions.yaml` file, so we want to package the same version to avoid breakages. This recently came up during CVE analysis.

**NO**

### DIFF
--- a/SPECS/virtiofsd/virtiofsd.spec
+++ b/SPECS/virtiofsd/virtiofsd.spec
@@ -1,4 +1,6 @@
 Name:           virtiofsd
+# The version should be kept in sync with the `assets.virtiofsd.version` field in
+# https://github.com/microsoft/kata-containers/blob/msft-main/versions.yaml
 Version:        1.8.0
 Release:        1%{?dist}
 Summary:        Virtio-fs vhost-user device daemon (Rust version)


SPECS/virtiofsd/virtiofsd.spec
```diff
@@ -1,4 +1,6 @@
 Name:           virtiofsd
+# The version should be kept in sync with the `assets.virtiofsd.version` field in
```
dallasd1, line 2: I'm not sure this is the place we want to mention this requirement since virtiofsd is a standalone package, and Kata has a dependency on it. Is it possible to instead add a comment in the Kata spec mentioning the version dependency on virtiofsd?

sprt: Virtiofsd being a standalone package is a good point, however when a CVE targets only virtiofsd, devs wouldn't look at the Kata spec.

dallasd1: It sounds like the issue is when this package gets updated for a CVE or otherwise, Kata could break? Is it possible to anchor the version of virtiofsd in the Kata spec to one that meets its requirement in versions.yaml?

And if this package is updated, whose responsibility is it to update the Kata versions.yaml (if it needs to be updated)? Should we make the Kata team a required reviewer if it's the Kata team's responsibility to keep versions.yaml aligned?

ms-mahuber: ``Is it possible to anchor the version of virtiofsd in the Kata spec to one that meets its requirement in versions.yaml?`` - that sounds like the right approach.

For CVE fixes for virtiofsd I would not worry. A CVE fix is supposed to fix a vulnerability and not change behavior.

sprt: > Is it possible to anchor the version of virtiofsd in the Kata spec to one that meets its requirement in versions.yaml?

Ah yes that's a good point, didn't think of that. How does that work in PMC? Will multiple versions of virtiofsd live there without issue? What happens if the 2.0 branch has Kata requiring virtiofsd 1.8.0 but the virtiofsd spec version is 1.9.0?

> And if this package is updated, whose responsibility is it to update the Kata versions.yaml (if it needs to be updated)?

The virtiofsd package update should be driven by versions.yaml, not the other way around. When upstream Kata (or optionally us, if we need newer virtiofsd features) updates versions.yaml, we'll have to upgrade the virtiofsd package as well to match. versions.yaml is what dictates the virtiofsd version used by Kata.

> For CVE fixes for virtiofsd I would not worry. A CVE fix is supposed to fix a vulnerability and not change behavior.

The issue is that we fix some CVEs by upgrading the package by more than one version, which could be a breaking change. Regardless the virtiofsd version used by Kata should follow versions.yaml.

ms-mahuber: - Yes, multiple versions of virtiofsd can live side by side on PMC (old versions are retained).
- I think we need to clear whether the virtiofsd package will be exclusively used by the Kata group or not. If this is not the case, the virtiofsd package can be updated by anyone to a newer version.
- Dallas' comment about anchoring the version in the Kata SPECS makes sense regardless of that in order to express the current version dependency stemming from versions.yaml.
- The process from a kata-only perspective would be: When merging the microsoft/kata-containers package sources against a newer version from kata-containers/kata-containers, versions.yaml will be updated and the changes will need to be inspected. If the virtiofsd version changes, we will have to update the virtiofsd package as well, as you already mentioned.

sprt: Gotcha, agree on all points. Package ownership TBC by EOW. And I will anchor the virtiofsd in the upcoming PR I'll make against kata-containers.

PR #7376, opened 2024-01-22
Shrinking `memcached`

osamaesmailmsft:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

I removed %{systemd_requires} from `memcached` to shrink the package size.

###### Change Log  

- Above

NO

###### Test Methodology

- Pipeline build id: 488403

### DIFF
--- a/SPECS/memcached/memcached.spec
+++ b/SPECS/memcached/memcached.spec
@@ -38,7 +38,6 @@ BuildRequires:  libseccomp-devel
 BuildRequires:  openssl-devel
 %endif
 Requires(pre):  shadow-utils
-#%{?systemd_requires}
 
 %description
 memcached is a high-performance, distributed memory object caching
--- a/SPECS/memcached/memcached.spec
+++ b/SPECS/memcached/memcached.spec
@@ -7,7 +7,7 @@
 Summary:        High Performance, Distributed Memory Object Cache
 Name:           memcached
 Version:        1.6.22
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        BSD
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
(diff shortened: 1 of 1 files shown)


SPECS/memcached/memcached.spec
```diff
@@ -38,7 +38,6 @@ BuildRequires:  libseccomp-devel
 BuildRequires:  openssl-devel
 %endif
 Requires(pre):  shadow-utils
-%{?systemd_requires}
```
PawelWMS, line 41: How do we know the removal of these dependencies don't break anything?

osamaesmailmsft: It still passes tests when we remove it, so I believe it works. is that good, or should there be some more behind-the-scenes testing?

PawelWMS: The package tests only run in the build environment, so they have access to whatever is provided by the `BuildRequires`. But I'm not sure, if package tests are the main worry here. Since the removed macro only removes run-time dependency on `systemd`, the package tests are probably not influenced, but the post-install scripts clearly require some systemd infrastructure.

Can you say which scenario we're fixing with this change? Is it for distroless containers or something else? I don't want to modify a valid spec configuration (`systemd` is indeed needed by this package's scripts) if we're targeting a narrow set of images. It may be, that we need a change to how some images are built. For instance, maybe we need to expand the clean-up step for distroless containers to also drop `systemd` - I think it may not be used in these images anyway.

osamaesmailmsft: So, while working on memcached, we realized it installs 300 MB but is a small package itself, and it seems like removing this line shrinks the package significantly. To test out functionality like you're suggesting, I can try building a container with my version of memcached.

osamaesmailmsft: I didn't run into issues when testing a container locally :)

PR #7425, opened 2024-01-23
Adding support for multiple cache inputs

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

In more complex scenarios package builds want to hydrate their external RPMs cache from more than one source. This change makes that possible by making the input cache parameters an array.

Similarly, in some scenarios it's required to export the cache of the current package build.

###### Change Log  

- Converted singe input cache into an arbitrary number of input caches.
- Added a parameter for exporting the RPMs cache.
- Updating the toolkit with the `compress-cached-rpms` target.
- Clarified the naming of the input and output parameters.

No.

###### Test Methodology

- PR check.

### DIFF
--- a/toolkit/scripts/pkggen.mk
+++ b/toolkit/scripts/pkggen.mk
@@ -233,6 +233,7 @@ $(preprocessed_file): $(cached_file) $(go-graphPreprocessor)
 
 ######## PACKAGE BUILD ########
 
+cache_archive	= $(OUT_DIR)/cache.tar.gz
 pkggen_archive	= $(OUT_DIR)/rpms.tar.gz
 srpms_archive  	= $(OUT_DIR)/srpms.tar.gz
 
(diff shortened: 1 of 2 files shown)


toolkit/scripts/pkggen.mk
```diff
@@ -233,6 +233,7 @@ $(preprocessed_file): $(cached_file) $(go-graphPreprocessor)
 
 ######## PACKAGE BUILD ########
 
+cache_archive	= $(OUT_DIR)/cache.tar.gz
```
mfrw, line 236: Should we be modifying the toolkit via fasttrack route ?

PawelWMS: This change is intended to make improvements in the fast-track builds, thus this route. A good question, though, fast-track is in general intended for everything that needs to be pushed out quickly. In this case that's making fast-track faster, so that we can push out other things quicker.;)

mandeepsplaha (approved):

PR #7428, opened 2024-01-24
OS modifier: fix a bug if condition to not skip processing sshkeys when sshkeypaths is empty

elainezhao96:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 
fix a bug if condition to not skip processing sshkeys when sshkeypaths is empty

###### Change Log  
fix a bug if condition to not skip processing sshkeys when sshkeypaths is empty

NO

###### Test Methodology
test emu invocation in trident in local dev, and in the log, key was copied into .ssh/authorized_keys, and i can ssh into the vm with the private key on my local machine
image
image
image

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -1495,22 +1495,6 @@ func ProvisionUserSSHCerts(installChroot safechroot.ChrootInterface, username st
 
 	homeDir := userutils.UserHomeDirectory(username)
 	userSSHKeyDir := filepath.Join(homeDir, ".ssh")
-
-	exists, err = file.PathExists(userSSHKeyDir)
-	if err != nil {
-		logger.Log.Warnf("Error accessing %s directory: %v", userSSHKeyDir, err)
-		return
-	}
-	if !exists {
-		logger.Log.Debugf("Directory %s does not exist. Creating directory...", userSSHKeyDir)
-
-		err = os.Mkdir(userSSHKeyDir, os.FileMode(sshDirectoryPermission))
-		if err != nil {
-			logger.Log.Warnf("Failed to create %s directory: %v", userSSHKeyDir, err)
-			return
-		}
-	}
-
 	authorizedKeysFile := filepath.Join(userSSHKeyDir, "authorized_keys")
 
 	exists, err = file.PathExists(authorizedKeysTempFile)
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -1495,6 +1495,29 @@ func ProvisionUserSSHCerts(installChroot safechroot.ChrootInterface, username st
 
 	homeDir := userutils.UserHomeDirectory(username)
 	userSSHKeyDir := filepath.Join(homeDir, ".ssh")
+
+	exists, err = file.PathExists(userSSHKeyDir)
+	if err != nil {
+		logger.Log.Warnf("Error accessing %s directory: %v", userSSHKeyDir, err)
+		return
+	}
+	if !exists {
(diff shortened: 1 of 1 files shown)


romoh (approved):

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1489,12 +1489,35 @@ func ProvisionUserSSHCerts(installChroot safechroot.ChrootInterface, username st
...
+	}
+	if !exists {
+		logger.Log.Debugf("Directory %s does not exist. Creating directory...", userSSHKeyDir)
+
+		var permissionVal uint64
+		permissionVal, err = strconv.ParseUint(sshDirectoryPermission, 8, 32)
+		if err != nil {
+			logger.Log.Warnf("Failed to parse permissions: %v", err)
+			return
+		}
```
PawelWMS, line 1512: It looks like `sshDirectoryPermission` is not used anywhere but here anyway. With that in mind, please just replace that constant with the value you're passing to `os.Mkdir`, so we can drop this section of the code completely.

PawelWMS (dismissed):
I'm not sure if tooling is the right place for this code. Can you provide more details about the issue you're trying to fix? This looks like a potential misconfiguration of a package expecting that directory to be present.

toolkit/tools/imagegen/installutils/installutils.go
elainezhao96: in the code below it will be used as string, but I can declare it as int first and avoid this parsing logic

elainezhao96:
> an you provide more details about the issue you're trying to fix? This looks like a potential misconfiguration of a package expecting that directory to be present.

as the image customizer supports adding users and setting up sshkey access for the user, there could be a case that the .ssh dir does not exist in the OS, in our case, we have Trident project which invoke MIC binary to do in-OS customization, and .ssh dir  does not exist in the OS beforehand, so we create it before copying over the sshkeys. in this case, i dont think there is misconfiguration of a package.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1485,16 +1485,32 @@ func ProvisionUserSSHCerts(installChroot safechroot.ChrootInterface, username st
 	const squashErrors = false
 	const authorizedKeysTempFilePerms = 0644
 	const authorizedKeysTempFile = "/tmp/authorized_keys"
-	const sshDirectoryPermission = "0700"
+	const sshDirectoryPermission = 0700
```
cwize1, line 1488: nit: `0o700`.

This just makes it more explicit that you are specifying an octal number.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1485,16 +1485,32 @@ func ProvisionUserSSHCerts(installChroot safechroot.ChrootInterface, username st
...
+	if !exists {
+		logger.Log.Debugf("Directory %s does not exist. Creating directory...", userSSHKeyDir)
+
+		err = os.Mkdir(userSSHKeyDir, os.FileMode(sshDirectoryPermission))
+		if err != nil {
+			logger.Log.Warnf("Failed to create %s directory: %v", userSSHKeyDir, err)
+			return
+		}
+	}
+
```
cwize1, line 1513: I believe the `err = installChroot.AddFiles(fileToCopy)` line below is what creates the .ssh directory. So, I don’t think this change is necessary.

(GitHub doesn’t allow you to add comments to lines that are not a part of the PR. )

cwize1 (changes requested):

toolkit/tools/imagegen/installutils/installutils.go
elainezhao96: you're right. I will revert it and update this pr to only include the bug in the if condition

cwize1 (approved):

PR #7429, opened 2024-01-24
Refactor IsoMaker so that it is usable from MIC

gmileka:
###### Merge Checklist  

- [ ] Ready to merge

---

Refactor IsoMaker so that it is usable from MIC

Update IsoMaker to support the following:
- Moved all the code except main to a package so that it can be re-used by other tools.
- The current constructor assumes an installer iso with a config file.
  For the iso maker to be used to create liveos isos, we need to make a few changes:
  - Added a new constructor that takes in:
    - a pre-built config data structure (not a file). This allows other tools to build
      the config data structure dynamically and pass it to iso maker. There will be no
      need to provide a text file with the data.
    - a full path to a grub.cfg that can override the default iso grub.cfg.
    - a flag to indicate whether to include the bios bootloaders or not. Mic iso generation
      will not support bios bootloaders for now.
  - made the resources folder (and the subsequent copying of stock files) optional.
    This change is required since when creating a liveos iso, we do not need anything
    from the resource folder. The resources folder provided a bios bootloader and a stock
    grub.cfg..
    - The bios bootloader will not be support for mic iso generation.
    - The stock grub.cfg cannot be used for a liveos.
  - removed the requirement that an image config must have either Packages ot PackageLists nodes.
    This removal is necessary since when creating an iso from a full disk image, there is live installer
    running to install rpms - so there is no need to require such a list.
  - updated the MountPoint object to expose some of its fields as read-only properties.
    This change is required so that callers that get MountPoint objects can inspect/use the
    mount information as they need without relying on the functionality of chroot (as is the
    case before the change)..

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -72,9 +72,6 @@ func (s *SystemConfig) IsValid() (err error) {
 		return fmt.Errorf("invalid [BootType]: %s. Expecting values of either 'efi', 'legacy', 'none' or empty string", s.BootType)
 	}
 
-	if len(s.PackageLists) == 0 && len(s.Packages) == 0 {
```
cwize1, line 75: Out of interest, why was this check removed?

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -276,16 +324,38 @@ func (im *IsoMaker) prepareWorkDirectory() {
 // copyStaticIsoRootFiles copies architecture-independent files from the
 // Mariner repo directories.
 func (im *IsoMaker) copyStaticIsoRootFiles() {
-	staticIsoRootFilesPath := filepath.Join(im.resourcesDirPath, "assets/isomaker/iso_root_static_files/*")
 
-	logger.Log.Debugf("Copying static ISO root files from '%s'.", staticIsoRootFilesPath)
+	if im.resourcesDirPath == "" && im.grubCfgPath == "" {
+		logger.Log.Panicf("Missing required parameters. Must specify either the resources directory or provide a grub.cfg.")
```
cwize1, line 329: If we are going to make this code into a proper module/library, then all of these panics really ought to be removed and replaced with proper error handling.

cwize1 (changes requested):

toolkit/tools/imagegen/configuration/systemconfig.go
gmileka: Image definitions expect to always have packages to install (whether it's for a vhd(x) build or an iso build).
When mic uses IsoMaker to generate an iso, there are really no packages to include in the generated iso (just the squashfs). 
IsoMaker interface is an image definition file which goes through this code for validation. I'm relaxing the check so that image definitions from mic->isomaker can go through.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: Makes sense... will change.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -66,6 +79,38 @@ func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersi
 		configFilePath:     configFilePath,
```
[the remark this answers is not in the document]
gmileka: probably not anymore - let me check...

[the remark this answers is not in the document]
gmileka: Removed.

toolkit/tools/imagegen/configuration/systemconfig.go
cwize1: Should this check be added back in the imager tool?

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -276,16 +321,38 @@ func (im *IsoMaker) prepareWorkDirectory() {
...
 // Mariner repo directories.
 func (im *IsoMaker) copyStaticIsoRootFiles() {
-	staticIsoRootFilesPath := filepath.Join(im.resourcesDirPath, "assets/isomaker/iso_root_static_files/*")
 
-	logger.Log.Debugf("Copying static ISO root files from '%s'.", staticIsoRootFilesPath)
+	if im.resourcesDirPath == "" && im.grubCfgPath == "" {
+		logger.Log.Panicf("Missing required parameters. Must specify either the resources directory or provide a grub.cfg.")
+	}
+
+	if im.resourcesDirPath != "" {
```
cwize1, line 329: Does it make sense for both `resourcesDirPath` and `grubCfgPath` to be specified considering that the only resource file is the `grub.cfg` file?

cwize1 (changes requested):

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -140,6 +140,18 @@ func NewOverlayMountPoint(chrootDir, source, target, lowerDir, upperDir, workDir
 	return
 }
 
+func (m *MountPoint) GetSource() string {
```
cwize1, line 143: Is this being used for anything?

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: Well, the `resources` folder contains both arch-specific files and non-arch-specific files.
For the non-arch-specific files, grub.cfg is indeed the only file there today. But from the way the code is written, there's an assumption that this non-arch-specific group could include other files and the code will not need to change should a new file is dropped there.
Given that we have only grub.cfg, we can remove unused functionality... I'm thinking of leaving since it will be useful if new files are added to that rresources directory...

toolkit/tools/internal/safechroot/safechroot.go
gmileka: It's being using mic (different PR not published yet).

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: Created a draft PR here for the elimination of 'panic': https://github.com/microsoft/CBL-Mariner/pull/7454/files
Of special interest is the `defer` function error propagation (since the original intent was to stop execution when a clean-up fails).
@cwize1

toolkit/tools/internal/safechroot/safechroot.go
cwize1: Move this to the other change then?

cwize1 (changes requested):
(A couple of open comments.)

dmcilvaney (approved):

toolkit/tools/imagegen/configuration/systemconfig.go
PawelWMS: Judging by the test, this check is also verifying the proper content of the image config JSON. If the `imageconfigvalidator` tool is always ran when we build images, then yes, let's move the check **and the test** there as part of this PR. However, if the validator is not part of the image build, I think we should leave the check here, since a proper image configuration expect a package list.

@gmileka, could you elaborate more on how MIC is making use of the ISO maker? It sounds like it's outside the regular flow of how the toolkit is used. I'm worried we're making changes for a custom, non-official scenario, while breaking the official behaviour.

toolkit/tools/internal/safechroot/safechroot.go
PawelWMS: Once the other PR is published, please also link both of them with #XXXX mentions, so it's easier to see the connection between the changes.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -1,7 +1,7 @@
 // Copyright Microsoft Corporation.
 // Licensed under the MIT License.
 
-package main
+package isomakerlib
```
PawelWMS, line 4: Just `isomaker` to follow how we've been naming our packages elsewhere. Similarly, we don't add `pkg` suffixes.
```suggestion
package isomaker
```

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -52,20 +55,59 @@ func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersi
...
+}
+
+func NewIsoMakerWithConfig(unattendedInstall, enableBiosBoot bool, baseDirPath, buildDirPath, releaseVersion, resourcesDirPath string, config configuration.Config, initrdPath, grubCfgPath, isoRepoDirPath, outputDir, imageNameBase, imageNameTag string) *IsoMaker {
+
+	if imageNameBase == "" {
+		imageNameBase = defaultImageNameBase
+	}
+
+	if imageNameTag != "" {
+		imageNameTag = "-" + imageNameTag
```
PawelWMS, line 92: What if `imageNameTag` already begins with a hyphen? Do we want it doubled?

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -52,20 +55,59 @@ func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersi
...
+	}
+
+	// verifyConfig() panic if an error occurs.
```
PawelWMS, line 95: Nit-pick:
```suggestion
	// verifyConfig() panics if an error occurs.
```

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -276,16 +321,38 @@ func (im *IsoMaker) prepareWorkDirectory() {
...
 
-	logger.Log.Debugf("Copying static ISO root files from '%s'.", staticIsoRootFilesPath)
+	if im.resourcesDirPath == "" && im.grubCfgPath == "" {
+		logger.Log.Panicf("Missing required parameters. Must specify either the resources directory or provide a grub.cfg.")
+	}
+
+	if im.resourcesDirPath != "" {
+		staticIsoRootFilesPath := filepath.Join(im.resourcesDirPath, "assets/isomaker/iso_root_static_files/*")
+
+		logger.Log.Infof("Copying static ISO root files from '%s' to '%s'.", staticIsoRootFilesPath, im.buildDirPath)
```
PawelWMS, line 332: Why do we want this log on info level? If I recall correctly, it contains paths from toolkit's internals. Not really something a regular user cares about. I'd prefer to keep such internal stuff on debug level to minimize noise.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -276,16 +321,38 @@ func (im *IsoMaker) prepareWorkDirectory() {
...
+
+	// im.grubCfgPath allows the user to overwrite the default grub.cfg that is
+	// copied from the resource folder.
+	if im.grubCfgPath != "" {
+		targetGrubCfg := filepath.Join(im.buildDirPath, "boot/grub2/grub.cfg")
+		targetGrubCfgDir := filepath.Dir(targetGrubCfg)
+		logger.PanicOnError(os.MkdirAll(targetGrubCfgDir, os.ModePerm), "Failed while creating directory '%s'.", targetGrubCfgDir)
+
+		logger.Log.Infof("Copying '%s' to '%s'.", im.grubCfgPath, targetGrubCfg)
+		shell.MustExecuteLive("cp", im.grubCfgPath, targetGrubCfg)
```
PawelWMS, line 345: Could we extract that to a separate function? Similar to `recursiveCopyDereferencingLinks`.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -276,16 +321,38 @@ func (im *IsoMaker) prepareWorkDirectory() {
...
+	}
 }
 
 // copyArchitectureDependentIsoRootFiles copies the pre-built UEFI modules required
 // to boot the ISO image.
 func (im *IsoMaker) copyArchitectureDependentIsoRootFiles() {
+	if im.resourcesDirPath == "" || !im.enableBiosBoot {
```
PawelWMS, line 352: Could you add a comment explaining why we want to exit when this condition is true?

toolkit/tools/pkg/isomakerlib/isomaker.go
PawelWMS: I got to the changes inside `buildIsoImageFilePath`. This change seems to be related to that one. I'd suggest to move this logic into `buildIsoImageFilePath` completely, so it's all in one place and it's easier to understand why we modify the tag name. It's a bit cryptic here.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -462,8 +529,11 @@ func (im *IsoMaker) initializePaths() {
 // buildIsoImageFilePath gets the output ISO file path from the config JSON file name
 // and the image build environment.
 func (im *IsoMaker) buildIsoImageFilePath() string {
-	imageBaseName := strings.TrimSuffix(filepath.Base(im.configFilePath), ".json")
-	isoImageFileName := fmt.Sprintf("%v-%v%v.iso", imageBaseName, im.releaseVersion, im.imageNameTag)
+	isoImageFileNameSuffix := ""
+	if im.releaseVersion != "" || im.imageNameTag != "" {
+		isoImageFileNameSuffix = fmt.Sprintf("-%v%v", im.releaseVersion, im.imageNameTag)
+	}
```
PawelWMS, line 535: Related to my comment in `NewIsoMakerWithConfig`: maybe let's just split this into multiple calls like:
```suggestion
	imageBaseName := strings.TrimSuffix(filepath.Base(im.configFilePath), ".json")
	imageBaseName = appendImageNameSuffix(imageBaseName, im.releaseVersion)
	imageBaseName = appendImageNameSuffix(imageBaseName, im.imageNameTag)
```

And `appendImageNameSuffix` would do the check for an empty string and a hyphen.

toolkit/tools/pkg/isomakerlib/isomaker.go
cwize1: Some people on the Mariner team have been advocating for putting reusable code under the `pkg` directory. (There was a PR a while back to do a mass move into `pkg`. But it stalled out.)

The `lib` suffix is added to ensure the package directory name is unique. It can be quite annoying having to rename module in import statements to avoid name conflicts.

(This is also following existing precedent set by the `imagecustomizer` tool.)

toolkit/tools/imagegen/configuration/systemconfig.go
gmileka: @PawelWMS - sure...
- IsoMaker takes a configuration file, and a pointer to rpms and user artifacts and creates an iso with all of those artifacts.
- Mic takes an existing image (vhd(x)/qcow), extracts artifacts (rootfs, etc), and creates an iso with all of those artifacts.

The main change in this PR is about allowing IsoMaker to:
- accept an image configuration that might not have the same attributes as what's required today.
- accept an image configuration through the code and by reading it from a config file and parsing it (since MIC will generate that image configuration at run time).

The change should not change/break the existing behavior in anyway. It allows different entry points into the code though + relaxied config schema requirements.

The way I see it is that there is an input front-end and an "engine" to create the ISO. This change is providing another input front-end and leverages the same "engine" to create the ISO.

toolkit/tools/internal/safechroot/safechroot.go
gmileka: My plane is to make check-ins to the lower level layers first (IsoMaker) and then follow-up with the higher level layers.
These changes are library changes. Applications that depend on them will come in separate PRs.
I'll revert this change since it's harder to fit within the 'allow IsoMaker lib to take in a config struct instead of a config file.'

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: My thinking is that this function copies a set of files, and then follows with some overrides - so, I kept them together. Both blocks of code deal with copying static architecture neutral files needed for the iso image. What do you think Pawel?

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: updated.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: I've removed the code in the constructor that added the '-'. Now, this `Sprintf` is the only place where the name is constructed. I think we should be fine not adding `appendImageNameSuffix()` now.

toolkit/tools/imagegen/configuration/systemconfig.go
gmileka: I have decided to take it out because, today, IsoMaker itself does not run the full config validation and hence I do not have to change it right now. However, there is still an issue about who validates the schema and whether we should make changes to the scheme.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: @PawelWMS plz, let me know if you feel strongly on this one.

cwize1 (approved):

toolkit/tools/internal/safechroot/safechroot.go
gmileka: Reverted.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: fixed.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: Not needed. Changed back to Debugf.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: Resolving. Plz, re-activate if needed.

PR #7445, opened 2024-01-24
Specialcased 5XX errors during package downloads.

PawelWMS:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Migrating #7373 to this PR as a hand-off from @reubeno.

We continue to intermittently see HTTP 502 errors (Bad Gateway) when TDNF downloads packages from packages.microsoft.com, usually in build pipelines. This PR introduces granular retry logic for that specific failure case. (We can't retry on any HTTP error, as 404 is an expected error that shows up in normal execution.)

###### Change Log  

- Update the `rpmrepcloner` to inspect stderr on TDNF download failure. If an error string indicating a 5XX error is found, then we retry with exponential backoff.
- Unified the retry logic for network downloads.
- Removed outdated logic from `rpmrepocloner` looking for errors related with the `toybox` package.

No.

###### Associated issues  

- #7373

###### Test Methodology

- [Fast-track PR check build simulating building the extended packages (tends to fail due to PMC issues)](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=490221&view=results). I was able to observe a few `Package cloning attempt 1/5 failed with a retriable error.` messages in the logs followed by a successful download.

### DIFF
--- a/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
+++ b/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
@@ -53,8 +53,8 @@ const (
 )
 
 var (
-	serverErrorsRegex = regexp.MustCompile(`(?m)Error: (5\d{2}) when downloading`)
-	errorCodeIndex    = 1
+	serverErrorsRegex    = regexp.MustCompile(`(?m)Error: (5\d{2}) when downloading`)
+	serverErrorCodeIndex = 1
 )
 
 // RpmRepoCloner represents an RPM repository cloner.
@@ -626,6 +626,7 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
 				if retriable {
 					logger.Log.Debugf("Package cloning attempt %d/%d failed with a retriable error.", retryNum, retry.DefaultDownloadRetryAttempts)
 				} else {
+					logger.Log.Debugf("Package cloning attempt %d/%d failed with an unrecoverable error. Cancelling.", retryNum, retry.DefaultDownloadRetryAttempts)
 					close(cancel)
 				}
 			}
@@ -832,8 +833,8 @@ func tdnfDownload(args ...string) (err error, retriable bool) {
 	//
 	if err != nil {
 		serverErrorMatch := serverErrorsRegex.FindStringSubmatch(stderr)
-		if len(serverErrorMatch) > errorCodeIndex {
-			logger.Log.Debugf("Encountered possibly intermittent HTTP %s error.", serverErrorMatch[errorCodeIndex])
+		if len(serverErrorMatch) > serverErrorCodeIndex {
+			logger.Log.Debugf("Encountered possibly intermittent HTTP %s error.", serverErrorMatch[serverErrorCodeIndex])
 			retriable = true
 		}
 	}
--- a/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
+++ b/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
@@ -645,9 +645,9 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
 
 func tdnfDownload(args ...string) (err error, retriable bool) {
 	const (
-		unresolvedOutputPrefix  = "No package"
 		toyboxConflictsPrefix   = "toybox conflicts"
 		unresolvedOutputPostfix = "available"
+		unresolvedOutputPrefix  = "No package"
 	)
 
 	var (
--- a/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
+++ b/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
@@ -9,8 +9,8 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
-	"time"
 
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/buildpipeline"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
@@ -50,6 +50,11 @@ const (
 
 	useSingleTransaction    = true
 	useMultipleTransactions = !useSingleTransaction
+)
+
+var (
+	serverErrorsRegex = regexp.MustCompile(`(?m)Error: (5\d{2}) when downloading`)
+	errorCodeIndex    = 1
 )
 
 // RpmRepoCloner represents an RPM repository cloner.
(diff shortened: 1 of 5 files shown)


toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -809,3 +793,50 @@ func (r *RpmRepoCloner) reposArgsHaveOnlyLocalSources(reposArgs []string) bool {
...
+	)
+
+	stdout, stderr, err := shell.Execute("tdnf", args...)
+
+	logger.Log.Debugf("stdout: %s", stdout)
+	logger.Log.Debugf("stderr: %s", stderr)
+
+	// ============== TDNF SPECIFIC IMPLEMENTATION ==============
+	//
+	// Check if TDNF could not resolve a given package. If TDNF does not find a requested package,
```
reubeno, line 810: Are we reasonably confident that, if tdnf prints "No package ... available", then retry is never useful? That sounds reasonable, but want to confirm based on what we've seen in the past.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -947,14 +939,14 @@ func hydrateFromRemoteSource(fileHydrationState map[string]bool, newSourceDir st
 			}
 		}
 
-		cancelled, internalErr := retry.RunWithExpBackoff(func() error {
+		cancelled, internalErr := retry.RunWithDefaultDownloadBackoff(func() error {
```
reubeno, line 942: Nice! Looks like you found and de-duplicated some other copies of the constants; thanks for cleaning them up while you were at it :)

reubeno (approved):
Thanks for picking this up! It's a bit tough to test without artificially simulated conditions, so I'll be interested to see if those test runs do trigger the logic. Presumably we can always iterate over time too.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: I think "reasonably confident" is a good way to describe the way I see it. Failures with the 502 error don't make TDNF assume the package is unavailable (the "No package... available" message is not printed out anywhere). I think it is "reasonable" to assume then, that other server or network errors won't cause this message to appear and we can thus trust that if TDNF says the package is not there, there's no point to retry.

toolkit/tools/srpmpacker/srpmpacker.go
PawelWMS: I only hope this is not overspecializing the `retry` package's public functions. But you're welcome.

PawelWMS:
> Thanks for picking this up! It's a bit tough to test without artificially simulated conditions, so I'll be interested to see if those test runs do trigger the logic. Presumably we can always iterate over time too.

The extended package builds had some issues lately, so I hope we can hit this now. And yes, we will most probably need to keep improving on this, but I hope this is already a good upgrade.

toolkit/tools/srpmpacker/srpmpacker.go
reubeno: I think it's an iterative improvement, now that we've de-duped some magic numbers. If we deem that we need different call sites to have different policy, then we can still do so intentionally but can at least now have a conversation about why its defaults ought to be different.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -50,6 +52,11 @@ const (
 	useMultipleTransactions = !useSingleTransaction
 )
 
+var (
+	serverErrorsRegex = regexp.MustCompile(`(?m)Error: (5\d{2}) when downloading`)
+	errorCodeIndex    = 1
```
cwize1, line 57: `errorCodeIndex` is a very generic name and there is nothing to suggest it is related to `serverErrorsRegex`. Perhaps rename to `serverErrorsRegexErrorCodeIndex`?

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -615,40 +617,22 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
...
+		// We run in a retry loop on errors deemed retriable.
+		cancel := make(chan struct{})
+		retryNum := 1
+		_, err = retry.RunWithDefaultDownloadBackoff(func() error {
+			downloadErr, retriable := tdnfDownload(finalArgs...)
+			if downloadErr != nil {
+				if retriable {
+					logger.Log.Debugf("Package cloning attempt %d/%d failed with a retriable error.", retryNum, retry.DefaultDownloadRetryAttempts)
+				} else {
+					close(cancel)
```
cwize1, line 629: nit: Might be worth adding a comment stating that this is triggering the cancel signal. (It is a little unintuitive.)

cwize1 (approved):

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: Good call, it would be nice to know its server errors. I'll skip the "Regex" part to stick to how we named the regex-index pairs elsewhere (similar beginning, but only the actual regular expression has "Regex" in the name to avoid any confusion).

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
PawelWMS: Good idea. I'll put it in form of a log, so we may also ease some debugging while we're at it.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -615,40 +617,23 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
...
-				err = fmt.Errorf(trimmedLine)
-				break
+		// We run in a retry loop on errors deemed retriable.
+		cancel := make(chan struct{})
+		retryNum := 1
+		_, err = retry.RunWithDefaultDownloadBackoff(func() error {
+			downloadErr, retriable := tdnfDownload(finalArgs...)
+			if downloadErr != nil {
+				if retriable {
+					logger.Log.Debugf("Package cloning attempt %d/%d failed with a retriable error.", retryNum, retry.DefaultDownloadRetryAttempts)
```
dmcilvaney, line 627: Maybe this should be at least info, mayb warning? Would be nice to keep track of how prevalent these are without needing to pull the log files. Ideally this will never happen so if PMC improves it just goes away.

dmcilvaney:
```suggestion
					logger.Log.Warnf("Package cloning attempt %d/%d failed with a retriable error.", retryNum, retry.DefaultDownloadRetryAttempts)
```

toolkit/tools/internal/retry/retry.go
```diff
@@ -8,6 +8,14 @@ import (
 	"time"
 )
 
+const (
```
dmcilvaney, line 11: This works for me, but if we start doing more of these maybe we can move to a dedicated config struct like:
```go
type RetryConfig {
    BackoffBase float
    RetryAttempts int
    RetryDuration  time.Duration
}

const DefaultDownloadConfig = RetryConfig{ 2.0, 5, time.Second}
```
(note: Just threw that together on a whim, not sure its valid)

dmcilvaney: (to be clear, current version is fine)

dmcilvaney (approved):

Issue #6363, opened 2023-10-06
golang package GOROOT_FINAL seems incorrect


cpuguy83:
In https://github.com/microsoft/CBL-Mariner/blob/ae993d0b3288b1bb291f5f468169fa36b681b4d4/SPECS/golang/golang.spec#L82

It seems to be setting `GOROOT` to `/usr/bin/go`.
But `/usr/bin/go` is a binary.
Actual `GOROOT` should be `/usr/lib/golang` in this case.

I found this while trying to build packages that depend on go using the mariner toolkit.
For whatever reason in the chroot during package build /etc/profile is not being sourced so GOROOT (and other GO* vars) are left unset leaving them to the compiled-in versions.
This yields a somewhat cryptic error (or at least very unexpected) like: `go: cannot find GOROOT directory: /usr/bin/go`

cpuguy83:
Relevant for go1.21: https://github.com/golang/go/issues/62047

sozercan:
cc @jslobodzian

mfrw:
Fixed as per upgrade to go1.21.6

PR #7446, opened 2024-01-24
Bump golang version to 1.21.6-1 and switch to Microsoft Go

dagood:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Update `golang` package in 3.0 to use Microsoft Go 1.21.6-1, remove `msft-golang` package.

###### Change Log  

- Bump version to 1.21.6-1
- Switch from upstream Go to the Microsoft build of Go
- Clean up Go 1.17, 1.18 spec files

Unknown

###### Test Methodology

- PR tests


gdams:
@jslobodzian this is the progress we've made so far. Once we've got trigger access in AzDo we'll continue testing.

dagood:
We're having trouble with the buddy build, seems to be getting 404s very early on:
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=491743
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=491808

dagood:
After rebasing on 4faeb04997942aca81a5d4f94ef5032a7f90d2f1, a new buddy build hit a different error: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=491974&view=results

I'm taking a look.

SPECS/golang/golang-1.17.spec
```diff
@@ -1,236 +0,0 @@
-%global goroot          %{_libdir}/golang
```
mfrw, line 1: We might want to rebase because of:
- #7508
- #7511

Also, with the current state, I was able to build locally :)

gdams: > Also, with the current state, I was able to build locally :)

Good to know, we couldn't get the buddy build to pass but I'm not sure it was our spec file at fault.

@dagood can you rebase?

dagood: Rebased onto the latest build, https://dev.azure.com/mariner-org/mariner/_build/results?buildId=493968, I'm fairly sure.

Kicked off this buddy build with the build number: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=495337
And this one with `latest`: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=495338&view=results

dagood:
Rebased again (because there was already a conflicting change merged since the latest build) and kicked off https://dev.azure.com/mariner-org/mariner/_build/results?buildId=495369.

gdams (approved):
LGTM

SPECS/golang/golang.spec
```diff
@@ -13,20 +15,21 @@
...
 Group:          System Environment/Security
-URL:            https://golang.org
-Source0:        https://golang.org/dl/go%{version}.src.tar.gz
+URL:            https://github.com/microsoft/go
+Source0:        https://github.com/microsoft/go/releases/download/v%{version}-%{ms_go_revision}/go.%{ms_go_buildid}.src.tar.gz
 Source1:        https://dl.google.com/go/go1.4-bootstrap-20171003.tar.gz
-Source2:        https://dl.google.com/go/go%{bootstrap_compiler_version}.src.tar.gz
+Source2:        https://github.com/microsoft/go/releases/download/%{bootstrap_compiler_version}/go.20230802.5.src.tar.gz
 Patch0:         go14_bootstrap_aarch64.patch
 Obsoletes:      %{name} < %{version}
```
jslobodzian, line 29: I think we might want to remove the obsoletes..  If we are ultimately supporting the n and n-1 versions concurrently.  It might be that conflicts is more appropriate because we cannot have both installed at the same time....unless there is a way to do that.

SPECS/golang/golang.spec
```diff
@@ -68,7 +71,7 @@ popd
 rm -rf %{_libdir}/golang
 
 # Make go%{bootstrap_compiler_version} as the new bootstrapper
-mv -v %{_topdir}/BUILD/go1.19.12 %{_libdir}/golang
+mv -v %{_topdir}/BUILD/go%{bootstrap_compiler_version} %{_libdir}/golang
 
 # Build current go version
 export GOHOSTOS=linux
```
jslobodzian, line 77: Can you verify that these exports are correct?  Esepcially the "GOROOT" one.  I have a bug against Mariner 2.0 (https://github.com/microsoft/CBL-Mariner/issues/6363) that suggests GOROOT is pointing to the wrong place. 

     It seems to be setting GOROOT to /usr/bin/go.
     But /usr/bin/go is a binary.
     Actual GOROOT should be /usr/lib/golang in this case.

mfrw: The GOROOT_FINAL support has been dropped in go1.21.x
https://github.com/golang/go/issues/62047

dagood: Yeah, that's the wrong place. It should be possible to simply remove it--since 1.9, GOROOT is automatic. Although there might be something odd necessary due to bootstrapping, I don't have practical experience with it.

Also, the only part of `/etc/profile.d/go-exports.sh` that I think makes a difference nowadays is `GOPATH`, but I don't know whether it's actually good or not to move it from a user-specific path (default) to a systemwide path. For a container I suppose it doesn't matter.

I tried `go env` before and after `. /etc/profile.d/go-exports.sh`, and the difference was:
```diff
-GOMODCACHE="/root/go/pkg/mod"
+GOMODCACHE="/usr/share/gocode/pkg/mod"
-GOPATH="/root/go"
+GOPATH="/usr/share/gocode"
```

(I don't know in what circumstances a bad `GOROOT_FINAL` results in an error, this isn't something I've seen before.)

I wasn't able to get https://github.com/microsoft/CBL-MarinerTutorials/tree/main/build-in-container to work, so I don't have a good dev loop to try things out. (We decided to go ahead with this PR because if URL substitution fails, it probably won't be a subtle problem. 😄) I might have been hitting the same issues as our buddy builds have been hitting. This is something I'll have to look into later.

SPECS/golang/golang.spec
dagood: I'm not super familiar with the intricacies of `Obsoletes` vs. `Conflicts`, but I'm not sure if either is needed here. From what I can see, the package name has always been `golang`, so installing a specific version will either upgrade or downgrade an existing installed version. It seems pretty natural that a package will conflict with another version of itself so I guess it doesn't need to be stated in the spec. 😄 

`Obsoletes` seems to be more about handling one package(name) that no longer depends on another one, making it so that when installing the new version, the removed dependency will get cleaned up. Looking at all the `Obsoletes`es in https://src.fedoraproject.org/rpms/golang/blob/rawhide/f/golang.spec is interesting.

`Conflicts` seems like it's about conflicts between two different package(name)s. Although the intricacies [look complicated](https://docs.fedoraproject.org/en-US/packaging-guidelines/Conflicts) and it seems like it should be treated as a last resort.

SPECS/golang/golang.spec
```diff
@@ -13,20 +15,20 @@
...
 Release:        1%{?dist}
 License:        BSD-3-Clause
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 Group:          System Environment/Security
-URL:            https://golang.org
-Source0:        https://golang.org/dl/go%{version}.src.tar.gz
+URL:            https://github.com/microsoft/go
+Source0:        https://github.com/microsoft/go/releases/download/v%{version}-%{ms_go_revision}/go.%{ms_go_buildid}.src.tar.gz
 Source1:        https://dl.google.com/go/go1.4-bootstrap-20171003.tar.gz
```
dagood, line 26: Maybe worth mentioning that we are also hosting this at https://github.com/microsoft/go/releases/tag/v1.4.0-1, but I don't see any pressing reason to change the URL in this spec. The checksum is the same, and a download cache is used to actually build the package.

SPECS/golang/golang.spec
jslobodzian: If the tooling properly allows installing them side-by-side and the package is installing the contents properly, then yeah, remove the Obsoletes and the conflicts.

dagood: The Go tools work side by side, but this spec file doesn't define the package or install the files in a way that allows side-by-side installs.

Even so, the `Obsoletes: %{name} < %{version}` is not needed (nor any `Conflicts`), because this is already being treated as an upgrade. For example, https://github.com/microsoft/CBL-Mariner/blob/2.0/SPECS/cmake/cmake.spec doesn't include `Obsoletes`, and it installs some non-side-by-side files like `/usr/bin/cmake`. When I run `tdnf install -y cmake-3.21.4-3.cm2.x86_64` (old) then run `tdnf install cmake`, it asks if I want to upgrade:

```
root [ / ]# tdnf install cmake

Upgrading:
cmake              x86_64        3.21.4-10.cm2      mariner-official-base  46.54M      17.51M  
Total installed size:  46.54M
Total download size:  17.51M
Is this ok [y/N]:
```

As far as I can tell, it's theoretically possible to force dnf to install two versions on the same machine, but it's involved enough that I don't think anyone would do it (to any package, let alone golang) without either explicitly installing it to a different location or expecting some bad results.

Creating a set of RPM packages that work side by side seems doable to me, but not in the scope of this PR.

(I pushed a change to remove `Obsoletes:      %{name} < %{version}`.)

jslobodzian (approved):

mfrw (commented):
I have triggered a test build:
- [PR-7446](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=501707&view=results) ✅ 

We might need one more rebase because of merge-conflict though :)
@dagood

Changes LGTM :)

dagood:
Thanks! Rebased, and for good measure, kicked off https://dev.azure.com/mariner-org/mariner/_build/results?buildId=502010.

SPECS/golang/golang.spec
```diff
@@ -141,6 +143,10 @@ fi
 %{_bindir}/*
 
 %changelog
+* Wed Jan 24 2024 Davis Goodin <dagood@microsoft.com> - 1.21.6-1
```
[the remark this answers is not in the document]
dagood: Yeah, we're aware of the need for this, but it looks like `scripts/update_spec.sh` still needs a list of packages to bump, and we're expecting one step beyond that: a script that bumps all the dependent packages automatically.

mfrw (commented):
From my local testing of the PR, I think `podman` would need an update for `gvproxy`.
as per pr:
- #7640 
- [link](https://github.com/microsoft/CBL-Mariner/pull/7640/files#diff-9f80ea84a8b08bcc5b79595bff7acbfda31b628809958402a0827c589931898eL32)

@dagood

SPECS/golang/golang.spec
PawelWMS: If it's the same, I think it's still worth updating the URL so that there's no confusion about why we're using partially Microsoft and partially Google's sources. And if the name of the file is different in Microsoft's URL, adding `#<cached_file_name>` at the end of the URL will make the tool still pick the right file from our cache.

PawelWMS (approved):

SPECS/golang/golang.spec
dagood: Sure, pushed. Thanks for the note about `#`, that might be handy if we end up hitting some wall in our automation.

In theory, it might be useful to keep upstream's URL to show people that this package does use the same bootstrap sources as ordinary Go distributions. So, added a comment, too.

dagood:
Kicked off https://dev.azure.com/mariner-org/mariner/_build/results?buildId=507600 for the latest changes.

> From my local testing of the PR, I think `podman` would need an update for `gvproxy`.

@mfrw It would be interesting to see the errors/logs from that. Normally Go should be backward compatible, although that spec *does* seem to be building in a more complicated way than I'm used to. (Maybe as a new issue or over email to avoid clutter in this PR, which I don't think is blocked by this?)

dagood:
In the meantime, 1.21.7-1 has been released. I'm planning to submit an update PR to that version, which will also make sure the list of spec+other changes for a patch is what we expect and we can set up automation around it.

1.22.0-1 is also out now, but it requires inserting another bootstrapping step, so that's something we'll work on separately from the 1.21.7-1 update.

PR #7450, opened 2024-01-24
Update dracut to allow supressing user confirmation prompt when the liveos overlay is backed by memory.

gmileka:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 
Update dracut to allow supressing user confirmation prompt when the liveos overlay is backed by memory.

Dracut allows the creation of a LiveOS using a read-only squashfs and an read-write overlay on top.

If the read-write overlay is backed by a ram-disk, Dracut halts booting and prompts the user to confirm
whether to continue or not.

This interaction during the boot process is not desired in all cases.

This change introduces a new flag (rd.live.overlay.nouserconfirmprompt) that when defined, it supresses
the prompt and allows the boot process to continue to completion without user interation.

There is no impact to existing configurations and their associated behavior. Only when the new switch
is explicitly define by the image build (as a kernel parameter), the new behavior will take effect.

###### Change Log  

- n/a

**NO**

###### Associated issues  

- n/a

###### Links to CVEs  
- n/a

###### Test Methodology

- Build dracut package.
- Build baremetal core image.
- generate a LiveOS iso where the new flag is included in the kernel parameters in grub.cfg.
- create a VM and boot from thee iso.
- verify the boot is successful and no user input is required.

### DIFF
--- a/SPECS/dracut/allow-liveos-overlay-no-user-confirmation-prompt.patch
+++ b/SPECS/dracut/allow-liveos-overlay-no-user-confirmation-prompt.patch
@@ -1,6 +1,30 @@
---- a/modules.d/90dmsquash-live/dmsquash-live-root.sh	2021-05-27 05:34:19.000000000 -0700
-+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh	2024-01-24 13:46:46.839171333 -0800
-@@ -25,6 +25,10 @@
+From 4d47f0bae243577a4cf634ae5e01b324cf78e7eb Mon Sep 17 00:00:00 2001
+From: George Mileka <gmileka@microsoft.com>
+Date: Thu, 25 Jan 2024 15:06:13 -0800
+Subject: [PATCH] Update dracut to allow supressing user confirmation prompt
+ when the liveos overlay is backed by memory.
+
+Dracut allows the creation of a LiveOS using a read-only squashfs and an read-write overlay on top.
+
+If the read-write overlay is backed by a ram-disk, Dracut halts booting and prompts the user to confirm
+whether to continue or not.
+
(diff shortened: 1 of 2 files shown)


SPECS/dracut/allow-liveos-overlay-no-user-confirmation-prompt.patch
```diff
@@ -0,0 +1,22 @@
+--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh	2021-05-27 05:34:19.000000000 -0700
```
romoh, line 1: ⁠Is this a patch from upstream? or is it your contribution to solve this problem?

gmileka: not from upstream. I've authored this patch to unblock the boot flow.

amritakohli:
clean up / update description

SPECS/dracut/allow-liveos-overlay-no-user-confirmation-prompt.patch
dmcilvaney, line 1: Minor style thing, but I tend to prefer using `git format-patch...` to create these instead of `git diff`. The patch is the same but we can include author, commit message, etc.

This is small enough to not really matter though, very clear what its trying to do.

dmcilvaney (approved):

SPECS/dracut/allow-liveos-overlay-no-user-confirmation-prompt.patch
gmileka: thx Daniel - I've re-generated the patch using `git format-patch` - much better.

gmileka:
Updated description.

PR #7454, opened 2024-01-25
The isomaker library should return errors on failure and not terminate …

gmileka:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

IsoMaker was originally written as a single go package.  However, as MIC is starting to support ISO, it needs to share some of this code, and the code has been refactored into a separate package.
The problem is that this code still panics when errors are returned. A software library should not painic and terminate the process. Instead, it should report errors to the caller, and the caller can decide how to handle it.
This change converts all function calls in the isomaker pkg to return error codes.

###### Change Log  

- Update pkg/isomakerlib so that it does not panic on error, and instead returns an error.

**NO**

###### Associated issues  

- [6755 Update iso maker golang pkg to return errors](https://dev.azure.com/mariner-org/ECF/_workitems/edit/6755)

###### Links to CVEs  
- n/a

###### Test Methodology

- Build an iso based on the baremetal image defintiion.
- Install the iso on a hyper-v VM.

### DIFF
--- a/toolkit/tools/pkg/isomakerlib/isomaker.go
+++ b/toolkit/tools/pkg/isomakerlib/isomaker.go
@@ -119,7 +119,7 @@ func (im *IsoMaker) Make() (err error) {
 			if err != nil {
 				err = fmt.Errorf("%w\n%w", err, cleanupErr)
 			} else {
-				err = fmt.Errorf("%w", cleanupErr)
+				err = cleanupErr
 			}
 		}
 	}()
@@ -414,7 +414,10 @@ func (im *IsoMaker) prepareWorkDirectory() error {
 	}
 	if exists {
 		logger.Log.Warningf("Unexpected: temporary ISO build path '%s' exists. Removing.", im.buildDirPath)
-		logger.PanicOnError(os.RemoveAll(im.buildDirPath), "Failed while removing directory '%s'.", im.buildDirPath)
+		err = os.RemoveAll(im.buildDirPath)
+		if err != nil {
+			return fmt.Errorf("failed while removing directory '%s':\n%w", im.buildDirPath, err)
+		}
 	}
 
 	err = os.Mkdir(im.buildDirPath, os.ModePerm)
@@ -505,7 +508,7 @@ func (im *IsoMaker) copyArchitectureDependentIsoRootFiles() error {
 	}
 
 	if im.resourcesDirPath == "" && im.enableBiosBoot {
-		logger.Log.Panicf("missing required parameters. Must specify the resources directory if BIOS bootloaders are to be included.")
+		return fmt.Errorf("missing required parameters. Must specify the resources directory if BIOS bootloaders are to be included.")
 	}
 
 	architectureDependentFilesDirectory := filepath.Join(im.resourcesDirPath, isoRootArchDependentDirPath, runtime.GOARCH, "*")
@@ -770,20 +773,17 @@ func (im *IsoMaker) deferIsoMakerCleanUp(cleanUpTask func() error) {
 // isoMakerCleanUp runs all clean-up tasks scheduled through "deferIsoMakerCleanUp".
 // Tasks are ran in reverse order to how they were scheduled.
 func (im *IsoMaker) isoMakerCleanUp() (err error) {
-	// We defer again to run the tasks in the correct order AND so that a potential
-	// panic from one of these doesn't prevent other ones from running.
-	for _, cleanUpTask := range im.isoMakerCleanUpTasks {
-		defer func() {
-			cleanupErr := cleanUpTask()
-			if cleanupErr != nil {
-				if err != nil {
-					err = fmt.Errorf("%w\n%w", err, cleanupErr)
-				} else {
-					err = fmt.Errorf("%w", cleanupErr)
-				}
+	for i := len(im.isoMakerCleanUpTasks) - 1; i >= 0; i-- {
+		cleanupErr := im.isoMakerCleanUpTasks[i]()
+		if cleanupErr != nil {
+			if err != nil {
+				err = fmt.Errorf("%w\n%w", err, cleanupErr)
+			} else {
+				err = cleanupErr
 			}
-		}()
+		}
 	}
+
 	return err
 }
 
--- a/toolkit/tools/pkg/isomakerlib/isomaker.go
+++ b/toolkit/tools/pkg/isomakerlib/isomaker.go
@@ -47,20 +47,25 @@ type IsoMaker struct {
 	imageNameBase      string               // Base name of the ISO to generate (no path, and no file extension).
 	imageNameTag       string               // Optional user-supplied tag appended to the generated ISO's name.
 
-	isoMakerCleanUpTasks []func() // List of clean-up tasks to perform at the end of the ISO generation process.
+	isoMakerCleanUpTasks []func() error // List of clean-up tasks to perform at the end of the ISO generation process.
 }
 
 // NewIsoMaker returns a new ISO maker.
-func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersion, resourcesDirPath, configFilePath, initrdPath, isoRepoDirPath, outputDir, imageNameTag string) *IsoMaker {
+func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersion, resourcesDirPath, configFilePath, initrdPath, isoRepoDirPath, outputDir, imageNameTag string) (*IsoMaker, error) {
 	if baseDirPath == "" {
 		baseDirPath = filepath.Dir(configFilePath)
 	}
 
 	imageNameBase := strings.TrimSuffix(filepath.Base(configFilePath), ".json")
 
-	// readConfigFile() and verifyConfig() panic if an error occurs.
-	config := readConfigFile(configFilePath, baseDirPath)
-	verifyConfig(config, unattendedInstall)
+	config, err := readConfigFile(configFilePath, baseDirPath)
+	if err != nil {
+		return nil, err
+	}
+	err = verifyConfig(config, unattendedInstall)
+	if err != nil {
+		return nil, err
+	}
 
 	return &IsoMaker{
 		enableBiosBoot:     true,
@@ -75,17 +80,19 @@ func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersi
 		outputDirPath:      outputDir,
 		imageNameBase:      imageNameBase,
 		imageNameTag:       imageNameTag,
-	}
+	}, nil
 }
 
-func NewIsoMakerWithConfig(unattendedInstall, enableBiosBoot bool, baseDirPath, buildDirPath, releaseVersion, resourcesDirPath string, config configuration.Config, initrdPath, grubCfgPath, isoRepoDirPath, outputDir, imageNameBase, imageNameTag string) *IsoMaker {
+func NewIsoMakerWithConfig(unattendedInstall, enableBiosBoot bool, baseDirPath, buildDirPath, releaseVersion, resourcesDirPath string, config configuration.Config, initrdPath, grubCfgPath, isoRepoDirPath, outputDir, imageNameBase, imageNameTag string) (*IsoMaker, error) {
 
 	if imageNameBase == "" {
 		imageNameBase = defaultImageNameBase
 	}
 
-	// verifyConfig() panics if an error occurs.
-	verifyConfig(config, unattendedInstall)
+	err := verifyConfig(config, unattendedInstall)
+	if err != nil {
+		return nil, err
+	}
 
 	return &IsoMaker{
 		enableBiosBoot:     enableBiosBoot,
@@ -101,25 +108,51 @@ func NewIsoMakerWithConfig(unattendedInstall, enableBiosBoot bool, baseDirPath, 
 		outputDirPath:      outputDir,
 		imageNameBase:      imageNameBase,
 		imageNameTag:       imageNameTag,
-	}
+	}, nil
 }
 
 // Make builds the ISO image to 'buildDirPath' with the packages included in the config JSON.
-func (im *IsoMaker) Make() {
-	defer im.isoMakerCleanUp()
+func (im *IsoMaker) Make() (err error) {
+	defer func() {
+		cleanupErr := im.isoMakerCleanUp()
+		if cleanupErr != nil {
+			if err != nil {
+				err = fmt.Errorf("%w\n%w", err, cleanupErr)
+			} else {
+				err = fmt.Errorf("%w", cleanupErr)
+			}
+		}
+	}()
 
-	im.initializePaths()
+	err = im.initializePaths()
+	if err != nil {
+		return err
+	}
 
-	im.prepareWorkDirectory()
+	err = im.prepareWorkDirectory()
+	if err != nil {
+		return err
+	}
 
-	im.createIsoRpmsRepo()
+	err = im.createIsoRpmsRepo()
+	if err != nil {
+		return err
+	}
 
-	im.prepareIsoBootLoaderFilesAndFolders()
+	err = im.prepareIsoBootLoaderFilesAndFolders()
+	if err != nil {
+		return err
+	}
 
-	im.buildIsoImage()
+	err = im.buildIsoImage()
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
 
-func (im *IsoMaker) buildIsoImage() {
+func (im *IsoMaker) buildIsoImage() error {
 	isoImageFilePath := im.buildIsoImageFilePath()
 
 	logger.Log.Infof("Generating ISO image under '%s'.", isoImageFilePath)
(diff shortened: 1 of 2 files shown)


toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -540,48 +759,64 @@ func (im *IsoMaker) buildIsoImageFilePath() string {
...
 
 // isoMakerCleanUp runs all clean-up tasks scheduled through "deferIsoMakerCleanUp".
 // Tasks are ran in reverse order to how they were scheduled.
-func (im *IsoMaker) isoMakerCleanUp() {
+func (im *IsoMaker) isoMakerCleanUp() (error err) {
 	// We defer again to run the tasks in the correct order AND so that a potential
 	// panic from one of these doesn't prevent other ones from running.
 	for _, cleanUpTask := range im.isoMakerCleanUpTasks {
-		defer cleanUpTask()
+		defer func() {
```
cwize1, line 772: The defer here might no longer be neccessary.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -540,48 +759,64 @@ func (im *IsoMaker) buildIsoImageFilePath() string {
...
 	}
 }
 
-func readConfigFile(configFilePath, baseDirPath string) configuration.Config {
+func readConfigFile(configFilePath, baseDirPath string) (configuration.Config, error) {
 	config, err := configuration.LoadWithAbsolutePaths(configFilePath, baseDirPath)
-	logger.PanicOnError(err, "Failed while reading config file from '%s' with base directory '%s'.", configFilePath, baseDirPath)
-	return config
+	if err != nil {
+		return configuration.Config{}, fmt.Errorf("Failed while reading config file from '%s' with base directory '%s'.\nError:%v", configFilePath, baseDirPath, err)
```
cwize1, line 788: nit: In Go, errors start with a lower case letter.

cwize1, line 788: When wrapping an existing error object, you should use `%w` not `%v`.

cwize1, line 788: nit: Instead of `.\nError:%v` use `:\n%w`.

You don't need to add the `Error:` since we already know it is an error.

Similar comment elsewhere.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -212,34 +282,63 @@ func (im *IsoMaker) setUpIsoGrub2Bootloader() {
...
+		return err
+	}
 
 	defer func() {
 		logger.Log.Debugf("Unmounting '%s'.", efiBootImgTempMountDir)
-		logger.PanicOnError(syscall.Unmount(efiBootImgTempMountDir, 0), "Failed to unmount '%s'.", efiBootImgTempMountDir)
+		tmpErr := syscall.Unmount(efiBootImgTempMountDir, 0)
+		if tmpErr != nil {
+			if err != nil {
+				err = fmt.Errorf("Failed to unmount '%s'.\nError:%v\nClean-up Error:%v", efiBootImgTempMountDir, err, tmpErr)
```
cwize1, line 295: nit: `"failed to unmount '%s':\n%w:\nclean-up error: %w"`

cwize1 (changes requested):

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: would having "error" help distinguish between the two error messages in the text?

gmileka: fixed.

gmileka: fixed.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: fixed.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: When tasks are added for later clean-up, they are added to an array (see `deferIsoMakerCleanUp`).
When they are executed, it is desired that they are executed in the opposite order.
The original implementation relied on defer to do that reversal.
We can traverse the array also in the opposite order during this clean-up.
I find traversing the array in the reverse order more readable. But I'm okay either way...
@cwize1 Do you have a preference?

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: removed.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -546,48 +763,65 @@ func (im *IsoMaker) buildIsoImageFilePath() string {
...
+func (im *IsoMaker) deferIsoMakerCleanUp(cleanUpTask func() error) {
 	im.isoMakerCleanUpTasks = append(im.isoMakerCleanUpTasks, cleanUpTask)
 }
 
 // isoMakerCleanUp runs all clean-up tasks scheduled through "deferIsoMakerCleanUp".
 // Tasks are ran in reverse order to how they were scheduled.
-func (im *IsoMaker) isoMakerCleanUp() {
+func (im *IsoMaker) isoMakerCleanUp() (err error) {
 	// We defer again to run the tasks in the correct order AND so that a potential
 	// panic from one of these doesn't prevent other ones from running.
```
cwize1, line 774: Existing comment mentions “panic”.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -546,48 +763,65 @@ func (im *IsoMaker) buildIsoImageFilePath() string {
...
 	for _, cleanUpTask := range im.isoMakerCleanUpTasks {
-		defer cleanUpTask()
+		defer func() {
+			cleanupErr := cleanUpTask()
+			if cleanupErr != nil {
+				if err != nil {
+					err = fmt.Errorf("%w\n%w", err, cleanupErr)
+				} else {
+					err = fmt.Errorf("%w", cleanupErr)
```
cwize1, line 782: Simplify to: `err = cleanupErr`?

toolkit/tools/pkg/isomakerlib/isomaker.go
cwize1: Yeah, I’d also prefer being explicit and iterating the list in reverse.

cwize1 (changes requested):

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -47,20 +47,25 @@ type IsoMaker struct {
 	imageNameBase      string               // Base name of the ISO to generate (no path, and no file extension).
 	imageNameTag       string               // Optional user-supplied tag appended to the generated ISO's name.
 
-	isoMakerCleanUpTasks []func() // List of clean-up tasks to perform at the end of the ISO generation process.
+	isoMakerCleanUpTasks []func() error // List of clean-up tasks to perform at the end of the ISO generation process.
 }
 
 // NewIsoMaker returns a new ISO maker.
-func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersion, resourcesDirPath, configFilePath, initrdPath, isoRepoDirPath, outputDir, imageNameTag string) *IsoMaker {
+func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersion, resourcesDirPath, configFilePath, initrdPath, isoRepoDirPath, outputDir, imageNameTag string) (*IsoMaker, error) {
```
PawelWMS, line 54: In cases of more errors (or even with one, recently) we're tring to use named output arguments:
```suggestion
func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersion, resourcesDirPath, configFilePath, initrdPath, isoRepoDirPath, outputDir, imageNameTag string) (isoMake *IsoMaker, err error) {
```

This may require a few changes below.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -145,30 +178,51 @@ func (im *IsoMaker) buildIsoImage() {
 		// Directory to convert to an ISO.
 		im.buildDirPath)
 
-	shell.MustExecuteLive("mkisofs", mkisofsArgs...)
+	err := shell.ExecuteLive(false /*squashErrors*/, "mkisofs", mkisofsArgs...)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
PawelWMS, line 186: If you'd like, you can also do this for such simpler cases:
```suggestion
	return shell.ExecuteLive(false /*squashErrors*/, "mkisofs", mkisofsArgs...)
```

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -101,25 +108,51 @@ func NewIsoMakerWithConfig(unattendedInstall, enableBiosBoot bool, baseDirPath,
...
 
 // Make builds the ISO image to 'buildDirPath' with the packages included in the config JSON.
-func (im *IsoMaker) Make() {
-	defer im.isoMakerCleanUp()
+func (im *IsoMaker) Make() (err error) {
+	defer func() {
+		cleanupErr := im.isoMakerCleanUp()
+		if cleanupErr != nil {
+			if err != nil {
+				err = fmt.Errorf("%w\n%w", err, cleanupErr)
```
PawelWMS, line 120: It might be worth to distinguish between the two errors here to make debugging easier.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -183,18 +237,34 @@ func (im *IsoMaker) setUpIsoGrub2Bootloader() {
...
+		return err
+	}
 
 	defer func() {
 		logger.Log.Debugf("Removing '%s'.", efiBootImgTempMountDir)
-		logger.PanicOnError(os.RemoveAll(efiBootImgTempMountDir), "Failed to remove temporary mount directory '%s'.", efiBootImgTempMountDir)
+		cleanupErr := os.RemoveAll(efiBootImgTempMountDir)
+		if cleanupErr != nil {
+			if err != nil {
+				err = fmt.Errorf("failed to remove temporary mount directory '%s':\n%w\n%w", efiBootImgTempMountDir, err, cleanupErr)
```
PawelWMS, line 263: Similar ask here to add a word or two to separate the cause of the error in `err` from `cleanupErr`.

toolkit/tools/pkg/isomakerlib/isomaker.go
PawelWMS: +1 for reverse iteration. I think with the previous use of panics a `defer` was just necessary.

toolkit/tools/pkg/isomakerlib/isomaker.go
PawelWMS: I think this may be fine. In case more clean-up tasks fail, we'd lose their errors.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: thx - updated.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: updated.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: If err is nil, it makes sense to have `err = cleanupErr` and not the `err = fmt.Errorf("%w", cleanupErr)`. I think that's what Chris meant. I've updated it to reflect that. Failures from subsequent calls will just accumulate in the err object.

cwize1 (approved):

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: I had `%w\nError:%w` before, but opted to simplify...
I find it useful to distinguish between clean-up errors and real errors that affect the output.
Would something like `%w:\nclean-up error:%w` be okay?
@cwize1 @PawelWMS

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: @PawelWMS - used named output arguments where it makes sense.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: updated functions that have this pattern.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: added `clean-up error:` prefix to error messages from clean-up...

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: updated.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -76,18 +81,22 @@ func NewIsoMaker(unattendedInstall bool, baseDirPath, buildDirPath, releaseVersi
...
 
 	if imageNameBase == "" {
 		imageNameBase = defaultImageNameBase
 	}
 
-	// verifyConfig() panics if an error occurs.
-	verifyConfig(config, unattendedInstall)
+	err = verifyConfig(config, unattendedInstall)
+	if err != nil {
+		return nil, err
```
PawelWMS, line 96: Purely informal, no need to change anything, if you prefer this format: with the named output, you can now simplify all returns like so.
```suggestion
		return
```

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -145,30 +182,41 @@ func (im *IsoMaker) buildIsoImage() {
...
 
 // copyInitrd copies a pre-built initrd into the isolinux folder.
-func (im *IsoMaker) copyInitrd() {
+func (im *IsoMaker) copyInitrd() error {
 	initrdDestinationPath := filepath.Join(im.buildDirPath, "isolinux/initrd.img")
 
 	logger.Log.Debugf("Copying initrd from '%s'.", im.initrdPath)
 
-	file.Copy(im.initrdPath, initrdDestinationPath)
+	return file.Copy(im.initrdPath, initrdDestinationPath)
```
PawelWMS, line 214: Nice catch.

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -245,102 +338,150 @@ func (im *IsoMaker) copyShimFromInitrd(efiBootImgTempMountDir, bootBootloaderFil
...
-	im.deferIsoMakerCleanUp(func() {
+	im.deferIsoMakerCleanUp(func() error {
 		logger.Log.Debugf("Removing '%s'.", im.buildDirPath)
-
-		logger.PanicOnError(os.RemoveAll(im.buildDirPath), "Failed to remove '%s'.", im.buildDirPath)
+		err := os.RemoveAll(im.buildDirPath)
+		if err != nil {
+			err = fmt.Errorf("failed to remove '%s':\n%w", im.buildDirPath, err)
+		}
+		return err
```
PawelWMS, line 423: The code is correct but to avoid confusion with the `err` variable from `prepareWorkDirectory` we tend to give distinct names to the error variables in anonymous functions:
```suggestion
		err := os.RemoveAll(im.buildDirPath)
		if cleanupErr != nil {
			cleanupErr = fmt.Errorf("failed to remove '%s':\n%w", im.buildDirPath, cleanupErr)
		}
		return cleanupErr
```

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -245,102 +338,150 @@ func (im *IsoMaker) copyShimFromInitrd(efiBootImgTempMountDir, bootBootloaderFil
...
 }
 
 // copyStaticIsoRootFiles copies architecture-independent files from the
 // Mariner repo directories.
-func (im *IsoMaker) copyStaticIsoRootFiles() {
+func (im *IsoMaker) copyStaticIsoRootFiles() (err error) {
 
 	if im.resourcesDirPath == "" && im.grubCfgPath == "" {
-		logger.Log.Panicf("missing required parameters. Must specify either the resources directory or provide a grub.cfg.")
+		return fmt.Errorf("missing required parameters. Must specify either the resources directory or provide a grub.cfg.")
```
PawelWMS, line 449: I spotted this just now: with `fmt.Errorf` it's a Golang convention we try to follow to not include a period at the end:
```suggestion
		return fmt.Errorf("missing required parameters. Must specify either the resources directory or provide a grub.cfg")
```

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -352,143 +493,197 @@ func (im *IsoMaker) copyArchitectureDependentIsoRootFiles() {
 	// enableBiosBoot will not affect those on-architecture dependent files
 	// though.
 	if im.resourcesDirPath == "" || !im.enableBiosBoot {
-		return
+		return nil
 	}
 
 	if im.resourcesDirPath == "" && im.enableBiosBoot {
-		logger.Log.Panicf("missing required parameters. Must specify the resources directory if BIOS bootloaders are to be included.")
+		return fmt.Errorf("missing required parameters. Must specify the resources directory if BIOS bootloaders are to be included.")
```
PawelWMS, line 500:
```suggestion
		return fmt.Errorf("missing required parameters. Must specify the resources directory if BIOS bootloaders are to be included")
```

PawelWMS (approved):
Just a few small comments but looks good.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: fixed.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: fixed.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: good point - fixed.

toolkit/tools/pkg/isomakerlib/isomaker.go
gmileka: I'll leave the explicit returns for now, Pawel. I find the explicitness to be more readible.

PR #7459, opened 2024-01-25
Sort mariner-required-configs.json for kernel configs and add a workflow check for it

vinceaperri:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR sorts the mariner-required-configs.json to make it human friendly. It also adds a workflow check to ensure it stays sorted. This check will print out a patch that can be applied if the check fails.

###### Change Log  
- Sort keys of mariner-required-configs.json using python3 json library.
- Add check_required_configs_fmt.py script and inject it into check-kernel-config workflow.
- Add hook to check-kernel-config workflow so it will run when mariner-required-configs.json is changed.

**NO**

###### Associated issues  

- Original PR that added this JSON: #5530

###### Test Methodology
- This PR is self-validated with the check below. See `Check mariner-required-configs.json format` step in `Kernel Required Configs Check / Kernel configs check (pull_request)`. It should pass since this PR also sorts the JSON.
- #7460 should fail at this step because config info keys aren't sorted
- #7461 should fail at this step because config values aren't sorted
- #7462 should fail at this step because config options themselves aren't sorted.


toolkit/scripts/check_required_config_fmt.py
```diff
@@ -0,0 +1,42 @@
...
+    # successfully.
+    try:
+        subprocess.check_call(['git', 'diff',
+            '--no-index',
+            required_configs,
+            tmp_file])
+    except subprocess.CalledProcessError:
+        sys.exit(1)
+    finally:
+        os.remove(tmp_file)
```
trungams, line 42:
```suggestion
    # Sort the JSON file and write it to a temporary file.
    with tempfile.NamedTemporaryFile(mode='w') as tmp_file:
        json.dump(content, tmp_file, indent=4, sort_keys=True)

        # Compare the sorted JSON file to the original. If they are different,
        # print the diff that can be applied to the JSON file to fix it and
        # exit with a non-zero exit code. Otherwise, print nothing and exit
        # successfully.
        try:
            subprocess.check_call(['git', 'diff',
                '--no-index',
                required_configs,
                tmp_file.name])
        except subprocess.CalledProcessError:
            sys.exit(1)
```

I think if we're already using tempfile, we don't really need the explicit `os.remove` call once we're done with the check

vinceaperri: Done! There were issues with that before but I have fixed it with a f.flush().

toolkit/scripts/check_required_config_fmt.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python3
```
trungams, line 1:
```suggestion
#!/usr/bin/env python3
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.
```

toolkit/scripts/sort_required_configs.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
```
trungams, line 1:
```suggestion
#!/usr/bin/env python3
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.
```

trungams (approved):
LGTM. Just some minor additions.

PR #7490, opened 2024-01-26
SymCrypt provider initial integration

mamckee:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR adds the SymCrypt provider for OpenSSL 3. It will be used by default if present and enabled. 
 
###### Change Log  

<ins>openssl</ins> 
- Re-adds the FIPS mode flag check patch to OpenSSL and modifies the previously removed forced FIPS patch to load the SymCrypt provider by default. 
    - In non-FIPS modeThe SymCrypt provider will be loaded as the default if present, but may be disabled in favor of other providers.
    - In FIPS mode, the SymCrypt provider will always be implicitly loaded.

<ins>SymCrypt</ins>
- Updates SymCrypt to 103.4.1 which includes changes needed by the SymCrypt provider.

<ins>SymCrypt-OpenSSL</ins>
- Updates the SymCrypt-OpenSSL to include the new SymCrypt provider for OpenSSL 3. 
    - The SymCrypt provider replaces the SymCrypt engine for OpenSSL 1.1.1, since the engine model will be deprecated.

**YES/NO**

###### Test Methodology

- Local build


SPECS/openssl/0032-Force-fips.patch
```diff
@@ -0,0 +1,82 @@
+From 97edba1a35e1cc40b8b700aef75dd8904bcb2f33 Mon Sep 17 00:00:00 2001
+From: Maxwell Moyer-McKee <mamckee@microsoft.com>
+Date: Thu, 25 Jan 2024 16:48:47 -0800
+Subject: [PATCH] Load the symcrypt provider by default if fips flag or
```
dmcilvaney, line 4: What is the flow if the symcrypt packge is not installed, but fips is on? Out of scope for this PR but we have a validation tool for image configurations, and some documentation for FIPS, that might need to be updated if so.

SPECS/openssl/openssl.spec
```diff
@@ -9,7 +9,7 @@
 Summary: Utilities from the general purpose cryptography library with TLS implementation
 Name: openssl
 Version: 3.1.4
-Release: 1%{?dist}
+Release: 2%{?dist}
```
tobiasb-ms, line 12: Let's leave it at Release 1 for now -- we'll lock the release in later on, but this way we can work incrementally without resulting in a huge release number at ship date.

tobiasb-ms (approved):

SPECS/openssl/0032-Force-fips.patch
dmcilvaney: I suppose it won't be without precedent... the system won't even boot without `dracut-fips` installed if you turn on fips mode....

Can we add this package to `imageconfigvalidator.go:validatePackages()` so we at least yell about it missing if fips is configured in the image json?

@tobiasb-ms We will want to update compliance and documentation to reflect this, or do we want to add a Requires on symcrpt?

dmcilvaney: Had a chat offline, we can just ignore this for now and fix it later in the 3.0 release. Figure out what we want to do.

dmcilvaney (approved):

SPECS/SymCrypt-OpenSSL/SymCrypt-OpenSSL.spec
```diff
@@ -34,26 +34,41 @@ mkdir bin; cd bin
...
-%{_includedir}/scossl.h
+%{_libdir}/engines-3/symcryptengine.so
+%{_libdir}/ossl-modules/symcryptprovider.so
+%{_includedir}/e_scossl.h
+%{_sysconfdir}/pki/tls/symcrypt_prov.cnf
 
 %changelog
+* Thu Dec 28 2023 Maxwell Moyer-McKee <mamckee@microsoft.com> - 1.4.0-1
+- Update SymCrypt-OpenSSL to v1.4.0.
+- Adds SymCryptp-OpenSSL provider for OpenSSL 3.
```
samuel-lee-msft, line 70: NP typo: SymCryptp-OpenSSL -> SymCrypt-OpenSSL

SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -40,15 +40,20 @@ A core cryptographic library written by Microsoft
...
 # Create a symbolic link as if jitterentropy-library has been pulled in as git submodule
-rm -rf jitterentropy-library
-ln -s jitterentropy-library-3.3.1 jitterentropy-library
+rm -rf 3rdparty/jitterentropy-library
+ln -s ../jitterentropy-library-3.3.1 3rdparty/jitterentropy-library
 
 %build
+SYMCRYPT_BRANCH=main \
+SYMCRYPT_COMMIT_HASH=c55c670 \
+SYMCRYPT_COMMIT_TIMESTAMP=2023-12-01T15:59:37-08:00 \
```
samuel-lee-msft, line 49: I think the hash and timestamp should reflect the hash and timestamp of the last commit in SymCrypt 103.4.1 just for clarity when these strings are built into the library:
SYMCRYPT_COMMIT_HASH=a84ffe1
SYMCRYPT_COMMIT_TIMESTAMP=2024-01-26T22:00:47-08:00

mamckee: Thanks for the catch. I forgot to update this after the release was published.

PR #7492, opened 2024-01-26
Image Customizer: Add support for SELinux

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This change adds support for SELinux to the image customizer tool.

This includes:

- Enabling SELinux on a base image with SELinux disabled.
- Disabling SELinux on a base image with SELinux enabled.
- When SELinux is enabled, ensuring all files are labeled correctly.

###### Change Log  

- Image Customizer: Add support for SELinux

NO

###### Test Methodology

- Manual testing of image customizer tool.


toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -226,6 +226,67 @@ So, any existing ExtraCommandLine value in the base image will be replaced.
...
 to the existing `grub.cfg` file.
 
+### SELinux
+
+Specifies the mode to set SELinux to.
+
+If this field is not specified, then the existing SELinux mode in the base image is
+maintained.
+
+If SELinux is enabled, then all the file-systems that support SELinux will be have their
```
amritakohli, line 236: will have*

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -226,6 +226,67 @@ So, any existing ExtraCommandLine value in the base image will be replaced.
...
+```
+
+### SELinuxPolicy
+
+When SELinux is enabled, the Mariner Image Customizer tool ensures the image has the
+`selinux-policy` package installed.
+This package contains the default SELinux rules and is required for SELinux enabled
+images to be functional.
+
+However, some images may wish to use their own custom SELinux policy package.
```
amritakohli, line 273: I would recommend rephrasing such that the functionality is described first and reasoning explained later. As a user, when looking at docs I'd want to be able to read through as quickly as possible to get the info I need.
Eg: This property allows the user to override the name of the policy package... (followed by the explanation / reasoning)

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1619,48 +1624,49 @@ func ProvisionUserSSHCerts(installChroot safechroot.ChrootInterface, username st
...
 		mode = configuration.SELinuxPermissive.String()
 	}
 
-	selinuxConfigPath := filepath.Join(installChroot.RootDir(), configFile)
-	selinuxMode := fmt.Sprintf("SELINUX=%s", mode)
-	err = sed(selinuxPattern, selinuxMode, "`", selinuxConfigPath)
+	selinuxConfigPath := filepath.Join(installChroot.RootDir(), SELinuxConfigFile)
+	selinuxProperty := fmt.Sprintf("SELINUX=%s", mode)
+	err = sed(selinuxPattern, selinuxProperty, "`", selinuxConfigPath)
 	return
```
amritakohli, line 1663: return err* ?

amritakohli, line 1663: maybe avoid a naked return and explicitly "return err" for better readability? (especially for people with less golang expertise)

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -394,3 +402,72 @@ func addCustomizerRelease(imageChroot *safechroot.Chroot, toolVersion string, bu
...
+				stdout = strings.TrimSpace(stdout)
+				return fmt.Errorf("SELinux is enabled so the policy package must be installed:\n%s:\n%w", stdout, err)
+			}
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+
+		// Get the list of point points.
```
amritakohli, line 453: mount points*

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -226,6 +226,67 @@ So, any existing ExtraCommandLine value in the base image will be replaced.
...
+    SELinux: enforcing
+
+  PackagesInstall:
+  - selinux-policy
+  - selinux-policy-modules
+  - setools-console
+  - policycoreutils-python-utils
+```
+
+### SELinuxPolicy
```
romoh, line 266: Is that equivalent of removing the selinux-policy package and re-installing the new one?

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -394,3 +402,72 @@ func addCustomizerRelease(imageChroot *safechroot.Chroot, toolVersion string, bu
...
+
+		// Check if the selinux-policy package is installed.
+		if selinuxPolicyPackage == "" {
+			selinuxPolicyPackage = configuration.SELinuxPolicyDefault
+		}
+
+		err = imageChroot.UnsafeRun(func() error {
+			stdout, _, err := shell.Execute("rpm", "-q", selinuxPolicyPackage)
+			if err != nil {
+				// Note: RPM's stdout will say somelike like "package selinux-policy is not installed".
```
amritakohli, line 443: 1) if we require user to specify certain packages along with SELinux we should add it more explicitly in the docs. i do see it's part of the example but wouldn't hurt to be more clear what other things are required / conditions / recommendations with a certain property. eg im not clear if the 3 packages mentioned other than selinux-policy are requirements or recommendations.
2) is there a good reason not to install this package ourselves anyway if the user has SELinux property set but doesn't mention this package explicitly in PackagesInstall? like could it not be the default behavior?

toolkit/tools/imagecustomizer/docs/configuration.md
amritakohli: as per the current implementation it seems the selinux-policy package is not installed by default - the user has to specify it in PackagesInstall

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -226,6 +226,67 @@ So, any existing ExtraCommandLine value in the base image will be replaced.
...
+
+When SELinux is enabled, the Mariner Image Customizer tool ensures the image has the
+`selinux-policy` package installed.
+This package contains the default SELinux rules and is required for SELinux enabled
+images to be functional.
```
amritakohli, line 271: this should be included in the above property instead for better readability and easy comprehension of the features

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -226,6 +226,67 @@ So, any existing ExtraCommandLine value in the base image will be replaced.
...
+However, some images may wish to use their own custom SELinux policy package.
+Such images can override the name of policy package using the `SELinuxPolicy` property.
+
+Example:
+
+```yaml
+SystemConfig:
+  KernelCommandLine:
+    SELinux: enforcing
+    SELinuxPolicy: my-selinux-policy
```
gmileka, line 282: why have field to specify the package name if it needs to be listed under `PackagesInstall:`? (where as if it is the default, we don't specify `SELinuxPolicy` and rely only on the user listing it under `PackagesInstall`.
Does MIC need to be aware of which policy package is in use?

toolkit/tools/imagegen/configuration/kernelcommandline.go
```diff
@@ -11,6 +11,10 @@ import (
 	"strings"
 )
 
+const (
+	SELinuxPolicyDefault = "selinux-policy"
```
amritakohli, line 15: IMO: i feel like this variable doesn't really fit in this file / package and could be declared locally in handleSelinux function instead (since currently that's it's only purpose)

amritakohli, line 15: would this perhaps be a better fit in configuration/selinux.go?

toolkit/tools/imagecustomizer/docs/configuration.md
amritakohli: i agree, left a similar comment - this is a bit confusing and redundant
ideally, if user specifies selinux and no policy - we should install selinux-policy, or else the policy that is specified in selinuxpolicy. and for either case not have them (mandatorily) also mention it in PackagesInstall. 
(this could also be added in a future PR)

toolkit/tools/imagegen/installutils/installutils.go
cwize1: I agree that this error code pattern is a little confusing. But this pattern is used quite extensively throughout the toolkit's code. And if I am modifying an existing function, then I am inclined to leave the error code pattern alone. ("When in Rome" and all that.)

toolkit/tools/imagegen/configuration/kernelcommandline.go
cwize1: No `selinux.go` contains the SELinux mode (confusing called just `SELinux`), not SELinux stuff in general.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+	if selinuxArgsMatch == nil {
+		// SELinux args are not in the grub.cfg file.
+		if selinuxMode == imagecustomizerapi.SELinuxDisabled {
+			// Nothing to do.
+			return nil
+		}
+
+		insertLocationMatch := linuxCommandLineInsertRegex.FindStringIndex(grub2Config)
+		if insertLocationMatch == nil {
+			return fmt.Errorf("failed to find location to insert SELinux kernel command-line args")
```
gmileka, line 97: would it be helpful if we print out the regex here for the user? or an explanation of what we're looking for?
I'm thinking of a case where the user has some grub.cfg, and it fails and there is not information to fix the problem.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+		}
+
+		// Use the end of the regex match as the point to do the insert at.
+		insertPoint := insertLocationMatch[1]
```
gmileka, line 101: mm... in the future, not in this PR, should we consider writing a grub.cfg parser/dom?

toolkit/tools/imagegen/configuration/kernelcommandline.go
amritakohli: okay, and why does this belong in kernelcommandline?

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
gmileka: or we tokenize the kernel parameters in a tree-like structure...

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+	err := os.WriteFile(grub2ConfigFilePath, []byte(grub2Config), 0)
 	if err != nil {
-		return fmt.Errorf("failed to write new grub2 config file: %w", err)
+		return fmt.Errorf("failed to write new grub2 config file:\n%w", err)
 	}
 
 	return nil
 }
+
+func getGrub2ConfigFilePath(imageChroot *safechroot.Chroot) string {
```
gmileka, line 236: these grub helpers can be useful to others who deal with grub. Does it make sense to  move them into their own file/object? (I have similar needs with the iso generation code)

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: No. This property only affects the sanity check that ensures the policy package is installed. It doesn't actually install the package.

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: Fixed.

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: Fixed.

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Probably not. There is a lot of stuff in the grub.cfg file that I don't care about.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: I'm kind of hoping that the error never triggers for a user. But on the off chance it does, then the unique error message ensures that we can quickly diagnose the problem.

toolkit/tools/imagegen/configuration/kernelcommandline.go
cwize1: Because `kernelcommandline.go` has the `KernelCommandLine.SELinuxPolicy` property, which this constant is directly related to.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: I've added comments to the YAML example stating what the packages are used for.

cwize1: In terms of auto-installing the package, I am mostly just following what the imager tool does. That being said, using `PackagesInstall` does give the user more flexibility in terms of which version of the package is used.

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: I have changed the property to be a boolean that disables the check. This should hopefully make the config less confusing and redundant.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: All the code in `customizeboot.go` is related to the grub.cfg file. So, it is already separated out.

If you ever need these functions to live in a shared module, then we can cross that bridge when we come to it.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -394,3 +402,70 @@ func addCustomizerRelease(imageChroot *safechroot.Chroot, toolVersion string, bu
...
+		}
+	}
+
+	if selinuxMode != imagecustomizerapi.SELinuxDisabled {
+		imagerSELinuxMode, err := selinuxModeToImager(selinuxMode)
+		if err != nil {
+			return err
+		}
+
+		if !selinuxDisablePackageCheck {
```
amritakohli, line 435: wouldnt we still want to check that the user has installed some selinux policy package (when it's custom)?

maybe we could have 
SELinuxMode:
SELinuxPolicy: (when not specified, assume default) 
-- pretty similar to the original implementation

I guess the feedback / question earlier was that the policy could be installed based on the above fields... but as you mentioned PackagesInstall allows the user to specify the version they want which makes sense to me. So the previous implementation was ok / better imo

cwize1: The check if really just a sanity check for the user. If they are using their own custom SELinux policy, then we'll trust that they know what they are doing and get out of their way.

Checking that the policy package that the user specified is installed seems kind of pointless. If the user knows that package should be installed, then can just make sure it is installed.

gmileka (approved):

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -226,6 +226,80 @@ So, any existing ExtraCommandLine value in the base image will be replaced.
...
+Example:
+
+```yaml
+SystemConfig:
+  KernelCommandLine:
+    SELinux: enforcing
+
+  PackagesInstall:
+  # Required packages for SELinux.
+  - selinux-policy
```
romoh, line 273: Should MIC install them under the hoods?

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -89,6 +91,12 @@ func doCustomizations(buildDir string, baseConfigPath string, config *imagecusto
 		return fmt.Errorf("failed to add extra kernel command line: %w", err)
 	}
 
+	err = handleSELinux(config.SystemConfig.KernelCommandLine.SELinux,
```
romoh, line 94: should selinux `setfiles` be called as the last step after running scripts?

romoh (commented):
For users who want to hotfix the current selinux policies, is the recommendation to do that through post-install scripts? If so, do we need to relabel with updated policies?

I suggest adding explicit documentation for such scenario.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: No, this is actually one of the meaningful differences between PostInstallScripts and FinalizeImageScripts, and mirrors the imager's behavior.

cwize1:
> For users who want to hotfix the current selinux policies, is the recommendation to do that through post-install scripts? If so, do we need to relabel with updated policies?
> 
> I suggest adding explicit documentation for such scenario.

Fixed.

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: Nah, I rather the user be in control of this. e.g. using the existing version in the base image, updating the package to a new version, etc.

amritakohli (approved):

romoh (approved):

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -394,3 +402,70 @@ func addCustomizerRelease(imageChroot *safechroot.Chroot, toolVersion string, bu
...
+
+		if !selinuxDisablePackageCheck {
+			// Check if the selinux-policy package is installed.
+			err = imageChroot.UnsafeRun(func() error {
+				stdout, _, err := shell.Execute("rpm", "-q", configuration.SELinuxPolicyDefault)
+				if err != nil {
+					// Note: RPM's stdout will say somelike like "package selinux-policy is not installed".
+					stdout = strings.TrimSpace(stdout)
+					return fmt.Errorf("SELinux is enabled so the policy package must be installed:\n%s:\n%w", stdout, err)
+				}
```
PawelWMS, line 443: In case RPM fails for another reason, we may be losing useful debugging output if we don't capture `stderr`. In such cases the error message may be also misleading. I'd suggest making the log a bit more robust. One idea:
```suggestion
				stdout, stderr, err := shell.Execute("rpm", "-q", configuration.SELinuxPolicyDefault)
				if err != nil {
					// Note: RPM's stdout will say somelike like "package selinux-policy is not installed".
					stdout = strings.TrimSpace(stdout)
					stderr = strings.TrimSpace(stderr)
					return fmt.Errorf("failed while verifying if an SELinux policy package is installed:\nStdout: %s:\nStderr: %s\nError: %w", stdout, stderr, err)
				}
```

We could also go smarter and separate the "package X is not installed" from other errors but in that case I'd extract that logic to a separate function.

PawelWMS, line 443: Suggestion: it might be worth to extract bits of code into well-named functions, if you feel they are complex enough they require a comment. That way the comment is encapsulated in the function's name and the intention/flow of the program may be easier to follow.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -394,3 +402,70 @@ func addCustomizerRelease(imageChroot *safechroot.Chroot, toolVersion string, bu
...
+			switch mountPoint.GetTarget() {
+			case "/dev", "/proc", "/sys", "/run", "/dev/pts":
+				// Skip special directories.
+				continue
+			}
+
+			mountPointToFsTypeMap[mountPoint.GetTarget()] = mountPoint.GetFSType()
+		}
+
+		// Set the SELinux config file and relabel all the files.
```
PawelWMS, line 463: It feels like this comment is a bit redundant considering the function's documentation. I'd just like to avoid over-commenting the code and save it for complex bits, which are hard to understand. I fell frequent, short comments in between working code slowly make it harder to see what's happening inside the code itself.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -394,3 +402,70 @@ func addCustomizerRelease(imageChroot *safechroot.Chroot, toolVersion string, bu
...
+	if selinuxMode != imagecustomizerapi.SELinuxDisabled {
+		imagerSELinuxMode, err := selinuxModeToImager(selinuxMode)
+		if err != nil {
+			return err
+		}
+
+		if !selinuxDisablePackageCheck {
+			// Check if the selinux-policy package is installed.
+			err = imageChroot.UnsafeRun(func() error {
+				stdout, _, err := shell.Execute("rpm", "-q", configuration.SELinuxPolicyDefault)
```
PawelWMS, line 438: Is it possible that a different SELinux policy package may be installed instead of the default one?

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -4,9 +4,12 @@ The Mariner Image Customizer is configured using a YAML (or JSON) file.
 
 ### Operation ordering
 
-1. Override the `/etc/resolv.conf` file with the version from the host OS.
+1. If partitions were specified in the config, customize the disk partitions and reset
```
PawelWMS, line 7: Thanks for updating the docs!

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -226,6 +235,86 @@ So, any existing ExtraCommandLine value in the base image will be replaced.
...
+Note: For images with SELinux enabled, the `selinux-policy` package must be installed.
+This package contains the default SELinux rules and is required for SELinux-enabled
+images to be functional.
+The Mariner Image Customizer tool will report an error if the package is missing from
+the image.
+
+Note: If you wish to apply additional SELinux policies on top of the base SELinux
+policy, then it is recommended to apply these new policies using
+([PostInstallScripts](#postinstallscripts-script)).
+After applying the policues, you do not need to call `setfiles` manually since it will
```
PawelWMS, line 276: Typo:
```suggestion
After applying the policies, you do not need to call `setfiles` manually since it will
```

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -226,6 +235,86 @@ So, any existing ExtraCommandLine value in the base image will be replaced.
...
+  # Optional packages that contain useful SELinux utilities.
+  - setools-console
+  - policycoreutils-python-utils
+```
+
+### SELinuxDisablePackageCheck
+
+Disables the check that ensures the `selinux-policy` package is installed when SELinux
+is enabled.
+This is useful for images that want to use their own custom SELinux policy package.
```
PawelWMS, line 300: Question: by allow to disable the check is it possible to misconfigure the image by thinking SELinux is enabled but accidently skipping the check and end up with no policies installed? Do you think it would make sense to replace `SELinuxDisablePackageCheck` with `SELinuxPolicyPackage` and always check for the policy package when SELinux is enabled but when `SELinuxPolicyPackage` is not explicitly set, it defaults to `selinux-policy`?

Related question, since I don't know much about SELinux: is there another way to verify if any SELinux policy exists on the system (in case it's installed through adding single files or post-install scripts but not through any RPM package)?

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -50,6 +50,11 @@ const (
 	// /boot directory should be only accesible by root. The directories need the execute bit as well.
 	bootDirectoryFileMode = 0400
 	bootDirectoryDirMode  = 0700
+
+	CmdlineSELinuxSettings       = "security=selinux selinux=1"
+	CmdlineSELinuxForceEnforcing = CmdlineSELinuxSettings + " enforcing=1"
+
+	SELinuxConfigFile = "etc/selinux/config"
```
PawelWMS, line 57: Small ask: since we're here, could we split the `const` sections into two? One for public constants and one for private ones? We tried to do it in many other places in the code and it usually makes it easier to distinguish the what is internal-only and what is intended to be used by other packages.

PawelWMS, line 57: Could we add documentation comments to the public constants? Same as you did in `selinux.go`.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -140,6 +140,18 @@ func NewOverlayMountPoint(chrootDir, source, target, lowerDir, upperDir, workDir
...
+}
+
+func (m *MountPoint) GetFSType() string {
+	return m.fstype
+}
+
+func (m *MountPoint) GetTarget() string {
+	return m.target
+}
+
```
PawelWMS, line 154: They are trivial, but please still add comments for all public methods.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -670,3 +682,9 @@ func extractWorkerTar(chroot string, workerTar string) (err error) {
 	_, _, err = shell.Execute("tar", "-I", gzipTool, "-xf", workerTar, "-C", chroot)
 	return
 }
+
+func (c *Chroot) GetMountPoints() []*MountPoint {
```
PawelWMS, line 686: Same ask to document this method.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -670,3 +682,9 @@ func extractWorkerTar(chroot string, workerTar string) (err error) {
...
+	// Create a copy of the list so that the caller can't mess with the list.
```
PawelWMS, line 687: Same ask to document this method. This comment then can be included in that documentation, so the behaviour is discoverable even without entering this file.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -9,12 +9,20 @@ import (
...
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/imagegen/installutils"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
 	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safechroot"
 )
 
 var (
 	linuxCommandLineRegex = regexp.MustCompile(`\tlinux .* (\$kernelopts)`)
+
+	linuxCommandLineSELinuxRegex = regexp.MustCompile(`\tlinux .* (security=selinux selinux=1( enforcing=1)?) `)
```
PawelWMS, line 21: To be on the safe side, what do you think about this small modification:
```suggestion
	linuxCommandLineSELinuxRegex = regexp.MustCompile(`\tlinux .* (security=selinux\s+selinux=1(\s+enforcing=1)?) `)
```

PawelWMS: Also, I see there's a space at the end of the regex. I'm not an expert here but I want to double-check: is it guaranteed that space is going to be there (the arguments won't end yet)?

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -9,12 +9,20 @@ import (
...
+	linuxCommandLineInsertRegex  = regexp.MustCompile(`\tlinux \$bootprefix/\$mariner_linux `)
+
+	// Finds the SELinux mode line in the /etc/selinux/config file.
+	selinuxConfigModeRegex = regexp.MustCompile(`(?m)^SELINUX=(\w+)$`)
```
PawelWMS, line 25: Thank you for using `(?m)` and not looping line by line like we did in many other places.:)

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: Yes. And there is actually a user already doing this.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+	switch selinuxMode {
+	case imagecustomizerapi.SELinuxForceEnforcing:
+		newSELinuxArgs = installutils.CmdlineSELinuxForceEnforcing
+
+	case imagecustomizerapi.SELinuxPermissive, imagecustomizerapi.SELinuxEnforcing:
+		newSELinuxArgs = installutils.CmdlineSELinuxSettings
+
+	case imagecustomizerapi.SELinuxDisabled:
+		newSELinuxArgs = ""
+
```
PawelWMS, line 124: What do you think about moving the new SELinux constants and arguments used in this fragment into `selinux.go`, since we already have it? That way all SELinux stuff would be in one place. Similarly, this mapping here could be a public function inside `selinux.go`.

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: The `selinux-policy` package provides the `/etc/selinux/config` file. This file specifies what policy to use and is required for SELinux to work. So, I can check for existence of that file. This is probably a better solution than looking for an installed package.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: If the image was created by the imager and wasn't manually modified by the user, then the end space is guaranteed to be there. Adding the end space to the regex means I don't need to worry about adding extra spaces when inserting the new args.

cwize1: `\s` includes newlines, which I am not keen to include in the regex.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+		}
+
+		insertLocationMatch := linuxCommandLineInsertRegex.FindStringIndex(grub2Config)
+		if insertLocationMatch == nil {
+			return fmt.Errorf("failed to find location to insert SELinux kernel command-line args")
+		}
+
+		// Use the end of the regex match as the point to do the insert at.
+		insertPoint := insertLocationMatch[1]
+		start = insertPoint
```
PawelWMS, line 102: With the comment and the `insertLocationMatch` variable name, do we even need the separate `insertPoint`?
```suggestion
		start = insertLocationMatch[1]
```

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+
+		// -1 is a sneaky way of adding a whitespace character after the inserted SELinux args.
+		end = insertPoint - 1
```
PawelWMS, line 105: With the comment and the `insertLocationMatch` variable name, do we even need the separate `insertPoint`?
```suggestion
		start = insertLocationMatch[1]

		// -1 is a sneaky way of adding a whitespace character after the inserted SELinux args.
		end = start - 1
```

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+	} else {
+		// Get the range of the existing SELinux args.
```
PawelWMS, line 107: We could also mention that we get them in order to replace them/cut them out.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+	if selinuxArgsMatch == nil {
+		// SELinux not present in the kernel command-line.
+		// Therefore, it is disabled.
+		return imagecustomizerapi.SELinuxDisabled, nil
+	}
+
+	// Check if "enforcing=1" is present.
+	// Note: A negative index means the sub-match is not present.
+	// Note: regexp returns index pairs. So, [4] is start index of the 2nd sub-match.
+	if selinuxArgsMatch[4] >= 0 {
```
PawelWMS, line 160: Could we use well-named constants for these indexes to avoid "magic numbers"? For instance, take a look at `rpmSpecBuiltRPMRegex` and its corresponding `rpmSpecBuiltRPMRegex*Index` constants inside `rpmsnapshot.go`.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+
+	return nil
+}
+
+func getCurrentSELinuxMode(imageChroot *safechroot.Chroot) (imagecustomizerapi.SELinux, error) {
+	logger.Log.Debugf("Get existing SELinux mode")
+
+	grub2Config, err := readGrub2ConfigFile(imageChroot)
+	if err != nil {
+		return "", err
```
PawelWMS, line 146: Nit-pick: I'm aware it's just a string but since we already defined the `SELinux` type, could we use its named values instead of a raw empty string? For the sake of consistency and to be sure we always return an valid value (even if there's a good chance it won't be used).

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+	// Update grub.cfg file.
+	err = writeGrub2ConfigFile(newGrub2Config, imageChroot)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getCurrentSELinuxMode(imageChroot *safechroot.Chroot) (imagecustomizerapi.SELinux, error) {
```
PawelWMS, line 141: If there's more than one output variable, could we follow the use of named variables? That's what we tend to stick to in most of the code.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+	return selinuxMode, nil
+}
+
+func getSELinuxModeFromConfigFile(imageChroot *safechroot.Chroot) (imagecustomizerapi.SELinux, error) {
+	selinuxConfigFilePath := filepath.Join(imageChroot.RootDir(), installutils.SELinuxConfigFile)
+
+	// Read the SELinux config file.
+	fileBytes, err := os.ReadFile(selinuxConfigFilePath)
+	if err != nil {
+		return "", fmt.Errorf("failed to read existing grub2 config file:\n%w", err)
```
PawelWMS, line 178: It might be worth to include the file path in this error message as well.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+	}
+
+	selinuxConfig := string(fileBytes)
+
+	match := selinuxConfigModeRegex.FindStringSubmatch(selinuxConfig)
+	if match == nil {
+		return "", fmt.Errorf("failed to find SELinux mode in %s file", installutils.SELinuxConfigFile)
+	}
+
+	selinuxConfigMode := match[1]
```
PawelWMS, line 188: Similar as to use well-named constants for match indexes.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+
+	switch selinuxConfigMode {
+	case "enforcing":
+		return imagecustomizerapi.SELinuxEnforcing, nil
+
+	case "permissive":
+		return imagecustomizerapi.SELinuxPermissive, nil
+
+	case "disabled":
+		return imagecustomizerapi.SELinuxDisabled, nil
```
PawelWMS, line 198: Since we're reading `selinuxConfigMode` from the config file and that's already validated earlier if it contains only the valid values, do we need this mapping and check here? If we're doing this to be extra safe, could we at least use the public variables instead of raw strings. Think should have the same values, if I'm getting this right.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+	fileBytes, err := os.ReadFile(selinuxConfigFilePath)
+	if err != nil {
+		return "", fmt.Errorf("failed to read existing grub2 config file:\n%w", err)
+	}
+
+	selinuxConfig := string(fileBytes)
+
+	match := selinuxConfigModeRegex.FindStringSubmatch(selinuxConfig)
+	if match == nil {
+		return "", fmt.Errorf("failed to find SELinux mode in %s file", installutils.SELinuxConfigFile)
```
PawelWMS, line 185: Could we stick to one log format as much as possible? I'm personally a fan of surrounding potentially-empty strings with brackets (like you a few lines below) or quotes, so empty strings don't print confusing logs.
```suggestion
		return "", fmt.Errorf("failed to find SELinux mode in (%s) file", installutils.SELinuxConfigFile)
```

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -57,10 +60,179 @@ func handleKernelCommandLine(extraCommandLine string, imageChroot *safechroot.Ch
...
+
+func readGrub2ConfigFile(imageChroot *safechroot.Chroot) (string, error) {
+	logger.Log.Debugf("Reading grub.cfg file")
+
+	grub2ConfigFilePath := getGrub2ConfigFilePath(imageChroot)
+
+	// Read the existing grub.cfg file.
+	grub2ConfigFileBytes, err := os.ReadFile(grub2ConfigFilePath)
+	if err != nil {
+		return "", fmt.Errorf("failed to read existing grub2 config file:\n%w", err)
```
PawelWMS, line 214: Might be worth to mention the path we tried to read.

toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
```diff
@@ -21,14 +21,20 @@ func customizePartitionsUsingFileCopy(buildDir string, baseConfigPath string, co
 	}
 	defer existingImageConnection.Close()
 
+	existingSELinuxMode, err := getCurrentSELinuxMode(existingImageConnection.Chroot())
```
PawelWMS, line 24: Nit-pick just for the sake of consistency: since we called the function `getCurrentSELinuxMode`, it might be worth to stick to this naming for the variable as well.
```suggestion
	currentSELinuxMode, err := getCurrentSELinuxMode(existingImageConnection.Chroot())
```

Or swap it the other way around.:)

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: This is reading from the `/etc/selinux/config` file, not the config file provided by the user to the tool.

The names used in the API do match the names used in the `/etc/selinux/config` file for most values. But there are some exceptions. e.g. `"force-enforcing"` is not a value that is supported by `/etc/selinux/config`.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -394,3 +402,70 @@ func addCustomizerRelease(imageChroot *safechroot.Chroot, toolVersion string, bu
...
+	// the SELinux args will already be correct and don't need to be updated.
+	if !partitionsCustomized {
+		// Update the SELinux kernel command-line args.
+		err := updateSELinuxCommandLine(selinuxMode, imageChroot)
+		if err != nil {
+			return fmt.Errorf("failed to update SELinux args in grub.cfg:\n%w", err)
+		}
+	}
+
+	if selinuxMode != imagecustomizerapi.SELinuxDisabled {
```
PawelWMS, line 429: The code behind this check is rather lengthy and has a few nested `if` statements as well. Could we:
1. Leave early to reduce the number of indents?
2. Extract the code behind this check into another function?
```suggestion
	if selinuxMode == imagecustomizerapi.SELinuxDisabled {
		return nil
	}
```

I'm mostly interested in the second bit to be honest. If the check stays the way it is but we have that separate function, then it will get cleaner anyway.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Hmm. `""` is a lot shorted than `SELinuxDefault`. This reduces the visual noise and IMHO makes it easier to read the error.

toolkit/tools/pkg/imagecustomizerlib/typeConversion.go
```diff
@@ -158,14 +158,51 @@ func mountIdentifierTypeToImager(mountIdentifierType imagecustomizerapi.MountIde
...
+	existingSELinuxMode imagecustomizerapi.SELinux,
+) (configuration.SELinux, error) {
+	if selinuxMode == imagecustomizerapi.SELinuxDefault {
+		selinuxMode = existingSELinuxMode
+	}
+
+	return selinuxModeToImager(selinuxMode)
+}
+
+func selinuxModeToImager(selinuxMode imagecustomizerapi.SELinux) (configuration.SELinux, error) {
```
PawelWMS, line 191: Suggestion: the `selinuxModeMaybeDefaultToImager` name feels a bit unclear. Since the main point is to get the `ToImager` value, what do you think about just moving the `if` into `selinuxModeToImager` and dropping this additional function?
```suggestion
func selinuxModeToImager(selinuxMode imagecustomizerapi.SELinux,
	existingSELinuxMode imagecustomizerapi.SELinux) (configuration.SELinux, error) {
	if selinuxMode == imagecustomizerapi.SELinuxDefault {
		return existingSELinuxMode, nil
	}
```

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
PawelWMS: Hmmm. In many places in the code we also follow the approach of named output, where you can then use a simple `return` without additional arguments. That might also help here to reduce the repetition completely.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: In the imagecustomizer code, we generally haven't been using named output variables, mainly because I prefer this style instead.

toolkit/tools/pkg/imagecustomizerlib/typeConversion.go
cwize1: `selinuxModeToImager` is used directly by the `handleSELinux()` function. In `handleSELinux()` function, the default value is already resolved. So, the `existingSELinuxMode` would just a dummy value, which I didn't like. Hence, why I split the functions.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: moot

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: fixed

toolkit/tools/imagegen/installutils/installutils.go
cwize1: fixed

toolkit/tools/internal/safechroot/safechroot.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: fixed

toolkit/tools/imagegen/installutils/installutils.go
cwize1: fixed

toolkit/tools/internal/safechroot/safechroot.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: I created a new set of constants and used them in the imager tool as well.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: The code is read in a lot of different contexts, some of which might not have a functioning IntelliSense IDE. So, I prefer to target the lowest common denominator.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: fixed

toolkit/tools/internal/file/file.go
```diff
@@ -86,6 +86,15 @@ func CopyAndChangeMode(src, dst string, dirmode os.FileMode, filemode os.FileMod
 	return copyWithPermissions(src, dst, dirmode, true, filemode)
 }
 
+// Write writes a string to the file dst.
```
PawelWMS, line 89: Forgot to update the comment:
```suggestion
// Read reads a string from the file src.
```

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -37,7 +37,27 @@ const (
...
+	// File path of the SELinux config file.
+	SELinuxConfigFile = "/etc/selinux/config"
+
+	// The valid string values for the "SELINUX" field in the /etc/selinux/config file.
+	SELinuxConfigEnforcing  = "enforcing"
+	SELinuxConfigPermissive = "permissive"
+	SELinuxConfigDisabled   = "disabled"
+
+	GrubCfgFile = "/boot/grub2/grub.cfg"
+	GrubDefFile = "/etc/default/grub"
```
PawelWMS, line 57: Thanks for organizing the constants! Could we add documentation for the `GrubCfg*` ones and start each documentation line with the name of the constant (similar to `NullDevice`)? We're tryint to follow [Golang's docs](https://go.dev/blog/godoc) about commenting the code.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1717,6 +1736,15 @@ func selinuxRelabelFiles(systemConfig configuration.SystemConfig, installChroot
...
 
 		err = installChroot.UnsafeRun(func() error {
+			runtime.LockOSThread()
+			defer runtime.UnlockOSThread()
+
+			err := setSELinuxExecContext("system_u:system_r:setfiles_mac_t:s0")
+			if err != nil {
+				return fmt.Errorf("failed to set SELinux exec context:\n%w", err)
+			}
+			defer resetSELinuxExecContext()
```
PawelWMS, line 1746: Could we add a comment about what's happening here? I'm guessing it may be obvious for people familiar with SELinux, but adding something for other folks may help.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
PawelWMS: Got it. In that case could we add a comment explaining why we guarantee the spaces?

And you're right about `\s`. I think `[[:blank:]]` might work better then but I have no strong need to modify this regex.

PawelWMS (approved):

toolkit/tools/imagegen/installutils/installutils.go
cwize1: Sorry, didn't mean to push this.

(This is hack to allow the code to run on Fedora. But it isn't applicable to Ubuntu. I need to figure out how to detect which distro we are running on in a sensible way.)

cwize1: fixed

toolkit/tools/imagegen/installutils/installutils.go
cwize1: fixed

toolkit/tools/internal/file/file.go
cwize1: fixed

PR #7495, opened 2024-01-26
Upgraded `ruby` to 3.3.0.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Upgraded `ruby` to 3.3.0. Also switching back to using the bigdecimal gem from the default gems in `ruby` instead of using a separate spec as we now have the ability to do so. In 2.0 we had a separate spec for bigdecimal with a version higher than what the default gem offered, so we had to keep maintaining it. With this update, this is no longer necessary. Thanks, @neha170 for pointing this out!

The `racc` gem has been removed from the default Ruby packages but still exists as a standalone. It is a `a LALR(1) parser generator`, which is a feature unlikely used by Azure Linux users, so I decided to not pro-actively re-add it as a separate package and take the responsibility of adding it in future in case I was wrong.

###### Change Log  

- Bumped `ruby` to version 3.3.0.
- Removed the standalone `rubygem-bigdecimal` package in favour of using the default one provided by the `ruby` package.

No.

###### Test Methodology

- Local build with tests.
- [Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=491893&view=results).


SPECS/rubygem-bigdecimal/rubygem-bigdecimal.spec
```diff
@@ -1,19 +1,19 @@
-%global debug_package %{nil}
 %global gem_name bigdecimal
+
 Summary:        BigDecimal provides arbitrary-precision floating point decimal arithmetic
-Name:           rubygem-bigdecimal
-Version:        3.1.2
```
neha170, line 5: Any specific reason we're upgrading to v3.1.6? Alternatively, we can get rid of the separate spec, and get v3.1.5 with ruby v3.3.0

PawelWMS: We want the latest version for 3.0 and that's just what's latest at this moment. At least that's how I see our 3.0 effort.:)

anphel31 (approved):

neha170 (approved):
Looks good to me.
Just fyi, we are removing rubygem-racc and rubygem-readline-ext as they are no longer being provided as 'default' gems with the updated ruby v3.3.0. If there is anyone using these gems already, this change might break their setup.

PawelWMS:
> Looks good to me. Just fyi, we are removing rubygem-racc and rubygem-readline-ext as they are no longer being provided as 'default' gems with the updated ruby v3.3.0. If there is anyone using these gems already, this change might break their setup.

You're right, @neha170. The `rubygem-readline-ext` gem was deprecated, so it's just not there in the new versions of Ruby.

When it comes to `racc`, it is a `a LALR(1) parser generator`, which is a feature unlikely used by Azure Linux users, so I decided to not pro-actively re-add it as a separate package and take the responsibility of adding it in future in case I was wrong. I should have, however, mentioned that in the summary (now updated).

PR #7512, opened 2024-01-27
Upgrade ksh: 1.0.0~beta.1 -> 1.0.8

BettyRain:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Upgrade ksh: 1.0.0~beta.1 -> 1.0.8 and make necessary changes. Remove the patches since they are not needed anymore.
Disabled pty tests (caused whole build to freeze), other tests are enabled.

**NO**

###### Test Methodology

- Local build
- Pipeline build id: [499592](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=499592&view=results)


SPECS/ksh/ksh.spec
```diff
@@ -37,6 +30,9 @@ Requires(post): grep
 Requires(post): systemd
 
 Requires(postun): sed
+Requires(rpmlib): rpmlib(CompressedFileNames) <= 3.0.4-1
+Requires(preun): /bin/sh
+Requires(verify): /bin/sh
```
mfrw, line 35: I do not think we need to specify these Requires.

BettyRain: Working on the tests in this package, will put this PR into draft

SPECS/ksh/ksh.spec
```diff
@@ -1,51 +1,52 @@
...
 BuildRequires:  gcc
 # regression test suite uses 'ps' from procps
 BuildRequires:  procps
 
+# regression test suite requirements
+BuildRequires: glibc-lang
+BuildRequires: ncurses
+BuildRequires: procps
+BuildRequires: tzdata
+BuildRequires: util-linux
```
mfrw, line 25: should we move these in the below block which gates the `check` dependencies :)

BettyRain: I removed them fully, we don't need them at all

mfrw (approved):
LGTM :)

SPECS/ksh/ksh.spec
```diff
@@ -32,20 +25,20 @@ BuildRequires:  sudo
 Requires:       coreutils
 Requires:       diffutils
 
-Requires(post): coreutils
-Requires(post): grep
-Requires(post): systemd
-
+Requires(post): grep, coreutils, systemd
```
PawelWMS, line 28: Please put a single dependency per line.

SPECS/ksh/ksh.spec
```diff
@@ -1,23 +1,16 @@
-%global       verBetaPrefix 1.0.0
-%global       verBetaSuffix 1
-%global       verBetaFull %{verBetaPrefix}-beta.%{verBetaSuffix}
-
+%global debug_package %{nil}
```
PawelWMS, line 1: This looks fishy. Why do we want to disable the debuginfo package now?

SPECS/ksh/ksh.spec
```diff
@@ -1,23 +1,16 @@
...
-Version:        %{verBetaPrefix}~beta.%{verBetaSuffix}
-Release:        5%{?dist}
+Version:        1.0.8
+Release:        1%{?dist}
 License:        EPL-1.0
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 URL:            http://www.kornshell.com/
-Source0:        https://github.com/ksh93/%{name}/archive/v%{verBetaFull}/%{name}-%{verBetaFull}.tar.gz
+Source0:        https://github.com/ksh93/%{name}/archive/refs/tags/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
```
PawelWMS, line 10: Thank you for updating this line.:)

SPECS/ksh/ksh.spec
```diff
@@ -32,20 +25,20 @@ BuildRequires:  sudo
...
+
 %description
 KSH-93 is the most recent version of the KornShell by David Korn of
 AT&T Bell Laboratories.
 KornShell is a shell programming language, which is upward compatible
 with "sh" (the Bourne Shell).
 
 %prep
-%autosetup -n %{name}-%{verBetaFull} -p1
+%autosetup -n %{name}-%{version}
```
PawelWMS, line 41: Not needed anymore. The default is `%{name}-%{version}` anyway.
```suggestion
%autosetup
```

SPECS/ksh/ksh.spec
```diff
@@ -80,12 +73,12 @@ touch %{buildroot}%{_bindir}/rksh
...
-useradd test -G root -m
-
-# Disabling tests as they tend to freez and the package is low-pri at the moment.
-false && sudo -u test ./bin/shtests --compile
+# disable pty tests as they tend to freeze
+shopt -s globstar; shopt -s nullglob; for test in bin/tests/**/*.sh; do
+  if [[ $test != *"pty.sh" ]]; then
+    $test || exit 1;
+  fi;
+done
```
PawelWMS, line 81: This looks a bit hacky. I think we're relying on finding and running the tests ourselves instead of the way upstream suggests running them. This is prone to break during version updates or we may miss a critical prep step for the tests.  And it looks like they do quite a few things inside `downloaded_sources/ksh-1.0.8/src/cmd/ksh93/tests/shtests`. Would looking for and removing the `pty.sh` work? It still won't resolve the potential break after a version upgrade but seems a bit more reliable and we can still run upstream's way of starting tests.

SPECS/ksh/ksh.spec
```diff
@@ -80,12 +73,12 @@ touch %{buildroot}%{_bindir}/rksh
 touch %{buildroot}%{_mandir}/man1/rksh.1.gz
 
 %check
-# We run more tests as non-root user
-chmod g+w . -R
-useradd test -G root -m
-
-# Disabling tests as they tend to freez and the package is low-pri at the moment.
-false && sudo -u test ./bin/shtests --compile
```
PawelWMS, line 88: Can we still run the test as a non-root user? Or are all tests now running even if it's from a root account?

SPECS/ksh/ksh.spec
BettyRain: I added this line to fix the build failure:

In SRPM:
time="2024-02-01T08:51:28Z" level=debug msg="Processing files: ksh-debuginfo-1.0.8-1.azl3.x86_64"
time="2024-02-01T08:51:28Z" level=debug msg="error: Empty %files file /usr/src/mariner/BUILD/ksh-1.0.8/debugfiles.list"

In test:
time="2024-02-01T08:51:32Z" level=debug msg="RPM build errors:"
time="2024-02-01T08:51:32Z" level=debug msg="error: Empty %files file /usr/src/mariner/BUILD/ksh-1.0.8/debugfiles.list"

SPECS/ksh/ksh.spec
```diff
@@ -80,12 +71,14 @@ touch %{buildroot}%{_bindir}/rksh
...
-# We run more tests as non-root user
-chmod g+w . -R
 useradd test -G root -m
 
-# Disabling tests as they tend to freez and the package is low-pri at the moment.
-false && sudo -u test ./bin/shtests --compile
+# disable pty tests as they tend to freeze
+shopt -s globstar; shopt -s nullglob; for test in bin/tests/**/*.sh; do
+  if [[ $test != *"pty.sh" ]]; then
+    $test || exit 1;
```
PawelWMS, line 79: We should never `exit` out of the `%check` section. This breaks the whole package build and not just indicate a failure of the tests. The way test failures are identified is the return value of the last command. If there's more than one test to run, you could for instance have a helper `tests_ok` variable, which you set to `false` whenever we detect a failing test and as the last line in the `%check` section to `$tests_ok`. Example in `edk2.spec`.

Having said that, please see my comment below about using upstream's way of running tests, which may make this comment irrelevant.

PawelWMS (changes requested):
A few comments to address in the `%check` section but a great job on re-enabling the tests!

SPECS/ksh/ksh.spec
PawelWMS: Got it. Could you in that case add include that in a comment above this line for future refernce?

SPECS/ksh/ksh.spec
```diff
@@ -80,12 +74,10 @@ touch %{buildroot}%{_bindir}/rksh
...
 %check
-# We run more tests as non-root user
-chmod g+w . -R
 useradd test -G root -m
-
-# Disabling tests as they tend to freez and the package is low-pri at the moment.
-false && sudo -u test ./bin/shtests --compile
+# disable pty tests as they tend to freeze
+find src/cmd/ksh93/tests -name "pty.sh" -delete
+bin/package test
```
PawelWMS, line 80: This is looking nice. But can we still run the tests as non-root user? I believe that allowed more tests to run.
```suggestion
sudo -u test bin/package test
```

If that's no longer the case, please ignore this comment.;)

PawelWMS (approved):
Just one more comment to try and run the tests as a non-root user. But already approving.

PR #7524, opened 2024-01-29
Fix coredns CVE-2023-44487 by patching vendor tar

dmcilvaney:
- Replace all uses of golang.org/x/net with version v0.17.0 in go.mod

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

`coredns` still has vendored code that is vulnerable to CVE-2023-44487 (`golang.org/x/net/http2`). The version requirement is coming from a 3rd order dependency (`github.com/openzipkin/zipkin-go v0.4.2` -> `github.com/IBM/sarama v1.40.1` -> `golang.org/x/net v0.12.0` for example). There were multiple conflicting requirements for the version of `net`, so a blanket module upgrade is used. (I'm hoping the `golang.org/x/net` module will generally be fairly stable and backwards compatible which may not be the case for the intermediate modules).

`IBM/sarama` does have a version that updates to use updated `x/net` ([version v1.41.3](https://github.com/IBM/sarama/blob/c42b2e0ed061fbdbade7fcb86ddf88a000a7bc4d/go.mod#L19)) however `zipkin` does not have [any version which consumes this updated version](https://github.com/openzipkin/zipkin-go/blob/e84b2cf6d2d915fe0ee57c2dc4d736ec13a2ef6a/go.mod#L6).

###### Change Log  

- Add `replace golang.org/x/net => golang.org/x/net v0.17.0` to `coredns`'s `go.mod` file.
- Add `%check` section to `coredns` (it only covers go unit tests, not true e2e testing).

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/48735735

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2023-44487

###### Test Methodology

- local build
- Buddy Build TBD

### DIFF
--- a/SPECS/coredns/coredns.spec
+++ b/SPECS/coredns/coredns.spec
@@ -54,12 +54,13 @@ make
 
 %check
 # From go.test.yml
-go install github.com/fatih/faillint@latest
-go test -v -race request/...
-go test -v -race core/...
-go test -v -race coremain/...
-go test -v -race plugin/...
-go test -v -race test/...
+go install github.com/fatih/faillint@latest && \
+(cd request && go test -v -race ./...) && \
+(cd core && go test -v -race ./...) && \
+(cd coremain && go test -v -race ./...) && \
+(cd plugin && go test -v -race ./...) && \
+(cd test && go test -v -race ./...) && \
+./coredns -version
 
 %install
 install -m 755 -d %{buildroot}%{_bindir}
--- a/SPECS/coredns/coredns.spec
+++ b/SPECS/coredns/coredns.spec
@@ -17,8 +17,8 @@ Source0:        %{name}-%{version}.tar.gz
 #   1. wget https://github.com/coredns/coredns/archive/v%%{version}.tar.gz -O %%{name}-%%{version}.tar.gz
 #   2. tar -xf %%{name}-%%{version}.tar.gz
 #   3. cd %%{name}-%%{version}
-#   5. go mod vendor
-#   6. tar  --sort=name \
+#   4. go mod vendor
+#   5. tar  --sort=name \
 #           --mtime="2021-04-26 00:00Z" \
 #           --owner=0 --group=0 --numeric-owner \
 #           --pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime \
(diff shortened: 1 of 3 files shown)


dmcilvaney:
Waiting on local build before uploading sources

SPECS/coredns/coredns.spec
```diff
@@ -17,8 +17,11 @@ Source0:        %{name}-%{version}.tar.gz
 #   1. wget https://github.com/coredns/coredns/archive/v%%{version}.tar.gz -O %%{name}-%%{version}.tar.gz
 #   2. tar -xf %%{name}-%%{version}.tar.gz
 #   3. cd %%{name}-%%{version}
-#   4. go mod vendor
-#   5. tar  --sort=name \
+#  >>>> Required for Patch1: CVE-2023-44487.patch
+#   4. patch -p1 CVE-2023-44487.patch
+#  >>>>
+#   5. go mod vendor
+#   6. tar  --sort=name \
```
PawelWMS, line 24: I think the `generate_source_tarball.sh` script is taking care of these steps now. I think we can just remove this whole comment and add the lines you've added here somewhere inside that script.

Also, I think we actually need to update the script to work with our automation.

SPECS/coredns/coredns.spec
```diff
@@ -48,6 +52,16 @@ export BUILDOPTS="-mod=vendor -v"
...
 
+%check
+# From go.test.yml
+go install github.com/fatih/faillint@latest && \
+(cd request && go test -v -race ./...) && \
+(cd core && go test -v -race ./...) && \
+(cd coremain && go test -v -race ./...) && \
+(cd plugin && go test -v -race ./...) && \
+(cd test && go test -v -race ./...) && \
+./coredns -version
```
PawelWMS, line 63: Cool, thanks for adding tests!

PawelWMS (commented):
We should update the `generate_source_tarball.sh` script as well before we merge it, so that our automation don't cut out the patch during the next update.

dmcilvaney:
Thinking about this more last night, this might be a repeated task across several golang packages. I'm wondering if it wouldn't be better to use a patch file to surgically target the affected http2 module instead. It would mean we don't need to update the vendor tars for each package at the cost of slightly more cumbersome spec file patching.

PawelWMS:
> Thinking about this more last night, this might be a repeated task across several golang packages. I'm wondering if it wouldn't be better to use a patch file to surgically target the affected http2 module instead. It would mean we don't need to update the vendor tars for each package at the cost of slightly more cumbersome spec file patching.

Yeah, I think that sounds better than having to patch the source before we upload it to our blobstore. Like you said, this way we can keep the same vendor tarball, but it's also in line with how we patch the upstream source and makes it easier to see, which patches were already applied. Btw. I see there's an example of what you're suggesting in #7542 now.

PawelWMS (approved):

dmcilvaney:
Rebased with latest

PR #7585, opened 2024-01-31
systemd: do not conflict with polkit dir

ddstreetmicrosoft:
This is needed since currently the `systemd` and `polkit` packages cannot be installed together on a system.

Also `polkit` FTBFS because it is not packaging one of the built libs.

### DIFF
--- a/SPECS/systemd/systemd.spec
+++ b/SPECS/systemd/systemd.spec
@@ -2,7 +2,7 @@
 Summary:        Systemd
 Name:           systemd
 Version:        254.5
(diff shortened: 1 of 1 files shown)


rlmenge (approved):

tobiasb-ms (approved):

ddstreetmicrosoft:
> nit: Expand the pull request description to include why this is useful/important (and maybe get rid of some of the template text).

ack, PR desc updated, thanks

christopherco (approved):

PR #7728, opened 2024-02-06
graphviz: add build dep gc-devel

ddstreetmicrosoft:
add missing build dep so the package will build


ddstreetmicrosoft:
> Looks like the spec file check is complaining about `%patchN` being used. Could you update the spec file to preferably use `%autosetup`? Otherwise the change looks good to me.

done, can you add a review? thanks!

trungams (approved):

PR #7744, opened 2024-02-07
[MIC ISO generation] Allow adding kernel parameters to the LiveOS iso grub.cfg.

gmileka:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Allow adding kernel parameters to the LiveOS iso grub.cfg.

This is needed to enable the cloud-init scenario where the data files are embedded on the iso media.

###### Change Log  

- [Allow adding kernel parameters to the LiveOS iso grub.cfg.](https://github.com/microsoft/CBL-Mariner/commit/df1c154d8b1466337bbd187973ea63018ed6fce7)

**NO**

###### Associated issues  

- [6781: Add cloud-init support for LiveOS ISOs](https://dev.azure.com/mariner-org/ECF/_workitems/edit/6781)

###### Links to CVEs  
- n/a

###### Test Methodology

- Local build of two configurations:
  - cloud-init data files on iso file system.
  - cloud-init data files on root file system.


toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -243,6 +246,45 @@ func (b *LiveOSIsoBuilder) prepareRootfsForDracut(writeableRootfsDir string) err
...
+	for _, field := range fields {
+		// If a field starts with 'ds=nocloud;s=' (for cloud-init), it means
+		// the user did not wrap it in single quotes as it should be. If left
+		// as-is, the grub parsing at runtime will drop the text after the ';'.
+		//
+		// Unfortunately, if the user adds single quotes, the higher-level
+		// toolkit and MIC schema validation fail because the Mariner Toolkit
+		// does not allow it.
+		//
+		// This Toolkit limitation is because it uses 'sed' to manipulate
```
cwize1, line 273: The escape character used with `sed` is the back-tick not single-quotes.

Also, the use of `sed` is causing problems then feel free to replace it with Go's regexp library. (Doing this is goodness regardless as we should avoid dependencies on external tools as much as possible.)

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -243,6 +246,45 @@ func (b *LiveOSIsoBuilder) prepareRootfsForDracut(writeableRootfsDir string) err
...
+		// grub.cfg and 'sed' relies on single quotes to parse its input
+		// expression. So, include a single quote will break the sed input
+		// parsing.
+		//
+		// The work around below is to restore the needed single quotes.
+		if strings.HasPrefix(field, "ds=nocloud;s=") {
```
cwize1, line 279: MIC is going to support strong back-compat. So, hacks like this aren't permitted.

The BASH/grub rules for parsing single quotes are actually very straightforward since there aren't any escape characters inside single-quote strings. So, it shouldn't be too difficult to expand the validity checking to correctly handle single-quotes.

toolkit/tools/imagecustomizerapi/iso.go
```diff
@@ -9,13 +9,24 @@ import (
 
 // Iso defines how the generated iso media should be configured.
 type Iso struct {
-	AdditionalFiles AdditionalFilesMap `yaml:"AdditionalFiles"`
+	KernelCommandLine KernelCommandLine  `yaml:"KernelCommandLine"`
```
cwize1, line 12: You don't seem to be supporting other parts of the `KernelCommandLine` type. So, you probably shouldn't be using the `KernelCommandLine` type here.

Feel free to place `ExtraCommandLine` directly under `Iso`. (Perhaps rename to `ExtraKernelCommandLine`?)

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: for some reason, i misread the sed character to be the single quote instead of the back-tick here:
https://github.com/microsoft/CBL-Mariner/blob/14f37351044fc9b943c3b9660957f3f97e4530c8/toolkit/tools/imagecustomizerapi/kernelcommandline.go#L33
Will revise...

gmileka:
I misread the sed character to be the single quote instead of the back-tick here:
https://github.com/microsoft/CBL-Mariner/blob/14f37351044fc9b943c3b9660957f3f97e4530c8/toolkit/tools/imagecustomizerapi/kernelcommandline.go#L33
Will revise...

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: updated the logic.

toolkit/tools/imagecustomizerapi/iso.go
gmileka: Placed it directly...
Can selilnux be supported for the iso file system though? If so, to not break schema compatibility, we'd need to use the `KernelCommandLine` and error out if somebody tries to use selinux while it is not implemented yet.

toolkit/tools/imagecustomizerapi/kernelextraparameters.go
```diff
@@ -0,0 +1,106 @@
...
+// if escapedCharacters is not empty, it means escape only if the next
+// character is one of escapedCharacters.
+func processEscapedCharacter(text string, start int, count int, escapedCharacters string) (lastProcessed int, err error) {
+	i := start + 1
+	if i < count {
+		if len(escapedCharacters) == 0 || bytes.IndexByte([]byte(escapedCharacters), text[i]) != -1 {
+			// character is meant to be escaped
+			return i, nil
+		} else {
+			// character is not meant to be escaped
```
cwize1, line 31: You might want to add a comment here explaining what this is doing. I think this is related to Bash's rule where in double-quoted strings, characters after a backslash that aren't `$`, back-tick, `"`, `\`, or newline don't result in the backslash being removed from the string.

cwize1 (approved):

toolkit/tools/imagecustomizerapi/iso.go
gmileka: Decided to report an error when selinux is specified under the iso configuration. That way, we avoid creating a new type, and then in the future, should we enable selinux for iso image, we have a consistent kernel schema.

toolkit/tools/imagecustomizerapi/kernelextraparameters.go
```diff
@@ -0,0 +1,106 @@
+package imagecustomizerapi
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+)
+
+// KernelExtraParameters defines one or more extra kernel parameters.
+type KernelExtraParameters string
```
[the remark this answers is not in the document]
gmileka: renamed across.

toolkit/tools/imagecustomizerapi/kernelextraparameters.go
gmileka: correct - that's how grub defined it too.
added comment.

PR #7752, opened 2024-02-07
Fix pytest version for python-virtualenv

corvus-callidus:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

The python-virtualenv package pip installs pytest for the check section. The latest release of pytest (v8.0.0) contains a breaking change, treating use of certain deprecated parameters as errors. This caused failures in the python-virtualenv tests since it's an older version which predates the API changes to pytest. To fix the tests, this PR pins the pytest version to `<8`, ensuring we use a compatible version of pytest.

###### Change Log  

- Added patch to pin pytest version to `<8`
- Fixed check section macro

NO

###### Test Methodology

- Local build
- Pipeline build id: [501105](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=501105&view=results)

### DIFF
--- a/SPECS/python-virtualenv/pin-pytest-version.patch
+++ b/SPECS/python-virtualenv/pin-pytest-version.patch
@@ -1,0 +1,28 @@
+From 40d446bf7f2c3ee200599ec8675a2b57ed024909 Mon Sep 17 00:00:00 2001
+From: corvus-callidus <108946721+corvus-callidus@users.noreply.github.com>
+Date: Wed, 7 Feb 2024 17:12:27 +0000
+Subject: [PATCH] Restrict pytest versions for compatibility
+
+Version 20.14.0 of virtualenv uses now-deprecated pytest APIs. Version 8 of
+pytest now treats the deprecated APIs as errors, so we need to restrict the
+version of pytest for compatibility with this older version of virtualenv.
+
+---
+ setup.cfg | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/setup.cfg b/setup.cfg
+index 991979b..0fbfac7 100644
+--- a/setup.cfg
++++ b/setup.cfg
+@@ -96,7 +96,7 @@ testing =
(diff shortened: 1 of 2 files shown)


SPECS/python-virtualenv/pin-pytest-version.patch
```diff
@@ -0,0 +1,28 @@
...
+diff --git a/setup.cfg b/setup.cfg
+index 991979b..0fbfac7 100644
+--- a/setup.cfg
++++ b/setup.cfg
+@@ -96,7 +96,7 @@ testing =
+ 	coverage>=4
+ 	coverage-enable-subprocess>=1
+ 	flaky>=3
+-	pytest>=4
++	pytest>=4,<8
```
ddstreetmicrosoft, line 23: i don't think we actually *want* to pip install any deps, because that results in build variability based on changes in external code. Instead I think this package should simply BuildRequires: pytest which will pull in *our* version of pytest which will remain stable (and should be the correct version).

ddstreetmicrosoft (changes requested):
I think using BuildRequires: is a more appropriate way to handle this, as we really don't want to *ever* use pip to install anything during a build, we should be using *only* our provided packages.

Also, what's the problem you are solving? Is this just to fix python-virtualenv build-time test failure?

SPECS/python-virtualenv/pin-pytest-version.patch
ddstreetmicrosoft: OH! Our pytest is version *3*? Ok wow, yeah, ignore what i said, we have to use the pip installed version.

ddstreetmicrosoft (approved):

SPECS/python-virtualenv/pin-pytest-version.patch
corvus-callidus: Yup. Too old. Forgot to mention that. In an ideal world, yes I'd use our package instead. :)

PR #7785, opened 2024-02-08
[CHERRYPICK FROM FASTTRACK/2.0] fix cve-2022-21698 in nmi (#7681)

tobiasb-ms:
Cherry-pick commit 916f9d28c179456d7658fcae4a5eb1caec4b5022 to main. Original PR: #7681

Buddy Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=502165&view=results

### DIFF
--- a/SPECS/nmi/nmi.spec
+++ b/SPECS/nmi/nmi.spec
@@ -1,8 +1,8 @@
 %global debug_package %{nil}
 Summary:        Node Managed Identity
 Name:           nmi
-Version:        1.8.7
-Release:        15%{?dist}
+Version:        1.8.11
+Release:        1%{?dist}
 License:        MIT
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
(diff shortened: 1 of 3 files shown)


SPECS/nmi/nmi.spec
```diff
@@ -34,14 +34,12 @@ NMI is the resource that is used when your pods look to use their identity.
 %autosetup -c -N -n %{name}-%{version}
 pushd aad-pod-identity-%{version}
 %patch0 -p1
```
trungams, line 36: unrelated to the changes in this PR but is this still needed? `%autosetup` is already used above

PawelWMS (approved):

SPECS/nmi/nmi.spec
tobiasb-ms: It uses `%autosetup -N`, which does makes it not apply patches.

PR #7837, opened 2024-02-12
Add compatibility flag to all createrepo commands

mbykhovtsev-ms:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

PR #7715 upgrades `createrepo_c` package from version `0.17.5` to `1.0.3` which introduces breaking changes to how `createrepo` command creates metadata files, their format changes entirely which makes `tdnf` unable to read it properly. However if run with `--compatibility` flag, it will preserve the format as much as it can and thus `tdnf` can use the created metadata files for the repo. This PR adds such flags to all of our `createrepo` commands used so that we can upgrade the package. This flag does not change behavior of the command in our current version of the package, so it won't break anything in adding this now.

###### Change Log  

- Added `--compatibility` flag to all `creatrepo` commands

**YES**

###### Test Methodology

- Pipeline build id: local

### DIFF
--- a/.pipelines/containerSourceData/Dockerfile-Initial
+++ b/.pipelines/containerSourceData/Dockerfile-Initial
@@ -7,7 +7,7 @@ RUN --mount=type=bind,source=./Stage/,target=/dockerStage/ \\\
     tdnf install -y createrepo; \\\
     cp -r ${RPMS_PATH} ${LOCAL_REPO_PATH}; \\\
     cat /dockerStage/marinerLocalRepo.repo >> /etc/yum.repos.d/local.repo; \\\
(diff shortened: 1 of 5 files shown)


suresh-thelkar:
All the code changes look good to me. 

It seems we might also need to modify dnf.patch to include --compatibility flag for createrepo for fontpackages.spec. Can you please take a look into this.

suresh-thelkar (approved):

mbykhovtsev-ms:
> All the code changes look good to me.
> 
> It seems we might also need to modify dnf.patch to include --compatibility flag for createrepo for fontpackages.spec. Can you please take a look into this.

Good catch, I did not see command there at first only comments but upon second look I have noticed it needs change too. Updated

dmcilvaney (approved):

PR #7873, opened 2024-02-14
Add automatic mode for DAILY_BUILD_ID

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Update the `Makefile` to support `DAILY_BUILD_ID=lkg` and `DAILY_BUILD_ID=lkg-force`. These will automatically invoke `setuplkgtoolchain.sh` to find the most recent valid repo to use.

`lkg` will just set the repo, while `lkg-force` will also update the local manifests.

If the local toolchain manifests have been changed, `lkg` will result in an error. `lkg-force` or an explicit ID must be used.

###### Change Log  

- Enhance `DAILY_BUILD_ID` to take `lkg` and `lkg-force` as special case options.

**YES**

###### Test Methodology

- Local tests
- Full pipeline tbd.

### DIFF
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -141,68 +141,8 @@ include $(SCRIPTS_DIR)/utils.mk
 
 ######## DAILY BUILD AND PREVIEW REPO SUPPORT ########
 
-##help:var:DAILY_BUILD_ID:{'lkg','lkg-force',<build_id>}='lkg' will auto select latest daily build repo. Unless LKG_MANIFESTS='n' is used, unmodifed manifests will be udpated to match. 'lkg-force' will overwrite changes. An explicit ID of the daily build repo can also be used.
-DAILY_BUILD_ID ?=
+include $(SCRIPTS_DIR)/daily_build.mk
 
-##help:var:LKG_MANIFESTS:{y,n}='y' will auto update the manifests to the latest daily build if used in conjunction with DAILY_BUILD_ID='lkg' or 'lkg-force'.
-LKG_MANIFESTS ?= y
-skip_manifest_updates = $(if $(filter y,$(LKG_MANIFESTS)),,--skip-manifest-updates)
-
-ifneq ($(DAILY_BUILD_ID),)
-   # If DAILY_BUILD_ID is set to "lkg" or "lkg-force" we should attempt to auto select the best daily build
-   ifeq ($(DAILY_BUILD_ID),lkg)
-      $(warning Auto detecting DAILY_BUILD_ID based on the latest known good build)
-      override DAILY_BUILD_ID := $(shell $(SCRIPTS_DIR)/setuplkgtoolchain.sh --auto-id $(skip_manifest_updates))
-      ifneq ($(.SHELLSTATUS),0)
-         $(error Failed to auto detect DAILY_BUILD_ID... try DAILY_BUILD_ID=lkg-force to overwrite manifests)
-      endif
-   endif
-   ifeq ($(DAILY_BUILD_ID),lkg-force)
-      # lkg-force is incompatible with LKG_MANIFESTS=n
-      ifneq ($(LKG_MANIFESTS),y)
-         $(error DAILY_BUILD_ID=lkg-force requires LKG_MANIFESTS=y)
-      endif
-      $(info Auto detecting DAILY_BUILD_ID based on the latest known good build AND overwriting manifests)
-      override DAILY_BUILD_ID := $(shell $(SCRIPTS_DIR)/setuplkgtoolchain.sh --auto-id --force $(skip_manifest_updates))
-      ifneq ($(.SHELLSTATUS),0)
-         $(error Failed to auto detect DAILY_BUILD_ID and update manifests)
-      endif
-   endif
-
-   $(warning Using Daily Build $(DAILY_BUILD_ID))
-   # Ensure build_arch is set
-   ifeq ($(build_arch),)
-      $(error build_arch must be set when using DAILY_BUILD_ID)
-   endif
-   # build_arch cannot be directly used because Azure storage do not support container names with '_' char
-   ifeq ($(build_arch),x86_64)
-      # The actual repo is found at <URL>/, while a duplicate copy of all the rpms can be found at <URL>/built_rpms_all/
-      # Include both so that the tools that expect a valid repo work, while the tools that expect a basic URL also work.
-      override PACKAGE_URL_LIST   := https://mariner3dailydevrepo.blob.core.windows.net/daily-repo-$(DAILY_BUILD_ID)-x86-64/built_rpms_all \
-                                       https://mariner3dailydevrepo.blob.core.windows.net/daily-repo-$(DAILY_BUILD_ID)-x86-64
-      override SRPM_URL_LIST      := https://mariner3dailydevrepo.blob.core.windows.net/daily-repo-$(DAILY_BUILD_ID)-x86-64/SRPMS
-   else
-      override PACKAGE_URL_LIST   := https://mariner3dailydevrepo.blob.core.windows.net/daily-repo-$(DAILY_BUILD_ID)-aarch64/built_rpms_all \
-                                       https://mariner3dailydevrepo.blob.core.windows.net/daily-repo-$(DAILY_BUILD_ID)-aarch64
-      override SRPM_URL_LIST      := https://mariner3dailydevrepo.blob.core.windows.net/daily-repo-$(DAILY_BUILD_ID)-aarch64/SRPMS
-   endif
-else
-   ifeq ($(USE_PREVIEW_REPO),y)
-      override PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/preview/base/$(build_arch)
-      override PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/preview/base/debuginfo/$(build_arch)
-      override PACKAGE_URL_LIST   += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/preview/Microsoft/$(build_arch)
-      override SRPM_URL_LIST      += https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR_ID)/preview/base/srpms
-      ifneq ($(wildcard $(PREVIEW_REPO)),)
-         override REPO_LIST += $(PREVIEW_REPO)
-      else
-         $(warning )
-         $(warning ######################### WARNING #########################)
-         $(warning 'USE_PREVIEW_REPO=y' set but '$(PREVIEW_REPO)' is missing. Regenerate toolkit's 'repos' directory. Remove 'USE_PREVIEW_REPO' for core builds.)
-         $(warning ######################### WARNING #########################)
-         $(warning )
-      endif
-   endif
-endif
 
 ######## REMAINING BUILD FLAGS ########
 
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -139,21 +139,31 @@ SRPM_URL_LIST      ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR
 # Variable prerequisite tracking
 include $(SCRIPTS_DIR)/utils.mk
 
-######## REMAINING BUILD FLAGS ########
+######## DAILY BUILD AND PREVIEW REPO SUPPORT ########
 
-##help:var:DAILY_BUILD_ID:{lkg,lkg-force,<build_id>}='lkg' will auto select latest daily build repo, 'lkg-force' will also replace the toolchain manifest files. An explicit ID of the daily build repo can also be used.
+##help:var:DAILY_BUILD_ID:{'lkg','lkg-force',<build_id>}='lkg' will auto select latest daily build repo. Unless LKG_MANIFESTS='n' is used, unmodifed manifests will be udpated to match. 'lkg-force' will overwrite changes. An explicit ID of the daily build repo can also be used.
+DAILY_BUILD_ID ?=
+
+##help:var:LKG_MANIFESTS:{y,n}='y' will auto update the manifests to the latest daily build if used in conjunction with DAILY_BUILD_ID='lkg' or 'lkg-force'.
+LKG_MANIFESTS ?= y
+skip_manifest_updates = $(if $(filter y,$(LKG_MANIFESTS)),,--skip-manifest-updates)
+
 ifneq ($(DAILY_BUILD_ID),)
    # If DAILY_BUILD_ID is set to "lkg" or "lkg-force" we should attempt to auto select the best daily build
    ifeq ($(DAILY_BUILD_ID),lkg)
       $(warning Auto detecting DAILY_BUILD_ID based on the latest known good build)
-      override DAILY_BUILD_ID := $(shell $(SCRIPTS_DIR)/setuplkgtoolchain.sh --auto-id)
+      override DAILY_BUILD_ID := $(shell $(SCRIPTS_DIR)/setuplkgtoolchain.sh --auto-id $(skip_manifest_updates))
       ifneq ($(.SHELLSTATUS),0)
          $(error Failed to auto detect DAILY_BUILD_ID... try DAILY_BUILD_ID=lkg-force to overwrite manifests)
       endif
    endif
    ifeq ($(DAILY_BUILD_ID),lkg-force)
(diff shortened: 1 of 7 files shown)


toolkit/Makefile
```diff
@@ -141,7 +141,24 @@ include $(SCRIPTS_DIR)/utils.mk
...
+      endif
+   endif
+   ifeq ($(DAILY_BUILD_ID),lkg-force)
+      $(info Auto detecting DAILY_BUILD_ID based on the latest known good build AND overwriting manifests)
+      override DAILY_BUILD_ID := $(shell $(SCRIPTS_DIR)/setuplkgtoolchain.sh --auto-id --force)
+      ifneq ($(.SHELLSTATUS),0)
+         $(error Failed to auto detect DAILY_BUILD_ID and update manifests)
+      endif
+   endif
+
```
PawelWMS, line 161: I have nothing against adding an interface to run `setuplkgtoolchain.sh` directly from a `make` command. I don't want to overload `DAILY_BUILD_ID`, however. With the current change we're adding special values modifying what setting `DAILY_BUILD_ID` does, which is modifying the URL lists to look for the toolchain packages under a different address. It's a different responsibility, so I think we should add a separate argument for this behaviour.

PawelWMS (changes requested):

toolkit/Makefile
anphel31: I don't really see a problem with this change. I think it's fine to overload the DAILY_BUILD_ID value with `lkg` or `lkg-force`. Just my opinion, but I think we have too many make targets already, and we should try to consolidate them and avoid creating new ones unless they are really necessary.

anphel31 (approved):

toolkit/Makefile
dmcilvaney: Was going to respond the same, I'd like to try and consolidate our flags more in the future. Its much easier to offer help/readme type documentation for a smaller number of commands that each do a few things. In this case I feel like the explicit and automatic modes are doing something VERY similar.

anphel31 (approved):

toolkit/Makefile
```diff
@@ -139,9 +139,36 @@ SRPM_URL_LIST      ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR
 # Variable prerequisite tracking
 include $(SCRIPTS_DIR)/utils.mk
 
-######## REMAINING BUILD FLAGS ########
+######## DAILY BUILD AND PREVIEW REPO SUPPORT ########
```
PawelWMS, line 142: Considering the comment, do you think it's time to create another `*.mk` file and move this logic there?;)

toolkit/Makefile
```diff
@@ -139,9 +139,36 @@ SRPM_URL_LIST      ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR
...
+DAILY_BUILD_ID ?=
+
+##help:var:LKG_MANIFESTS:{y,n}='y' will auto update the manifests to the latest daily build if used in conjunction with DAILY_BUILD_ID='lkg' or 'lkg-force'.
+LKG_MANIFESTS ?= y
+skip_manifest_updates = $(if $(filter y,$(LKG_MANIFESTS)),,--skip-manifest-updates)
 
 ifneq ($(DAILY_BUILD_ID),)
+   # If DAILY_BUILD_ID is set to "lkg" or "lkg-force" we should attempt to auto select the best daily build
+   ifeq ($(DAILY_BUILD_ID),lkg)
+      $(warning Auto detecting DAILY_BUILD_ID based on the latest known good build)
```
PawelWMS, line 154: Why not display this warning for `lkg-force` as well?

toolkit/Makefile
```diff
@@ -139,9 +139,36 @@ SRPM_URL_LIST      ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR
...
+      override DAILY_BUILD_ID := $(shell $(SCRIPTS_DIR)/setuplkgtoolchain.sh --auto-id $(skip_manifest_updates))
+      ifneq ($(.SHELLSTATUS),0)
+         $(error Failed to auto detect DAILY_BUILD_ID... try DAILY_BUILD_ID=lkg-force to overwrite manifests)
+      endif
+   endif
+   ifeq ($(DAILY_BUILD_ID),lkg-force)
+      # lkg-force is incompatible with LKG_MANIFESTS=n
+      ifneq ($(LKG_MANIFESTS),y)
+         $(error DAILY_BUILD_ID=lkg-force requires LKG_MANIFESTS=y)
+      endif
```
PawelWMS, line 164: Suggestion: since we're already running the script and make it do its own validation, maybe we can keep this check there? Mostly to reduce logic in makefiles.

toolkit/scripts/setuplkgtoolchain.sh
```diff
@@ -23,39 +28,97 @@ get_lkg() {
 }
 
 check_for_modified_manifests() {
+    auto=$1
+    force=$2
```
PawelWMS, line 32: Consistency nit-pick:
```suggestion
    local auto=$1
    local force=$2
```

Or remove `local` from the other function.

toolkit/scripts/setuplkgtoolchain.sh
```diff
@@ -23,39 +28,97 @@ get_lkg() {
...
 EOF
-
-echo -e "Finished syncing toolchain to LKG ('$DAILY_BUILD_ID' - '$GIT_COMMIT')"
-echo -e "To download LKG toolchain, run:\n\tsudo make toolchain -j$(nproc) REBUILD_TOOLCHAIN=n REBUILD_TOOLS=y DAILY_BUILD_ID=$DAILY_BUILD_ID"
+    echo -e "Finished syncing toolchain to LKG ('$DAILY_BUILD_ID' - '$GIT_COMMIT')"
+    echo -e "To download LKG toolchain, run:\n\tsudo make toolchain -j$(nproc) REBUILD_TOOLCHAIN=n REBUILD_TOOLS=y DAILY_BUILD_ID=$DAILY_BUILD_ID"
+else
+    # In auto mode, just print the LKG build ID so we can use it in the calling script
+    echo $DAILY_BUILD_ID
+fi
```
PawelWMS, line 122: I'm a worried about this bit, that it's now tightly coupling a script that can be run independently with one specific use in our toolkit. Would it be problematic to have the script print its original output and add a grep in the caller like `grep -oP (?<=DAILY_BUILD_ID=')[^']+`?

toolkit/scripts/setuplkgtoolchain.sh
```diff
@@ -23,39 +28,97 @@ get_lkg() {
...
 
 check_for_modified_manifests() {
+    auto=$1
+    force=$2
     if git status --porcelain | grep -qP "(pkggen_core|toolchain)_$(uname -p).txt"; then
-        echo "Local modifications to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt' detected."
-        echo -e "\nNOTE: Changes to manifests were detected, and these will be overwritten. Hit Ctrl+C within 10 seconds to cancel...\n"
-        sleep 10s
+        if $force; then
+            echo "Forcing overwrite of local changes to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt'" 1>&2
```
PawelWMS, line 35: Why a redirect to `stderr`? Looks like intended behaviour.

toolkit/scripts/setuplkgtoolchain.sh
```diff
@@ -23,39 +28,97 @@ get_lkg() {
...
+        else
+            if $auto; then
+                echo "Local modifications to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt' detected, will not auto update without --force" 1>&2
+                exit 1
+            else
+                echo "Local modifications to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt' detected." 1>&2
+                echo -e "\nNOTE: Changes to manifests were detected, and these will be overwritten. Hit Ctrl+C within 10 seconds to cancel...\n" 1>&2
+                sleep 10s
+            fi
+        fi
```
PawelWMS, line 45: Idea to reduce indentation and one `fi`:
```suggestion
        elif $auto; then
            echo "Local modifications to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt' detected, will not auto update without --force" 1>&2
            exit 1
        else
            echo "Local modifications to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt' detected." 1>&2
            echo -e "\nNOTE: Changes to manifests were detected, and these will be overwritten. Hit Ctrl+C within 10 seconds to cancel...\n" 1>&2
            sleep 10s
        fi
```

toolkit/scripts/setuplkgtoolchain.sh
```diff
@@ -23,39 +28,97 @@ get_lkg() {
...
-        sleep 10s
+        if $force; then
+            echo "Forcing overwrite of local changes to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt'" 1>&2
+        else
+            if $auto; then
+                echo "Local modifications to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt' detected, will not auto update without --force" 1>&2
+                exit 1
+            else
+                echo "Local modifications to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt' detected." 1>&2
+                echo -e "\nNOTE: Changes to manifests were detected, and these will be overwritten. Hit Ctrl+C within 10 seconds to cancel...\n" 1>&2
```
PawelWMS, line 42: Similar question about `stderr` redirects here.

toolkit/scripts/setuplkgtoolchain.sh
```diff
@@ -23,39 +28,97 @@ get_lkg() {
...
+update_manifest() {
+    local filename
+    local url
+    local filepath
+    filename="$1"
+    url="https://raw.githubusercontent.com/microsoft/CBL-Mariner/$GIT_COMMIT/toolkit/resources/manifests/package/${filename}"
+    filepath="$ROOT_FOLDER/toolkit/resources/manifests/package/${filename}"
+    filepath_tmp="${filepath}.tmp"
+
+    echo "Downloading $filename manifest from $url" 1>&2
```
PawelWMS, line 61: OK, now I'm really curious why we want to print everything to `stderr`. I'm not saying we don't, I just want to understand.:)

toolkit/scripts/setuplkgtoolchain.sh
```diff
@@ -23,39 +28,97 @@ get_lkg() {
...
     if git status --porcelain | grep -qP "(pkggen_core|toolchain)_$(uname -p).txt"; then
-        echo "Local modifications to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt' detected."
-        echo -e "\nNOTE: Changes to manifests were detected, and these will be overwritten. Hit Ctrl+C within 10 seconds to cancel...\n"
-        sleep 10s
+        if $force; then
+            echo "Forcing overwrite of local changes to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt'" 1>&2
+        else
+            if $auto; then
+                echo "Local modifications to 'pkggen_core_$(uname -p).txt' or 'toolchain_$(uname -p).txt' detected, will not auto update without --force" 1>&2
+                exit 1
```
neha170, line 39: Why is it `exit 1`? If the user chooses `auto`, it is not an error

toolkit/scripts/setuplkgtoolchain.sh
PawelWMS: I thought about it a bit more and I think the fact we're not making this tool return two different things required for the toolkit to function (update the manifests and return the build ID, which is now an input for something else) maybe causing us headaches now (echoing two different output depending on `auto_id` and forcing us to print to `stderr`) and in the future.

In addition to that, we're not introducing a new way for the toolkit to update/create files that is I think also unusual for `make` builds. Instead of using a target, we have a shell call done while we assign a value to the `DAILY_BUILD_ID` argument. This makes it even harder to understand what's happening in the `Makefile`.

What do you think about splitting the logic:
- Have a separate target, which updates the manifests (and prints whatever information or warnings it likes). This way we also won't have to check its error code, since `make` will do it for us.
- Have a separate script dedicated only to returning the build ID or other values but making no file changes. This is still not ideal (the script may still fail and require a check of the error code) but I think it's a consequence of using URLs directly encoded inside arguments like `PACKAGE_URL_LIST`. Maybe we can remedy that somehow later (move the URLs to files?).

toolkit/Makefile
dmcilvaney: I was debating it actually... I can do that.

toolkit/Makefile
dmcilvaney: ah, typo... they both need to be warnings so they go to std/err. They will pollute the output otherwise (messing with stuff like `printvar` etc.)

toolkit/Makefile
dmcilvaney: Sure, I can move that. Probably not a big deal to simply this if so.

toolkit/Makefile
```diff
@@ -139,43 +139,13 @@ SRPM_URL_LIST      ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR
...
 include $(SCRIPTS_DIR)/utils.mk
 
+######## DAILY BUILD AND PREVIEW REPO SUPPORT ########
+
+include $(SCRIPTS_DIR)/daily_build.mk
+
+
 ######## REMAINING BUILD FLAGS ########
 
-ifneq ($(DAILY_BUILD_ID),)
```
dmcilvaney, line 144: Note: This is copied verbatim to the new makefile.

toolkit/Makefile
```diff
@@ -139,43 +139,13 @@ SRPM_URL_LIST      ?= https://packages.microsoft.com/cbl-mariner/$(RELEASE_MAJOR
 # Variable prerequisite tracking
 include $(SCRIPTS_DIR)/utils.mk
 
+######## DAILY BUILD AND PREVIEW REPO SUPPORT ########
+
+include $(SCRIPTS_DIR)/daily_build.mk
+
+
```
dmcilvaney, line 146:
```suggestion
```

toolkit/scripts/daily_build.mk
```diff
@@ -0,0 +1,90 @@
...
+#	- Rules to update manifests
+#	- Mechanism to update DAILY_BUILD_ID
+
+TOOLCHAIN_MANIFEST ?= $(TOOLCHAIN_MANIFESTS_DIR)/toolchain_$(build_arch).txt
+WORKER_CHROOT_MANIFEST = $(TOOLCHAIN_MANIFESTS_DIR)/pkggen_core_$(build_arch).txt
+
+lkg_workdir = $(BUILD_DIR)/daily_build_id
+
+######## CONFIGURE FLAGS ########
+
```
dmcilvaney, line 14:
```suggestion

# DAILY_BUILD_ID is the main point of entry. It can be one of 'lkg', 'lkg-force', or <build-id>. Based on this selection 'LKG_MANIFESTS' and 'FORCE_MANIFEST_UPDATES' are selected by default. They may be overwritten if desired.

# LKG_MANIFESTS will update the manifest files to match the commit in the LKG file. This does not make sense for arbitrary build-ids however, so only default to 'y' for 'lkg' or 'lkg-force'

# PKG_MANIFEST_OVERWRITE will ignore local changes to the manifests and replace them with the LKG version.

```

toolkit/scripts/update_manifest.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
```
dmcilvaney, line 4:
```suggestion

# Updates the requested manifest file based on the contents of the LKG json.

```

toolkit/scripts/setuplkgtoolchain.sh
dmcilvaney: I've split these off, and added a dedicated `dailly_build.mk` file.

toolkit/scripts/setuplkgtoolchain.sh
dmcilvaney: When doing `variable = $(shell some_command.sh arg)`, all output on `stdout` will be sent to the variable. That means that any helpful debug output or status to the user needs to be on `stderr`. We can check the return with `ifneq ($(.SHELLSTATUS),0)...` but that won't explain why it failed,  just that it was non-zero.

dmcilvaney:
Superceded by https://github.com/microsoft/azurelinux/pull/9899

PR #7887, opened 2024-02-14
[RFC] tools: srpmpacker: refactor as a library

mfrw:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This is a part of a larger set of refactors which was initially started within the 2.0 timeframe but we decided to target this to 3.0
The PR only refactors business logic from the main package and introduces a new library package of the same name.
The scope of the PR is strictly a 1:1 refactor and no change from the existing behaviour.

N/A - It is a toolkit specific change.

###### Associated issues  

- Related 2.0 PR: #2983
- NA [WIll create a tracking issue for all the PRs in this class]

###### Links to CVEs  
- NA

###### Test Methodology

- Pipeline build id: Local Build
- Full Build: [link](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=506221&view=results)


toolkit/tools/pkg/srpmpacker/srpmpacker.go
```diff
@@ -0,0 +1,994 @@
...
+	timestamp.StopEvent(nil)
+
+	// A pack list may be provided, if so only pack this subset.
+	// If non is provided, pack all srpms.
+	packList, err := packagelist.ParsePackageListFile(cfg.PackListFile)
+	logger.PanicOnError(err)
+
+	err = createAllSRPMsWrapper(cfg.SpecsDir, cfg.DistTag, cfg.BuildDir, cfg.OutDir, cfg.WorkerTar, cfg.Workers, cfg.ConcurrentNetOps, cfg.NestedSourcesDir, cfg.RepackAll, cfg.RunCheck, packList, templateSrcConfig)
+
+	return nil
```
cwize1, line 136: `return err`?

toolkit/tools/pkg/srpmpacker/srpmpacker.go
```diff
@@ -0,0 +1,994 @@
...
+	case SignatureEnforceString:
+		templateSrcConfig.signatureHandling = signatureEnforce
+	case SignatureSkipCheckString:
+		logger.Log.Warn("Skipping signature enforcement")
+		templateSrcConfig.signatureHandling = signatureSkipCheck
+	case SignatureUpdateString:
+		logger.Log.Warn("Will update signature files as needed")
+		templateSrcConfig.signatureHandling = signatureUpdate
+	default:
+		logger.Log.Fatalf("Invalid signature handling encountered: %s. Allowed: %s", cfg.SignatureHandling, cfg.ValidSignatureLevels)
```
cwize1, line 101: I suspect these panics/fatals only exist because this code was originally under the `main` function. They should probably be converted to normal error handling.

toolkit/tools/pkg/srpmpacker/srpmpacker.go
```diff
@@ -0,0 +1,994 @@
...
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/logger"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/network"
+	packagelist "github.com/microsoft/CBL-Mariner/toolkit/tools/internal/packlist"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/retry"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/rpm"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/CBL-Mariner/toolkit/tools/internal/timestamp"
+)
+
+const (
```
cwize1, line 30: This should probably live in `config.go`.

toolkit/tools/pkg/srpmpacker/srpmpacker.go
```diff
@@ -0,0 +1,994 @@
...
+
+// specState holds the state of a SPEC file: if it should be packed and the resulting SRPM if it is.
+type specState struct {
+	specFile string
+	srpmFile string
+	toPack   bool
+	err      error
+}
+
+func (cfg *Config) CreateAllSrpmsWrapper() error {
```
cwize1, line 87: I don't think this should be a member function of `Config`. It should take `Config` as its first parameter,

cwize1 (changes requested):

toolkit/tools/pkg/srpmpacker/srpmpacker.go
```diff
@@ -0,0 +1,994 @@
+package srpmpacker
```
cwize1, line 1: It is generally a good idea to ensure that all modules have unique directory names, as this avoids needing to rename modules in imports.

toolkit/tools/pkg/srpmpacker/srpmpacker.go
mfrw: Yes I agree, but I think in the scope of the current PR, IMHO we should keep them as is. We should do a treewide refactor in a subsequent PR where we remove all panics and Fatals from any leaf level functions & have a top-level function that does the panic.
Our API design IMO should be to bubble errors up from a function instead of `painc`/`Fatal` in a given function.

What are your thoughts ?

toolkit/tools/pkg/srpmpacker/srpmpacker.go
mfrw: Yes, I agree.

toolkit/tools/pkg/srpmpacker/srpmpacker.go
mfrw: This was an intentional design choice and ofcourse is up for discussion.
The rationale behind this is our public interface function should have a `context` & maybe a `logger Interface` and rest flows via the config.
The `context` gives us the capability to bail out prematurely if we want. Having a `logger Interface` gives us capability to switch loggers and also maybe have a no-op logger for testing.
As it stands right now we have a global logger which is very closely coupled with the `logrus` library.
Also a few local experiments that I have done, give an indication that our logger implementation has an explicit `Mutex`;
If we have an API design that allows us to switch logger implementations at fly, there is a chance we would get some more perf.
What do you think ?

toolkit/tools/pkg/srpmpacker/srpmpacker.go
mfrw: Can you please elaborate on a little more.
Also, this idiom/pattern has taken from the `kubernetes` project. e.g `kubelet` is a `main package` as well as a library.

toolkit/tools/pkg/srpmpacker/srpmpacker.go
cwize1: Yes, ideally there should only a single panic/fatal call in the main function. Everything else should be normal error handling.

Tree-wide refactors are generally pretty painful to get through PR. So, it might be better to break it up into separate changes for each tool, like you are doing in this PR.

Also, it might be better to fix the error handling first and then do this change where you split-off the code into a separate module. That way, you don’t have an intermediate time where the code is in a bad state.

toolkit/tools/pkg/srpmpacker/srpmpacker.go
cwize1: Eh, I guess the current code is fine. In theory, nobody will ever import the tool’s main package. So, this shouldn’t be a problem.

toolkit/tools/pkg/srpmpacker/srpmpacker.go
cwize1: There isn’t a limit to the number of parameters a function can take.

For example:

```golang
func (ctx context.Context, logger logrus.StdLogger, cfg *Config) { … }
```

The question at hand is if this is an operation on a config object or an operation that uses the config object as input? I’m pretty sure it is the latter.

cwize1: There have been some cases where it would have been handy if we supported `context` in the toolkit. But fixing this will be a very large undertaking. And I’m not sure the benefit justifies the cost at the moment.

`context` really comes in handy when you are doing a lot of network operations, which may fail or you may want to cancel for taking too long. But a lot of what the toolkit does doesn’t use the network.

cwize1: In regards to logging, not needing to pass around a logger object is somewhat nice and it reduces visual noise within the codebase. Also, changing the logging pattern would be a massive undertaking. So, there would need to be strong justification for changing it.

If you think there is performance improvements to be had, then you could write an experimental change and provide the % improvement you observed.

PR #7890, opened 2024-02-14
use strings.Fields to handle whitespace between items in list

### LINKED ISSUE #7895: List parsing for build tools doesn't handle multiple spaces correctly.
**Describe the bug**
For build parameters like `SRPM_PACK_LIST` we take a space-separated list, like `SRPM_PACK_LIST="openssl SymCrypt"`. The [code that parses](https://github.com/microsoft/CBL-Mariner/blob/2c4b506d7a1582164be7c0151ab049e7318f7dc3/toolkit/tools/internal/exe/exe.go#L58C1-L66C2) those into an array of packages doesn't handle multiple spaces between the package names.

**To Reproduce**
1. In a command-line, add multiple spaces between package names in a parameter, like `SRPM_PACK_LIST="foo   bar"`.
2. Build fails, complaining that it can't find one or more packages named `''`.

**Expected behavior**
Parsing code should handle multiple spaces.

tobiasb-ms:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 
For build parameters like `SRPM_PACK_LIST` we take a space-separated list, like `SRPM_PACK_LIST="openssl SymCrypt"`. Currently, the code that parses those into an array of packages doesn't handle multiple spaces between the package names. For example, if you add an extra space between the package names, like `"openssl  SymCrypt"`, it will parse it into three packages: "`openssl`", "``", and "`SymCrypt`" -- note the empty package name. This makes our build tools fail down the line, as they try to handle an empty package name.

###### Change Log  
- Updated the list parsing code to handle multiple spaces/whitespace between package names.
- Added tests.

**YES**

###### Associated issues  
- fixes #7895

###### Test Methodology
- Tested locally
- Full Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=505261&view=results

### DIFF
--- a/toolkit/tools/internal/exe/exe_test.go
+++ b/toolkit/tools/internal/exe/exe_test.go
@@ -1,0 +1,48 @@
+package exe
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseListArgument(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected []string
+	}{
+		{
+			name:     "empty input",
+			input:    "",
+			expected: []string{},
+		},
+		{
+			name:     "single value",
+			input:    "value",
+			expected: []string{"value"},
+		},
+		{
+			name:     "multiple values",
+			input:    "value1 value2 value3",
+			expected: []string{"value1", "value2", "value3"},
+		},
+		{
+			name:     "leading/trailing spaces",
+			input:    "  value1  value2  value3  ",
+			expected: []string{"value1", "value2", "value3"},
+		},
+		{
+			name:     "extra spaces",
+			input:    "value1  value2   value3",
+			expected: []string{"value1", "value2", "value3"},
+		},
+	}
+
+	for _, tt := range tests {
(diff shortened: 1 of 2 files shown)


dmcilvaney (approved):

toolkit/tools/internal/exe/exe_test.go
```diff
@@ -0,0 +1,58 @@
...
+		},
+		{
+			name:     "mixed whitespace",
+			input:    "value1\tvalue2  value3",
+			expected: []string{"value1", "value2", "value3"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
neha170, line 53: Not familiar with golang testing, so just making sure. This would show exactly which testcase failed and which passed?

dmcilvaney: Hah, actually had this chat with Tobias as well. This seems to be the "recommended" form for tests.  Each sub-test will have its name printed if it fails, although unfortunately the line numbers refer back to the call to `Run()` so a bit of extra leg work is needed.

PR #7918, opened 2024-02-16
Fixed toolchain spec packing for hosts with RPM < 4.18.

PawelWMS:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Toolchain packages are being packed into SRPMs using the host's RPM (Azure Linux worker chroot is not available at that time). For RPM versions < 4.18 (common on Ubuntu 22.04 and earlier) the only valid `%patch` macro usage is either `%patch[number]` or `%patch -P [number]`. Since the former is deprecated starting from RPM 4.18, I've updated toolchain specs to use `-P [number]`.

###### Change Log  

- Updated toolchain specs' `%patch` macros.
- Enabled missing 2 patches in `perl.spec`.
- Extended the spec PR check to prevent invalid `%patch` macro usage for toolchain packages.

Yes.

###### Associated issues  

- #7136

###### Test Methodology

- Local tests of the PR check and package builds.
- [Full toolchain build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=506999&view=results).


SPECS/zip/zip.spec
```diff
@@ -31,6 +30,9 @@ make prefix=%{buildroot}/%{_prefix} MANDIR=%{buildroot}/usr/share/man/man1 -f un
 %{_mandir}/*
 
 %changelog
+*   Thu Feb 15 2024 Pawel Winogrodzki <pawelwi@microsoft.com> - 3.0-6
+-   - Updated patch application macros.
```
cwize1, line 34: nit: Extra `-`

SPECS/perl/perl.spec
```diff
@@ -6847,6 +6838,10 @@ popd
 
 # Old changelog entries are preserved in CVS.
 %changelog
+* Thu Feb 15 2024 Pawel Winogrodzki <pawelwi@microsoft.com> - 4:5.38.2-504
+- Updated patch application macros.
+- Applying missed patches 14 and 202.
```
cwize1, line 6843: Do we know why these were excluded?

toolkit/scripts/check_spec_guidelines.py
```diff
@@ -121,7 +126,31 @@ def check_source_attribution(spec_path: str):
     return False
 
 
-def check_spec(spec_path):
+def check_toolchain_patch_lines(spec_path: str, toolchain_specs: list):
```
cwize1, line 129: Should `toolchain_specs` be `set` not `list`?

SPECS/perl/perl.spec
PawelWMS: Yes, I spoke with @nicogbg. Turns out we accidentally missed them during an upgrade but there was no reason to skip them.

cwize1 (approved):

PR #7982, opened 2024-02-21
fix error when using tdnf inside of functions from setup_functions.sh

tobiasb-ms:
Fix errors in the `containerized-rpmbuild` environment for 3.0 when using helper functions that call `tdnf`.

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 
In the `containerized-rpmbuild` environment, we have an alias for `tdnf` that adds a `--releasever` flag in all cases and for `3.0` builds adds some flags to fix up what repo we will go to. However, as currently implemented the extra flags for `3.0` are added after sourcing `setup_functions.sh`. Because aliases are expanded when functions are read, not when they are executed, this means that any function from `setup_functions.sh` (like `install_dependencies`) will have the wrong flags. This results in seeing a failure, though oddly the packages still get installed.

To fix this, I changed `tdnf` from an alias to a function and simplified the flag settings a little.

###### Change Log  
- Change docker file to pipe a variable through to the container that has the specified version.
- Define a function `tdnf` in `setup_functions.sh` that uses an array of parameters which is conditionally defined based on that variable.

**NO**

###### Test Methodology

- Tested locally with `containerized-rpmbuild`

### DIFF
--- a/toolkit/scripts/containerized-build/resources/mariner.Dockerfile
+++ b/toolkit/scripts/containerized-build/resources/mariner.Dockerfile
@@ -10,7 +10,7 @@ LABEL containerized-rpmbuild=$mariner_repo/build
 COPY resources/local_repo /etc/yum.repos.d/local_repo.disabled_repo
 
 # This is used by setup_functions.sh to determine tdnf default arguments
-ENV CONTAINERIZED_RPMBUILD_MARINER_VERSION=${version}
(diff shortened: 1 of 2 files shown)


neha170 (approved):
Thanks for root causing and fixing this.
Minor suggestion: we can rename the `CONTAINERIZED_RPMBUILD_MARINER_VERSION` to `CONTAINERIZED_RPMBUILD_AZL_VERSION`

PR #8013, opened 2024-02-21
Refactor dist, azl macros

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

We define `%dist` as `.azl3` (previously `.cm2`) in `/usr/lib/rpm/macros.d/dist.macros`. This is currently part of `mariner-rpm-macros` package. This has a few issues:
- mariner-rpm-macros is not included in many images, it is generally only pulled in with `rpm-build` etc.
- `%dist` can't be used easily for conditional flags in spec files (`%if 0%{?rhel}` is a common pattern).

Define a new macro `%azl 3` (which will be defined in `macros.dist` in the future). Since the `DIST_TAG` can be customized we set this macro explicitly from values baked into the go tools.

https://microsoft.sharepoint.com/:w:/t/Mariner/EQ5GV3mMEZRLsLNIwxFXj-gBxBalfkJpba-xHiz-iOJc-Q?e=Dw2Min

###### Change Log  

- Add `%azl` to build macros
- Update specs that are using similar macros to use the new versions.

**YES**

###### Test Methodology

- Pipeline build id: ~https://dev.azure.com/mariner-org/mariner/_build/results?buildId=510616&view=results~ https://dev.azure.com/mariner-org/mariner/_build/results?buildId=515509&view=results


SPECS/mariner-rpm-macros/mariner-rpm-macros.spec
```diff
@@ -79,7 +87,9 @@ install -p -m 755 -t %{buildroot}%{rcdir} gen-ld-script.sh
 install -p -m 755 -t %{buildroot}%{rcdir} generate-package-note.py
 install -p -m 755 -t %{buildroot}%{rcdir} verify-package-notes.sh
 
-sed -e 's|@DIST@|%{dist}|g' %{SOURCE28} > macros.dist
+major_version=`echo %{mariner_release_version} | cut -d'.' -f1`
+sed -e 's|@DIST@|%{dist}|g' %{SOURCE28} | sed -e "s|@MAJORVER@|$major_version|g" > macros.dist
```
PawelWMS, line 91: Just an idea to keep things shorter, not an actual issue:
```suggestion
sed -e 's|@DIST@|%{dist}|g' -e "s|@MAJORVER@|$major_version|g" %{SOURCE28} > macros.dist
```

SPECS/systemd/systemd.spec
```diff
@@ -1,10 +1,3 @@
-
-# Temporary until azure-release-common is updated to carry macros.dist that defines %azure
-%if %{defined azure}
```
PawelWMS, line 3: Thanks for the clean-up!

toolkit/scripts/rpmops.sh
```diff
@@ -14,7 +14,21 @@ fi
 
 # Additional macros required to parse spec files.
 DIST_TAG=$(make -s -f $REPO_ROOT/toolkit/Makefile get-dist-tag)
-DEFINES=(-D "with_check 1" -D "dist $DIST_TAG")
+RELEASE_NUM=$(make -s -f $REPO_ROOT/toolkit/Makefile get-release-major)
```
PawelWMS, line 17: I think we may need something similar inside a few more places where we manually define the macros:
- `toolkit/scripts/toolchain/check_manifests.sh`,
- `toolkit/scripts/toolchain/build_official_toolchain_rpms.sh`.

toolkit/scripts/rpmops.sh
```diff
@@ -14,7 +14,21 @@ fi
...
+# 'azl<ver> 1': Strip the leading '.' from the DIST_TAG
+distro_def="${DIST_TAG#.} 1"
+
+# 'azl <ver>': Strip the leading '.' and the trailing number from the DIST_TAG
+distro_name="$(echo ${DIST_TAG#.} | sed 's/[0-9]\+$//')"
+distro_majver="$(echo $RELEASE_NUM | cut -d'.' -f1)"
+distro_ver="${distro_name} ${distro_majver}"
+
+DEFINES=(-D "with_check 1" -D "dist $DIST_TAG" -D "$distro_def" -D "$distro_ver")
+echo "EXTRA DEFINES: ${DEFINES[*]}"
```
PawelWMS, line 31: Please don't echo anything extra to `stdout` for a successful run - this is meant to mimic `rpmspec` and is used by tooling parsing its output.

dmcilvaney: Sorry, I had cleaned that up locally but my git upstream got messed up, re-pushed.

ddstreetmicrosoft:
I would really prefer my PR over this one #8028

ddstreetmicrosoft:
> I would really prefer my PR over this one #8028

specifically, i don't want to see the mariner-rpm-macros package get *bigger*, it should get split up into separate packages. And the macros.dist file absolutely should belong to the azurelinux-release package, not mariner-rpm-macros (or a subpackage).

toolkit/scripts/rpmops.sh
dmcilvaney: Hmm, I realize now that the toolchain call to srpmpacker needs these as well

dmcilvaney: Ok, we need to define these explicitly in the go tools. I noticed we only call `DefaultDefines()` in one place (srpmpacker), and we use the distro tag version everywhere else. I unified the two into a single function and removed the special case handling from srpmpacker.

dmcilvaney:
> > I would really prefer my PR over this one #8028
> 
> specifically, i don't want to see the mariner-rpm-macros package get _bigger_, it should get split up into separate packages. And the macros.dist file absolutely should belong to the azurelinux-release package, not mariner-rpm-macros (or a subpackage).

I'm happy to merge these changes together somehow, I'm not married to the exact implementation. Especially given I've now realized that the tools will need to define these macros independently of the actual macro file (at least for the toolchain build) the exact location of the macros isn't important to me anymore.

dmcilvaney:
@ddstreetmicrosoft Ok, what I'll do is remove my changes to `macros.dist`. The build tools will be setting the values of these macros anyways, so the exact location/source of the file isn't critical. 

The one wrinkle there is that we need to set all the macros we care about inside the tools. How many of the new dist macros are critical? I'll leave `azl3 1` and `azl 3` defined (unless you are really keen on `azure`?), but we might need to add more to the go tools in the future.

PawelWMS:
> > > I would really prefer my PR over this one #8028
> > 
> > 
> > specifically, i don't want to see the mariner-rpm-macros package get _bigger_, it should get split up into separate packages. And the macros.dist file absolutely should belong to the azurelinux-release package, not mariner-rpm-macros (or a subpackage).
> 
> I'm happy to merge these changes together somehow, I'm not married to the exact implementation. Especially given I've now realized that the tools will need to define these macros independently of the actual macro file (at least for the toolchain build) the exact location of the macros isn't important to me anymore.

I think I also like the idea of `macros.dist` coming from `azurelinux-release`. If we're going to make that switch, however, we have to keep a few things in mind:
- Current "source of truth" for the `dist` macro is the `DIST_TAG` variable defined in `toolkit/scripts/build_tag.mk`. It's used during the toolchain build (when no packages are available, yet) and it determines the value of `dist` everywhere else. One of the features, that it's meant to support, is to allow non-core Azure Linux teams to build packages with their own `dist` tag. We should make sure we don't introduce regressions here.
- If we want to make use of the new macros defined in `azurelinux-release`, we'd need to either add that package to the toolchain or extend the logic inside `rpm.go:DefaultDefines()` to make sure they are applied during packing of the SRPMs. Ideally, I'd prefer to rely on the `macros.dist` and not toolkit's magic as we do now, but that's a lot bigger tooling change, which I doubt we'll be able to handle right now. Or at least I wouldn't try to bundle it all up into one change.

With that in mind and Daniel's revert of the `mariner-rpm-macros.spec` changes in this PR my suggestion is:
- Take this PR first, since it's a smaller change, which will at least help us establish one version of the `azl`/`azure` macros across our specs.
- Rebase the `azurelinux-release` PR on top of this change, since it's more involved considering the two points I mentioned above.

What do you think?

ddstreetmicrosoft:
> > > > I would really prefer my PR over this one #8028
> > > 
> > > 
> > > specifically, i don't want to see the mariner-rpm-macros package get _bigger_, it should get split up into separate packages. And the macros.dist file absolutely should belong to the azurelinux-release package, not mariner-rpm-macros (or a subpackage).
> > 
> > 
> > I'm happy to merge these changes together somehow, I'm not married to the exact implementation. Especially given I've now realized that the tools will need to define these macros independently of the actual macro file (at least for the toolchain build) the exact location of the macros isn't important to me anymore.
> 
> I think I also like the idea of `macros.dist` coming from `azurelinux-release`. If we're going to make that switch, however, we have to keep a few things in mind:
> 
> * Current "source of truth" for the `dist` macro is the `DIST_TAG` variable defined in `toolkit/scripts/build_tag.mk`. It's used during the toolchain build (when no packages are available, yet) and it determines the value of `dist` everywhere else. One of the features, that it's meant to support, is to allow non-core Azure Linux teams to build packages with their own `dist` tag. We should make sure we don't introduce regressions here.
> * If we want to make use of the new macros defined in `azurelinux-release`, we'd need to either add that package to the toolchain or extend the logic inside `rpm.go:DefaultDefines()` to make sure they are applied during packing of the SRPMs. Ideally, I'd prefer to rely on the `macros.dist` and not toolkit's magic as we do now, but that's a lot bigger tooling change, which I doubt we'll be able to handle right now. Or at least I wouldn't try to bundle it all up into one change.
> 
> With that in mind and Daniel's revert of the `mariner-rpm-macros.spec` changes in this PR my suggestion is:
> 
> * Take this PR first, since it's a smaller change, which will at least help us establish one version of the `azl`/`azure` macros across our specs.

If this is *only* changing the toolkit then go for it, as I said I don't care about changes to the toolkit. Please remove the systemd.spec changes though.

> * Rebase the `azurelinux-release` PR on top of this change, since it's more involved considering the two points I mentioned above.
> 
> What do you think?

ddstreetmicrosoft:
> > @ddstreetmicrosoft Ok, what I'll do is remove my changes to `macros.dist`. The build tools will be setting the values of these macros anyways, so the exact location/source of the file isn't critical.
> 
> I don't particularly care what the toolkit does, as long as it results in rpms that are identical to what a 'normal' (i.e. non-toolkit) build produces.
> 
> > The one wrinkle there is that we need to set all the macros we care about inside the tools. How many of the new dist macros are critical? I'll leave `azl3 1` and `azl 3` defined (unless you are really keen on `azure`?), but we might need to add more to the go tools in the future.
> 
> %azure only please. No %azl nor %azl3.

To clarify - %azl seems like we're doing the acronym thing like %rhel, but while %rhel shortens 'red hat enterprise linux' and there isn't really any part of that that makes sense standalone (e.g. %redhat or %enterprise wouldn't really make sense as far as referring to a specific product), we have 'azure linux' so %azl isn't even acronymming correctly, and we also can simply use %azure like %fedora does - e.g. 'fedora linux' uses %fedora' and 'azure linux' can use %azure. That's what makes the most sense to me. But if we really don't want to use %azure then just %azl is the next best option IMHO. For sure it makes no sense at all to have %azl3.

Guess we should also see what @christopherco @jslobodzian think

dmcilvaney:
I've... broken GitHub...
Commits on my branch aren't showing up.

ddstreetmicrosoft:
> I feel like `%azure` is too broad, but that's just my personal feelings. I'll get an opinion from Chris/Jon and go from there.
> 
> As for `azl3`, I based my macros on https://docs.fedoraproject.org/en-US/packaging-guidelines/DistTag/#_conditionals, specifically the fedora `%{fc#}` and `%{el#}`. Probably not super critical for our core packages since we split our releases into separate branches today, but it would be handy for down stream teams building packages, or maybe in the future if we split our packages away. This allows a single spec to easily target multiple versions.

I don't have any problem with defining %azl3, I just think *using* it is a mistake in 99% of cases (instead, the unversioned %azure or %azl macro should be used).

dmcilvaney:
> I don't have any problem with defining %azl3, I just think _using_ it is a mistake in 99% of cases (instead, the unversioned %azure or %azl macro should be used).

100% agree, seems very specialized and personally I'm always going to recommend using the numerical value of `azl` ( ~`%if %{azl} > 2` ).

If we want to omit it I'm certainly not too hung up on it, if you don't see any real value I'll go ahead and remove it. Any maybe by the time I'm done GH will fix itself...

dmcilvaney:
Had a chat, going with `azl`, but I'll drop `azl#`.

ddstreetmicrosoft (approved):
LGTM but probably should get a review from someone else more familiar with the toolkit also

toolkit/scripts/rpmops.sh
```diff
@@ -14,7 +14,17 @@ fi
 
 # Additional macros required to parse spec files.
 DIST_TAG=$(make -s -f $REPO_ROOT/toolkit/Makefile get-dist-tag)
-DEFINES=(-D "with_check 1" -D "dist $DIST_TAG")
+RELEASE_NUM=$(make -s -f $REPO_ROOT/toolkit/Makefile get-release-major)
+
+# Define 'azl <ver>' macro for use in spec files. Normally they would be defined in the 'azurelinux-release' package.
+# Dist tag will be of the form '.<distro><ver>', e.g. '.azl3'., rease number will be of the form '3.0'.
```
PawelWMS, line 20:
```suggestion
# Dist tag will be of the form '.<distro><ver>', e.g. '.azl3'., relase number will be of the form '3.0'.
```

toolkit/scripts/toolchain/check_manifests.sh
```diff
@@ -8,19 +8,30 @@ TOOLCHAIN_SPEC_LIST_FILE=$1
...
+distro_ver="${distro_name} ${distro_majver}"
+
+distro_defines=(--define "with_check 1" --define "dist $DIST_TAG" --define "$distro_ver")
 
 write_rpms_from_spec () {
     # $1 = spec file
     # $2 = file to save to
     spec_dir=$(dirname $1)
-    exclusiveArch=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="dist $DIST_TAG" --qf="[%{EXCLUSIVEARCH} ]" --srpm 2>/dev/null)
+    exclusiveArch=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" "${distro_defines[@]}" --qf="[%{EXCLUSIVEARCH} ]" --srpm 2>/dev/null)
```
PawelWMS, line 28: Now we're changing the behaviour here a bit. We're starting with `--define="with_check 0"` and then pass `--define "with_check 1"` insice `distro_defines`. I'm not sure, if anything may break but if we're making that change anyway, I'm curious, if we could switch to calling `mariner_rpmspec` altogether. This way the logic setting the defines would be kept in one place less.

toolkit/tools/internal/rpm/rpm.go
```diff
@@ -243,16 +247,29 @@ func executeRpmCommandRaw(program string, args ...string) (stdout string, err er
 	return
 }
 
-// DefaultDefinesWithDist returns a new map of default defines that can be used during RPM queries that also includes
-// the dist tag.
-func DefaultDefinesWithDist(runChecks bool, distTag string) map[string]string {
-	defines := DefaultDefines(runChecks)
-	defines[DistTagDefine] = distTag
+// DefaultDistroDefines returns a new map of default defines that can be used during RPM queries that also includes
+// the distro tags like '%dist', '%zal', '%alz<VER>'.
```
PawelWMS, line 251:
```suggestion
// the distro tags like '%dist', '%azl'.
```

toolkit/tools/internal/rpm/rpm.go
```diff
@@ -243,16 +247,29 @@ func executeRpmCommandRaw(program string, args ...string) (stdout string, err er
...
+		// distTag is of the form ".<distro><version>", extract the distro and version
+		// This will be used to generate '%<distro> <version>'
+		distTagSplit := distTagRegex.FindStringSubmatch(distTag)
+		if len(distTagSplit) == 3 {
+			distroName := distTagSplit[1]
+			distroVersion := distTagSplit[2]
+			defines[distroName] = distroVersion
+		} else {
+			logger.Log.Warnf("Invalid distro tag (%s), won't generate derived distro defines", distTag)
+		}
```
PawelWMS, line 266: I'm not convinced about this part for these reasons:
- We extract `distroName` from `distTag`, so there's no guarantee it's always going to match the hard-coded value of `azl` in our specs.
- We expect the distro name and version to be set correctly for our specs to work, so the warning log is too weak in my opinion.

Ideas:
1. We could turn the warning into an error. That would be a lot of pluming changes in the code, however, and it would not help with making sure we always land with `azl` as the distro name. And I'd like to avoid hard-coding `azl`, since that would create another place, which would require an update in case of any re-branding changes in the future.
2. Since `distTag`, `distroName` and `distroVersion` are something we want to control globally for the whole repository (right now from the toolkit with a potential switch to reading the values provided by a package providing `system-release`), this looks similar to what we're doing with `exe.go::ToolkitVersion`. What do you think about compiling `distTag`, `distroName` and `distroVersion` into our tooling and completely dropping the `distTag` argument from `DefaultDistroDefines`? This would require either splitting `DIST_TAG` inside our makefiles a bit or parsing it to extract the other values like you do it here (but also immediately failing in case the format is incorrect).

dmcilvaney: I think we still want to have DIST_TAG configurable, but the other macros I agree can be hard-coded.

toolkit/scripts/build_tag.mk
```diff
@@ -34,6 +33,15 @@ RELEASE_VERSION := $(RELEASE_VERSION)
...
 IMAGE_TAG          ?=
 
+# Basis of the distro macros. The abreveation of the distro name and the major version are encoded into the go tools so
+# that they can be used to create the distro specific macro '<DIST_NAME_ABRV> <DIST_MAJOR_ID>'.
+DIST_NAME_MACRO     ?= azl
+dist_major_macro    := $(shell echo $(RELEASE_MAJOR_ID) | cut -d'.' -f1)
+distro_macro        := $(DIST_NAME_MACRO) $(dist_major_macro)
+
+# Dist tag will be of the form '.<ABRV><MAJOR_ID>', unless overridden by the user.
+DIST_TAG            ?= .$(DIST_NAME_MACRO)$(DIST_MAJOR_ID)
```
PawelWMS, line 43: Consistency nit-pick: you use "ABRV", "MACRO" interchangeably making it a bit confusing to understand what's going on. Also, I think the comment above `DIST_TAG` is not adding any more information, since it's only visible here right above the actual value for `DIST_TAG`. Saying `<ABRV><MAJOR_ID>` in the comments while passing `$(DIST_NAME_MACRO)$(DIST_MAJOR_ID)` in reality adds to the confusion in my opinion.

What do you think about this:
```suggestion
##help:var:DIST_NAME_ABRV:<dist_name_abrv>=Basis of the distro macros. The abbreviation of the distro name and the major version are encoded into the go tools so that they can be used to create the distro specific macro '<DIST_NAME_ABRV> <dist_major_version_number>', where 'dist_major_version_number' is the major version number extracted from '<RELEASE_MAJOR_ID>'.
DIST_NAME_ABRV            ?= azl
dist_major_version_number := $(shell echo $(RELEASE_MAJOR_ID) | cut -d'.' -f1)
dist_version_macro        := $(DIST_NAME_ABRV) $(dist_major_version_number)

##help:var:DIST_TAG:<dist_tag>=Distribution tag, defines the "dist" macro used by the specs. Default: con.
DIST_TAG            ?= .$(DIST_NAME_ABRV)$(DIST_MAJOR_ID)
```

**NOTE**: I've re-named `distro_macro` to `dist_version_macro` to avoid confusion with `DIST_TAG` and use "dist" consistently.

toolkit/scripts/toolchain/check_manifests.sh
```diff
@@ -8,19 +8,19 @@ TOOLCHAIN_SPEC_LIST_FILE=$1
 SPECS_DIR=$2
 MANIFESTS_DIR=$3
 DIST_TAG=$4
-ARCH=$5
+DISTRO_MACRO=$5
```
PawelWMS, line 11: Are we still using `ARCH`? It looks like we're still passing it as an argument.
```suggestion
DISTRO_MACRO=$5
ARCH=$6
```

toolkit/scripts/toolchain/check_manifests.sh
```diff
@@ -8,19 +8,19 @@ TOOLCHAIN_SPEC_LIST_FILE=$1
...
 
 write_rpms_from_spec () {
     # $1 = spec file
     # $2 = file to save to
     spec_dir=$(dirname $1)
-    exclusiveArch=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="dist $DIST_TAG" --qf="[%{EXCLUSIVEARCH} ]" --srpm 2>/dev/null)
+    exclusiveArch=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="$DIST_TAG" --define="$DISTRO_MACRO" --qf="[%{EXCLUSIVEARCH} ]" --srpm 2>/dev/null)
```
PawelWMS, line 17: `DIST_TAG` only has the value, not the name of the macro.
```suggestion
    exclusiveArch=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="dist $DIST_TAG" --define="$DISTRO_MACRO" --qf="[%{EXCLUSIVEARCH} ]" --srpm 2>/dev/null)
```

toolkit/scripts/toolchain/check_manifests.sh
```diff
@@ -8,19 +8,19 @@ TOOLCHAIN_SPEC_LIST_FILE=$1
...
     if [[ -n "$exclusiveArch" && ! "$exclusiveArch" =~ "$ARCH" ]]; then
         return 0
     fi
 
-    version=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="dist $DIST_TAG" --qf="%{VERSION}" --srpm 2>/dev/null)
-    rpmWithoutExtension=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="dist $DIST_TAG" --target=$ARCH --qf="%{nvra}\n" 2>/dev/null)
+    version=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="$DIST_TAG" --define="$DISTRO_MACRO" --qf="%{VERSION}" --srpm 2>/dev/null)
+    rpmWithoutExtension=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="$DIST_TAG" --define="$DISTRO_MACRO" --target=$ARCH --qf="%{nvra}\n" 2>/dev/null)
```
PawelWMS, line 23:
```suggestion
    version=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="dist $DIST_TAG" --define="$DISTRO_MACRO" --qf="%{VERSION}" --srpm 2>/dev/null)
    rpmWithoutExtension=$(rpmspec -q $1 --define="with_check 0" --define="_sourcedir $spec_dir" --define="dist $DIST_TAG" --define="$DISTRO_MACRO" --target=$ARCH --qf="%{nvra}\n" 2>/dev/null)
```

toolkit/tools/internal/rpm/rpm.go
```diff
@@ -89,8 +91,19 @@ var (
 	//
 	//	D: ========== +++ systemd-devel-239-42.azl3 x86_64-linux 0x0
 	installedRPMRegex = regexp.MustCompile(`^D: =+ \+{3} (\S+) (\S+)-linux.*$`)
+
+	// For most use-cases, the distro name abbreviation and major version are set by the exe package. However, if the
+	// module is used outside of the main Mariner build system, the caller can override these values with SetDistroMacros().
+	distNameAbreviation = exe.DistroNameAbreviation
+	distMajorVersion    = exe.DistroMajorVersion
```
PawelWMS, line 98: We could now also move the error checks inside `DefaultDistroDefines` to here, so that they run when the application gets started. This way we'd fail faster for invalid values and avoid having to push the `error` through the code.

Something along these lines:
```suggestion
	distNameAbreviation = exe.DistroNameAbreviation
	distMajorVersion    = ldFlagToPositiveInt(exe.DistroMajorVersion, "DistroMajorVersion")
	
	func ldFlagToPositiveInt(name, value string) int {
	result, err := strconv.Atoi(name)
	if err != nil || result <= 0 {
		err = fmt.Errorf("failed to parse Go ldflag '%s'='%s'. The value must be a positive integer", name, value)
		panic(err)
	}

	return result
}
```

toolkit/scripts/build_tag.mk
dmcilvaney: I was changing a few bits here, merged your suggestions in.

toolkit/scripts/toolchain/check_manifests.sh
dmcilvaney: Good catch, mangled that one during a revert I think.

toolkit/scripts/tools.mk
```diff
@@ -91,6 +91,11 @@ clean-go-tools:
 	rm -rf $(TOOL_BINS_DIR)
 	rm -rf $(BUILD_DIR)/tools
 
+go_ldflags := 	-X github.com/microsoft/CBL-Mariner/toolkit/tools/internal/exe.ToolkitVersion=$(RELEASE_VERSION) \
+				-X github.com/microsoft/CBL-Mariner/toolkit/tools/pkg/imagecustomizerlib.ToolVersion=$(IMAGE_CUSTOMIZER_FULL_VERSION) \
+				-X github.com/microsoft/CBL-Mariner/toolkit/tools/internal/exe.DistroNameAbreviation=$(DIST_NAME_MACRO) \
```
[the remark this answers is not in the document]
dmcilvaney: Handled locally across all instansces

toolkit/tools/pkg/specreaderutils/specreaderutil.go
```diff
@@ -296,8 +296,9 @@ func readSpecWorker(requests <-chan string, results chan<- *parseResult, cancel
 
 	defer wg.Done()
 
-	noCheckDefines := rpm.DefaultDefinesWithDist(false, distTag)
-	checkDefines := rpm.DefaultDefinesWithDist(true, distTag)
+	// These variables are common to all specs and can be reused, but we need to cache the error messages.
```
dmcilvaney, line 299:
```suggestion
```

toolkit/scripts/build_tag.mk
```diff
@@ -34,6 +33,14 @@ RELEASE_VERSION := $(RELEASE_VERSION)
 # Image tag - empty by default. Does not apply to the initrd.
 IMAGE_TAG          ?=
 
+##help:var:DIST_NAME_ABRV:<dist_name_abrv>=Basis of the distro macros. The abbreviation of the distro name and the major version are encoded into the go tools so that they can be used to create the distro specific macro '<DIST_NAME_ABRV> <dist_major_version_number>', where 'dist_major_version_number' is the major version number extracted from '<RELEASE_MAJOR_ID>'.
+DIST_NAME_ABRV       ?= azl
+dist_major_version_number := $(shell echo $(RELEASE_MAJOR_ID) | cut -d'.' -f1)
+DIST_VERSION_MACRO := $(DIST_NAME_ABRV) $(dist_major_version_number)
+
+##help:var:DIST_TAG:<dist_tag>=Distribution tag, defines the "dist" macro used by the specs. Default: '.<DIST_NAME_ABRV><dist_major_version_number>'
```
dmcilvaney, line 41:
```suggestion
##help:var:DIST_TAG:<dist_tag>=Distribution tag, defines the "dist" macro used by the specs. Default: '.<DIST_NAME_ABRV><dist_major_version_number>' e.g., ".azl3"
```

cwize1 (approved):

toolkit/tools/internal/rpm/rpm.go
```diff
@@ -89,8 +91,48 @@ var (
...
+	distNameAbreviation, distMajorVersion = loadLdDistroFlags()
 )
 
+// checkDistroMacros validates the distro macro values.
+func checkDistroMacros(nameAbreviation string, majorVersion int) (string, int, error) {
+	if majorVersion < 0 || nameAbreviation == "" {
+		err := fmt.Errorf("failed to set distro defines (%s, %d), invalid name or version", nameAbreviation, majorVersion)
+		return "", 0, err
+	}
+	return nameAbreviation, majorVersion, nil
```
PawelWMS, line 106: We're not setting or updating these values here, so no need to return them. I think we might as well completely drop this function and just move the `if majorVersion < 0 || nameAbreviation == ""` directly to the caller, since there's only one. Alternatively, the function could return a boolean and leave it to the caller to construct a proper error message.

dmcilvaney: Both `loadLdDistroFlags()` and `SetDistroMacros()` call it, but I agree it doesn't need to return the values (hold over from when it was calling atoi).

toolkit/tools/internal/rpm/rpm_test.go
```diff
@@ -199,3 +201,56 @@ func TestExtractNameFromRPMPath(t *testing.T) {
 		})
 	}
 }
+
+func TestDistroDefines(t *testing.T) {
```
PawelWMS, line 205: Would it be possible to test one scenario per test function? It's going to be clearer what we're checking for and what exactly failed in case of a failure.

PawelWMS: Basically, instead of a comment line for each test case in this one function, my suggestion is to have a separate test function with a name that reflects what it's testing.

PawelWMS (approved):

PR #8015, opened 2024-02-21
add SymCrypt and SymCrypt-OpenSSL to images with openssl

tobiasb-ms:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 
Add `SymCrypt` and `SymCrypt-OpenSSL` as `Recommends` packages for `openssl` and add them to images with `openssl`.

Note that an initial version of this commit with incorrect capitalization on `SymCrypt-OpenSSL` was merged in #7977 and then reverted in #8001.

In AZL 3 `openssl` relies on `SymCrypt` for `FIPS` mode. As such, we want `SymCrypt` and the related `SymCrypt-OpenSSL` packages installed where `openssl` is installed. To accomplish this, we do two things:
1. Add these packages as `Recommends` in the `openssl` spec. This means that if the package manager can find these packages, it will install them when `openssl` is installed, but it won't refuse to install `openssl` if it can't find them.
2. Add these packages to any image that has `openssl`.

One may wonder why we didn't simply add these packages as `Requires` to `openssl`. The issue here is that `SymCrypt` has a `BuildRequires` on `python`, which in turn requires `openssl`, so there's a circular dependency.

###### Change Log  
- Add `SymCrypt` and `SymCrypt-OpenSSL` as `Recommends` in the `openssl` spec file.
- Add `SymCrypt` and `SymCrypt-OpenSSL` to any image that already has `openssl`
- 
**YES**

###### Test Methodology
- Tested locally to make sure the relevant `Recommends` and `Requires` gets into the built rpms.
- Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=510542&view=results

### DIFF
--- a/SPECS/core-packages/core-packages.spec
+++ b/SPECS/core-packages/core-packages.spec
@@ -71,6 +71,8 @@ Requires:       rpm
 Requires:       rpm-libs
 Requires:       sed
 Requires:       sqlite-libs
+Requires:       SymCrypt
+Requires:       SymCrypt-OpenSSL
 Requires:       tdnf
 Requires:       tdnf-plugin-repogpgcheck
 Requires:       xz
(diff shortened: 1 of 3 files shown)


dmcilvaney (approved):

mfrw (commented):
/cc @kanikanema for awareness.

PR #8028, opened 2024-02-22
Updates to azurelinux-release

ddstreetmicrosoft:
Several updates to the azurelinux-release package:
- remove %%config(noreplace) from *-release files                                                                                                                                             
- define dist_version and use local macros                                                                                                                                                    
- move *-release and issue files under %%_libdir and make %%_sysconfdir symlinks                                                                                                              
- use consistent creation of here documents                                                                                                                                                   
- add issue.d dir                                                                                                                                                                             
- move macros.dist into this package                                                                                                                                                          
- Add default systemd presets


SPECS/azurelinux-release/90-default.preset
```diff
@@ -0,0 +1,184 @@
...
+enable dmraid-activation.service
+
+enable mdmonitor.service
+enable mdmonitor-takeover.service
+
+enable raid-check.timer
+
+enable spice-vdagentd.service
+
+enable qemu-guest-agent.service
```
romoh, line 93: why is this enabled by default? 
AFAIK the common usage is not qemu VMs. This should only be enabled in the `qemu-guest` references image

ddstreetmicrosoft: These are simply the presets, these don't mean the package will actually be installed.

For `qemu-guest-agent` in particular, you can see it's `WantedBy=dev-virtio\x2dports-org.qemu.guest_agent.0.device` so, *when enabled*, it will only register itself to actually be *started* when that virtual device is detected in the system. Meaning that even if the package is installed onto a system, and the service is enabled, it won't actually get started unless it's actually a qemu guest instance.

I have no particularly strong feeling about any of the presets (taken mostly from fedora's presets), and I also have no problem with specific image builds adding (or overriding) any presets. However this one in particular seems to make sense to leave enabled here, unless I'm missing something.

nicogbg (approved):

SPECS/mariner-rpm-macros/mariner-rpm-macros.signatures.json
```diff
@@ -12,7 +12,6 @@
   "gpgverify": "db0e050f56b694497d70603a6f5c17dd60ddbcf7cee670616851cd389f6767c4",
   "macros": "b7ad5c17d6ce105427eeef54f43317cb1cc8cc114bfeca7693069242ad747d20",
   "macros.check": "79367176c3c7d10c0158b6e5d881e0fc3c8fd50c5957dad2f097c2d4a37833e7",
-  "macros.dist": "817653f151349adff8c658143cf01ad1f8b51168be2087e4e02778224da85d63",
```
PawelWMS, line 15: Does the full toolchain and packages build show no regressions after this change? I believe we should be fine but want to make sure.

ddstreetmicrosoft: here's a fresh pipeline run:
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=513845&view=results

SPECS/systemd/systemd.spec
```diff
@@ -1,10 +1,4 @@
 
-# Temporary until azure-release-common is updated to carry macros.dist that defines %azure
-%if %{defined azure}
-%{error:Once azure macro is defined in macros.dist, please remove this}
-%endif
-%global azure 3
```
PawelWMS, line 6: Does `systemd` still build correctly after this change? I think it doesn't have the `azurelinux-release` package present during the build, so the `azure` macro may not be set.

dmcilvaney: With the changes from https://github.com/microsoft/CBL-Mariner/pull/8013 we will also have these defined by the build tools explicitly (but agreed the buildrequires should also cover it).

SPECS/azurelinux-release/azurelinux-release.spec
```diff
@@ -1,67 +1,125 @@
...
+ln -sv ..%{_libdir}/issue.net %{buildroot}%{_sysconfdir}/issue.net
+
+install -d -m 755 %{buildroot}%{_sysconfdir}/issue.d
+
+cat <<-"EOF" > %{buildroot}%{_rpmmacrodir}/macros.dist
+# dist macros.
+
+%%__bootstrap         ~bootstrap
+%%azure               %{dist_version}
+%%azl%{dist_version}  1
```
dmcilvaney, line 79: We decided that `azl<ver> 1` does not give much value over `azl <ver>`
```suggestion
%%azl                %{dist_version}
```

SPECS/systemd/systemd.spec
PawelWMS: Adding the BR would cause a circular dependency between `systemd` and `azurelinux-release`. With the changes from #8013, which inject the `azl` and `dist` macros, I think we should be fine with building `systemd` the way it's defined in this commit. We need to work on unifying the source of truth for these macros but that's completely out of scope for this PR and #8013.

PawelWMS (commented):
I really like these changes. The only reason I haven't approved yet, is that we probably need to wait for #8013 to define the `azl` macro `systemd.spec` depends. For now, unfortunately, we can't deliver it from `azurelinux-release` for that spec. Once that's in and `systemd` builds, I think we'll be good to go.

PawelWMS (approved):
Changes look good. The full build you linked a while earlier failed due to missing signatures, though. The signatures are definitely here in this version of the PR, so I hope that's not going to fail anymore. I just started [another build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516648&view=results). Before you merge, please make sure there are no regressions in the toolchain and package builds.

ddstreetmicrosoft:
> Changes look good. The full build you linked a while earlier failed due to missing signatures, though. The signatures are definitely here in this version of the PR, so I hope that's not going to fail anymore. I just started [another build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516648&view=results). Before you merge, please make sure there are no regressions in the toolchain and package builds.

unfortunately that failed due to a mismatch in the nftables signatures, but it did get through the toolchain build.

I rebased and added a commit to fix the nftables signatures also.

christopherco (approved):

PR #7618, opened 2024-02-01
feat: EnableGrubMkconfig by default; deprecate legacy tools-grub.cfg flow

Camelron:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 

What does the PR accomplish, why was it needed?
Remove the old way of installing our /boot/grub2/grub.cfg; instead generate everything by grub2-mkconfig by default. In our imageconfigs, EnableGrubMkconfig is now `true` by default.  

###### Change Log  

- Remove legacy /boot/grub2/grub.cfg and its related artifacts (/boot/mariner.cfg.. etc)
- Remove $bootprefix-related logic from /boot/efi/boot/grub2/grub.cfg. Instead populate that information directly in $prefix (the variable used natively by grub2-mkconfig). 
- Make EnableGrubMkconfig the standard behavior for all images, differentiate between standard image configs and container configs with config.IsContainerImage
- Automatically install grub2 on image configurations that use grub2-mkconfig in a similar fashion to how kernel is added. 

###### Does this affect the toolchain?  

YES

###### Test Methodology

- Local build of core-efi, verify that the produced vhd boots and can regenerate its grub.cfg
- pipelines running.

PR #8029, opened 2024-02-22
Remove installkernel package since it is provided by systemd-udev

ddstreetmicrosoft:
The 'installkernel' script is provided by systemd-udev; this gets rid of the old custom 'installkernel' package.


ddstreetmicrosoft:
@christopherco @rlmenge @Camelron @BettyRain I think we'll need to evaluate what currently is *using* installkernel, as the one provided by systemd-udev probably isn't doing exactly everything our custom one is doing - for example, the specific filename of the vmlinuz and initrd, and our custom script is dropping the hardcoded /boot/linux-*.cfg file that our hardcoded grub.cfg uses.

*However* our kernel rpm also does some/all of that stuff, so I think the 'installkernel' script is only getting used by manual runs of 'make install' from a kernel build. Anyone know for sure if anything in our kernel package and/or image build is actually using 'installkernel' for anything?

christopherco:
> _However_ our kernel rpm also does some/all of that stuff, so I think the 'installkernel' script is only getting used by manual runs of 'make install' from a kernel build. Anyone know for sure if anything in our kernel package and/or image build is actually using 'installkernel' for anything?

Correct. Our `installkernel` package & script is only used in manual runs of `make install`. It is not used by our kernel package or image builds.

Camelron:
> @christopherco @rlmenge @Camelron @BettyRain I think we'll need to evaluate what currently is _using_ installkernel, as the one provided by systemd-udev probably isn't doing exactly everything our custom one is doing - for example, the specific filename of the vmlinuz and initrd, and our custom script is adding the hardcoded /boot/linux-*.cfg file that our hardcoded grub.cfg uses.
> 
> _However_ our kernel rpm also does some/all of that stuff, so I think the 'installkernel' script is only getting used by manual runs of 'make install' from a kernel build. Anyone know for sure if anything in our kernel package and/or image build is actually using 'installkernel' for anything?

The Mariner-specific /boot/linux-*.cfg has been removed in https://github.com/microsoft/CBL-Mariner/pull/7618

Camelron (approved):

christopherco (approved):

PR #8044, opened 2024-02-23
cal10n: upgrade to 0.8.1

miz060:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Update cal10n to v0.8.1

###### Change Log  

- Update cal10n to v0.8.1

**NO**

###### Test Methodology

- BuddyBuild: [549557](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=549557&view=results)


SPECS/cal10n/cal10n.spec
```diff
@@ -16,16 +15,15 @@
...
+Release:        1%{?dist}
 License:        MIT
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 Group:          Development/Libraries/Java
 URL:            http://cal10n.qos.ch
-Source0:        https://github.com/qos-ch/cal10n/archive/refs/tags/v_%{version}.tar.gz#/%{name}-%{version}.tar.gz
-Source1:        build.xml-0.7.7.tar.xz
-Patch0:         cal10n-0.7.7-sourcetarget.patch
+Source0:        %{_distro_sources_url}/cal10n-0.8.1.tar.xz
```
anphel31, line 25: These should point to the upstream source URL, not our mirror of the sources (`%{_distro_sources_url}`) / `https://cblmarinerstorage.blob.core.windows.net/sources/core`

cgmanifest.json
```diff
@@ -1417,8 +1417,8 @@
         "type": "other",
         "other": {
           "name": "cal10n",
-          "version": "0.7.7",
-          "downloadUrl": "https://github.com/qos-ch/cal10n/archive/refs/tags/v_0.7.7.tar.gz"
+          "version": "0.8.1",
+          "downloadUrl": "https://cblmarinerstorage.blob.core.windows.net/sources/core/cal10n-0.8.1.tar.xz"
```
anphel31, line 1421: How was this source generated? Can we link to the upstream source if it's available?

anphel31 (changes requested):

SPECS/cal10n/cal10n.spec
miz060: I updated this spec based on openSUSE's v0.8.1 spec, but their source tarball has an extra folder compared to the upstream source (https://github.com/qos-ch/cal10n/archive/v_0.8.1.tar.gz). 

Do you happen to know how openSUSE generates their source tarballs?

cgmanifest.json
miz060: It's borrowed from openSUSE's cal10n spec. Their source tarball has an extra folder that makes it work with their build.tar.xz file. The upstream source does not include this folder and would cause the build to fail. 
Is it ok for us to keep the current setup, or we would need to switch to a maven build instead?

SPECS/cal10n/cal10n.spec
```diff
@@ -16,16 +15,15 @@
...
 License:        MIT
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 Group:          Development/Libraries/Java
 URL:            http://cal10n.qos.ch
-Source0:        https://github.com/qos-ch/cal10n/archive/refs/tags/v_%{version}.tar.gz#/%{name}-%{version}.tar.gz
-Source1:        build.xml-0.7.7.tar.xz
-Patch0:         cal10n-0.7.7-sourcetarget.patch
+Source0:        https://code.opensuse.org/adrianSuSE/%{name}/blob/factory/f/%{name}-%{version}.tar.xz
+Source1:        %{name}-build.tar.xz
```
PawelWMS, line 26: Please also clarify where this tarball is coming from and how to recreated it later, if needer or during an upgrade.

miz060: Good point. Added comment and changelog entry to point to the source.

> Please also clarify where this tarball is coming from and how to recreated it later, if needer or during an upgrade.

anphel31 (approved):

SPECS/cal10n/cal10n.spec
```diff
@@ -16,16 +15,16 @@
...
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 Group:          Development/Libraries/Java
 URL:            http://cal10n.qos.ch
-Source0:        https://github.com/qos-ch/cal10n/archive/refs/tags/v_%{version}.tar.gz#/%{name}-%{version}.tar.gz
-Source1:        build.xml-0.7.7.tar.xz
-Patch0:         cal10n-0.7.7-sourcetarget.patch
+Source0:        https://code.opensuse.org/adrianSuSE/%{name}/blob/factory/f/%{name}-%{version}.tar.xz
+# cal10n-build.tar.gz imported from https://code.opensuse.org/adrianSuSE/cal10n/blob/factory/f/cal10n-build.tar.xz
+Source1:        %{name}-build.tar.xz
```
PawelWMS, line 27: If it's coming from an upstream URL, then you can use the same format as for `Source0`. Having said that, can you add a comment explaining what this source is used for, so it's easier for the next person to understand how this package is built and what they may need to look into during a future update?
```suggestion
Source1:        https://code.opensuse.org/adrianSuSE/cal10n/blob/factory/f/%{name}-build.tar.xz
```

SPECS/cal10n/cal10n.spec
```diff
@@ -73,58 +74,54 @@ http://cal10n.qos.ch/manual.html
...
+mkdir -p lib
+build-jar-repository -s lib \
+%if %{with tests}
+    ant-antunit \
+%endif
+    ant/ant
+%{ant} \
+%if %{without tests}
+    -Dtest.skip=true \
+%endif
```
PawelWMS, line 86: We don't use `with tests` but the `with_check` macro. Please update the spec - right now we'd never run the `with test` branch.

miz060: Got it. Updated.

SPECS/cal10n/cal10n.spec
miz060: Got it. Updated source1 and added comment.

SPECS/cal10n/cal10n.spec
```diff
@@ -73,58 +74,54 @@ http://cal10n.qos.ch/manual.html
...
+build-jar-repository -s lib \
+%if %{with_check}
+    ant-antunit \
+%endif
+    ant/ant
+%{ant} \
+%if %{without tests}
+    -Dtest.skip=true \
+%endif
+    package javadoc
```
PawelWMS, line 87: We're closer but we still have `without tests` and want to use `?`:
```suggestion
%if 0%{?with_check}
    ant-antunit \
%endif
    ant/ant
%{ant} \
%if ! 0%{?with_check}
    -Dtest.skip=true \
%endif
    package javadoc
```

PawelWMS (approved):

Issue #7342, opened 2024-01-19
widespread use of invalid %{with_check} macro prevents using many standard tools with our spec files


ddstreetmicrosoft:
**Describe the bug**

many of our spec files contain the invalid macro %{with_check}, which is not a valid macro and only defined when using our 'make' process. This results in the inability to use any of these invalid spec files with standard rpm tooling (e.g. rpmbuild, spectool, etc).

**To Reproduce**

```
# rpmbuild -bb wget.spec 
error: parse error in expression:  %{with_check}
error:                             ^
error: /usr/src/mariner/SPECS/wget.spec:12: bad %if condition:  %{with_check}

```

ddstreetmicrosoft:
@dmcilvaney @PawelWMS for 3.0 (at least) we need to replace use of this invalid macro with something standard, which will require toolkit changes as well as fixing all the spec files.

fyi @christopherco @jslobodzian @Britel

mfrw:
I sometimes get over it by `rpmbuild -D "with_check 1" .....` {1/0 depending if i want check or not}

dmcilvaney:
Would it be sufficient to define a default in our macros package? Or is the request that we use a standard mechanism (%bcond*) and update the packages to use that? Obviously the macro path is quicker, but would mean that the tools in question either need to explicitly import the mariner macros files (we do this for PR gates on Ubuntu), or be run on Mariner.

PR #8088, opened 2024-02-26
tree-wide: change %{with_check} to 0%{?with_check}

ddstreetmicrosoft:
Fixes: #7342


PawelWMS (commented):
Looks good to me. The CG manifest check failure is unrelated to this change (looks like 4 URLs have gotten stale), so we can override that. The toolchain manifests check needs looking into but once that's done, let me know and I'll override CG manifest check.

ddstreetmicrosoft:
> Looks good to me. The CG manifest check failure is unrelated to this change (looks like 4 URLs have gotten stale), so we can override that. The toolchain manifests check needs looking into but once that's done, let me know and I'll override CG manifest check.

Should be ready now.

PR #8119, opened 2024-02-27
Updates to systemd

ddstreetmicrosoft:
Updates to systemd spec, and add python-pytest-flakes

pipeline build https://dev.azure.com/mariner-org/mariner/_build/results?buildId=515257&view=results


SPECS/systemd/systemd.spec
nisamson, line 1: Should the release number of systemd be bumped?

PawelWMS (approved):

dmcilvaney (approved):

PR #8131, opened 2024-02-28
Image Customizer: Restore legacy grub.cfg file.

cwize1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

The Image Customizer tool needs the ability to customize Mariner 2.0 images. This includes needing the ability to reset the bootloader configuration (e.g. when the partition layout is changed). However, in a previous change (#7618) the legacy grub.cfg file was removed from the toolkit since it is no longer being used in Mariner 3.0. This change restores this file but only for the Image Customizer tool.

###### Change Log  

- Image Customizer: Restore legacy grub.cfg file.

NO

###### Test Methodology

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -989,6 +989,7 @@ func ConfigureDiskBootloader(bootType string, encryptionEnable bool, readOnlyVer
 	partitionSettings []configuration.PartitionSetting, kernelCommandLine configuration.KernelCommandLine,
 	installChroot *safechroot.Chroot, diskDevPath string, installMap map[string]string,
 	encryptedRoot diskutils.EncryptedRootDevice, readOnlyRoot diskutils.VerityDevice, enableGrubMkconfig bool,
+	includeLegacyGrubCfg bool,
 ) (err error) {
 	timestamp.StartEvent("configuring bootloader", nil)
 	defer timestamp.StopEvent(nil)
@@ -1056,7 +1057,7 @@ func ConfigureDiskBootloader(bootType string, encryptionEnable bool, readOnlyVer
 
 	// Grub will always use filesystem UUID, never PARTUUID or PARTLABEL
 	err = InstallGrubDefaults(installChroot.RootDir(), rootDevice, bootUUID, bootPrefix, encryptedRoot,
-		kernelCommandLine, readOnlyRoot, isBootPartitionSeparate)
+		kernelCommandLine, readOnlyRoot, isBootPartitionSeparate, includeLegacyGrubCfg)
 	if err != nil {
 		err = fmt.Errorf("failed to install main grub config file: %s", err)
 		return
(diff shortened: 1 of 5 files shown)


ddstreetmicrosoft (changes requested):
Well...I mean...no? Absolutely not?

We for sure should not be using the legacy hardcoded grub config in azl3 for anything.

cwize1:
> Well...I mean...no? Absolutely not?
> 
> We for sure should not be using the legacy hardcoded grub config in azl3 for anything.

Please read the PR description. This is only for the Image Customizer tool. And only when customizing Mariner 2.0 images.

gmileka (approved):

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -989,6 +989,7 @@ func ConfigureDiskBootloader(bootType string, encryptionEnable bool, readOnlyVer
 	partitionSettings []configuration.PartitionSetting, kernelCommandLine configuration.KernelCommandLine,
 	installChroot *safechroot.Chroot, diskDevPath string, installMap map[string]string,
 	encryptedRoot diskutils.EncryptedRootDevice, readOnlyRoot diskutils.VerityDevice, enableGrubMkconfig bool,
+	includeLegacyGrubCfg bool,
```
romoh, line 992: Can't that be derived from enableGrubMkconfig? I assume we'd have either or?

cwize1: While that might be true, having it as a separate field makes it clear that this is functionality being kept around for the sake of the image customizer. So, I think it is better to keep it as is.

Camelron:
> > Well...I mean...no? Absolutely not?
> > We for sure should not be using the legacy hardcoded grub config in azl3 for anything.
> 
> Please read the PR description. This is only for the Image Customizer tool. And only when customizing Mariner 2.0 images.

One remaining question: under what circumstances are we using the 3.0 version of Image Customizer on a Mariner 2.0 image? Wouldn't it make more sense for them to use MIC 2.0?

cwize1:
> > > Well...I mean...no? Absolutely not?
> > > We for sure should not be using the legacy hardcoded grub config in azl3 for anything.
> > 
> > 
> > Please read the PR description. This is only for the Image Customizer tool. And only when customizing Mariner 2.0 images.
> 
> One remaining question: under what circumstances are we using the 3.0 version of Image Customizer on a Mariner 2.0 image? Wouldn't it make more sense for them to use MIC 2.0?

The Image Customizer executable will be capable of targeting all versions of Mariner that still receive support. There will not be separate builds of the Image Customizer for different versions of Mariner.

The Image Customizer currently builds out of 3.0-dev, which is where all the development work is being merged. When the Image Customizer does official releases, it will fork off a release branch, as documented [here](https://dev.azure.com/mariner-org/ECF/_git/ecf-docs?path=/network-cloud/mib-v2/ReleaseProcess.md&_a=preview).

toolkit/tools/internal/resources/assets/grub2/grub.cfg
```diff
@@ -0,0 +1,26 @@
+set timeout=0
+set bootprefix={{.BootPrefix}}
+search -n -u {{.BootUUID}} -s
+
+load_env -f $bootprefix/mariner.cfg
```
Camelron, line 5: Worth noting that these .cfg files only are present if we are installing 2.0 (.cm2) packages (specifically the kernel, systemd) 

Can we add a sanity check when includeLegacyGrubCfg=true that we are modifying a 2.0 image?

cwize1: Considering that only the Image Customizer is passing in `true`, I think it makes sense to leave the image version detection to the Image Customizer code.

cwize1: And in fact, I'll probably do this as a detection of if the image is using grub-mkconfig.

ddstreetmicrosoft (approved):
i still don't understand why this needs to be in the 3.0 branch, but go for it

dmcilvaney (approved):

Camelron (approved):

PR #8152, opened 2024-02-28
Fix/simplify some of our macro definitions

ddstreetmicrosoft:
+- Remove incorrect code making %check nonfatal (EDIT: this commit removed/deferred)
+- fix %_topdir macro
+- remove %_lib and %_lib64 macros (EDIT: this commit removed/deferred, with added comment to macros file)
+- remove %_docdir
+- remove redundant macros

EDIT: added commit to replace several hardcoded `/usr/src/mariner` uses in the toolkit

See specific commits for details.


SPECS/mariner-rpm-macros/macros.check
```diff
@@ -1,4 +1,3 @@
-# RPM macros for overriding the default behavior of %check to be nonfatal.
```
dmcilvaney, line 1: This needs to wait, we need to make this change inline with the tools/pipelines. (Tools still expect this legacy behaviour and will scan for the `"====== CHECK DONE %{name}. EXIT STATUS $?"` I think).

SPECS/mariner-rpm-macros/macros
```diff
@@ -1,29 +1,17 @@
...
+# Fedora/RHEL have long since adopted the move, while SUSE does have a
+# fallback that causes root (and/or anyone with write access) to use
+# /usr/src/packages, which was the original "system" _topdir. For
+# obvious reasons, it's important for non-root users to be able to
+# build packages (in fact root should generally NEVER be used to build
+# any package), but there may be lots of external code that has
+# hardcoded the /usr/src/mariner directory for building. So this
+# adopts SUSE's approach, while using our /usr/src/mariner directory
+# instead of the traditional /usr/src/packages.
+%_topdir            %{expand:%%global _topdir %{lua:if posix.access(rpm.expand("/usr/src/mariner"), "w") then print "/usr/src/mariner" else print "%{getenv:HOME}/rpmbuild" end} \
```
dmcilvaney, line 13: We will want to do a few tests on this one, make sure it works as expected. I can try some local tools if you queue a full end-to-end build (I'd turn on the raw toolchain boostrap as well, since that might have some knock on effects).

dmcilvaney (changes requested):

SPECS/mariner-rpm-macros/macros.check
ddstreetmicrosoft: ack, i removed this commit from this pr and i'll open a separate PR later for the check-related changes

SPECS/mariner-rpm-macros/macros
ddstreetmicrosoft: here's pipeline run:
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516244&view=results

SPECS/filesystem/filesystem.spec
```diff
@@ -36,6 +36,7 @@ install -vdm 755 %{buildroot}/media
 install -dv -m 0750 %{buildroot}/root
 install -dv -m 1777 %{buildroot}/tmp %{buildroot}/var/tmp
 install -vdm 755 %{buildroot}/usr/{,local/}{bin,include,lib,sbin,src}
+install -vdm 755 %{buildroot}/usr/src/mariner
```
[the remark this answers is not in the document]
ddstreetmicrosoft: > is /usr/src/"mariner" the right thing here? there is rebranding afoot for "azurelinux"

The only reason to leave this is in case there is any end users who have hardcoded this path, and are expecting to build packages in this path.

Otherwise, we don't need the fallback behavior (of root still using /usr/src/mariner) and we should simply use standard topdir location that rpm has provided for over 16 years.

Personally I would *much* rather drop the fallback behavior and go with the rpm standard.

SPECS/filesystem/filesystem.spec
```diff
@@ -660,6 +661,7 @@ return 0
 %dir /usr/share/terminfo
 %dir /usr/share/zoneinfo
 %dir /usr/src
+%dir /usr/src/mariner
```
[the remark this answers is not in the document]
ddstreetmicrosoft: same answer

SPECS/filesystem/filesystem.spec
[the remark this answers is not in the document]
ddstreetmicrosoft: ah i see @PawelWMS is already planning to rename the location so there's no need to keep backwards compatibility; I'll update the PR to completely move to the standard then (i.e. remove our %_topdir override completely)

SPECS/mariner-rpm-macros/macros
ddstreetmicrosoft: new pipeline build with latest push https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516306&view=results

SPECS/filesystem/filesystem.spec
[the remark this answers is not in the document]
ddstreetmicrosoft: latest push has our %_topdir override removed completely

SPECS/mariner-rpm-macros/macros
dmcilvaney: Found a few spots its hard-coded in the tools:
`build_mariner_toolchain.sh`, `build_official_toolchain_rpms.sh`, `toolchain.mk` (aside these should probably just be inheriting from each other if they truly need to be hard-coded...) @anphel31

Also the containerized build env. will need a polish up @neha170 .

I'm supportive of getting this change in as well, but at minimum the toolchain bits will need adjusting.

dmcilvaney:
hash for `macros` should be `2fd9d7600f1c5c1abcf2c7eddba05eb52eca24be60d916163e1fedbac8abac97` by the looks of it.

ddstreetmicrosoft:
> hash for `macros` should be `2fd9d7600f1c5c1abcf2c7eddba05eb52eca24be60d916163e1fedbac8abac97` by the looks of it.

ack, i always forget to update it since the files are checked into git and don't really need to be in the 'signatures' file.

updated in latest push

SPECS/mariner-rpm-macros/macros
dmcilvaney: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516389&view=results

ddstreetmicrosoft: I updated all the places in the toolkit that grep showed it was hardcoded and tried to fix them (but @anphel31 and @neha170 should review it).

And also as you said above, we probably shouldn't be hardcoding rpm build locations into the toolkit, at least not in multiple different places.

ddstreetmicrosoft:
started a new full pipeline buidl https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516812&view=results

ddstreetmicrosoft:
looks like we can't remove our custom %_lib just yet as there's some places hardcoded to /usr/lib (during package build/install, so the merged-usr doesn't help at that point)

new pipeline https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516868

ddstreetmicrosoft:
pipeline build https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516880&view=results

ddstreetmicrosoft:
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516943&view=results

ddstreetmicrosoft:
another pipeline run after the rebase https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516985&view=results

dmcilvaney:
> another pipeline run after the rebase https://dev.azure.com/mariner-org/mariner/_build/results?buildId=516985&view=results

Pipeline failed, but I'm not sure it's related to this PR... @anphel31 I've seen this a few times:
```
time="2024-02-29T18:20:09UTC" + /usr/lib/rpm/mariner/brp-python-bytecompile /usr/bin/python3 1 0
time="2024-02-29T18:20:09UTC" /usr/var/tmp/rpm-tmp.idVeEg: line 561: /usr/lib/rpm/mariner/brp-python-bytecompile: No such file or directory
time="2024-02-29T18:20:09UTC" error: Bad exit status from /usr/var/tmp/rpm-tmp.idVeEg (%install)
```
I don't see anything in filesystem.spec that would have affected this lately...

ddstreetmicrosoft:
another pipeline run https://dev.azure.com/mariner-org/mariner/_build/results?buildId=518491&view=results

@dmcilvaney I think this is just waiting on your review

ddstreetmicrosoft:
another rebase, and another pipeline run: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=523385&view=results

@dmcilvaney i think this is just waiting on your review?

dmcilvaney:
@anphel31 Any idea on this one? https://dev.azure.com/mariner-org/mariner/_build/results?buildId=523388&view=logs&j=db98f19e-da46-5e4d-a4ba-372cf3771a92&t=8831c5ce-9476-5d69-9bf5-bfd7485723f1&l=225727

PR #8158, opened 2024-02-28
systemd: skip sm3 digest in test-openssl

ddstreetmicrosoft:
We don't provide the sm3 digest, so skip it in test-openssl


SPECS/systemd/systemd.spec
```diff
@@ -135,6 +135,9 @@ Patch0490:      use-bfq-scheduler.patch
 # Adjust upstream config to use our shared stack
 Patch0491:      fedora-use-system-auth-in-pam-systemd-user.patch
 
+# Patches for Azure Linux
+Patch: do-not-test-openssl-sm3.patch
```
hbeberman, line 139: Patch should have a number

hbeberman (approved):

PR #8208, opened 2024-03-01
qemu: update glibc-static dep version

ddstreetmicrosoft:


bfjelds (approved):

kanikanema:
Thanks for the fix @ddstreetmicrosoft. The last time I build qemu before merging it was with glibc-static 2.38-1. user-static is turned for azlinux so this shouldn't impact the actual build. Our pre-checks maybe doing a simpler check instead of taking into account the conditions.

PR #8222, opened 2024-03-02
create separate python-pip package

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Separate python-pip from python3.spec. Will allow updating python3 and python-pip versions independently. Also fixes issue where python-pip package version is showing the python3 version (ex: 3.12.0)

###### Change Log  

- Romove python-pip subpackage from python3.spec
- Add new python-pip.spec

**YES**

###### Test Methodology

- X64 toolchain: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=518532&view=results


SPECS/python-pip/python-pip.spec
```diff
@@ -0,0 +1,44 @@
...
+%description -n python3-pip %{_description}
+
+%prep
+%autosetup -n %{srcname}-%{version}
+
+%build
+%py3_build_wheel
+
+%install
+pip3 install --no-cache-dir --no-index --ignore-installed --root %{buildroot} \
```
ddstreetmicrosoft, line 33: does this work? Using `pip3` to install pip into the buildroot? Since this package is what actually provides `pip3` I'm surprised this works (without having to create a "circular dependency")...

I would expect something like `%{python3} dist/%{python_wheel_name}/pip install ...` to use the in-package pip instead of the (non-existent?) system pip...

ddstreetmicrosoft (approved):
one comment on the `%install` section, otherwise lgtm

SPECS/python-pip/python-pip.spec
anphel31: It does work in the toolchain build environment, since the python3.12 installation installs `pip3`. It will probably  not work in the buddybuild/pkggen build environment, though. Let me see if I can modify this to build in both environments properly.

SPECS/python-pip/python-pip.spec
```diff
@@ -0,0 +1,44 @@
...
+
+%files -n python3-pip
+%defattr(-,root,root,755)
+%{_bindir}/pip*
+%{python3_sitelib}/pip*
+
+%changelog
+* Tue Feb 13 2024 Andrew Phelps anphel@microsoft.com - 24.0-1
+- License verified
+- Original version for CBL-Mariner
```
PawelWMS, line 44: The PR check has been recently tweaked, so this should now work:
```suggestion
- Original version for Azure Linux.
```

SPECS/python-pip/python-pip.spec
PawelWMS: I'm also curious how this worked. The 2.0 version of Python seems not to provide any binary/script/file in general called `pip3` or `pip`. It's only available from the `python3-pip` packages, which we're removing right now. I think this may work in the toolchain build, if the raw toolchain installs `pip` but let's also make it work for a regular build. @ddstreetmicrosoft's suggestion to run the `pip` we just built sounds very reasonable to me.

PawelWMS (approved):
LGTM assuming the spec also builds during "regular" builds and in the buddy build pipeline.

dmcilvaney (approved):

anphel31:
> LGTM assuming the spec also builds during "regular" builds and in the buddy build pipeline.

It does not currently build in the buddybuild pipeline, just in the toolchain environment. I will merge now but come back to revisit this issue.

SPECS/python-pip/python-pip.spec
anphel31: I've fixed the building of this package in buddybuild/pkggen environment by adding ensurepip to install pip3, and adding a BR on python3-wheel. Note that this BR is temporarily disabled as it will cause a pkg build error until python3-wheel is moved into the toolchain. (toolchain packages cannot build require non-toolchain packages).

PR #8226, opened 2024-03-03
Upgrade python rpm generator scripts

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Azure Linux 3.0-dev rpms are not built with correct python provides. They are missing automatically provided python3dist(foo) metadata. The macros that python generators are currently out of date and this PR upgrades them to be compatible with python 3.12.

###### Change Log  

Toolchain changes:
- Move the following packages into the toolchain (they are already part of the "raw toolchain" build environment: `python-packaging`, `python-wheel`, `python-flit-core`, `python-rpm-generators`
- Add the following packages to default pkggen worker chroot: (to ensure rpms built in pkggen environment get automatic python provides: `python-rpm-generators` and `python-packaging`
- Install `rpm`, `python-setuptools`, `pyproject-rpm-macros`, `python-packaging`, and `python-rpm-generators` in toolchain build environment. This allows the python generators to work when building toolchain packages.
- Remove old and unneeded python generators from raw toolchain (ex: `python.attr`)
- Fix raw `rpm` to use "--with-vendor=azl" instead of "--with-vendor=mariner"
- Add `python-packaging` to the toolchain, since `pythonbundles.py` depends on it

Package changes:
- Promote the `python-rpm-generators` package from SPECS-EXTENDED to SPECS, and upgrade to version 14. Add this package to the toolchain and worker chroot, so it's available for all package builds.
- Remove the python generator scripts (ex: `python.attr`) from the rpm package. These were outdated and replaced by versions in `python-rpm-generators`.
- Upgrade `pyproject-rpm-macros` to version 1.12.0
- `azurelinux-rpm-macros`: Remove `pythondist.attr`, add `python.lua`
- Fix three packages that error out with the following: (`python-daemon`, `python-importlib-metadata`, `python-zipp`)
```
"Error: The version in the Python package metadata 0.0.0 normalizes to zero."
"It's likely a packaging error caused by missing version information"
"(e.g. when using a version control system snapshot as a source)."
"Try providing the version information manually when building the Python package,"
"for example by setting the SETUPTOOLS_SCM_PRETEND_VERSION environment variable if the package uses setuptools_scm."
"If you are confident that the version of the Python package is intentionally zero,"
"you may %define the _python_dist_allow_version_zero macro in the spec file to disable this check."

"error: Dependency tokens must begin with alpha-numeric, '_' or '/': *** PYTHON_PROVIDED_VERSION_NORMALIZES_TO_ZERO___SEE_STDERR ***"
"Provides: python-zipp = 3.17.0-1.azl3 python3-zipp = 3.17.0-1.azl3 python3.12-zipp = 3.17.0-1.azl3"
```

See: https://rpm-software-management.github.io/rpm/manual/dependency_generators.html 

**YES**

###### Test Methodology

- Full build: 3.0.20240307-anphel-522214
- X64 Toolchain - https://dev.azure.com/mariner-org/mariner/_build/results?buildId=522217&view=results
- X64 BuildRpms - https://dev.azure.com/mariner-org/mariner/_build/results?buildId=522295&view=results


SPECS/python-rpm-generators/python-rpm-generators.spec
```diff
@@ -1,23 +1,28 @@
...
+# Originally, this was simplified to GPL-2.0-or-later, but "effective license" analysis is no longer allowed
+License:        GPL-2.0-or-later AND LGPL-2.1-or-later AND (LicenseRef-Fedora-Public-Domain OR LGPL-2.1-or-later OR GPL-2.0-or-later)
 Url:            https://src.fedoraproject.org/python-rpm-generators
-# Commit is the last change in following files
+
+# Originally the following files were part of RPM, so the license is inherited: GPL-2.0-or-later
+# The COPYING file is grabbed from the last commit that changed the files
 Source0:        https://raw.githubusercontent.com/rpm-software-management/rpm/102eab50b3d0d6546dfe082eac0ade21e6b3dbf1/COPYING
 Source1:        python.attr
 Source2:        pythondist.attr
```
anphel31, line 17: https://src.fedoraproject.org/rpms/python-rpm-generators/blob/f40/f/pythondist.attr

SPECS/azurelinux-rpm-macros/azurelinux-rpm-macros.spec
```diff
@@ -43,6 +42,7 @@ Source24:       macros.suse
 Source25:       gen-ld-script.sh
 Source26:       generate-package-note.py
 Source27:       verify-package-notes.sh
+Source28:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/python.lua
```
PawelWMS, line 45: Fedora includes that file in its `python-srpm-macros` package (subpackage of `python-rpm-macros.spec`). I know we agreed to adding that package ourselves and moving out the files provided by it from this spec, but could we group all files in this spec, which would eventually come from `python-rpm-macros` (`macros.python`, `macros.pybytecompile`, `macros.python-srpm`, etc.) and add a comment saying where we can look for newer version, which would be compatible with each other?

corvus-callidus (approved):

SPECS/azurelinux-rpm-macros/macros.python-srpm
```diff
@@ -1,28 +1,95 @@
+# There are multiple Python 3 versions packaged, but only one can be the "main" version
```
PawelWMS, line 1: I think it might be safe to also update other files we have in this spec, which come from Fedora's `python-rpm-macros` spec. Otherwise, there's a greater risk of hitting some obscure issues stemming from the fact, that some of these macro files and script are from an older version of Fedora, than this one.

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
```diff
@@ -1,13 +1,23 @@
...
 Name:           pyproject-rpm-macros
+
+%if 0%{?with_check}
+%bcond tests 1
+%else
+%bcond tests 0
+%endif
+# pytest-xdist and tox are not desired in RHEL
+#%%bcond pytest_xdist %%{undefined rhel}
+#%%bcond tox_tests %%{undefined rhel}
```
PawelWMS, line 11: I think we might as well replace all `%if 0%{with test}` with `%if 0%{with_check}`, since we regulate test vs non-test builds with the `with_check` macro.

And I think we could also remove the RHEL-specific comment below and any macros we're not setting like `pytest_xdist` or `tox_tests`.

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
```diff
@@ -41,25 +53,42 @@ Source303:      test_RECORD
...
+BuildRequires:  python3dist(pyyaml)
 BuildRequires:  python3dist(packaging)
 BuildRequires:  python3dist(pip)
-BuildRequires:  python3dist(pyyaml)
 BuildRequires:  python3dist(setuptools)
+%if %{with tox_tests}
 BuildRequires:  python3dist(tox-current-env) >= 0.0.6
+%endif
 BuildRequires:  python3dist(wheel)
+#BuildRequires:  (python3dist(tomli) if python3 < 3.11)
```
PawelWMS, line 69: Can probably be removed. We don't have anything below Python 3.12 anyway.

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
anphel31: ok, I can do that, but wanted to minimize the changes from the fedora spec, where possible

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
```diff
@@ -75,48 +104,82 @@ They work for:
...
 %prep
 # Not strictly necessary but allows working on file names instead
 # of source numbers in install section
-%setup -q -c -T
+%setup -c -T
 cp -p %{sources} .
 
+%generate_buildrequires
+# nothing to do, this is here just to assert we have that functionality
+
```
PawelWMS, line 128: I think it would be better to remove these lines, since the comment is misleading. We (Azure Linux) actually do not support `generate_buildrequires` right now. As in: RPM will work with it correctly but it's ignored by out toolkit's package build.

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
```diff
@@ -41,25 +53,42 @@ Source303:      test_RECORD
...
+# We use the following tools outside of coreutils
+Requires:       findutils
+Requires:       sed
+
+# This package requires the %%generate_buildrequires functionality.
+# It has been introduced in RPM 4.15 (4.14.90 is the alpha of 4.15).
+# What we need is rpmlib(DynamicBuildRequires), but that is impossible to (Build)Require.
+Requires:       rpm-build >= 4.14.90
+BuildRequires:  rpm-build >= 4.14.90
```
PawelWMS, line 91: Not needed: we always had a newer version of RPM. Also, we ignore BR produced by `generate_buildrequires`, so this may be misleading.

SPECS/pyproject-rpm-macros/macros.aaa-pyproject-srpm
```diff
@@ -0,0 +1,7 @@
+# This file is called macros.aaa-pyproject-srpm
+# to sort alphabetically before macros.pyproject.
+# When this file is installed but macros.pyproject is not
+# this macro will cause the package with the real macro to be installed.
+# When macros.pyproject is installed, it overrides this macro.
+# Note: This needs to maintain the same set of options as the real macro.
+%pyproject_buildrequires(rRxtNwe:C:) echo 'pyproject-rpm-macros' && exit 0
```
PawelWMS, line 7: I don't think the idea behind this file is Fedora-specific. It might be better to just drop this file to avoid confusion. Especially since having this fille installed but not `macros.pyproject` will still not install `pyproject-rpm-macros`.

SPECS/python-daemon/python-daemon.spec
```diff
@@ -10,6 +10,8 @@ URL:            https://pypi.python.org/pypi/python-daemon/
 Source0:        https://files.pythonhosted.org/packages/source/p/python-daemon/%{name}-%{version}.tar.gz
 BuildArch:      noarch
 
+%define _python_dist_allow_version_zero 1
```
PawelWMS, line 13: For all such lines, could we add a comment explaining why it needs to be there?

SPECS/python-rpm-generators/python-rpm-generators.spec
```diff
@@ -26,11 +31,11 @@ BuildArch:      noarch
...
 %package -n python3-rpm-generators
 Summary:        %{summary}
-Requires:       python3-setuptools
-# The point of split
-Conflicts:      rpm-build < 4.13.0.1-2
-# Breaking change, change a way how depgen is enabled
-Conflicts:      python-rpm-macros < 3-35
+Requires:       python3-packaging
+# We have parametric macro generators, we need RPM 4.16 (4.15.90+ is 4.16 alpha)
+Requires:       rpm > 4.15.90-0
```
PawelWMS, line 36: Not needed.
```suggestion
```

SPECS/python-zipp/python-zipp.spec
```diff
@@ -3,19 +3,20 @@
 Summary:        Backport of pathlib-compatible object wrapper for zip files
 Name:           python-%{pypi_name}
 Version:        3.17.0
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        MIT
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 URL:            https://github.com/jaraco/zipp
-Source0:        %{pypi_source}
```
PawelWMS, line 11: The `pypi_source` macro should work correctly, if we've got everything right. Can you check, if we can revert the change to this one line?

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
anphel31: removed

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
anphel31: removed

SPECS/azurelinux-rpm-macros/macros.python-srpm
anphel31: What do you feel about making those additional changes in a follow-up PR? I tried to limit this PR to python-related stuff and it's already a pretty large PR

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
anphel31: fixed

SPECS/python-daemon/python-daemon.spec
anphel31: sure

SPECS/python-rpm-generators/python-rpm-generators.spec
anphel31: It still needs RPM, right? just not the `> 4.15.90-0`?

SPECS/python-zipp/python-zipp.spec
anphel31: where is this set?

anphel31: ok, I see now that many other specs are also using this, and it appears to work. I've reverted this change

SPECS/python-daemon/python-daemon.spec
anphel31: I found the root cause for the version error in this package and fixed it (had to remove a bogus sed command)

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
anphel31: fixed

SPECS/python-rpm-generators/python-rpm-generators.spec
PawelWMS: Sorry, right.:)

SPECS/pyproject-rpm-macros/macros.aaa-pyproject-srpm
anphel31: I've removed this

SPECS/azurelinux-rpm-macros/macros.python-srpm
anphel31: I've pushed updates to all of these files

SPECS/azurelinux-rpm-macros/azurelinux-rpm-macros.spec
anphel31: I updated the PR to group these files together, add source URLs, and update to the latest contents from Fedora 40

SPECS/azurelinux-rpm-macros/azurelinux-rpm-macros.spec
```diff
@@ -18,31 +18,37 @@ Source2:        default-hardened-cc1
...
+Source23:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/macros.python-srpm
+Source24:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/brp-python-bytecompile
+Source25:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/macros.pybytecompile
+Source26:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/compileall2.py
+Source27:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/python.lua
+Source28:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/clamp_source_mtime.py
+Source29:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/pathfix.py
+Source30:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/brp-fix-pyc-reproducibility
+Source31:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/brp-python-hardlink
+Source32:       https://src.fedoraproject.org/rpms/python-rpm-macros/blob/f40/f/import_all_modules.py
```
PawelWMS, line 50: From this set of files, I see we're installing `pathfix.py`, `import_all_modules.py`, `clamp_source_mtime.py`, and `python.lua`. I think we may be still missing `brp-python-hardlink` and `brp-fix-pyc-reproducibility`.

SPECS/azurelinux-rpm-macros/azurelinux-rpm-macros.spec
```diff
@@ -6,7 +6,7 @@
 %global __brp_python_bytecompile %{nil}
 Summary:        Azure Linux specific rpm macro files
 Name:           azurelinux-rpm-macros
-Version:        %{azl}.0
+Version:        3.0
 Release:        1%{?dist}
```
PawelWMS, line 10: From the size of this change, it might be worth to bump `Release` and add a note in the changelog.

SPECS/ca-certificates/ca-certificates.spec
```diff
@@ -44,7 +44,7 @@ Name:           ca-certificates
 
 # When updating, "Epoch, "Version", AND "Release" tags must be updated in the "prebuilt-ca-certificates*" packages as well.
 Epoch:          1
-Version:        %{azl}.0.0
+Version:        3.0.0
```
PawelWMS, line 47: I think these changes got add accidentally after a recent fix in `3.0-dev`. Maybe a merge/rebase of the latest `3.0-dev` will get rid of these.

SPECS/libsemanage/libsemanage.spec
```diff
@@ -102,15 +102,13 @@ ln -sf  %{_libdir}/libsemanage.so.2 %{buildroot}/%{_libdir}/libsemanage.so
 %files python3
 %{python3_sitelib}/*.so
 %{python3_sitelib}/semanage.py*
+%{python3_sitelib}/__pycache__/semanage.cpython*.pyc
```
PawelWMS, line 105: Now I'm getting worried. Is this a merge artifact or you're reverting some of the changes done during Python 3.12 upgrade, because we now generate these files again?

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
```diff
@@ -87,36 +107,45 @@ cp -p %{sources} .
...
+install -pm 644 pyproject_convert.py %{buildroot}%{_rpmconfigdir}/azl/
+install -pm 644 pyproject_save_files.py  %{buildroot}%{_rpmconfigdir}/azl/
+install -pm 644 pyproject_preprocess_record.py %{buildroot}%{_rpmconfigdir}/azl/
+install -pm 644 pyproject_construct_toxenv.py %{buildroot}%{_rpmconfigdir}/azl/
+install -pm 644 pyproject_requirements_txt.py %{buildroot}%{_rpmconfigdir}/azl/
+install -pm 644 pyproject_wheel.py %{buildroot}%{_rpmconfigdir}/azl/
 
 %check
-pip3 install more_itertools pytest>=3.9 toml 'tox>=3.27.1,<4.0.0'
+%if 0%{?with_check}
```
PawelWMS, line 120: I think this is not needed - `with_check` will always be set, if we're already in the `%check` stage.

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
```diff
@@ -87,36 +107,45 @@ cp -p %{sources} .
...
 %{_rpmconfigdir}/azl/pyproject_preprocess_record.py
 %{_rpmconfigdir}/azl/pyproject_construct_toxenv.py
 %{_rpmconfigdir}/azl/pyproject_requirements_txt.py
+%{_rpmconfigdir}/azl/pyproject_wheel.py
+
+%doc README.md
+%license LICENSE
+
+%files -n pyproject-srpm-macros
+%license LICENSE
```
PawelWMS, line 143: Do we even need this as a separate package? Wouldn't a `Provides:` do the trick?

toolkit/scripts/toolchain/cgmanifest.json
```diff
@@ -5,11 +5,11 @@
                 "Type": "DockerImage",
                 "DockerImage": {
                     "Name": "mcr.microsoft.com/cbl-mariner/base/core",
-                    "Digest": "sha256:ca9eb22782f247189c15a4145dfde7f21bca25e545a6a3782a3fe2dbe3b128f0",
-                    "Tag": "2.0.20231004"
+                    "Digest": "sha256:82314abb594a695fd8817774e8b7f101934902cc1d99b3075e80acbc8b9b23ee",
```
PawelWMS, line 8: This one also looks like it's from a different PR.

SPECS/azurelinux-rpm-macros/azurelinux-rpm-macros.spec
anphel31: Good idea. There actually was a bump to version 2.0-27 from Daniel, who also helped with some of these changes. But now that we're on version 3.0, I'll bump it again.

SPECS/azurelinux-rpm-macros/azurelinux-rpm-macros.spec
anphel31: There is a catch-all below that installs `brp-*` that should handle these new files.

SPECS/firewalld/firewalld.spec
```diff
@@ -3,7 +3,7 @@
 Summary:        A firewall daemon with D-Bus interface providing a dynamic firewall
 Name:           firewalld
 Version:        2.0.2
-Release:        3%{?dist}
+Release:        2%{?dist}
```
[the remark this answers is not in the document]
anphel31: When I upgraded python to 3.12, it broke a handful of packages due to missing __pycache__ files. I updated those packages in the same PR - see https://github.com/microsoft/azurelinux/pull/7858/files#diff-4b714dcc3b08c555d4b13cea1e3c8790f4253060696205fda66e3bd3d83df107. But since this PR fixes the pycache issues, I decided to revert all of the temporary fixes

SPECS/libsemanage/libsemanage.spec
anphel31: Yes, this change was intentional - to remove the temporary fixes I put in place here: https://github.com/microsoft/azurelinux/pull/7858/files#diff-414a05781908e0638662dce8f649e00533b15999686b03c7884b8a047ad259fc
I decided not to bump the release number since this package has subsequently been updated already. I think this should be fine given we are still in the development phase for 3.0

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
anphel31: thanks, will remove

SPECS/pyproject-rpm-macros/pyproject-rpm-macros.spec
anphel31: We probably don't need it, but in the interest of minimizing deltas between our specs and Fedora, (and incase something we need is added to this subpackage in the future), I plan to leave it as-is for now.

toolkit/scripts/toolchain/build_official_toolchain_rpms.sh
```diff
@@ -541,6 +541,17 @@ chroot_and_install_rpms docbook-style-xsl
...
 
+build_rpm_in_chroot_no_install python-setuptools
+chroot_and_install_rpms python-setuptools python3-setuptools
+
+build_rpm_in_chroot_no_install pyproject-rpm-macros
+chroot_and_install_rpms pyproject-rpm-macros pyproject-rpm-macros
+chroot_and_install_rpms pyproject-rpm-macros pyproject-srpm-macros
+
+# python-packaging requires pyproject-rpm-macros
+build_rpm_in_chroot_no_install python-packaging
```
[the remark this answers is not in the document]
anphel31: yes, will fix this

toolkit/scripts/toolchain/cgmanifest.json
anphel31: yea, I wanted to unblock this branch so i could run a test build

PawelWMS (approved):

PR #8243, opened 2024-03-05
Image Customizer: Adding skippable frames with unique metadata to generated partitions

Adub17030MS:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When MIC customizes Mariner images, it produces multiple output image files (with the `.zst` extension) that correspond to different partitions on the system. Currently, combining partition images from multiple runs of MIC is not supported and has resulted in customer bugs. Today, we have no way to detect whether a user has accidentally mixed images from different runs.

Zstd has a concept of “skippable frames",  which are blocks of data that can be included in a compressed stream but do not impact the result of decompression (https://github.com/facebook/zstd/blob/release/doc/zstd_compression_format.md#skippable-frames). Given a zst compressed file, a skippable frame can be prepended to the start to produce an equivalent zst file carrying additional metadata. 

The proposal here is to have MIC write a skippable frame at the start of all partition images it produces. All partition images from the same run would carry the same skippable frame. 

###### Change Log  

- Added functionality in MIC to write a skippable frame at the start of all the partition images it produces. 
- All partition images from the same run would carry the same skippable frame metadata and this is a unique identifier for that run.
- Added unit test with verification functions that validated that the outputted `.zst` files had the valid skippable frame with the correct metadata and that they could be decompressed correctly with the skippable frame added in

**NO**

###### Test Methodology

- Manually tested this functionality by locally running MIC tool with the `-output-split-partitions-format` flag set to  `raw-zst`
- Added unit test for this feature
image

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
@@ -68,9 +68,14 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
 // Extract raw-zst partition
 func extractRawZstPartition(partitionRawFilepath string, skippableFrameMetadata [SkippableFramePayloadSize]byte, partitionFilename string, outDir string) (partitionFilepath string, err error) {
 	// Compress raw partition with zstd
-	tempPartitionFilepath, err := compressWithZstd(partitionRawFilepath)
+	tempPartitionFilepath, err := compressWithZstd(partitionRawFilepath, partitionFilename, outDir)
 	if err != nil {
 		return "", err
+	}
+	// Remove raw file since output partition format is raw-zst
+	err = os.Remove(partitionRawFilepath)
+	if err != nil {
+		return "", fmt.Errorf("failed to remove raw file %s:\n%w", partitionRawFilepath, err)
 	}
 	// Create a skippable frame containing the metadata and prepend the frame to the partition file
 	partitionFilepath, err = addSkippableFrame(tempPartitionFilepath, skippableFrameMetadata, partitionFilename, outDir)
@@ -81,11 +86,6 @@ func extractRawZstPartition(partitionRawFilepath string, skippableFrameMetadata 
 	err = os.Remove(tempPartitionFilepath)
 	if err != nil {
 		return "", fmt.Errorf("failed to remove temp file %s:\n%w", partitionRawFilepath, err)
-	}
-	// Remove raw file since output partition format is raw-zst
-	err = os.Remove(partitionRawFilepath)
-	if err != nil {
-		return "", fmt.Errorf("failed to remove raw file %s:\n%w", partitionRawFilepath, err)
 	}
 	return partitionFilepath, nil
 }
@@ -114,14 +114,16 @@ func copyBlockDeviceToFile(outDir, devicePath, name string) (filename string, er
 }
 
 // Compress file from .raw to .raw.zst format using zstd.
-func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, err error) {
+func compressWithZstd(partitionRawFilepath string, partitionFilename string, outDir string) (partitionFilepath string, err error) {
+	// Define output file.
+	outputPartitionFilepath := outDir + "/" + partitionFilename + "_temp.raw.zst"
 	// Using -f to overwrite a file with same name if it exists.
-	err = shell.ExecuteLive(true, "zstd", "-f", "-9", "-T0", partitionRawFilepath)
+	err = shell.ExecuteLive(true, "zstd", "-f", "-9", "-T0", partitionRawFilepath, "-o", outputPartitionFilepath)
 	if err != nil {
 		return "", fmt.Errorf("failed to compress %s with zstd:\n%w", partitionRawFilepath, err)
 	}
 
-	return partitionRawFilepath + ".zst", nil
+	return outputPartitionFilepath, nil
 }
 
 // Prepend a skippable frame with the metadata to the specified partition file.
@@ -134,7 +136,7 @@ func addSkippableFrame(tempPartitionFilepath string, skippableFrameMetadata [Ski
 	// Create a skippable frame
 	skippableFrame := createSkippableFrame(SkippableFrameMagicNumber, SkippableFramePayloadSize, skippableFrameMetadata)
 	// Define the final partition file path
-	partitionFilepath = outDir + "/" + partitionFilename + "_final" + ".raw.zst"
+	partitionFilepath = outDir + "/" + partitionFilename + ".raw.zst"
 	// Create partition file
 	finalFile, err := os.Create(partitionFilepath)
 	if err != nil {
--- a/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
@@ -3,24 +3,21 @@ package imagecustomizerlib
 import (
 	"bytes"
 	"crypto/rand"
-	"crypto/sha256"
 	"encoding/binary"
-	"encoding/hex"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"strconv"
 
 	"github.com/microsoft/azurelinux/toolkit/tools/imagegen/diskutils"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/file"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
 )
 
-// Define global constants
 const (
-	MagicNumber uint32 = 0x184D2A50
-	FrameSize   uint32 = 16
+	SkippableFrameMagicNumber uint32 = 0x184D2A50
+	SkippableFrameSize        uint32 = 16
 )
 
 // Extract all partitions of connected image into separate files with specified format.
@@ -33,20 +30,17 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
 	}
 
 	// Create skippable frame metadata defined as a random 128-Bit number
-	metadata, err := createSkippableFrameMetadata()
+	skippableFrameMetadata, err := createSkippableFrameMetadata()
 	if err != nil {
 		return err
 	}
-
-	// Encode metadata into a payload defined as a uint32 array
-	payload := Encode128BitLittleEndian(metadata)
 
 	for partitionNum := 0; partitionNum < len(diskPartitions); partitionNum++ {
 		if diskPartitions[partitionNum].Type == "part" {
 			rawFilename := basename + "_" + strconv.Itoa(partitionNum) + ".raw"
 			partitionLoopDevice := diskPartitions[partitionNum].Path
 
-			partitionFilePath, err := copyBlockDeviceToFile(outDir, partitionLoopDevice, rawFilename)
+			partitionFilepath, err := copyBlockDeviceToFile(outDir, partitionLoopDevice, rawFilename)
 			if err != nil {
 				return err
 			}
@@ -55,28 +49,28 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
 			case "raw":
 				// Do nothing for "raw" case.
 			case "raw-zst":
-				partitionRawFilePath := partitionFilePath
-				partitionFilepath, err := compressWithZstd(partitionRawFilePath)
+				partitionRawFilepath := partitionFilepath
+				partitionFilepath, err := compressWithZstd(partitionRawFilepath)
 				if err != nil {
 					return err
 				}
-				// Create a skippable frame containing the metadata payload and prepend the frame to the partition file
-				err = addSkippableFrame(partitionFilepath, MagicNumber, FrameSize, payload)
+				// Create a skippable frame containing the metadata and prepend the frame to the partition file
+				err = addSkippableFrame(partitionFilepath, SkippableFrameMagicNumber, SkippableFrameSize, skippableFrameMetadata)
 				if err != nil {
 					return err
 				}
 				// Verify decompression with skippable frame
-				err = verifySkippableFrameDecompression(partitionRawFilePath, partitionFilepath)
+				err = verifySkippableFrameDecompression(partitionRawFilepath, partitionFilepath)
 				if err != nil {
 					return err
 				}
 				// Remove raw file since output partition format is raw-zst.
-				err = os.Remove(partitionRawFilePath)
+				err = os.Remove(partitionRawFilepath)
 				if err != nil {
-					return fmt.Errorf("failed to remove raw file %s:\n%w", partitionRawFilePath, err)
+					return fmt.Errorf("failed to remove raw file %s:\n%w", partitionRawFilepath, err)
 				}
 				// Verify skippable frame metadata
-				err = verifySkippableFrameMetadataFromFile(partitionFilepath, MagicNumber, FrameSize, metadata)
+				err = verifySkippableFrameMetadataFromFile(partitionFilepath, SkippableFrameMagicNumber, SkippableFrameSize, skippableFrameMetadata)
 				if err != nil {
 					return err
 				}
@@ -84,7 +78,7 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
 			default:
 				return fmt.Errorf("unsupported partition format (supported: raw, raw-zst): %s", partitionFormat)
 			}
-			logger.Log.Infof("Partition file created: %s", partitionFilePath)
+			logger.Log.Infof("Partition file created: %s", partitionFilepath)
 		}
 	}
 	return nil
@@ -124,8 +118,8 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
 	return partitionRawFilepath + ".zst", nil
 }
 
-// Prepend a skippable frame with the metadata payload to the specified partition file.
-func addSkippableFrame(partitionFilepath string, magicNumber uint32, frameSize uint32, payload [4]uint32) (err error) {
+// Prepend a skippable frame with the metadata to the specified partition file.
+func addSkippableFrame(partitionFilepath string, magicNumber uint32, frameSize uint32, skippableFrameMetadata [SkippableFrameSize]byte) (err error) {
 	// Read existing data from the partition file.
 	existingData, err := os.ReadFile(partitionFilepath)
 	if err != nil {
@@ -133,7 +127,7 @@ func addSkippableFrame(partitionFilepath string, magicNumber uint32, frameSize u
 	}
 
 	// Create a skippable frame.
-	skippableFrame := createSkippableFrame(magicNumber, frameSize, payload)
+	skippableFrame := createSkippableFrame(magicNumber, frameSize, skippableFrameMetadata)
 
 	// Combine the skippable frame and existing data.
 	newData := append(skippableFrame, existingData...)
@@ -148,9 +142,9 @@ func addSkippableFrame(partitionFilepath string, magicNumber uint32, frameSize u
 }
 
 // Creates a skippable frame.
-func createSkippableFrame(magicNumber uint32, frameSize uint32, payload [4]uint32) (skippableFrame []byte) {
+func createSkippableFrame(magicNumber uint32, frameSize uint32, skippableFrameMetadata [SkippableFrameSize]byte) (skippableFrame []byte) {
 	// Calculate the length of the byte array
-	lengthOfByteArray := 4 + 4 + (4 * len(payload))
+	lengthOfByteArray := 4 + 4 + len(skippableFrameMetadata)
 	// Define the Skippable frame
 	skippableFrame = make([]byte, lengthOfByteArray)
 	// Magic_Number
@@ -158,10 +152,7 @@ func createSkippableFrame(magicNumber uint32, frameSize uint32, payload [4]uint3
 	// Frame_Size
 	binary.LittleEndian.PutUint32(skippableFrame[4:8], frameSize)
 	// User_Data
-	binary.LittleEndian.PutUint32(skippableFrame[8:12], payload[0])
-	binary.LittleEndian.PutUint32(skippableFrame[12:16], payload[1])
-	binary.LittleEndian.PutUint32(skippableFrame[16:20], payload[2])
-	binary.LittleEndian.PutUint32(skippableFrame[20:24], payload[3])
+	copy(skippableFrame[8:24], skippableFrameMetadata[:])
 
 	logger.Log.Infof("Skippable frame has been created with the following metadata: %d", skippableFrame[8:24])
 
@@ -169,18 +160,18 @@ func createSkippableFrame(magicNumber uint32, frameSize uint32, payload [4]uint3
 }
 
 // Create user metadata that will be inserted into the skippable frame.
-func createSkippableFrameMetadata() (metadata [16]byte, err error) {
-	// Set the metadata to be a random 128-Bit number
-	metadata, err = generateRandom128BitNumber()
+func createSkippableFrameMetadata() (skippableFrameMetadata [SkippableFrameSize]byte, err error) {
+	// Set the skippableFrameMetadata to be a random 128-Bit number
+	skippableFrameMetadata, err = generateRandom128BitNumber()
 	if err != nil {
-		return metadata, err
+		return skippableFrameMetadata, err
 	}
-	return metadata, nil
+	return skippableFrameMetadata, nil
 }
 
 // Generates a Random 128-Bit number.
-func generateRandom128BitNumber() ([16]byte, error) {
-	var randomBytes [16]byte
+func generateRandom128BitNumber() ([SkippableFrameSize]byte, error) {
+	var randomBytes [SkippableFrameSize]byte
 	_, err := rand.Read(randomBytes[:])
 	if err != nil {
 		return randomBytes, err
@@ -188,50 +179,21 @@ func generateRandom128BitNumber() ([16]byte, error) {
 	return randomBytes, nil
 }
 
-// Encodes a 128-Bit number into an array of uint32 in little-endian order.
-func Encode128BitLittleEndian(number [16]byte) [4]uint32 {
-	var encoded [4]uint32
-	for i := 0; i < 4; i++ {
-		offset := i * 4
-		encoded[i] = binary.LittleEndian.Uint32(number[offset : offset+4])
-	}
-	return encoded
-}
-
-// Calculates sha256sum hash for a given file
-func calculateSHA256(filePath string) (string, error) {
-	file, err := os.Open(filePath)
-	if err != nil {
-		return "", err
-	}
-	defer file.Close()
-
-	hash := sha256.New()
-	if _, err := io.Copy(hash, file); err != nil {
-		return "", err
-	}
-
-	hashInBytes := hash.Sum(nil)
-	hashString := hex.EncodeToString(hashInBytes)
-
-	return hashString, nil
-}
-
 // Decompress the .raw.zst partition file and verifies the hash matches with the source .raw file
 func verifySkippableFrameDecompression(rawPartitionFilepath string, rawZstPartitionFilepath string) (err error) {
 	// Decompressing .raw.zst file
-	decompressedPartitionFilePath := "build/decompressed.raw"
-	err = shell.ExecuteLive(true, "zstd", "-d", rawZstPartitionFilepath, "-o", decompressedPartitionFilePath)
+	decompressedPartitionFilepath := "build/decompressed.raw"
+	err = shell.ExecuteLive(true, "zstd", "-d", rawZstPartitionFilepath, "-o", decompressedPartitionFilepath)
 	if err != nil {
 		return fmt.Errorf("failed to decompress %s with zstd:\n%w", rawZstPartitionFilepath, err)
 	}
 
 	// Calculating hashes
-	rawPartitionFileHash, err := calculateSHA256(rawPartitionFilepath)
+	rawPartitionFileHash, err := file.GenerateSHA256(rawPartitionFilepath)
 	if err != nil {
 		return fmt.Errorf("error: %w", err)
 	}
-	decompressedPartitionFileHash, err := calculateSHA256(decompressedPartitionFilePath)
+	decompressedPartitionFileHash, err := file.GenerateSHA256(decompressedPartitionFilepath)
 	if err != nil {
 		return fmt.Errorf("error: %w", err)
 	}
@@ -240,19 +202,19 @@ func verifySkippableFrameDecompression(rawPartitionFilepath string, rawZstPartit
 	if rawPartitionFileHash != decompressedPartitionFileHash {
 		return fmt.Errorf("decompressed partition file hash does not match source partition file hash: %s != %s", decompressedPartitionFileHash, rawPartitionFilepath)
 	}
-	logger.Log.Infof("Decompressed partition file hash matches source partition file hash!")
+	logger.Log.Debugf("Decompressed partition file hash matches source partition file hash!")
 
 	// Removing decompressed file
-	err = os.Remove(decompressedPartitionFilePath)
+	err = os.Remove(decompressedPartitionFilepath)
 	if err != nil {
-		return fmt.Errorf("failed to remove raw file %s:\n%w", decompressedPartitionFilePath, err)
+		return fmt.Errorf("failed to remove raw file %s:\n%w", decompressedPartitionFilepath, err)
 	}
 
 	return nil
 }
 
 // Verifies that the skippable frame has been correctly prepended to the partition file with the correct data
-func verifySkippableFrameMetadataFromFile(partitionFilepath string, magicNumber uint32, frameSize uint32, metadata [16]byte) (err error) {
+func verifySkippableFrameMetadataFromFile(partitionFilepath string, magicNumber uint32, frameSize uint32, skippableFrameMetadata [SkippableFrameSize]byte) (err error) {
 	// Read existing data from the partition file.
 	existingData, err := os.ReadFile(partitionFilepath)
 	if err != nil {
@@ -275,9 +237,9 @@ func verifySkippableFrameMetadataFromFile(partitionFilepath string, magicNumber 
 	}
 	logger.Log.Infof("Skippable frame frameSize field is correct...")
 
-	// verify that the skippable frame has the correct inserted metadata by validating metadata
-	if !bytes.Equal(existingData[8:8+frameSize], metadata[:]) {
-		return fmt.Errorf("skippable frame metadata does not match the inserted metadata:\n %d != %d", existingData[8:8+frameSize], metadata[:])
+	// verify that the skippable frame has the correct inserted metadata by validating skippableFrameMetadata
+	if !bytes.Equal(existingData[8:8+frameSize], skippableFrameMetadata[:]) {
+		return fmt.Errorf("skippable frame metadata does not match the inserted metadata:\n %d != %d", existingData[8:8+frameSize], skippableFrameMetadata[:])
 	}
 	logger.Log.Infof("Skippable frame is valid and contains the correct metadata!")
 
--- a/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
@@ -195,6 +195,7 @@ func Encode128BitLittleEndian(number [16]byte) [4]uint32 {
 	return encoded
 }
 
+// Calculates sha256sum hash for a given file
 func calculateSHA256(filePath string) (string, error) {
 	file, err := os.Open(filePath)
 	if err != nil {
--- a/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
@@ -5,7 +5,6 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
-	"io/fs"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -16,10 +15,9 @@ import (
 )
 
 const (
-	SkippableFrameMagicNumber uint32      = 0x184D2A50
-	SkippableFramePayloadSize uint32      = 16
-	SkippableFrameHeaderSize  int         = 8
-	PartitionFilePermissions  fs.FileMode = 0644
+	SkippableFrameMagicNumber uint32 = 0x184D2A50
+	SkippableFramePayloadSize uint32 = 16
+	SkippableFrameHeaderSize  int    = 8
 )
 
 // Extract all partitions of connected image into separate files with specified format.
@@ -70,7 +68,7 @@ func extractRawZstPartition(partitionRawFilepath string, skippableFrameMetadata 
 	// Define file path for temporary partition
 	tempPartitionFilepath := outDir + "/" + partitionFilename + "_temp.raw.zst"
 	// Compress raw partition with zstd
-	tempPartitionFilepath, err = compressWithZstd(partitionRawFilepath, tempPartitionFilepath)
+	err = compressWithZstd(partitionRawFilepath, tempPartitionFilepath)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -20,29 +26,62 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
 		return err
 	}
 
+	// Define magicNumber and frameSize for skippable frames
```
vinceaperri, line 29: How did you determine these values? Should they be global constants?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -20,29 +26,62 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
+	magicNumber := uint32(0x184D2A50)
+	frameSize := uint32(16)
```
vinceaperri, line 31: 16 seems to be hardcoded throughout the code added. Is there a way to eliminate this redundancy and define and reuse a single global constant?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +118,165 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
...
+	hashString := hex.EncodeToString(hashInBytes)
+
+	return hashString, nil
+}
+
+// Decompress the .raw.zst partition file and verifies the hash matches with the source .raw file
+func verifySkippableFrameDecompression(rawPartitionFilepath string, rawZstPartitionFilepath string) (err error) {
+	// Decompressing .raw.zst file
+	decompressedPartitionFilePath := "build/decompressed.raw"
+	err = shell.ExecuteLive(true, "zstd", "-d", rawZstPartitionFilepath, "-o", decompressedPartitionFilePath)
```
cwize1, line 224: The files will potentially be pretty large. So, fully decompressing them will take a non-trivial amount of time. So, we probably shouldn't be doing this in the product code.

You should move this verification logic into UTs.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -20,29 +26,62 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
 			if err != nil {
 				return err
 			}
 
 			switch partitionFormat {
 			case "raw":
 				// Do nothing for "raw" case.
+				logger.Log.Infof("Partition file created: %s", partitionRawFilepath)
 			case "raw-zst":
-				partitionFilepath, err = compressWithZstd(partitionFilepath)
```
vinceaperri, line 37: To avoid renaming on line 47 and moving/copying logger.Log.Infof lines, have you tried:

```
case "raw-zst":
  partitionRawFilePath := partitionFilePath
  partitionFilepath, err = compressWithZstd(partitionFilepath)
  ...
```

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -20,29 +26,62 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
 		return err
 	}
 
+	// Define magicNumber and frameSize for skippable frames
+	magicNumber := uint32(0x184D2A50)
```
cwize1: Put these two numbers in `const` at the top of the file.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +118,165 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
 		return "", fmt.Errorf("failed to compress %s with zstd:\n%w", partitionRawFilepath, err)
 	}
 
-	// Remove raw file since output partition format is raw-zst.
-	err = os.Remove(partitionRawFilepath)
+	return partitionRawFilepath + ".zst", nil
+}
+
+// Prepend a skippable frame with the metadata payload to the specified partition file.
+func addSkippableFrame(partitionFilepath string, magicNumber uint32, frameSize uint32, payload [4]uint32) (err error) {
```
vinceaperri, line 125: This payload size of 4 is also a good candidate for a global constant. It is reused many times in the newly added code. Is it somehow determined by the framesize?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +118,165 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
...
+	}
+
+	// Create a skippable frame.
+	skippableFrame := createSkippableFrame(magicNumber, frameSize, payload)
+
+	// Combine the skippable frame and existing data.
+	newData := append(skippableFrame, existingData...)
+
+	// Write the combined data back to the partition file.
+	err = os.WriteFile(partitionFilepath, newData, 0644)
```
vinceaperri, line 139: Where does the mode of 0644 come from? Is this how the file was originally created? If so, can you reuse a constant value here?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -20,29 +26,62 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
+				partitionFilepath, err := compressWithZstd(partitionRawFilepath)
+				if err != nil {
+					return err
+				}
+				// Create a skippable frame containing the metadata payload and prepend the frame to the partition file
+				err = addSkippableFrame(partitionFilepath, magicNumber, frameSize, payload)
+				if err != nil {
+					return err
+				}
+				// Verify decompression with skippable frame
```
vinceaperri, line 66: verifySkippableFrameDecompression and verifySkippableFrameMetadataFromFile seem to be better suited as unit tests. As sanity-tests, they seem a bit expensive to call in this function.

vinceaperri (dismissed):

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +118,165 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
...
+	// Calculate the length of the byte array
+	lengthOfByteArray := 4 + 4 + (4 * len(payload))
+	// Define the Skippable frame
+	skippableFrame = make([]byte, lengthOfByteArray)
+	// Magic_Number
+	binary.LittleEndian.PutUint32(skippableFrame, magicNumber)
+	// Frame_Size
+	binary.LittleEndian.PutUint32(skippableFrame[4:8], frameSize)
+	// User_Data
+	binary.LittleEndian.PutUint32(skippableFrame[8:12], payload[0])
```
cwize1, line 161: Converting the 128-bits into u32s and then back into bytes seems a little pointless. You should be able to use the copy function.

```golang
var payload [16]byte
...
copy(skippableFrame[8:24], payload[:])
```

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +118,165 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
...
+	binary.LittleEndian.PutUint32(skippableFrame[12:16], payload[1])
+	binary.LittleEndian.PutUint32(skippableFrame[16:20], payload[2])
+	binary.LittleEndian.PutUint32(skippableFrame[20:24], payload[3])
+
+	logger.Log.Infof("Skippable frame has been created with the following metadata: %d", skippableFrame[8:24])
```
cwize1, line 166: Move to log message to debug level. The info level is reserved for stuff of interest to the end-user.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: These values are defined here: https://github.com/facebook/zstd/blob/release/doc/zstd_compression_format.md#skippable-frames

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +118,165 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
...
+	var encoded [4]uint32
+	for i := 0; i < 4; i++ {
+		offset := i * 4
+		encoded[i] = binary.LittleEndian.Uint32(number[offset : offset+4])
+	}
+	return encoded
+}
+
+// Calculates sha256sum hash for a given file
+func calculateSHA256(filePath string) (string, error) {
```
cwize1, line 202: There is already the `file.GenerateSHA256` function.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
cwize1, line 128: These functions are prime candidates for UTs. Please add some.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Yes, they should be global constants. I will change these values to be that👍

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Good point, I will add this change in

Adub17030MS: Added this in, resolving comment.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Added this in, resolving the conversation.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +118,165 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
...
 
-	// Remove raw file since output partition format is raw-zst.
-	err = os.Remove(partitionRawFilepath)
+	return partitionRawFilepath + ".zst", nil
+}
+
+// Prepend a skippable frame with the metadata payload to the specified partition file.
+func addSkippableFrame(partitionFilepath string, magicNumber uint32, frameSize uint32, payload [4]uint32) (err error) {
+	// Read existing data from the partition file.
+	existingData, err := os.ReadFile(partitionFilepath)
```
cwize1, line 130: These are potentially fairly large files. So, reading them into memory could potentially take a non-trivial amount of time and will consume a non-trivial amount of memory.

The fancy way of fixing this is:

1. Open file in Go.
2. Add the skippable frame.
3. Call `zstd` without the `-o` so that it outputs to stdout. Do this by calling [exec.Command](https://pkg.go.dev/os/exec#Command) directly (don't use the functions in the 'shell' package). and set the open file (from step 1) on the [Stdout](https://pkg.go.dev/os/exec#Cmd) field.

This will result in the file handle being passed to the zstd process, so that zstd writes directly to the file.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: I added it as a global constant and will replace all locations where it is hardcoded, but should I replace locations where it is a part of a function definition parameter (i.e. the byte array parameters of functions like `createSkippableFrameMetadata` and `createSkippableFrameMetadata`?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -34,15 +55,36 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
 			case "raw":
 				// Do nothing for "raw" case.
 			case "raw-zst":
-				partitionFilepath, err = compressWithZstd(partitionFilepath)
+				partitionRawFilePath := partitionFilePath
```
cwize1, line 58: This is a little too much code for a `case` statement. I recommend putting the code in a helper function.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
vinceaperri: Can you make them variable length?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -20,12 +32,21 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
+	// Encode metadata into a payload defined as a uint32 array
+	payload := Encode128BitLittleEndian(metadata)
+
 	for partitionNum := 0; partitionNum < len(diskPartitions); partitionNum++ {
 		if diskPartitions[partitionNum].Type == "part" {
 			rawFilename := basename + "_" + strconv.Itoa(partitionNum) + ".raw"
 			partitionLoopDevice := diskPartitions[partitionNum].Path
 
-			partitionFilepath, err := copyBlockDeviceToFile(outDir, partitionLoopDevice, rawFilename)
+			partitionFilePath, err := copyBlockDeviceToFile(outDir, partitionLoopDevice, rawFilename)
```
amritakohli, line 49: nit: Filepath* for consistency

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
amritakohli, line 58: nit: Filepath* for consistency (in both variables)

amritakohli (changes requested):

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
amritakohli: please keep this consistent everywhere in the file

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Addressed this comment.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Addressed this comment.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -11,6 +17,12 @@ import (
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
 )
 
+// Define global constants
+const (
+	MagicNumber uint32 = 0x184D2A50
```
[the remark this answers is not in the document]
Adub17030MS: Updated the names.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Updated to use this instead.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Updated.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Just updated it to use this logic.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -20,12 +32,21 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
 	}
 
+	// Create skippable frame metadata defined as a random 128-Bit number
+	metadata, err := createSkippableFrameMetadata()
+	if err != nil {
+		return err
+	}
+
+	// Encode metadata into a payload defined as a uint32 array
+	payload := Encode128BitLittleEndian(metadata)
```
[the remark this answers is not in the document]
Adub17030MS: Updated the code to directly insert the `skippableFrameMetadata` byte array into the `skippableFrame` with `copy()` as per your suggestion earlier to avoid converting to uint32 and back to byte array, and this `payload` variable no longer exists now.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Updated the code and it no longer makes use of this `payload` variable and also all the sizes are now defined based on the constants so none are hardcoded anymore👍.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Updated the code to use the global constants and not hardcoded values

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Added a helper function that contains this code, resolving this comment.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
amritakohli: can you add a comment for this in the file?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Moved them into unit test file.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Moved them into unit test file

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Moved them into unit test file

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -34,20 +47,39 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
 				return fmt.Errorf("unsupported partition format (supported: raw, raw-zst): %s", partitionFormat)
 			}
-
 			logger.Log.Infof("Partition file created: %s", partitionFilepath)
 		}
 	}
 	return nil
 }
 
+// Extract raw-zstd partition
```
amritakohli, line 63: nit: raw-zst*

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: I am a little confused here:
for step 1 I am opening the file in go with `os.OpenFile()`...
for step 2, how would I "add" the skippable frame to the open file? 
for step 3, I am writing to the open file with zstd?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Updated.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -34,20 +47,39 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
-
 			logger.Log.Infof("Partition file created: %s", partitionFilepath)
 		}
 	}
 	return nil
 }
 
+// Extract raw-zst partition
+func extractRawZstPartition(partitionRawFilepath string, skippableFrameMetadata [SkippableFrameSize]byte) (partitionFilepath string, err error) {
+	// Compress raw partition with zstd and output it
```
amritakohli, line 65: nit: this comment could simply be "Compress raw partition file with zstd"

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -34,20 +47,39 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
+	partitionFilepath, err = compressWithZstd(partitionRawFilepath)
+	if err != nil {
+		return "", err
+	}
+	// Create a skippable frame containing the metadata and prepend the frame to the partition file
+	err = addSkippableFrame(partitionFilepath, SkippableFrameMagicNumber, SkippableFrameSize, skippableFrameMetadata)
+	if err != nil {
+		return "", err
+	}
+	// Remove raw file since output partition format is raw-zst.
```
amritakohli, line 75: nit: remove period for consistency

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+func createTestRawPartitionFile(filename string) (string, error) {
+	// Test data
+	data := []byte{0x01, 0x02, 0x03, 0x04, 0x05}
+
+	err := writeToFile(filename, data)
+	if err != nil {
+		return "", err
+	} else {
+		logger.Log.Infof("File %s created successfully with dummy data.", filename)
```
amritakohli, line 57: nit: colon, not period at the end with \n ?

amritakohli, line 57: nit: follow this format -
logger.Log.Infof("Partition file created: %s", partitionFilepath)

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,142 @@
...
+		return filename, nil
+	}
+}
+
+func writeToFile(fileName string, data []byte) error {
```
amritakohli, line 62: nit: call this variable filename for consistency

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,142 @@
...
+	file, err := os.Create(fileName)
```
amritakohli, line 63: file.Write already does os.Create so this part is not needed

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	_, err = file.Write(data)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Decompress the .raw.zst partition file and verifies the hash matches with the source .raw file
```
amritakohli, line 77: nit: verify*

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,142 @@
...
+	decompressedPartitionFilepath := "decompressed.raw"
+	err = shell.ExecuteLive(true, "zstd", "-d", rawZstPartitionFilepath, "-o", decompressedPartitionFilepath)
+	if err != nil {
+		return fmt.Errorf("failed to decompress %s with zstd:\n%w", rawZstPartitionFilepath, err)
+	}
+
+	// Calculating hashes
+	rawPartitionFileHash, err := file.GenerateSHA256(rawPartitionFilepath)
+	if err != nil {
+		return fmt.Errorf("error: %w", err)
```
amritakohli, line 89: nit: error generating SHA256:\n
(reasoning - it seems the GenerateSHA256 function doesn't return any meaningful error string so best to mention it here so it's easier to identify where the error is coming from)
similar comment for line 93

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Verifies that the skippable frame has been correctly prepended to the partition file with the correct data
+func verifySkippableFrameMetadataFromFile(partitionFilepath string, magicNumber uint32, frameSize uint32, skippableFrameMetadata [SkippableFrameSize]byte) (err error) {
+	// Read existing data from the partition file.
+	existingData, err := os.ReadFile(partitionFilepath)
+	if err != nil {
+		return fmt.Errorf("failed to read partition file %s:\n%w", partitionFilepath, err)
+	}
+
+	// verify that the skippable frame has been prepended to the partition file by validating magicNumber
```
amritakohli, line 119: nit: title case for the comments in this function for consistency

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,142 @@
...
+	var magicNumberByteArray [4]byte
```
amritakohli, line 120: create variables for these numbers? as reader idk why it's 4 so having a variable could help explain that choice

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
Adub17030MS: In this piece fo code, `file, err := os.Create(fileName)` creates a `file` object. Then `file.Write` adda the byte array sotred in the `data` variable to the `file` object, so we do need this since `Write()` needs to be called from a `file` object, which is what is being created in line 63.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
Adub17030MS: Updated.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Added constant and comment.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAddSkippableFrame(t *testing.T) {
+	// Create a skippable frame containing the metadata and prepend the frame to the partition file
+	skippableFrameMetadata, err := createSkippableFrameMetadata()
+	assert.NoError(t, err)
+
+	// Create test raw partition file
+	partitionRawFilepath, err := createTestRawPartitionFile("test.raw")
```
cwize1, line 22: This package has a [main_test.go](https://github.com/microsoft/azurelinux/blob/3.0-dev/toolkit/tools/pkg/imagecustomizerlib/main_test.go) that sets up a temporary directory for the test code. It would probably be goodness to use it here as well.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
Adub17030MS: Changed the code so we are no longer using this

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Using the alternate solution and it works! Thanks! closing this comment.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
Adub17030MS: Added this change in.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,127 @@
...
+		return fmt.Errorf("failed to read partition file %s:\n%w", partitionFilepath, err)
+	}
+
+	// Verify that the skippable frame has been prepended to the partition file by validating magicNumber
+	var magicNumberByteArray [4]byte
+	binary.LittleEndian.PutUint32(magicNumberByteArray[:], magicNumber)
+	if !bytes.Equal(existingData[0:4], magicNumberByteArray[:]) {
+		return fmt.Errorf("skippable frame has not been prepended to the partition file:\n %d != %d", existingData[0:4], magicNumberByteArray[:])
+	}
+	logger.Log.Infof("Skippable frame had been correctly prepended to the partition file...")
```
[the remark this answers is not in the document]
Adub17030MS: Added this change in.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -11,6 +14,12 @@ import (
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
 )
 
+const (
+	SkippableFrameMagicNumber uint32      = 0x184D2A50
+	SkippableFrameSize        uint32      = 16
```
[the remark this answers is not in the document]
Adub17030MS: Added this change in.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -20,9 +31,16 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
+	// Create skippable frame metadata defined as a random 128-Bit number
+	skippableFrameMetadata, err := createSkippableFrameMetadata()
+	if err != nil {
+		return err
+	}
+
 	for partitionNum := 0; partitionNum < len(diskPartitions); partitionNum++ {
 		if diskPartitions[partitionNum].Type == "part" {
-			rawFilename := basename + "_" + strconv.Itoa(partitionNum) + ".raw"
+			partitionfileName := basename + "_" + strconv.Itoa(partitionNum)
```
amritakohli, line 42: nit: partitionFilename*

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +121,72 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
 		return "", fmt.Errorf("failed to compress %s with zstd:\n%w", partitionRawFilepath, err)
 	}
 
-	// Remove raw file since output partition format is raw-zst.
-	err = os.Remove(partitionRawFilepath)
+	return partitionRawFilepath + ".zst", nil
+}
+
+// Prepend a skippable frame with the metadata to the specified partition file.
+func addSkippableFrame(tempPartitionFilepath string, skippableFrameMetadata [SkippableFramePayloadSize]byte, partitionfileName string, outDir string) (partitionFilepath string, err error) {
```
amritakohli, line 128: partitionFilename*

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -34,20 +52,44 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
 			}
-
 			logger.Log.Infof("Partition file created: %s", partitionFilepath)
 		}
 	}
 	return nil
 }
 
+// Extract raw-zst partition
+func extractRawZstPartition(partitionRawFilepath string, skippableFrameMetadata [SkippableFramePayloadSize]byte, partitionfileName string, outDir string) (partitionFilepath string, err error) {
```
amritakohli, line 69: partitionFilename*

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +121,72 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
...
+	// Open tempPartitionFile for reading
+	tempPartitionFile, err := os.OpenFile(tempPartitionFilepath, os.O_RDWR, PartitionFilePermissions)
 	if err != nil {
-		return "", fmt.Errorf("failed to remove raw file %s:\n%w", partitionRawFilepath, err)
+		return "", fmt.Errorf("failed to open partition file %s:\n%w", tempPartitionFilepath, err)
 	}
+	// Create a skippable frame
+	skippableFrame := createSkippableFrame(SkippableFrameMagicNumber, SkippableFramePayloadSize, skippableFrameMetadata)
+	// Define the final partition file path
+	partitionFilepath = outDir + "/" + partitionfileName + "_final" + ".raw.zst"
```
cwize1: The output filename shouldn't have `_final` in it.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -34,20 +52,44 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
+	if err != nil {
+		return "", err
+	}
+	// Remove temp partition file
+	err = os.Remove(tempPartitionFilepath)
+	if err != nil {
+		return "", fmt.Errorf("failed to remove temp file %s:\n%w", partitionRawFilepath, err)
+	}
+	// Remove raw file since output partition format is raw-zst
+	err = os.Remove(partitionRawFilepath)
```
cwize1, line 86: You can delete `partitionRawFilepath` after `compressWithZstd()`.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -34,20 +52,44 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
 			logger.Log.Infof("Partition file created: %s", partitionFilepath)
 		}
 	}
 	return nil
 }
 
+// Extract raw-zst partition
+func extractRawZstPartition(partitionRawFilepath string, skippableFrameMetadata [SkippableFramePayloadSize]byte, partitionfileName string, outDir string) (partitionFilepath string, err error) {
+	// Compress raw partition with zstd
+	tempPartitionFilepath, err := compressWithZstd(partitionRawFilepath)
```
amritakohli, line 71: where is partitionRawFilepath coming from?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
amritakohli, line 137: we do not want to change the naming convention of the output files

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAddSkippableFrame(t *testing.T) {
+	// Create a skippable frame containing the metadata and prepend the frame to the partition file
+	skippableFrameMetadata, err := createSkippableFrameMetadata()
+	assert.NoError(t, err)
+
+	// Create test raw partition file
+	partitionfileName := "test"
```
amritakohli, line 22: partitionFilename *

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +121,72 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
...
 
-	// Remove raw file since output partition format is raw-zst.
-	err = os.Remove(partitionRawFilepath)
+	return partitionRawFilepath + ".zst", nil
+}
+
+// Prepend a skippable frame with the metadata to the specified partition file.
+func addSkippableFrame(tempPartitionFilepath string, skippableFrameMetadata [SkippableFramePayloadSize]byte, partitionfileName string, outDir string) (partitionFilepath string, err error) {
+	// Open tempPartitionFile for reading
+	tempPartitionFile, err := os.OpenFile(tempPartitionFilepath, os.O_RDWR, PartitionFilePermissions)
```
cwize1, line 130: You can just use `os.ModePerm` (i.e. `0o777`) instead of `0o644`. The [umask](https://en.wikipedia.org/wiki/Umask) will also apply to the permissions and will ensure it is something sensible.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: It is defined in line 46 - it is the filepath of the raw partition file

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -34,20 +52,44 @@ func extractPartitions(imageLoopDevice string, outDir string, basename string, p
...
+	}
+	// Create a skippable frame containing the metadata and prepend the frame to the partition file
+	partitionFilepath, err = addSkippableFrame(tempPartitionFilepath, skippableFrameMetadata, partitionFilename, outDir)
+	if err != nil {
+		return "", err
+	}
+	// Remove temp partition file
+	err = os.Remove(tempPartitionFilepath)
+	if err != nil {
+		return "", fmt.Errorf("failed to remove temp file %s:\n%w", partitionRawFilepath, err)
```
amritakohli, line 83: tempPartitionFilepath*?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
amritakohli: // Write writes a string to the file dst.
func Write(data string, dst string) (err error) {
	logger.Log.Debugf("Writing to (%s)", dst)

	dstFile, err := os.Create(dst)
	if err != nil {
		return
	}
	defer dstFile.Close()

	_, err = dstFile.WriteString(data)
	return
}

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Fixed the naming so that the names of the final partition files are the same as before.

Adub17030MS: Fixed the naming so that the names of the final partition files are the same as before.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Updated

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -79,11 +121,72 @@ func compressWithZstd(partitionRawFilepath string) (partitionFilepath string, er
...
+	tempPartitionFile, err := os.OpenFile(tempPartitionFilepath, os.O_RDWR, PartitionFilePermissions)
 	if err != nil {
-		return "", fmt.Errorf("failed to remove raw file %s:\n%w", partitionRawFilepath, err)
+		return "", fmt.Errorf("failed to open partition file %s:\n%w", tempPartitionFilepath, err)
 	}
+	// Create a skippable frame
+	skippableFrame := createSkippableFrame(SkippableFrameMagicNumber, SkippableFramePayloadSize, skippableFrameMetadata)
+	// Define the final partition file path
+	partitionFilepath = outDir + "/" + partitionFilename + "_final" + ".raw.zst"
+	// Create partition file
```
amritakohli, line 140: I would recommend writing a Prepend() function to handle all of this such that the resulting file name is the same
something like:
1) create a new temp file
2) add the data you want to prepend to it
3) add the data from old file to new file
4) delete the old file
5) rename / mv the temp file to the original filename
you could add this to file.go perhaps
https://github.com/microsoft/azurelinux/blob/4cb4cf97332d38647b4648157e0a826cc6ff73e9/toolkit/tools/internal/file/file.go

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Would I need to change the PartitionFilePermissions to 0777 too?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Updated

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -114,18 +116,18 @@ func copyBlockDeviceToFile(outDir, devicePath, name string) (filename string, er
...
+func compressWithZstd(partitionRawFilepath string, outputPartitionFilepath string) (partitionFilepath string, err error) {
 	// Using -f to overwrite a file with same name if it exists.
-	err = shell.ExecuteLive(true, "zstd", "-f", "-9", "-T0", partitionRawFilepath)
+	err = shell.ExecuteLive(true, "zstd", "-f", "-9", "-T0", partitionRawFilepath, "-o", outputPartitionFilepath)
 	if err != nil {
 		return "", fmt.Errorf("failed to compress %s with zstd:\n%w", partitionRawFilepath, err)
 	}
 
-	return partitionRawFilepath + ".zst", nil
+	return outputPartitionFilepath, nil
```
cwize1, line 126: Do you still need to return `outputPartitionFilepath` anymore?

cwize1 (approved):

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Updated to use this, thanks!

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
Adub17030MS: Good point, I do not, I just updated it

PR #8248, opened 2024-03-05
Disable empty checks for 5 python packages

0xba1a:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Remove check section for check-disabled python packages

###### Change Log  

Remove check section for
 - python-hatch-vcs
 - python-hatchling
 - python-pathspec
 - python-trove-classifiers
 - python-platformdirs

**NO**

###### Test Methodology

- Pipeline build id: local build

### DIFF
--- a/SPECS/python-hatch-vcs/python-hatch-vcs.spec
+++ b/SPECS/python-hatch-vcs/python-hatch-vcs.spec
@@ -69,8 +69,8 @@ Summary:        %{summary}
 %pyproject_save_files hatch_vcs
 
 
-%check
 %if 0%{?with_check}
+%check
 %if %{with tests}
 %pytest
 %else
--- a/SPECS/python-hatchling/python-hatchling.spec
+++ b/SPECS/python-hatchling/python-hatchling.spec
@@ -75,8 +75,8 @@ install -t '%{buildroot}%{_mandir}/man1' -D -p -m 0644 \
     '%{SOURCE500}'
 
 
-%check
 %if 0%{?with_check}
+%check
 %pytest
 %endif
 
--- a/SPECS/python-pathspec/python-pathspec.spec
+++ b/SPECS/python-pathspec/python-pathspec.spec
@@ -57,8 +57,8 @@ is derived from Rsync's wildmatch. Git uses wildmatch for its gitignore files.
 %pyproject_save_files pathspec
 
 
-%check
 %if 0%{?with_check}
+%check
 %pytest
 %endif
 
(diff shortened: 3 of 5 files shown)


SPECS/python-hatch-vcs/python-hatch-vcs.spec
```diff
@@ -69,8 +69,8 @@ Summary:        %{summary}
 %pyproject_save_files hatch_vcs
 
 
-%check
 %if 0%{?with_check}
```
PawelWMS, line 72: Please remove `%global with_check 0` at the top of this spec and its in the `%check` stage. That macro is controlled by the build tooling and shouldn't be overridden by the spec. Also, there's no need to surround `%check` with it - we skip this stage automatically, when we build without tests.

SPECS/python-hatchling/python-hatchling.spec
```diff
@@ -75,8 +75,8 @@ install -t '%{buildroot}%{_mandir}/man1' -D -p -m 0644 \
     '%{SOURCE500}'
 
 
-%check
 %if 0%{?with_check}
```
PawelWMS, line 78: Same ask here: let's remove `%global with_check 0` from this spec and `%if 0%{?with_check}` from this line.

SPECS/python-hatch-vcs/python-hatch-vcs.spec
0xba1a: This is understandable. But my motivation is not to lose track of the code and explicitly disable the check section as I know it is a temporary case. This section will be disabled until the necessary packages are onboarded to Mariner.

At the same time, removing `%global with_check 0` will flag failures in PTest. As it is expected case, these failures don't add-up any value. So, I want to avoid the unnecessary PTest failures too.

PawelWMS: It would be better to keep the failures visible, so we don't lose track of them, in case we run out of time to come back to this immediately. Right now it's especially easy for that to happen, since we run the tests but don't anything in them, which marks them as passing - very misleading.

0xba1a: This PR completely disables the tests. So, we'll avoid the falsehood of misleading PTest success.

I don't think there is any solid usefulness in running the tests when we know it will fail. Historically we tend to disable failures if we're not going to fix them in the near future. As I don't have bandwidth to fix the test cases in the near future, IMHO it's the best option to disable them. And the comments in the SPEC file are indication of a backlog.

PR #8262, opened 2024-03-06
Fix VHD build: Make grub2 explictily require systemd-udev

Camelron:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Tools no longer pull in systemd into the setupchroot as a side effect of its set of explicit dependencies changing in 3.0 (grub2-pc -> grub2...). Additionally, a critical component from systemd, udevadm, has been moved to systemd-udev. The result of this was that grub2-install was lacking udevadm, causing it to break during the VHD builds.

Explicitly make grub2 depend on systemd-udev.

###### Change Log  

- Explicitly install systemd, systemd-udev, and its deps in setupchroot

NO

###### Test Methodology

- local build
image

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -95,7 +95,6 @@ func GetRequiredPackagesForInstall() []*pkgjson.PackageVer {
 	if runtime.GOARCH == "amd64" {
 		packageList = append(packageList, &pkgjson.PackageVer{Name: "grub2-pc"})
 	}
-	
 	packageList = append(packageList, &pkgjson.PackageVer{Name: "systemd-rpm-macros"})
 
 	return packageList
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -96,6 +96,10 @@ func GetRequiredPackagesForInstall() []*pkgjson.PackageVer {
 		packageList = append(packageList, &pkgjson.PackageVer{Name: "grub2-pc"})
 	}
 
+	packageList = append(packageList, &pkgjson.PackageVer{Name: "systemd"})
+	packageList = append(packageList, &pkgjson.PackageVer{Name: "systemd-udev"})
+	packageList = append(packageList, &pkgjson.PackageVer{Name: "systemd-rpm-macros"})
+
 	return packageList
 }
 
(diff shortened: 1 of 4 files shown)


Camelron:
VHD built successfully in pipelines: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=521502&view=results

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -96,6 +96,10 @@ func GetRequiredPackagesForInstall() []*pkgjson.PackageVer {
 		packageList = append(packageList, &pkgjson.PackageVer{Name: "grub2-pc"})
 	}
 
+	packageList = append(packageList, &pkgjson.PackageVer{Name: "systemd"})
+	packageList = append(packageList, &pkgjson.PackageVer{Name: "systemd-udev"})
```
PawelWMS, line 100: You mentioned making `systemd-udev` a run-time dependency of the package, which provides that tool running `udevadm`. I think that makes sense, if it cannot run properly without it. Could we do that instead of hard-coding the missing run-time dependencies in the toolkit's code?

PawelWMS: Btw. the tooling installs `grub2-pc` along with its dependencies, so adding a `Requires` should cause `systemd-udev` to be installed.

Camelron: I agree, adding an explicit dependency makes more sense. 

In practice, this tooling change is required because the set of packages added to the local repo mounted into the chroots is determined by GetRequiredPackagesForInstall. https://github.com/microsoft/azurelinux/blob/2.0/toolkit/tools/imagepkgfetcher/imagepkgfetcher.go#L154

I see that those packages are cloned with --alldeps in the tdnf transaction. In practice, I couldn't get it to pull systemd-rpm-macros, which causes the transaction to fail in the chroot.

Camelron: As for why systemd-rpm-macros doesn't end up cloned, it may be due to the weird macros not being accounted for in our tooling:

Requires(meta): (%{name}-rpm-macros = %{version}-%{release} if rpm-build)

I'm testing a permutation where grub2-pc requires systemd-udev and the tools explicitly pull rpm-macros.

PawelWMS: This is a bit puzzling. The downloads and installations are performed by the package manager (TDNF) - the tooling doesn't analyze the run-time dependencies at this point. And from a local check it looks like TDNF is trying to install `systemd-rpm-macros` along with `systemd` or `systemd-devel`. I'll try to find you soon and we can take a look at the problem together. I think adding a BR on `systemd-udev` into `grub2` should be enough.

Btw., I think RPM macros shouldn't really be needed for anything during a image build.

Camelron: Note: it is actually grub2 itself which packages grub2-install/grub2-probe which are the things that depend on systemd-udev. This experiment succeeded, see changes below.

PawelWMS (approved):

gjswalling (approved):

PR #8306, opened 2024-03-08
[3.0] dhcp/dhclient.conf: request classless static routes

cjp256:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Azure configures classless routes and cloud-init requires the configuration to be included in the lease file to behave correctly.

###### Change Log  

- Add rfc3442-classless-static-routes option to dhclient.conf

No

###### Test Methodology

Build & installed RPM locally and validated cloud-init behavior.

### DIFF
--- a/SPECS/dhcp/dhcp.spec
+++ b/SPECS/dhcp/dhcp.spec
@@ -1,7 +1,7 @@
 Summary:        Dynamic host configuration protocol
 Name:           dhcp
 Version:        4.4.3
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        MPLv2.0
 Url:            https://www.isc.org/dhcp/
 Source0:        ftp://ftp.isc.org/isc/dhcp/%{version}/%{name}-%{version}.tar.gz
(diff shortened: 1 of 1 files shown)


ddstreetmicrosoft (changes requested):
First, I'd much prefer if we updated to the latest (i.e. *last ever*) version of isc-dhcp, which already includes the classless-static-routes option (i.e. we shouldn't be defining it ourselves).
(EDIT: it's actually the Fedora package that [adds the classless route option](https://src.fedoraproject.org/rpms/dhcp/blob/rawhide/f/0012-RFC-3442-Classless-Static-Route-Option-for-DHCPv4-51.patch)...from looking at the patch, i'm not sure this PR to simply manually define option 121 will actually result in the dhcp client correctly handling the static routes in the server response)

Second, the entire isc-dhcp package has been [dead for almost 1.5 years](https://www.isc.org/download/#DHCP), and it pains me to see us continuing to actively use it, although I understand it's still used by cloud-init (until https://github.com/canonical/cloud-init/pull/4912). But I'll work through the pain, so ignore this item ;-)

Finally, I'm not entirely sure that requesting classless static routes *by default for all users* is the best thing to do? This option, if provided by the dhcp server, will cause the client to use *only* the static routes, and *not* use a default route (aka "gateway") provided by the dhcp server (at least, assuming the client follows the RFC correctly). And note that, IIUC, this change wouldn't affect *only* cloud-init's use of isc-dhcp-client, it would affect *all* use of isc-dhcp-client on the system. I'm not saying that is wrong, just that I'm not sure it is the behavior that *everyone* wants...can you provide a bit more detail on why it's the right thing to do for *all* users?

cjp256:
cloud-init 24.1 is introducing support for alternative dhcp clients.  Ubuntu 24.04 will be using dhcpcd out of the box.

FWIW Debian and Ubuntu's dhclient.conf includes these options so I don't imagine there's a breaking case here wrt requesting classless routes.  Azure DHCP responds with it regardless, so it's not really needed I guess.

cjp256:
OK I see what you're saying.  Yeah the patch to support that request isn't in Mariner so it's a moot point, we can just ditch it.  The `option rfc3442-classless-static-routes code 121 = array of unsigned integer 8;` is the critical part.

PR #8312, opened 2024-03-08
dhcp: update to final isc-dhcp version, which has been EOL since Dec …

ddstreetmicrosoft:
…2022

related to #8306

### DIFF
--- a/SPECS/dbus/dbus.spec
+++ b/SPECS/dbus/dbus.spec
@@ -2,7 +2,7 @@
 Summary:        DBus for systemd
 Name:           dbus
 Version:        1.15.8
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv2+ OR AFL
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
@@ -49,6 +49,7 @@ autoreconf -i
     --docdir=%{_versioneddocdir}  \
     --enable-libaudit=yes \
     --enable-selinux=yes \
+    --runstatedir=/run \
     --with-console-auth-dir=/run/console
 
 make %{?_smp_mflags}
@@ -91,6 +92,10 @@ make %{?_smp_mflags} check
 %{_libdir}/*.so
 
 %changelog
+* Wed Mar 20 2024 Dan Streetman <ddstreet@microsoft.com> - 1.15.8-2
+- specify runstatedir to avoid "Line references path below legacy
+  directory /var/run/" warnings
+
 * Thu Jan 04 2024 Brian Fjeldstad <bfjelds@microsoft.com> - 1.15.8-1
 - Upgrade to 1.15.8
 
--- a/SPECS/dhcp/dhcp.spec
+++ b/SPECS/dhcp/dhcp.spec
@@ -1,10 +1,10 @@
 Summary:        Dynamic host configuration protocol
 Name:           dhcp
-Version:        4.4.3
-Release:        2%{?dist}
+Version:        4.4.3.P1
+Release:        1%{?dist}
 License:        MPLv2.0
 Url:            https://www.isc.org/dhcp/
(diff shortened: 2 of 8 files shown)


cjp256 (approved):
LGTM, thanks!

cjp256:
If you want to address `option rfc3442-classless-static-routes code 121 = array of unsigned integer 8;` at the same time I can close the other PRs.

cjp256:
Will these be done for 2.0 as well?

ddstreetmicrosoft:
> If you want to address `option rfc3442-classless-static-routes code 121 = array of unsigned integer 8;` at the same time I can close the other PRs.

can you provide more detail about the specific results with and without this? for example, provide the dhcpdump of a dhcp session both with and without dhcp client conf change? and also the ip r difference(s)? I'd like to better understand exactly what our current dhcp will do differently, so we can make sure that we don't need any additional patch(es), like what's included in the Fedora packaging.

note that i don't think azure linux provides dhcpdump so you might need to use the [fedora rpm](https://packages.fedoraproject.org/pkgs/dhcpdump/dhcpdump/) or just compile [from source](https://github.com/bbonev/dhcpdump)

ddstreetmicrosoft:
> Will these be done for 2.0 as well?

since 2.0 is a stable release, i don't think these changes should be done there. of course, end users can modify the `dhclient.conf` in their image however they want.

cjp256:
> > Will these be done for 2.0 as well?
> 
> since 2.0 is a stable release, i don't think these changes should be done there. of course, end users can modify the `dhclient.conf` in their image however they want.

The primary NIC check in cloud-init 23.4 is failing because the static routes configured by Azure aren't processed.   This would be addressed by configuring dhclient to handle the classless static routes.  We'll need to add the following to /etc/dhcp/dhclient.conf:
```
option rfc3442-classless-static-routes code 121 = array of unsigned integer 8;
```



The lease file currently produced by dhclient looks something like:
```
lease {
  interface "eth0";
  <...cut...>
  option unknown-121 0:a:0:0:1:20:a8:3f:81:10:a:0:0:1:20:a9:fe:a9:fe:a:0:0:1;
  <...cut...>
}
```

After this change the lease file will show something like:

```
lease {
  interface "eth0";
  <...cut...>
  option rfc3442-classless-static-routes 0,10,0,0,1,32,168,63,129,16,10,0,0,1,32,169,254,169,254,10,0,0,1;
  <...cut...>
}

ddstreetmicrosoft:
> > > Will these be done for 2.0 as well?
> > 
> > 
> > since 2.0 is a stable release, i don't think these changes should be done there. of course, end users can modify the `dhclient.conf` in their image however they want.
> 
> The primary NIC check in cloud-init 23.4 is failing because the static routes configured by Azure aren't processed. This would be addressed by configuring dhclient to handle the classless static routes. We'll need to add the following to /etc/dhcp/dhclient.conf:
> 
> ```
> option rfc3442-classless-static-routes code 121 = array of unsigned integer 8;
> ```
> 
> The lease file currently produced by dhclient looks something like:
> 
> ```
> lease {
>   interface "eth0";
>   <...cut...>
>   option unknown-121 0:a:0:0:1:20:a8:3f:81:10:a:0:0:1:20:a9:fe:a9:fe:a:0:0:1;
>   <...cut...>
> }
> ```
> 
> After this change the lease file will show something like:
> 
> ```
> lease {
>   interface "eth0";
>   <...cut...>
>   option rfc3442-classless-static-routes 0,10,0,0,1,32,168,63,129,16,10,0,0,1,32,169,254,169,254,10,0,0,1;
>   <...cut...>
> }
> ```

I don't think a PR is the right place to get into the details of a bug; can you please open a github issue and describe the details of the problem?

For this PR, I think it should remain restricted to simply updating the version of isc-dhcp in azl 3.0.

christopherco (approved):

PR #8331, opened 2024-03-11
procps-ng: suppress debug messages to console by default

christopherco:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

During normal use, many low level debug messages are printing to the console which is causing usability issues for the user.

To deal with this, install a sysctl configuration to run on boot which will change the default kernel printk value to "4 4 1 7".

/etc/sysctl.d is supplied by systemd, so we need to also have a weak recommends on systemd. This is not a full requires because it would induce a circular dependency, and would pull in systemd in situations where it may not be truly required (i.e., inside minimal container images).

**YES**

###### Test Methodology

- Buddy Build: [524643](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=524643&view=results)

### DIFF
--- a/SPECS/procps-ng/10-console-messages.conf
+++ b/SPECS/procps-ng/10-console-messages.conf
@@ -1,0 +1,2 @@
+# Suppress debug messages on console
+kernel.printk = 4 4 1 7
--- a/SPECS/procps-ng/procps-ng.spec
+++ b/SPECS/procps-ng/procps-ng.spec
@@ -1,15 +1,17 @@
 Summary:        Programs for monitoring processes
 Name:           procps-ng
 Version:        4.0.4
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv2 AND LGPLv2
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 Group:          Applications/System
 URL:            https://gitlab.com/procps-ng/procps
 Source0:        https://sourceforge.net/projects/procps-ng/files/Production/%{name}-%{version}.tar.xz
+Source1:        10-console-messages.conf
 BuildRequires:  ncurses-devel
 Requires:       ncurses
+Recommends:     systemd
 Conflicts:      toybox
 Provides:       /bin/ps
 Provides:       procps = %{version}-%{release}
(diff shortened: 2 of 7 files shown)


jslobodzian (approved):

SPECS/procps-ng/procps-ng.spec
```diff
@@ -55,10 +57,21 @@ ln -s %{_bindir}/pidof %{buildroot}%{_sbindir}/pidof
 find %{buildroot} -type f -name "*.la" -delete -print
 %find_lang %{name} --all-name --with-man
 
+install -Dm 644 %{SOURCE1} %{buildroot}%{_sysconfdir}/sysctl.d/10-console-messages.conf
```
christopherco, line 60: @ddstreetmicrosoft would this be better supplied by `systemd` since `systemd` owns `/etc/sysctl.d`? I could also see this eventually live in `azurelinux-profile` packages, but these are not ready/available yet.

ddstreetmicrosoft: as `procps-ng` doesn't provide any configuration at all currently, it probably doesn't make a lot of sense to put this here. Putting it into `systemd` or `azurelinux-release` (or `azurelinux-release-$VARIANT` if we want to add that) probably makes sense. It would be good to have a better overall design of where we want config like this, but for now probably `systemd` is the best place to put it, as you suggested.

Note that it should go into `%{_prefix}/lib/sysctl.d/`, and not into `%{_sysconfdir}/sysctl.d`. Distro-provided config should go into `/usr/lib` while `/etc` is normally reserved for end-user-provided config.

SPECS/procps-ng/10-console-messages.conf
```diff
@@ -0,0 +1,2 @@
+# Suppress debug messages on console
+kernel.printk = 4 4 1 7
```
ddstreetmicrosoft, line 2: I notice that on fedora, they don't provide any overriding sysctl like this, yet i don't see lots of debug printed on their console...have we made sure there isn't some other change we should make?

christopherco:
Converting back to draft. Updating this change to supply the conf file from systemd package instead of procps-ng package

SPECS/procps-ng/10-console-messages.conf
christopherco: Good point. I thought Fedora doesn't need this because their default kernel printk from the kernel might be set lower than `=7` but this is a guess. Will double-check

christopherco: Confirmed that on f39 workstation, they use their kernel built-in defaults rather than an overriding sysctl. Instead, on the workstation image, by default it boots with `quiet` on the kernel command line, so their kernel's `CONFIG_CONSOLE_LOGLEVEL_QUIET=3` setting takes effect, rather than `CONFIG_CONSOLE_LOGLEVEL_DEFAULT=7`.

PR #8332, opened 2024-03-11
imagecustomizer: Optimize MIC Overlay supporting multiple volumes.

liulanze:
###### Merge Checklist  
- [x] Ready to merge

---

###### Summary 

Optimize the existing MIC Overlay API to support multiple volumes. Users can assign different persistent volumes to distinct Overlays. If no volumes are set, a volatile Overlay will be created.

**NO**

###### Associated issues  

- (internal) https://dev.azure.com/mariner-org/ECF/_sprints/taskboard/Mariner%20Operator%20Edge/ECF/2403?workitem=7107

###### Test Methodology

- Local build

### DIFF
--- a/SPECS/dracut/20overlayfs/overlayfs-mount.sh
+++ b/SPECS/dracut/20overlayfs/overlayfs-mount.sh
@@ -1,8 +1,8 @@
 #!/bin/sh
 
 # Description: This script is designed to mount a DM-Verity root filesystem and
-# set up an OverlayFS. It is driven by kernel parameters and is invoked during
-# the dracut initramfs phase.
+# set up OverlayFS. It is driven by kernel parameters and is invoked during the
+# dracut initramfs phase.
 
 # Kernel Parameters:
 # - root: Specifies the path to the root filesystem. This script is designed to
@@ -12,11 +12,8 @@
 #   setups, the script will proceed with the standard OverlayFS setup, ensuring
 #   versatility in its application.
 # - rd.overlayfs: A comma-separated list defining the OverlayFS configuration.
-#   Each entry should specify the overlay, upper, and work directories for an
-#   OverlayFS instance.
-# - rd.overlayfs_persistent_volume: Specifies the path to a persistent storage
-#   volume to be used by OverlayFS. If not provided, a volatile (tmpfs) overlay
-#   is created.
+#   Each entry should specify the overlay, upper, work directories, and optional
+#   volume for an OverlayFS instance.
 
 # Behavior:
 # - Verifies the presence of the 'dracut-lib' for necessary utilities.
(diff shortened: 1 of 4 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeoverlay.go
```diff
@@ -46,6 +46,9 @@ func updateGrubConfigForOverlay(imageChroot *safechroot.Chroot, overlays []image
 
 	// Iterate over each Overlay configuration
 	for _, overlay := range overlays {
+		if overlay.Partition == nil {
+			formattedPartition = ""
+		}
 		if overlay.Partition != nil {
 			formattedPartition, err = systemdFormatPartitionId(overlay.Partition.IdType, overlay.Partition.Id)
 			if err != nil {
```
Adub17030MS, line 54: This can be simplified by initializing `formattedPartition = ""` as the default value outside the if statement and only changing the value if `overlay.Partition != nil`. So we only need the second if statement here.

Adub17030MS (changes requested):
Can you provide some more information in the PR description about how this was validated and tested?

toolkit/tools/pkg/imagecustomizerlib/customizeoverlay.go
liulanze: Fixed.

cwize1 (approved):

Adub17030MS (approved):

SPECS/dracut/20overlayfs/overlayfs-mount.sh
```diff
@@ -54,49 +51,51 @@ parse_cmdline_args() {
...
         echo "No overlayfs persistent volume specified. Creating a volatile overlay."
-        mount -t tmpfs tmpfs -o ${OVERLAY_MNT_OPTS} "${OVERLAY_MOUNT}" || \
-            die "Failed to create overlay tmpfs at ${OVERLAY_MOUNT}"
+        mount -t tmpfs tmpfs -o ${OVERLAY_MNT_OPTS} "${_overlay_mount}" || \
+            die "Failed to create overlay tmpfs at ${_overlay_mount}"
     else
         # Check if /etc/mdadm.conf exists.
         if [ -f "/etc/mdadm.conf" ]; then
-            mdadm --assemble ${partition} || \
+            mdadm --assemble ${_volume} || \
```
dmcilvaney, line 71: Not needed to fix here, but dracut should already be able to assemble mdadm raid on its own. If you add an entry to the init queue to wait for it I think that should work?

dmcilvaney (approved):

SPECS/dracut/20overlayfs/overlayfs-mount.sh
liulanze: Mark here, thanks.

PR #8336, opened 2024-03-11
glibc: provide C.utf8 locale

ddstreetmicrosoft:
We should provide the `C.utf8` locale, as most (but not all) distros provide it, and it's the "default" system locale set by systemd (in the `/etc/locale.conf` file, unless changed using `localectl` or other method of setting locale).

Currently we only provide `en_US.utf8`, which is fine if we want to provide *only* that, but we would need to adjust the systemd-provided "default" locale. I think it makes more sense to provide the common `C.utf8` locale.

Note that if the system locale (i.e. what's in `/etc/locale.conf`) is not actually provided by the system, bad things might happen, like segfaults (e.g. from readline lib, and probably other places)


ddstreetmicrosoft:
ugh i really hate the `glibc-static` pr check

PawelWMS (approved):

anphel31 (approved):
I built an image with this fix and confirmed it resolves the login issue related to cloud-init

PR #8337, opened 2024-03-11
Image Customizer: Changes to bootloader reset.

cwize1:
The change does two things.

Firstly, it moves bootloader reset operation so that it is done after package installation. If a new partition layout requires additional dependencies (e.g. the grub2 package), this allows those dependencies to be easily installed before the bootloader is reset.

Secondly, a new config field called `ResetBootLoaderType` has been added. This field doesn't add any additional logic. Instead, it used to make the user explicitly acknowledge that the bootloader must be reset when partitions are customized.

###### Merge Checklist  

- [x] Ready to merge

---

NO

###### Test Methodology

### DIFF
--- a/toolkit/tools/imagecustomizer/docs/configuration.md
+++ b/toolkit/tools/imagecustomizer/docs/configuration.md
@@ -35,9 +35,12 @@ The Azure Linux Image Customizer is configured using a YAML (or JSON) file.
 
 10. Run post-install scripts. ([PostInstallScripts](#postinstallscripts-script))
 
-11. If [ResetBootLoaderType](#resetbootloadertype-string) is specified, then reset the
-    boot-loader.
-    Otherwise, append kernel command-line args to existing boot-loader config.
+11. If [ResetBootLoaderType](#resetbootloadertype-string) is set to `hard-reset`, then
+    reset the boot-loader.
+
+    If [ResetBootLoaderType](#resetbootloadertype-string) is not set, then
+    append the [ExtraCommandLine](#extracommandline-string) value to the existing
+    `grub.cfg` file.
 
 12. Change SELinux mode and, if SELinux is enabled, call `setfiles`.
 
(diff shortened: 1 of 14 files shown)


toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -36,7 +35,9 @@ The Mariner Image Customizer is configured using a YAML (or JSON) file.
 
 10. Run post-install scripts. ([PostInstallScripts](#postinstallscripts-script))
 
-11. Apply kernel command-line args, if the partitions weren't customized.
+11. If [ResetBootLoaderType](#resetbootloadertype-string) is specified, then reset the
+    boot-loader.
+    Otherwise, append kernel command-line args to existing boot-loader config.
```
romoh, line 40: Not clear if that appends the kernel cmdline from the input images or from the config or both? 
Also maybe call out explicitly that this is the default behavior in the paritions API section?

romoh: Reading the rest of the PR, you have it explained wel later. Though it looks like ResetBootLoaderType is mandatory so you may want to reword this a bit to call out "ResetBootLoaderType is set to `hard-reset` instead of the "If" statement

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -323,12 +326,10 @@ Options for configuring the kernel.
...
 
-If the partitions are customized, then the `grub.cfg` file will be reset to handle the
-new partition layout.
-So, any existing ExtraCommandLine value in the base image will be replaced.
-
-If the partitions are not customized, then the `ExtraCommandLine` value will be appended
-to the existing `grub.cfg` file.
+If [ResetBootLoaderType](#resetbootloadertype-string) is set to `"hard-reset"`, then the
+`ExtraCommandLine` value will be appended to the new `grub.cfg` file.
+Otherwise, the `ExtraCommandLine` value will be appended to the existing `grub.cfg`
```
romoh, line 331: Is that an option?

cwize1: It is. I've rewrote this to hopefully make it clearer.

toolkit/tools/imagecustomizer/docs/configuration.md
cwize1: It is only mandatory if partitions are customized.

romoh (approved):

PR #8346, opened 2024-03-12
Prevent -devel package inclusion in core images (from elfutils)

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Prevent *-devel packages from being pulled into core images. These should not be needed in the core images, and they increase the disk usage.
Reduces core EFI VHDX by ~4MB.

###### Change Log  

- Remove automatic requirements (AutoReq) from being generated in core `elfutils` package. This was automatically adding dependencies on development libraries, which pulled the following devel packages when elfutils is installed: `elfutils-devel`, `elfutils-libelf-devel`, `xz-devel`, `zlib-devel`, `zstd`, `zstd-devel`
Reduces core image size by ~4MB. 

**YES**

###### Test Methodology

- Buddybuild: 
- Local build. Built core VHDX before and after this change. Confirm that `-devel` packages are no longer in image.
`time sudo make image CONFIG_FILE=./imageconfigs/core-efi.json REBUILD_TOOLS=y SRPM_PACK_LIST="" DAILY_BUILD_ID=3-0-20240311 REBUILD_PACKAGES=n`
Boot the VHDX in Hyper-V and run "df -h" to compare disk usage, and diff the `build/logs/imggen/imager.log` file to compare the packages pulled into the image.


SPECS/python-setuptools/python-setuptools.spec
```diff
@@ -19,14 +19,15 @@ Source0:        https://pypi.org/packages/source/s/setuptools/setuptools-%{versi
 %package -n python3-setuptools
 Summary:        %{summary}
 
-Requires:       python3-devel
+Requires:       python3
```
chalamalasetty, line 22: Move this Requires below after BuildRequires entries.

chalamalasetty (approved):

SPECS/python-setuptools/python-setuptools.spec
anphel31: will change

SPECS/elfutils/elfutils.spec
```diff
@@ -15,15 +15,14 @@ Source0:        https://sourceware.org/elfutils/ftp/%{version}/%{name}-%{version
 BuildRequires:  bison >= 1.875
 BuildRequires:  bzip2-devel
 BuildRequires:  flex >= 2.5.4a
-BuildRequires:  gcc >= 4.1.2-33
+BuildRequires:  gcc
 BuildRequires:  gettext
-BuildRequires:  glibc >= 2.7
 BuildRequires:  m4
+AutoReq:        no
```
PawelWMS, line 21: Disabling `AutoReq` is risky because this forces us to figure out all of `elfutils`'s run-time dependencies manually and keep them up-to-date in the spec. In case of any reshuffling of files in other packages this one depends on, or an addition of a new dependency due to a version change here, we may miss it and the package may not work correctly.

If `elfutils` are currently pulling in the `-devel` packages, then it's possible the specs, which generate these `-devel` packages may be putting run-time libraries into them. In other words - this looks more like a misconfiguration in other packages.

Also, I see that `elfutils` only directly depends on two `-devel` packages: `bzip2-devel` and `elfutils-devel`.
image

All other ones are probably run-time dependencies of these dependencies, so it's even more possible we're going to lose valid run-time dependencies here by adding `AutoReq: no`.

anphel31: The automatic dependencies added are to shared object files provided by devel packages, for example:
```
$ sudo rpm -q --requires ../build/toolchain/built_rpms_all/elfutils-0.189-1.azl3.x86_64.rpm^C
liblzma.so.5()(64bit)
liblzma.so.5(XZ_5.0)(64bit)
libz.so.1()(64bit)
libz.so.1(ZLIB_1.2.2.3)(64bit)
libzstd.so.1()(64bit)
```
Which pulls in zstd-devel, xz-devel, etc.
I'll see if there is a better way to remove these dependencies. It doesn't look like the Fedora eflutils requires them: https://rpmfind.net/linux/RPM/fedora/updates/testing/40/x86_64/Packages/e/elfutils-0.191-2.fc40.x86_64.html

anphel31: Thanks for the help, the better fix for this is in a fresh PR: https://github.com/microsoft/azurelinux/pull/8373

anphel31:
I've split this into 2 PRs:
python-setuptools: #8363 
elfutils: #8373

PR #8364, opened 2024-03-12
Image Customizer: Fix bugs for ISO gen + kernel-mshv

cwize1:
Fix a couple of bugs related to taking image that uses the kernel-mshv package and turning it into an ISO.

Firstly, when uninstalling the kernel package, the /usr/lib/modules/<version> directory is emptied but it isn't removed. This confuses the image customizer's logic to discover the list of installed kernels. This change updates the logic to filter out empty directories.

Secondly, when generating the ISO, the files under the /boot directory are copied into the ISO. But this logic fails if it encounters a broken symlink. For example, a `mariner.cfg` file that wasn't removed when the `kernel` package was uninstalled. This change ensures that symlinks are copied as symlinks.

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

### DIFF
--- a/toolkit/tools/internal/file/file.go
+++ b/toolkit/tools/internal/file/file.go
@@ -86,19 +86,7 @@ func Move(src, dst string) (err error) {
 // Copy copies a file from src to dst, creating directories for the destination if needed.
 // dst is assumed to be a file and not a directory. Will preserve permissions.
 func Copy(src, dst string) (err error) {
-	return copyWithPermissions(src, dst, os.ModePerm, false, os.ModePerm, false)
-}
-
-// CopyAndChangeMode copies a file from src to dst, creating directories with the given access rights for the destination if needed.
-// dst is assumed to be a file and not a directory. Will change the permissions to the given value.
-func CopyAndChangeMode(src, dst string, dirmode os.FileMode, filemode os.FileMode) (err error) {
-	return copyWithPermissions(src, dst, dirmode, true, filemode, false)
-}
-
-// CopyNoDereference copies a file from src to dst, creating directories for the destination if needed.
-// dst is assumed to be a file and not a directory. Will preserve permissions and symlinks.
-func CopyNoDereference(src, dst string) (err error) {
-	return copyWithPermissions(src, dst, os.ModePerm, false, os.ModePerm, true)
+	return NewFileCopyBuilder(src, dst).Run()
 }
 
 // Read reads a string from the file src.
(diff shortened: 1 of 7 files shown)


toolkit/tools/internal/file/file.go
```diff
@@ -39,6 +39,17 @@ func IsFile(path string) (isFile bool, err error) {
 	return !info.IsDir(), nil
 }
 
+// IsFileOrSymlink returns true if the provided path is a file or a symlink.
+func IsFileOrSymlink(path string) (isFile bool, err error) {
```
neha170, line 43: nit: since we already have `isFile` could you restrict this function to just `isSymlink` and use both in your usecase?

toolkit/tools/internal/file/file.go
```diff
@@ -245,27 +262,56 @@ func GetAbsPathWithBase(baseDirPath, inputPath string) string {
...
+
 // copyWithPermissions copies a file from src to dst, creating directories with the requested mode for the destination if needed.
 // Depending on the changeMode parameter, it may also change the file mode.
-func copyWithPermissions(src, dst string, dirmode os.FileMode, changeMode bool, filemode os.FileMode) (err error) {
+func copyWithPermissions(src, dst string, dirmode os.FileMode, changeMode bool, filemode os.FileMode,
+	noDereference bool,
+) (err error) {
 	const squashErrors = false
 
 	logger.Log.Debugf("Copying (%s) -> (%s)", src, dst)
```
neha170, line 283: nit: could probably make two variables here, `isFile` and `isSymlink`. If-else would read better imo

toolkit/tools/internal/file/file.go
```diff
@@ -75,13 +86,19 @@ func Move(src, dst string) (err error) {
...
 }
 
 // CopyAndChangeMode copies a file from src to dst, creating directories with the given access rights for the destination if needed.
 // dst is assumed to be a file and not a directory. Will change the permissions to the given value.
 func CopyAndChangeMode(src, dst string, dirmode os.FileMode, filemode os.FileMode) (err error) {
-	return copyWithPermissions(src, dst, dirmode, true, filemode)
+	return copyWithPermissions(src, dst, dirmode, true, filemode, false)
+}
+
+// Copy copies a file from src to dst, creating directories for the destination if needed.
```
amritakohli, line 98: CopyNoDereference*

amritakohli (changes requested):

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -572,6 +572,7 @@ func (im *IsoMaker) copyAndRenameConfigFiles() (err error) {
 // files can be used by custom initrd/LiveOS images that will look for them
 // on the iso media.
 func (im *IsoMaker) copyIsoAdditionalFiles() (err error) {
+	logger.Log.Debugf("Copying ISO additional files.")
```
neha170, line 575: nit: please remove `.` at the end, as part of the cleanup effort for logging

toolkit/tools/internal/file/file.go
```diff
@@ -245,27 +262,56 @@ func GetAbsPathWithBase(baseDirPath, inputPath string) string {
...
+
+	empty := len(contents) == 0
+	return empty, nil
+}
+
 // copyWithPermissions copies a file from src to dst, creating directories with the requested mode for the destination if needed.
 // Depending on the changeMode parameter, it may also change the file mode.
-func copyWithPermissions(src, dst string, dirmode os.FileMode, changeMode bool, filemode os.FileMode) (err error) {
+func copyWithPermissions(src, dst string, dirmode os.FileMode, changeMode bool, filemode os.FileMode,
+	noDereference bool,
```
amritakohli, line 278: can you elaborate in comments somewhere on what noDereference is

toolkit/tools/internal/file/file.go
```diff
@@ -75,13 +86,19 @@ func Move(src, dst string) (err error) {
...
 // CopyAndChangeMode copies a file from src to dst, creating directories with the given access rights for the destination if needed.
 // dst is assumed to be a file and not a directory. Will change the permissions to the given value.
 func CopyAndChangeMode(src, dst string, dirmode os.FileMode, filemode os.FileMode) (err error) {
-	return copyWithPermissions(src, dst, dirmode, true, filemode)
+	return copyWithPermissions(src, dst, dirmode, true, filemode, false)
+}
+
+// Copy copies a file from src to dst, creating directories for the destination if needed.
+// dst is assumed to be a file and not a directory. Will preserve permissions and symlinks.
+func CopyNoDereference(src, dst string) (err error) {
```
amritakohli, line 100: would CopyWithoutDereference be a better name to explain the behavior?

neha170, line 100: random thought, probably an overkill: with so many versions of func copy*, do you think having a builder would be more appropriate?

neha170: Since argument to `cp` is `--no-dereference` I personally think `CopyNoDereference` suits better

toolkit/tools/internal/file/file.go
cwize1: That would mean 2 syscalls instead of 1, which seems a little wasteful.

toolkit/tools/internal/file/file.go
cwize1: Fixed

toolkit/tools/pkg/isomakerlib/isomaker.go
cwize1: Fixed

toolkit/tools/internal/file/file.go
cwize1: Duplicate comment.

amritakohli (approved):

toolkit/tools/internal/file/file.go
cwize1: Fixed

dmcilvaney (approved):

neha170 (approved):
Thanks for making the changes, and adding a new test :)

PR #8368, opened 2024-03-12
Auto clean tdnf cache

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

The toolkit should be able to handle cleaning up the `tdnf` cache itself, no need to oblige users to call a post-install script for ever image.

###### Change Log  

- By default run `tdnf clean all` and remove `/var/cache/tdnf` at the end of image builds.
- Add `PreserveTdnfCache` to `systemconfig` to keep old behavior.

**NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- local build of baremetal and core-efi

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -562,6 +562,17 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 
 	// Run post-install scripts from within the installroot chroot
 	err = runPostInstallScripts(installChroot, config)
+	if err != nil {
+		return
+	}
+
+	// The system should be fully populated with packages, we can clear the tdnf cache now to free up space.
+	if !config.PreserveTdnfCache {
+		err = cleanupTdnfCache(installChroot)
+		if err != nil {
+			return
+		}
+	}
 	return
 }
 
@@ -2180,6 +2191,33 @@ func runPostInstallScripts(installChroot *safechroot.Chroot, config configuratio
 	}
 
 	return
+}
+
+func cleanupTdnfCache(installChroot *safechroot.Chroot) error {
+	const (
+		squashErrors      = false
+		rpmCacheDirectory = "/var/cache/tdnf"
+	)
+
+	ReportActionf("Cleaning tdnf cache")
+	err := installChroot.UnsafeRun(func() error {
+		chrootErr := shell.ExecuteLive(squashErrors, "tdnf", "clean", "all")
+		return chrootErr
+	})
+
+	if err != nil {
+		err = fmt.Errorf("failed to cleanup tdnf cache:\n%w", err)
+		return err
+	}
+
+	cacheDir := filepath.Join(installChroot.RootDir(), rpmCacheDirectory)
+	err = os.RemoveAll(cacheDir)
+	if err != nil {
+		err = fmt.Errorf("failed to remove tdnf cache directory (%s):\n%w", cacheDir, err)
+		return err
+	}
+
+	return nil
 }
 
 func RunFinalizeImageScripts(installChroot *safechroot.Chroot, config configuration.SystemConfig) (err error) {
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -5,6 +5,7 @@ package installutils
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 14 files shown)


toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -2159,6 +2170,33 @@ func runPostInstallScripts(installChroot *safechroot.Chroot, config configuratio
...
+		return err
+	}
+
+	cacheDir := filepath.Join(installChroot.RootDir(), rpmCacheDirectory)
+	err = os.RemoveAll(cacheDir)
+	if err != nil {
+		err = fmt.Errorf("failed to remove tdnf cache directory (%s):\n%w", cacheDir, err)
+		return err
+	}
+
```
romoh, line 2196: Is that really needed? 
I am not in favor of removing dirs as in the past that led to SELinux labels being messed up on boot.

mbykhovtsev-ms: I believe the right way to clean tdnf cache is to remove `/var/cache/tdnf`. But alternatively we could just remove everything in that folder itself

dmcilvaney: fwiw this is a direct copy of what we are doing today via script. Won't claim one way or the other which is correct.

mbykhovtsev-ms: Yeah, I can confirm that it is necessary to at *least* delete everything in `/var/cache/tdnf`, I worked added this to our core images scripts and found that just running `tdnf -y clean all` was not sufficient.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -2159,6 +2170,33 @@ func runPostInstallScripts(installChroot *safechroot.Chroot, config configuratio
...
 
+func cleanupTdnfCache(installChroot *safechroot.Chroot) error {
+	const (
+		squashErrors      = false
+		rpmCacheDirectory = "/var/cache/tdnf"
+	)
+
+	ReportActionf("Cleaning tdnf cache")
+	err := installChroot.UnsafeRun(func() error {
+		chrootErr := shell.ExecuteLive(squashErrors, "tdnf", "clean", "all")
```
mbykhovtsev-ms, line 2181: Let me know if this line works. When I tried the same I was running into some error, I assumed I needed to provide `--installroot` parameter

dmcilvaney: Might be running from the setup chroot rather than the install chroot, that would likely require --installroot.

toolkit/tools/imagegen/installutils/installutils.go
romoh: can we remove files under /var/cache/tdnf instead? 
I don't believe any of the current images that uses that have SELinux on.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -2159,6 +2171,56 @@ func runPostInstallScripts(installChroot *safechroot.Chroot, config configuratio
...
+		if d.IsDir() {
+			return fs.SkipDir
+		} else {
+			return nil
+		}
+	})
+	if err != nil {
+		err = fmt.Errorf("failed to scan tdnf cache directory for files to remove(%s):\n%w", cacheDir, err)
+		return err
+	}
```
PawelWMS, line 2220: Can we move this into `file.go`? Seems like a useful function.

Also, we could do something similar to (or identical) `RemoveContents` from [here](https://stackoverflow.com/a/33451503/11313995). It doesn't require separate logic for skipping the top directory and subdirectories.

PawelWMS (approved):

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: Done

toolkit/tools/internal/file/file_test.go
```diff
@@ -43,3 +43,83 @@ func TestRemoveFileDoesNotExist(t *testing.T) {
 	err := RemoveFileIfExists(fileName)
 	assert.NoError(t, err)
 }
+func TestRemoveDirectoryContents(t *testing.T) {
```
dmcilvaney, line 46:
```suggestion

func TestRemoveDirectoryContents(t *testing.T) {
```

dmcilvaney:
Queuing build to check images: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=542246&view=results

dmcilvaney:
Build failed due to unrelated signature issues. Rebased and rebuilding.

dmcilvaney:
ISO build deletes the full /var/cache folder, handle the case where the directory is already gone.

dmcilvaney:
rerunning images: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=545265&view=results

PR #8380, opened 2024-03-13
Fix: Signature check not handling files in subfolders

bfjelds:
Signature check was not handling PRs that changed files embedded in subfolders.

This BuidlRpms break (https://dev.azure.com/mariner-org/mariner/_build/results?buildId=526527&view=logs&j=75cba72d-09cb-503a-3927-4ab61ac4a179&t=88a626d5-bd2d-5bcb-0f14-68ce5ab12667) happened because a .sh file had a mismatched signature.  The .sh file was not a file in the same folder as the signatures/spec file, it was in a subfolder.

The changes here:
* look for SPECS|SPECS-EXTENDED folder that have changed in PR to get signatures from, search all signatures.json with corresponding spec file.
* when finding files to check signatures for, search all subfolders for a file with the matching name

### DIFF
--- a/toolkit/scripts/check_signatures.py
+++ b/toolkit/scripts/check_signatures.py
@@ -34,8 +34,8 @@ def find_file_and_check(path, filename, expected_signature) -> Optional[bool]:
         path_to_check = os.path.join(path, content)
         if os.path.isdir(path_to_check):
             result = find_file_and_check(path_to_check, filename, expected_signature)
-            if result is True:
-                return True
+            if result is not None:
+                return result
 
     return None
 
--- a/toolkit/scripts/check_signatures.py
+++ b/toolkit/scripts/check_signatures.py
@@ -9,6 +9,7 @@ import argparse
 import hashlib
 import json
 import os
+import re
 import sys
 
 def getSignature(fileName) -> str:
@@ -48,26 +49,19 @@ def find_name_of_all_spec_and_signatures_json_pairs(path: str) -> List[str]:
     return names
 
 def find_spec_folder_with_signatures_json(path: str) -> Optional[str]:
-    # Use this path if there are any spec files (XXX.spec) that have
-    # a matching signature file (XXX.signatures.json)
-    names = find_name_of_all_spec_and_signatures_json_pairs(path)
-    if len(names) > 0:
-        return path
-
-    # No spec/signatures.json combo found in this folder,
-    # check the parent folder (unless the parent folder IS
-    # THE SAME as this folder)
-    current = Path(path)
-    parent = current.parent
-    if parent != current:
-        return find_spec_folder_with_signatures_json(f"{parent}")
-
-    # If nothing is found, return None
+    # Assume that spec/signatures.json files are only found in
+    # SPECS/XX and SPECS-EXTENDED/XX.  Find an ancestor of path
+    # that adheres to this assuption.  Return None if not found.
+    regex = f".*SPECS(-EXTENDED)?{os.sep}[^{os.sep}]+"
+    matching_path = re.search(regex, path)
+    if matching_path is not None:
+        return matching_path.group(0)
     return None
 
 def check_folder(folder):
     signatures_correct = True
 
+    # get SPECS/XX ancestor of input
     # find YY (maybe ancestor of path) that has xx/YY/YY.spec
     path = find_spec_folder_with_signatures_json(folder)
     if path is None:
--- a/toolkit/scripts/check_signatures.py
+++ b/toolkit/scripts/check_signatures.py
@@ -3,7 +3,7 @@
 # Licensed under the MIT License.
 
 from pathlib import Path
-from typing import Optional
+from typing import List, Optional
 
 import argparse
 import hashlib
@@ -39,14 +39,35 @@ def find_file_and_check(path, filename, expected_signature) -> Optional[bool]:
 
     return None
 
+def find_name_of_all_spec_and_signatures_json_pairs(path: str) -> List[str]:
+    names: List[str] = []
+    # Search for all spec files (XXX.spec)
+    for spec_path in Path(path).glob("*.spec"):
+        with open(spec_path, "r") as f:
+            name = Path(spec_path).stem
+            signature_path = os.path.join(path, f"{name}.signatures.json")
+            with open(signature_path, "r") as ff:
+                # If there is a matching signature file (XXX.signatures.json),
+                # add it to list
+                names.append(name)
+    return names
+
 def find_spec_folder_with_signatures_json(path: str) -> Optional[str]:
+    # Use this path if there are any spec files (XXX.spec) that have
+    # a matching signature file (XXX.signatures.json)
+    names = find_name_of_all_spec_and_signatures_json_pairs(path)
+    if len(names) > 0:
+        return path
+
+    # No spec/signatures.json combo found in this folder,
+    # check the parent folder (unless the parent folder IS
+    # THE SAME as this folder)
     current = Path(path)
-    if Path(os.path.join(path, f"{current.name}.spec")).is_file():
-        if Path(os.path.join(path, f"{current.name}.signatures.json")).is_file():
-            return path
     parent = current.parent
     if parent != current:
         return find_spec_folder_with_signatures_json(f"{parent}")
+
+    # If nothing is found, return None
     return None
 
 def check_folder(folder):
@@ -58,7 +79,9 @@ def check_folder(folder):
         # no spec/signature files found in path or its ancestors
         return signatures_correct
 
-    for signature_path in Path(path).glob("*.signatures.json"):
+    names = find_name_of_all_spec_and_signatures_json_pairs(path)
+    for name in names:
+        signature_path = os.path.join(path, f"{name}.signatures.json")
         with open(signature_path, "r") as f:
             signatures_json = json.load(f)
             for file_to_check, expected_signature in signatures_json["Signatures"].items():
--- a/toolkit/scripts/check_signatures.py
+++ b/toolkit/scripts/check_signatures.py
@@ -22,15 +22,11 @@ def getSignature(fileName) -> str:
             sha256sum.update(read_data)
     return sha256sum.hexdigest()
 
-def find_file(path, filename) -> Optional[str]:
-    return_value : Optional[str] = None
+def find_matching_files(path, filename) -> List[str]:
+    return_value : List[str] = []
     for matching_file in Path(path).glob(f"**/{filename}"):
         if os.path.exists(matching_file):
-            if return_value is not None:
-                print(f"ERROR: detected multiple {filename}: [{matching_files}]")
-                return None
-            
-            return_value = matching_file
+            return_value.append(str(matching_file))
 
     return return_value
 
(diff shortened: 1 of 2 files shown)


toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +21,56 @@ def getSignature(fileName) -> str:
...
+            return path
+    parent = current.parent
+    if parent != current:
+        return find_spec_folder_with_signatures_json(f"{parent}")
+    return None
+
+def check_folder(folder):
     signatures_correct = True
+
+    print(f"check {folder}")
```
bfjelds, line 55: not needed

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +21,52 @@ def getSignature(fileName) -> str:
...
+            if result is not None:
+                return result
+
+    return None
+
+def find_spec_folder_with_signatures_json(path: str) -> Optional[str]:
+    current = Path(path)
+    if Path(os.path.join(path, f"{current.name}.spec")).is_file():
+        if Path(os.path.join(path, f"{current.name}.signatures.json")).is_file():
+            return path
```
bfjelds, line 46: this assumes that for SPECS/XXX/YYY.spec and SPECS/XXX/ZZZ.signatures.json, XXX==YYY and XXX==ZZZ.  hopefully this is true, it seems to be?

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +21,52 @@ def getSignature(fileName) -> str:
...
+def check_folder(folder):
     signatures_correct = True
+
+    # find YY (maybe ancestor of path) that has xx/YY/YY.spec
+    path = find_spec_folder_with_signatures_json(folder)
+    if path is None:
+        # no spec/signature files found in path or its ancestors
+        return signatures_correct
+
     for signature_path in Path(path).glob("*.signatures.json"):
```
PawelWMS, line 61: Btw. only valid signatures file is `<spec_name>.signatures.json`. In an off chance there's more than one `.signatures.json` files, all other ones will be ignored by the toolkit anyway, so it might be worth to make the check similarly strict as well.

toolkit/scripts/check_signatures.py
PawelWMS: Not quite. There can be multiple `.spec` files inside a folder (see: `golang` for instance). The expected structure is that for each `XXX.spec` there's an optional corresponding `XXX.signatures.json`. The signatures are optional, because some specs just don't have `Source` tags, so there are no sources to sign.

bfjelds: ok ... will code something up with that in mind :)

toolkit/scripts/check_signatures.py
bfjelds: not a valid assumption ... adjusted accordingly

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +21,75 @@ def getSignature(fileName) -> str:
...
+    names: List[str] = []
+    # Search for all spec files (XXX.spec)
+    for spec_path in Path(path).glob("*.spec"):
+        with open(spec_path, "r") as f:
+            name = Path(spec_path).stem
+            signature_path = os.path.join(path, f"{name}.signatures.json")
+            with open(signature_path, "r") as ff:
+                # If there is a matching signature file (XXX.signatures.json),
+                # add it to list
+                names.append(name)
```
PawelWMS, line 52: Why do we need to open files if we're not reading anything from them?

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +21,75 @@ def getSignature(fileName) -> str:
...
+    return None
+
+def find_name_of_all_spec_and_signatures_json_pairs(path: str) -> List[str]:
+    names: List[str] = []
+    # Search for all spec files (XXX.spec)
+    for spec_path in Path(path).glob("*.spec"):
+        with open(spec_path, "r") as f:
+            name = Path(spec_path).stem
+            signature_path = os.path.join(path, f"{name}.signatures.json")
+            with open(signature_path, "r") as ff:
```
PawelWMS, line 49: There might be not signatures JSON for some spec files and `open()` will then crash.

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +21,75 @@ def getSignature(fileName) -> str:
             sha256sum.update(read_data)
     return sha256sum.hexdigest()
 
-def check_folder(path):
+def find_file_and_check(path, filename, expected_signature) -> Optional[bool]:
```
PawelWMS, line 24: We're overloading the meaning of the return value for `find_file_and_check` where true/false say if the check was OK and `None` indicates the file was missing. To make function have a single responsibility, what do you think about extracting the `Path(path_to_check).is_file()` check outside the function?

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +21,75 @@ def getSignature(fileName) -> str:
...
+        actual_signature = getSignature(path_to_check)
+        if actual_signature == expected_signature:
+            return True
+        print(f"ERROR: detected a mismatched signature for {filename}, expected [{expected_signature}] does not equal actual [{actual_signature}]")
+        return False
+
+    for content in os.listdir(path):
+        path_to_check = os.path.join(path, content)
+        if os.path.isdir(path_to_check):
+            result = find_file_and_check(path_to_check, filename, expected_signature)
```
PawelWMS, line 36: It feels like we're re-implementing a recursive file search. What do you think about using [Python modules](https://stackoverflow.com/questions/2186525/how-to-use-to-find-files-recursively), that would do it for us?

toolkit/scripts/check_signatures.py
bfjelds: just a way to make sure the file exists.  maybe there is a better way, i'm a python novice

toolkit/scripts/check_signatures.py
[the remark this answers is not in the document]
bfjelds: open doesn't crash, but i can use os.path.exists instead

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +21,75 @@ def getSignature(fileName) -> str:
...
+def find_spec_folder_with_signatures_json(path: str) -> Optional[str]:
+    # Use this path if there are any spec files (XXX.spec) that have
+    # a matching signature file (XXX.signatures.json)
+    names = find_name_of_all_spec_and_signatures_json_pairs(path)
+    if len(names) > 0:
+        return path
+
+    # No spec/signatures.json combo found in this folder,
+    # check the parent folder (unless the parent folder IS
+    # THE SAME as this folder)
```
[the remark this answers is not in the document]
bfjelds: if the PR had only changed the script file in SPECS/dracut/20overlay/script.sh, SPECS/dracut/20overlay is passed into the py code ... the code needs to look at SPECS/dracut to see the signatures file.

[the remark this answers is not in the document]
bfjelds: i wanted to minimize the assumptions, so i didn't do much filtering from the git commands the yml executed.

toolkit/scripts/check_signatures.py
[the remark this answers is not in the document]
bfjelds: changed to use os.path.exists

toolkit/scripts/check_signatures.py
bfjelds: changed to use os.path.exists

toolkit/scripts/check_signatures.py
bfjelds: this code doesn't really bother me much given how small it is.

but if needed, i could change the logic so that there was a find-file that returned an Optional[Path] and then check the signature of that path.

toolkit/scripts/check_signatures.py
bfjelds: maybe, but the time it took me to find something vs just writing this didn't feel justified.

toolkit/scripts/check_signatures.py
bfjelds: separated search from signature-check

toolkit/scripts/check_signatures.py
bfjelds: replaced this code with Path.glob.

toolkit/scripts/check_signatures.py
PawelWMS: Also, I noticed I was imprecise: there can be more `.signature.json` files when there's more `.spec` files but what I had in mind is `.signature.json` files, which don't match any `.spec` files. They may be source files for one of the `.spec` files then but I'd be very surprised to see that.

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +21,75 @@ def getSignature(fileName) -> str:
...
+def find_file_and_check(path, filename, expected_signature) -> Optional[bool]:
+    path_to_check = os.path.join(path, filename)
+    if Path(path_to_check).is_file():
+        actual_signature = getSignature(path_to_check)
+        if actual_signature == expected_signature:
+            return True
+        print(f"ERROR: detected a mismatched signature for {filename}, expected [{expected_signature}] does not equal actual [{actual_signature}]")
+        return False
+
+    for content in os.listdir(path):
```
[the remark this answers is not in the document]
PawelWMS: I think I just mistook `path` and `path_to_check`. Sorry for the noise.

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +22,65 @@ def getSignature(fileName) -> str:
...
+                # add it to list
+                names.append(name)
+
+    return names
+
+def find_spec_folder_ancestor(path: str) -> Optional[str]:
+    # Assume that spec/signatures.json files are only found in
+    # SPECS/XX and SPECS-EXTENDED/XX.  Find an ancestor of path
+    # that adheres to this assuption.  Return None if not found.
+    regex = f".*SPECS(-EXTENDED)?{os.sep}[^{os.sep}]+"
```
PawelWMS, line 55: There's also `SPECS-SIGNED`.:P

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +22,65 @@ def getSignature(fileName) -> str:
...
-                if Path(path_to_check).is_file():
-                    actual_signature = getSignature(path_to_check)
+                file_to_check_path = find_file(path, file_to_check)
+                if file_to_check_path is not None:
+                    actual_signature = getSignature(file_to_check_path)
                     if actual_signature != expected_signature:
                         print(f"ERROR: detected a mismatched signature for {file_to_check}, expected [{expected_signature}] does not equal actual [{actual_signature}]")
                         signatures_correct = False
                 else:
                     print(f"{file_to_check} is not found in CBL-Mariner, build to verify signature")
```
PawelWMS, line 82: Highly unlikely but if `find_file` finds more than one file (and returns `None` as a result of that), this message will make things confusing. I'd suggest we print file-related error messages either from here or from `find_file`, so that they are consistent.

toolkit/scripts/check_signatures.py
```diff
@@ -20,20 +22,65 @@ def getSignature(fileName) -> str:
...
+
+    return names
+
+def find_spec_folder_ancestor(path: str) -> Optional[str]:
+    # Assume that spec/signatures.json files are only found in
+    # SPECS/XX and SPECS-EXTENDED/XX.  Find an ancestor of path
+    # that adheres to this assuption.  Return None if not found.
+    regex = f".*SPECS(-EXTENDED)?{os.sep}[^{os.sep}]+"
+    matching_path = re.search(regex, path)
+    if matching_path is not None:
```
PawelWMS, line 57: Not an actual change request. You mentioned you're new to Python, so just in case you find this syntax more appealing and you weren't aware of it:
```suggestion
    if matching_path:
```

This will return `False` for `None`, empty strings, empty collections (sets, lists, dictionaries), and maybe other things that I'm not aware of.

PawelWMS (approved):

PR #8382, opened 2024-03-13
Fix bug 48763270 Mariner's go downloader clobbers prev logs, making it hard to diagnose toolchain download errors

Xiaohong-Deng:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

The for loop at line 187 of toolchain.mk is meant to write and keep the logs from all the iterations but the logs from the current iteration overwrites the previous one. For more detail please refer to work item BUG 48763270

###### Change Log  

- Every invocation of the target `toolchain_rpms_rehydrated` flushes the log file.
-  Move the for loop from the `toolchain.mk` to  the `downloader.go`, passing url list to the downloader.

**YES**

###### Associated issues  

- BUG 48763270 is an internal work item.

###### Links to CVEs  
- None

###### Test Methodology

- This affects QUICK_REBUILD_TOOLCHAIN=y in local build

### DIFF
--- a/toolkit/tools/downloader/downloader.go
+++ b/toolkit/tools/downloader/downloader.go
@@ -115,9 +115,9 @@ func main() {
 		err = downloadFile(srcUrl, localDstFile, caCerts, tlsCerts)
 		if err != nil {
 			if idx == len(*srcUrls)-1 {
-				logger.Log.Fatalf("Failed to download (%s) to (%s). Error:\n%s", srcUrl, *dstFile, err)
+				logger.Log.Fatalf("Failed to download (%s) to (%s). Error:\n%s", srcUrl, localDstFile, err)
 			} else {
-				logger.Log.Errorf("Failed to download (%s) to (%s). Error:\n%s", srcUrl, *dstFile, err)
+				logger.Log.Errorf("Failed to download (%s) to (%s). Error:\n%s", srcUrl, localDstFile, err)
 			}
 		} else {
 			break
--- a/toolkit/tools/downloader/downloader.go
+++ b/toolkit/tools/downloader/downloader.go
@@ -43,8 +43,6 @@ var (
 func main() {
 	app.Version(exe.ToolkitVersion)
 	kingpin.MustParse(app.Parse(os.Args[1:]))
-	// log-file is created in InitBestEffort and should be used for all log messages
-	// for all the URLs in srcURLs
 	logger.InitBestEffort(logFlags)
 	if *noVerbose {
 		logger.Log.SetLevel(logrus.WarnLevel)
--- a/toolkit/tools/downloader/downloader.go
+++ b/toolkit/tools/downloader/downloader.go
@@ -150,7 +150,3 @@ func downloadFile(srcUrl, dstFile string, caCerts *x509.CertPool, tlsCerts []tls
 	}
 	return
 }
-
-func getRpmUrl(url string, rpmFilename string) (result string, err error) {
-	return url + "/" + rpmFilename, nil
-}
--- a/toolkit/tools/downloader/downloader.go
+++ b/toolkit/tools/downloader/downloader.go
@@ -34,14 +34,17 @@ var (
 	tlsClientCert = app.Flag("certificate", "TLS client certificate to use when downloading files.").String()
 	tlsClientKey  = app.Flag("private-key", "TLS client key to use when downloading files.").String()
 
-	dstFile   = app.Flag("output-file", "Destination file to download to").Short('O').String()
-	prefixDir = app.Flag("directory-prefix", "Directory to download to").Short('P').String()
-	srcUrl    = app.Arg("url", "URL to download").Required().String()
+	dstFile       = app.Flag("output-file", "Destination file to download to").Short('O').String()
+	prefixDir     = app.Flag("directory-prefix", "Directory to download to").Short('P').String()
+	rpmFilename   = app.Arg("rpm-filename", "RPM to download").Required().String()
+	srcUrls       = app.Arg("url-list", "URLs to download").Required().Strings()
 )
 
 func main() {
 	app.Version(exe.ToolkitVersion)
 	kingpin.MustParse(app.Parse(os.Args[1:]))
+	// log-file is created in InitBestEffort and should be used for all log messages
+	// for all the URLs in srcURLs
 	logger.InitBestEffort(logFlags)
 	if *noVerbose {
 		logger.Log.SetLevel(logrus.WarnLevel)
@@ -75,17 +78,6 @@ func main() {
 	if *dstFile != "" && *prefixDir != "" {
 		logger.Log.Fatalf("Cannot specify both --output-file and --directory-prefix")
 	}
-	if *dstFile == "" {
-		// strip query strings from url
-		u, err := url.Parse(*srcUrl)
-		if err != nil {
-			logger.Log.Fatalf("Invalid URL (%s), Error:\n%s", *srcUrl, err)
-		}
-		*dstFile = filepath.Base(u.Path)
-		if *prefixDir != "" {
-			*dstFile = filepath.Join(*prefixDir, *dstFile)
-		}
-	}
 
 	if *noClobber {
 		exists, err := file.PathExists(*dstFile)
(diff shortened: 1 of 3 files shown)


toolkit/tools/internal/logger/log.go
```diff
@@ -86,7 +86,7 @@ func initLogFile(filePath string, color string) (err error) {
 		return
 	}
 
-	file, err := os.Create(filePath)
+	file, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
```
PawelWMS, line 89: This function is used across all of our tools and we don't want to append to all logs by default - people often run multiple builds, which end up logging to the same files, so this would get messy very soon.

Xiaohong-Deng: Would adding a new param to the function to switch between append and truncate and let truncate be the default minimize the impact?

No truncation in a single invocation of the toolkit and starting a new log file from scratch in each toolkit invocation are the goals here.

dmcilvaney:
I believe I sent this to m40 back in Jan as an alternative approach:
```diff
diff --git a/toolkit/scripts/toolchain.mk b/toolkit/scripts/toolchain.mk
index 15b6c8e7a..168bc3f63 100644
--- a/toolkit/scripts/toolchain.mk
+++ b/toolkit/scripts/toolchain.mk
@@ -191,7 +191,10 @@ $(toolchain_rpms_rehydrated): $(TOOLCHAIN_MANIFEST) $(go-downloader)
  echo "Attempting to download toolchain RPM: $$rpm_filename" | tee -a "$$log_file" && \
  mkdir -p $$rpm_dir && \
  cd $$rpm_dir && \
+  url_num=0 && \
  for url in $(PACKAGE_URL_LIST); do \
+    log_file="$(toolchain_downloads_logs_dir)/$${rpm_filename}.url$${url_num}.log" && \
+    url_num=$$(($$url_num + 1)) && \
    $(go-downloader) --no-verbose --no-clobber $$url/$$rpm_filename \
      $(if $(TLS_CERT),--certificate=$(TLS_CERT)) \
      $(if $(TLS_KEY),--private-key=$(TLS_KEY)) \
@@ -310,7 +313,10 @@ $(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(depend_REBUILD_TOOLCHAIN) $(go-downlo
  echo "Downloading toolchain RPM: $$rpm_filename" | tee -a "$$log_file" && \
  mkdir -p $$rpm_dir && \
  cd $$rpm_dir && \
+  url_num=0 && \
  for url in $(PACKAGE_URL_LIST); do \
+    log_file="$(toolchain_downloads_logs_dir)/$${rpm_filename}.url$${url_num}.log" && \
+    url_num=$$(($$url_num + 1)) && \
    $(go-downloader) --no-verbose --no-clobber $$url/$$rpm_filename \
      $(if $(TLS_CERT),--certificate=$(TLS_CERT)) \
      $(if $(TLS_KEY),--private-key=$(TLS_KEY)) \
```

Its a bit cumbersome but will avoid stomping logs while also not requiring changes to the logging library.

toolkit/tools/internal/logger/log.go
dmcilvaney: We definitely want fresh logs for each tool when it starts by default, I think append should be the special case.

PawelWMS: I agree that the append would be a special case. In general I'm fine with making append an option but in this particular scenario, I think the issue lies with us trying to combine makefile's output with logs coming from a single tool. To keep the tool isolated and simpler, I'd allow it to continue to only worry about printing its own logs into a file and leave it up to the caller, if they want to do something more fancy. For instance we could print the tool's output to different files for each iteration of the `for url in $(PACKAGE_URL_LIST); do` loop. Or we could print it to a temporary file log and append the main log with its contents for each loop iteration.

toolkit/tools/downloader/downloader.go
```diff
@@ -75,17 +78,6 @@ func main() {
 	if *dstFile != "" && *prefixDir != "" {
 		logger.Log.Fatalf("Cannot specify both --output-file and --directory-prefix")
 	}
-	if *dstFile == "" {
-		// strip query strings from url
-		u, err := url.Parse(*srcUrl)
```
Xiaohong-Deng, line 80: It seems this check of url validity only happens when dstFile is not specified. It's more like a step to extract the rpmFilename. Since the urls and rpmFilename are passed separately, do we still want to call `url.Path`? @dmcilvaney @PawelWMS

dmcilvaney: I'm fine with adding more validation of the URLs.

toolkit/tools/downloader/downloader.go
```diff
@@ -34,14 +34,17 @@ var (
 	tlsClientCert = app.Flag("certificate", "TLS client certificate to use when downloading files.").String()
 	tlsClientKey  = app.Flag("private-key", "TLS client key to use when downloading files.").String()
 
-	dstFile   = app.Flag("output-file", "Destination file to download to").Short('O').String()
-	prefixDir = app.Flag("directory-prefix", "Directory to download to").Short('P').String()
-	srcUrl    = app.Arg("url", "URL to download").Required().String()
+	dstFile     = app.Flag("output-file", "Destination file to download to").Short('O').String()
+	prefixDir   = app.Flag("directory-prefix", "Directory to download to").Short('P').String()
+	rpmFilename = app.Arg("rpm-filename", "RPM to download").Required().String()
+	srcUrls     = app.Arg("url-list", "URLs to download").Required().Strings()
```
dmcilvaney, line 40: To avoid confusion, I think we probably want this to be a flag rather than an arg (ie require `--url-list` to be passed explicilty)

dmcilvaney, line 40: This tool is mean to roughly mimic `wget`, I'd like to keep the basic behavior roughly the same.

Thoughts on having `--url-list` require fully formed URLs and drop `--rpm-filename`? That would keep the behaviour (we try each URL in the list in order, failing if they all fail).

Then we could just generate the URL list in make.

Xiaohong-Deng: Agree that it's an improvement.

Xiaohong-Deng: Move that URL generating work to make is doable. Does the downloader have the potential to be used elsewhere? I can imagine if the trailing part is completely different for the URLs in other scenarios, dropping `--rpm-filename` makes it more generic.

reubeno:
@dmcilvaney @Xiaohong-Deng @PawelWMS  -- great to see the good discussion! I see the change is currently targeting `main`. What do y'all think of retargeting this change against `3.0-dev`? That should give you breathing room to make the change the way y'all think it should look without any risk to the 2.0 LTS.

toolkit/tools/downloader/downloader.go
Xiaohong-Deng: I'm thinking maybe we can do the validation regardless whether `dstFile` is specified? @dmcilvaney

toolkit/tools/downloader/downloader.go
PawelWMS: +1 to a simple change where `--url-list` can have more than one entry. This way the tool is still a generic downloader (like @dmcilvaney said) but we have a way to ask it to download multiple files at once.

Having said that, now we have to deal with how we handle new scenarios:
1. Download failures, where we can:
    - fail on the first unsuccessful download,
    - fail when there was at least one unsuccessful download but try to download all files before exiting,
    - never fail but display a warning.
2. Downloading same-named file from separate sources: how do we save them, so they don't override each other?

I understand the current change is to address the issue of downloading one file but from a set of X potential sources. That's just not what this tool is for, however - it's meant to accept a URL and try to download the file using it. We could try now to extend the use cases it's suppose to handle but I think that'll unnecessarily complicate the implementation, so my suggestion would be to handle that on the caller's side.

PawelWMS (changes requested):
Current changes seem to make the tool less generic and more tight-fitted for just a single use case. Please see earlier comments for suggestion how to differently approach the issue.

PawelWMS:
> @dmcilvaney @Xiaohong-Deng @PawelWMS -- great to see the good discussion! I see the change is currently targeting `main`. What do y'all think of retargeting this change against `3.0-dev`? That should give you breathing room to make the change the way y'all think it should look without any risk to the 2.0 LTS.

+1. We definitely want to reflect all 2.0 changes in 3.0. Depending on how invasive or not this one is going to be, it may still meet the bar to merge it to `main` as well. Having said that, I've got nothing against making this change `3.0-dev`-only.

toolkit/tools/downloader/downloader.go
Xiaohong-Deng: The current changes reflect the original behavior from `toolchain.mk` and `downloader.go` combined, that is, mark a success if it downloads the file from a URL, skip the remaining URLs and move on to the next RPM file, or mark a failure if it is unable to download the file from any of the given URLs and move on to the next RPM file.

@dmcilvaney and I proposed handling the bug from the caller's side exclusively as one of the possible routes but I liked the idea of moving the for loop into the downloader by @dmcilvaney. IMHO with `--url-list` being dropped it stays generic and becomes compatible with more use cases at the cost of internal complexity. The positive side is it reduces the complexity of the Makefile which is harder to read and more error-prone than Golang, which is why I chose this route.

In order to reduce the complexity on the Golang side, maybe we can try to keep the original downloader as much as possible in a helper function.

kgodara912:
Closing as 2.0 is EOL now. Will reopen if required.

PR #8389, opened 2024-03-13
Smaller base images (baremetal & qemu-guest)

romoh:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

This PR attempts to make the base images smaller by:
* Move away from using the common core-packages and instead switch to a smaller subset absolutely needed to boot.
* Expand `cleanup.sh` script to clean up extra directories.
* Drop a macros file to skip installing docs. This can be overridden by consumers of the images.

**NO**

###### Test Methodology

- Local built and tested the image.


jiria (approved):

toolkit/imageconfigs/packagelists/base-image-packages.json
```diff
@@ -0,0 +1,27 @@
+{
+    "packages": [
```
dmcilvaney, line 2: Wonder if this can be in `core-packges.spec` as a `core-packages-minimal` or something like that. I don't mind it either way. I do see some differences like `initramfs` which isn't included in the core-packages-* lists...

dmcilvaney (approved):

toolkit/imageconfigs/packagelists/base-image-packages.json
romoh: I thought about that, but then we are still trying to create another smaller subset for recovery os scenarios, and this will not have some of the stuff in here (ie: cloud-init), that would be the true `minimum`. I like the idea of putting that in `core-packages.spec` 👍

PR #8398, opened 2024-03-13
Image Customizer: Fix ISO gen with SELinux disabled base image.

cwize1:
Currently, the ISO gen disables SELinux in the image because there is an issue causing the ISO not to boot if SELinux is enabled. However, the logic to disable SELinux fails if SELinux is already disabled. This change fixes this problem.

This change also combines the two different implementations for changing the SELinux mode (in the kernel CLI args). It also makes the logic more robust to the ordering of the kernel CLI args.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology


toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -138,10 +127,10 @@ func setLinuxPath(inputGrubCfgContent string, linuxPath string) (outputGrubCfgCo
 func setInitrdPath(inputGrubCfgContent string, initrdPath string) (outputGrubCfgContent string, oldInitrdPath string, err error) {
 	match := initPathRegex.FindStringSubmatchIndex(inputGrubCfgContent)
 	if match == nil {
-		return "", "", fmt.Errorf("failed to find initrd command line.")
+		return "", "", fmt.Errorf("failed to find initrd command line")
 	}
 	if len(match) <= initPathRegexPathIndexEnd {
-		return "", "", fmt.Errorf("failed to find initrd command line path.")
+		return "", "", fmt.Errorf("failed to find initrd command line path")
```
romoh, line 133: nit - use a different error so you can differentiate between this failure and the previous one

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -170,40 +159,78 @@ func appendKernelCommandLineArguments(inputGrubCfgContent string, extraCommandLi
...
+	}
 
-	oldValue = inputGrubCfgContent[start:end]
-	outputGrubCfgContent = inputGrubCfgContent[:start] + value + inputGrubCfgContent[end:]
+	outputGrubCfgContent = fmt.Sprintf("%s%s=%s%s", inputGrubCfgContent[:start], name, value, inputGrubCfgContent[end:])
 
 	return outputGrubCfgContent, oldValue, nil
 }
 
+func tryRemoveKernelCommandLineArgumentValue(inputGrubCfgContent string, name string) (string, int, error) {
```
romoh, line 217: why is this tryRemove..()? what expectations should caller of this have?

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: They are different. By 1 word.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Add comment.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -170,40 +159,90 @@ func appendKernelCommandLineArguments(inputGrubCfgContent string, extraCommandLi
...
 }
 
-func replaceKernelCommandLineArgumentValue(inputGrubCfgContent string, name string, value string) (outputGrubCfgContent string, oldValue string, err error) {
+// Tries to find the specified kernel CLI arg. Does not fail if the arg is not found.
+//
+// Returns:
+// - startIndex: The string index where the arg was found. Or -1 if not found.
+// - endIndex: The string index of the end of the arg. Or -1 if not found.
+// - value: The value of the arg. Or "" if not found.
+func findKernelCommandLineArgumentArg(grub2Config string, name string) (int, int, string, error) {
```
dmcilvaney, line 168: edge case, but what happens if the same key is used twice? Some cmdline args support that additively (crash kernel size for example), while others just take the last in the list.

cwize1: This code actually has a lot of problems, such as being unable to handle escape sequences and single/double quoted strings. So, only really aiming for good enough at the moment.

romoh (approved):

amritakohli (approved):

PR #8406, opened 2024-03-14
Update shim to 15.8

ddstreetmicrosoft:
Rework our 3.0 shim, as well as updating it to 15.8.

Upstream review opened: https://github.com/rhboot/shim-review/issues/427


ddstreetmicrosoft:
@christopherco @adithyaj

ddstreetmicrosoft:
note that I got rid of the `shim-unsigned` package as it didn't make sense to me...but if we should keep it i'm open to discussion

christopherco:
> note that I got rid of the `shim-unsigned` package as it didn't make sense to me...but if we should keep it i'm open to discussion

Supportive of removing `shim-unsigned`.

Also, I noticed you are renaming `shim-unsigned-aarch64` to `shim-unsigned-aa64`. Any concerns on keeping `shim-unsigned-aarch64`? The main benefit is that it aligns with familiar Fedora package naming.

SPECS/shim/TESTING
```diff
@@ -10,21 +10,216 @@ separate keys, and the public part of the key used to sign mm/fb
...
+$ sudo certutil -d /etc/pki/pesign -L -n KeyInShim -r > key-in-shim.der
+
+Because the shim package can only have a single embedded certificate,
+you have 2 options on how to embed the test certificate in the shim.
+
+Option a) Replace the production Azure Linux certificate
+
+This will result in a shim that contains *only* the test certificate,
+and will only be able to load other EFI binaries signed by the test
+certificate. This means the shim will not be load the
```
gjswalling, line 73: minor typo, "will not be load" should be "will not load"

ddstreetmicrosoft:
> > note that I got rid of the `shim-unsigned` package as it didn't make sense to me...but if we should keep it i'm open to discussion
> 
> Supportive of removing `shim-unsigned`.
> 
> Also, I noticed you are renaming `shim-unsigned-aarch64` to `shim-unsigned-aa64`. Any concerns on keeping `shim-unsigned-aarch64`? The main benefit is that it aligns with familiar Fedora package naming.

yeah migth as well just keep the naming the same, i adjusted to remove the rename in latest push

SPECS/shim/TESTING
ddstreetmicrosoft: ack thnx

ddstreetmicrosoft:
@christopherco @gjswalling one q for this pr; I currently have the new shim building with `EFIDIR=azurelinux`, while our Mariner 2 shim doesn't have any `EFIDIR=` set at all.

What that means is that for Mariner 2, our shim will look in the same directory (i.e. `/boot/efi/EFI/BOOT`) for its default boot loader (i.e. `grubx64.efi`), while in Azure Linux 3, our shim will look in the `EFIDIR` (i.e. `/boot/efi/EFI/azurelinux`). That then will require our grub2 package to place its EFI into `/boot/efi/EFI/azurelinux` instead of `/boot/efi/EFI/BOOT`. I don't know if there is any "right" approach (e.g. Fedora appears to use `EFIDIR=` while OpenSUSE appears to not).

I don't have a strong opinion either direction, though I'm inclined to shift over to using `EFIDIR=` like Fedora does. I'll include an update to the grub2 package in this PR to adjust the `grubx64.efi` location, but I can remove it if you think it would be better to not use `EFIDIR=`.

ddstreetmicrosoft:
> I don't have a strong opinion either direction, though I'm inclined to shift over to using `EFIDIR=` like Fedora does. I'll include an update to the grub2 package in this PR to adjust the `grubx64.efi` location, but I can remove it if you think it would be better to not use `EFIDIR=`.

Actually on second thought, I'm just gonna leave the `grub2` package unchanged (i.e. install to `/boot/efi/EFI/BOOT/grubx64.efi`) and remove the use of `EFIDIR=` from the azl3 shim package. That will keep everything inside the `/boot/efi/EFI/BOOT/` dir, as was the case with Mariner 2.

I'm open to changing it to use a `EFIDIR=` (and I think that's probably "better") but only if either of you think we should.

ddstreetmicrosoft:
> > I don't have a strong opinion either direction, though I'm inclined to shift over to using `EFIDIR=` like Fedora does. I'll include an update to the grub2 package in this PR to adjust the `grubx64.efi` location, but I can remove it if you think it would be better to not use `EFIDIR=`.
> 
> Actually on second thought, I'm just gonna leave the `grub2` package unchanged (i.e. install to `/boot/efi/EFI/BOOT/grubx64.efi`) and remove the use of `EFIDIR=` from the azl3 shim package. That will keep everything inside the `/boot/efi/EFI/BOOT/` dir, as was the case with Mariner 2.
> 
> I'm open to changing it to use a `EFIDIR=` (and I think that's probably "better") but only if either of you think we should.

hmm actually the later shim version might require `EFIDIR=`, as it doesn't appear to be booting at all without it. I'll look into this more.

ddstreetmicrosoft:
@christopherco @gjswalling @adithyaj i think this is ready for another round of reviews...still needed is:

1) wait for PMC for 3.0 to get the `pesign` package published
2) update the shim-review submission to correct the Dockerfile and add docker logs (using the PMC `pesign` instead of building from the Dockerfile)
3) work with @adithyaj to update the signing pipeline(s) to sign the mm/fb binaries from the `shim-unsigned` packages and provide them to the `shim` package
4) do secure boot test (following TESTING doc in this pr)

and then of course submit the shim-review upstream and wait for that process to complete (and get the shims signed), and add the signed shim efi binaries into the `shim` package source

ddstreetmicrosoft:
Also note that I added a 'spec entanglement' thingie so that the `shim-unsigned-x64`, `shim-unsigned-aarch64`, and `shim` packages all must be at exactly the same version-release. That *might* mean that we can't make any changes to the `shim-unsigned-*` packages without going through a new upstream shim-review, but that is probably what we actually want. Only reason I mention it is if e.g. we want to patch the `mmx64.efi` to fix something in mokmanager...that might not actually need to get the shim reviewed, if we're only changing mokmanager. it's possible the shim binary would remain binary-identical in which case there wouldn't be a problem (we could just re-use the existing signed shim binary). In any case, I think it's a low probability of it being a problem and even if it is we can change the build process later if needed, so TL;DR is I think having the specs 'entangled' is ok.

ddstreetmicrosoft:
> I added a 'spec entanglement' thingie

also just to clarify, I added the 'spec entanglement' PR check thingie (which only is effective inside a PR), but also I added a real 'entanglement' by setting the `shim` package to `BuildRequires: shim-unsigned-%{efiarch} = %{version}-%{release}`

dmcilvaney:
Just an fyi: our tools are happy to have multiple .spec files share a folder (ie llvm, clang, nodejs etc.). 

Your call on if this simplifies your life or makes it more complicated, just wanted to offer the option.

ddstreetmicrosoft:
> Just an fyi: our tools are happy to have multiple .spec files share a folder (ie llvm, clang, nodejs etc.).
> 
> Your call on if this simplifies your life or makes it more complicated, just wanted to offer the option.

thanks, I'll keep that in mind, though I think in this case it probably isn't needed.

ddstreetmicrosoft:
@adithyaj @christopherco to handle the 'must sign after building all packages' design requirement i just created a symlink so the source spec files will be identical (i.e. `SPECS-SIGNED/shim-signed/shim.spec -> ../../SPECS/shim/shim.spec`). I think this is the best signing design that still fits into the limitations of the 'build from scratch' design; do you think this will work with our signing pipeline?

ddstreetmicrosoft:
> @adithyaj @christopherco to handle the 'must sign after building all packages' design requirement i just created a symlink so the source spec files will be identical (i.e. `SPECS-SIGNED/shim-signed/shim.spec -> ../../SPECS/shim/shim.spec`). I think this is the best signing design that still fits into the limitations of the 'build from scratch' design; do you think this will work with our signing pipeline?

minor update - on thinking about it, more than just the spec file is needed so i changed the symlink to the entire dir, i.e. `SPECS-SIGNED/shim -> ../SPECS/shim`

SPECS/shim/ucs2le.bom
adithyaj, line 1: super minor change - but could we maybe rename this to ucs2le-header.bom? It's only the header and we still need the BOOT<arch>.CSV for it.

SPECS/shim-unsigned-aarch64/azurelinux-ca-20230216.der
adithyaj, line 1: Is this from the new certchain?

SPECS-SIGNED/grub2-efi-binary-signed/grub2-efi-binary-signed.spec
```diff
@@ -58,6 +61,8 @@ specifically created for installing on %{buildarch} systems
 Summary:        GRand Unified Bootloader
 Group:          Applications/System
 Requires:       grub2-tools-minimal = %{version}-%{release}
+Recommends:     shim-x64 >= 15.8-1
```
adithyaj, line 64: Curious - why the recommendation for shim-x64?

adithyaj (changes requested):

SPECS/shim/shim.spec
```diff
@@ -1,43 +1,187 @@
...
+        touch ${f}-is-signed
+    else
+        echo "Using unsigned $efi"
+        cp -vf %{shimdir}/$efi
+    fi
+done
+
+%build
+echo "%{name}" > dnf-protected.conf
+cat ucs2le.bom %{shimdir}/BOOT%{EFIARCH}.CSV > BOOT%{EFIARCH}.CSV
```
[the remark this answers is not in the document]
ddstreetmicrosoft: yes

SPECS/shim-unsigned-aarch64/azurelinux-ca-20230216.der
ddstreetmicrosoft: yes

SPECS-SIGNED/grub2-efi-binary-signed/grub2-efi-binary-signed.spec
ddstreetmicrosoft: it may make more sense if you review the commits individually, e.g. https://github.com/microsoft/azurelinux/commit/6d6b59851daaa43d6e14c8f70b70d6ef4fd30ddd

SPECS/shim/ucs2le.bom
ddstreetmicrosoft: I don't quite understand the reason for the renaming? Maybe it would be easier to just dynamically generate the file.

adithyaj: Yeah that dymanic generation actually seems a bit cleaner to me too, thanks!

adithyaj (changes requested):

SPECS/shim-unsigned-aarch64/shim-unsigned-aarch64.spec
```diff
@@ -1,65 +1,216 @@
...
-ExclusiveArch:  aarch64
+%global __debug_package 1
+%global _binaries_in_noarch_packages_terminate_build 0
+%global __debug_install_post %{SOURCE100} %{efiarch}
+%undefine _debuginfo_subpackages
+
+# currently here's what's in our dbx: nothing
+%global dbxfile %{nil}
+
+Name:		shim-unsigned-aarch64
```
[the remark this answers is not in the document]
ddstreetmicrosoft: thanks, i added a `Provides: shim-unsigned-%%{efiarch}`

SPECS/shim/shim.spec
```diff
@@ -1,43 +1,186 @@
...
+# binary from the shim-unsigned package. We do not actually check if
+# the file is signed here; that will be verified in %%check.
+for f in mm fb; do
+    efi=${f}%{efiarch}.efi
+    if [[ -s $efi ]]; then
+        echo "Using signed $efi"
+        touch ${f}-is-signed
+    else
+        echo "Using unsigned $efi"
+        cp -vf %{shimdir}/$efi
```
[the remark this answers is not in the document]
ddstreetmicrosoft: thanks, added missing .

SPECS/shim-unsigned-aarch64/shim-unsigned-aarch64.signatures.json
```diff
@@ -1,7 +1,9 @@
 {
  "Signatures": {
-  "cbl-mariner-ca-20210127.der": "34f7436bc56c7b4e526e55dbc207ad6311dbace0403e1d462ad1f9ce3479d386",
-  "sbat.csv.in": "895ecdff532c9ed1cfbcd1e4d8fd9b29b87628394cc9c9101ec5be47f66e504d",
-  "shim-15.4.tar.bz2": "8344473dd10569588b8238a4656b8fab226714eea9f5363f8c410aa8a5090297"
+  "azurelinux-ca-20230216.der": "fac39cd3bbb87d6326eb87020eb7ab1f01058d9b94d3c47d4264da05601bc89a",
+  "sbat.azurelinux.csv": "bb37f33ec9bb6be2fa8929243b613337244f007efb0761196f2f6b3caa03926c",
+  "shim-15.8.tar.bz2": "a79f0a9b89f3681ab384865b1a46ab3f79d88b11b4ca59aa040ab03fffae80a9",
+  "shim-find-debuginfo.sh": "80a43905c7ec5578794adcbf277e01a6d7eb05f2c12bd97bddc8d727066266df",
```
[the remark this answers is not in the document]
ddstreetmicrosoft: > shim-find-debuginfo.sh 's hash is invalid, needs to updated to match the signature in shim-unsigned-x64.

thanks, updated

> 
> Sidenote: I think we're running into a similar issue we were trying to avoid with the symlink where we have the same dependencies between x86_64 and aarch64

If you mean separate src packages for each shim-unsigned arch, then I agree they could be combined into one, but the difficulty then is if one arch needs to be updated to a newer version while the other arch can't, maybe due to a arch-specific bug. Fedora handles this with separate src packages for shim-unsigned and also having a 'map' of what shim-unsigned package version is used by the actual shim package. I don't know if that's the best way to handle it, but in any case I think it will require more thinking/designing to come up with the best way to arrange the packages for both/all archs, which probably is best done for the next version instead of 3.0.

ddstreetmicrosoft:
rebased

SPECS/shim/shim.spec
```diff
@@ -1,43 +1,188 @@
...
+%global shimversiondir %{shimrootdir}/%{version}
+%global shimdir %{shimversiondir}/%{efiarch}
+
 Summary:        First stage UEFI bootloader
-Name:           shim
-Version:        15.4
-Release:        2%{?dist}
+Name:           shim-%{efiarch}
+Version:        15.8
+Release:        3%{?dist}
```
trungams, line 35: Release number should be 1?

gjswalling (approved):

PR #8422, opened 2024-03-18
add patch for cloud-init pkg install error

rmhsawyer:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Mariner in 2.0 inherited some cloud-init provisioning functions from photon in [upstream](https://github.com/canonical/cloud-init/blob/f7c1c76104870b629d3b5e25618544431f7d8427/cloudinit/distros/photon.py) which includes package install function. This function logs error even on benign warning message. For example, doing tdnf upgrade with message "None or nothing to do" will be consider as an error. This causes some issues for Mariner users trying to use cloud-init package install provisioning function. Here we add a change to prevent exec_cmd returning failure if stderr is present. If something comes up in the package install process, we will still be able to catch it in the cloud-init.log.

###### Change Log  

- add a patch to prevent return error in cloud-init exec_cmd function
- adjust unit test accordingly

NO

###### Associated issues  

- [Bug 48639553](https://microsoft.visualstudio.com/OS/_workitems/edit/48639553): For Mariner 2: 2 ClusterConfiguration Package download is failing for multiple packages

###### Test Methodology

- local package build with RUN_CHECK=y
- Buddy build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=529575&view=results
- Azure VM provisioning test using package install module


cjp256:
Looks like exec_cmd() assumes return failure if stderr is present.  https://github.com/canonical/cloud-init/blob/82a5fb3588102b088c84def3a263d4c2d61fcb0f/cloudinit/distros/photon.py#L48

Maybe it's worth patching that instead?

ddstreetmicrosoft:
> Looks like exec_cmd() assumes return failure if stderr is present. https://github.com/canonical/cloud-init/blob/82a5fb3588102b088c84def3a263d4c2d61fcb0f/cloudinit/distros/photon.py#L48
> 
> Maybe it's worth patching that instead? Doesn't look like an appropriate assumption in the general case.

I agree this is a better approach.

rmhsawyer:
> > Looks like exec_cmd() assumes return failure if stderr is present. https://github.com/canonical/cloud-init/blob/82a5fb3588102b088c84def3a263d4c2d61fcb0f/cloudinit/distros/photon.py#L48
> > Maybe it's worth patching that instead? Doesn't look like an appropriate assumption in the general case.
> 
> I agree this (patching exec_cmd) is a better approach.

cjp256:
LGTM, thanks!  Perhaps that stderr check appears to have been for setting hostname? Maybe we can try to set an invalid hostname and make sure it fails properly on Mariner?

ddstreetmicrosoft (approved):
lgtm

PR #8444, opened 2024-03-18
Adding additionalDirs support to MIC API 

Adub17030MS:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Today, the MIC API only has support for file-by-file copy in the additionalFiles schema. This PR adds the **additionalDirs** schema to the MIC API, which adds support for enabling directory-by-directory copy functionality.

Shown below is an example implementation of how this schema is to be used:
additionalDirs:
  - sourcePath: "path/to/local/dir"
    destinationPath: "/path/to/dest/dir"
    newDirPermissions: _value_
    mergedDirPermissions: _value_
    childFilePermissions: _value_

sourcePath: The path in the target OS that the directory will be copied to.

destinationPath: The path in the target OS that the directory will be copied to.

_the following three parameters are optional_

newDirPermissions: The permissions to set on all of the new directories being created on the target OS (including the top-level directory). Note: If this value is not specified in the config, the permissions for these directories will be set to 0755.

mergedDirPermissions: The permissions to set on the directories being copied that already do exist on the target OS (including the top-level directory). Note: If this value is not specified in the config, the permissions for this field will be the same as that of the pre-existing directory.

childFilePermissions: The permissions to set on the children file of the directory. Note: If this value is not specified in the config, the permissions for these directories will be set to 0755.

###### Change Log  

- Implemented API changes to integrate the **additionalDirs** schemea

**NO**

###### Test Methodology

Directories being copied: 
Screenshot 2024-04-26 181603
- Setup:
image
- Output (on the customized MIC image)
image

- Setup (without permissions set):
image
- Output (without permissions set):
image
(Note, the existing /usr directory had permissions of 755 already set to it, so in this case where the merge permissions is not defined, the permissions are not overridden)


toolkit/tools/imagecustomizerapi/dirpermissions.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+	"os"
+)
+
+// The file permissions to set on the directory.
+//
+// Accepted formats:
+//
+// - Octal string (e.g. "660")
+type DirPermissions os.FileMode
```
cwize1, line 18: Why not reuse the `FilePermissions` type?

toolkit/tools/internal/file/file.go
```diff
@@ -89,6 +89,72 @@ func Copy(src, dst string) (err error) {
...
+	isDstExist, err := PathExists(dst)
+	if err != nil {
+		return err
+	}
+	if isDstExist {
+		isDstDir, err := IsDir(dst)
+		if err != nil {
+			return err
+		}
+		if isDstDir {
```
cwize1, line 104: Should you error out if the destination directory exists and isn't a directory?

toolkit/tools/internal/file/file.go
```diff
@@ -89,6 +89,72 @@ func Copy(src, dst string) (err error) {
...
+			if err := CopyDir(srcPath, dstPath, childPermissions, childPermissions); err != nil {
+				return err
+			}
+		} else {
+			// If it's a file, copy it
+			if err := Copy(srcPath, dstPath); err != nil {
+				return err
+			}
+			// Set file permissions
+			if err := os.Chmod(dstPath, childPermissions); err != nil {
```
cwize1, line 148: You should use `NewFileCopyBuilder` to copy the file, so that you don't need to reimplement setting the file permissions.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -206,6 +211,22 @@ func copyAdditionalFiles(baseConfigPath string, additionalFiles imagecustomizera
 	return nil
 }
 
+func copyAdditionalDirs(baseConfigPath string, additionalDirs imagecustomizerapi.DirConfigList, imageChroot *safechroot.Chroot) error {
+	for _, dirConfigElement := range additionalDirs {
+		absSourceDir := file.GetAbsPathWithBase(baseConfigPath, dirConfigElement.SourcePath)
+
+		logger.Log.Infof("Copying %s into %s", absSourceDir, dirConfigElement.DestinationPath)
+
+		err := imageChroot.AddDirs(absSourceDir, dirConfigElement.DestinationPath, fs.FileMode(*dirConfigElement.Permissions), fs.FileMode(*dirConfigElement.ChildPermissions))
```
cwize1, line 220: `Permissions` and `ChildPermissions` are optional fields. But you are assuming here that they are always provided.

toolkit/tools/imagecustomizerapi/dirconfig.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+)
+
+// DirConfigList is a list of destination files where the source file will be copied to in the final image.
+// This type exists to allow a custom marshaller to be attached to it.
+type DirConfigList []DirConfig
+
+type DirConfig struct {
+	// The path in the target OS that the directory will be copied to.
+	SourcePath string `yaml:"SourcePath"`
```
cwize1, line 18: Use camelCase for the YAML names.

toolkit/tools/imagecustomizerapi/dirconfig.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// The path in the target OS that the directory will be copied to.
+	DestinationPath string `yaml:"DestinationPath"`
+
+	// The permissions to set on the directory.
+	Permissions *DirPermissions `yaml:"Permissions"`
+
+	// The permissions to set on the children of the directory.
+	ChildPermissions *DirPermissions `yaml:"ChildPermissions"`
```
cwize1, line 27: Looking at the implementation, `ChildPermissions` seems to the permissions applied to all the descendent files (that are copied) of the directory. So, perhaps rename it to `DescendentFilesPermissions`? Or `CopiedFilesPermissions`?

toolkit/tools/imagecustomizerapi/dirconfig.go
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//
+
+package imagecustomizerapi
```
cwize1, line 6: You need to update the `toolkit/tools/imagecustomizer/docs/configuration.md` file.

toolkit/tools/imagecustomizerapi/dirconfig.go
```diff
@@ -0,0 +1,69 @@
...
+
+type DirConfig struct {
+	// The path in the target OS that the directory will be copied to.
+	SourcePath string `yaml:"SourcePath"`
+
+	// The path in the target OS that the directory will be copied to.
+	DestinationPath string `yaml:"DestinationPath"`
+
+	// The permissions to set on the directory.
+	Permissions *DirPermissions `yaml:"Permissions"`
```
cwize1, line 24: Looking at the implementation, `Permissions` seems to be the permissions that is given to top-level directory if it doesn't already exist. It might be good to clarify the comment.

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/dirconfig.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// The permissions to set on the children of the directory.
+	ChildPermissions *DirPermissions `yaml:"ChildPermissions"`
+}
+
+func (l *DirConfigList) IsValid() (err error) {
+	if len(*l) <= 0 {
+		return fmt.Errorf("list is empty")
```
amritakohli, line 32: is there a case where this would be less than 0?

toolkit/tools/imagecustomizerapi/dirpermissions.go
Adub17030MS: I thought to separate the types to not confuse the design since this type is being set to the directory permissions. Would you suggest to reus the `FilePermissions` type instead here?

toolkit/tools/imagecustomizerapi/dirconfig.go
Adub17030MS: I do not think so, but I chose to keep this format since it is also used in `fileconfig.go`: https://github.com/microsoft/azurelinux/blob/862c25d1ce38634a8a8b20453d8b8d6ec1ac6b42/toolkit/tools/imagecustomizerapi/fileconfig.go#L35

toolkit/tools/imagecustomizerapi/dirconfig.go
Adub17030MS: I chose `childPermissions` since it is also being applied to the subdirectories being copied over, so specifically adding files to the name of this feild might be misleading

toolkit/tools/imagecustomizerapi/dirconfig.go
Adub17030MS: done.

toolkit/tools/internal/file/file.go
Adub17030MS: implemented with `NewFileCopyBuilder`

toolkit/tools/imagecustomizerapi/dirconfig.go
Adub17030MS: Added more clarity to the comment.

toolkit/tools/imagecustomizerapi/dirpermissions.go
Adub17030MS: Will remove this and use the `FilePermissions`

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
Adub17030MS: Now they are implemeneted as optional

toolkit/tools/imagecustomizerapi/dirconfig.go
Adub17030MS: Changed name

toolkit/tools/imagecustomizerapi/dirconfig.go
Adub17030MS: I will update the readme file once all other design/code changes have been made

toolkit/tools/internal/file/file.go
```diff
@@ -89,6 +89,77 @@ func Copy(src, dst string) (err error) {
...
+				}
+			}
+		}
+	}
+
+	if !isDstExist {
+		logger.Log.Infof("Creating destination directory on chroot (%s)", dst)
+		// Create dst dir
+		if newDirPermissions != nil {
+			err = os.MkdirAll(dst, *newDirPermissions)
```
cwize1, line 118: Considering that the default permissions is a value `0755`, instead of some other behavior, you could drop the pointer on `newDirPermissions` and `childFilePermissions` and let the caller (i.e. `copyAdditionalDirs()`) handle the default behavior.

toolkit/tools/internal/file/file.go
```diff
@@ -89,6 +89,77 @@ func Copy(src, dst string) (err error) {
...
+			if err != nil {
+				return err
+			}
+		} else {
+			err = os.MkdirAll(dst, 0755)
```
cwize1, line 123: nit: `0o755`. This just makes it explicit that the value is an octal.

Similar comment elsewhere.

toolkit/tools/imagecustomizerapi/dirconfig.go
```diff
@@ -0,0 +1,77 @@
...
+	// The permissions to set on the children file of the directory.
+	// Note: If this value is not specified in the config, the permissions for these directories will be set to 0755.
+	ChildFilePermissions *FilePermissions `yaml:"childFilePermissions"`
+}
+
+func (l *DirConfigList) IsValid() (err error) {
+	for i, dirConfig := range *l {
+		err = dirConfig.IsValid()
+		if err != nil {
+			return fmt.Errorf("invalid [dirConfig] at index %d: %w", i, err)
```
cwize1, line 40: There isn't anything in the YAML called "dirConfig". So, this error might be a bit confusing to the user. So perhaps, `"invalid value at index %d:\n%w"`.

toolkit/tools/imagecustomizerapi/dirconfig.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	if d.DestinationPath == "" {
+		return fmt.Errorf("invalid [destinationPath] value: empty string")
+	}
+
+	// Permissions
+	if d.NewDirPermissions != nil {
+		err = d.NewDirPermissions.IsValid()
+		if err != nil {
+			return fmt.Errorf("invalid [newDirPermissions] value: %w", err)
```
cwize1, line 60: nit: `:\n%w`. This makes the combined error message a little easier to read.

Same comment elsewhere.

toolkit/tools/internal/file/file.go
cwize1, line 123: nit: Move `0755` into a constant, instead of repeating the value everywhere.

toolkit/tools/internal/file/file_test.go
```diff
@@ -123,3 +125,115 @@ func TestRemoveDirectoryContentsNonExistent(t *testing.T) {
 	err = RemoveDirectoryContents(tempDir)
 	assert.Error(t, err)
 }
+
+func TestCopyDir(t *testing.T) {
+	workingDir, err := os.Getwd()
+	if err != nil {
```
cwize1, line 131: nit:

```golang
if !assert.NoError(err) {
    return
}
```

Instead of using `Panicf`.

toolkit/tools/internal/file/file_test.go
```diff
@@ -123,3 +125,115 @@ func TestRemoveDirectoryContentsNonExistent(t *testing.T) {
...
+	assert.True(t, equal)
+
+	// Removing all test files and directories
+	err = os.RemoveAll(testDir)
+	assert.NoError(t, err)
+}
+
+func CreateTestFiles(filename string, outputDir string) error {
+	// Test data
+	testData := []byte{0x01, 0x02, 0x03, 0x04, 0x05}
```
cwize1, line 167: If you are so inclined, you can add the test files to the git repo. Typically, such files are placed under a directory called `testdata`.

toolkit/tools/internal/file/file_test.go
```diff
@@ -123,3 +125,115 @@ func TestRemoveDirectoryContentsNonExistent(t *testing.T) {
...
+	equal, err := AreDirectoriesEqual(src, dst)
+	assert.NoError(t, err)
+	assert.True(t, equal)
+
+	// Removing all test files and directories
+	err = os.RemoveAll(testDir)
+	assert.NoError(t, err)
+}
+
+func CreateTestFiles(filename string, outputDir string) error {
```
cwize1, line 165: nit: Rename to `createTestFiles` to make the function private. Same thing for the other helper functions.

cwize1 (changes requested):

toolkit/tools/internal/file/file_test.go
Adub17030MS: Fixed this.

toolkit/tools/internal/file/file_test.go
Adub17030MS: Fixed this.

toolkit/tools/internal/file/file.go
Adub17030MS: Added constant.

toolkit/tools/imagecustomizerapi/dirconfig.go
Adub17030MS: Added this.

toolkit/tools/imagecustomizerapi/dirconfig.go
Adub17030MS: added this

toolkit/tools/internal/file/file.go
Adub17030MS: Added this and made it into a constant.

toolkit/tools/internal/file/file.go
Adub17030MS: Is there an added benefit for doing this? I prefer the current implementation since it is clearer.

toolkit/tools/internal/file/file_test.go
Adub17030MS: I think its okay to leave the current implementation as is since this `testdata` directory will only serve a purpose for this specific test case. I like this function implementations since its cleaner as well.

toolkit/tools/internal/file/file.go
Adub17030MS: When you say drop the pointer, do you mean to change the signature of the `CopyDir()` function to be this:

`func CopyDir(src, dst string, newDirPermissions, childFilePermissions fs.FileMode, mergedDirPermissions*fs.FileMode) (err error)`

toolkit/tools/internal/file/file.go
```diff
@@ -101,6 +103,9 @@ func CopyDir(src, dst string, newDirPermissions, mergedDirPermissions, childFile
 		if err != nil {
 			return err
 		}
+		if !isDstDir {
+			return fmt.Errorf("destination exists but is not a directory (%s)", dst)
+		}
 		if isDstDir {
```
cwize1, line 109: nit: Remove the `if isDstDir {` check. It is redundant.

toolkit/tools/internal/file/file.go
cwize1: Yes.

cwize1: The benefit would be that the default would be stored within the MIC specific code. So, it less likely to be accidentally changed by non-MIC toolkit changes.

cwize1 (changes requested):

toolkit/tools/internal/file/file.go
Adub17030MS: removed.

toolkit/tools/internal/file/file.go
Adub17030MS: Added this change in and verified that the functionality described in the description still works.

toolkit/tools/imagecustomizerapi/dirconfig.go
Adub17030MS: Updated the `toolkit/tools/imagecustomizer/docs/configuration.md` file

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -409,6 +420,59 @@ os:
...
+Example:
+
+```yaml
+os:
+  additionalDirs:
+    - sourcePath: "home/files/targetDir"
+      destinationPath: "usr/project/targetDir"
+```
+
+### newDirPermissions [int]
```
cwize1, line 448: Please match the existing documentation for fileConfig.permissions.

The permission value is a string because at some point we may add support for the alternative formats that chmod supports.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -409,6 +420,59 @@ os:
...
+- Int containing an octal value. e.g. `664`
+
+Example:
+
+```yaml
+os:
+  additionalFiles:
+    - sourcePath: "home/files/targetDir"
+      destinationPath: "usr/project/targetDir"
+      newDirPermissions: 0644
```
cwize1, line 471: You don’t need the `0` prefix.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -409,6 +420,59 @@ os:
...
+```yaml
+os:
+  additionalDirs:
+    - sourcePath: "home/files/targetDir"
+      destinationPath: "usr/project/targetDir"
+```
+
+### newDirPermissions [int]
+
+The permissions to set on all of the new directories being created on the target OS (including the top-level directory). **Note:** If this value is not specified in the config, the permissions for these directories will be set to `0755`.
```
cwize1, line 450: You can simplify the second sentence to `Default value: 755`.

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/dirconfig.go
```diff
@@ -14,10 +14,10 @@ import (
 type DirConfigList []DirConfig
 
 type DirConfig struct {
-	// The path in the target OS that the directory will be copied to.
+	// The absolute path to the source directory that will be copied.
```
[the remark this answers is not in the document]
Adub17030MS: It does work with both implementations (both the absolute and relative path). So what should we change comment to here, maybe this: `The path to the source directory that will be copied (can be relative or absolute path).`

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -409,6 +420,59 @@ os:
...
+      destinationPath: "usr/project/targetDir"
+```
+
+### newDirPermissions [string]
+
+The permissions to set on all of the new directories being created on the target OS (including the top-level directory). Default value: `755`.
+
+### mergedDirPermissions [string]
+
+The permissions to set on the directories being copied that already do exist on the target OS (including the top-level directory). **Note:** If this value is not specified in the config, the permissions for this field will be the same as that of the pre-existing directory.
```
cwize1, line 454: Please line wrap at 88.

toolkit/tools/imagecustomizerapi/dirconfig.go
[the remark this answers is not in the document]
cwize1: Sure.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -206,6 +212,42 @@ func copyAdditionalFiles(baseConfigPath string, additionalFiles imagecustomizera
...
+	var (
+		newDirPermissionsValue    fs.FileMode
+		childFilePermissionsValue fs.FileMode
+	)
+	for _, dirConfigElement := range additionalDirs {
+		absSourceDir := file.GetAbsPathWithBase(baseConfigPath, dirConfigElement.SourcePath)
+		logger.Log.Infof("Copying %s into %s", absSourceDir, dirConfigElement.DestinationPath)
+
+		// Setting permissions values. They are set to a default value if they have not been specified.
+		if dirConfigElement.NewDirPermissions == nil {
```
cwize1, line 225: nit:

```golang
newDirPermissionsValue := defaultFilePermissions
if dirConfigElement.NewDirPermissions != nil {
    newDirPermissionsValue = *(*fs.FileMode)(dirConfigElement.NewDirPermissions)
}
```

2 fewer lines. (You can also remove the `var` line.)

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -409,6 +420,59 @@ os:
...
+
+### childFilePermissions [string]
+
+The permissions to set on the children file of the directory. Default value: `755`.
+
+Supported formats for permission values:
+
+- Int containing an octal value. e.g. `664`
```
cwize1, line 462: An integer can't contain an octal value. Octal is a string representation of a number.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -97,90 +99,99 @@ os:
 
 ## Schema Overview
```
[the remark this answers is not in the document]
cwize1: Please revert.

cwize1 (changes requested):

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -24,47 +24,49 @@ The Azure Linux Image Customizer is configured using a YAML (or JSON) file.
 4. Update hostname. ([hostname](#hostname-string))
 
 5. Copy additional files. ([additionalFiles](#additionalfiles-mapstring-fileconfig))
+  
+6. Copy additional directories. ([additionalDirs](#additionaldirs-dirconfig))
```
cwize1, line 28: Should we copy directories before files?

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -24,47 +24,49 @@ The Azure Linux Image Customizer is configured using a YAML (or JSON) file.
...
-19. If [--shrink-filesystems](./cli.md#shrink-filesystems) is specified, then shrink
+20. If [--shrink-filesystems](./cli.md#shrink-filesystems) is specified, then shrink
     the file systems.
 
-20. If ([verity](#verity-type)) is specified, then create the hash tree and update the
+21. If ([verity](#verity-type)) is specified, then create the hash tree and update the
     grub config.
 
-21. if the output format is set to `iso`, copy additional iso media files.
+22. if the output format is set to `iso`, copy additional iso media files.
```
cwize1, line 69: You should still add `additionalDirs` to the `Schema Overview` section.

cwize1 (approved):

PR #8452, opened 2024-03-19
filesystem: fix nobody GID

ddstreetmicrosoft:
The 'nogroup' group has differing implementations across major distros:

Ubuntu:        'nogroup' GID 65534
Fedora/RHEL:   'nobody'  GID 65534
OpenSUSE/SLES: 'nobody'  GID 65534 *and* 'nogroup' GID 65533

However, *all* distros provide the 'nobody' user with *both* UID and GID 65534.

This brings our group settings in line with OpenSUSE/SLES (providing *both* 'nogroup' and 'nobody' groups), and brings our 'nobody' user settings in line with *all* other distros (uid:gid of 65534:65534).

### DIFF
--- a/SPECS/filesystem/filesystem.spec
+++ b/SPECS/filesystem/filesystem.spec
@@ -1,7 +1,7 @@
 Summary:      Default file system
 Name:         filesystem
 Version:      1.1
-Release:      19%{?dist}
+Release:      20%{?dist}
 License:      GPLv3
 Group:        System Environment/Base
 Vendor:       Microsoft Corporation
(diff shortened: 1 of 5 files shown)


anphel31 (commented):
We also set this in the toolchain bootstrap code here; https://github.com/microsoft/azurelinux/blob/3.0-dev/toolkit/scripts/toolchain/container/toolchain_initial_chroot_setup.sh#L57
Looks to me that it's already set correctly, but can you also please check that spot?

anphel31 (approved):

christopherco (approved):

ddstreetmicrosoft:
> We also set this in the toolchain bootstrap code here; https://github.com/microsoft/azurelinux/blob/3.0-dev/toolkit/scripts/toolchain/container/toolchain_initial_chroot_setup.sh#L57 Looks to me that it's already set correctly, but can you also please check that spot?

*EEK*! we're creating basic system files in two totally separate places? Should we be doing that?

In any case yeah that looks fine, though it's slightly different (it looks like it's taking the Ubuntu approach).

PR #8469, opened 2024-03-20
cryptsetup: enable ssh-token

ddstreetmicrosoft:
Previously there were "circular dependency" problems, but `openssh` now build deps the "bootstrap" version of `systemd-rpm-macros` which avoids the "circular dependency".


reubeno (approved):
Would you mind updating the PR or commit info with some insights as to why this circular dependency is no longer an issue (i.e., what changed)?

PR #8495, opened 2024-03-21
Image customizer: kernel module API refactor

elainezhao96:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

refactor mic kernel module api

###### Change Log  

-  Add kernel module Options
- refactored base on
image
- Add unit tests for the helper functions
- Addressed comments from previous pr (closed that one as that was against 2.0, see comments in https://github.com/microsoft/azurelinux/pull/6487)

**NO**

###### Test Methodology

tested with the modules-config.yaml 
image
image


toolkit/tools/pkg/imagecustomizerlib/kernelmoduleutils.go
```diff
@@ -0,0 +1,313 @@
...
+		content := strings.Join(optionsToAppend, "\n") + "\n"
+		if _, err = file.WriteString(content); err != nil {
+			return fmt.Errorf("failed to write module options to %s: %w", moduleOptionsFilePath, err)
+		}
+	}
+
+	return nil
+}
+
+func aggregateModuleOptions(aggregatedOptions []string, moduleOptionsFilePath string, moduleName string, optionKey string, optionValue string) ([]string, error) {
```
cwize1, line 259: The behavior of this function seems very strange. Why not read the `moduleOptionsFilePath` file in once, do all the changes required, and then write out the file once?

toolkit/tools/pkg/imagecustomizerlib/kernelmoduleutils_test.go
```diff
@@ -0,0 +1,235 @@
...
+	err := chroot.Initialize("", []string{}, []*safechroot.MountPoint{}, false)
+	assert.NoError(t, err)
+	defer chroot.Close(false)
+
+	modprobePath := filepath.Join(chroot.RootDir(), "etc/modprobe.d")
+	moduleDisableFilePath := filepath.Join(modprobePath, "blacklist.conf")
+	err = os.MkdirAll(modprobePath, os.ModePerm)
+	assert.NoError(t, err)
+
+	// Prepopulate the file with a blacklisted module
```
cwize1, line 78: nit: `disabled module`

We should avoid using the word "blacklist" in our own code/comments.

toolkit/tools/pkg/imagecustomizerlib/kernelmoduleutils.go
```diff
@@ -0,0 +1,313 @@
...
+type LoadMode string
+
+const (
+	LoadModeBoot    LoadMode = "boot"
+	LoadModeAuto    LoadMode = "auto"
+	LoadModeDisable LoadMode = "disable"
+	LoadModeInherit LoadMode = "inherit"
+)
+
+func loadOrDisableModules(modules []imagecustomizerapi.Module, imageChroot *safechroot.Chroot) error {
```
cwize1, line 26: Considering that this function only uses `imageChroot.RootDir()`, it might be goodness to remove the `imageChroot` parameter and pass the rootDir as a string instead. This would make it easier to write tests for this function.

cwize1, line 26: Add UTs for this function as well?

toolkit/tools/pkg/imagecustomizerlib/kernelmoduleutils_test.go
```diff
@@ -0,0 +1,235 @@
...
+)
+
+func TestEnsureModulesLoaded(t *testing.T) {
+	if os.Geteuid() != 0 {
+		t.Skip("Test must be run as root because it uses a chroot")
+	}
+
+	// Setup environment.
+	proposedDir := filepath.Join(tmpDir, "TestEnsureModulesLoaded")
+	chroot := safechroot.NewChroot(proposedDir, false)
```
cwize1, line 23: Is `chroot` actually needed here? It seems like you are just reading/writing files.

cwize1 (changes requested):

cwize1 (changes requested):

PR #8514, opened 2024-03-22
Revert "systemd: add Provides: for the "bootstrap" package names"

dmcilvaney:
This reverts commit e2f7ca6f2f705b2447aa3e98f6ef8bf76ec6f08f.

This commit exposes a limitation in our build-cycle resolution code. Will need to fix it, but revert in the meantime.
We should not have been able to publish this package as-is, so I believe the revert of the release number is ok.

Bug: https://microsoft.visualstudio.com/OS/_workitems/edit/49601294

Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=533021&view=results

### DIFF
--- a/SPECS/systemd/systemd.spec
+++ b/SPECS/systemd/systemd.spec
@@ -50,7 +50,7 @@ Version:        255
 # determine the build information from local checkout
 Version:        %(tools/meson-vcs-tag.sh . error | sed -r 's/-([0-9])/.^\1/; s/-g/_g/')
 %endif
-Release:        10%{?dist}
+Release:        9%{?dist}
 
 # FIXME - hardcode to 'stable' for now as that's what we have in our blobstore
 %global stable 1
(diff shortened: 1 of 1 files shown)


ddstreetmicrosoft:
I won't make any comment on what the toolkit needs or how to fix it, but keep in mind that package upgrades to the entire systemd package *WILL BE BROKEN* until this revert is unreverted. The systemd rpm macros *MUST* be available when building systemd.

ddstreetmicrosoft:
> I won't make any comment on what the toolkit needs or how to fix it, but keep in mind that package upgrades to all the systemd packages _WILL BE BROKEN_ until this revert is unreverted. The systemd rpm macros _MUST_ be available when building systemd.

oh sorry I thought you were reverting #8471? The bug referenced that PR, not #8467?

Reverting #8467 is simply annoying, so that's fine (but disappointing that the toolkit can't handle it). If that's really all the toolkit needs to unbreak itself.

PR #8318, opened 2024-03-08
[2.0] dhcp/dhclient.conf: add option rfc3442-classless-static-routes in dhclient.conf

cjp256:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [ ] The toolchain has been rebuilt successfully (or no changes were made to it)
- [ ] The toolchain/worker package manifests are up-to-date
- [ ] Any updated packages successfully build (or no packages were changed)
- [ ] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [ ] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [ ] All package sources are available
- [ ] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [ ] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [ ] All source files have up-to-date hashes in the `*.signatures.json` files
- [ ] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [ ] Documentation has been updated to match any changes to the build system
- [ ] Ready to merge

---

###### Summary 

What does the PR accomplish, why was it needed?

Azure configures classless routes and cloud-init requires the configuration to be included in the lease file to behave correctly.

###### Change Log  

- Add rfc3442-classless-static-routes option to dhclient.conf

###### Does this affect the toolchain?  

No

###### Test Methodology

Build & installed RPM locally and validated cloud-init behavior.

PR #8517, opened 2024-03-22
[2.0] dhcp/dhclient.conf: add option rfc3442-classless-static-routes …

ddstreetmicrosoft:
…in dhclient.conf (#8318)

### DIFF
--- a/SPECS/dhcp/dhcp.spec
+++ b/SPECS/dhcp/dhcp.spec
@@ -1,7 +1,7 @@
 Summary:        Dynamic host configuration protocol
 Name:           dhcp
 Version:        4.4.2
-Release:        6%{?dist}
+Release:        7%{?dist}
 License:        MPLv2.0
 Url:            https://www.isc.org/dhcp/
 Source0:        ftp://ftp.isc.org/isc/dhcp/%{version}/%{name}-%{version}.tar.gz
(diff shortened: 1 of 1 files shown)


ddstreetmicrosoft:
I am not able to figure out what the PR check failure is unhappy about, sorry.

SPECS/dhcp/dhcp.spec
```diff
@@ -72,6 +72,8 @@ cat > %{buildroot}/etc/dhcp/dhclient.conf << "EOF"
 #
 # Basic dhclient.conf(5)
 
+option rfc3442-classless-static-routes code 121 = array of unsigned integer 8;
```
PawelWMS, line 75: Can you add more details in the PR's summary about why we want to make this change through fast-track? We limit the changes on this branch to critical changes, which cannot wait for a regular monthly release (fixes for criticial/high CVEs, severely broken configuration, other critical security issues, etc.). I'm not an expert on DHCP configuration and couldn't tell, what this change is exactly about and how critical it may or may not be.

PawelWMS:
> I am not able to figure out what the PR check failure is unhappy about, sorry.

The pipelines failed on the `Verify all tests passed` step. If you look at the failing tests, you'll see that `libguestfs` (one of the packages depending on `dhcp` during its build) failed its package test.

The `libguestfs` is one of the packages, where its tests are still failing consistently, so we can skip over that check for the time being. Personally, I'm not a fan but we're still working on getting 100% of tests to always pass and I doubt this failure has anything to do with the change from this PR.

PawelWMS:
Had a chat with Christopher Patterson to understand what we're doing here a bit better. Fortunately, it's a low impact issue at the moment but we definitely want this fix in. The full release will contain updated images. Merging to fast-track as well, since it's already here and it's a very-risk change.

PR #8539, opened 2024-03-25
[DO NOT  MERGE] Testing new signature check pr gate

dmcilvaney:


.github/workflows/check-source-signatures.yml
```diff
@@ -0,0 +1,111 @@
...
+    steps:
+      # Checkout the branch of our repo that triggered this action
+      - name: Workflow trigger checkout
+        uses: actions/checkout@v4
+
+      # For consistency, we use the same major/minor version of Python that CBL-Mariner ships
+      - name: Setup Python 3.9
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
```
anphel31, line 26: 3.0 has python 3.12 now; 2.0 is on 3.9. Not sure that it matters here though

dmcilvaney: Sure, I can bump those while I'm at it. Just copy/paste from the other actions

.github/workflows/check-source-signatures.yml
```diff
@@ -0,0 +1,111 @@
...
+          if [ -n "${{ env.changed_pkgs_extended }}" ]; then
+            # We want to ignore errors here, as we want to check all the packages that have been modified. Capture the error code and check it later.
+            set +e
+            echo "Running: sudo make -C toolkit input-srpms REBUILD_TOOLS=y DAILY_BUILD_ID=${{ env.LKG_BUILD_ID }} SRPM_PACK_LIST='${{ env.changed_pkgs_extended }}' SPECS_DIR=../SPECS-EXTENDED"
+            sudo make -C toolkit -j$(nproc) input-srpms REBUILD_TOOLS=y DAILY_BUILD_ID=${{ env.LKG_BUILD_ID }} SRPM_PACK_LIST="${{ env.changed_pkgs_extended }}" SPECS_DIR=../SPECS-EXTENDED
+            extended_err=$?
+            set -e
+          fi
+
+          # Print results
```
anphel31, line 99: print a result in the successful case as well?

.github/workflows/check-source-signatures.yml
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+name: Source Signature Check
+
+on:
+  push:
+    branches: [3.0*]
```
anphel31, line 8: we can probably enable for all branches?

.github/workflows/check-source-signatures.yml
```diff
@@ -0,0 +1,111 @@
...
+          fi
+
+          # Print results
+          if [ $core_err -ne 0 ] || [ $extended_err -ne 0 ]; then
+            printf "\n\n******************************"
+            echo "Failed to check the signatures of the modified packages."
+            echo "Check the logs above for details on the mismatches files and their expected hashes."
+            if [ $core_err -ne 0 ]; then
+              echo "Consider running: sudo make -C toolkit input-srpms REBUILD_TOOLS=y SRPM_PACK_LIST='${{ env.changed_pkgs }}'"
+            elif [ $extended_err -ne 0 ]; then
```
anphel31, line 104: this should be a separate if statement, in case both SPECS and SPECS-EXTENDED have issues

.github/workflows/check-source-signatures.yml
dmcilvaney: It relies on the LKG, once we normalize the way to do repos I agree we should turn on.

.github/workflows/check-source-signatures.yml
dmcilvaney: good point.

.github/workflows/check-source-signatures.yml
```diff
@@ -0,0 +1,116 @@
...
+          # Print results
+          if [ $core_err -ne 0 ] || [ $extended_err -ne 0 ]; then
+            printf "\n\n******************************"
+            echo "Failed to check the signatures of the modified packages."
+            echo "Check the logs above for details on the mismatches files and their expected hashes."
+            if [ $core_err -ne 0 ]; then
+              echo "Consider running: sudo make -C toolkit input-srpms REBUILD_TOOLS=y SRPM_PACK_LIST='${{ env.changed_pkgs }}'"
+            fi
+            if [ $extended_err -ne 0 ]; then
+              echo "Consider running: sudo make -C toolkit input-srpms REBUILD_TOOLS=y SRPM_PACK_LIST='${{ env.changed_pkgs_extended }} SPECS_DIR=../SPECS-EXTENDED'"
```
anphel31, line 110: missing single quote after the SRPM_PACK_LIST=

anphel31 (approved):

PR #8597, opened 2024-03-28
Bugfix on Overlay Dracut module resilient without verity.

liulanze:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Add if check statement to make Overlay module resilient when there is no verity configured, it still can be worked properly. Also, adjust root mount logic to avoid `block:` prefix, which makes the mount works properly.

**NO**

###### Associated issues  

- Internal bug tracking: https://dev.azure.com/mariner-org/ECF/_sprints/taskboard/Mariner%20Operator%20Edge/ECF/2403?workitem=7251

###### Test Methodology

- Pipeline build id: buddybuild:


amritakohli (approved):

SPECS/dracut/20overlayfs/overlayfs-mount.sh
```diff
@@ -136,4 +137,9 @@ mount_overlayfs() {
 }
 
 parse_kernel_cmdline_args
-mount_overlayfs
+
+if [ -n "${overlayfs}" ]; then
```
romoh, line 141: You may want to add a comment explaining the logic here.
Looks like parse_kernel_cmdline_args should never fail and you rely on checking the variable to check if parsing was successful or not.

SPECS/dracut/20overlayfs/overlayfs-mount.sh
```diff
@@ -102,7 +102,8 @@ mount_overlayfs() {
     else
         echo "Mounting regular root"
         mkdir -p "${VERITY_MOUNT}"
-        mount -o ro,defaults "$root" "${VERITY_MOUNT}" || \
+        root_device="${root#block:}"  # Remove 'block:' prefix if present.
```
romoh, line 105: curious, what's scenario where block will be present?

liulanze: When `getarg` consume `$root=` from the kernel args, it will return something for example `block:/dev/disk/by-partuuid/301574d7-db66-4f13-9492-6208859a1b27`.

SPECS/dracut/20overlayfs/overlayfs-mount.sh
liulanze: Fixed, thanks.

SPECS/dracut/20overlayfs/overlayfs-mount.sh
dmcilvaney: I'm not sure if this is still the standard way to handle root paths, but consider https://github.com/microsoft/azurelinux/blob/592875ba994cf7fc55449f95bfcb3d26f618a50d/SPECS/verity-read-only-root/20verity-mount/verity-parse.sh#L111

liulanze: Fixed, please review again, thanks.

liulanze:
Re-run buddybuild with the latest revision: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=541586&view=results

dmcilvaney (approved):

PR #8617, opened 2024-03-29
add named args to check-manifest.sh

neha170:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

check-manifests.sh: replace positional args with named, populate values by default

###### Change Log  

- check-manifests.sh: replace positional args with named
- check-manifests.sh: populate values by default
- this will make it easier for user to call check-manifests.sh script directly, instead of relying on the make target
- call the script directly in check-manifests.yml
- we can remove the now defunct make target, making our makefile leaner

**NO**

###### Test Methodology

- Locally: successfully invoked check-manifests.sh directly
- Locally: successfully invoked modified check-manifests make target
- Pipeline: pipeline successfully passed when no toolchain specs were modified, and successfully caught error when toolchain spec was modified, without supporting change in manifest file

### DIFF
--- a/.github/workflows/check-manifests.yml
+++ b/.github/workflows/check-manifests.yml
@@ -25,12 +25,20 @@ jobs:
 
     - name: Check x86_64 manifests
       run: |
+        echo ##########
+        echo "Ensure toolchain and pkggen manifests (./toolkit/resources/manifests/package/*) match the versions in the .spec files"
+        echo "Run './scripts/toolchain/check_manifests.sh -a "x86_64"' to validate locally"
+        echo ##########
         pushd toolkit
         ./scripts/toolchain/check_manifests.sh -a "x86_64"
         popd
 
     - name: Check aarch64 manifests
       run: |
+        echo ##########
+        echo "Ensure toolchain and pkggen manifests (./toolkit/resources/manifests/package/*) match the versions in the .spec files"
+        echo "Run './scripts/toolchain/check_manifests.sh -a "aarch64"' to validate locally"
+        echo ##########
         pushd toolkit
         ./scripts/toolchain/check_manifests.sh -a "aarch64"
         popd
--- a/.github/workflows/check-manifests.yml
+++ b/.github/workflows/check-manifests.yml
@@ -26,12 +26,12 @@ jobs:
     - name: Check x86_64 manifests
       run: |
         pushd toolkit
-        make check-x86_64-manifests
+        ./scripts/toolchain/check_manifests.sh -a "x86_64"
         popd
 
     - name: Check aarch64 manifests
       run: |
         pushd toolkit
-        make check-aarch64-manifests
+        ./scripts/toolchain/check_manifests.sh -a "aarch64"
         popd
       if: always()
--- a/SPECS/sed/sed.spec
+++ b/SPECS/sed/sed.spec
@@ -1,7 +1,7 @@
 Summary:        Stream editor
 Name:           sed
 Version:        4.9
-Release:        2%{?dist}
+Release:        1%{?dist}
 License:        GPLv3
 URL:            https://www.gnu.org/software/sed
 Group:          Applications/Editors
--- a/SPECS/sed/sed.spec
+++ b/SPECS/sed/sed.spec
@@ -1,7 +1,7 @@
 Summary:        Stream editor
 Name:           sed
 Version:        4.9
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv3
 URL:            https://www.gnu.org/software/sed
 Group:          Applications/Editors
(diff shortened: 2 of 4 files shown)


SPECS/sed/sed.spec
```diff
@@ -1,7 +1,7 @@
 Summary:        Stream editor
 Name:           sed
 Version:        4.9
-Release:        1%{?dist}
+Release:        2%{?dist}
```
neha170, line 4: This is a test change to see if check-manifests pipeline (PR check) is working correctly

.github/workflows/check-manifests.yml
```diff
@@ -26,12 +26,12 @@ jobs:
     - name: Check x86_64 manifests
       run: |
         pushd toolkit
-        make check-x86_64-manifests
+        ./scripts/toolchain/check_manifests.sh -a "x86_64"
```
dmcilvaney, line 29: I feel like we want to have the pipelines use the tools the same way a user does. Is there a reason we want to change the users' behavior here?

dmcilvaney (approved):

.github/workflows/check-manifests.yml
```diff
@@ -26,12 +26,12 @@ jobs:
...
         popd
 
     - name: Check aarch64 manifests
       run: |
         pushd toolkit
-        make check-aarch64-manifests
+        ./scripts/toolchain/check_manifests.sh -a "aarch64"
         popd
       if: always()
```
dmcilvaney, line 37: What I have been trying to do is add a corrective action for the user into the pipeline.
Something like:
```bash
echo ##########
echo "Ensure toolchain and pkggen manifests (./toolkit/resources/manifests/package/*) match the versions in the .spec files"
echo "Run 'make check-manifests' to validate locally"
echo ##########
```

toolkit/scripts/toolchain/check_manifests.sh
```diff
@@ -4,12 +4,52 @@
...
+DISTRO_MACRO=$(make --no-print-directory -s printvar-DIST_VERSION_MACRO  2> /dev/null)
+
+if [ "$#" -eq 0 ]
+then
+    echo -e "ERROR: Missing required arg\n\n"
+    help >&2
+    exit 1
+fi
+
+while (( "$#")); do
```
cwize1, line 33: nit: Use `getopts` instead  of processing it manually.

toolkit/scripts/toolchain/check_manifests.sh
```diff
@@ -4,12 +4,52 @@
...
+    echo "[OPTIONAL] -t -> space-separated list of toolchain specs"
+    echo "[OPTIONAL] -s -> SPECS_DIR"
+    echo "[OPTIONAL] -m -> TOOLCHAIN_MANIFESTS_DIR"
+    echo "[OPTIONAL] -d -> DIST_TAG"
+    echo "[OPTIONAL] -i -> DISTRO_MACRO"
+    echo "[OPTIONAL] -h -> print this help dialogue and exit"
+}
+
+# assign default value
+TOOLCHAIN_SPEC_LIST=$(make --no-print-directory -s printvar-toolchain_spec_list  2> /dev/null)
```
cwize1, line 20: So, the `Makefile` is dependent on this script and this script is dependent on the `Makefile`? I can't say I'm particularly a fan of interdependencies like this, even if it does "work".

Would it make sense to add variants of `check-aarch64-manifests` and `check-x86_64-manifests` that don't include dependency lists?

toolkit/scripts/toolchain.mk
```diff
@@ -97,23 +97,9 @@ copy-toolchain-rpms:
 # check that the manifest files only contain RPMs that could have been generated from toolchain specs.
 check-manifests: check-x86_64-manifests check-aarch64-manifests
 check-aarch64-manifests: $(toolchain_files)
-	cd $(SCRIPTS_DIR)/toolchain && \
-		./check_manifests.sh \
-			"$(toolchain_spec_list)" \
-			"$(SPECS_DIR)" \
```
cwize1, line 103: What if any of these values are overridden by the user? I think `SPECS_DIR` is commonly overridden value by product teams.

neha170: It has the optional parameter "-s" for overriding default value of SPECS_DIR

cwize1: Yes, but the `Makefile` isn't passing it down.

.github/workflows/check-manifests.yml
neha170: I think it's a great step in empowering the user to call scripts/ tools directly without having to invoke make. Would help us redesign/ get rid of make in the future more easily.

PR #8631, opened 2024-04-01
cmake: fix package test

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

cmake: fix package test

###### Change Log  

- Change fix cmake ptest

**YES**

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=541180&view=results

### DIFF
--- a/SPECS/cmake/cmake.spec
+++ b/SPECS/cmake/cmake.spec
@@ -2,7 +2,7 @@
 Summary:        Cmake
 Name:           cmake
 Version:        3.28.2
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        BSD AND LGPLv2+
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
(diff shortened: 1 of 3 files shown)


SPECS/cmake/cmake.spec
```diff
@@ -61,6 +64,7 @@ sed -i -e "s|@@CMAKE_VERSION@@|%{version}|" -e "s|@@CMAKE_MAJOR_VERSION@@|%{majo
 # Should be removed once the issue is fixed upstream and we apply the fix: https://gitlab.kitware.com/cmake/cmake/-/issues/22470.
 rm -f %{_lib64dir}/lib{stdc++,gfortran}.a
 
+export JAVA_HOME=/usr/lib/jvm/msopenjdk-17
```
PawelWMS, line 67: Would this work:
```suggestion
export JAVA_HOME="%{java_home}"
```

I think this was a way to avoid hard-coding a specific version but I'm still trying to find the place where this macro is defined.

PawelWMS: Ah, it's inside `javapackages-tools`, which should be available -it's pulled in by `BuildRequires: javapackages-generators`.

PawelWMS: I see `javapackages-tools` is still on `msopenjdk-11`, so this may not work after all. Unless we also update that package.

anphel31: good idea. it does look like it works, and javapackages-tools is already updated for version 17

PR #8648, opened 2024-04-01
MIC ISO: support for grub-noprefix package use scenario.

gmileka:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When building an Azure Linux image, a user may choose to use the grub-noprefix package. This package installs a grub bootloader with the built-in `prefix` grub env variable set to an empty string. This is typically used for PXE booting where the bootloader and its configuration are placed in the same folder.

The grub-noprefix package installs grubx64-noprefix.efi - but MIC iso generation and the shim (bootx64.efi) both expect a grubx64.efi file. This causes MIC to fail.

This change updates MIC iso generation to:
- look for grubx64-noprefix.efi and understand that it's the same as grubx64.efi (in terms of purpose). 
- place grub.cfg file be placed under <iso root>/EFI/BOOT if the noprefix flow is detected. This is basically the folder the bootloader will look for grub.cfg in with this configuration.

###### Change Log  

- [MIC ISO: support for grub-noprefix package use scenario](https://github.com/microsoft/azurelinux/pull/8648/commits/160705f3468429cece29039184539a0d80652dc7)

**NO**

###### Associated issues  

- [7112 Image Customizer: ISO gen fails if grub2-efi-binary-noprefix package is used](https://dev.azure.com/mariner-org/ECF/_workitems/edit/7112)

###### Links to CVEs  
- None

###### Test Methodology

- Create LiveOS ISO using the following two configuration:
  - (1) baremetal 2.0 image + no package changes.
  - (2) baremetal 2.0 image - grub2-efi-binary + grub2-efi-binary-noprefix


toolkit/tools/imagecustomizer/docs/iso.md
```diff
@@ -2,8 +2,9 @@
 
 ## Overview
 
-Given a full disk image, the Azure Linux Image Customizer can generate a
-LiveOS ISO image when the `--output-image-format` is set to `iso`.
+Given a full disk image generated by the Azure Linux toolkit, the Azure Linux
```
cwize1, line 5: We should avoid mentioning the "toolkit" in the docs. It is probably better to say something like: official Azure Linux images.

toolkit/tools/imagecustomizer/docs/iso.md
```diff
@@ -98,4 +104,20 @@ os:
...
+## Input Image Layout Assumptions
+
+### Full Disk Image
+
+The input full disk image must satisfy the following requirements in order for
+MIC to be able to generate an iso image out of it:
+- File layout (after all partitions have been mounted):
+  - `/boot/grub2/grub.cfg` must exist and is the 'main' grub configuration (not
+    a redirection grub configuration file for example).
+  - `bootx64.efi` and `grubx64.efi` (or `grubx64-noprefix.efi`) must be under
```
cwize1, line 118: As an aside, in ARM64 these files are called `bootaa64.efi` and `grubaa64.efi`.

toolkit/tools/imagecustomizer/docs/iso.md
```diff
@@ -98,4 +104,20 @@ os:
...
+    the `/boot` folder (exact location is irrelevant).
```
cwize1, line 119: What does "exact location is irrelevant" mean? Are you saying that the files just have to be under "/boot" or any of its subdirectories?

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -277,26 +304,48 @@ func (b *LiveOSIsoBuilder) updateGrubCfg(grubCfgFileName string, extraCommandLin
...
+
+	usingGrubNoPrefix := containsGrubNoPrefix(bootFolderFilePaths)
+
+	for _, sourcePath := range bootFolderFilePaths {
 
 		for _, exclusion := range exclusions {
 			match := exclusion.FindStringIndex(sourcePath)
 			if match != nil {
-				return nil
+				logger.Log.Debugf("Skipping %s", sourcePath)
```
cwize1, line 347: The log message seems a little too vague.

toolkit/tools/imagecustomizer/docs/iso.md
gmileka: Does MIC support images that are built by the user using the Azure Linux toolkit though? Those are not official Azure Linux image... but are using official tooling. I am concerned that if I change it to 'Official Azure Linux images', it would mean that we do not support images created by the official tooling (unless that's really the case).

toolkit/tools/imagecustomizer/docs/iso.md
gmileka: Does MIC currently support ARM64?

gmileka: I've updated the docs. I've also opened a work item to update mic iso to look for the right files when the target is amd64.

toolkit/tools/imagecustomizer/docs/iso.md
gmileka: fixed.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: Fixed.

toolkit/tools/imagecustomizer/docs/iso.md
cwize1: ARM64 is something we should support. I think someone tested it and it mostly works.

toolkit/tools/imagecustomizer/docs/iso.md
```diff
@@ -115,8 +115,10 @@ MIC to be able to generate an iso image out of it:
 - File layout (after all partitions have been mounted):
   - `/boot/grub2/grub.cfg` must exist and is the 'main' grub configuration (not
     a redirection grub configuration file for example).
-  - `bootx64.efi` and `grubx64.efi` (or `grubx64-noprefix.efi`) must be under
-    the `/boot` folder (exact location is irrelevant).
+  - The bootloader and the shim must exist under the `/boot` folder or any of
+    its subdirectories.
+    - For x64, `bootx64.efi` and `grubx64.efi` (or `grubx64-noprefix.efi`).
+    - For ARM64, `bootaa64.efi` and `grubaa64.efi` (or `grubx64-noprefix.efi`).
```
cwize1, line 121: `grubaa64-noprefix.efi`

toolkit/tools/imagecustomizer/docs/iso.md
cwize1: It is probably possible to create images using the toolkit that will confuse MIC. So, it is probably best to not say we explicitly support all images created by the toolkit. (That being said, we don't need to explicitly say we don't support custom images either.)

cwize1 (approved):

toolkit/tools/imagecustomizer/docs/iso.md
gmileka: will fix.

PR #8654, opened 2024-04-02
opentofu: add new package opentofu v1.6.2

trungams:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

- Add new package opentofu v1.6.2

###### Change Log  

- Add new package opentofu v1.6.2

**NO**

###### Test Methodology

- Local build
- Buddy build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=542460&view=results


amritakohli (approved):

SPECS/opentofu/opentofu.spec
```diff
@@ -0,0 +1,87 @@
...
+Release:        3%{?dist}
+Summary:        OpenTofu lets you declaratively manage your cloud infrastructure
+Vendor:         Microsoft Corporation
+Distribution:   Azure Linux
+
+License:        0BSD AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND MIT AND MPL-2.0
+URL:            https://github.com/opentofu/opentofu
+Source0:        https://github.com/opentofu/opentofu/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
+
+# Below is a manually created tarball, no download link.
```
mfrw, line 14: We have started a convention of adding a shell script that generates the vendor tarball automatically.
I think we might as well add that. 
E.G look at moby-engine/docker-cli etc etc

mfrw (commented):
Maybe we should add a shell script to generate vendor tarball

SPECS/opentofu/opentofu.spec
PawelWMS: +1, our automation relies on the `generate_source_tarball.sh` scripts with specific input arguments. There should be plenty of example to use for reference and they all have a very similar structure, so a copy-paste with minor tweaks should do the trick.

SPECS/opentofu/opentofu.spec
```diff
@@ -0,0 +1,87 @@
...
+BuildRequires:  golang
+
+%description
+OpenTofu lets you declaratively manage your cloud infrastructure.
+
+%prep
+%autosetup -p1 %{?with_vendor:-b1}
+
+%build
+tar -xf %{SOURCE1} --no-same-owner
```
PawelWMS, line 43: Can we move the untarring into the `%prep` stage where the first source is extracted? You might be able to make use of the `%autosetup (...) -a 1` flag there as well.

SPECS/opentofu/opentofu.spec
```diff
@@ -0,0 +1,87 @@
...
+
+export LDFLAGS="-X github.com/opentofu/opentofu/version.dev=no"
```
PawelWMS, line 45: Why do we need to overwrite our default `LDFLAGS`?  And what's the point of setting this part here and then `-X github.com/opentofu/opentofu/version=$VERSION` separately below?

SPECS/opentofu/opentofu.spec
```diff
@@ -0,0 +1,87 @@
...
+install -m 0755 -vp ./bin/*             %{buildroot}%{_bindir}/
+
+%check
+for test in "TestResourceProvider_ApplyCustomWorkingDirectory" \
+            "TestGet_cancel" \
+            "TestInit_cancelModules" \
+            "TestFileExists" \
+; do
+awk -i inplace '/^func.*'"$test"'\(/ { print; print "\tt.Skip(\"disabled failing test\")"; next}1' $(grep -rl $test)
+done
```
PawelWMS, line 68: Why are we disabling these 4 tests? Please add a comment of explanation, so we know what's wrong with them.

SPECS/opentofu/opentofu.spec
trungams: I added a `generate_source_tarball.sh` for this package.

SPECS/opentofu/opentofu.spec
trungams: updated

SPECS/opentofu/opentofu.spec
trungams: Good point. I was mostly following fedora's spec after expanding all their macros. I cleaned up the %build section now.

SPECS/opentofu/opentofu.spec
trungams: The first 3 come from the original fedora spec file. After checking again I found out they work fine, so they could have failed due to some differences in fedora's build environment. Only the last one, `TestFileExists`, fails for us now. Added a comment in the spec file for that.

Camelron (approved):

PR #8741, opened 2024-04-09
Add configurability in systemd to control default value of UseDomains parameter

henryli001:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

- Adds parameters to /etc/systemd/networkd.conf to allow controlling the default value of the UseDomain= parameter (for both dhcp v4 and v6). 
- Backported from this [change](https://github.com/systemd/systemd/pull/32194) in the upstream

###### Change Log  

- Add patch to allow configurability of "UseDomains=" for networkd

NO

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=550114&view=results


ddstreetmicrosoft:
technically looks ok, but these kind of user-facing parameters are dangerous to add into stable releases *before* upstreaming; we do not want to add and release a config param that our users start using, and then have upstream decide they want to do it differently, leaving us to clean up our own mess of having a parameter that (at best) is confusing to end users and adds maintenance burden or (at worst) directly conflicts with how upstream decides to implement it.

So with the caveat that this needs to be upstreamed *first*, it LGTM.

henryli001:
> technically looks ok, but these kind of user-facing parameters are dangerous to add into stable releases _before_ upstreaming; we do not want to add and release a config param that our users start using, and then have upstream decide they want to do it differently, leaving us to clean up our own mess of having a parameter that (at best) is confusing to end users and adds maintenance burden or (at worst) directly conflicts with how upstream decides to implement it.
> 
> So with the caveat that this needs to be upstreamed _first_, it LGTM.

Makes sense to me. Will start working on the upstream change first and post the link here.

henryli001:
> > technically looks ok, but these kind of user-facing parameters are dangerous to add into stable releases _before_ upstreaming; we do not want to add and release a config param that our users start using, and then have upstream decide they want to do it differently, leaving us to clean up our own mess of having a parameter that (at best) is confusing to end users and adds maintenance burden or (at worst) directly conflicts with how upstream decides to implement it.
> > So with the caveat that this needs to be upstreamed _first_, it LGTM.
> 
> Makes sense to me. Will start working on the upstream change first and post the link here.

Upstream changes have been merged in: https://github.com/systemd/systemd/pull/32194. Updated the patch here to align with the upstream changes.

henryli001:
> LGTM but please patch azl3 also (preferably first)

3.0 patching complete. Here's the PR link: https://github.com/microsoft/azurelinux/pull/8799

dallasd1 (approved):

hbeberman (approved):

PR #8776, opened 2024-04-11
Add configurable docs and locales

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

We want to be able to disable documentation and locale files to minimize storage space. `rpm` offers two macros to help with this:
- `%_excludedocs 1`: Disable any `%doc` file in a package
- `%_install_langs en:es:fr`: Only include locale files that start with one of the prefixes. By setting it to `NONE` this matches no locales and effectively disables them.

These can be controlled by new settings in `SystemConfig`:
``` json
"DisableRpmDocs": true,
"DisableRpmLocales": true,
"OverrideRpmLocales": "en:fr:es"
```

`DisableRpmLocales` and `OverrideRpmLocales` are mutually exclusive.

For example, setting `DisableRpmLocales` will create `/usr/lib/rpm/macros.d/macros.installercustomizations_customize_locales`

``` rpm
# This macro file was dynamically generated by the Azure Linux Toolkit image generator
# based on the configuration used at image creation time.

# This stops locale files from being installed. %%_install_langs acts as a filter for locales
# which start with the provides strings. Setting it to an invalid value (ie 'NONE') will
# prevent any locale files from being installed.
# To enable locale files, remove this file, or comment out '%%_install_langs <LOCALE STRING>'
# Any packages which are already installed must be reinstalled for this change to take effect.

%_install_langs NONE
```

###### Change Log  

- Add `DisableDocumentationDefines()`, `DisableLocaleDefines()`, `OverrideLocaleDefines()` to `rpm.go` to create these macros.
- Add `customizationmacros` package which can generate macro files
  - `AddMacroFile()` will create a new macro file inside the given root directory that contains each of the macros given.
  - `AddCustomizationMacros()` sets the new flags defined in `SystemConfig` along with some helpful comments.

**NO**

###### Test Methodology

- Local builds of core-efi.json


toolkit/imageconfigs/core-efi.json
```diff
@@ -49,10 +49,15 @@
                 "packagelists/core-packages-image.json",
                 "packagelists/cloud-init-packages.json"
             ],
+            "Packages": [
+                "wget"
+            ],
```
PawelWMS, line 54: Is this intended or just a test leftover?

dmcilvaney: ah, good catch. Was just a leftover

dmcilvaney, line 54:
```suggestion
```

toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -41,6 +41,8 @@ type SystemConfig struct {
 	PreserveTdnfCache    bool                      `json:"PreserveTdnfCache"`
 	ReadOnlyVerityRoot   ReadOnlyVerityRoot        `json:"ReadOnlyVerityRoot"`
 	EnableHidepid        bool                      `json:"EnableHidepid"`
+	DisableDocs          bool                      `json:"DisableDocs"`
+	DisableLocales       bool                      `json:"DisableLocales"`
```
PawelWMS, line 45: Since we're already working on locales/languages: what do you think about making this option reflect RPM's configuration and accept a list of languages instead of boolean (internally it would need to be converted to a colon-separated string)? This way we're not unnecessarily limiting ourselves and the users.

toolkit/tools/internal/customizationmacros/customizationmacros.go
```diff
@@ -0,0 +1,112 @@
...
+	"sort"
+	"strings"
+
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/file"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/rpm"
+)
+
+const (
+	// macro files to customize the installer and final images
```
PawelWMS, line 19: I think we shouldn't mention the installer or final image here. This is a helper package, which shouldn't really need to have any understand the installer vs final image concepts.

toolkit/tools/imager/imager.go
```diff
@@ -590,6 +591,15 @@ func buildImage(mountPointMap, mountPointToFsTypeMap, mountPointToMountArgsMap,
 	}
 	timestamp.StopEvent(nil) // install chroot packages
 
+	// Configure setup macros for the setup environment. We run 'rpm' from outside the install chroot since it starts
+	// empty. So the macros must be defined here before we install packages.
+	logger.Log.Infof("Adding setup environment customization macros if needed")
```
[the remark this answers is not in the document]
dmcilvaney: I swapped these around while I'm doing some cleanup on the comments

toolkit/tools/internal/customizationmacros/customizationmacros.go
dmcilvaney: See above, reworkign comments

toolkit/tools/imager/imager.go
```diff
@@ -590,6 +591,15 @@ func buildImage(mountPointMap, mountPointToFsTypeMap, mountPointToMountArgsMap,
 	}
 	timestamp.StopEvent(nil) // install chroot packages
 
+	// Configure setup macros for the setup environment. We run 'rpm' from outside the install chroot since it starts
```
dmcilvaney, line 594:
```suggestion
	// Configure rpm install macros for the setup environment. We run 'rpm' from outside the install chroot since it starts
```

toolkit/tools/imagegen/configuration/systemconfig.go
dmcilvaney: You can now just set an override as a string, I figure its up to the user to write the string.

toolkit/tools/imager/imager.go
```diff
@@ -590,6 +591,15 @@ func buildImage(mountPointMap, mountPointToFsTypeMap, mountPointToMountArgsMap,
 	}
 	timestamp.StopEvent(nil) // install chroot packages
 
+	// Configure rpm install macros for the setup environment. We run 'rpm' from outside the install chroot since it starts
+	// empty. So the macros must be defined here before we install packages.
```
romoh, line 595: Does that mean the resulted image won't have the macros?

dmcilvaney: I call this twice, so both the setup environment and the final image should have them. I initially had a second set of flags for the final image but I think this is simpler.

toolkit/tools/internal/customizationmacros/customizationmacros_test.go
```diff
@@ -0,0 +1,288 @@
...
+			tempDir := t.TempDir()
+			err := AddCustomizationMacros(tempDir, tc.disableDocs, tc.DisableRpmLocales, tc.OverrideRpmLocales)
+
+			if tc.expectError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+
+				// If neither are enabled, ensure no file
+				if !tc.disableDocs && !tc.DisableRpmLocales && tc.OverrideRpmLocales == "" {
```
cwize1, line 250: In general, a test helper shouldn't reimplement the product code. While this isn't a full reimplementation, it does seem like a partial reimplementation. So, it might be worth considering how to make the test case data more generic.

For example, perhaps the test cast data can contain a list of files that are expected to exist under the `/usr/lib/rpm/macros.d` directory and a (non-exhaustive) list of lines expected to be in each file.

dmcilvaney: I ran into the same with the other tests as well, ended up with the list of expected results. I can do the same here.

toolkit/tools/internal/customizationmacros/customizationmacros.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+	}
+	return nil
+}
+
+// AddMacroFile adds a macro file to the specified root directory with the specified macros. The macro file
+// is created in the default rpm macros directory. The macro file will include a default header, with additional comments
+// if desired. Each extra comment should start with a '#' character.
+func AddMacroFile(rootDir string, macros map[string]string, macroFileName string, extraComments []string) error {
+	const macrosPath = "/usr/lib/rpm/macros.d"
```
PawelWMS, line 96: To consider: it would be safer to retrieve this value from RPM itself in case it ever changes or we're building some unusual image. Unlikely, so not a high priority in my opinion.

How to get this from RPM:
```bash
rpm --eval %_rpmmacrodir
```

toolkit/tools/internal/customizationmacros/customizationmacros.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/file"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/rpm"
+)
+
+const (
+	// macro files to customize the installer and final images
+	DisableDocsMacroFile         = "macros.installercustomizations_disable_docs"
+	ConfigureRpmLocalesMacroFile = "macros.installercustomizations_customize_locales"
```
PawelWMS, line 25: Do we need these to be global (and public)? Seems like we only need the constants in one function.

toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -188,6 +191,11 @@ func (s *SystemConfig) IsValid() (err error) {
 
 	//Validate Encryption
 
+	// Validate locales
+	if len(s.OverrideRpmLocales) > 0 && s.DisableRpmLocales {
+		return fmt.Errorf("invalid [OverrideRpmLocales]: cannot be set when [DisableRpmLocales] is true")
+	}
```
PawelWMS, line 197: What do you think of merging these two configs into one? That way we wouldn't need to add verification logic like the one here and in `customizationmacros.go`.

We replace them with a config like `InstalledRpmLocales` (or `InstalledRPMLocales`, since I'm nit-picky;)) or `InstalledPackagesLocales`. In the docs, in addition to your comment that this works like the RPM `_install_langs` macro, we'd then have to only mention that setting this value to "NONE" would only install the default locale.

Btw., would the value of "en" have the same effect as "NONE"? If so, we wouldn't really need to add any additional comments. Unless we wanted to clarify, how to limit the locales as much as possible.

toolkit/tools/internal/customizationmacros/customizationmacros.go
```diff
@@ -0,0 +1,133 @@
...
+
+var (
+	// Universal header comments for all customization macro files
+	customizationMacroHeaderComments = []string{
+		"# This macro file was dynamically generated by the Azure Linux Toolkit image generator",
+		"# based on the configuration used at image creation time.",
+		"",
+	}
+	docComments []string = []string{
+		"# This stops documentation files (anything fine in a package which is %%doc) from being installed.",
```
PawelWMS, line 36: Not sure I understood the "anything fine in a package which is %%doc" part. If we're saying that `_excludedocs` removes files marked with `%doc`, then that's also not the full picture - the macro makes RPM skip all files it considers documentation (that includes files under a bunch of hard-coded paths).

toolkit/tools/internal/customizationmacros/customizationmacros.go
```diff
@@ -0,0 +1,133 @@
...
+	header := customizationMacroHeaderComments
+	if len(extraComments) > 0 {
+		extraComments = append(extraComments, "")
+		header = append(customizationMacroHeaderComments, extraComments...)
+	}
+
+	err := validateComments(extraComments)
+	if err != nil {
+		return fmt.Errorf("failed to validate extra comments while adding macro file:\n%w", err)
+	}
```
PawelWMS, line 111: I suggest we drop `validateComments` and just add "#" at the beginning of each line ourselves here. It'll be easier to write the comments and harder to make mistakes.

jslobodzian (approved):

dmcilvaney:
Need to wait on another PR first to fix a bunch of packages. Several packages include license files as `%doc` which should really be `%license`. We need to fix those before we ship images that disable documentation.

toolkit/tools/internal/rpm/rpm.go
```diff
@@ -283,6 +286,33 @@ func DefaultDistroDefines(runChecks bool, distTag string) map[string]string {
 	return defines
 }
 
+// DisableBuildRequiresDefines sets the macro to disable documentation files when installing RPMs.
+// - defines: optional map of defines to update. If nil, a new map will be created.
+func DisableDocumentationDefines(defines map[string]string) map[string]string {
```
PawelWMS, line 291: Setting the same result through both (optionally) modifying the input and through a separate return variable may be a bit confusing. It looks like we're trying to combine two responsibilities (updating an existing map and creating a one with only the disabled docs) in one function. Also, this forces us to always set an input variable even if we don't need to update anything.

What do you think about simplifying these new functions to not take in any input maps? We don't seem to be making any updates to the defines anywhere now. If needed, we could instead create a more generic `MergeMaps` or `UpdateMaps` function, so that both these new functions are clearer with a single responsibility, and we have a tool for combining arbitrary maps instead of only adding pre-defined values to and existing one.

toolkit/tools/internal/customizationmacros/customizationmacros.go
dmcilvaney: Make this path more general, no need to limit to root.

toolkit/tools/imagegen/configuration/systemconfig.go
dmcilvaney: I don't like the idea of tying "disabled" to a user-selectible string, if we realize there is some issue with the implementation we would need to reach out to customers to get them to fix it. If we have a simple `bool` we can hide the implementation details from them.

toolkit/tools/internal/customizationmacros/customizationmacros.go
cwize1: Feel free to make them private. If MIC needs them, then it isn't too difficult to make them public again.

toolkit/tools/internal/customizationmacros/customizationmacros.go
dmcilvaney: Ok, updated to preferentially use rpm --eval. It will fall back to the default value if rpm isn't there (useful for unit tests).

toolkit/tools/internal/customizationmacros/customizationmacros.go
dmcilvaney: Private it is, I'll leave them as a const up top since that makes more sense to me.

toolkit/tools/internal/customizationmacros/customizationmacros.go
dmcilvaney: I'll just say "# This stops anything rpm considers a documentation files from being installed."

toolkit/tools/internal/customizationmacros/customizationmacros.go
```diff
@@ -0,0 +1,133 @@
...
+	docComments []string = []string{
+		"# This stops documentation files (anything fine in a package which is %%doc) from being installed.",
+		"# To enable documentation files, remove this file, or comment out '%%_excludedocs 1'",
+		"# Any packages which are already installed must be reinstalled for this change to take effect.",
+	}
+	localeComments []string = []string{
+		"# This stops locale files from being installed. %%_install_langs acts as a filter for locales",
+		"# which start with the provides strings. Setting it to an invalid value (ie 'NONE') will",
+		"# prevent any locale files from being installed.",
```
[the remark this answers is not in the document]
dmcilvaney: "en" isn't actually a default, there can be "en*" locale files. The default is whatever is baked into the binaries. They way the filter works is: For each of "<loc1>:<loc2>:<loc3>..." set, it will include any locale path that looks like: `<loc1>*/`, `<loc2>*/`, `<loc3>*/`. So since no locale will start with "NONE", no locale paths make it past the filter.

toolkit/tools/internal/customizationmacros/customizationmacros.go
dmcilvaney: Updated that function to instead ensure the # is present.

toolkit/tools/internal/rpm/rpm.go
dmcilvaney: done

toolkit/tools/internal/rpm/rpm.go
```diff
@@ -168,6 +172,29 @@ func SetMacroDir(newMacroDir string) (origenv []string, err error) {
...
 
+// Queries rpm for the current macro directory via --eval %_rpmmacrodir
+func GetMacroDir() (macroDir string, err error) {
+	const (
+		macro         = "%_rpmmacrodir"
+		defaultRpmDir = "/usr/lib/rpm/macros.d"
+	)
+
+	// This should continue to work even if the rpm command is not available (ie unit tests).
+	_, err = exec.LookPath(rpmProgram)
```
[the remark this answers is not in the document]
dmcilvaney: I added a private helper function that takes that param that can be called from UT.

toolkit/tools/internal/customizationmacros/customizationmacros.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+	return nil
+}
+
+// formatComments ensures that the comments are valid for a macro file: ie they are empty or start with '#'
+func formatComments(comments []string) (formattedComments []string) {
+	for _, comment := range comments {
+		strippedComment := strings.TrimSpace(comment)
+		if !strings.HasPrefix(strippedComment, "#") && comment != "" {
```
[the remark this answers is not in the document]
dmcilvaney: Got that feedback elsewhere as well, just felt friendly but I'll simplify. I supose there really isn't any need since no one else will really be using this.

toolkit/tools/internal/customizationmacros/customizationmacros_test.go
```diff
@@ -0,0 +1,403 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/file"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
+	"github.com/stretchr/testify/assert"
+)
+
+var expectedHeader = []string{
```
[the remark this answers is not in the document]
dmcilvaney: This is the "formatted" version, I'd have to re-calculate it dynamically to add the '#' on each line.

toolkit/tools/internal/customizationmacros/customizationmacros.go
```diff
@@ -87,11 +87,11 @@ func AddCustomizationMacros(rootDir string, disableDocs, disableRpmLocales bool,
 // formatComments ensures that the comments are valid for a macro file: ie they are empty or start with '#'
 func formatComments(comments []string) (formattedComments []string) {
 	for _, comment := range comments {
-		strippedComment := strings.TrimSpace(comment)
-		if !strings.HasPrefix(strippedComment, "#") && comment != "" {
-			comment = "# " + comment
+		if strings.TrimSpace(comment) == "" {
+			formattedComments = append(formattedComments, "")
+		} else {
+			formattedComments = append(formattedComments, strings.TrimSpace("# "+comment))
```
cwize1, line 93: Should this be `"# " + strings.TrimSpace(comment)`?

cwize1 (approved):

dmcilvaney:
Need more testing for locales, will leave them enabled in configs for now.

toolkit/tools/imagegen/configuration/systemconfig.go
PawelWMS: I'm still not convinced on this one:
- With both the override and the boolean we make it easier for the users to make a mistake. And they won't learn about it until later into the build, when this check is triggered.
- It's easy to add the boolean switch, if we learn users have problems with just the override. It's hard to remove a feature once it's published. Or to put it differently: preemptively adding feature we think the users may need can run us into a corner without any clear benefits.

dmcilvaney: Allright, I'll switch.

I still much prefer the explicit boolean though (I feel like adding a half-implemented feature which is unfriendly to use can be just as bad)

Camelron (approved):

toolkit/docs/formats/imageconfig.md
```diff
@@ -194,6 +194,29 @@ A sample PackageLists entry pointing to three files containing package lists:
...
+
+#### Custom Locales
+
+A specific locale string may also be set using:
+
+``` json
+"OverrideRpmLocales": "en:fr:es"
+```
+
+This may be any value compatible with the `%_install_langs` rpm macro.
```
PawelWMS, line 218: Could you also add a link to the RPM docs explaining how this configuration works, so it's easier for people to learn and understand what are their options and how to use this config?

toolkit/docs/formats/imageconfig.md
```diff
@@ -194,6 +194,29 @@ A sample PackageLists entry pointing to three files containing package lists:
...
+For size constrained images it may be desirable to omit documentation and non-default locales from an image.
+
+``` json
+"DisableRpmDocs": true,
+"OverrideRpmLocales": "NONE",
+```
+
+`DisableRpmDocs` will configure the image to never install the documentation. Selecting `OverrideRpmLocales=NONE` will omit all locales from the image.
+
+These customizations are encoded into `/usr/lib/rpm/macros.d/macros.installercustomizations_*` files on the final system.
```
PawelWMS, line 208: I'd also exclusively mention that we're using RPM's `%_excludedocs ` macro to disable the installation of docs on the final system. Similar to how you mention `%_install_langs` for the locales below.

PawelWMS (approved):

toolkit/docs/formats/imageconfig.md
dmcilvaney: Best I can find are some tangentially related fedora planning docs and a few mailing list posts. No official documents that I could find unfortunately.

toolkit/docs/formats/imageconfig.md
dmcilvaney:
```suggestion
These customizations are encoded into `/usr/lib/rpm/macros.d/macros.installercustomizations_*` files on the final system. They set values for `%_excludedocs` and `%_install_langs` respectively.
```

PR #8779, opened 2024-04-12
Fix PTest for python-platformdirs

0xba1a:
* Add new package python-iniconfig to SPECS-EXTENDED
 * Update toolkit to include extended repos for build with check section enabled

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Fix PTest for python-platformdirs

###### Change Log  

 - Add new package python-iniconfig to SPECS-EXTENDED
 - Update toolkit to include extended repos for build with check section enabled

**NO**

###### Test Methodology

- Pipeline build id: local
Buiddybuild will not work as it can't consider both SPECS and SPECS-EXTENDED at the same time and also the upstream extended repo is yet to be published.


toolkit/Makefile
```diff
@@ -141,6 +141,10 @@ include $(SCRIPTS_DIR)/utils.mk
 
 ######## REMAINING BUILD FLAGS ########
 
+ifeq ($(RUN_CHECK),y)
+	override PACKAGE_URL_LIST   += https://packages.microsoft.com/azurelinux/$(RELEASE_MAJOR_ID)/prod/extended/$(build_arch)
```
dmcilvaney, line 145: Was this actually needed? `PACKAGE_URL_LIST` in theory should only be needed for generating a toolchain. I would think adding to the REPO_LIST would be better.

dmcilvaney:
Thoughts on making extended available to check builds, I've got two ideas:
1) This needs to be something done through the caching code, and very selectively. We are moving towards doing our package builds and checks in a single pass (see FastTrack) since it is much faster. This means that any change we make to the build has the possibility of affecting "real" package builds.

The way I could see us doing this is to have a second cache dir for testing only stuff, where we don't care as much. The fetcher would need to support that.

We would then need to add a new mount point, and a new repo source, for the builders.

2) When running a test, I believe the workers have network. We could add a fallback mechanism only for test nodes, where if we can't find a cached copy we try to use tdnf to get it during the build. This would add a lot of complexity to the graph/scheduler though, since those packages are indistinguishable from a normal unresolved remote node which would normally be considered a build failure.

SPECS-EXTENDED/python-iniconfig/python-iniconfig.spec
```diff
@@ -0,0 +1,133 @@
...
+BuildRequires:      python3-devel
+BuildRequires:      pyproject-rpm-macros
+BuildRequires:      python3-pip
+BuildRequires:      python3-setuptools
+BuildRequires:      python3-setuptools_scm >= 8.0.3-1
+BuildRequires:      python3-six
+BuildRequires:      python3-wheel
+
+# pytest 6+ needs this and this uses pytest for tests
+%bcond_without tests
```
PawelWMS, line 23: Please use the `with_check` macro instead - that's what Azure Linux uses. `tests` is local to the spec and we don't really use the `--with-X` logic.

SPECS-EXTENDED/python-iniconfig/python-iniconfig.spec
```diff
@@ -0,0 +1,133 @@
...
+
+%install
+%pyproject_install
+%pyproject_save_files iniconfig
+
+
+%if %{with tests}
+%check
+%pytest -v
+%endif
```
PawelWMS, line 74: Redundant - we control running the `%check` section outside the spec.
```suggestion
%check
%pytest -v
```

SPECS-EXTENDED/python-iniconfig/python-iniconfig.spec
```diff
@@ -0,0 +1,133 @@
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
```
PawelWMS, line 2: Please move the vendor and distribution to the other tags below.

SPECS/python-platformdirs/python-platformdirs.spec
```diff
@@ -27,6 +26,12 @@ BuildRequires:  python3-pluggy
 BuildRequires:  python3-tomli
 BuildRequires:  python3-trove-classifiers
 BuildArch:      noarch
+%if %{with tests}
```
PawelWMS, line 29:
```suggestion
%if 0%{?with_check}
```

SPECS/python-platformdirs/python-platformdirs.spec
```diff
@@ -1,5 +1,4 @@
 # Disable tests as it requires new package python-exceptiongroup
-%global with_check 0
 %global srcname platformdirs
 %bcond_without tests
```
PawelWMS, line 3: Please remove as well - that's not how running tests is controlled in our build infrastructure and this will just override the global settings.

toolkit/Makefile
PawelWMS: Could you explain in more detail what is the goal of this change? Since it doesn't influence regular package builds, I'm not sure what we're trying to do.

0xba1a:
Abandoning as it requires significant changes in toolkit which will be taken separately.

PR #8785, opened 2024-04-12
[mic][iso] Added the ability to customize an existing mic-generated iso.

gmileka:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR allows MIC to take a mic-generated LiveOS iso and customize it.
Customization spans both the squashfs os image as well as the iso file system.

###### Change Log  

- [[mic][iso] Added the ability to customie an existing mic-generated iso.](https://github.com/microsoft/azurelinux/commit/973a1a09a3e2cabc3a1120170d5644b7f64391ae)

**NO**

###### Associated issues  

- [10 - [Existing LiveOS ISO] Add support for customizing the ISO file system contents](https://dev.azure.com/mariner-org/ECF/_workitems/edit/6782)
- [12 - [Existing LiveOS ISO] Add support for customizing LiveOS root file system (squashfs)](https://dev.azure.com/mariner-org/ECF/_workitems/edit/6785)

###### Links to CVEs  
- none

###### Test Methodology

- Created the following flavors and ensured they all boot successfully and we can log-in with mic/ci credentials:
  - vhdx->vhdx
    - os customizations
  - vhdx->iso
    - os with ci customizations + iso customizations {additional files + kernel params }
    - os customizations + iso with ci customizations {additional files + kernel params }
      - iso {from previous step} -> iso
        - iso customization
        - os customization
        - both os and iso customization


toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+	duStdout, _, err := shell.Execute("du", duParams...)
+	if err != nil {
+		return 0, fmt.Errorf("failed to find the size of the specified folder (%s):\n%w", rootDir, err)
+	}
+
+	// parse and get count and unit
+	diskSizeRegex := regexp.MustCompile(`^(\d+)([KMGT])?.*`)
+	matches := diskSizeRegex.FindStringSubmatch(duStdout)
+	if matches == nil || len(matches) < 3 {
+		return 0, fmt.Errorf("disk size (%s) has incorrect format. Expecting: ^[0-9]+[KMGT]?\t.* (e.g. 100M, 1G)", duStdout)
```
cwize1, line 1138: You should make this error message more specific to what this code is doing. Perhaps something along the lines of "unrecognized 'du' output format".

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+//
+//   - returns the size in mega bytes.
+func getDiskSizeEstimateInMBs(rootDir string, safetyFactor uint64) (size uint64, err error) {
+
+	sizeInBytes, err := getSizeOnDiskInBytes(rootDir)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get folder size on disk while estimating total disk size:\n%w", err)
+	}
+
+	sizeInMBs := sizeInBytes/1024/1024 + 1
```
cwize1, line 1198: Use `diskutils.MiB`.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+	defer func() {
+		unmountParams := []string{squashMountDir}
+		cleanupErr := shell.ExecuteLive(false, "umount", unmountParams...)
+		if cleanupErr != nil {
+			err = fmt.Errorf("%w:\nfailed to clean-up (%s): %w", err, squashMountDir, cleanupErr)
+		}
+	}()
+
+	// estimate the new disk size
+	safeDiskSizeMB, err := getDiskSizeEstimateInMBs(squashMountDir, 2 /*safety factor*/)
```
cwize1, line 1246: Might be goodness to put the safety factor in a constant at the top of the file.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+//
+//   - 'rootDir':
+//     root folder to calculate its size.
+//   - 'safetyFactor':
+//     a multiplier used with the total number of bytes calculated.
+//
+// outputs:
+//
+//   - returns the size in mega bytes.
+func getDiskSizeEstimateInMBs(rootDir string, safetyFactor uint64) (size uint64, err error) {
```
cwize1, line 1191: Should safety factor be a floating point?

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
cwize1, line 1246: Why is the safety factor so large? Being off by an order of magnitude suggests that something is wrong.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+func extractIsoImageContents(buildDir string, isoImageFile string, isoExpansionFolder string) (err error) {
+
+	mountDir, err := ioutil.TempDir(buildDir, "tmp-iso-mount-")
+	if err != nil {
+		return fmt.Errorf("failed to create temporary mount folder for iso:\n%w", err)
+	}
+	defer os.RemoveAll(mountDir)
+
+	mountParams := []string{isoImageFile, mountDir}
+	err = shell.ExecuteLive(false, "mount", mountParams...)
```
cwize1, line 925: Use `safemount.Mount` type. Unmounting safely and reliably is actually pretty difficult.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -43,6 +43,37 @@ var (
 	ToolVersion = ""
 )
 
+type ImageCustomizer struct {
```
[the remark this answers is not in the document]
gmileka: re. #1. The refactoring of the code into three stages (convert to writeable, customize, convert to final image) is related to this change as it makes the code more readable now we have the iso scenarios intermingled with customizations. The core of mic customization code has not actually changed at all (diff looks bad because of the parameters) - it has its own function right now. The parts that changed a bit is the 'convert to writeable' and the 'convert to final image'.

re. #2. I debated between the two approaches, Chris. But having flat functions here makes it look very unmanageable - especially with the increase in the number of parameters. I leaned towards centralizing the core and derived parameters in one place so that it's more consistent as well as easier to pass to consumers.
I agree that there's a chance of having side effects, maybe I should just make this structure a read-only parameter and pass it into the various functions. All its state is setup when it is created and should not change after that (also, intentionally - since I agree re. no side effects).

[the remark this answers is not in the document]
cwize1: If you want to create a parameters struct, I'm ok with that.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+//   - 'isoExpansionFolder'
+//     folder where the extracts contents will be copied to.
+//
+// outputs:
+//
+//   - creates a local folder with the same structure and contents as the provided
+//     iso image.
+func extractIsoImageContents(buildDir string, isoImageFile string, isoExpansionFolder string) (err error) {
+
+	mountDir, err := ioutil.TempDir(buildDir, "tmp-iso-mount-")
```
cwize1, line 918: Use `os.MkdirTemp`

cwize1: Same below.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+	// estimate the new disk size
+	safeDiskSizeMB, err := getDiskSizeEstimateInMBs(squashMountDir, 2 /*safety factor*/)
+	if err != nil {
+		return fmt.Errorf("failed to calculate the disk size of %s:\n%w", squashMountDir, err)
+	}
+
+	logger.Log.Debugf("safeDiskSizeMB = %d", safeDiskSizeMB)
+
+	// define a disk layout with a boot partition and a rootfs partition
+	var maxDiskSizeMB imagecustomizerapi.DiskSize
```
cwize1, line 1254: nit: `maxDiskSizeMB := imagecustomizerapi.DiskSize(safeDiskSizeMB * 1024 * 1024)`

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+	maxDiskSizeMB = imagecustomizerapi.DiskSize(safeDiskSizeMB * 1024 * 1024)
```
cwize1, line 1255: Use `diskutils.MiB`.

cwize1, line 1255: Add the ESP partition to the disk's max size?

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+
+	// mount squash fs
+	squashMountDir, err := ioutil.TempDir(buildDir, "tmp-squashfs-mount-")
+	if err != nil {
+		return fmt.Errorf("failed to create temporary mount folder for squashfs:\n%w", err)
+	}
+	defer os.RemoveAll(squashMountDir)
+
+	squashMountParams := []string{b.artifacts.squashfsImagePath, squashMountDir}
+	err = shell.ExecuteLive(false, "mount", squashMountParams...)
```
cwize1, line 1233: Use `safemount` package.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+//     root folder to calculate its size.
+//
+// outputs:
+//
+//   - returns the size in bytes.
+func getSizeOnDiskInBytes(rootDir string) (size uint64, err error) {
+	logger.Log.Debugf("Estimating disk size for (%s)", rootDir)
+
+	duParams := []string{"-sh", rootDir}
+	duStdout, _, err := shell.Execute("du", duParams...)
```
cwize1, line 1129: nit: `shell.Execute("du", "-sh", rootDir)`

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+	num, err := strconv.ParseUint(numString, 0, 10)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse disk size (%s):\n%w", numString, err)
+	}
+	logger.Log.Debugf("du returned=%d %s", num, unit)
+
+	// convert to bytes
+	multiplier := uint64(1)
+	switch unit {
+	case "K":
```
cwize1, line 1152: If you drop the `-h` to `du`, you won't need to parse the unit.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+	// parse and get count and unit
+	diskSizeRegex := regexp.MustCompile(`^(\d+)([KMGT])?.*`)
+	matches := diskSizeRegex.FindStringSubmatch(duStdout)
+	if matches == nil || len(matches) < 3 {
+		return 0, fmt.Errorf("disk size (%s) has incorrect format. Expecting: ^[0-9]+[KMGT]?\t.* (e.g. 100M, 1G)", duStdout)
+	}
+
+	numString := matches[1]
+	unit := matches[2]
+	num, err := strconv.ParseUint(numString, 0, 10)
```
cwize1, line 1143: I think you meant `strconv.ParseUint(numString, 10, 0)`.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -864,3 +894,419 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+	logger.Log.Debugf("Estimating disk size for (%s)", rootDir)
+
+	duParams := []string{"-sh", rootDir}
+	duStdout, _, err := shell.Execute("du", duParams...)
+	if err != nil {
+		return 0, fmt.Errorf("failed to find the size of the specified folder (%s):\n%w", rootDir, err)
+	}
+
+	// parse and get count and unit
+	diskSizeRegex := regexp.MustCompile(`^(\d+)([KMGT])?.*`)
```
cwize1, line 1135: You don't need the `.*` at the end of the regex. It is implied if you don't have `$`.

cwize1, line 1135: Add `\s+` after number?

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,124 +102,252 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
+
+		// If the input is a LiveOS iso and there are OS customizations are
+		// defined, we create a writeable disk image so that mic can modify
+		// it. If no OS customizations are defined, we can skip this step and
+		// just re-use the existing squashfs.
+		if ic.customizeOSPartitions {
+			err := ic.isoBuilder.createWriteableImageFromSquashfs(ic.buildDir, ic.rawImageFile)
+			if err != nil {
+				return fmt.Errorf("failed to create writeable image:\n%w", err)
+			}
```
cwize1, line 259: Should `ic.isoBuilder` be deleted here? When `customizeOSPartitions` is `true`, `ic.isoBuilder` isn't used after this point.

toolkit/tools/imagecustomizer/docs/iso.md
```diff
@@ -2,20 +2,22 @@
...
+(vhd/vhdx/qcow2/raw) or another LiveOS iso image. The LiveOS iso output format
+is specified by setting the `--output-image-format` command line parameter to
+`iso`.
+
+The LiveOS iso image is a bootable image that boots into a rootfs image
+included on the iso media without the need to have anything pre-installed
+on the target machine. The rootfs is kept read-only and an overlay in memory
+is created so that processes can successfully write state.
+
+Customizations are made to the input rootfs as usual using MIC, and will
```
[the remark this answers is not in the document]
gmileka: pending.

toolkit/tools/imagecustomizer/docs/iso.md
```diff
@@ -118,8 +120,13 @@ MIC to be able to generate an iso image out of it:
...
+    - For ARM64, `bootaa64.efi` and `grubaa64.efi` (or `grubaa64-noprefix.efi`).
   - All grub configurations and related files must be stored under the `/boot`
     folder. For example, grub.cfg cannot reference files outside that folder.
     If it does, those referenced files will not be copied to the iso and may
     cause grub to fail booting the desired operating system.
+
+### LiveOS ISO Image
+
+The input LiveOS iso image must be previously generated by The Azure Linux Image
+Customizer.
```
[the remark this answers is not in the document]
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -45,7 +45,7 @@ func doCustomizations(buildDir string, baseConfigPath string, config *imagecusto
 		return err
 	}
 
-	err = addRemoveAndUpdatePackages(buildDir, baseConfigPath, &config.OS, imageChroot, rpmsSources,
+	err = addRemoveAndUpdatePackages(buildDir, baseConfigPath, config.OS, imageChroot, rpmsSources,
```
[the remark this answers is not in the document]
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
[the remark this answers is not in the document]
gmileka: sadly, no read-only marking in golang. Passed it as a parameters struct.
pending.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,124 +102,252 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
+		return nil, err
+	}
+
+	// input image
+	ic.inputImageFile = inputImageFile
+
+	// configuration
+	ic.configPath = configPath
+	ic.config = config
+	ic.customizeOSPartitions = (config.Storage != nil) || (config.OS != nil)
```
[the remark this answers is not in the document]
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: fair. pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: I've changed it to 1.5. It's hard to estimate the size without some more calculations. I'm trying to err on the safe side.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

gmileka: It's already included since the squashfs has both partitions included.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: cool! pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: Changed to explicit values: base=10, bitSize=64.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

gmileka: pending (not that we don't have a unity).

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: It actually needs to survive post this point...
isoBuilder has populated its internal state with all the relevant artifacts. That state can be used to regenerate the iso again without re-building any of these artifacts.
I've moved isoBuild out of the parameter structure so that it's not a side effect anymore. It's now returned from `convertInputImageToWriteableFormat` and is fed into `convertWriteableFormatToOutputImage`.

toolkit/tools/imagecustomizer/docs/iso.md
```diff
@@ -2,26 +2,29 @@
...
-
-All customizations can still be made to the input full disk image rootfs as
-usual using MIC, and will naturally carry over into the LiveOS ISO. This
-includes customizations for kernel modules, dracut, and other early boot-time
-actions.
-
-While converting the input full disk image into a LiveOS ISO involves copying
+The Azure Linux Image Customizer can customize an input image and package the
+output as a LiveOS iso image. The input image can be a full disk image
+(vhd/vhdx/qcow2/raw) or another LiveOS iso image. The LiveOS iso output format
```
romoh, line 7: does it have to be `LiveOS` iso only? what will happen if a regular iso is used?

toolkit/tools/imagecustomizer/docs/iso.md
```diff
@@ -111,15 +105,20 @@ os:
...
-    - For ARM64, `bootaa64.efi` and `grubaa64.efi` (or `grubx64-noprefix.efi`).
+    - For ARM64, `bootaa64.efi` and `grubaa64.efi` (or `grubaa64-noprefix.efi`).
   - All grub configurations and related files must be stored under the `/boot`
     folder. For example, grub.cfg cannot reference files outside that folder.
     If it does, those referenced files will not be copied to the iso and may
     cause grub to fail booting the desired operating system.
+
+### LiveOS ISO Image
+
+The input LiveOS iso image must be previously generated by The Azure Linux Image
```
romoh, line 123: nit - lower case `The` or remove it completely.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -402,6 +406,9 @@ func handleBootLoader(baseConfigPath string, config *imagecustomizerapi.Config,
 
 	switch config.OS.ResetBootLoaderType {
 	case imagecustomizerapi.ResetBootLoaderTypeHard:
+		if config.Storage == nil {
```
romoh, line 409: I thought storage is not necessary to `resetBootLoaderType` but the other way around is mandated (where resetBootLoaderType is a must when storage is specified), is that not the case anymore?

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,124 +163,209 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
 
-	// Create 'buildDir' directory.
-	err = os.MkdirAll(buildDirAbs, os.ModePerm)
+	imageCustomizerParameters, err := createImageCustomizerParameters(buildDir, imageFile,
+		baseConfigPath, config,
+		useBaseImageRpmRepos, rpmsSources, enableShrinkFilesystems, outputSplitPartitionsFormat,
+		outputImageFormat, outputImageFile)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create image customizer:\n%w", err)
```
romoh, line 199: nit - improve error message. append `parameters` to the existing message?

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,124 +163,209 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
+		logger.Log.Infof("Skipping customizing OS partitions because none is specified")
+	}
+
+	err = convertWriteableFormatToOutputImage(imageCustomizerParameters, isoBuilder)
 	if err != nil {
-		return fmt.Errorf("failed to convert image file to raw format:\n%w", err)
+		return fmt.Errorf("failed to convert customized raw image to output format:\n%w", err)
+	}
+
+	logger.Log.Infof("Success!")
```
romoh, line 243: unrelated to this PR - I always think it would be good to print the final generated image full path as part of the success message.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -41,6 +49,10 @@ const (
 	dracutConfig = `add_dracutmodules+=" dmsquash-live "
 add_drivers+=" overlay "
 `
+	// the total size of a collection of files is multiplied by the
+	// expansionSafetyFactor to estimate a disk size sufficient to hold those
+	// files.
+	expansionSafetyFactor = 1.5
```
romoh, line 55: curious - is this based of anything specific or just a buffered estimate?

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -915,24 +921,23 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+		return fmt.Errorf("failed to create loop device for (%s):\n%w", isoImageFile, err)
 	}
-	defer func() {
-		unmountParams := []string{mountDir}
-		cleanupErr := shell.ExecuteLive(false, "umount", unmountParams...)
-		if cleanupErr != nil {
-			err = fmt.Errorf("%w:\nfailed to clean-up (%s): %w", err, mountDir, cleanupErr)
-		}
-	}()
+	defer loopDevice.CleanClose()
```
cwize1, line 934: Use `Close` with `defer`. And call `CleanClose` at the bottom of the function, checking the error. There are plenty of examples throughout the codebase.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,124 +163,209 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
+		}
+	case ImageFormatIso:
+		if ic.customizeOSPartitions {
+			err := createLiveOSIsoImage(ic.buildDir, ic.configPath, ic.config.Iso, ic.rawImageFile, ic.outputImageDir, ic.outputImageBase)
+			if err != nil {
+				return fmt.Errorf("failed to create LiveOS iso image:\n%w", err)
+			}
+		} else {
+			if isoBuilder == nil {
+				return fmt.Errorf("internal error: isoBuilder cannot be nul when the output format is .iso and there are not OS customizations.")
```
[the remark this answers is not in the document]
cwize1: `nil` is the term used in Go.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -43,6 +43,98 @@ var (
...
+	ic.buildDir = buildDir
+
+	buildDirAbs, err := filepath.Abs(buildDir)
+	if err != nil {
+		return nil, err
+	}
+
+	ic.buildDirAbs = buildDirAbs
+
+	err = os.MkdirAll(ic.buildDirAbs, os.ModePerm)
```
cwize1, line 95: Please move the resource creation to be outside of the `createImageCustomizerParameters` function.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: But you create a new copy of the `LiveOSIsoBuilder` type in the `createLiveOSIsoImage` function. So, surely the resources being held by `ic.isoBuilder` aren't needed?

cwize1 (changes requested):

toolkit/tools/imagecustomizer/docs/iso.md
gmileka: Correct, the input iso has to an image previously generated by mic. If the input iso does not have the same exact names and location of bits as the one generated by mic, customization will fail.

gmileka: [we need to locate bootloaders, kernel, initrd, rootfs, grub.cfg in specific locations and with specific names]

toolkit/tools/imagecustomizer/docs/iso.md
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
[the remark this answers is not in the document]
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: not based on anything specific - just a buffered estimate.
I've found out that `du` returns size in disk blocks - not in actual file contents. So, `du` will return 4k for 2 byte file, for example.  This makes me think we can lower the safety factor, since choosing a block size that's not 4k is not common. Open to suggestions here...

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,124 +163,209 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
+		return fmt.Errorf("failed to convert customized raw image to output format:\n%w", err)
+	}
+
+	logger.Log.Infof("Success!")
+
+	return nil
+}
+
+func convertInputImageToWriteableFormat(ic *ImageCustomizerParameters) (isoBuilder *LiveOSIsoBuilder, err error) {
+
```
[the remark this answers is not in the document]
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -43,6 +43,98 @@ var (
...
+	// output image
+	outputImageFormat     string
+	qemuOutputImageFormat string
+	outputImageFile       string
+	outputImageDir        string
+	outputImageBase       string
+
+	// iso builder helper
+	isoBuilder *LiveOSIsoBuilder
```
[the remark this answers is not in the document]
gmileka: though i have removed this. yes, old code. pending.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: I see you're point. You're right. I think we can make that decision right here.
pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: Are we using that pattern because we can't capture the error of a deferred function?

gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: btw:
I've found out that du returns the total size of disk blocks used. It does not just add the exact size of each file. So, du will return 4k for 2 byte file, for example. This makes me think we can lower the safety factor, since choosing a block size that's not 4k is not common. Open to suggestions here...

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: I've found a flow where we do need the input iso builder object even when we're re-creating the squashfs. So, I need to keep it around.
Basically, when doing iso->iso creation, even if we re-create the squashfs (boot + rootfs), we still need the original iso state to determine what extra files the user has packed into it (in a previous customization), and copy them unto the new iso as well.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,124 +163,209 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
-		err = shell.ExecuteLiveWithErr(1, "qemu-img", "convert", "-O", qemuOutputImageFormat, rawImageFile, outputImageFile)
-		if err != nil {
-			return fmt.Errorf("failed to convert image file to format: %s:\n%w", outputImageFormat, err)
-		}
-	case ImageFormatIso:
-		err = createLiveOSIsoImage(buildDir, baseConfigPath, config.Iso, rawImageFile, outputImageDir, outputImageBase)
+	// If outputSplitPartitionsFormat is specified, extract the partition files.
+	if ic.outputSplitPartitionsFormat != "" {
+		logger.Log.Infof("Extracting partition files")
+		err = extractPartitionsHelper(ic.rawImageFile, ic.outputImageDir, ic.outputImageBase, ic.outputSplitPartitionsFormat)
```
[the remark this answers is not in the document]
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
cwize1: `Close` requests the resource to be closed but doesn't wait for it. `CleanClose` waits for the close to succeed, with a timeout.

`Close` is typically used in error cases, where you just want to clean everything up but you don't want/need to wait for it. Whereas, `CleanClose` ensures all the bits have fully flushed before moving on.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,133 +152,236 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
 	return nil
 }
 
+func cleanUp(ic *ImageCustomizerParameters) error {
+
+	var allErrors error
```
cwize1, line 157: Why is this called `allErrors`?

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,133 +152,236 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
 	return nil
 }
 
+func cleanUp(ic *ImageCustomizerParameters) error {
+
```
cwize1, line 156: There shouldn't be an empty newline after `{`.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,133 +152,236 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
+}
+
+func convertInputImageToWriteableFormat(ic *ImageCustomizerParameters) (*LiveOSIsoBuilder, error) {
+	logger.Log.Infof("Converting input image to a writeable format")
+
+	if ic.inputImageFormat == ".iso" {
+
+		inputIsoArtifacts, err := createIsoBuilderFromIsoImage(ic.buildDir, ic.buildDirAbs, ic.inputImageFile)
+		if err != nil {
+			var cleanUpError error
```
cwize1, line 244: This cleanup should occur in `createIsoBuilderFromIsoImage`, not in the caller of `createIsoBuilderFromIsoImage`.

It might be helpful to create a `createIsoBuilderFromIsoImageHelper` function that handles all the initialization logic. Then have the `createIsoBuilderFromIsoImage` call the `createIsoBuilderFromIsoImageHelper` function and call the cleanup logic if `createIsoBuilderFromIsoImageHelper` fails.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,133 +152,236 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
+func customizeOSContents(ic *ImageCustomizerParameters) error {
+
+	// If the user has defined customizations, then should proceed as usual.
+	// However, there are no customizations, and the input is an iso, we
+	// should just return - but call out any command line switches that might
+	// have requested unsupported changes.
+	if !ic.customizeOSPartitions && ic.inputImageFormat == ".iso" {
+
+		if ic.enableShrinkFilesystems {
+			return fmt.Errorf("shrinking file systems is not supported when the input image is an iso.")
```
cwize1, line 287: Error messages should not contain periods (`.`).

cwize1, line 287: Ideally, these errors would be reported before we do any work. (e.g. extracting the ISO contents.)

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,133 +152,236 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
+	return nil, nil
+}
+
+func customizeOSContents(ic *ImageCustomizerParameters) error {
+
+	// If the user has defined customizations, then should proceed as usual.
+	// However, there are no customizations, and the input is an iso, we
+	// should just return - but call out any command line switches that might
+	// have requested unsupported changes.
+	if !ic.customizeOSPartitions && ic.inputImageFormat == ".iso" {
```
cwize1, line 284: It might be worth putting the 'is ISO' check in `ImageCustomizerParameters`.

cwize1, line 284: The `".iso"` string should probably be a global constant.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -71,133 +152,236 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, imageFile
...
+		}
+
+		if ic.outputSplitPartitionsFormat != "" {
+			return fmt.Errorf("extracting partitions is not support when the input image is an iso.")
+		}
+
+		return nil
+	}
+
+	if ic.config.OS == nil {
```
cwize1, line 297: Might be worth adding a comment here.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -869,10 +912,468 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+	}
+	defer os.RemoveAll(mountDir)
+
+	isoImageLoopDevice, err := safeloopback.NewLoopback(isoImageFile)
+	if err != nil {
+		return fmt.Errorf("failed to create loop device for (%s):\n%w", isoImageFile, err)
+	}
+	defer isoImageLoopDevice.Close()
+
+	isoImageMount, err := safemount.NewMount(isoImageLoopDevice.DevicePath(), mountDir, "iso9660" /*fstype*/, unix.MS_RDONLY /*flags*/, "" /*data*/, false /*makeAndDelete*/)
```
cwize1, line 976: Line is a tag long.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -869,10 +912,468 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, isoConfig *imagecusto
...
+	}
+	defer os.RemoveAll(squashMountDir)
+
+	squashfsLoopDevice, err := safeloopback.NewLoopback(b.artifacts.squashfsImagePath)
+	if err != nil {
+		return fmt.Errorf("failed to create loop device for (%s):\n%w", b.artifacts.squashfsImagePath, err)
+	}
+	defer squashfsLoopDevice.Close()
+
+	isoImageMount, err := safemount.NewMount(squashfsLoopDevice.DevicePath(), squashMountDir, "squashfs" /*fstype*/, 0 /*flags*/, "" /*data*/, false /*makeAndDelete*/)
```
cwize1, line 1275: Line is a tag long.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1, line 284: Should we also error out if the input is an ISO and the config has a `.storage` value?

cwize1, line 284: Should we also error out if the input is an ISO and the output isn't an ISO?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: This function used to clean-up multiple resources and would not return after the first error to attempt to do further clean-up. Right now, this is unnecessary - so, I'll revert to simple err check/return pattern.
pending.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: pending

gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: makes sense. I've moved `isobuilder.cleanUp()` into `createIsoBuilderFromIsoImage()` and set it such that it gets called only if there's an error. This 'only if' avoid having to add the call to each error handling in the rest of `createIsoBuilderFromIsoImage()` (and I think we can do without the helper function).
```
func someFunc() (err error ) {
    x := &SomeType{}
    defer func() { if err != nil {x.cleanUp} }()
    err = anotherFunc()
    if err != nil { return err } // -> this invokes the clean-up code.
    return nil // -> this does not invoke the clean-up code.
}

File: toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go [hunk 40]
```diff
@@ -110,17 +113,47 @@ func createImageCustomizerParameters(buildDir string,
...
 	}
 
+	// If there are OS customizations, then we proceed as usual.
+	// If there are no OS customizations, and the input is an iso, we just
+	// return because this function is mainly about OS customizations.
+	// This function also supports shrinking/exporting partitions. While
+	// we could support those functions for input isos, we are choosing to
+	// not support them until there is an actual need/a future time.
+	// We explicitly inform the user of the lack of support here.
+	if !ic.customizeOSPartitions && ic.inputIsIso {
```
cwize1, line 135: Would it make sense to disallow `enableShrinkFilesystems` and `outputSplitPartitionsFormat` when the input is an ISO, regardless of the value of `customizeOSPartitions`?

cwize1 (approved):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
gmileka: it does. I've updated the code.

PR #8807, opened 2024-04-16
fix: move stopGPGAgent into safechroot to stop /dev from being left b…

SeanDougherty:
…ehind in chroot cleanup

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

**What does the PR accomplish, why was it needed?**
This broadens the stopping of the gpgAgent to all safechroots. Previously, Image Customizer was using a safechroot to create customized Azure Linux images and was failing to do a final close-out of the chroot because keyboxd was holding the /dev mount open and the loopback device would fail to be detached.

Demo of before the fix and after the fix:
image

###### Change Log  

- Remove defer stopGPGAgent from installutils' PopulateInstallChroot() because it is now handled when the chroot is closed
- Move stopGPGAgent from installutils into safechroot
- Add stopGPGAgent to safechroot's close() before unmounting

NO

###### Test Methodology

- [full build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=551400&view=results)

### DIFF
--- a/toolkit/tools/internal/safechroot/safechroot.go
+++ b/toolkit/tools/internal/safechroot/safechroot.go
@@ -695,12 +695,12 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
 	return mountPoints
 }
 
-// stopGPGAgent stops gpg-agent and keyboxd if they are running inside the chroot.
+// stopGPGComponents stops gpg-agent and keyboxd if they are running inside the chroot.
 //
-// A GPG agent may have been started while the chroot was in use.
+// A GPG agent may have been started while the chroot was in use. Newer versions of "gnupg2" will also start keyboxd.
 // E.g. when installing the azurelinux-repos-shared package, a GPG import occurs. This starts the gpg-agent process inside the chroot.
 // To be able to cleanly exit the setup chroot, we must stop it.
-func (c *Chroot) stopGPGAgent() (err error) {
+func (c *Chroot) stopGPGComponents() (err error) {
 	_, err = exec.LookPath("gpgconf")
 	if err != nil {
 		// gpgconf is not installed, so gpg-agent is not running.
--- a/toolkit/tools/internal/safechroot/safechroot.go
+++ b/toolkit/tools/internal/safechroot/safechroot.go
@@ -6,6 +6,7 @@ package safechroot
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"os/signal"
 	"path/filepath"
 	"sync"
@@ -416,6 +417,10 @@ func (c *Chroot) Close(leaveOnDisk bool) (err error) {
 	activeChrootsMutex.Lock()
 	defer activeChrootsMutex.Unlock()
 
+	// Stops gpg-agent and keyboxd if they are running inside the chroot.
+	// This is to avoid leaving folders like /dev mounted when the chroot folder is forcefully deleted in cleanup.
+	c.stopGPGAgent()
+
 	if buildpipeline.IsRegularBuild() {
 		index := -1
 		for i, chroot := range activeChroots {
@@ -688,4 +693,34 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
 	// Create a copy of the list so that the caller can't mess with the list.
 	mountPoints := append([]*MountPoint(nil), c.mountPoints...)
 	return mountPoints
+}
+
+// stopGPGAgent stops gpg-agent and keyboxd if they are running inside the chroot.
+//
+// It is possible that one of the packages or post-install scripts started a GPG agent.
+// e.g. when installing the azurelinux-repos SPEC, a GPG import occurs. This starts the gpg-agent process inside the chroot.
+// To be able to cleanly exit the setup chroot, we must stop it.
+func (c *Chroot) stopGPGAgent() {
+	_, err := exec.LookPath("gpgconf")
+	if err != nil {
+		// gpgconf is not installed, so gpg-agent is not running.
+		logger.Log.Warnf("gpgconf is not installed, so gpg-agent is not running: %s", err)
+		return
+	}
+
+	c.UnsafeRun(func() error {
+		err := shell.ExecuteLiveWithCallback(logger.Log.Debug, logger.Log.Warn, false, "gpgconf", "--kill", "gpg-agent")
+		if err != nil {
+			// This is non-fatal, as there is no guarantee the image has gpg agent started.
+			logger.Log.Warnf("Failed to stop gpg-agent. This is expected if it is not installed or is no longer running: %s", err)
+		}
+
+		err = shell.ExecuteLiveWithCallback(logger.Log.Debug, logger.Log.Warn, false, "gpgconf", "--kill", "keyboxd")
+		if err != nil {
+			// This is non-fatal, as there is no guarantee the image has gpg agent started.
+			logger.Log.Warnf("Failed to stop keyboxd. This is expected if it is not installed or is no longer running: %s", err)
+		}
+
+		return nil
+	})
 }
--- a/toolkit/tools/internal/safechroot/safechroot.go
+++ b/toolkit/tools/internal/safechroot/safechroot.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
@@ -433,7 +434,7 @@ func (c *Chroot) Close(leaveOnDisk bool) (err error) {
 
 		// Stops gpg-agent and keyboxd if they are running inside the chroot.
 		// This is to avoid leaving folders like /dev mounted when the chroot folder is forcefully deleted in cleanup.
-		c.stopGPGAgent()
+		c.stopGPGComponents()
 
 		// mount is only supported in regular pipeline
 		err = c.unmountAndRemove(leaveOnDisk, unmountTypeNormal)
@@ -709,16 +710,29 @@ func (c *Chroot) stopGPGComponents() (err error) {
 	}
 
 	err = c.UnsafeRun(func() error {
-		err := shell.ExecuteLiveWithCallback(logger.Log.Debug, logger.Log.Warn, false, "gpgconf", "--kill", "gpg-agent")
-		if err != nil {
-			// This is non-fatal, as there is no guarantee the image has gpg agent started.
-			logger.Log.Debugf("Failed to stop gpg-agent. This is expected if it is not installed or is no longer running: %s", err)
+		stdout, stderr, err := shell.Execute("gpgconf", "--list-components")
+
+		logger.Log.Debugf("gpgconf --list-components output:\n%s", stdout)
+
+		if err != nil || stderr != "" {
+			logger.Log.Errorf("Failed to list gpg components: %s\nUnable to check and stop GPG components.", err)
+			return err
 		}
 
-		err = shell.ExecuteLiveWithCallback(logger.Log.Debug, logger.Log.Warn, false, "gpgconf", "--kill", "keyboxd")
-		if err != nil {
-			// This is non-fatal, as there is no guarantee the image has gpg agent started.
-			logger.Log.Debugf("Failed to stop keyboxd. This is expected if it is not installed or is no longer running: %s", err)
+		componentsToKill := []string{"gpg-agent", "keyboxd"}
+		// Split --list-components stdout on newline and iterate over each line
+		for _, line := range strings.Split(stdout, "\n") {
+			for _, component := range componentsToKill {
+				// Check if the line contains a component to kill
+				if strings.Contains(line, component) {
+					logger.Log.Debugf("Found %s running inside chroot. Stopping it.", component)
+					_, _, err = shell.Execute("gpgconf", "--kill", component)
+					if err != nil {
+						logger.Log.Errorf("Failed to stop %s: %s", component, err)
+						return err
+					}
+				}
+			}
 		}
 
 		return err
--- a/toolkit/tools/internal/safechroot/safechroot.go
+++ b/toolkit/tools/internal/safechroot/safechroot.go
@@ -4,6 +4,7 @@
 package safechroot
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 	"os/exec"
@@ -437,7 +438,7 @@ func (c *Chroot) Close(leaveOnDisk bool) (err error) {
 		err = c.stopGPGComponents()
 		if err != nil {
 			// Don't want to leave a stale root if gpg components fail to exit. Logging a Warn and letting close continue...
-			logger.Log.Warnf("Failed to stop GPG components: %s", err)
+			logger.Log.Warnf("Failed to stop GPG components while tearing down the (%s) chroot: %s", c.rootDir, err)
 		}
 
 		// mount is only supported in regular pipeline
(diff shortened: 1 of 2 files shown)


toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +694,33 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
 	mountPoints := append([]*MountPoint(nil), c.mountPoints...)
 	return mountPoints
 }
+
+// stopGPGAgent stops gpg-agent and keyboxd if they are running inside the chroot.
+//
+// It is possible that one of the packages or post-install scripts started a GPG agent.
+// e.g. when installing the azurelinux-repos SPEC, a GPG import occurs. This starts the gpg-agent process inside the chroot.
```
PawelWMS, line 701:
```suggestion
// E.g. when installing the azurelinux-repos-shared package, a GPG import occurs. This starts the gpg-agent process inside the chroot.
```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +694,33 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
 	mountPoints := append([]*MountPoint(nil), c.mountPoints...)
 	return mountPoints
 }
+
+// stopGPGAgent stops gpg-agent and keyboxd if they are running inside the chroot.
+//
+// It is possible that one of the packages or post-install scripts started a GPG agent.
```
PawelWMS, line 700: I'd even simplify and say that a GPG agent might have been started while the chroot was in use. There's many ways we use chroots, no need to suggest it's only a package installation, which might trigger the agent.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +694,33 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+// e.g. when installing the azurelinux-repos SPEC, a GPG import occurs. This starts the gpg-agent process inside the chroot.
+// To be able to cleanly exit the setup chroot, we must stop it.
+func (c *Chroot) stopGPGAgent() {
+	_, err := exec.LookPath("gpgconf")
+	if err != nil {
+		// gpgconf is not installed, so gpg-agent is not running.
+		logger.Log.Warnf("gpgconf is not installed, so gpg-agent is not running: %s", err)
```
PawelWMS, line 707: I'd suggest either an info or debug level log - this is expected behaviour. Warnings in cases like these tend to lead to confusion for the users and during debugging attempts.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +694,33 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+		return
+	}
+
+	c.UnsafeRun(func() error {
+		err := shell.ExecuteLiveWithCallback(logger.Log.Debug, logger.Log.Warn, false, "gpgconf", "--kill", "gpg-agent")
+		if err != nil {
+			// This is non-fatal, as there is no guarantee the image has gpg agent started.
+			logger.Log.Warnf("Failed to stop gpg-agent. This is expected if it is not installed or is no longer running: %s", err)
```
PawelWMS, line 715: This is no longer expected and should be treated as an error at this point. Same for the next call.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +694,33 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+		}
+
+		err = shell.ExecuteLiveWithCallback(logger.Log.Debug, logger.Log.Warn, false, "gpgconf", "--kill", "keyboxd")
+		if err != nil {
+			// This is non-fatal, as there is no guarantee the image has gpg agent started.
+			logger.Log.Warnf("Failed to stop keyboxd. This is expected if it is not installed or is no longer running: %s", err)
+		}
+
+		return nil
```
PawelWMS, line 724: The function doesn't return anything right now. It should probably return `error`, though.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -416,6 +417,10 @@ func (c *Chroot) Close(leaveOnDisk bool) (err error) {
 	activeChrootsMutex.Lock()
 	defer activeChrootsMutex.Unlock()
 
+	// Stops gpg-agent and keyboxd if they are running inside the chroot.
+	// This is to avoid leaving folders like /dev mounted when the chroot folder is forcefully deleted in cleanup.
+	c.stopGPGAgent()
```
PawelWMS, line 422: Let's move it past the `if index < 0` check. It should be OK to try and close the same chroot twice and this call will probably now start failing, if you try to run it inside a chroot with `/dev` already unmounted. Also, I think there's no need to stop the agent for the non-regular builds branch - the chroot is not being torn down in this scenario.

toolkit/tools/internal/safechroot/safechroot.go
SeanDougherty: What if this is called multiple times on the same chroot? Or if gpgconf is present in the chroot, but there are no active processes/services? Wouldn't erroring out in these scenarios be the opposite of the desired behavior?

toolkit/tools/internal/safechroot/safechroot.go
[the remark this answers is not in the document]
SeanDougherty: Fixed!

toolkit/tools/internal/safechroot/safechroot.go
SeanDougherty: Updated!

toolkit/tools/internal/safechroot/safechroot.go
SeanDougherty: I opted for debug logging when this occurs, happy to discuss further though!

toolkit/tools/internal/safechroot/safechroot.go
SeanDougherty: Fixed!

toolkit/tools/internal/safechroot/safechroot.go
SeanDougherty: Updated!

toolkit/tools/internal/safechroot/safechroot.go
SeanDougherty: Updated!

toolkit/tools/internal/safechroot/safechroot.go
SeanDougherty: Updated!

toolkit/tools/internal/safechroot/safechroot.go
PawelWMS: Good questions, made me investigate more:
- For multiple calls it shouldn't matter - as long as `gpg-agent` is listed with `gpgconf --list-components`, you can ask to kill `gpg-agent` as many times as you want. The command will succeed even if it's not running.
- If the component we want to kill is not listed, I agree we shouldn't error out.

I would like to be intentional with error treatment and clear about our expectations. If we first check, if a component is installed, then we know we can either skip the `--kill` call or killing it should succeed. With that in place, all errors should be unexpected and the function should error out. Then it's up to the caller to handle that. I'd have the caller log a message on the error level at least. We may still attempt to continue unmounting the chroot. Similarly, attempting to kill `keyboxd` even if `gpg-agent` failed seems fine as long as we don't lose the errors and return them to the caller.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +694,35 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
 	mountPoints := append([]*MountPoint(nil), c.mountPoints...)
 	return mountPoints
 }
+
+// stopGPGAgent stops gpg-agent and keyboxd if they are running inside the chroot.
+//
+// A GPG agent may have been started while the chroot was in use.
+// E.g. when installing the azurelinux-repos-shared package, a GPG import occurs. This starts the gpg-agent process inside the chroot.
+// To be able to cleanly exit the setup chroot, we must stop it.
+func (c *Chroot) stopGPGAgent() (err error) {
```
PawelWMS, line 703: One more thing that came to my mind right now: since we're now stopping more than just the GPG agent, it wouldn't hurt to re-name this function and update the comment a bit. One suggestion:
```suggestion
// stopGPGComponents stops gpg-agent and keyboxd if they are running inside the chroot.
//
// A GPG agent may have been started while the chroot was in use. Newer versions of "gnupg2" will also start keyboxd.
// E.g. when installing the azurelinux-repos-shared package, a GPG import occurs. This starts the gpg-agent process inside the chroot.
// To be able to cleanly exit the setup chroot, we must stop it.
func (c *Chroot) stopGPGComponents() (err error) {
```

toolkit/tools/internal/safechroot/safechroot.go
SeanDougherty: I see, I've gone ahead and implemented your recommendation. I believe this is what you were looking for 👍

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +699,48 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+		componentsToKill := []string{"gpg-agent", "keyboxd"}
+		// Split --list-components stdout on newline and iterate over each line
+		for _, line := range strings.Split(stdout, "\n") {
+			for _, component := range componentsToKill {
+				// Check if the line contains a component to kill
+				if strings.Contains(line, component) {
+					logger.Log.Debugf("Found %s running inside chroot. Stopping it.", component)
+					_, _, err = shell.Execute("gpgconf", "--kill", component)
+					if err != nil {
+						err = fmt.Errorf("failed to stop gpg component \"%s\": \n%s", component, err)
```
PawelWMS, line 735:
```suggestion
						err = fmt.Errorf("failed to stop gpg component \"%s\":\n%w", component, err)
```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -430,6 +432,14 @@ func (c *Chroot) Close(leaveOnDisk bool) (err error) {
 			return
 		}
 
+		// Stops gpg-agent and keyboxd if they are running inside the chroot.
+		// This is to avoid leaving folders like /dev mounted when the chroot folder is forcefully deleted in cleanup.
+		err = c.stopGPGComponents()
+		if err != nil {
+			// Don't want to leave a stale root if gpg components fail to exit. Logging a Warn and letting close continue...
+			logger.Log.Warnf("Failed to stop GPG components: %s", err)
```
PawelWMS, line 440:
```suggestion
			logger.Log.Warnf("Failed to stop GPG components while tearing down the (%s) chroot: %s", c.rootDir, err)
```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +699,48 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+		logger.Log.Debugf("gpgconf --list-components output:\n%s", stdout)
+
+		if err != nil || stderr != "" {
+			err = fmt.Errorf("failed to list gpg components.\nerr: %s\nstderr: %s\nUnable to check and stop GPG components", err, stderr)
+			return
+		}
+
+		componentsToKill := []string{"gpg-agent", "keyboxd"}
+		// Split --list-components stdout on newline and iterate over each line
+		for _, line := range strings.Split(stdout, "\n") {
```
PawelWMS, line 728: Summary of an offline discussion: let's refactor by extracting the components listing to a separate function.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +700,66 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+	})
+
+	return
+}
+
+func (c *Chroot) listGPGComponents() (components map[string]bool, err error) {
+	stdout, stderr, err := shell.Execute("gpgconf", "--list-components")
+
+	if err != nil {
+		err = fmt.Errorf("failed to list GPG components.\nerr: %w\nstderr: %s", err, stderr)
```
SeanDougherty, line 744: stderr is not of type Error and therefore can't be expanded with %w

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +700,66 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+	components = make(map[string]bool)
+	for scanner.Scan() {
+		line := scanner.Text()
+		components[strings.Split(line, ":")[0]] = true
+	}
+
+	err = scanner.Err()
+	if err != nil {
+		err = fmt.Errorf("error parsing gpgconf --list-components output: %w", err)
+	}
```
PawelWMS, line 762: I think the line parsing you had earlier was fine. I was probably confusing when I tried to explain my main concern: it was about the decreased readability of multiple nested indentations, the logic itself worked fine. And I did suggest a regex but that was just a dumb suggestion on my end, when I read more closely what we need to do here in terms of parsing.

Anyway, what do you think about this:
```suggestion
	components = make(map[string]bool)
	for _, line := range strings.Split(stdout, "\n") {
		components[strings.Split(line, ":")[0]] = true
	}
```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +700,66 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+			}
+		}
+
+		return
+	})
+
+	return
+}
+
+func (c *Chroot) listGPGComponents() (components map[string]bool, err error) {
```
PawelWMS, line 740: Since `listGPGComponents()` is not accessing any of the `Chroot` members:
```suggestion
		components, err := listGPGComponents()
		// List of components to kill. The names must be verbatim identical to the name tag that is used by `gpgconf`
		componentsToKill := []string{"gpg-agent", "keyboxd"}
		for _, component := range componentsToKill {
			// Check if the line contains a component to kill
			if components[component] {
				logger.Log.Debugf("Found %s running inside chroot. Stopping it.", component)
				_, stderr, err := shell.Execute("gpgconf", "--kill", component)
				if err != nil {
					err = fmt.Errorf("failed to stop GPG component (%s):\nerr: %w\nstderr: %s", component, err, stderr)
					return err
				}
			}
		}

		return
	})

	return
}

func listGPGComponents() (components map[string]bool, err error) {
```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +700,66 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+		components, err := c.listGPGComponents()
+		// List of components to kill. The names must be verbatim identical to the name tag that is used by `gpgconf`
+		componentsToKill := []string{"gpg-agent", "keyboxd"}
+		for _, component := range componentsToKill {
+			// Check if the line contains a component to kill
+			if components[component] {
+				logger.Log.Debugf("Found %s running inside chroot. Stopping it.", component)
+				_, stderr, err := shell.Execute("gpgconf", "--kill", component)
+				if err != nil {
+					err = fmt.Errorf("failed to stop GPG component (%s):\nerr: %w\nstderr: %s", component, err, stderr)
```
PawelWMS, line 728:
```suggestion
					err = fmt.Errorf("failed to stop GPG component (%s):\nerr:\n%w\nstderr: %s", component, err, stderr)
```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +700,66 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+				}
+			}
+		}
+
+		return
+	})
+
+	return
+}
+
```
PawelWMS, line 739: Making sure we don't lose the `err`. Aside from that only a suggestion to trim the indentations.

```suggestion
		return killGPGComponents(componentsToKill, components)
	})

	return
}

func killGPGComponents(componentsToKill []string, availableComponents map[string]bool) (err error) {
	for _, component := range componentsToKill {
		if availableComponents[component] {
			logger.Log.Debugf("Found %s running inside chroot. Stopping it.", component)
			_, stderr, err := shell.Execute("gpgconf", "--kill", component)
			if err != nil {
				return fmt.Errorf("failed to stop GPG component (%s):\nerr: %w\nstderr: %s", component, err, stderr)
			}
		}
	}

	return
}

```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +700,66 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+	_, err = exec.LookPath("gpgconf")
+	if err != nil {
+		// gpgconf is not installed, so gpg-agent is not running.
+		logger.Log.Debugf("gpgconf is not installed, so gpg-agent is not running: %s", err)
+		return
+	}
+
+	err = c.UnsafeRun(func() (err error) {
+		// Split --list-components stdout into a list of name tags, one for each component
+		components, err := c.listGPGComponents()
```
PawelWMS, line 719:
```suggestion
		components, err := c.listGPGComponents()
		if err != nil {
			return err
		}
```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +700,66 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+func (c *Chroot) stopGPGComponents() (err error) {
+	_, err = exec.LookPath("gpgconf")
+	if err != nil {
+		// gpgconf is not installed, so gpg-agent is not running.
+		logger.Log.Debugf("gpgconf is not installed, so gpg-agent is not running: %s", err)
+		return
+	}
+
+	err = c.UnsafeRun(func() (err error) {
+		// Split --list-components stdout into a list of name tags, one for each component
```
PawelWMS, line 718: This comment might now be better located above the definition of the `listGPGComponents` function, since that's closer to its implementation.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -689,3 +700,66 @@ func (c *Chroot) GetMountPoints() []*MountPoint {
...
+//
+// A GPG agent may have been started while the chroot was in use. Newer versions of "gnupg2" will also start keyboxd.
+// E.g. when installing the azurelinux-repos-shared package, a GPG import occurs. This starts the gpg-agent process inside the chroot.
+// To be able to cleanly exit the setup chroot, we must stop it.
+func (c *Chroot) stopGPGComponents() (err error) {
+	_, err = exec.LookPath("gpgconf")
+	if err != nil {
+		// gpgconf is not installed, so gpg-agent is not running.
+		logger.Log.Debugf("gpgconf is not installed, so gpg-agent is not running: %s", err)
+		return
```
PawelWMS, line 714:
```suggestion
		return nil
```

PR #8837, opened 2024-04-18
upgrade bpftrace to 0.20.3 and add patch for llvm18 support

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

upgrade bpftrace to 0.20.3 and add patch for llvm18 support

###### Change Log  

- upgrade bpftrace to 0.20.3
- patch bpftrace for llvm18 support

**NO**

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=553091&view=results

### DIFF
--- a/SPECS/bpftrace/bpftrace-0.20-llvm18.patch
+++ b/SPECS/bpftrace/bpftrace-0.20-llvm18.patch
@@ -1,0 +1,515 @@
+From 686b73cbb7ae5ab24d1f8ad69c608225d14c94e2 Mon Sep 17 00:00:00 2001
+From: Daniel Xu <dxu@dxuuu.xyz>
+Date: Tue, 12 Mar 2024 13:13:01 -0600
+Subject: [PATCH 1/5] Update CHANGELOG
+
+---
+ CHANGELOG.md | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/CHANGELOG.md b/CHANGELOG.md
+index e699c473..c574ad3e 100644
+--- a/CHANGELOG.md
++++ b/CHANGELOG.md
+@@ -9,6 +9,8 @@ and this project adheres to
+ ## Unreleased
+ 
+ #### Added
++- Add LLVM 18 support
++  - [#3051](https://github.com/bpftrace/bpftrace/pull/3051)
+ #### Changed
+ #### Deprecated
+ #### Removed
+-- 
+2.33.8
+
+
+From 79e9282b4d2a87a3269d7fd2c595efa784fda7c4 Mon Sep 17 00:00:00 2001
+From: Wentao Zhang <wentao.zhang@windriver.com>
+Date: Tue, 15 Aug 2023 11:18:36 +0800
+Subject: [PATCH 2/5] replace python with python3 in the test
+
+"runtime:call" in ptest gets the following FAILED:
+python: No such file or directory
+replace python with python3 in the test scripts.
+
+$export BPFTRACE_RUNTIME_TEST_EXECUTABLE=/usr/bin
+$cd /usr/lib/bpftrace/ptest/tests
+$python3 runtime/engine/main.py --filter="call.*"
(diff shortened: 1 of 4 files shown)


jslobodzian (approved):

mfrw (approved):

SPECS/bpftrace/bpftrace-0.20-llvm18.patch
```diff
@@ -0,0 +1,515 @@
+From 686b73cbb7ae5ab24d1f8ad69c608225d14c94e2 Mon Sep 17 00:00:00 2001
```
ddstreetmicrosoft, line 1: Personally, I prefer if our patch file commit ids actually match the upstream commit id (makes it much easier to compare against the actual upstream commits). For example this patch upstream has commit id ce9a456d184d9aac21b2fefcbd46212391b6ebea

SPECS/bpftrace/bpftrace-0.20-llvm18.patch
```diff
@@ -0,0 +1,515 @@
...
+-RUN {{BPFTRACE}} -e 'BEGIN { printf("%s - %s\n", strftime("1%f", 1000000123000), strftime("1%f", 0)); exit(); }' | tail -n +2 | xargs -I{} python -c "print({})"
++RUN {{BPFTRACE}} -e 'BEGIN { printf("%s - %s\n", strftime("1%f", 1000000123000), strftime("1%f", 0)); exit(); }' | tail -n +2 | xargs -I{} python3 -c "print({})"
+ EXPECT 123
+ TIMEOUT 1
+ 
+-- 
+2.33.8
+
+
+From 0cd1b0a341c3562a478c4f9a86fe3f765b5ec2bb Mon Sep 17 00:00:00 2001
```
ddstreetmicrosoft, line 80: Multiple patches in a single file *technically* works but IMHO is not great. I am really not a fan of doing this, and I'm not aware of any technical reason for doing it.

ddstreetmicrosoft (approved):
LGTM besides the comments re: the patch file, but I'll leave it up to you if you want to split up the patch file and/or fix the commit ids.

SPECS/bpftrace/bpftrace-0.20-llvm18.patch
anphel31: I combined them all since they were part of a single upstream PR (which was not squash merged)

ddstreetmicrosoft: as i said, technically it works. but my experience is that there is no actual benefit in doing it, and it makes anyone trying to maintain the package at a later date work significantly harder to work with, e.g. comparing each patch to the real upstream patch, or even just reviewing the individual patches. if you are concerned about future devs being able to easily see that separate patch files are related, my preference is to use a simple patch file naming pattern that prefixes each patch file with info about the 'bug' corresponding to the patch or patch set, for example:
https://launchpad.net/ubuntu/+source/systemd/245.4-4ubuntu3.12

for a fix with a single patch, it's just the actual `git format-patch` filename with the leading `0001` replaced with `lp123456` (where `lp` is for `launchpad` and the number is the corresponding launchpad bug number). for a fix with multiple patches, create a `lp123456` dir and place all the files under it.

Since rpms don't quite allow having files in subdirs it probably would apply to us more easily as just prefixing *all* patch files with the corresponding bug indication tag (for 1 or more than 1 patch file). and also since we don't actually use any bugs to track fixes, it probably would be easier to use something else like the github pr number.

anyway, that's not 'official' for our team at all (though IMHO something like that should be, but i just work here) so i leave it up to you however you want to handle the patch file(s)

anphel31: Will stick with this format for now. I'm hoping we can get rid of the patch as soon as the next bpftrace release is available.

PR #8918, opened 2024-04-25
Image Customizer: Make grub config customization more robust.

cwize1:
This change adds a tokenizer for grub config files. While not a full parser, it does ensure that semicolons, character escapes, single-quotes, and double-quotes are all parsed properly. Using this tokenizer will allow the image customizer to handle a greater variety of grub config files.

In addition, all existing image customizer code that modifies the grub config file has been updated to use the grub tokenizer instead of using regexes and the like.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Added UTs for grub tokenizer.


toolkit/tools/internal/grub/grubtokenizer.go
```diff
@@ -0,0 +1,509 @@
...
+
+	// Check if the word starts with '#'.
+	// Note: A '#' in the middle of a word doesn't start a comment.
+	if c == '#' {
+		err := t.parseComment()
+		if err != nil {
+			return err
+		}
+
+		return nil
```
PawelWMS, line 120: Nit-pick: the check seems redundant in this case - might as well just return `err`.
```suggestion
		return nil
```

PawelWMS, line 120: Nit-pick: the check seems redundant in this case - might as well just return `err`.
```suggestion
		return t.parseComment()
```

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -113,41 +111,25 @@ func updateGrubConfig(dataPartitionIdType imagecustomizerapi.IdType, dataPartiti
...
-			line = "set rootdevice=/dev/mapper/root"
-			updatedLines = append(updatedLines, line)
-		} else {
-			// Add other lines unchanged
-			updatedLines = append(updatedLines, line)
-		}
+		return fmt.Errorf("failed to read grub config:\n%w", err)
+	}
+
+	grub2Config, err = replaceKernelCommandLineArguments(grub2Config, []string{"rd.systemd.verity", "roothash",
```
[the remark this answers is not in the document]
cwize1: `replaceKernelCommandLineArguments` removes existing args and then adds new ones. Perhaps I should rename it to `updateKernelCommandLineArguments`?

toolkit/tools/pkg/imagecustomizerlib/customizeoverlay.go
```diff
@@ -68,40 +67,20 @@ func updateGrubConfigForOverlay(imageChroot *safechroot.Chroot, overlays []image
...
+
+	grub2Config, err := file.Read(grubCfgPath)
 	if err != nil {
-		return fmt.Errorf("failed to read grub config: %w", err)
+		return fmt.Errorf("failed to read grub config:\n%w", err)
 	}
 
-	var updatedLines []string
-	linuxLineRegex, err := regexp.Compile(`^linux .*rd.overlayfs=.*`)
+	grub2Config, err = updateKernelCommandLineArguments(grub2Config, []string{"rd.overlayfs"}, newArgs)
```
gmileka, line 76: why both the array argument names and a formatted string of name/value pairs?
Would it make sense that this function takes an array where each element is a pair of strings?
`struct x { name, value}`

gmileka: To avoid parsing the input - since in most cases, I think, the caller can easily build the array of structures.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -359,11 +423,82 @@ func updateSELinuxCommandLineHelper(grub2Config string, selinuxMode imagecustomi
...
+			return "", fmt.Errorf("grub config has a set command that doesn't have an equals symbol")
+		}
+
+		// Check if the name matches.
+		if name == varName {
+			setVarLines = append(setVarLines, line)
+		}
+	}
+
+	// Ensure there is only 1 set command.
```
gmileka, line 482: would it make sense to move this validation logic to load time? Once successfully loaded, we know that the input grub is valid.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -301,54 +365,54 @@ func updateSELinuxCommandLine(selinuxMode imagecustomizerapi.SELinuxMode, imageC
...
-func updateSELinuxCommandLineHelper(grub2Config string, selinuxMode imagecustomizerapi.SELinuxMode) (string, error) {
-	// Remove all the existing SELinux args.
-	grub2Config, securityIndex, err := tryRemoveKernelCommandLineArgumentValue(grub2Config, "security")
-	if err != nil {
-		return "", err
-	}
-
-	grub2Config, selinuxIndex, err := tryRemoveKernelCommandLineArgumentValue(grub2Config, "selinux")
+// Finds all the kernel command-line args that match the provided names, then insert replacement arg(s).
+func updateKernelCommandLineArguments(grub2Config string, argsToRemove []string, newArg string) (string, error) {
```
gmileka, line 369: It is more consistent if the interface accepts either arrays or strings when expecting arguments/parameters. I'd vote for arrays as they are structured and help us avoid the need to re-parse the contents.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -107,143 +93,221 @@ func replaceToken(inputGrubCfgContent string, oldToken string, newToken string)
...
+	if insertAtToken == nil {
+		return nil, grub.Token{}, fmt.Errorf("failed to find $kernelopts in linux command line")
 	}
 
-	outputGrubCfgContent = fmt.Sprintf("%s%s=%s%s", inputGrubCfgContent[:start], name, value, inputGrubCfgContent[end:])
+	return args, *insertAtToken, nil
+}
 
-	return outputGrubCfgContent, oldValue, nil
+// Filters a list of kernel command-line args to only those that match the provides list of names.
```
gmileka, line 250: nit: provided instead of provides

toolkit/tools/pkg/imagecustomizerlib/customizeoverlay.go
cwize1: Writing it this way is subtlety more powerful. It allows the args that are removed to be different from the args that are set. For example, when setting the SELinux mode, sometimes you want to add `enforcing=1` and sometimes you want to completely remove it.

cwize1: The args that are added should really be an array of strings, instead of a single combined string. I was just being lazy.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Even if we added pre-validation steps, I would probably still keep the on-the-fly checks just as a form of sanity check.

(If you stop the search early, it'll still be O(n).)

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: `newArgs` is now a string array.

toolkit/tools/pkg/imagecustomizerlib/customizeoverlay.go
```diff
@@ -68,43 +67,25 @@ func updateGrubConfigForOverlay(imageChroot *safechroot.Chroot, overlays []image
...
-			newLinuxLine := overlayRegex.ReplaceAllString(trimmedLine, newArgs)
-			updatedLines = append(updatedLines, newLinuxLine)
-		} else if strings.HasPrefix(trimmedLine, "linux ") {
-			// Append new overlay arguments if no existing overlay arguments are found.
-			updatedLines = append(updatedLines, line+" "+newArgs)
-		} else {
-			// Add other lines unchanged.
-			updatedLines = append(updatedLines, line)
-		}
+		return fmt.Errorf("failed to set verity kernel command line args:\n%w", err)
```
romoh, line 83: nit - `overlay` instead of  `verity`

toolkit/tools/internal/grub/tokentests/CommonLineCommandLine.test
```diff
@@ -0,0 +1 @@
+linux $bootprefix/$mariner_linux rd.auto=1 root=$rootdevice $mariner_cmdline lockdown=integrity sysctl.kernel.unprivileged_bpf_disabled=1 $systemd_cmdline console=tty0 console=ttyS0 $kernelopts
```
romoh, line 1: suggestion - maybe add a test for a more involed kernel cmdline. ie
```
console=tty0 console=ttyS0 intel_iommu=on iommu=pt
      processor.max_cstate=1 intel_idle.max_cstate=0 msr.allow_writes=on
      default_hugepagesz=1G hugepagesz=1G hugepages=40 skew_tick=1 nohz=on
      nohz_full=2-31,34-63 rcu_nocbs=2-31,34-63 systemd.cpu_affinity=0,1,32,33
      rcu_nocb_poll nmi_watchdog=0 idle=poll kthread_cpus=0,1,32,33
      irqaffinity=0,1,32,33 isolcpus=nohz,domain,managed_irq,2-31,34-63
      tsc=reliable iomem=relaxed nosoftlockup net.ifnames=1 audit=0 rd.luks=0
```

toolkit/tools/internal/grub/grubtokenizer.go
```diff
@@ -0,0 +1,509 @@
...
+loop:
+	for {
+		c, eof := t.scanner.Peek()
+		if eof {
+			break loop
+		}
+
+		switch {
+		// Name characters.
+		case (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '_':
```
romoh, line 436: should that also include dots `.` ?

romoh (commented):
Left minor comments, otherwise LGTM

toolkit/tools/internal/grub/grubtokenizer.go
cwize1: Variable names in Bash/grub.cfg can't contain dots.

toolkit/tools/pkg/imagecustomizerlib/customizeoverlay.go
cwize1: Thanks! Fixed.

toolkit/tools/internal/grub/tokentests/CommonLineCommandLine.test
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Fixed.

romoh (approved):

toolkit/tools/internal/grub/grubtokenizer.go
```diff
@@ -0,0 +1,509 @@
...
+		if eof {
+			break loop
+		}
+
+		switch c {
+		case '\n':
+			break loop
+
+		default:
+			sb.WriteRune(c)
```
gmileka, line 214: sb (the comment) is not captured, correct?

gmileka: should the parser be able to re-construct the input file from the tokens? if so, will the commends be preserved?

toolkit/tools/internal/grub/grubtokenizer.go
```diff
@@ -0,0 +1,509 @@
...
+			Start: locStart,
+			End:   locEnd,
+		},
+		Type:       tokenType,
+		RawContent: rawContent,
+	}
+	return token
+}
+
+func (t *grubConfigTokenizer) newSubWord(locStart, locEnd filescanner.SourceLoc, subWordType SubWordType, value string,
```
gmileka, line 493: would it make sense to document/define what a `token`, a `subword`, and a `word` are?

toolkit/tools/internal/grub/tokentests/CommonLineCommandLine.result
```diff
@@ -0,0 +1,29 @@
...
+WORD[1:61-1:77][60-76]
+  VAR_EXPANSION[1:61-1:77][60-76]("mariner_cmdline")
+WORD[1:78-1:96][77-95]
+  KEYWORD_STRING[1:78-1:96][77-95]("lockdown=integrity")
+WORD[1:97-1:138][96-137]
+  KEYWORD_STRING[1:97-1:138][96-137]("sysctl.kernel.unprivileged_bpf_disabled=1")
+WORD[1:139-1:155][138-154]
+  VAR_EXPANSION[1:139-1:155][138-154]("systemd_cmdline")
+WORD[1:156-1:168][155-167]
+  KEYWORD_STRING[1:156-1:168][155-167]("console=tty0")
```
gmileka, line 25: curious, why not breaks such property=value into smaller tokens? Would that make reset values easier (easier search of properties, and easier setting).

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -107,143 +93,228 @@ func replaceToken(inputGrubCfgContent string, oldToken string, newToken string)
 	return outputGrubCfgContent, nil
 }
 
-func setLinuxPath(inputGrubCfgContent string, linuxPath string) (outputGrubCfgContent string, oldKernelPath string, err error) {
-	match := linuxPathRegex.FindStringSubmatchIndex(inputGrubCfgContent)
-	if match == nil {
-		return "", "", fmt.Errorf("failed to find kernel command line")
+// Find the linux command within the grub config file.
+func findLinuxLine(inputGrubCfgContent string) ([]grub.Token, error) {
+	linuxLine, err := findSingularGrubCommand(inputGrubCfgContent, "linux")
```
gmileka, line 98: how does this handle the case where there are if statements/multiple boot menu items where each has its own linux/initrd commands?

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -107,143 +93,228 @@ func replaceToken(inputGrubCfgContent string, oldToken string, newToken string)
...
-		argValueGroupEndIndex   = argValueGroupIndex*2 + 1
-	)
-
-	escapedName := regexp.QuoteMeta(name)
-
-	// "linux ... <name>=<value> ..."
-	argumentRegexString := "(?m)[\\t ]+linux[\\t ](.*[\\t ])?(" + escapedName + "(=(\\S*)?)?)[\\t ]"
-	argumentRegex, err := regexp.Compile(argumentRegexString)
+// - args: A list of kernel command-line arguments.
+// - insertToken: A tokenizer token that represents an appropriate insert point for any new args.
```
gmileka, line 180: We probably need some clarification on what 'appropriate' means for the `insertToekn`. It sounds like it means right before `$kernelopts`?

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -107,143 +93,228 @@ func replaceToken(inputGrubCfgContent string, oldToken string, newToken string)
...
+		name, value, foundEqSymbol := strings.Cut(argString, "=")
+		if !foundEqSymbol && hasVarExpansion {
+			// Arg has a variable expansion (e.g. $a) that is part of the arg name.
+			// There isn't any easy way to handle such args. So, just ignore them.
+			continue
+		}
+
+		if hasVarExpansion {
+			// The arg string value isn't known because it contains a variable expansion.
+			value = ""
```
gmileka, line 235: why set it to empty string and not just keep the `$variable`?

toolkit/tools/internal/grub/grubtokenizer.go
cwize1: Old code. I originally gave comments their own token type. But decided that made it more difficult to interpret the tokens.

As it stands, you can't perfectly reconstruct the original file due to missing tokens for whitespace and comments.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -107,143 +93,228 @@ func replaceToken(inputGrubCfgContent string, oldToken string, newToken string)
...
+// Filters a list of kernel command-line args to only those that match the provided list of names.
+func findMatchingCommandLineArgs(args []grubConfigLinuxArg, names []string) []grubConfigLinuxArg {
+	matching := []grubConfigLinuxArg(nil)
+
+argsLoop:
+	for _, arg := range args {
+		for _, name := range names {
+			if arg.Name == name {
+				matching = append(matching, arg)
+				continue argsLoop
```
gmileka, line 263: what happens when some arguments are repeated?

toolkit/tools/internal/grub/tokentests/CommonLineCommandLine.result
cwize1: The `=` symbol only has meaning in certain contexts. So, it is probably better to only handle `=` in the contexts where it matters.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: It errors out.

If a user ever requests that we handle grub.cfg files with multiple boot entries, then we can figure out how to handle the situation.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -107,143 +93,228 @@ func replaceToken(inputGrubCfgContent string, oldToken string, newToken string)
...
+	lastArg := foundArgs[len(foundArgs)-1]
+	if lastArg.ValueHasVarExpansion {
+		return "", fmt.Errorf("kernel arg (%s) has variable expansion in value", name)
+	}
+
+	return lastArg.Value, nil
+}
+
+// Finds an existing kernel command-line arg and replaces its value.
+func replaceKernelCommandLineArgumentValue(inputGrubCfgContent string, name string, value string,
```
gmileka, line 290: nit: consistency Argument vs Arg in the function name

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -107,143 +93,228 @@ func replaceToken(inputGrubCfgContent string, oldToken string, newToken string)
...
 	if err != nil {
-		return "", 0, err
+		return "", "", err
 	}
 
-	if start == -1 {
-		// Arg not found.
-		return inputGrubCfgContent, -1, nil
+	foundArgs := findMatchingCommandLineArgs(args, []string{name})
+	if len(foundArgs) < 1 {
```
gmileka, line 301: would it make sense to have replace be more forgiving in case it does not find the existing instance - maybe through an extra flag to say `appendIfAbsent` or something along those lines?
In some cases, we do not know if the argument is present or not - but we want to replace it if it exists, or append if it does not. So, instead of two callls, one for presence, and one for replacement, maybe replacement can do both.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -107,143 +93,228 @@ func replaceToken(inputGrubCfgContent string, oldToken string, newToken string)
...
+		return "", "", fmt.Errorf("failed to find kernel arg (%s)", name)
 	}
+	if len(foundArgs) > 1 {
```
gmileka, line 304: Also, here, can we just eliminate all those duplicate instances and replacement them with our new single instance?

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: The 'replace or insert' behavior is covered by the `updateKernelCommandLineArguments` function.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: The `updateKernelCommandLineArguments` function also already handles removing duplicate args.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Added clarification to code comment.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Added clarification to code comment.

gmileka (approved):

toolkit/tools/internal/grub/grubtokenizer.go
cwize1: Added some code comments on the types. Please let me know if you think they need more details.

toolkit/tools/internal/grub/commands.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package grub
```
neha170, line 4: Overall comment on directory structure
Instead of having these libraries in internal/, we can have it in pkg/. Since both internal/ and package are rooted at tools/, they essentially have the same "visibility" in our codebase.

toolkit/tools/internal/grub/filescanner/filescanner.go
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package filescanner
```
neha170, line 4: If this filescanner functionality is specific to grub files, I would put this under the `package grub`. `filescanner` sounds like `fileutils` to me (usable across the board, not specific to grub files).
Please disregard if the implementation is not specific to grub usecase. I wasn't sure because you have placed it under grub/.

toolkit/tools/internal/grub/commands.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if len(line) > 0 {
+		lines = append(lines, line)
+	}
+
+	return lines
+}
+
+// Checks if a token is keyword.
+func IsTokenKeyword(token Token, keyword string) bool {
```
neha170, line 40: nit: we usually word function description as 
`IsTokenKeyword checks if a token is keyword`

toolkit/tools/internal/grub/grubtokenizer.go
```diff
@@ -0,0 +1,507 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/grub/filescanner"
+)
+
+func TokenizeGrubConfigFile(name string) ([]Token, error) {
```
neha170, line 19: nit: since this is already a part of `grub` package, we can consider dropping grub from the function names.

toolkit/tools/internal/grub/commands.go
cwize1: The visibility is the same within our codebase, yes. But if a package from outside our codebase uses this code, then they'll be able to access stuff under `pkg` but not `internal`. So, I think it makes sense to limit `pkg` to only the library equivalents of the toolkit's tools.

toolkit/tools/internal/grub/commands.go
cwize1: Fixed.

toolkit/tools/internal/grub/grubtokenizer.go
cwize1: Fixed.

toolkit/tools/internal/grub/filescanner/filescanner.go
cwize1: It's actually a fairly generic library. I've moved the package up a directory.

PR #8930, opened 2024-04-26
kernel: add kernel-uki package

ddstreetmicrosoft:
Generate a `kernel-uki` package containing a prebuilt kernel uki. This is initially intended for use with CVMs and eventually should be signed.


ddstreetmicrosoft:
@rlmenge @christopherco fyi

SPECS/kernel/kernel-uki.spec
```diff
@@ -0,0 +1,65 @@
...
+# Note - while Fedora's kernel version includes the %%_target_cpu as a
+# suffix, our kernel version does not.
+%define kernelver %{version}-%{release}
+
+%define cmdline console=ttyS0
+
+Summary:        Linux Kernel
+Name:           kernel-uki
+Version:        6.6.22.1
+Release:        2%{?dist}
```
ms-mahuber, line 16: Why do we start with '2'? 

Generally, this is aligned with the kernel.spec. Is our intent to update this SPEC each time the kernel gets updated (version)?

ddstreetmicrosoft: right, the %{release} is 2 here because the kernel.spec is also 2. this *should* be simply a subpackage of the normal `kernel.spec` but as stated in the `kernel-uki.spec` it can't be, due to "circular dependencies" that are a result of our build design (i.e. systemd BuildRequires: bpftool which is built by the kernel, and kernel-uki BuildRequires: systemd-ukify, so we can't put kernel-uki into the kernel.spec because it would create a "circular dependency" because we "build from scratch").

SPECS/kernel/kernel-uki-dracut.conf
```diff
@@ -0,0 +1,38 @@
...
+dracutmodules+=" crypt crypt-loop tpm2-tss "
+
+# modules: support root on virtiofs
+dracutmodules+=" virtiofs "
+
+# modules: use sysext images (see 'man systemd-sysext')
+dracutmodules+=" systemd-sysext "
+
+# drivers: virtual buses, pci
+drivers+=" virtio-pci virtio-mmio "      # qemu-kvm
```
Camelron, line 24: These drivers are in addition to the base set required by defaults.conf packaged in dracut, yes?

trungams: This original config is here: https://src.fedoraproject.org/rpms/kernel/blob/rawhide/f/dracut-virt.conf.

I don't think we are using `defaults.conf` for our UKI.

SPECS/kernel/kernel-uki.spec
```diff
@@ -0,0 +1,69 @@
...
+BuildRequires:  systemd-ukify
+BuildRequires:  dracut
+BuildRequires:  binutils
+BuildRequires:  systemd-boot-unsigned
+BuildRequires:  systemd-udev
+BuildRequires:  system-release
+BuildRequires:  tpm2-tools
+BuildRequires:  cryptsetup
+BuildRequires:  device-mapper
+
```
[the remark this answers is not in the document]
trungams: The output of this package is an EFI binary, so no runtime dependency should be required.

SPECS/kernel/kernel-uki-dracut.conf
Camelron: This is inconsequential because it is only used during rpmbuild time and neither packaged nor installed on a target system.

Camelron: In other words, this configuration is isolated to the initrd built, added to the UKI image which is okay.

Camelron (approved):

PawelWMS (approved):

neha170 (approved):

rlmenge (approved):

PR #8941, opened 2024-04-29
azurelinux-sysinfo: new system information properties for host type and architecture

amritakohli:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/SPECS/azurelinux-sysinfo/sysinfo-schema-v1.json
+++ b/SPECS/azurelinux-sysinfo/sysinfo-schema-v1.json
@@ -254,17 +254,38 @@
             "$id": "#/properties/system_info",
             "type": "object",
             "title": "System Information",
-            "description": "Information about the system-wide settings",
+            "description": "Information about the system",
             "properties": {
                 "selinux_mode": {
-                    "$id": "#/properties/package_info/properties/package_count",
+                    "$id": "#/properties/system_info/properties/selinux_mode",
                     "type": "string",
                     "title": "SELinux Mode",
-                    "description": "Enforced or Permissive"
+                    "description": "Provides information about whether SELinux mode is enforcing or permissive"
+                },
+                "system_manufacturer": {
+                    "$id": "#/properties/system_info/properties/system_manufacturer",
+                    "type": "string",
+                    "title": "System Manufacturer",
+                    "description": "Provides information about the system manufacturer"
+                },
+                "computer_architecture": {
+                    "$id": "#/properties/system_info/properties/computer_architecture",
+                    "type": "string",
+                    "title": "Computer Architecture",
+                    "description": "Provides information about the machine hardware"
+                },
+                "host_type": {
+                    "$id": "#/properties/system_info/properties/host_type",
+                    "type": "string",
+                    "title": "Host Type",
+                    "description": "Provides information about whether the host is physical or virtual"
                 }
             },
             "required": [
-                "selinux_mode"
+                "selinux_mode",
+                "system_manufacturer",
+                "computer_architecture",
+                "host_type"
             ],
             "additionalProperties": false
         }


SPECS/azurelinux-sysinfo/sysinfo-schema-v1.json
```diff
@@ -254,17 +254,38 @@
...
             "properties": {
                 "selinux_mode": {
-                    "$id": "#/properties/package_info/properties/package_count",
+                    "$id": "#/properties/system_info/properties/selinux_mode",
                     "type": "string",
                     "title": "SELinux Mode",
-                    "description": "Enforced or Permissive"
+                    "description": "Provides information about whether SELinux mode is enforcing or permissive"
+                },
+                "computer_architecture": {
```
romoh, line 265: would recommend keeping it as just `architecture`

SPECS/azurelinux-sysinfo/collect-sysinfo
```diff
@@ -168,7 +168,7 @@ def collect_resource_utilization():
 def collect_package_info():
     print("Collecting package info...")
     get_package_list = subprocess.run(
-        ["rpm", "-qa"], capture_output=True, text=True, check=True
```
romoh, line 171: removing check=True means we fail on non-zero exist code - is that intentional?

romoh: yes exactly (my comment had a typo, I mean `we won't fail`). Is that a desired state? 
IMO, there should be mandatory fields and optional fields. For mandatory, we should fail if command failed while for things like `cloud-init blame` , we can ignore exit-code.

amritakohli: Okay, added back check for now and will think of how to handle failures in a separate task

SPECS/azurelinux-sysinfo/collect-sysinfo
```diff
@@ -220,10 +220,36 @@ def get_selinux_mode():
         ["getenforce"], capture_output=True, text=True, check=True
     ).stdout.strip()
 
+def get_system_manufacturer():
+    return subprocess.run(
+        ["cat", "/sys/devices/virtual/dmi/id/sys_vendor"], capture_output=True, text=True
```
romoh, line 225: no `check=True` here

romoh: I'm not sure about this as a mechanism for detecting virtual vs physical. This will show the vendor and doesn't necessarily have to be QEMU.
For example my linux hyper-v VM shows "Microsoft" under that path,

SPECS/azurelinux-sysinfo/collect-sysinfo
```diff
@@ -220,10 +220,36 @@ def get_selinux_mode():
...
+    ).stdout.strip()
+
+def get_host_type(sys_vendor):
+    if sys_vendor == "QEMU":
```
romoh, line 229: Maybe you can compare this to a list of vendors instead, so it captures the different scenarios - ie: ['VMware', 'Microsoft Corporation', 'QEMU']

SPECS/azurelinux-sysinfo/collect-sysinfo
amritakohli: that means your hyper-v vm is on microsoft device? if you want to capture whether the machine is a VM that is different. this simply identifies whether the **host** machine is virtual or physical. eg if you create a vm on your hyperv vm it will show QEMU under the path

amritakohli: and yes i know theres no check=True there - dont think we need the service to fail in this case right now

PR #8952, opened 2024-04-29
config: azurelinux 3.0 kernel config

allenpais:
Optimize and Converge Kernel Configuration for Enhanced
    Performance and Maintainability

    This config is a based on:
    https://github.com/microsoft/azurelinux/blob/3.0/SPECS/kernel/config

    Key Changes:
      - Deprecated outdated configuration options that no longer
        provide benefits in current hardware and software contexts.

      - We are suggesting these optimizations for our WSL/LCOW
        convergence to azurelinux.

    Signed-off-by: Allen Pais <apais@linux.microsoft.com>


SPECS/kernel/config
```diff
@@ -293,11 +293,8 @@ CONFIG_TRACEPOINTS=y
 #
 CONFIG_CRASH_CORE=y
 CONFIG_KEXEC_CORE=y
-CONFIG_HAVE_IMA_KEXEC=y
-CONFIG_KEXEC=y
-CONFIG_KEXEC_FILE=y
-# CONFIG_KEXEC_SIG is not set
-# CONFIG_KEXEC_JUMP is not set
+# CONFIG_KEXEC is not set
+# CONFIG_KEXEC_FILE is not set
```
tyhicks, line 297: @allenpais `CONFIG_KEXEC_FILE` still needs to be enabled. We just want to disable `CONFIG_KEXEC` because it doesn't support signature verification.

PR #8972, opened 2024-04-30
Image Customizer: Cleanup warning messages.

cwize1:
1. Don't call `sfdisk --delete` on disks that are known to be empty.

2. Don't call `gpgconf --kill` in chroot contexts that don't have the default mounts (e.g. /proc) since the call won't work.

3. Don't log the stderr of `dd` and `mkisofs` to WARN since those commands use stderr for information messages.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Ran UTs.
- Manually built and tested installer ISO.

### DIFF
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -424,16 +424,18 @@ func WaitForDevicesToSettle() error {
 
 // CreatePartitions creates partitions on the specified disk according to the disk config
 func CreatePartitions(diskDevPath string, disk configuration.Disk, rootEncryption configuration.RootEncryption,
-	readOnlyRootConfig configuration.ReadOnlyVerityRoot,
+	readOnlyRootConfig configuration.ReadOnlyVerityRoot, diskKnownToBeEmpty bool,
 ) (partDevPathMap map[string]string, partIDToFsTypeMap map[string]string, encryptedRoot EncryptedRootDevice, readOnlyRoot VerityDevice, err error) {
 	const timeoutInSeconds = "5"
 	partDevPathMap = make(map[string]string)
 	partIDToFsTypeMap = make(map[string]string)
 
 	// Clear any old partition table info to prevent errors during partition creation
-	_, stderr, err := shell.Execute("sfdisk", "--delete", diskDevPath)
-	if err != nil {
-		logger.Log.Warnf("Failed to clear partition table. Expected if the disk is blank: %v", stderr)
+	if !diskKnownToBeEmpty {
+		_, stderr, err := shell.Execute("sfdisk", "--delete", diskDevPath)
+		if err != nil {
+			logger.Log.Warnf("Failed to clear partition table. Expected if the disk is blank: %v", stderr)
+		}
 	}
 
 	// Create new partition table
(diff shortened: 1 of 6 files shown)


toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -424,16 +424,18 @@ func WaitForDevicesToSettle() error {
...
 	partIDToFsTypeMap = make(map[string]string)
 
 	// Clear any old partition table info to prevent errors during partition creation
-	_, stderr, err := shell.Execute("sfdisk", "--delete", diskDevPath)
-	if err != nil {
-		logger.Log.Warnf("Failed to clear partition table. Expected if the disk is blank: %v", stderr)
+	if !diskKnownToBeEmpty {
+		_, stderr, err := shell.Execute("sfdisk", "--delete", diskDevPath)
+		if err != nil {
+			logger.Log.Warnf("Failed to clear partition table. Expected if the disk is blank: %v", stderr)
```
romoh, line 437: Are there cases where this can still happen with the given error message `Expected if disk is blank`?

toolkit/tools/pkg/isomakerlib/isomaker.go
```diff
@@ -200,7 +200,7 @@ func (im *IsoMaker) buildIsoImage() error {
 		// Directory to convert to an ISO.
 		im.buildDirPath)
 
-	return shell.ExecuteLive(false /*squashErrors*/, "mkisofs", mkisofsArgs...)
+	return shell.ExecuteLive(true /*squashErrors*/, "mkisofs", mkisofsArgs...)
```
romoh, line 203: nit - maybe worth adding a comment here for why (same for dd)

romoh (approved):
minor comments

toolkit/tools/imagegen/diskutils/diskutils.go
cwize1: Yes. If you are dealing with a real disk, then the disk might still have data on it from the previous install. Or the disk might be completely clean (e.g. new).

PawelWMS: I think I didn't get this part. What does `diskKnownToBeEmpty` mean exactly if it's still possible for the disk to be blank (according to the warning message) even when `diskKnownToBeEmpty` is `false`? And since at this point we're interesting in making sure the disk is cleared regardless, if it had any data on it or not, would it be possible to check that first and then expect no errors from `sfdisk` instead of passing another argument and having to still potentially ignore legitimate errors.

cwize1: If `diskKnownToBeEmpty` is `true`, then the disk is guaranteed to be empty. But if `diskKnownToBeEmpty` is `false`, then the status of the disk is not known.

I don't think there is a reliable way to tell if an existing disk has data or not. It's just a blob of bits at the end of the day.

cwize1: My goal for this change is to reduce number of the situations where the "Failed to clear partition table. Expected if the disk is blank" is seen.

toolkit/tools/pkg/isomakerlib/isomaker.go
cwize1: Fixed.

anphel31 (approved):

dmcilvaney (approved):

PR #8996, opened 2024-05-02
Image Customizer: Increase timeout for loopback detach

romoh:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Issues are reported when detaching the loopback devices where it will timeout before the detach is completely done.
```
2024/05/01 23:02:48 image customization failed: timed out waiting for loopback device (/dev/loop36) for disk (/customizer/image.raw) to close
```
The change increments the number of attempts which also increases the delay with each and overall timeout.

**YES/**NO****

###### Test Methodology

Local testing, no repro anymore with the change.

### DIFF
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -379,7 +379,7 @@ func WaitForLoopbackToDetach(devicePath string, diskPath string) error {
 	}
 
 	delay := 100 * time.Millisecond
-	attempts := 10
+	attempts := 8
 	for failures := 0; failures < attempts; failures++ {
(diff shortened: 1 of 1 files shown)


toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -379,7 +379,7 @@ func WaitForLoopbackToDetach(devicePath string, diskPath string) error {
 	}
 
 	delay := 100 * time.Millisecond
-	attempts := 5
+	attempts := 10
```
cwize1, line 382: This sets the total timeout to 102 seconds, which might be a tad long. I think 8 loops for a total of 25 seconds might be a bit more reasonable.

cwize1 (approved):

PR #9060, opened 2024-05-08
Add new license validator tool

dmcilvaney:
##### Merge Checklist  

- [x] Ready to merge

---

##### Summary 

Some of our packages have poorly configured license files (e.g. `COPYING`, `license.txt`, etc.) that should accompany the packages when installed. Of most critical concern are packages which incorrectly mark these files as `%doc`, either explicitly or automatically based on file location. We care about this since [we would like to be able to disable documentation](https://github.com/microsoft/azurelinux/pull/8776) when making size-constrained images. If the license files are incorrectly tagged as `%doc` they will not be installed in the images. This is important if we release these images.  (The most critical issues affecting the core images are fixed [here](https://github.com/microsoft/azurelinux/pull/9147)).

A tool was required to find these mis-configured packages to fix them, and ideally, we want to ensure that future packages do not regress. This is the first version of that regression checker. It currently is a stand-alone tool which can be run for all locally build packages, all packages used in an image, or an arbitrary directory containing `*.rpm` files.

In the future this tool will be integrated with the scheduler to validate rpms when they are built, and the imager tool to detect issues (especially if documentation is disabled).

###### Invoking the tool

This PR adds new make targets:
- `sudo make license-check SRPM_PACK_LIST="pkg1 pkg2 pkg3"` will build `pgk`, `pkg2`, `pgk3` and then validate their licenses
- `sudo make license-check-img CONFIG_FILE="./imageconfigs/core-efi.json"` will gather all the packages needed for `core-efi`, then validate their licenses
- `sudo make license-check LICENSE_CHECK_DIRS="./path/to/rpms ./other/path/to/more/rpms"` will scan for any `*.rpm` files in those directories, recursively.

###### Implementation

The changes are split into two parts: the `licensecheck` command line, and the common `licensecheck` package. The common package is designed to also be used as part of the image and package build tooling.

There are two parts to the package, a wrapper tool based on the `simpletoolchroot` package. It manages a worker chroot and provides `New()`, `CleanUp()`, `CheckLicenses()`, `GetAllResults()`/`FormatResults()`. These take a directory, mounts it into a chroot worker, and scans it, and then offers the results either as lists or a formatted message. These functions in turn call the second set of functions. The second set are used to scan individual RPMs and are suitable for use in any mariner-like environment. `CheckRpmLicenses()` will scan a specific RPM for license issues. `IsALicenseFile()` and `IsASkippedLicenseFile()` are used to interact with the matching heuristic (see below).

The tool outputs several types of results:
- `BadDocs`: A critical issue where a `%doc` file is actually a license file. These are must-fix.
- `BadFiles`: Less critical, a general file that the tool thinks is a license file. Still considered an error.
- `DuplicatedDocs`: A warning, a license file appears to be duplicated in the docs.

###### Exceptions

The tool allows exceptions for bad detections. Exceptions can be added to ` 
toolkit/resources/manifests/package/license_file_exceptions.json` and it will be ignored by the tools.

###### Heuristic

The matching is one based on filename only currently. In `toolkit/tools/pkg/licensecheck/licensenames.go` there are three lists, `licenseNamesFuzzy`, `licenseNamesVerbatim`, `licenseNamesSkip`. The fuzzy names are case-insensitive regexes that are matched against the file paths. They will match any substrings inside the file path, ignoring some well-known prefixes (e.g. `/usr/share/licenses/pkg/licenses/somefile.txt` will be converted to `licenses/somefile.txt` before matching, it would match against `(?i).*license.*`). The verbatim matches target only the basename of files exactly (e.g. `/usr/share/licenses/pkg/licenses/MIT` will be converted to `MIT`, it would match `^MIT$`). Finally the skip lists behave the same as the first set of matches, but are files to ignore.

###### Test Data

A pair of tools can be used to validate the heuristic. `generate_test_data.py` will scan the currently active repos of its host machine (use azl3 container to get azl3 data for example) and generate lists of files: `all_docs_<date>.txt`, `all_licenses_<date>.txt` and `all_other_files_<date>.txt` (The all other files list is git ignored since it is very large). These can be used to calculate false positive/ false negative data with the go tool `licensetestchecker.go` (See `README_license_check.md` for details on regenerating and testing the lists).

As of 2024-05-07 the results are:
- 66 false negative results against real %license files (1.91%)
- 383 false positive against real %doc files (0.25%)
- 1535 false positive against all other (not doc or license) files (0.47%)

**NOTE: Some of these "false positives" are actually correct detections of legitimately bad license files**

###### Limitations

The tool does not currently examine the contents of the files, only their paths. It therefor can only detect well-known styles of licenses.

It also is not able to determine if a package is MISSING licenses, only if the ones already included are packaged correctly. This has two parts: A package may completely omit required licenses, but it may also include them in the wrong sub-package. e.g. `pkg-common` which is required by both `pkg`, `pkg-libs`. The license files may be included in `pkg-common` and since it is required by the others it will bring the required license files. However, if the licenses were instead included in `pkg`, `pkg-libs` could be installed without pulling in the licenses (assuming there is no dependency from `pkg-libs` -> `pkg`). 

###### Future Work

This tool must be manually run currently. Future features:
- Document via contributing guide
- Add a test to our build pipelines to flag these
- Integrate into the package build tool (first as a warning, then an error)
- Consider if we should handle the limitations listed above

###### Example output
``` bash
sudo make license-check-img CONFIG_FILE=./imageconfigs/core-efi.json
```
Gives...
```
...

/home/damcilva/repos/CBL-Mariner/toolkit/out/tools/licensecheck \
        --rpm-dirs="/home/damcilva/repos/CBL-Mariner/build/imagegen/core-efi/package_repo" \
        --exception-file="/home/damcilva/repos/CBL-Mariner/toolkit/resources/manifests/package/license_file_exceptions.json" \
        --worker-tar="/home/damcilva/repos/CBL-Mariner/build/worker/worker_chroot.tar.gz" \
        --build-dir="/home/damcilva/repos/CBL-Mariner/build/license_check_tool" \
        --dist-tag=.azl3 \
         \
        --results-file="/home/damcilva/repos/CBL-Mariner/out/license_check/license_check_results_image_core-efi.json"
INFO[0000][licensecheck] Preparing license check environment for /home/damcilva/repos/CBL-Mariner/build/imagegen/core-efi/package_repo... 
INFO[0005][licensecheck] Scanning /home/damcilva/repos/CBL-Mariner/build/imagegen/core-efi/package_repo for license issues... 
INFO[0005][licensecheck] Queuing 217 rpms for search                  
INFO[0005][licensecheck] Searching rpms...                            
INFO[0005][licensecheck] Checked 22/217 rpms (10%)          
          
...

INFO[0006][licensecheck] Checked 217/217 rpms (100%)                  
INFO[0006][licensecheck] Search results for /home/damcilva/repos/CBL-Mariner/build/imagegen/core-efi/package_repo:
WARN: 'libaio-0.3.113-1.azl3.x86_64.rpm' has license warnings:
        duplicated license files:
                /usr/share/doc/libaio/COPYING
WARN: 'slang-2.3.3-1.azl3.x86_64.rpm' has license warnings:
        duplicated license files:
                /usr/share/doc/slang-2.3.3/COPYING
WARN: 'sudo-1.9.15p5-1.azl3.x86_64.rpm' has license warnings:
        duplicated license files:
                /usr/share/doc/sudo-1.9.15p5/LICENSE.md
ERROR: 'libdb-5.3.28-7.azl3.x86_64.rpm' has license errors:
        bad general file:
                /usr/share/licenses/LICENSE
                /usr/share/licenses/README
ERROR: 'lmdb-libs-0.9.31-1.azl3.x86_64.rpm' has license errors:
        bad %doc files:
                /usr/share/doc/lmdb/COPYRIGHT
        bad general file:
                /usr/share/licenses/lmdb/LICENSE
ERROR: 'userspace-rcu-0.14.0-1.azl3.x86_64.rpm' has license errors:
        bad %doc files:
                /usr/share/doc/userspace-rcu/LICENSE 
INFO[0006][licensecheck] Writing results to file: /home/damcilva/repos/CBL-Mariner/out/license_check/license_check_results_image_core-efi.json 
INFO[0006][licensecheck] 
Errors/warnings fall into three buckets:
        1. 'bad %doc files': A %doc documentation file that the tool believes to be a license file.
        2. 'bad general file': A file that is placed into '/usr/share/licenses/' that is not flagged as
                a license file. These files should use %license instead of %doc. Ideally whey should also
                not be placed in a directory manually. (e.g. prefer '%license COPYING' over
                '%license %{_docdir}/%{name}/COPYING')
        3. 'duplicated license files': A license file that is both a %license and a %doc file, pick one.")
                This is a warning, unless the tool is run in pedantic mode, in which case it is an error.
How to fix:
        - 'False positives': In all cases, a detection may be suppressed by using the exception file:
                /home/damcilva/repos/CBL-Mariner/toolkit/resources/manifests/package/license_file_exceptions.json.
                This file contains per-package and global exceptions in the form of regexes.
        - 'bad %%doc files': Mark it using %license, ideally without using a buildroot path (e.g. use '%license COPYING').
        - 'bad general file': Mark it using %license, ideally without using a buildroot path (e.g. use '%license COPYING').
        - 'duplicated license files': If they are actually equivalent, remove the copy in the documentation.
        - Query package contents with 'rpm -ql <package>.rpm' to see all files, 'rpm -qL <package>.rpm' to
                see only the license files, and 'rpm -qd <package>.rpm' to see only the documentation files. 
ERRO[0006][licensecheck] Found (3) packages with license errors.      
WARN[0006][licensecheck] Found (3) packages with non-fatal license issues. 
make: *** [/home/damcilva/repos/CBL-Mariner/toolkit/scripts/analysis.mk:128: license-check-img] Error 1
```

##### Change Log  

- Add `licensecheck` tool
- Add `license-check` and `license-check-img` build targets
- Add `licensecheck` package
- Add testing infrastructure to validate the tool

##### Does this affect the toolchain?  

**NO**

##### Associated issues  

- https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/50220148
- https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/50276253

##### Test Methodology

- Local testing


toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,452 @@
...
+		err = fmt.Errorf("failed to walk directory. Error:\n%w", err)
+		return nil, err
+	}
+	return foundSrpmPaths, nil
+}
+
+// queueWorkers queues up workers to search the RPMs in parallel. Each worker will wait on the jobSemaphore before starting.
+func (l *LicenseChecker) queueWorkers(rpmsToSearchPaths []string, resultsChannel chan LicenseCheckResult, cancel chan struct{}) {
+	for _, rpmPath := range rpmsToSearchPaths {
+		go func(rpmPath string) {
```
cwize1, line 333: Ideally, the goroutine shouldn't be created until a job slot has freed up.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,452 @@
...
+	"github.com/microsoft/azurelinux/toolkit/tools/pkg/simpletoolchroot"
+)
+
+const (
+	chrootName      = "license_chroot"
+	licensePrefix   = "/usr/share/licenses"
+	numParallelJobs = 20
+)
+
+var jobSemaphore = make(chan struct{}, numParallelJobs) // Limit the number of parallel jobs
```
cwize1, line 53: Please make this a non-global variable.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,452 @@
...
+	if err != nil {
+		return LicenseCheckResult{RpmPath: rpmPath, err: fmt.Errorf("failed to query package contents. Error:\n%w", err)}, nil
+	}
+
+	pkgNameLines, err := rpm.QueryPackage(rpmPath, "%{NAME}", defines)
+	if err != nil {
+		return LicenseCheckResult{RpmPath: rpmPath, err: fmt.Errorf("failed to query package. Error:\n%w", err)}, nil
+	}
+	if len(pkgNameLines) != 1 {
+		return LicenseCheckResult{RpmPath: rpmPath, err: fmt.Errorf("failed to query package. Error: expected 1 package name, got %d", len(pkgNameLines))}, nil
```
cwize1, line 373: nit: `"failed to query package:\nexpected 1 package name, got %d"`

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,452 @@
...
+
+	sort.Strings(badDocFiles)
+	sort.Strings(duplicatedDocs)
+	sort.Strings(badOtherFiles)
+
+	return badDocFiles, badOtherFiles, duplicatedDocs
+}
+
+// isDocumentInLicenseFiles checks if a document file is in the list of license files (based on basename of the file).
+func isDocumentInLicenseFiles(documentFile string, licenseFiles []string) bool {
```
cwize1, line 426: Why are you only checking this filename, instead of the full path?

toolkit/tools/pkg/licensecheck/testdata/README.md
```diff
@@ -0,0 +1,33 @@
+# Test data for the license checker
```
cwize1, line 1: Markdown files should also be human readable. So, it is goodness to line wrap (at 88).

toolkit/tools/pkg/licensecheck/testdata/generate_test_data.py
```diff
@@ -0,0 +1,139 @@
...
+# Write the results to a file.
+def write_to_file(file_list: list[(str,str)], output_file: str):
+    print(f"Writing to {output_file}")
+    file_list.sort()
+    with open(output_file, "w") as f:
+        for pkg_name, full_path_list  in file_list:
+            for full_path in full_path_list:
+                f.write(f"{pkg_name}`{full_path}\n")
+
+# Put the debug info packages first since they tend to be really big,
```
cwize1, line 96: It is generally a good idea to always use the Python's main func pattern:

```python
def main():
   ...

if __name__ == '__main__':
    main()
```

toolkit/tools/pkg/licensecheck/testdata/licensetestchecker.go
```diff
@@ -0,0 +1,126 @@
...
+		logger.Log.Fatalf("Failed to load license exceptions: %v", err)
+	}
+
+	realLicenseLines, err := file.ReadLines(*licenses)
+	if err != nil {
+		logger.Log.Fatalf("failed to read input file: %v", err)
+	}
+
+	// Each line will be of the form <pkg>`/usr/share/licenses/<pkg>/<subpath>
+	filesNotDetectedAsLicense := []string{}
```
cwize1, line 50: It might be good to split this function up a little to make it easier to read.

toolkit/tools/pkg/licensecheck/testdata/licensetestchecker.go
```diff
@@ -0,0 +1,126 @@
...
+	for _, line := range realLicenseLines {
+		splitParts := strings.SplitN(line, delimiter, 2)
+		if len(splitParts) != 2 {
+			logger.Log.Fatalf("invalid line format: %s", line)
+		}
+		pkgName, licensePath := splitParts[0], splitParts[1]
+		// Check if we consider any of the test data to not be a license file
+		if !licensecheck.IsALicenseFile(pkgName, licensePath) || exceptions.ShouldIgnoreFile(pkgName, licensePath) {
```
cwize1, line 58: `IsALicenseFile` and `ShouldIgnoreFile` seem to always be called together. It might be good to move the common code into its own function.

toolkit/tools/pkg/licensecheck/testdata/licensetestchecker.go
```diff
@@ -0,0 +1,126 @@
...
+
+	falseNegativeRatio := float64(len(filesNotDetectedAsLicense)) / float64(len(realLicenseLines))
+	err = file.WriteLines(filesNotDetectedAsLicense, *licensesOut)
+	if err != nil {
+		logger.Log.Fatalf("failed to write output file: %v", err)
+	}
+	logger.Log.Infof("Wrote %d invalid entries to '%s' (%.2f%% false negative)", len(filesNotDetectedAsLicense), *licensesOut, falseNegativeRatio*100)
+
+	// Validate false positives with docs
+	invalidDocs := []string{}
```
cwize1, line 72: The `invalidDocs` code seems almost identical to the `filesNotDetectedAsLicense`. Is it possible to extract the common logic?

toolkit/tools/pkg/licensecheck/licensenames.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+// licenseNamesFuzzy is a list of license names that should be matched in a case-insensitive sub-string search
+var licenseNamesFuzzy = []*regexp.Regexp{
+	regexp.MustCompile(`(?i).*copying.*`),
```
cwize1, line 17: Would it make sense to move these names into resource file under `toolkit/resources/manifests/package` as well?

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,452 @@
...
+const (
+	chrootName      = "license_chroot"
+	licensePrefix   = "/usr/share/licenses"
+	numParallelJobs = 20
+)
+
+var jobSemaphore = make(chan struct{}, numParallelJobs) // Limit the number of parallel jobs
+
+// LicenseCheckResult is the result of a license check on an single RPM
+type LicenseCheckResult struct {
```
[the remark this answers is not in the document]
dmcilvaney: Sure, I was on the fence about that.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Oops, old hack that I didn't clean up. Moved to the struct.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+	}
+	return all, warnings, errors
+}
+
+// FormatResults formats the results of the search to a string. Results will be ordered as follows:
+// - Packages with warnings only, sorted alphabetically
+// - Packages with errors (and possibly warnings), sorted alphabetically
+// If pedantic is true, warnings will be treated as errors.
+func (l *LicenseChecker) FormatResults(pedantic bool) string {
```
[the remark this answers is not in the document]
dmcilvaney: I'm going to be working next on integrating this into the build scheduler / image build. I had hoped to re-use it there. If it doesn't work out I'll move it to the cmdline package.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Many packages will do something like:
```
%install
cp LICENSE.txt %{buidlroot}/some/random/folder/LICENSE.txt
...
%files
%license LICENSE.txt
%doc some/random/folder/LICENSE.txt
```

I couldn't think of a better way to try and detect that duplication. It will likely lead to some false-positive hits but we can use exceptions if needed. I think in general for a  well behaved package (ie no files like LICENSE.txt in %doc) this becomes a non-issue (since there are no docs to check in the first place).

toolkit/tools/pkg/licensecheck/testdata/README.md
dmcilvaney: Ah, didn't even consider it. My editor automatically wraps markdown.

toolkit/tools/pkg/licensecheck/testdata/generate_test_data.py
dmcilvaney: Done

toolkit/tools/pkg/licensecheck/testdata/licensetestchecker.go
dmcilvaney: Refactored the data gen stuff. Did this as well as moved to .json format.

toolkit/tools/pkg/licensecheck/testdata/licensetestchecker.go
dmcilvaney: I don't really consider this tool a "real" tool, and they are only used together in one place in the normal code. Not sure its worth the extra function just for that.

toolkit/tools/pkg/licensecheck/testdata/licensetestchecker.go
dmcilvaney: done with refactor

toolkit/tools/pkg/licensecheck/licensenames.go
dmcilvaney: Not a bad idea, will do.

jslobodzian (approved):

toolkit/Makefile
```diff
@@ -77,6 +77,14 @@ ENABLE_CPU_PROFILE              ?= n
 ENABLE_MEM_PROFILE              ?= n
 ENABLE_TRACE                    ?= n
 
+# License checking tool
+##help:var:LICENSE_CHECK_DIRS:"<rpm_dir_1> <rpm_dir_2"=Space separated list of directories to recursively validate with 'license-check'. Default is RPMS_DIR.
+LICENSE_CHECK_DIRS ?= $(RPMS_DIR)
```
PawelWMS, line 82: Could you also add the documentation for the new tool into the readmes?

toolkit/resources/manifests/package/license_file_exceptions.json
```diff
@@ -0,0 +1,37 @@
...
+            "PackageName": "gcc",
+            "IgnoredFilesRegexList": [
+                "^/usr/share/man/man7/gpl.7.gz$"
+            ]
+        },
+        {
+            "PackageName": "libdb-docs",
+            "IgnoredFilesRegexList": [
+                "^/usr/share/doc/libdb-5.3.28/installation/build_unix_freebsd.html$",
+                "^/usr/share/doc/libdb-5.3.28/license/license_db.html$"
```
PawelWMS, line 19: Do we want to be tied to the 5.3.28 version?

toolkit/Makefile
```diff
@@ -77,6 +77,14 @@ ENABLE_CPU_PROFILE              ?= n
 ENABLE_MEM_PROFILE              ?= n
 ENABLE_TRACE                    ?= n
 
+# License checking tool
+##help:var:LICENSE_CHECK_DIRS:"<rpm_dir_1> <rpm_dir_2"=Space separated list of directories to recursively validate with 'license-check'. Default is RPMS_DIR.
```
PawelWMS, line 81:
```suggestion
##help:var:LICENSE_CHECK_DIRS:"<rpm_dir_1> <rpm_dir_2>"=Space separated list of directories to recursively validate with 'license-check'. Default is RPMS_DIR.
```

toolkit/scripts/analysis.mk
```diff
@@ -86,3 +89,49 @@ sodiff-check: $(BUILT_PACKAGES_FILE) | $(SODIFF_REPO_FILE)
...
+	@echo Verifying no mountpoints present in $(license_check_build_dir)
+	$(SCRIPTS_DIR)/safeunmount.sh "$(license_check_build_dir)" && \
+	rm -rf $(license_check_build_dir) && \
+	rm -rf $(license_out_dir)
+
+# If we are using the default RPMS_DIR as LICENSE_CHECK_DIRS, we are responsible for building the rpms as needed.
+# Can set REBUILD_PACKAGES=n to skip rebuilding the rpms if they are already built.
+ifeq ($(LICENSE_CHECK_DIRS),$(RPMS_DIR))
+license-check: $(LICENSE_CHECK_DIRS)
+endif
```
PawelWMS, line 113: I'm not sure about this part, because it by default ties the licensing tool to the package building functionality of the toolkit. And it's conditional in an obscure way for the user - they change `LICENSE_CHECK_DIRS` and suddenly you see different behaviour. If anything, I'd go to the other way around: conditionally add license checking to package building but let the licensing tool be its own, independent component.

toolkit/scripts/analysis.mk
```diff
@@ -86,3 +89,49 @@ sodiff-check: $(BUILT_PACKAGES_FILE) | $(SODIFF_REPO_FILE)
...
+		--exception-file="$(LICENSE_CHECK_EXCEPTION_FILE)" \
+		--name-file="$(LICENSE_CHECK_NAME_FILE)" \
+		--worker-tar="$(chroot_worker)" \
+		--build-dir="$(license_check_build_dir)" \
+		--dist-tag=$(DIST_TAG) \
+		$(if $(filter y,$(LICENSE_CHECK_PEDANTIC)),--pedantic) \
+		--results-file="$(license_results_file_pkg)"
+
+##help:target:license-check-img=Validate all packages needed for an image for license compliance. Must set CONFIG_FILE=<path_to_config>.
+license-check-img: $(go-licensecheck) $(chroot_worker) $(image_package_cache_summary) $(LICENSE_CHECK_EXCEPTION_FILE) $(LICENSE_CHECK_NAME_FILE)
```
PawelWMS, line 128: Is my understanding correct, that the `image_package_cache_summary` dependency is to force the download of the image packages? Also, what happens when you run this for two images one after another without cleaning in the meantime? Will we end up verifying a union of the RPMs in the second run?

toolkit/tools/internal/rpm/rpm.go
```diff
@@ -337,6 +337,50 @@ func QueryPackage(packageFile, queryFormat string, defines map[string]string, ex
 	return executeRpmCommand(rpmProgram, args...)
 }
 
+// QueryPackageContents queries an RPM for its file contents. The results are split into several categories:
+// - allFilesAndDirectories: all files and directories in the package
+// - files: all files in the package (ie allFilesAndDirectories minus directories)
+// - directories: all directories in the package (ie allFilesAndDirectories minus files, symlinks etc.)
+// - documentFiles: all files marked as documentation (%doc)
+// - licenseFiles: all files marked as license (%license)
+func QueryPackageContents(packageFile string, defines map[string]string,
```
PawelWMS, line 346: I think the name could be clearer. A package has many contents (metadata, scripts, files, signatures, etc.). If we're only doing files, why not `QueryPackageFiles` or `ListPackageFiles`?

toolkit/tools/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,134 @@
...
+		printExplanation()
+		logger.Log.Errorf("Found (%d) packages with license errors.", totalFailedPackages)
+		logger.Log.Warnf("Found (%d) packages with non-fatal license issues.", totalWarningPackages)
+		os.Exit(1)
+	} else if totalWarningPackages > 0 {
+		printExplanation()
+		logger.Log.Warnf("Found (%d) packages with non-fatal license issues", totalWarningPackages)
+	} else {
+		logger.Log.Infof("No license issues found")
+	}
```
PawelWMS, line 76: Instead of a comment, why not:
```suggestion
	results, failedPackages, warningPackages, err := scanDirectories(...)
	
	if *resultFile != "" {
		logger.Log.Infof("Writing results to file: %s", *resultFile)
		err := licensecheck.SaveLicenseCheckResults(*resultFile, totalResults)
		if err != nil {
			logger.Log.Fatalf("Failed to write results to file. Error: %v", err)
		}
	}

	printSummary(failedPackages, failedWarnings)
```

The results saving part seems pretty clear, since we have both the log saying what we're doing as well as we only call one function with a very clear name.

toolkit/tools/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// validateRpmDir scans the given directory for RPMs and validates their licenses. It will return all findings split into warnings and failures.
+// Each call to this function will generate a new chroot environment and clean it up after the scan.
+func validateRpmDir(buildDirPath, workerTar, rpmDir, nameFile, exceptionFile, distTag string, pedantic bool,
+) (warningResults, failedResults []licensecheck.LicenseCheckResult, err error) {
+	logger.Log.Infof("Preparing license check environment for %s...", rpmDir)
+	licenseChecker, err := licensecheck.New(buildDirPath, workerTar, rpmDir, nameFile, exceptionFile, distTag)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to initialize RPM license checker:\n%v", err)
```
PawelWMS, line 108:
```suggestion
		return nil, nil, fmt.Errorf("failed to initialize RPM license checker:\n%w", err)
```

toolkit/tools/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,134 @@
...
+	}()
+
+	logger.Log.Infof("Scanning %s for license issues...", rpmDir)
+	err = licenseChecker.CheckLicenses()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to generate license scan:\n%w", err)
+	}
+
+	resultsString := licenseChecker.FormatResults(pedantic)
+	logger.Log.Infof("Search results for %s:\n%s", rpmDir, resultsString)
```
PawelWMS, line 129: Similar ask for these logged strings.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/rpm"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/sliceutils"
+	"github.com/microsoft/azurelinux/toolkit/tools/pkg/simpletoolchroot"
+)
+
+const (
+	chrootName      = "license_chroot"
+	licensePrefix   = "/usr/share/licenses"
+	numParallelJobs = 20
```
PawelWMS, line 49: Why hard-code it to any specific value? It might be more flexible to use some multiple of `runtime.NumCPU()`.

PawelWMS, line 49: The `chrootName` and `numParallelJobs` are used only inside the same one function. No need to make them globals in such case and keeping them closer to the code would also make it easier to read.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+) (newLicenseChecker *LicenseChecker, err error) {
+	newLicenseChecker = &LicenseChecker{
+		distTag:          distTag,
+		simpleToolChroot: &simpletoolchroot.SimpleToolChroot{},
+		jobSemaphore:     make(chan struct{}, numParallelJobs),
+	}
+
+	err = newLicenseChecker.simpleToolChroot.InitializeChroot(buildDirPath, chrootName, workerTarPath, rpmDirPath)
+	if err != nil {
+		newLicenseChecker.CleanUp()
```
PawelWMS, line 79: The fact that we have to run `newLicenseChecker.CleanUp()` when `InitializeChroot` fails seems like an issue in `InitializeChroot`, which should clean-up after itself, if it fails.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+		err = fmt.Errorf("failed to initialize chroot:\n%w", err)
+		return nil, err
+	}
+
+	newLicenseChecker.licenseNames, err = LoadLicenseNames(nameFilePath)
```
PawelWMS, line 84: This looks a bit fishy: why is `LoadLicenseNames` a public function if it's only used internally here and in the tests? We can always make it public, if it makes sense to run on its own but until then, there's no need to expose it. Especially since I understand we want to use the OO model, where all interactions with the module are handled through the `LicenseChecker` objects.

toolkit/tools/pkg/licensecheck/licensenames.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return false
+}
+
+// LoadLicenseNames loads the license name regexes from the given .json file into a LicenseNames struct
+func LoadLicenseNames(file string) (LicenseNames, error) {
+	config := LicenseNames{}
+	err := jsonutils.ReadJSONFile(file, &config)
+	if err != nil {
+		return LicenseNames{}, fmt.Errorf("failed to read license names file:\n%w", err)
```
PawelWMS, line 83: Please include the file path in the error message.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+	if err != nil {
+		newLicenseChecker.CleanUp()
+		err = fmt.Errorf("failed to load license names:\n%w", err)
+		return nil, err
+	}
+
+	if exceptionFilePath != "" {
+		newLicenseChecker.exceptions, err = LoadLicenseExceptions(exceptionFilePath)
+		if err != nil {
+			newLicenseChecker.CleanUp()
```
PawelWMS, line 94: Idea to avoid repetition and potential bugs, if we ever modify this function: you can run the clean-up in a `defer` whenever `err != nil`.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+		l.simpleToolChroot = nil
+	}
+	return nil
+}
+
+// CheckLicenses will scan all .rpm files in the chroot for bad licenses. The results can be accessed with FormatResults() or GetAllResults().
+func (l *LicenseChecker) CheckLicenses() error {
+	if l.simpleToolChroot == nil {
+		return fmt.Errorf("chroot has not been initialized")
```
PawelWMS, line 118: The chroot is part of the internal implementation of the tool, which the user shouldn't really worry about nor will this particular message immediately help them. The main thing that they missed is to correctly create `LicenseChecker` with a call to `New`, so that might be a more user-friendly message.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+		return searchErr
+	})
+	if err != nil {
+		return fmt.Errorf("failed to scan for license issues:\n%w", err)
+	}
+
+	for _, result := range l.results {
+		if result.err != nil {
+			logger.Log.Errorf("Failed to scan (%s) for license issues: %v", result.RpmPath, result.err)
+		}
```
PawelWMS, line 134: Do we want to exit `CheckLicenses` without an error in this case?

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+				// Otherwise print the warning now.
+				sb.WriteString(printWarning(result))
+			}
+		}
+	}
+
+	return sb.String()
+}
+
+func printWarning(result LicenseCheckResult) string {
```
PawelWMS, line 198: Naming may be a bit confusing: we're not printing anything in here and in `printError`. How about `formatWarning` or `buildWarning`?

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+	var sb strings.Builder
+	sb.WriteString(fmt.Sprintf("WARN: '%s' has license warnings:\n", filepath.Base(result.RpmPath)))
```
PawelWMS, line 200: I think we're trying to consistently surround strings in brackets. Also, I've noticed you put the in brackets in other places in this PR, so it would be good to keep one format.

toolkit/tools/pkg/licensecheck/licensecheck.go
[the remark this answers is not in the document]
PawelWMS: +1 to @cwize1's comment. It feels like we're mixing in additional responsibilities into the `LicenseChecker`. In addition to checking the licenses, it also deals with formatting the results, making it more tightly coupled to a single use case. I think it would make the code more flexible/modular, if we moved the formatting logic to a separate element, which would for instance accept the set of results as input. This way it's also going to be easier to extend the "formatters", if we need different ones in the future.

toolkit/scripts/analysis.mk
dmcilvaney: Correct, and each image build gets its own separate cache. The only pollution you might see is if you run the same image back-to-back after changing something in the config to add/remove packages.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+	}
+	return all, warnings, errors
+}
+
+// FormatResults formats the results of the search to a string. Results will be ordered as follows:
+// - Packages with warnings only, sorted alphabetically
+// - Packages with errors (and possibly warnings), sorted alphabetically
+// If pedantic is true, warnings will be treated as errors.
+func (l *LicenseChecker) FormatResults(pedantic bool) string {
```
PawelWMS, line 170: Similar concern about `pedantic`. Making the `LicenseChecker` understand what it means, makes it more tightly coupled to its current single use, making it harder to re-use in the future. That also adds additional complexity to this module's implementation. The user of the `LicenseChecker` can do the filtering of the results on their own without influencing the implementation of the checker itself.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+			lastReportPercent = percentProcessed
+		}
+		results = append(results, result)
+	}
+	return
+}
+
+// findRpmPaths walks the chroots's mount directory to find all *.rpm files. The paths are returned relative to the
+// chroot's root.
+func (s *LicenseChecker) findRpmPaths() (foundSrpmPaths []string, err error) {
```
PawelWMS, line 266: Why `s` for the `LicenseChecker` pointer?

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+	})
+	if err != nil {
+		err = fmt.Errorf("failed to walk directory:\n%w", err)
+		return nil, err
+	}
+	return foundSrpmPaths, nil
+}
+
+// queueWorkers queues up workers to search the RPMs in parallel. Each worker will wait on the jobSemaphore before starting.
+func (l *LicenseChecker) queueWorkers(rpmsToSearchPaths []string, resultsChannel chan LicenseCheckResult, cancel chan struct{}) {
```
PawelWMS, line 290: Since we're already switching to `Context` in another PR, maybe we should replace the `cancel` channel with one while we're here?

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+			}
+			resultsChannel <- searchResult
+		}(rpmPath)
+	}
+}
+
+// CheckRpmLicenses checks the licenses of an RPM at the given path. It returns result struct holding all the license
+// issues found. This function will use the host's macros to query the RPM so it is expected to be called in a chroot.
+// - rpmPath: The path to the RPM to check relative to the chroot's root.
+func CheckRpmLicenses(rpmPath, distTag string, licenseNames LicenseNames, exceptions LicenseExceptions) (result LicenseCheckResult, err error) {
```
PawelWMS, line 319: Why a public function? This makes this module be in a "split" position, where some functionality is exposed through the `LicenseChecker` struct and some through separate public functions. It seems like `CheckRpmLicenses` is only used internally by the `LicenseChecker` anyway, so I'd suggest keeping it private at least for now. If we learn in the future, that it's useful to expose this functionality publicly, that's going to be easy to do. Doing it upfront now may cause unnecessary confusion about how to use this module.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+	if err != nil {
+		err = fmt.Errorf("failed to walk directory:\n%w", err)
+		return nil, err
+	}
+	return foundSrpmPaths, nil
+}
+
+// queueWorkers queues up workers to search the RPMs in parallel. Each worker will wait on the jobSemaphore before starting.
+func (l *LicenseChecker) queueWorkers(rpmsToSearchPaths []string, resultsChannel chan LicenseCheckResult, cancel chan struct{}) {
+	for _, rpmPath := range rpmsToSearchPaths {
```
PawelWMS, line 291: Mostly curious: why spin up and tear down a new worker for each path instead of having X workers grabbing paths from a channel and waiting for them to finish? Since we're working with files, I don't think there's going to be any performance impact, thus only curious. Was the code cleaner?

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+
+			logger.Log.Debugf("Searching (%s)", filepath.Base(rpmPath))
+			searchResult, err := CheckRpmLicenses(rpmPath, l.distTag, l.licenseNames, l.exceptions)
+			logger.Log.Debugf("Finished searching (%s)", filepath.Base(rpmPath))
+			if err != nil {
+				logger.Log.Errorf("License check worker failed with error: %v", err)
+				resultsChannel <- LicenseCheckResult{err: err}
+				return
+			}
+			resultsChannel <- searchResult
```
PawelWMS, line 311: There's something fishy about the `LicenseCheckResult` and `CheckRpmLicenses` here. The `LicenseCheckResult` seems to combine two responsibilities:
- Holding the results of the check.
- Holding the error of the check.

This also forces us to ignore the `err` field during JSON serialization of `LicenseCheckResult`.

I'd suggest splitting the data from the error with a wrapper struct for `LicenseCheckResult` and `Error`, which can then be constructed here and sent through `resultsChannel`.

toolkit/scripts/analysis.mk
```diff
@@ -86,3 +89,49 @@ sodiff-check: $(BUILT_PACKAGES_FILE) | $(SODIFF_REPO_FILE)
...
+# If we are using the default RPMS_DIR as LICENSE_CHECK_DIRS, we are responsible for building the rpms as needed.
+# Can set REBUILD_PACKAGES=n to skip rebuilding the rpms if they are already built.
+ifeq ($(LICENSE_CHECK_DIRS),$(RPMS_DIR))
+license-check: $(LICENSE_CHECK_DIRS)
+endif
+
+##help:target:license-check=Validate all packages in the RPMS_DIR for license compliance. Set LICENSE_CHECK_DIRS to override source.
+license-check: $(go-licensecheck) $(chroot_worker) $(LICENSE_CHECK_EXCEPTION_FILE) $(LICENSE_CHECK_NAME_FILE)
+	$(go-licensecheck) \
+		$(foreach license_dir, $(LICENSE_CHECK_DIRS),--rpm-dirs="$(license_dir)" ) \
```
neha170, line 118: Can we shift the logic of handling the list into the tool
`$(foreach license_dir, $(LICENSE_CHECK_DIRS),--rpm-dirs="$(license_dir)" )`
we just pass the entire list of license_check_dirs to the tool.
This will help us with our goal of making Makefiles leaner wrt logic

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+	if pedantic && len(result.DuplicatedDocs) > 0 {
+		sb.WriteString(fmt.Sprintf("\tduplicated license files:\n\t\t%s\n", strings.Join(result.DuplicatedDocs, "\n\t\t")))
+	}
+	return sb.String()
+}
+
+// runLicenseCheckInChroot searches for bad licenses amongst the RPMs mounted into the chroot. This function is meant
+// to be called from inside the chroot's context.
+func (l *LicenseChecker) runLicenseCheckInChroot() (results []LicenseCheckResult, err error) {
+	const searchReportIntervalPercent = 10
```
neha170, line 223: would be nice to have a comment on what this const is

toolkit/tools/pkg/licensecheck/licensecheck.go
neha170, line 266: Not sure why we have named this as `foundSrpmPaths`. Could be `rpmPaths`

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+
+	// Find all the rpms in the chroot
+	rpmsToSearchPaths, err := l.findRpmPaths()
+	if err != nil {
+		return nil, fmt.Errorf("failed to walk srpm directory:\n%w", err)
```
neha170, line 228: Why are we saying srpm?

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+		logger.Log.Warnf("No rpms found in %s", l.simpleToolChroot.ChrootRelativeMountDir())
+		return nil, nil
+	}
+
+	// Scan each rpm in parallel
+	resultsChannel := make(chan LicenseCheckResult, len(rpmsToSearchPaths))
+	cancel := make(chan struct{})
+	logger.Log.Infof("Queuing %d rpms for search", len(rpmsToSearchPaths))
+	l.queueWorkers(rpmsToSearchPaths, resultsChannel, cancel)
+	logger.Log.Infof("Searching rpms...")
```
neha170, line 240: nit: can get rid of trailing `...`

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,411 @@
...
+	for range rpmsToSearchPaths {
+		result := <-resultsChannel
+		if result.err != nil {
+			// Signal the workers to stop if there is an error
+			err = fmt.Errorf("failed to search srpm:\n%w", result.err)
+			close(cancel)
+			return
+		}
+		numProcessed++
+		percentProcessed := (numProcessed * 100) / len(rpmsToSearchPaths)
```
neha170, line 254: nit: would be nice to have a comment about what we're calculating here and why

neha170 (commented):
I think we could get rid of this 
`sudo make license-check SRPM_PACK_LIST="pkg1 pkg2 pkg3" will build pgk, pkg2, pgk3 and then validate their licenses`
as it mixes two things, building packages and validating licenses.
If we get rid of the building packages part, we will only be validating licenses for rpms already built. Can we reuse the same chroot for all rpms that need to be validated? Might save us time and resources in creating/tearing down chroots

toolkit/resources/manifests/package/license_file_exceptions.json
```diff
@@ -0,0 +1,37 @@
...
+    "_comment2": "Each PkgException entry is a {'PackageName' and 'IgnoredFilesRegexList'}",
+    "_comment3": "The 'PackageName' is the name of the (sub)package to which the exception applies.",
+    "_comment4": "The 'IgnoredFilesRegexList' is a list of regular expressions that match files to be omitted from the license check.",
+    "_comment5": "The 'GlobalExceptionsRegexList' is a list of regular expressions that match all packages.",
+
+    "PkgExceptions": [
+        {
+            "PackageName": "gcc",
+            "IgnoredFilesRegexList": [
+                "^/usr/share/man/man7/gpl.7.gz$"
```
Camelron, line 12: I struggle to understand how these files would be caught up in the regexes listed in FuzzyLicenseNamesRegexList, etc. from license_file_names.json?

toolkit/Makefile
dmcilvaney: Added to `1_initial_prep.md` with the other tools.
I expect most people will interact with this feature through the integration with the other build tools (scheduler, image fetcher?).

toolkit/resources/manifests/package/license_file_exceptions.json
dmcilvaney: I was initially going to to explicit, but thinking back I agree that is probably going to be an ongoing pain to stay on top of. I'll replace with `[0-9\\.]+` which should match the version numbers.

toolkit/resources/manifests/package/license_file_exceptions.json
dmcilvaney: `(?i).*gpl.*` is the one flagging this file.

toolkit/resources/manifests/package/license_file_exceptions.json
dmcilvaney: Also realized that the `.` in the paths should be `\\.`.

toolkit/scripts/analysis.mk
dmcilvaney: I think it's important to be consistent, and our other tools can automatically call dependencies when it makes sense. I'll add a 3rd target that doesn't do so if that is a sticking point.

toolkit/tools/internal/rpm/rpm.go
dmcilvaney: `QueryPackageFiles()`

toolkit/tools/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,134 @@
...
+) (warningResults, failedResults []licensecheck.LicenseCheckResult, err error) {
+	logger.Log.Infof("Preparing license check environment for %s...", rpmDir)
+	licenseChecker, err := licensecheck.New(buildDirPath, workerTar, rpmDir, nameFile, exceptionFile, distTag)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to initialize RPM license checker:\n%v", err)
+	}
+	defer func() {
+		cleanupErr := licenseChecker.CleanUp()
+		if cleanupErr != nil {
+			if err == nil {
```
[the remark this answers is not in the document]
dmcilvaney: I think that muddles things a bit, this code path can run even if the main tool didn't return an error. I've gone with
```go
		if cleanupErr != nil {
			if err == nil {
				err = fmt.Errorf("failed to cleanup after RPM license checker:\n%w", cleanupErr)
			} else {
				// Append the cleanup error to the existing error
				err = fmt.Errorf("%w\nfailed to cleanup after RPM license checker failed:\n%w", err, cleanupErr)
			}
		}
```

toolkit/tools/licensecheck/licensecheck.go
dmcilvaney: Done

toolkit/scripts/analysis.mk
dmcilvaney: I think that is the wrong direction here. The argument parsing library we use has explicit code to take lists of directories etc. I think being scared of basic makefile functionality is the wrong way to go about this. We should try to minimize the complex logic, but in this case we should leverage the error checking intrinsic in the arg parsing library.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Our other tools just set a reasonable value for similar things (although they do let you override that which I didn't bother with). This is a bit different though, since it is a library... I'll have it always use `NumCPU()*2` which should be an ok compromise.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Probably just being overly defensive. One of those spots where it feels like we should err on the side of caution and always clean up after ourselves even if in this specific case it may not be needed today.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Its used in the test program `licensetestchecker.go`. If we really want to I can write a wrapper around it and `LoacLicenseExceptions, but I don't see the harm in having a public factory/constructor function for the `LicenseNames` and `LicenseExceptions` types.

(I had a bug in the test program that was using the old pattern to call `IsALicenseFile()`, fixed now.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Done

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: This is dead code, `l.runLicenseCheckInChroot()` will return an error if any worker does, so this can just go.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Copy/paste of code from my source code search tool, thought I got them all

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: My personal take, it makes the code much easier to read. go routines are super lightweight so there isn't really an issue with spinning them all up right away.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Will do

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Added a new private type
```
type licenseCheckReturn struct {
	finding LicenseCheckResult
	err     error
}
```

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: This will be needed by the build workers, this is just part one of the changes. The build worker has no need to stand up a full chroot object, it just needs to scan some files.

Go packaging best practices is to include all related functionality in a single package, don't over split.

dmcilvaney:
> I think we could get rid of this `sudo make license-check SRPM_PACK_LIST="pkg1 pkg2 pkg3" will build pgk, pkg2, pgk3 and then validate their licenses` as it mixes two things, building packages and validating licenses. If we get rid of the building packages part, we will only be validating licenses for rpms already built. Can we reuse the same chroot for all rpms that need to be validated? Might save us time and resources in creating/tearing down chroots

The idea is to be able to scan a list of packages (via `SRPM_PACK_LIST` since that controls that behavior in the tools). The scanner tool only ever uses a single chroot.

Phase two of this tool is to integrate it into the package build (and image build). That's the reason I exposed some of the basic functions as public. The package build chroot can call the functions on each `*.rpm` it has built without needing to spin up a dedicated chroot.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Having used the API for all its intended purposes, using an existing chroot was too much trouble, easier to just use the dedicated chroot using `New()`. Will make this private.

toolkit/Makefile
```diff
@@ -77,6 +77,14 @@ ENABLE_CPU_PROFILE              ?= n
 ENABLE_MEM_PROFILE              ?= n
 ENABLE_TRACE                    ?= n
 
+# License checking tool
+##help:var:LICENSE_CHECK_DIRS:"<rpm_dir_1> <rpm_dir_2>"=Space separated list of directories to recursively validate with the manual 'license-check' target.
+LICENSE_CHECK_DIRS ?=
+LICENSE_CHECK_EXCEPTION_FILE ?= $(MANIFESTS_DIR)/package/license_file_exceptions.json
+LICENSE_CHECK_NAME_FILE      ?= $(MANIFESTS_DIR)/package/license_file_names.json
+##help:var:LICENSE_CHECK_MODE:{none,warn,fatal,pedantic}=Set the license check mode during package and image builds. 'none' will disable the license check, 'warn' will print warnings, 'fatal' will stop the build on errors, 'pedantic' will stop the build on warnings and errors.
+LICENSE_CHECK_MODE ?= warn
```
dmcilvaney, line 86:
```suggestion
LICENSE_CHECK_MODE ?= none
```

toolkit/tools/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,172 @@
...
+	summaryFile = app.Flag("summary-file", "File to save the license check summary to.").String()
+)
+
+func main() {
+	app.Version(exe.ToolkitVersion)
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+	logger.InitBestEffort(logFlags)
+
+	mode := licensecheck.LicenseCheckMode(*mode)
+	if !licensecheck.IsValidLicenseCheckMode(mode) {
```
cwize1, line 46: `kingpin` has a `.Enum()` function you could use on the `Flag` if you were so inclined.

toolkit/tools/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,172 @@
...
+		logger.Log.Infof("Writing results to file (%s)", *resultFile)
+		err := licensecheck.SaveLicenseCheckResults(*resultFile, results)
+		if err != nil {
+			logger.Log.Fatalf("Failed to write results to file:\n%v", err)
+		}
+	}
+
+	if *summaryFile != "" {
+		logger.Log.Infof("Writing summary to file (%s)", *summaryFile)
+		resultsString := licensecheckformat.FormatResults(results, mode)
```
cwize1, line 64: `licensecheckformat.FormatResults` doesn't have the word summary in it, even though it seems it is creating a summary.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,359 @@
...
+			return fmt.Errorf("failed to cleanup chroot:\n%w", err)
+		}
+		l.simpleToolChroot = nil
+	}
+	return nil
+}
+
+// CheckLicenses will scan all .rpm files in the chroot for bad licenses. New unfiltered results will be returned but
+// also appended to the internal results list which can be accessed with GetResults().
+func (l *LicenseChecker) CheckLicenses(quiet bool) (latestResults []LicenseCheckResult, err error) {
```
cwize1, line 123: Does anything ever call `CheckLicenses` with `quiet` set to `true`?

cwize1 (approved):

toolkit/tools/licensecheck/licensecheck.go
dmcilvaney: Hmm, not sure why I added this, probably force of habit. I'm already using the enum to validate this.

toolkit/tools/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,172 @@
...
+
+func main() {
+	app.Version(exe.ToolkitVersion)
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+	logger.InitBestEffort(logFlags)
+
+	mode := licensecheck.LicenseCheckMode(*mode)
+	if !licensecheck.IsValidLicenseCheckMode(mode) {
+		logger.Log.Fatalf("Invalid license check mode (%s)", mode)
+	}
```
dmcilvaney, line 48:
```suggestion
```

toolkit/tools/licensecheck/licensecheck.go
dmcilvaney: I was 50/50 on including the big text blob, but decided to just have the basic formatted text. Maybe summary isn't the best word for it, but its more readable than json.

toolkit/tools/pkg/licensecheck/licensecheck.go
dmcilvaney: Follow on PR uses it as part of the package builds. I don't want it polluting the log output in that use-case when we already have an in-depth results output.

PR #9068, opened 2024-05-09
Image Customizer: Handle grub-mkconfig images.

cwize1:
Adds support for handling images that use `grub-mkconfig`, particularly Azure Linux 3.0. For the most part, this means that grub is configured using the `/etc/default/grub` file instead of by directly modifying the `grub.cfg` file.

This change blocks ISO output and verity support for `grub-mkconfig` images, since those features need more advanced `grub.cfg` changes that aren't immediately obvious how to implement using `/etc/default/grub`.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology


toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
```diff
@@ -314,105 +345,57 @@ func replaceKernelCommandLineArgValue(inputGrubCfgContent string, name string, v
...
-	logger.Log.Infof("Updating SELinux kernel command-line args")
-
-	grub2Config, err := readGrub2ConfigFile(imageChroot)
-	if err != nil {
-		return err
-	}
-
-	newGrub2Config, err := updateSELinuxCommandLineHelper(grub2Config, selinuxMode)
+// Finds all the kernel command-line args that match the provided names, then insert replacement arg(s).
+func updateKernelCommandLineArgs(grub2Config string, argsToRemove []string, newArgs []string) (string, error) {
```
gmileka, line 349: - nit: would be good to doc the output
- nit: I wonder if it makes sense to use the same pattern with other function that modify content:
   - `modifySomething(inputGrubConfig string,...) (modifiedGrubConfig string,...)`

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -466,34 +472,80 @@ func handleBootLoader(baseConfigPath string, config *imagecustomizerapi.Config,
 	return nil
 }
 
+// Inserts new kernel command-line args into the grub config file.
+func addKernelCommandLine(kernelExtraArguments imagecustomizerapi.KernelExtraArguments,
+	imageChroot *safechroot.Chroot,
+) error {
+	var err error
+
+	extraCommandLine := strings.TrimSpace(string(kernelExtraArguments))
```
gmileka, line 481: nit: this can be part of `AddKernelCommandLine()`

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) Microsoft Corporation.
```
gmileka, line 1: would it make sense to move `bootcustomizer.go`, `customizeboot.go`, and `defaultgrubutils.go` into their own packages so that the code is separated by the public interface?
I think it will be a useful package for tools other than the image customizer...

cwize1: Possibly at some point. But it might be worth doing that after ISO and verity is fixed, so we have a good idea of exactly what the public API should be.

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
cwize1: I also added to `AddKernelCommandLine`. I kept this one as it is useful to avoid the `"Setting KernelCommandLine.ExtraCommandLine"` log message.

toolkit/tools/pkg/imagecustomizerlib/customizeboot.go
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/defaultgrubutils.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	// Rewrite GRUB_CMDLINE_LINUX line.
+	defaultGrubContent = rewriteDefaultGrubLine(defaultGrubContent, cmdLineVarAssign, value)
+	return defaultGrubContent, nil
+}
+
+func rewriteDefaultGrubLine(defaultGrubContent string, varAssign grubDefaultVarAssign, newValue string) string {
```
gmileka, line 232: would it make sense to rename this to replace token? The name is a bit hard to understand

toolkit/tools/pkg/imagecustomizerlib/defaultgrubutils.go
```diff
@@ -0,0 +1,256 @@
...
+	args, err := parseCommandLineArgs(grubTokens)
+	if err != nil {
+		err = fmt.Errorf("failed to parse %s's value args:\n%w", varName, err)
+		return grubDefaultVarAssign{}, nil, 0, err
+	}
+
+	return cmdLineVarAssign, args, insertAt, nil
+}
+
+func addExtraCommandLineToDefaultGrubFile(defaultGrubContent string, extraCommandLine string) (string, error) {
```
gmileka, line 179: would it make sense to group functions that manipulate the default grub file in a type by itself? The naming is a bit tricky across various functions given that the default can mean the default grub file, or a default property/value. It might be a bit clearer that I use `defaultGrub.DoSomething()` `grubCfg.DoSomething()` to distinguish which file I'm targeting. Or, abstract that for the user so they do not need to know the difference (if possible)...[I need to read about who the files interact]

cwize1: The `BootCustomizer` type is intended to be the abstraction point between the two files. Though, it doesn't cover all scenarios yet.

I changed the naming so that it is "defaultGrubFile" instead of only "defaultGrub".

toolkit/tools/pkg/imagecustomizerlib/defaultgrubutils.go
```diff
@@ -0,0 +1,256 @@
...
+			return varAssign, nil
+		}
+	}
+
+	err := fmt.Errorf("failed to find %s variable assignment (%s)", installutils.GrubDefFile, name)
+	return grubDefaultVarAssign{}, err
+}
+
+func getGrubDefaultLinuxArgs(defaultGrubContent string, varName string,
+) (grubDefaultVarAssign, []grubConfigLinuxArg, int, error) {
```
gmileka, line 133: What would grubDefaultVarAssign contain?
The entire: `GRUB_CMDLINE_LINUX_DEFAULT=...` line?

gmileka: If the function is to return/find the linux arguments (from the name), why does it take a varName?
I think there's ambiguity re. varName - is this `GRUB_CMDLINE_LINUX_DEFAULT` or a kernel command line argument argument?

cwize1: Alas, there are two variables in the `/etc/default/grub` file that contains command line args: `GRUB_CMDLINE_LINUX_DEFAULT` and `GRUB_CMDLINE_LINUX`. `varName` specifies which variable to look at.

cwize1: Yes, the `grubDefaultVarAssign` value contains details for the `GRUB_CMDLINE_LINUX_DEFAULT=...` line.

The `grubDefaultVarAssign.Token` field contains the start and end locations of the line in the `/etc/default/grub` file.

gmileka (approved):

PR #9106, opened 2024-05-15
Fix PTests for Python Packages Initially Missing iniconfig

sameluch:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

A smattering of tests were failing due to a missing python module iniconfig. This PR fixes those tests

###### Change Log  

- python-html5libs, add pip install iniconfig for check section, add patch to fix test warnings causing failures
- python-importlib-metadata, add pip install iniconfig for check section
- python-mpmath, add pip install iniconfig for check section
- python-pluggy, add pip install iniconfig for check section
- python-rich, add pip install iniconfig for check section, add patch to fix test warning, add patch to skip tests requiring new python-pygments dependency
- python-wcwidth, add pip install iniconfig for check section
- python-websocket-client, add pip install iniconfig for check section
- python-wheel, add pip install iniconfig for check section

**YES**

###### Test Methodology

- Full Pipeline Run (toolchain build): https://dev.azure.com/mariner-org/mariner/_build/results?buildId=570165&view=results
- Buddy Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=571074&view=results


SPECS/python-html5lib/python-html5lib.spec
```diff
@@ -59,14 +61,17 @@ sed -i 's/from mock import/from unittest.mock import/' html5lib/tests/test_meta.
 %pyproject_save_files html5lib
 
 %check
-pip3 install more-itertools umsgpack webencodings
+pip3 install more-itertools umsgpack webencodings iniconfig
```
PawelWMS, line 64: Could you also try replacing installing `more-itertools` through pip by adding `BuildRequires: python3-more-itertools`?

SPECS/python-rich/0001-Skip-tests-for-mariner.patch
```diff
@@ -0,0 +1,176 @@
...
+ 
+ from rich.__main__ import make_test_card
+ from rich.console import Console, RenderableType
+ 
+ from ._card_render import expected
+ 
++if sys.version_info >= (3, 8):
++    from importlib.metadata import Distribution
++else:
++    from importlib_metadata import Distribution
```
PawelWMS, line 31: Is `else` even possible for Azure Linux 3.0 and up?

SPECS/python-rich/0001-Skip-tests-for-mariner.patch
```diff
@@ -0,0 +1,176 @@
...
++
++PYGMENTS_VERSION = Distribution.from_name("pygments").version
++OLD_PYGMENTS = PYGMENTS_VERSION == "2.5.2"
```
PawelWMS, line 34: I'm guessing we won't go down with `python-pygments` but does this mean that updating to `python-pygments` version 2.5.3 or 2.6.0 will already fix this issue? And if so, can we try the upgrade instead of the patch? Patches, which don't have a good chance of disappearing with version upgrades (like CVE patches for instance) add an additional maintenance burden, since they tend to break and require updates themselves.

SPECS/python-rich/python-rich.spec
```diff
@@ -10,7 +10,12 @@ Source0:        %{url}/archive/v%{version}/rich-%{version}.tar.gz
 
 BuildArch:      noarch
 
-Patch0:          3229.patch
+Patch0:         3229.patch
+Patch1:         ptest-warning.patch
+# This patch modifies the tests to set the OLD_PYGMENTS version
+# to the current one in Azure Linux, 2.5.2. Once python-pygments
+# version is updated >= 2.14.0, this patch can be removed.
```
PawelWMS, line 17: The patch targets only the 2.5.2 version. If we ever update to 2.6.0 for instance, the test will get re-enabled, while they will still fail, if I understand the comment correctly.

SPECS/python-rich/python-rich.spec
```diff
@@ -10,7 +10,12 @@ Source0:        %{url}/archive/v%{version}/rich-%{version}.tar.gz
 
 BuildArch:      noarch
 
-Patch0:          3229.patch
+Patch0:         3229.patch
+Patch1:         ptest-warning.patch
```
PawelWMS, line 14: Could you add more comments about this patch? Either here or in the patch itself?

SPECS/python-rich/python-rich.spec
```diff
@@ -51,6 +59,7 @@ code, tracebacks, and more — out of the box.
 
 %check
 # add below to make sure initial build will catch runtime import errors
+pip3 install iniconfig pygments
```
PawelWMS, line 62: I'm confused. If we're grabbing the latest `pygments` here, what is the 0001 patch for?

SPECS/python-rich/0001-Skip-tests-for-mariner.patch
sameluch: I talked with Andrew about upgrading the package. The required upgrade is to >=2.14.0. I started down the avenue of upgrading to latest (2.18.0) and found the package needed many additional dependencies as a part of the upgrade. While this is not normally an issue, since 'python-pygments' is a part of the toolchain, these dependencies would also need to be added as such there. Instead we decided it is best to simply skip these tests for the time being as they are mainly just comparing loaded style options from the library anyway.

I left a comment in the spec file detailing this patch's removal after such an upgrade.

SPECS/python-rich/python-rich.spec
sameluch: That is correct, in this regard I took the existing logic from `tests/test_syntax.py` and added the logic to additional files and modified the checked version for the one present in azure linux 3.0 (the original check only targeted 2.13.0, hence the tests failing for us in the first place).

In the event that we do upgrade to a version of `python-pygments` lower than 2.14.0, this patch would need to be modified in the future.

SPECS/python-rich/python-rich.spec
sameluch: This was here for testing, I will remove. Good catch!

SPECS/python-rich/python-rich.spec
sameluch: Sure I will add a comment to the spec

SPECS/python-html5lib/python-html5lib.spec
sameluch: Sure, new buddy build started here with changes https://dev.azure.com/mariner-org/mariner/_build/results?buildId=570210&view=results

SPECS/python-rich/0001-Skip-tests-for-mariner.patch
sameluch: uhh likely not, I can remove these instances. This is merely a copy paste error from other code within python-rich.

sameluch: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=570211&view=results - new build running with updated patch with else clauses removed.

SPECS/python-rich/python-rich.spec
sameluch: removed

SPECS/python-rich/python-rich.spec
sameluch: added.

SPECS/python-rich/python-rich.spec
sameluch: Links to the code in question and commit which changed these tests accordingly:
https://github.com/Textualize/rich/blob/master/tests/test_syntax.py#L29
https://github.com/Textualize/rich/pull/3139/files

SPECS/python-html5lib/python-html5lib.spec
sameluch: Done for both `pluggy` and `more-itertools` for all modified specs.

SPECS/python-rich/python-rich.spec
PawelWMS: Since we're interested in pre-2.14.0 versions couldn't we compare if `pygments` is < 2.14.0 now, so we don't have to worry about this until we reach 2.14.0 or higher? This looks like a promising way to [compare versions in Python](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python) you could try. It may require an `BuildRequires: python3-packaging` but I think that shouldn't be a problem - if it's inside the `0%{?with_check}` section, then there's no risk of a circular dependency as well.

sameluch: Updated to use the Version changes as detailed. Replace the comparison with a '<=' 2.13.0 as that is the targeted version from rich to start. Seems to work as expected.

buddy build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=571115&view=results

PawelWMS (approved):

PR #9137, opened 2024-05-16
Allow a custom path to be specified for custom snapshot for the pre-cacher

ellie-di:
Allow a custom path to be specified for the rpm snapshot summary file.

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

`make rpm_snapshot ...` generates the RPMs snapshot under a specific folder, with a hard coded name (rpms_snapshot.json) and `make pre-cache ...` refers to that generated snapshot file without any consideration for a custom file path/name. This is an inconvenience for builds that run on different platform architecture, that we often prefer to have the arch as part of the file name.
This PR allows setting PRECACHER_SNAPSHOT parameter for `make pre-cache`.

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Test Methodology

Tested in Azure Linux derivative image, where rpms snapshot is implemented and pre-cacher uses a custom path for the snapshot.
- Pipeline build id: xxxx

### DIFF
--- a/toolkit/scripts/precache.mk
+++ b/toolkit/scripts/precache.mk
@@ -26,13 +26,13 @@ clean-precache:
 # it does, so add the phony target as a dependency to the flag file
 .PHONY: precache_always_run_phony
 pre-cache: $(STATUS_FLAGS_DIR)/precache.flag
-$(STATUS_FLAGS_DIR)/precache.flag: $(go-precacher) $(chroot_worker) $(rpms_snapshot) precache_always_run_phony 
+$(STATUS_FLAGS_DIR)/precache.flag: $(go-precacher) $(chroot_worker) $(PRECACHER_SNAPSHOT) precache_always_run_phony
 	@if [ "$(DISABLE_UPSTREAM_REPOS)" = "y" ]; then \
 		echo "ERROR: Upstream repos are disabled (DISABLE_UPSTREAM_REPOS=y), cannot precache RPMs"; \
 		exit 1; \
 	fi
 	$(go-precacher) \
-		--snapshot "$(rpms_snapshot)" \
+		--snapshot "$(PRECACHER_SNAPSHOT)" \
 		--output-dir "$(remote_rpms_cache_dir)" \
 		--output-summary-file "$(precache_downloaded_files)" \
 		--repo-urls-file "$(repo_urls_file)" \
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -121,6 +121,8 @@ RPMS_DIR        ?= $(OUT_DIR)/RPMS
 SRPMS_DIR       ?= $(OUT_DIR)/SRPMS
 IMAGES_DIR      ?= $(OUT_DIR)/images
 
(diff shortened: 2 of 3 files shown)


neha170 (approved):
Looks good.
Could you add a line on `CUSTOM_PRECACHE_FILE` to the docs, https://github.com/microsoft/azurelinux/blob/3.0-dev/toolkit/docs/building/building.md
Thanks!

toolkit/scripts/precache.mk
```diff
@@ -26,13 +26,15 @@ clean-precache:
 # it does, so add the phony target as a dependency to the flag file
 .PHONY: precache_always_run_phony
 pre-cache: $(STATUS_FLAGS_DIR)/precache.flag
-$(STATUS_FLAGS_DIR)/precache.flag: $(go-precacher) $(chroot_worker) $(rpms_snapshot) precache_always_run_phony 
+# If we set CUSTOM_PRECACHE_FILE, we want to use that instead of the default rpms_snapshot
```
PawelWMS, line 29: Trying to keep the toolkit PRs cleaner, because we've noticed we later come back to them to get more context and miss information about what we were exactly doing:
- Could you clarify the PR's title? I think this change is more about providing a custom snapshot for the pre-cacher and not a summary file in general (which are mainly used for reproducible builds and only re-used for the pre-cacher).
- Could you also extend the summary a bit? Say what this change enabled. which scenarios it's meant to be used for, etc.?

toolkit/scripts/precache.mk
```diff
@@ -26,13 +26,15 @@ clean-precache:
...
+precache_input_snapshot = $(if $(CUSTOM_PRECACHE_FILE),$(CUSTOM_PRECACHE_FILE),$(rpms_snapshot))
```
PawelWMS, line 30: Please move it to `Makefile` and set it to a default value of `$(rpms_snapshot)`. We keep other input arguments modifiable by the user in that file. Afterwards we could also drop the "CUSTOM_" prefix as it's always going to be used with an option to be modified by the user. I'd maybe re-name it something along the lines of "PRECACHER_SNAPSHOT", so it's clearer what kind of file this may be.

Example:

```
PRECACHER_SNAPSHOT ?= $(rpms_snapshot)
```

The `precache_input_snapshot ` variable should be replacable with the new variable with that change as well.

PawelWMS (changes requested):
A few requests before we can pull this in.

toolkit/scripts/precache.mk
ellie-di: @PawelWMS Move to which `Makefile`? I see $(rpms_snapshot) is first defined in toolkit.mk and its default value depends on a bunch of other variables.

toolkit/scripts/precache.mk
ellie-di: Ok

PawelWMS (approved):

toolkit/scripts/precache.mk
PawelWMS: Thank you!

PR #9161, opened 2024-05-20
Saulparedes/lsg release

Redent0r:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
...
+%global clh_name cloud-hypervisor
+%global clh_type cvm
+
+Name:           %{clh_name}-%{clh_type}
+Summary:        Cloud Hypervisor CVM is an open source Virtual Machine Monitor (VMM) that enables running SEV SNP enabled VMs on top of MSHV using the IGVM file format as payload.
+Version:        %{gitver}.%{gitnum}
+Release:        1000.%{gitrev}%{?dist}
+License:        ASL 2.0 or BSD-3-clause
+Group:          Applications/System
+URL:            https://github.com/cloud-hypervisor/cloud-hypervisor
```
ms-mahuber, line 25: are we sure we don't want to reference the microsoft fork link?

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
...
+Source0:        https://github.com/microsoft/cloud-hypervisor/archive/refs/tags/msft/v%{version}.tar.gz#/%{clh_name}-%{version}.tar.gz
```
ms-mahuber, line 26: Our sources may collide with proper cloud-hypervisor. Any chance we can still rename the tar (and vendor) file to contain clh_type?

SPECS/cloud-hypervisor-cvm/update_src
```diff
@@ -0,0 +1,72 @@
+#/bin/bash
```
ms-mahuber, line 1: is this file required?

SPECS/cloud-hypervisor-cvm/fetch-private-deps
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
ms-mahuber, line 1: is this file required?

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
+# This spec file assumes you're building on an environment where:
```
ms-mahuber, line 1: We should remove those comments and in generally align with cloud-hypervisor where possible

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
...
+# * You have rustup installed on your system
+# * You have both x86_64-unknown-linux-gnu and x86_64-unknown-linux-musl, or
+# * You have both aarch64-unknown-linux-gnu and aarch64-unknown-linux-musl
+#   targets installed.
+
+%define using_rustup 0
+%define using_musl_libc 0
+%define using_vendored_crates 1
+
+# Generated by update script
```
ms-mahuber, line 12: We should remove this comment. Why do we need gitver, gitnum, gitrev variables? Those are only used once below.

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
...
+%global gitver   38.0
+%global gitnum   72
+%global gitrev   8f7209bdc
+%global clh_name cloud-hypervisor
```
ms-mahuber, line 16: We can get rid of those two variables if we can rename our tar files to avoid ambiguity with the proper CLH tar files

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
...
+%if 0%{?using_vendored_crates}
+%define cargo_offline --offline
+%endif
+
+%description
+Cloud Hypervisor is an open source Virtual Machine Monitor (VMM) that runs on top of KVM. The project focuses on exclusively running modern, cloud workloads, on top of a limited set of hardware architectures and platforms. Cloud workloads refers to those that are usually run by customers inside a cloud provider. For our purposes this means modern Linux* distributions with most I/O handled by paravirtualised devices (i.e. virtio), no requirement for legacy devices and recent CPUs and KVM.
+
+%prep
+
+%autosetup -p1 -n %{clh_name}-%{version}
```
ms-mahuber, line 79: in the regular CLH package we use %setup. Best to align. In general we can align this block a bit more with its regular CLH counterpart.

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
...
+pushd %{_builddir}/%{clh_name}-%{version}
+%if 0%{?using_vendored_crates}
+tar xf %{SOURCE1}
+mkdir -p .cargo
+cp %{SOURCE2} .cargo/config.toml
+%endif
+popd
+
+%define ch cloud-hypervisor
```
ms-mahuber, line 88: Is this %define really required?

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
...
+
+%install
+rm -rf %{buildroot}
+install -d %{buildroot}%{_bindir}
+install -D -m755  ./target/%{rust_def_target}/release/%{ch} %{buildroot}%{_bindir}
+
+%if 0%{?using_musl_libc}
```
ms-mahuber, line 95: vhost_user_net, vhost_user_blk is not required here?

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
...
+cp %{SOURCE2} .cargo/config.toml
+%endif
+popd
+
+%define ch cloud-hypervisor
+
+%install
+rm -rf %{buildroot}
+install -d %{buildroot}%{_bindir}
+install -D -m755  ./target/%{rust_def_target}/release/%{ch} %{buildroot}%{_bindir}
```
ms-mahuber, line 93: ch_remote, vhost_user_blk, vhost_user_net is not required here?

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -0,0 +1,229 @@
...
+export OPENSSL_NO_VENDOR=1
+%endif
+cargo build --release --target=%{rust_def_target} %{cargo_pkg_feature_opts} %{cargo_offline}
+
+%if 0%{?using_musl_libc}
+cargo build --release --target=%{rust_musl_target} %{cargo_pkg_feature_opts} %{cargo_offline}
+
+%endif
+
+%clean
```
ms-mahuber, line 142: if we need this, why does the regular CH SPEC not need this?

Redent0r:
Closing in favor of https://github.com/microsoft/azurelinux/pull/9261

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
Redent0r: They should not because clh-cvm always has the patch number in its version (eg 38.0.72). Vs upstream only uses major and minor (eg 38.0). Still good to make it clear these are cvm sources

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
Redent0r: Looks like we haven't been installing those https://dev.azure.com/mariner-org/mariner/_git/CBL-Mariner-Proprietary?path=/SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec&version=GBmain&line=89&lineEnd=99&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

I'll look into it more

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
Redent0r: Will also need to adapt Kata CI to rename the tars (add `-cvm` ) given by LSG

SPECS/cloud-hypervisor-cvm/fetch-private-deps
Redent0r: Double checked and doesn't look like we need this. This seems to be a helper script used by LSG to fetch the sources from the package feed

SPECS/cloud-hypervisor-cvm/update_src
Redent0r: Double checked and no. Same as https://github.com/microsoft/azurelinux/pull/9161/files#discussion_r1621506207. We can get the source tar by looking at the tags page in https://github.com/microsoft/cloud-hypervisor/tags , and we can generate the vendor tar by following https://github.com/microsoft/azurelinux/pull/9261/files#diff-5d949bbc496656a772fe7a22601f9837efcc1dffcf513a7ccffd6aa6d53ae2e4R16-R22 (adding this comment to cvm spec)

May be worth documenting this comment in the form of a useful script for the future. It's 3 or 4 straightforward commands

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
Redent0r: Would like to take this a step forward and remove gitrev from release number altogether. We can find that info by looking at the version used in https://github.com/microsoft/cloud-hypervisor/tags eg

image

This would also further align with normal clh

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
Redent0r: Just double checked and this isn't as straightforward as expected. The macro `%setup` expects the untared folder to also be called *-cvm

image

And LSG doesn't renames them to *-cvm. We could rename them ourselves, but this would alter the hash sum. This makes me want to leave them as is (eg `cloud-hypervisor-38.0.72.tar.gz`), and rely on the fact that there won't be a name collision with normal clh sources, because clh-cvm will always have a patch number in its source tar

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
Redent0r: ch-remote: utility to interact with cloud-hypervisor api (https://www.cloudhypervisor.org/docs/prologue/commands/#ch-remote-binary). Seems we can already do that using curl https://dev.azure.com/mariner-org/mariner/_wiki/wikis/mariner.wiki/2539/Booting-the-UVM-manually-on-a-node-via-CH-API-socket

vhost_user_blk and vhost_user net: They seem to be "backends" that provide functionality for network and block devices (https://www.qemu.org/docs/master/system/devices/vhost-user.html) . I think kata(cc) handles network and block devices differently and doesn't need these. Also considering they weren't being installed before

PR #9210, opened 2024-05-23
Remove requirement on awk for dracut device caching

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

To allow dracut to work with `selinux` we upstreamed a fix https://github.com/dracutdevs/dracut/pull/2324 to dracut to use `awk` instead of `grep -P`. In our new 3.0 images we are looking to remove `gawk` since it is fairly large. Unfortunately this results in a lot of (non-fatal) error messages from dracut:
```
WARN[0022][imager] initramfs 3.0-3.azl3 posttrans               
WARN[0023][imager] dracut: Disabling early microcode, because kernel does not support it. CONFIG_MICROCODE_[AMD|INTEL]!=y 
WARN[0023][imager] /usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found 
WARN[0023][imager] /usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found 
WARN[0023][imager] /usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found 
WARN[0023][imager] /usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found 
WARN[0023][imager] /usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found 
WARN[0023][imager] /usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found 
WARN[0023][imager] /usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found 
WARN[0023][imager] /usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found 
WARN[0023][imager] /usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found 
...
```

This is just a cache miss so is non-fatal, but it looks very ugly. Add a patch to only use the cache lookup if `awk` is available.

###### Change Log  

- Suppress `/usr/lib/dracut/dracut-functions.sh: line 247: awk: command not found ` errors from `dracut` if `awk` is unavailable.

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/50085799
- 
###### Test Methodology

- Local builds


Camelron (approved):
Thank you, sir. LGTM

rlmenge (approved):

anphel31 (approved):

christopherco (approved):
LGTM. Consider upstreaming this too

SPECS/dracut/0012-Skip-maj_min-cache-lookup-if-awk-unavailable.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/dracut-functions.sh b/dracut-functions.sh
+index 05d02c2..15083d0 100755
+--- a/dracut-functions.sh
++++ b/dracut-functions.sh
+@@ -243,7 +243,7 @@ get_maj_min() {
+     local _majmin
+     local _out
+ 
+-    if [[ $get_maj_min_cache_file ]]; then
++    if [[ $get_maj_min_cache_file ]] && awk -V > /dev/null 2>&1; then
```
PawelWMS, line 19: Instead of disabling the caching, I think it might be better to just drop the dependency on `awk`. We could then keep the functionality and have `dracut` require one less tool to work correctly/most efficiently. For instance, we could replace

```bash
grep -m1 -oE "^$1 \S+$" "$get_maj_min_cache_file" | awk '{print $NF}'
```

with 

```bash
grep -m1 -oE "^$1 \S+$" "$get_maj_min_cache_file" | grep -oE "\S+$"
```

PawelWMS: We could then also upstream it again.

PawelWMS (changes requested):
I suggest dropping the dependency on `awk` instead of disabling caching when it's not available.

dmcilvaney:
> I suggest dropping the dependency on `awk` instead of disabling caching when it's not available.

`awk` is used is several places in dracut, I don't expect they will be interested in the churn needed to remove it in this one spot. This is 100% a cosmetic fix for us.

`awk` is a core component of the posix standard set of tools and all sane environments should have it (we don't in many of our images now because `gawk` pulls ~4MB of stuff with it). 

I don't mind changing the behavior in the patch to do a double grep, but I don't think it's worth the effort on either side to upstream it just because our environment is really odd.

dmcilvaney:
Redid with double grep as suggested. I also formatted the patch so we could upstream it (assuming dracut-ng has similar requirements to old dracut).

PawelWMS:
> > I suggest dropping the dependency on `awk` instead of disabling caching when it's not available.
> 
> `awk` is used is several places in dracut, I don't expect they will be interested in the churn needed to remove it in this one spot. This is 100% a cosmetic fix for us.
> 
> `awk` is a core component of the posix standard set of tools and all sane environments should have it (we don't in many of our images now because `gawk` pulls ~4MB of stuff with it).
> 
> I don't mind changing the behavior in the patch to do a double grep, but I don't think it's worth the effort on either side to upstream it just because our environment is really odd.
> 
> Also, `dracut` upstream is abandoned and we would need to jump to `dracut-ng` to upstream.

Thanks for the context! In that case can we switch to `grep`, so we can keep using this functionality? I understand there is no need to disable if, if not necessary.

PawelWMS:
> > > I suggest dropping the dependency on `awk` instead of disabling caching when it's not available.
> > 
> > 
> > `awk` is used is several places in dracut, I don't expect they will be interested in the churn needed to remove it in this one spot. This is 100% a cosmetic fix for us.
> > `awk` is a core component of the posix standard set of tools and all sane environments should have it (we don't in many of our images now because `gawk` pulls ~4MB of stuff with it).
> > I don't mind changing the behavior in the patch to do a double grep, but I don't think it's worth the effort on either side to upstream it just because our environment is really odd.
> > Also, `dracut` upstream is abandoned and we would need to jump to `dracut-ng` to upstream.
> 
> Thanks for the context! In that case can we switch to `grep`, so we can keep using this functionality? I understand there is no need to disable if, if not necessary.

Nevermind, GitHub just refreshed...

PawelWMS (approved):

dmcilvaney:
Quick rebase, only change was release number + date

PR #9227, opened 2024-05-24
Fixing Perl automatic requires and provides.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

### Cherry-pick of #9226.

Our version of the `perl-libs` package had a run-time dependency on `perl`. This is incorrect as `perl-libs` has no components, which would require the default `perl` package. The additional dependency was added by RPM's built-in Perl dependencies generator `perl.req`. This generator is only present in the raw toolchain and thus during the build of the regular toolchain. The final version of the `rpm` package has that file removed, so it doesn't conflict with a better version of this generator coming from the `perl-generators` package.

This change switches the order of the toolchain package builds to install `perl-generators` as soon as possible and re-builds 3 Perl packages, which were needed to be installed in order to allow `perl-generators` to build.

The additional bumps of the `Release` tags in the other Perl specs are required to force a re-build of the other toolchain Perl packages to fix their requires and provides and publish them in the next release.

Yes.

###### Test Methodology

- [Full dev build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=575660&view=results).


toolkit/scripts/toolchain/build_official_toolchain_rpms.sh
```diff
@@ -390,6 +390,20 @@ chroot_and_install_rpms zlib
...
 
+# perl-generators requires perl-Fedora-VSP
+# All perl packages need perl-generators to correctly
+# generate their run-time provides and requires.
+build_rpm_in_chroot_no_install perl-Fedora-VSP
+chroot_and_install_rpms perl-Fedora-VSP
+build_rpm_in_chroot_no_install perl-generators
+chroot_and_install_rpms perl-generators
+
+# Rebuilding perl packages with perl-generators installed.
```
dmcilvaney, line 401: I feel like we might have an issue here with the delta build... lets assume `pkg.rpm` is NOT available for delta, but we are running a delta build otherwise.

We place our prospective delta build candidates into `$CHROOT_RPMS_DIR` on line 125.

In `build_rpm_in_chroot_no_install()` we check if we have a delta rpm via `$(find $CHROOT_RPMS_DIR -name "$rpm" -print -quit)`. If we are missing one (as is our case here), we will run `rpmbuild` which I believe also places files into ` $CHROOT_RPMS_DIR`.

We then collect our final product via `copy_built_rpms()` which takes them from `$CHROOT_RPMS_DIR` to `$FINISHED_RPM_DIR`.

When we now run the second build, I'd expect that `$(find $CHROOT_RPMS_DIR -name "$rpm" -print -quit)` will now see the first run files already present, and skip the second build.

dmcilvaney (changes requested):
Just to double check my understanding, the delta build won't mind since we check for the candidate packages in

toolkit/scripts/toolchain/build_official_toolchain_rpms.sh
dmcilvaney: https://github.com/microsoft/azurelinux/pull/9264 should fix

dmcilvaney (approved):
Approve if merged alongside https://github.com/microsoft/azurelinux/pull/9264

toolkit/scripts/toolchain/build_official_toolchain_rpms.sh
PawelWMS: Merged, thank you, @dmcilvaney!

PR #9251, opened 2024-05-29
Improve network behavior, use contexts instead of cancel channel

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

https://github.com/microsoft/azurelinux/pull/9198 creates an issue where a long-running download may block a tool from exiting. While it is possible to enhance the existing `cancel <-chan struct{}` pattern, it is difficult to make a double layered cancel thread safe (srpmpacker cancel + 404 abort cancel). The better approach is to use `ctx context.Context` objects. They provide a similar `chan` (`<-ctx.Done()`) we can use, and if created via `contex.WithCancel()` a cancelation function which replaces `close(chan)`. The cancelation function is safe to call from multiple locations and multiple times (unlike `close()` which will panic on a double close).

This PR replaces (most of) the channel based cancels with contexts. The scheduler is left as-is since it is quite complex and is working ok for now. We can revisit later if desired.

It also unifies the problematic download code into the `network` package via `DownloadFileWithRetry()`.

###### Change Log  

- Update cancel mechanisms to use `context.Context`
- Move `downloader.DownloadFile()` code into `network.DownloadFileWithRetry()`
  - Use this new function where applicable
- 404 errors skip download retries (https://github.com/microsoft/azurelinux/pull/9198)
- `precacher` errors on all download failures (https://github.com/microsoft/azurelinux/pull/9198)
- Increase network timeout (https://github.com/microsoft/azurelinux/pull/9198)

**NO**

###### Test Methodology

- Local testing


toolkit/tools/internal/retry/retry.go
```diff
@@ -40,21 +41,25 @@ func calculateLinearDelay(failCount int, sleep time.Duration) time.Duration {
...
 }
 
-// backoffSleep sleeps for the given duration, unless the cancel channel is triggered or closed first. The cancel channel
-// may be nil in which case the sleep will always complete.
-func backoffSleep(delay time.Duration, cancel <-chan struct{}) (cancelled bool) {
+// backoffSleep sleeps for the given duration, unless the context is cancelled. The context
+// may be nil in which case the background context will be used and the sleep will always complete.
+func backoffSleep(delay time.Duration, ctx context.Context) (cancelled bool) {
+	// Context is optional, if not provided, use a background context.
+	if ctx == nil {
```
dmcilvaney, line 48: ended up moving this back into the private function since there are multiple paths that reach here from public functions.

toolkit/tools/precacher/precacher.go
```diff
@@ -135,30 +135,30 @@ func downloadMissingPackages(rpmSnapshot *repocloner.RepoContents, packagesAvail
 	wg := new(sync.WaitGroup)
 	netOpsSemaphore := make(chan struct{}, concurrentNetOps)
 	results := make(chan downloadResult)
-	doneChannel := make(chan struct{})
```
cwize1, line 138: The `doneChannel` here is more of a synchronization primitive than a cancel signal. IMHO, you should leave this code as-is to avoid confusion.

That being said, you could simplify this code by calling `close(results)` instead of `close(done)`. Then in `monitorProgress` change the `result := <-results` to `result, ok := <-results` and use `ok` to break the loop.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -237,19 +237,22 @@ func createAllSRPMs(specsDir, distTag, buildDir, outDir string, workers, concurr
 	timestamp.StartEvent("packing SRPMS", nil)
 	defer timestamp.StopEvent(nil)
 
+	rootCtx, cancelFunc := context.WithCancel(context.Background())
```
cwize1, line 240: Is this needed? You don't seem to be using `cancelFunc` here.

cwize1 (changes requested):

toolkit/tools/internal/network/network.go
```diff
@@ -31,13 +47,62 @@ func JoinURL(baseURL string, extraPaths ...string) string {
...
+// will not be retried.
+// srcUrl: The URL to download from.
+// dstFile: The local file to save the download to.
+// caCerts: The CA certificates to use for the download.
+// tlsCerts: The TLS certificates to use for the download.
+// externalCancel: An optional channel that can be used to cancel the download, pass nil if not needed.
+// returns: wasCancelled: true if the download was cancelled via the external cancel channel, false otherwise.
+// returns: err: An error if the download failed (including being cancelled), nil otherwise.
+func DownloadFileWithRetry(srcUrl, dstFile string, caCerts *x509.CertPool, tlsCerts []tls.Certificate, ctx context.Context) (wasCancelled bool, err error) {
+	if ctx == nil {
```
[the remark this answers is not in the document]
dmcilvaney: Docs agree, and actually suggest having it as 1st param. Will change. https://go-language.org/go-docs/context/

toolkit/tools/internal/network/network.go
```diff
@@ -31,13 +47,62 @@ func JoinURL(baseURL string, extraPaths ...string) string {
...
+}
+
 // DownloadFile downloads `url` into `dst`. `caCerts` may be nil. If there is an error `dst` will be removed.
 func DownloadFile(url, dst string, caCerts *x509.CertPool, tlsCerts []tls.Certificate) (err error) {
 	logger.Log.Debugf("Downloading (%s) -> (%s)", url, dst)
 
 	dstFile, err := os.Create(dst)
 	if err != nil {
-		return
+		return fmt.Errorf("%w:\nfailed to create file:\n%w", ErrDownloadFileOther, err)
```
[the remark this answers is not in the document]
dmcilvaney: It should format out to be "download error: failed to create file: ..." I think it's ok? I can modify it to be "failed to create destination file" maybe? The file download operation DID fail, even if its not due to a network issue.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -616,7 +617,7 @@ func (r *RpmRepoCloner) clonePackage(baseArgs []string) (preBuilt bool, err erro
 		finalArgs := append(baseArgs, reposArgs...)
 
 		// We run in a retry loop on errors deemed retriable.
-		cancel := make(chan struct{})
+		ctx, cancelFunc := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
dmcilvaney: done

toolkit/tools/srpmpacker/srpmpacker.go
dmcilvaney: Probably don't even need a root ctx here at all, each function can handle their own ctx

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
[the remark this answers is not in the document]
PawelWMS: I think it's fine (and expected) to not cancel a context. If I understand this mechanism correctly, we only want to cancel when something special happens. In this case it's an unrecoverable error, which allows us to exit early, since retries won't help.

toolkit/tools/precacher/precacher.go
dmcilvaney: Agreed, always forget about the `ok` mechanism. Will use the results flow.

toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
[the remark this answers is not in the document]
dmcilvaney: Per the examples in https://go-language.org/go-docs/context/#example_WithCancel the suggested pattern seems to be to use defer. 

A context should be closed when the work is done, not just in the case of error. I agree that "cancel" is an odd name for it, but that does seem to be the expected flow.

toolkit/tools/downloader/downloader.go
```diff
@@ -96,7 +97,7 @@ func main() {
 		}
 	}
 
-	err = downloader.DownloadFile(*srcUrl, *dstFile, caCerts, tlsCerts)
+	_, err = network.DownloadFileWithRetry(context.Background(), *srcUrl, *dstFile, caCerts, tlsCerts)
```
mfrw, line 100: +1 on use of context.

I feel like we should have a context with deadline here in the top level, with a sane max timeout - under which our asssumption should be that the downloader even after retries and all is done. This would put a upper level bounding on our downloader.
What do you folks think ?

dmcilvaney: Not opposed to it... we would probably need to adjust the behavior of the retry to always return an error if it is cancelled though (which isn't the worst idea... currently it returns the error from the last time it did a retry attempt).

toolkit/tools/internal/retry/retry.go
```diff
@@ -40,21 +44,25 @@ func calculateLinearDelay(failCount int, sleep time.Duration) time.Duration {
...
+// backoffSleep sleeps for the given duration, unless the context is cancelled. The context
+// may be nil in which case the background context will be used and the sleep will always complete.
+func backoffSleep(ctx context.Context, delay time.Duration) (cancelled bool, err error) {
+	if ctx == nil {
+		return false, ErrNilContext
+	}
+
 	// Check if we were cancelled before we sleep to avoid a race condition with
 	// delay=0 and an immediate cancel.
 	select {
```
Camelron, line 60: So if I'm understanding this right, the cancellation function will cause ctx.Done() to start reporting true? 

So at any point if any other section of the code calls the cancaelFunc, next time backoffSleep is called it will propagate up the chain ending loops and gracefully exiting the context?

toolkit/tools/downloader/downloader.go
dmcilvaney: Ok, gave it a go

toolkit/tools/internal/retry/retry.go
dmcilvaney: Correct (although to be accurate `ctx.Done()` is a `chan`, and reading from a closed channel in a `select` block allows that branch to be executed). The idea of a context is it handles all the synchronization of canceling a job for you.

Camelron (approved):
Treating this as a refactor, the changes look good. Not sure I fully understand what was wrong with the original channel implementation or how it can block the tool from gracefully exiting.

toolkit/tools/internal/network/network.go
```diff
@@ -31,13 +58,69 @@ func JoinURL(baseURL string, extraPaths ...string) string {
...
+// tlsCerts: The TLS certificates to use for the download.
+// timeout: The maximum duration for the download operation, use network.NoTimeout for no timeout, or network.DefaultTimeout for a default timeout.
+// returns: wasCancelled: true if the download was cancelled via the external cancel channel, false otherwise.
+// returns: err: An error if the download failed (including being cancelled), nil otherwise.
+func DownloadFileWithRetry(ctx context.Context, srcUrl, dstFile string, caCerts *x509.CertPool, tlsCerts []tls.Certificate, timeout time.Duration) (wasCancelled bool, err error) {
+	if ctx == nil {
+		return false, fmt.Errorf("context is nil")
+	}
+	if timeout == 0 {
+		timeout = NoTimeout
```
cwize1, line 78: You should probably switch between `context.WithTimeout` and `context.WithCancel`. Setting `timeout` to max feels hacky.

cwize1 (changes requested):

toolkit/tools/internal/network/network.go
dmcilvaney: Ah, not a bad idea. Will do.

cwize1 (approved):

toolkit/tools/internal/network/network.go
```diff
@@ -62,12 +146,12 @@ func DownloadFile(url, dst string, caCerts *x509.CertPool, tlsCerts []tls.Certif
 
 	response, err := client.Get(url)
 	if err != nil {
-		return
+		return fmt.Errorf("%w:\nrequest failed:\n%w", ErrDownloadFileOther, err)
```
PawelWMS, line 149: While here, we should also make use of `http.NewRequestWithContext`. On way to do it, is to replace `client.Get(url)` with:

```golang
	request, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
	if err != nil {
		return fmt.Errorf("%w:\nfailed to create request:\n%w", ErrDownloadFileOther, err)
	}

	response, err := client.Do(request)
```

toolkit/tools/precacher/precacher.go
```diff
@@ -169,8 +168,11 @@ func monitorProgress(total int, results chan downloadResult, doneChannel chan st
 
 	for done := false; !done; {
 		// Wait for a result from a worker, or the done channel to be closed (which means all workers are done)
-		select {
-		case result := <-results:
+		result, ok := <-results
+		if !ok {
+			// All workers are done, finish this iteration of the loop and then return
+			done = true
+		} else {
```
PawelWMS, line 175: Unnecessary `if-else` indent making the code look a bit harder to read.

PawelWMS, line 175: I think we still left the code a bit messy and are not utilizing what the language gives us.

Since we don't have `doneChannel` now, then we can also dump `done` and replace `for done := false; !done;` with `for result := range results`. It will exit on its own after reading the last result.

PR #9269, opened 2024-05-31
Image Customizer: Modify /etc/selinux/config on SELinux disable.

cwize1:
Currently, when the image customizer disables SELinux in an image it doesn't bother modifying the `/etc/selinux/config` file and instead only modifies the kernel command-line args. While this is technically fine, it is confusing for the user. So, this change fixes this so that `/etc/selinux/config` is kept in sync with the kernel command-line args.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -64,10 +64,10 @@ const (
 	// SELinuxConfigEnforcing is the string value to set SELinux to enforcing in the /etc/selinux/config file.
 	SELinuxConfigEnforcing = "enforcing"
 
-	// SELinuxConfigEnforcing is the string value to set SELinux to permissive in the /etc/selinux/config file.
+	// SELinuxConfigPermissive is the string value to set SELinux to permissive in the /etc/selinux/config file.
 	SELinuxConfigPermissive = "permissive"
 
-	// SELinuxConfigEnforcing is the string value to set SELinux to disabled in the /etc/selinux/config file.
+	// SELinuxConfigDisabled is the string value to set SELinux to disabled in the /etc/selinux/config file.
 	SELinuxConfigDisabled = "disabled"
 
 	// GrubCfgFile is the filepath of the grub config file.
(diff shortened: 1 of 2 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -513,46 +513,40 @@ func handleSELinux(selinuxMode imagecustomizerapi.SELinuxMode, resetBootLoaderTy
 		}
 	}
 
-	if selinuxMode != imagecustomizerapi.SELinuxModeDisabled {
-		err = updateSELinuxModeInConfigFile(selinuxMode, imageChroot)
-		if err != nil {
-			return selinuxMode, err
-		}
+	err = updateSELinuxModeInConfigFile(selinuxMode, imageChroot)
```
Britel, line 516: Looks like `bootCustomizer.GetSELinuxMode` will only check the kernel-cmd setting for SElinux - at least if its `SELinuxModeDefault`.

In that case, and with the check on line 491, could we still end up in a case when the config and kernel-cmd are misaligned?

cwize1: Fair. Fixed.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1812,6 +1812,8 @@ func SELinuxUpdateConfig(selinuxMode configuration.SELinux, installChroot *safec
 		mode = SELinuxConfigEnforcing
 	case configuration.SELinuxPermissive:
 		mode = SELinuxConfigPermissive
+	case configuration.SELinuxOff:
+		mode = SELinuxConfigDisabled
```
PawelWMS, line 1816: Since we're on `SELinuxConfigDisabled`, could you also update the comments above this and the `SELinuxConfigPermissive` constants in `imagegen/installutils/installutils.go`? They both start with "// **SELinuxConfigEnforcing** is the string..." - it's just about using the correct constant name.

dmcilvaney (approved):

toolkit/tools/imagegen/installutils/installutils.go
cwize1: Fixed.

PawelWMS (approved):

gmileka (approved):

PR #9284, opened 2024-06-03
Fix CVE-2024-3154 in package cri-o

0xba1a:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Fix CVE-2024-3154 in package cri-o

###### Change Log  

- Patch CVE-2024-3154 for cri-o

**NO**

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2024-3154

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=580253&view=results

Arbitrary sytemd pod annotations can be copied to the OCI. So, it may affect the host. So, explicitly hard-code the systemd annotations to AllAllowedAnnotations file.

Ref: [[annotations: add OCI runtime specific annotations to the AllowedAnnot… · haircommander/cri-o@976ab1f (github.com)](https://github.com/haircommander/cri-o/commit/976ab1f4c916099fc1f2e6569f13e45df2f26b4f)](https://github.com/haircommander/cri-o/commit/976ab1f4c916099fc1f2e6569f13e45df2f26b4f)

**Explanation:** ValidateRuntimeAllowedAnnotations function in pkg/config/config.go uses the following logic to validate the annotations.

1. add all annotations in AllAllowedAnnotations to a temporary disallowedAnnotations
2. loop through the allowed annotations, if it is in disallowedAnnotations, remove it from disallowedAnnotations.
3. If it is not there, return error as not an expected Annotations
4. The leftover disallowedAnnotations were added to the global DisallowedAnnotations list

If an unknown annotation is there, it is filtered in the step-3. So, it looks like an unknown annotation is flagged by the ValidateRuntimeAllowedAnnotations function. But it is not explicitly added to the DisallowedAnnotations variable. So, if the caller doesn’t honor the response of ValidateRuntimeAllowedAnnotations and just rely on DisallowedAnnotations, this patch is worth adding. But it looks like that is not possible, because upon an unknown annotations, the DisallowedAnnotations also will be partial.

Anyway adding the patch as it is not causing any harm.

*Note: If you can’t find the function grep for `AllAllowedAnnotations` in the entire repo.*

**Not patching the vendor packages:**

Below is the actual patch mentioned in the CVE websites.

[[features: implement returning potentiallyUnsafeConfigAnnotations list by AkihiroSuda · Pull Request #4217 · opencontainers/runc (github.com)](https://github.com/opencontainers/runc/pull/4217/commits/eefc6ae2544a6819da9f92c5aa8e65d356da4c96)](https://github.com/opencontainers/runc/pull/4217/commits/eefc6ae2544a6819da9f92c5aa8e65d356da4c96)

But it is not applicable to us. Because the file getting patched is features.go. But this file is not in the vendor package. Instead only the `libcontainer` directory was cloned into the `vendor/github.com/opencntainers/runc` directory. Also, it is not mentioned by the CVE tools as well. So, patching this not relevant to cri-o

### DIFF
--- a/SPECS/cri-o/cri-o.spec
+++ b/SPECS/cri-o/cri-o.spec
@@ -1792,4 +1792,4 @@ mkdir -p /opt/cni/bin
 - Add cri-o package: CRI-O is meant to provide an integration path between OCI
   conformant runtimes and the kubelet. Specifically, it implements the Kubelet
   Container Runtime Interface (CRI) using OCI conformant runtimes. The scope of
-
+  CRI-O is tied to the scope of the CRI.
--- a/SPECS/cri-o/cri-o.spec
+++ b/SPECS/cri-o/cri-o.spec
@@ -26,7 +26,7 @@ Summary:        OCI-based implementation of Kubernetes Container Runtime Interfa
 # Define macros for further referenced sources
 Name:           cri-o
 Version:        1.22.3
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        ASL 2.0
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
@@ -63,6 +63,7 @@ Patch7:         CVE-2022-21698.patch
 Patch8:         CVE-2023-44487.patch
 Patch9:         CVE-2024-28180.patch
 Patch10:        CVE-2024-21626.patch
+Patch11:        CVE-2024-3154.patch
 BuildRequires:  btrfs-progs-devel
 BuildRequires:  device-mapper-devel
 BuildRequires:  fdupes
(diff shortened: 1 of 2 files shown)


PawelWMS (approved):

SPECS/cri-o/cri-o.spec
```diff
@@ -1788,4 +1792,4 @@ mkdir -p /opt/cni/bin
 - Add cri-o package: CRI-O is meant to provide an integration path between OCI
   conformant runtimes and the kubelet. Specifically, it implements the Kubelet
   Container Runtime Interface (CRI) using OCI conformant runtimes. The scope of
-  CRI-O is tied to the scope of the CRI.
```
christopherco, line 1791: Was this change intentional?

christopherco (changes requested):

SPECS/cri-o/cri-o.spec
PawelWMS: Good catch, I completely missed that in my review. I don't think we should be modifying someone else's changelog entries. I'll update the PR in Bala's name.

christopherco (approved):

PR #9291, opened 2024-06-04
Updated context to be used for HTTP request + refactoring

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

- Expanded the use of context to pass it to our HTTP requests.
- Added a context `nil` check in `azureblobstorage.go`.
- Re-named `cancelCtx` to `closeCtx` to better reflect the purpose of some of our `defer` calls.
- 1 loop refactoring to use the `for element := range someChannel` loops.

No.

###### Associated issues  

- #9251

###### Test Methodology

- Local tests.
- [Full dev build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=581443&view=results).


dmcilvaney (commented):
While it seems slightly pointless, everywhere I look says to always `defer cancelFunc()`. By its nature it should always be safe to do so. In fact I believe `go vet` will actually flag any un-deferred cancel functions.

I think the best argument is that maybe today its ok, but what about  in 6 months? A developer will see a context, and assume that we followed best practices. They will add a go routine that only exits on the cancel signal, and now we have a memory leak.

tldr, `defer cancelFunc()` is a really easy and safe way to ensure that a context is cleaned up no mater what code flow we end up following.

toolkit/tools/internal/network/network_test.go
```diff
@@ -139,9 +139,7 @@ func TestDownloadFile(t *testing.T) {
 			dstDir := t.TempDir()
```
dmcilvaney, line 139: https://github.com/microsoft/azurelinux/pull/9307#discussion_r1626770961 We should add a test here to check early cancel propagates correctly.

dmcilvaney (approved):

sameluch (approved):

PR #9325, opened 2024-06-05
Improve handling of preview repo files

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

The way we handle preview for the core repo is very frustrating. When we package a toolkit we generate a `./toolkit/repos/` directory that holds all our repo files. The preview repo in there is used when you select `USE_PREVIEW_REPO=y`. 

If you are building from the core repo however, this directory is missing. The tools would print a warning message, but continue regardless. This is a bad idea because the toolchain RPMS will download but all subsequent tools will fail to use the preview repos without generating a good error. Instead you just end up with unexpected dependency errors at the end of the build.

There is no reason we can't automatically reach out to our checked-in copy of the preview repo file, and if we still can't find a valid preview repo file we should actually generate an error immediately and stop.

**As part of this the `PREVIEW_REPO` path is being made private and should no longer be used.**

###### Change Log  

- Move preview repo handling to a dedicated makefile
- Generate error if all sources of preview repo files are missing
- Support using the copy of the preview repo from the `./SPECS` directory if needed.

**NO**

###### Test Methodology

- Local testing


toolkit/Makefile
```diff
@@ -136,6 +135,12 @@ PACKAGE_URL_LIST   += https://packages.microsoft.com/azurelinux/$(RELEASE_MAJOR_
 REPO_LIST          ?=
 SRPM_URL_LIST      ?= https://packages.microsoft.com/azurelinux/$(RELEASE_MAJOR_ID)/prod/base/srpms
 
+##help:var:PREVIEW_REPO=<path>=Path to the preview repo file to use if 'USE_PREVIEW_REPO=y' is set. If not set, the default preview repo file will be used.
+PREVIEW_REPO                    ?=
+preview_repo_primary_default    ?= $(toolkit_root)/repos/azurelinux-official-preview.repo
+# Intentionally not using $(SPECS) here to avoid issues with custom SPECS directories
+preview_repo_backup_default     ?= $(PROJECT_ROOT)/SPECS/azurelinux-repos/azurelinux-official-preview.repo
+
```
PawelWMS, line 143: This made me think we may also be able to backpaddle with 3.0 a bit and make `PREVIEW_REPO` an "internal" variable. We're allowed to make breaking changes and it's never been documented, which makes it a good candidate and can help us simplify the logic below.

toolkit/Makefile
```diff
@@ -168,14 +173,38 @@ else
...
+               $(warning )
+               $(warning ######################### ERROR #########################)
+               $(warning 'USE_PREVIEW_REPO=y' is set but all of the following failed:)
+               $(warning - 'PREVIEW_REPO=<path>' is not set)
+               $(warning - default '$(preview_repo_primary_default)' is missing)
+               $(warning - backup  '$(preview_repo_backup_default)' is missing)
+               $(warning ######################### ERROR #########################)
+               $(error )
+            endif
+         endif
```
PawelWMS, line 207: Here's the code for the idea I've mentioned in our offline discussion:
```Makefile
# Different Makefile. For instance "toolkit.mk":
preview_repo_source = $(PROJECT_ROOT)/SPECS/azurelinux-repos/azurelinux-official-preview.repo

# Make the "PREVIEW_REPO" variable "internal":
preview_repo = $(toolkit_root)/repos/mariner-official-preview.repo

ifeq ($(DAILY_BUILD_ID),)
# Leave only the file existence checks from your current `else` branch:
   ifeq ($(USE_PREVIEW_REPO),y)
      ifeq ($(wildcard $(preview_repo)),)
         ifeq ($(wildcard $(preview_repo_source)),)
           # Error message about the toolkit being neither packaged nor the user running from inside the core repository.
        endif
      endif
   endif

   override PACKAGE_URL_LIST   += https://packages.microsoft.com/azurelinux/$(RELEASE_MAJOR_ID)/preview/base/$(build_arch)
   override PACKAGE_URL_LIST   += https://packages.microsoft.com/azurelinux/$(RELEASE_MAJOR_ID)/preview/base/debuginfo/$(build_arch)
   override PACKAGE_URL_LIST   += https://packages.microsoft.com/azurelinux/$(RELEASE_MAJOR_ID)/preview/ms-oss/$(build_arch)
   override SRPM_URL_LIST      += https://packages.microsoft.com/azurelinux/$(RELEASE_MAJOR_ID)/preview/base/srpms
   override REPO_LIST          += $(preview_repo)
endif

# A native "make" approach:
$(preview_repo): $(preview_repo_source)
	@mkdir -p "$(dir $(preview_repo))
	@cp $(preview_repo_source) $(preview_repo)

# Other Makefiles for targets using "REPO_LIST":
$(graph_file): (...) $(REPO_LIST)

# Btw. we have a bunch of targets using "REPO_LIST" but not mentioning it in the prerequisites. Should we add it?
```

It assumes we make `PREVIEW_REPO` internal.

My main suggestion here is to:
1. remove one set of the manual file checking and 
2. use native `make` to do some of the work for us, since it's already centered around existence of files.

I think we can simplify a whole bunch of logic this way. The error message would also get simpler, since we'd now tell the user to either use a packaged toolkit from the core repo or run from the core itself.

toolkit/Makefile
dmcilvaney: Sure, sounds like a plan. Might ruffle a few feathers if people are doing any weird work-arounds but now is the time.

toolkit/Makefile
dmcilvaney: I'd have to do a bit reshuffle of the makefile to move those variables into another sub file, we do the includes after setting this stuff up.

dmcilvaney: Actually, just moved all this into its own file

PawelWMS (approved):

PawelWMS (approved):

PawelWMS (commented):
Can you also update the summary to say we're making `PREVIEW_REPO` private for 3.0?

anphel31 (approved):

toolkit/.gitignore
```diff
@@ -8,4 +8,7 @@ scripts/toolchain/container/CVE-2021-38185.patch
 scripts/toolchain/container/linker-script-readonly-keyword-support.patch
 scripts/toolchain/container/rpm-define-RPM-LD-FLAGS.patch
 scripts/toolchain/container/texinfo-perl-fix.patch
-scripts/toolchain/container/toolchain-local-wget-list
\ No newline at end of file
+scripts/toolchain/container/toolchain-local-wget-list
+
+# This is generated by the toolkit packaging process, or by a backup generation from the SPECS directory.
+repos/
```
cwize1, line 14: This will ignore any descendent directory named `repos` not just directly within this directory. Is that your intended behavior?

Write it as `/repos/` to limit it to only this directory.

(You might also want to fix the above entries.)

toolkit/scripts/repoquerywrapper.mk
```diff
@@ -55,4 +55,3 @@ $(STATUS_FLAGS_DIR)/repoquerywrapper.flag: $(go-repoquerywrapper) $(chroot_worke
 		$(if $(filter y,$(ENABLE_TRACE)),--enable-trace) \
 		--timestamp-file=$(TIMESTAMP_DIR)/repoquerywrapper.jsonl && \
 	scripts/update-target-if-output-changed.sh $@ $(QUERY_OUTPUT_FILE)-old $(QUERY_OUTPUT_FILE)
-
```
cwize1, line 58: nit: Restore deleted empty line.

toolkit/.gitignore
dmcilvaney: interesting, will do

toolkit/scripts/repoquerywrapper.mk
dmcilvaney: Ah, my editor considers multiple newlines at the end to be trailing whitespace. Will re-save without autoformat

PR #9326, opened 2024-06-05
Rework dracut defaults into separate subpackages for more granular image configuration

Camelron:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Due to a change in 3.0 which has us using dracut instead of mkinitrd (https://github.com/microsoft/azurelinux/pull/8126), we now have a required set of kernel modules which must be present for a given kernel to generate an initramfs. kernel-mshv has a much more minimal set of kernel drivers built; it by definition does not need to support being virtualized over xen, for example. Because we have this universally required set of modules and it is the superset required for all scenarios, kernel-mshv is unable to generate an initramfs which breaks our Dom0 boot. 

Mariner has historically installed the superset of kernel modules needed for all scenarios. Now, we care about installing only the relevant set of kernel modules needed for early boot in a given image's scenario. We accomplish this by separating out our dracut's defaults.conf into multiple smaller .conf files. 

###### Change Log  

- Split dracut defaults into separate packages.
- Flag new dracut subpackages as relevant to initramfs regeneration
- Configure existing image configs with reasonable sets of dracut modules

NO

###### Test Methodology

- Tested a wide set of scenarios to ensure that images are booting in all cases. 
- Core-efi image:
    - Local hyper-v vm - success
    - bvt tests - success
    - kdump - success
     - fips enablement of core image - success
     - full iso image install - success
- Marketplace image:
      - AKS image - success
      - lisa tests - success, see above summary
      - x86 image - success
      - Arm image - success
      - kdump - success
Baremetal images - success


SPECS/dracut/00-hostonly.conf
```diff
@@ -0,0 +1,4 @@
+# kdump currently uses the host system's initrd when enrolling a crash kernel
```
christopherco, line 1: Curious what would change if we had #9333 to regenerate the crash kernel initrd rather than using the host system's version

SPECS/dracut/dracut.signatures.json
```diff
@@ -1,10 +1,14 @@
...
+  "00-vrf.conf": "e2885a4b090d8ca3771e60ce6dcd8b849e28ce5002a5c7b71ff796a92deb2810",
+  "00-xen.conf": "8b7a89b7716cb40a9c0d681caed6994d81ff4dfad4fe50cea15cd47b885dc5a6",
   "dracut-059.tar.gz": "eabf0bb685420c1e1d5475b6855ef787104508f0135ff570312845256e0fcecf",
   "lgpl-2.1.txt": "dc626520dcd53a22f727af3ee42c770e56c97a64fe3adb063799d8ab032fe551",
   "megaraid.conf": "914824cdbe0c525b71efa05a75e453335b0068beb8bc28bef2a5866d74bf7dd4",
   "module-setup.sh": "8f5a0d3cc393d78bcb523b0d53c578d2767d688f00e84b525355bbb31d753da4",
   "overlayfs-mount.sh": "f049d0ad2e4a1d98acb329765900690883520fdc8fc9463f5081f6e7fb679695"
  }
-}
+}
```
christopherco, line 14: nit - newline at EOF

SPECS/dracut/dracut.spec
```diff
@@ -75,6 +79,20 @@ Requires:       nss
 This package requires everything which is needed to build an
 initramfs with dracut, which does an integrity check.
 
+%package hostonly
```
christopherco, line 82: Isn't there an implied linkage between these dracut subpackages and whether the kernel supplies the underlying modules? Should each kernels "Requires" their relevant dracut subpackages?

SPECS/dracut/00-hostonly.conf
Camelron: I think we would make the kdump initrd gen logic pass -H to explictly turn on hostonly. We would then not install hostonly.conf in our image configs.

SPECS/dracut/dracut.spec
Camelron: A given kernel doesn't require any module to go in initramfs. A given image being deployed on a certain **platform** is what defines the need for certain modules. Hence, we are choosing to compose the set of dracut packages in the image at the imageconfig level.

cwize1 (approved):

SPECS/dracut/dracut.spec
christopherco: Makes sense about platforms defining what modules are in the initramfs. As we were chatting offline and you should consider if a "Requires: dracut >= X" in the kernel spec would help protect against regressions due to kernel module changes in kernel package updates

SPECS/dracut/00-hostonly.conf
christopherco: Makes sense - I'll add this in my kdump changes

christopherco (approved):

PR #9329, opened 2024-06-05
Improve user/group creation in image build

ellie-di:
This PR is a combination of the following changes that's been in review process once:
1- https://github.com/microsoft/azurelinux/pull/6957
2- https://github.com/microsoft/azurelinux/pull/6876

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR accomplishes the following goals:
1- Add support to specify home directory and primary group when creating users. We needed this to overwrite those defaults for some of the service's users that are predefined in an Azure Linux derivative image. Also, added a validation step in imageconfigvalidator to ensure "shadow-utils" package is added to the packages list. 
2- Add support to define a list of users and groups when building a rootFS image type and do it earlier in the image build stage. The reason for shifting users/groups creation to an earlier time is that some of the upstream packages also define their own users or groups, and the chances are some of the users/groups get defined by installing packages in the build chroot and later on addUsers/addGroups will fail because some of the users/groups are already defined. Also, added a validation step in imageconfigvalidator to enforce each user can only belong to the groups that are defined in the config.

###### Change Log  

- Change
- Change
- Change

**NO**

###### Test Methodology

Tested locally to build an image with a list of users with predefined home directory and primary group

### DIFF
--- a/toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
+++ b/toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
@@ -85,6 +85,11 @@ func ValidateConfiguration(config configuration.Config) (err error) {
 		return
 	}
 
+	err = validateUsersGroups(config)
+	if err != nil {
+		return
+	}
+
 	err = validateKickStartInstall(config)
 	return
 }
@@ -118,6 +123,7 @@ func validatePackages(config configuration.Config) (err error) {
 		verityDebugPkgName = "verity-read-only-root-debug-tools"
 		dracutFipsPkgName  = "dracut-fips"
 		fipsKernelCmdLine  = "fips=1"
+		userAddPkgName     = "shadow-utils"
 	)
 
 	for _, systemConfig := range config.SystemConfigs {
@@ -129,6 +135,7 @@ func validatePackages(config configuration.Config) (err error) {
 		foundVerityInitramfsPackage := false
 		foundVerityInitramfsDebugPackage := false
 		foundDracutFipsPackage := false
+		foundUserAddPackage := false
 		kernelCmdLineString := systemConfig.KernelCommandLine.ExtraCommandLine
 		selinuxPkgName := systemConfig.KernelCommandLine.SELinuxPolicy
 		if selinuxPkgName == "" {
@@ -151,6 +158,9 @@ func validatePackages(config configuration.Config) (err error) {
 			if pkg == selinuxPkgName {
 				foundSELinuxPackage = true
 			}
+			if pkg == userAddPkgName {
+				foundUserAddPackage = true
+			}
 		}
 		if systemConfig.ReadOnlyVerityRoot.Enable {
 			if !foundVerityInitramfsPackage {
@@ -168,6 +178,47 @@ func validatePackages(config configuration.Config) (err error) {
 		if systemConfig.KernelCommandLine.SELinux != configuration.SELinuxOff {
 			if !foundSELinuxPackage {
 				return fmt.Errorf("%s: [SELinux] selected, but '%s' package is not included in the package lists", validateError, selinuxPkgName)
+			}
+		}
+		if len(systemConfig.Users) > 0 || len(systemConfig.Groups) > 0 {
+			if !foundUserAddPackage {
+				return fmt.Errorf("%s: add users require '%s' package that is not included in the package lists", validateError, userAddPkgName)
+			}
+		}
+	}
+
+	return
+}
+
+// Function to check if a user belongs to groups that are defined
+func userBelongsToGroup(user configuration.User, groups map[string]bool) bool {
+	if !groups[user.PrimaryGroup] {
+		return false
+	}
+
+	for _, group := range user.SecondaryGroups {
+		if !groups[group] {
+			return false
+		}
+	}
+
+	return true
+}
+
+func validateUsersGroups(config configuration.Config) (err error) {
+	timestamp.StartEvent("validate users groups", nil)
+	defer timestamp.StopEvent(nil)
+
+	for _, systemConfig := range config.SystemConfigs {
+		// Create a map of valid group IDs for quick lookup
+		groupMap := make(map[string]bool)
+		for _, group := range systemConfig.Groups {
+			groupMap[group.GID] = true
+		}
+		// Check each user
+		for _, user := range systemConfig.Users {
+			if !userBelongsToGroup(user, groupMap) {
+				fmt.Errorf("User %s (UID: %s) does not belong to any valid group\n", user.Name, user.UID)
 			}
 		}
 	}
--- a/toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
+++ b/toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
@@ -177,7 +177,7 @@ func validatePackages(config configuration.Config) (err error) {
 		}
 		if len(systemConfig.Users) > 0 || len(systemConfig.Groups) > 0 {
 			if !foundUserAddPackage {
-				return fmt.Errorf("%s: add users require '%s' package that is not included in the package lists", validateError, userAddPkgName)
+				return fmt.Errorf("%s: the '%s' package must be included in the package lists when the image is configured to add users or groups", validateError, userAddPkgName)
 			}
 		}
 	}
--- a/toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
+++ b/toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
@@ -85,11 +85,6 @@ func ValidateConfiguration(config configuration.Config) (err error) {
 		return
 	}
 
-	err = validateUsersGroups(config)
-	if err != nil {
-		return
-	}
-
 	err = validateKickStartInstall(config)
 	return
 }
@@ -183,42 +178,6 @@ func validatePackages(config configuration.Config) (err error) {
 		if len(systemConfig.Users) > 0 || len(systemConfig.Groups) > 0 {
 			if !foundUserAddPackage {
 				return fmt.Errorf("%s: add users require '%s' package that is not included in the package lists", validateError, userAddPkgName)
-			}
-		}
-	}
-
-	return
-}
-
-// Function to check if a user belongs to groups that are defined
-func userBelongsToGroup(user configuration.User, groups map[string]bool) bool {
-	if !groups[user.PrimaryGroup] {
-		return false
-	}
-
-	for _, group := range user.SecondaryGroups {
-		if !groups[group] {
-			return false
-		}
-	}
-
-	return true
-}
-
-func validateUsersGroups(config configuration.Config) (err error) {
-	timestamp.StartEvent("validate users groups", nil)
-	defer timestamp.StopEvent(nil)
-
-	for _, systemConfig := range config.SystemConfigs {
-		// Create a map of valid group IDs for quick lookup
-		groupMap := make(map[string]bool)
-		for _, group := range systemConfig.Groups {
-			groupMap[group.GID] = true
-		}
-		// Check each user
-		for _, user := range systemConfig.Users {
-			if !userBelongsToGroup(user, groupMap) {
-				fmt.Errorf("User %s (UID: %s) does not belong to any valid group\n", user.Name, user.UID)
 			}
 		}
 	}
--- a/toolkit/tools/imageconfigvalidator/imageconfigvalidator_test.go
+++ b/toolkit/tools/imageconfigvalidator/imageconfigvalidator_test.go
@@ -301,3 +301,44 @@ func TestShouldSucceedSELinuxPackageDefinedInline(t *testing.T) {
 	}
 	assert.Fail(t, "Could not find "+targetPackage+" to test")
 }
+
+func TestShouldFailMissingShadowUtilsPackageWithUsers(t *testing.T) {
+	const (
+		configDirectory   = "../../imageconfigs/"
+		targetPackage     = "core-efi.json"
+		targetPackageList = "core-packages-image.json"
+	)
+	configFiles, err := os.ReadDir(configDirectory)
+	assert.NoError(t, err)
+
+	// Pick the core-efi config file, then add a user, then remove shadow-utils from the package list (via dropping core... its a bit hacky)
+	for _, file := range configFiles {
+		if !file.IsDir() && strings.Contains(file.Name(), targetPackage) {
+			configPath := filepath.Join(configDirectory, file.Name())
+
+			fmt.Println("Corrupting ", configPath)
+
+			config, err := configuration.LoadWithAbsolutePaths(configPath, configDirectory)
+			for i, list := range config.SystemConfigs[0].PackageLists {
+				// Delete the packagelist from the config
+				if strings.Contains(list, targetPackageList) {
+					config.SystemConfigs[0].PackageLists = append(config.SystemConfigs[0].PackageLists[:i], config.SystemConfigs[0].PackageLists[i+1:]...)
+				}
+			}
+			assert.NoError(t, err)
+
+			config.SystemConfigs[0].Users = []configuration.User{
+				{
+					Name: "testuser",
+				},
+			}
+
+			err = ValidateConfiguration(config)
+			assert.Error(t, err)
+			assert.Equal(t, "failed to validate package lists in config: add users require 'shadow-utils' package that is not included in the package lists", err.Error())
+
+			return
+		}
+	}
+	assert.Fail(t, "Could not find "+targetPackage+" to test")
+}
--- a/toolkit/tools/imageconfigvalidator/imageconfigvalidator_test.go
+++ b/toolkit/tools/imageconfigvalidator/imageconfigvalidator_test.go
@@ -335,7 +335,7 @@ func TestShouldFailMissingShadowUtilsPackageWithUsers(t *testing.T) {
 
 			err = ValidateConfiguration(config)
 			assert.Error(t, err)
-			assert.Equal(t, "failed to validate package lists in config: add users require 'shadow-utils' package that is not included in the package lists", err.Error())
+			assert.Equal(t, "failed to validate package lists in config: the 'shadow-utils' package must be included in the package lists when the image is configured to add users or groups", err.Error())
 
 			return
 		}
--- a/toolkit/tools/imagecustomizerapi/user.go
+++ b/toolkit/tools/imagecustomizerapi/user.go
@@ -19,6 +19,7 @@ type User struct {
 	PrimaryGroup        string    `yaml:"primaryGroup"`
 	SecondaryGroups     []string  `yaml:"secondaryGroups"`
 	StartupCommand      string    `yaml:"startupCommand"`
+	HomeDirectory       string    `yaml:"HomeDirectory"`
 }
 
 func (u *User) IsValid() error {
--- a/toolkit/tools/imagecustomizerapi/user.go
+++ b/toolkit/tools/imagecustomizerapi/user.go
@@ -19,7 +19,7 @@ type User struct {
 	PrimaryGroup        string    `yaml:"primaryGroup"`
 	SecondaryGroups     []string  `yaml:"secondaryGroups"`
 	StartupCommand      string    `yaml:"startupCommand"`
-	HomeDirectory       string    `yaml:"HomeDirectory"`
+	HomeDirectory       string    `yaml:"homeDirectory"`
 }
 
 func (u *User) IsValid() error {
--- a/toolkit/tools/imagegen/configuration/systemconfig.go
+++ b/toolkit/tools/imagegen/configuration/systemconfig.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/asaskevich/govalidator"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/sliceutils"
 )
 
 // SystemConfig defines how each system present on the image is supposed to be configured.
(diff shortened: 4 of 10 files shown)


snambakam (approved):

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -493,6 +494,15 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		return
 	}
 
+	if len(config.Users) > 0 || len(config.Groups) > 0 {
```
dmcilvaney, line 497: Maybe something along these lines? Or failing that ensure that `shadow-utils` is in the install list in code.

```suggestion
	// imageconfigvalidator should have ensured that we intend to install shadow-utils, so we ca go ahead and do that here.
	if len(config.Users) > 0 || len(config.Groups) > 0 {
```

dmcilvaney: Or
```suggestion
	if len(config.Users) > 0 || len(config.Groups) > 0 {
		if ! sliceutils.ContainsValue(packagesToInstall, "shadow-utils") {
			err = fmt.Errorf("shadow-utils package is required when setting users or groups")
			return
		}
```

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -493,6 +494,15 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
...
+		shadowUtilsInstalled := 0
+		shadowUtilsInstalled, err = TdnfInstallWithProgress(shadowUtilsPkg, installRoot, packagesInstalled, totalPackages, true)
+		if err != nil {
+			return
+		}
+		packagesInstalled += shadowUtilsInstalled
```
dmcilvaney, line 503: Looks like `TdnfInstallWithProgress()` takes care of incrementing the number of packages if you pass the current number in.
```suggestion
		packagesInstalled, err = TdnfInstallWithProgress(shadowUtilsPkg, installRoot, packagesInstalled, totalPackages, true)
		if err != nil {
			return
		}
```

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -493,6 +494,15 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
...
+	}
```
dmcilvaney, line 504: Bit of a nit, and we don't do this for the filesystem above so... 🤷
Also not 100% sure this works (I think its good though...)
```suggestion
		// Remove shadow-utils from the list of packages to install
		packagesToInstall = sliceutils.FindMatches(packagesToInstall, func(pkg string) bool { return pkg != "shadow-utils" })

	}
```

toolkit/tools/imagegen/installutils/installutils.go
ellie-di: Done.

toolkit/tools/imagegen/installutils/installutils.go
ellie-di: Applied this suggestion anyway and confirmed it works.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -493,6 +495,22 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
...
+			return
+		}
+
+		packagesInstalled, err = TdnfInstallWithProgress(shadowUtilsPkg, installRoot, packagesInstalled, totalPackages, true)
+		if err != nil {
+			return
+		}
+
+		// Remove shadow-utils from the list of packages to install
+		packagesToInstall = sliceutils.FindMatches(packagesToInstall, func(pkg string) bool { return pkg != "shadow-utils" })
```
dmcilvaney, line 511: Sorry, forgot there was a new const. We should use that.
```suggestion
		packagesToInstall = sliceutils.FindMatches(packagesToInstall, func(pkg string) bool { return pkg != shadowUtilsPkg })
```

jslobodzian (approved):

toolkit/tools/internal/userutils/userutils.go
```diff
@@ -76,14 +76,20 @@ func UserExists(username string, installChroot safechroot.ChrootInterface) (bool
 	return userExists, nil
 }
 
-func AddUser(username string, hashedPassword string, uid string, installChroot safechroot.ChrootInterface) error {
+func AddUser(username string, homeDir string, primaryGroup string, hashedPassword string, uid string, installChroot *safechroot.Chroot) error {
```
cwize1, line 79: Why did you change `installChroot` from `ChrootInterface` to `*safechroot.Chroot`? Please revert.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -502,6 +520,18 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		}
 	}
 
+	// Add groups
+	err = addGroups(installChroot, config.Groups)
```
cwize1, line 524: What it intentional to move `addGroups` out of the `if !config.IsRootFS()` check?

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1380,6 +1398,9 @@ func addGroups(installChroot *safechroot.Chroot, groups []configuration.Group) (
 		err = installChroot.UnsafeRun(func() error {
 			return shell.ExecuteLive(squashErrors, "groupadd", args...)
 		})
+		if err != nil {
+			return fmt.Errorf("failed to add group: %w", err)
```
cwize1, line 1402: nit: `"failed to add group:\n%w"`

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -1476,7 +1497,7 @@ func createUserWithPassword(installChroot *safechroot.Chroot, user configuration
 		}
 		isRoot = true
 	} else {
-		err = userutils.AddUser(user.Name, hashedPassword, user.UID, installChroot)
+		err = userutils.AddUser(user.Name, user.HomeDirectory, user.PrimaryGroup, hashedPassword, user.UID, installChroot)
```
cwize1, line 1500: If you are going to set the user's primary group here, then you should skip the `ConfigureUserGroupMembership` call in `addUsers` when the user is added (as opposed to the user being updated).

toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -66,6 +67,34 @@ func (s *SystemConfig) GetMountpointPartitionSetting(mountPoint string) (partiti
...
+				logger.Log.Warnf("Primary group (%s) for user (%s) not defined", user.PrimaryGroup, user.Name)
+			}
+		}
+
+		for _, group := range user.SecondaryGroups {
+			if !sliceutils.Contains(s.Groups, group, groupMatchFunc) {
+				// Unclear how to validate these while allowing users to be added to existing system groups. If we
+				// define a group in the config that already exists it will cause errors.
+				// Maybe we can scrape /etc/group and /etc/passwd from filesystem.sepc to validate these?
+				logger.Log.Warnf("secondary group (%s) for user (%s) not defined", group, user.Name)
```
cwize1, line 90: nit: `Secondary`. (Logs start with an uppercase letter. Errors start with a lowercase letter.)

toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -66,6 +67,34 @@ func (s *SystemConfig) GetMountpointPartitionSetting(mountPoint string) (partiti
...
+		groupMatchFunc := func(groupName interface{}, groupObj interface{}) bool {
+			return groupName == groupObj.(Group).Name
+		}
+
+		if user.PrimaryGroup != "" {
+			if !sliceutils.Contains(s.Groups, user.PrimaryGroup, groupMatchFunc) {
+				// Unclear how to validate these while allowing users to be added to existing system groups. If we
+				// define a group in the config that already exists it will cause errors.
+				// Maybe we can scrape /etc/group and /etc/passwd from filesystem.sepc to validate these?
+				logger.Log.Warnf("Primary group (%s) for user (%s) not defined", user.PrimaryGroup, user.Name)
```
cwize1, line 81: "not defined" is probably specific enough. Perhaps, "not defined in SystemConfig.Groups"?

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
```diff
@@ -315,7 +315,7 @@ func addOrUpdateUser(user imagecustomizerapi.User, baseConfigPath string, imageC
 		}
 
 		// Add the user.
-		err = userutils.AddUser(user.Name, hashedPassword, uidStr, imageChroot)
+		err = userutils.AddUser(user.Name, user.HomeDirectory, user.PrimaryGroup, hashedPassword, uidStr, imageChroot.(*safechroot.Chroot))
```
cwize1, line 318: You need to handle the home directory for the case where the user is updated instead of added.

If it isn't possible (or practical) to move the home directory, then you should report an error if the home directory value is specified.

cwize1, line 318: The `ConfigureUserGroupMembership` call below should be skipped if the user is added.

cwize1 (changes requested):

toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
```diff
@@ -170,6 +175,11 @@ func validatePackages(config configuration.Config) (err error) {
 				return fmt.Errorf("%s: [SELinux] selected, but '%s' package is not included in the package lists", validateError, selinuxPkgName)
 			}
 		}
+		if len(systemConfig.Users) > 0 || len(systemConfig.Groups) > 0 {
+			if !foundUserAddPackage {
+				return fmt.Errorf("%s: add users require '%s' package that is not included in the package lists", validateError, userAddPkgName)
```
PawelWMS, line 180: Suggestion to make it clearer for the user what's missing:
```suggestion
				return fmt.Errorf("%s: the '%s' package must be included in the package lists when the image is configured to add users or groups", validateError, userAddPkgName)
```

toolkit/tools/imageconfigvalidator/imageconfigvalidator_test.go
```diff
@@ -301,3 +301,44 @@ func TestShouldSucceedSELinuxPackageDefinedInline(t *testing.T) {
...
+
+			err = ValidateConfiguration(config)
+			assert.Error(t, err)
+			assert.Equal(t, "failed to validate package lists in config: add users require 'shadow-utils' package that is not included in the package lists", err.Error())
+
+			return
+		}
+	}
+	assert.Fail(t, "Could not find "+targetPackage+" to test")
+}
```
PawelWMS, line 344: Why use non-test files and modify them by code here? Why not just use a test config file, which will be dedicated just for the tests? We're adding unnecessary logic in the test code and mixing test and non-test components.

toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -66,6 +67,34 @@ func (s *SystemConfig) GetMountpointPartitionSetting(mountPoint string) (partiti
 	return FindMountpointPartitionSetting(s.PartitionSettings, mountPoint)
 }
 
+func (s *SystemConfig) validateUsersAndGroups() (err error) {
+	for _, user := range s.Users {
+		groupMatchFunc := func(groupName interface{}, groupObj interface{}) bool {
+			return groupName == groupObj.(Group).Name
+		}
```
PawelWMS, line 74: Is there a need to define `groupMatchFunc` for each iteration of the loop?
```suggestion
	groupMatchFunc := func(groupName interface{}, groupObj interface{}) bool {
		return groupName == groupObj.(Group).Name
	}

	for _, user := range s.Users {
```

toolkit/tools/imagegen/configuration/systemconfig.go
```diff
@@ -66,6 +67,34 @@ func (s *SystemConfig) GetMountpointPartitionSetting(mountPoint string) (partiti
...
+
+		if user.PrimaryGroup != "" {
+			if !sliceutils.Contains(s.Groups, user.PrimaryGroup, groupMatchFunc) {
+				// Unclear how to validate these while allowing users to be added to existing system groups. If we
+				// define a group in the config that already exists it will cause errors.
+				// Maybe we can scrape /etc/group and /etc/passwd from filesystem.sepc to validate these?
```
PawelWMS, line 80: Can you be a bit more clear with the comment? What do you mean by "these" in "Unclear how to validate these" for instance? Also, I'm not sure, if the last comments is helpful. If we can check the files, let's do it. If not, why add the comment?

toolkit/tools/imagegen/configuration/systemconfig.go
PawelWMS: Also add a period at the end, please.

toolkit/tools/imagegen/configuration/systemconfig.go
PawelWMS, line 81: I'd also add a bit of explanation to the warning saying what is expected and what will happen, if the expectation is not met.

toolkit/tools/imagegen/configuration/user.go
```diff
@@ -25,6 +25,7 @@ type User struct {
 	PrimaryGroup        string   `json:"PrimaryGroup"`
 	SecondaryGroups     []string `json:"SecondaryGroups"`
 	StartupCommand      string   `json:"StartupCommand"`
+	HomeDirectory       string   `json:"HomeDirectory"`
```
PawelWMS, line 28: Is there documentation we should extend for this new field?

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -493,6 +495,22 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		return
 	}
 
+	// imageconfigvalidator should have ensured that we intend to install shadow-utils, so we can go ahead and do that here.
+	if len(config.Users) > 0 || len(config.Groups) > 0 {
+		if !sliceutils.ContainsValue(packagesToInstall, "shadow-utils") {
+			err = fmt.Errorf("shadow-utils package is required when setting users or groups")
```
PawelWMS, line 501: I'd mention that `shado-utils` must be listed explicitly. Otherwise, it's going to be confusing, if someone adds a package, which pulls in `shadow-utils`.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -493,6 +495,22 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
...
+		}
+
+		packagesInstalled, err = TdnfInstallWithProgress(shadowUtilsPkg, installRoot, packagesInstalled, totalPackages, true)
+		if err != nil {
+			return
+		}
+
+		// Remove shadow-utils from the list of packages to install
+		packagesToInstall = sliceutils.FindMatches(packagesToInstall, func(pkg string) bool { return pkg != shadowUtilsPkg })
+	}
```
PawelWMS, line 512: This section looks long and complex enough to move to a separate function.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -493,6 +495,22 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 		return
 	}
 
+	// imageconfigvalidator should have ensured that we intend to install shadow-utils, so we can go ahead and do that here.
+	if len(config.Users) > 0 || len(config.Groups) > 0 {
+		if !sliceutils.ContainsValue(packagesToInstall, "shadow-utils") {
+			err = fmt.Errorf("shadow-utils package is required when setting users or groups")
+			return
```
PawelWMS, line 502: Doesn't this belong in the config validation code earlier on?

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -493,6 +495,22 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
...
+		}
+
+		packagesInstalled, err = TdnfInstallWithProgress(shadowUtilsPkg, installRoot, packagesInstalled, totalPackages, true)
+		if err != nil {
+			return
+		}
+
+		// Remove shadow-utils from the list of packages to install
+		packagesToInstall = sliceutils.FindMatches(packagesToInstall, func(pkg string) bool { return pkg != shadowUtilsPkg })
```
PawelWMS, line 511: Why do we need to install `shadow-utils` separately? And why do we remove it later from `packagesToInstall`?

toolkit/tools/imagegen/installutils/installutils.go
PawelWMS: Please also say which group we failed to add. It's not guaranteed that name is going to be mentioned the output from `groupadd`.

toolkit/tools/imageconfigvalidator/imageconfigvalidator_test.go
dmcilvaney: Don't disagree, but I don't want to re-invent how we do testing for the validator right now. This is copy/pasted from another test above.

toolkit/tools/imagegen/configuration/systemconfig.go
dmcilvaney: Readability is better to me, and this is such a small loop the perf uplift didn't seem important. I originally had it as an inline function in the call to sliceutils but it was a bit hard to read.

toolkit/tools/imagegen/configuration/systemconfig.go
dmcilvaney:
```suggestion
				// Unclear how to validate that users and groups match together correctly while still allowing
				// users to be added to existing system groups. If we define a group in the config that already
				// exists it will cause errors, but we don't have insight into the default users/groups already in
				// an image before we install the filesystem package.
```

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney, line 501:
```suggestion
			err = fmt.Errorf("shadow-utils package must be added to the image's package lists when setting users or groups")
```

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: We put as much validation as possible natively as part of each config structure. The validator tool will use these functions when it runs its validation, but the config validator's job is to check against external things that aren't encoded in the .json. All the user bits are part of the .json. This is a bit odd since its now just a warning while it was originally an actual error.

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: The point of this change is to add users before installing any other packages. To do that we need shadow-utils (so at best we can try to minimize the number of other packages added). It feels cleaner to not re-install shadowutils if its already installed.

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: Can fix later, we need to get this merged for testing as early as possible.

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney:
```suggestion
			return fmt.Errorf("failed to add group (%s):\n%w",  group.Name, err)
```

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: Unclear why we would want to gate groups behind rootfs while we don't with users.

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: Removed and renamed function to `ConfigureUserSecondaryGroupMembership()`.

toolkit/tools/internal/userutils/userutils.go
dmcilvaney: Sorry, didn't see that when I was going through the changes. Will revert.

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: Actually... that might break MIC...

toolkit/tools/pkg/imagecustomizerlib/customizeutils.go
dmcilvaney: We need to configure secondary groups still, even if the primary group is ok. I'm wondering if the extra code complexity is worth it here to decide if we need to pass "" or `user.PrimaryGroup` in...

dmcilvaney: ok, added conditional `var primaryGroup string`.

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: Split into `ConfigureUserPrimaryGroupMembership()` and `ConfigureUserSecondaryGroupMembership()` so we can target what we need.

toolkit/tools/imagecustomizerapi/user.go
```diff
@@ -19,6 +19,7 @@ type User struct {
 	PrimaryGroup        string    `yaml:"primaryGroup"`
 	SecondaryGroups     []string  `yaml:"secondaryGroups"`
 	StartupCommand      string    `yaml:"startupCommand"`
+	HomeDirectory       string    `yaml:"HomeDirectory"`
```
cwize1, line 22: `homeDirectory`

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/user.go
dmcilvaney:
```suggestion
	HomeDirectory       string    `yaml:"homeDirectory"`
```

cwize1 (approved):

PR #8846, opened 2024-04-19
gcc: set baseline architectures to x86-64-v3 and armv8.1-a

anphel31:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [x] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [x] LICENSE-MAP files are up-to-date (`./SPECS/LICENSES-AND-NOTICES/data/licenses.json`, `./SPECS/LICENSES-AND-NOTICES/LICENSES-MAP.md`, `./SPECS/LICENSES-AND-NOTICES/LICENSE-EXCEPTIONS.PHOTON`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 

What does the PR accomplish, why was it needed?
gcc: set baseline architectures to x86-64-v3 and armv8.1-a. These changes modify our gcc compiler to use newer architectures by default, which contain additional instructions that can lead to performance improvements. Details on each choice of architecture are below. 

x86-64-v3:
This feature level appears to be supported on all currently available Azure virtual machines. It is also available on a few developer machines I tested. Some Azure VMs I tried supported v4, although not my development machines. 
x86 options: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
```
x86-64:         CMOV, CMPXCHG8B, FPU, FXSR, MMX, FXSR, SCE, SSE, SSE2
                First available: ~2003
x86-64-v2:      (close to Nehalem) CMPXCHG16B, LAHF-SAHF, POPCNT, SSE3, SSE4.1, SSE4.2, SSSE3
                First available: ~2008
[...]

PR #9353, opened 2024-06-07
gcc: revert change to baseline arm64 architecture

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Revert change to baseline arm64 architecture from #8846 

###### Change Log  

- gcc: revert change to baseline arm64 architecture

**YES*

###### Associated issues  

- #xxxx

###### Test Methodology

- Full build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=582680&view=results
- Booted the ARM64 ISO on Ampere eMag arm64 machine and ran installed with qemu/kvm


toolkit/scripts/toolchain/container/toolchain_build_temp_tools.sh
```diff
@@ -51,7 +51,7 @@ case $(uname -m) in
     sed -e '/m64=/s/lib64/lib/' -i.orig gcc/config/i386/t-linux64
   ;;
   aarch64)
-    GCC_CONFIG_WITH_ARCH="armv8.1-a"
+    GCC_CONFIG_WITH_ARCH=""
```
christopherco, line 54: Is setting `--with-arch=""` equivalent to not supplying `--with-arch` at all in the configure script? I'm concerned that specifying "" is actually different than what the default value would be (and if true, which one is the correct one we want?)

Unfortunately https://gcc.gnu.org/install/configure.html is light on information about this...

anphel31: Good question. It appears to be the same, although I've made a change so this is not set at all for arm64

anphel31: I tested the build both ways and shared some outputs below.

Configured with: ../configure … --with-arch= …
COLLECT_GCC_OPTIONS='-v' '-mlittle-endian' '-mabi=lp64' '-dumpdir' 'a-'

Configured with: ../configure … (do not call --with-arch at all)
COLLECT_GCC_OPTIONS='-v' '-mlittle-endian' '-mabi=lp64' '-dumpdir' 'a-'

Configured with: ../configure … --with-arch=armv8.1-a …
COLLECT_GCC_OPTIONS='-v' '-march=armv8.1-a' '-mlittle-endian' '-mabi=lp64' '-dumpdir' 'a-'

christopherco (approved):

snambakam (approved):

PR #9366, opened 2024-06-10
User/apais/azlinux 3.0 arm64 config

allenpais:
arm64: Clean up Azure Linux kernel configuration

This commit streamlines the ARM64 kernel configuration for Azure Linux environments by removing unused or redundant configuration options and updating settings to optimize performance and stability. Changes include:

- Disabling unused platforms
- Disabling unused device drivers specific to other platforms.

These adjustments ensure that the ARM64 kernel is better tailored to the operational characteristics of Azure, enhancing both efficiency and the overall user experience in cloud environments.

Signed-off-by: Allen Pais <apais@linux.microsoft.com>


rlmenge (changes requested):
Looks like this is intended for 3.0. Please change the target branch to 3.0-dev :)

tyhicks:
@allenpais how about disabling `CONFIG_HWPOISON_INJECT` on arm64 to match what we're doing on x86_64? It is a testing feature that doesn't need to be enabled in production.

PR #9400, opened 2024-06-13
Fix ordering issues with new preview repo handling in derivative builds

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Fix an issue for derivative builds using `make package-toolkit` if the destination didn't have a `./SPECS/.../*.repo file` file to grab, and `make` thought the existing version was out of date.

###### Change Log  

- Fix USE_PREVIEW_REPO=y for derivative builds.

**NO**

###### Test Methodology

- Local tests

### DIFF
--- a/toolkit/scripts/preview.mk
+++ b/toolkit/scripts/preview.mk
@@ -18,20 +18,24 @@ preview_repo_source := $(PROJECT_ROOT)/SPECS/azurelinux-repos/azurelinux-officia
 
 # Stop immediately if the preview repo file is missing and cannot be created.
 ifeq ($(wildcard $(preview_repo_source))$(wildcard $(preview_repo)),)
-	$(warning )
-	$(warning ######################### ERROR #########################)
-	$(warning 'USE_PREVIEW_REPO=y' is set but all of the following failed:)
-	$(warning - packaged default '$(preview_repo)' is missing)
-	$(warning - backup source    '$(preview_repo_source)' is missing)
-	$(warning ######################### ERROR #########################)
-	$(error )
+    $(warning )
+    $(warning ######################### ERROR #########################)
+    $(warning 'USE_PREVIEW_REPO=y' is set but all of the following failed:)
+    $(warning - packaged default '$(preview_repo)' is missing)
+    $(warning - backup source    '$(preview_repo_source)' is missing)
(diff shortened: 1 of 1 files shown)


cwize1 (approved):

anphel31 (approved):

toolkit/scripts/preview.mk
```diff
@@ -18,20 +18,24 @@ preview_repo_source := $(PROJECT_ROOT)/SPECS/azurelinux-repos/azurelinux-officia
...
+    $(warning - backup source    '$(preview_repo_source)' is missing)
+    $(warning ######################### ERROR #########################)
+    $(error )
 endif
 
-# Copy the preview repo file to the expected repo location.
+# Copy the preview repo file to the expected repo location, but only if we  have a target to copy from.
+ifeq ($(wildcard $(preview_repo_source)),)
+$(preview_repo): ;
+else
```
PawelWMS, line 33: Why is this case needed? In a core repo, `preview_repo_source` is expected to be there, so we should crash, if it's not. In the packaged toolkit case, `preview_repo` is expected to be present, so the target below should be a no-op and if `preview_repo` is missing, then the error above should already catch it, right? I think I'm misunderstanding some `make` magic.

dmcilvaney: I'm not 100% sure why but sometimes that 'no-op' isn't a 'no-op' and it tries to check the timestamps. It can't find the src file to check against so it creates an error.

PawelWMS: I think It might be worth understand that better, because something seems fishy. So many projects use `make` and I think they don't have to do manual checking like we do here. I'm not sure what's wrong, but it feels like we're masking the real issue.

dmcilvaney: I did some more looking, as best I can tell make should always generate an error. I'm less sure why I didn't get one. i.e. if you have a dependency, that file must either be `.PHONY`, exist as a file, or have a rule that creates it. tldr I'm not sure why this ever worked (and I can't reproduce it working anymore).

dmcilvaney: Also, as far as makefile jank goes, I'm not sure this even moves the needle :P

PR #9425, opened 2024-06-18
Patch bash to load startup files when invoked with '-bash'

dmcilvaney:
i.e., when invoked via 'su - <user> -c ...'. This is done by Fedora and Ubuntu (and other distros).

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Bash loads startup scripts when it runs, based on some logic to detect if it is a login/normal prompt, and interactive/non-interactive. (profile, bashrc, etc..., depending on those modes).

Tools like `su` in `util-linux` can explicitly request login/non-login mode (`su - other_user ...` (`-` == `-l` == `--login`). It informs bash of this choice by  prepending a '-' to `argv[0]` (i.e., `-bash`). By default `bash` ignores this, but can be [configured to respect it via a "hidden" define](https://git.savannah.gnu.org/cgit/bash.git/tree/config-top.h#n105) `NON_INTERACTIVE_LOGIN_SHELLS`. Most distros set this flag, so lets add it.

In containers:
```bash
tdnf -y install util-linux shadow-utils # 'dnf -y install util-linux' on fedora, ubuntu works out of box
useradd -m test
echo "export test=test" >> /etc/profile
su test -c 'echo $test' # Expected to be ""
su - test -c 'echo $test' # SHOULD BE "TEST", this works on ubuntu/fedora!
```
image

###### Change Log  

- Enable `NON_INTERACTIVE_LOGIN_SHELLS` for bash package.

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/46864804

###### Test Methodology

- Local build/test in vm & container
- Buddybuild: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=588741&view=results (uninstall test fails since removing bash from the env. breaks the test).

### DIFF
--- a/SPECS/bash/bash.spec
+++ b/SPECS/bash/bash.spec
@@ -332,7 +332,7 @@ fi
 %changelog
 * Mon Jun 17 2024 Daniel McIlvaney <damcilva@microsoft.com> - 5.2.15-2
 - When non-interactive shells are started with '-bash' load startup files. From
-  Fedora upstream: https://src.fedoraproject.org/rpms/bash/blob/f40/f/bash-2.03-profile.patch
+- Fedora upstream: https://src.fedoraproject.org/rpms/bash/blob/f40/f/bash-2.03-profile.patch
 
 * Tue Nov 21 2023 Andrew Phelps <anphel@microsoft.com> - 5.2.15-1
 - Upgrade to version 5.2.15
--- a/SPECS/bash/bash.spec
+++ b/SPECS/bash/bash.spec
@@ -1,7 +1,7 @@
 Summary:        Bourne-Again SHell
 Name:           bash
 Version:        5.2.15
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv3
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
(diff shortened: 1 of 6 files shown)


SPECS/bash/bash.spec
```diff
@@ -328,6 +330,10 @@ fi
 %defattr(-,root,root)
 
 %changelog
+* Mon Jun 17 2024 Daniel McIlvaney <damcilva@microsoft.com> - 5.2.15-2
+- When non-interactive shells are started with '-bash' load startup files. From
+  Fedora upstream: https://src.fedoraproject.org/rpms/bash/blob/f40/f/bash-2.03-profile.patch
```
anphel31, line 335: can you add a dash to the start of this comment?

anphel31 (approved):
I re-ran the buddybuild with the option set to not remove bash, to confirm it's passing - https://dev.azure.com/mariner-org/mariner/_build/results?buildId=589017&view=results

SPECS/bash/bash.spec
dmcilvaney: This is auto-formatted via `rpmdev-bumpspec`, but I can add.

SPECS/bash/bash.spec
```diff
@@ -328,6 +330,10 @@ fi
...
+
```
dmcilvaney, line 336:
```suggestion
* Mon Jun 17 2024 Daniel McIlvaney <damcilva@microsoft.com> - 5.2.15-2
- When non-interactive shells are started with '-bash' load startup files. From
- Fedora upstream: https://src.fedoraproject.org/rpms/bash/blob/f40/f/bash-2.03-profile.patch

```

PR #9448, opened 2024-06-19
Add support for generating confidential virtual machine images

trungams:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

- Introducing a set of changes for generating Azl3 CVM image.

###### Change Log  

- kernel-uki: include i18n dracut module so UKI systemd-vconsole-setup service works
- toolkit: add support for setting partition type UUID via image config
- image config: add CVM image definition

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/51858044/

###### Test Methodology

- Local build produces a VHD image that boots successfully on Azure CVM instance.


toolkit/imageconfigs/scripts/setup_cvm_image.sh
```diff
@@ -0,0 +1,31 @@
...
+}
+
+# symlink /boot/efi to ../efi
+cp -a /boot/efi/. /efi
+rm -rf /boot/efi
+ln -s ../efi /boot/efi
+
+# switching to systemd-boot
+# copy sd-boot EFI binary over the grub EFI binary
+cp /lib/systemd/boot/efi/systemd-bootx64.efi /efi/EFI/$EFIDIR/grubx64.efi
```
ms-mahuber, line 22: if there's a dependency on how things are setup/installed in another packages, let's mention that here.

toolkit/imageconfigs/scripts/setup_cvm_image.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+set -e
+
+EFIDIR="BOOT"
+KERNEL_VERSION=""
+
+get_kernel_version() {
```
ms-mahuber, line 10: looks a bit hacky. can we think of a more stable way. looks like we want this for "vmlinuz-uki.efi".

Does it make sense to either query using a command, or even add the version to "vmlinuz-uki.efi" when we build the respective package?

ms-mahuber: Ex. in UKI SPEC build and file section
%files
/lib/modules/%{kernelver}/vmlinuz-uki.efi

=> %{kernelver} as file name?

+ In this script, can we query the location/version of this binary by checking the version of the UKI package?

ms-mahuber (approved):

toolkit/imageconfigs/cvm.json
```diff
@@ -0,0 +1,73 @@
...
+                {
+                    "ID": "efi",
+                    "Flags": [
+                        "esp",
+                        "boot"
+                    ],
+                    "Start": 1,
+                    "End": 513,
+                    "FsType": "fat32",
+                    "TypeUUID": "c12a7328-f81f-11d2-ba4b-00a0c93ec93b"
```
cwize1, line 22: It might be goodness to also provide an alternative API that provides friendly names for the GUIDs:

e.g.

```json
"Type": "esp"
```

or:

```json
"Type": "linux-root-amd64"
```

toolkit/imageconfigs/scripts/setup_cvm_image.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
cwize1, line 1: Should this code be added directly to the `imager` instead of as a script?

toolkit/imageconfigs/packagelists/cvm-packages.json
```diff
@@ -0,0 +1,13 @@
...
+        "shim",
+        "systemd-boot-unsigned",
+        "kernel-uki",
+        "tpm2-tools",
+        "ca-certificates",
+        "cronie-anacron",
+        "logrotate",
+        "core-packages-base-image"
+    ],
+    "_comment": ""
```
cwize1, line 12: Is there any point to adding an empty comment?

toolkit/imageconfigs/scripts/setup_cvm_image.sh
trungams: This customization is quite specific to the CVM boot flow `UEFI -> shim -> sd-boot -> UKI`, and the setup is subject to change, so I think using this script is easier to maintain, at least for now.

toolkit/imageconfigs/cvm.json
trungams: Agreed, but I don't think we can set that via `parted` at the moment: https://www.gnu.org/software/parted/manual/parted.html#type

trungams: hmm maybe we can maintain our own map for these values..

cwize1: Yeah, my suggestion was to maintain our own list. (But keep `TypeUUID` as a fallback.)

toolkit/imageconfigs/scripts/setup_cvm_image.sh
```diff
@@ -0,0 +1,35 @@
...
+# Image generation is done in a chroot environment, so running `uname -r`
+# will return the version of the host running kernel. This function works
+# under the assumption that exactly one kernel is installed in the end image.
+get_kernel_version() {
+    kernel_modules_dir="/usr/lib/modules"
+    KERNEL_VERSION="$(ls $kernel_modules_dir)"
+}
+
+# symlink /boot/efi to ../efi
+cp -a /boot/efi/. /efi
```
Camelron, line 19: Who expects the esp partition at /efi?

Camelron (approved):

toolkit/imageconfigs/cvm.json
dmcilvaney: Second this, bit of boilerplate but we can add a `filesystemtype.go` (name pending) similar to `partitionflag.go`

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -620,14 +620,22 @@ func InitializeSinglePartition(diskDevPath string, partitionNumber int, partitio
...
+	// Set partition friendly name and partition type UUID (only for gpt)
 	if partitionTableType == "gpt" {
 		partitionName := partition.Name
 		_, stderr, err := shell.Execute("flock", "--timeout", timeoutInSeconds, diskDevPath, "parted", diskDevPath, "--script", "name", partitionNumberStr, partitionName)
 		if err != nil {
 			logger.Log.Warnf("Failed to set partition friendly name using parted: %v", stderr)
 			// Not-fatal
 		}
+
+		if partition.TypeUUID != "" {
```
dmcilvaney, line 632: Just do double check, is this going to generate a reasonable error on old parted if the `type` parameter is not supported, or will it be some cryptic undecipherable error?

(for background 2.0's version of parted is too old to support this if I recall right)

dmcilvaney: We could also consider adding support in `imageconfigvalidator` that makes sure you have a new enough version of `parted` if you set custom values for the type.

trungams: `parted` will just print out its help string to stdout and return exit code 0. The type parameter is effectively ignored.

toolkit/imageconfigs/scripts/setup_cvm_image.sh
trungams: systemd-boot, which is what we're using over grub for CVM images

toolkit/imageconfigs/cvm.json
trungams: Added a `Type` parameter, along with a limited list type name -> type UUID of common types

cwize1 (approved):
Signing off on tooling code.

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -620,14 +620,28 @@ func InitializeSinglePartition(diskDevPath string, partitionNumber int, partitio
...
+				typeUUID = partition.TypeUUID
+			} else {
+				typeUUID = configuration.PartitionTypeNameToUUID[partition.Type]
+			}
+			_, stderr, err := shell.Execute("flock", "--timeout", timeoutInSeconds, diskDevPath, "parted", diskDevPath, "--script", "type", partitionNumberStr, typeUUID)
+			if err != nil {
+				logger.Log.Warnf("failed to set partition type using parted: %v", stderr)
+				// Not-fatal
+			}
+		}
```
PawelWMS, line 644: This function is getting pretty lengthy. Could you extract the setting of partition type (and name above) into helper functions?

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -620,14 +620,28 @@ func InitializeSinglePartition(diskDevPath string, partitionNumber int, partitio
...
+		if partition.TypeUUID != "" || partition.Type != "" {
+			var typeUUID string
+			if partition.TypeUUID != "" {
+				typeUUID = partition.TypeUUID
+			} else {
+				typeUUID = configuration.PartitionTypeNameToUUID[partition.Type]
+			}
+			_, stderr, err := shell.Execute("flock", "--timeout", timeoutInSeconds, diskDevPath, "parted", diskDevPath, "--script", "type", partitionNumberStr, typeUUID)
+			if err != nil {
+				logger.Log.Warnf("failed to set partition type using parted: %v", stderr)
```
PawelWMS, line 641: Nit-pick:
```suggestion
				logger.Log.Warnf("Failed to set partition type using parted: %v", stderr)
```

dmcilvaney (approved):

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -620,14 +620,28 @@ func InitializeSinglePartition(diskDevPath string, partitionNumber int, partitio
...
+				// Not-fatal
```
PawelWMS, line 642: If it's non-fatal, we should also `err = nil`, so that the caller doesn't accidentally crash. We should probably do the same thing in the code above, which is setting the friendly name.

Camelron (approved):

toolkit/tools/imagegen/diskutils/diskutils.go
trungams: done

PR #9494, opened 2024-06-25
fix: remove override_glibc logic from libxcrypt

christopherco:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Now that glibc no longer builds libcrypt.so, we can directly use libxcrypt's version of libcrypt.so
and remove the override macros in the libxcrypt spec.

The override_glibc logic was added by us to workaround an issue with adding this improved version of libcrypt into Mariner 1.0 while it was already released. And in 1.0, 2.0, and pre-Nov 2023 3.0, the macro was %bcond_without override_glibc, but after Nov 2023, the macro was changed to %bcond_with override_glibc which means all these overrides are currently inert and are now just diffs from the original spec.

The only real functional change here is the addition of `Requires: glibc%{?_isa} >= %{glibc_minver}` 

**YES**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/47158360

###### Test Methodology

- Pipeline build id: [592855](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=592855&view=results)

### DIFF
--- a/SPECS/libxcrypt/libxcrypt.spec
+++ b/SPECS/libxcrypt/libxcrypt.spec
@@ -46,9 +46,6 @@
 # Fail linking if there are undefined symbols.
 # Required for proper ELF symbol versioning support.
 %global _ld_strict_symbol_defs 1
-# override_glibc and glibcversion are temporary to make libxcrypt install on top of glibc
-%define glibcversion 2.38
-%bcond_with override_glibc
 # Build the static library?
 %bcond_with new_api
 %bcond_with compat_pkg
@@ -102,7 +99,7 @@
 Summary:        Extended crypt library for descrypt, md5crypt, bcrypt, and others
 Name:           libxcrypt
 Version:        4.4.36
-Release:        1%{?dist}
+Release:        2%{?dist}
 # For explicit license breakdown, see the
 # LICENSING file in the source tarball.
 License:        LGPLv2+ AND BSD AND Public Domain
@@ -142,12 +139,7 @@ Provides:       %{name}-common = %{version}-%{release}
 Obsoletes:      %{name}-compat < %{version}-%{release}
 %endif
 # We need a version of glibc, that doesn't build libcrypt anymore.
-#Requires:       glibc%{?_isa}         >= %{glibc_minver}
-%if %{with override_glibc}
-# Require a specific glibc version so the post macro is compatible.
-BuildRequires:  glibc-devel = %{glibcversion}
-Requires:       glibc = %{glibcversion}
-%endif
+Requires:       glibc%{?_isa} >= %{glibc_minver}
 %if 0%{?fedora} >= 30
 Recommends:     mkpasswd
 %endif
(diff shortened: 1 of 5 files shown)


anphel31:
Looks good, but can we run a toolchain build for extra validation? The buddybuild environment is slightly different. (you might need to push a branch to this repo to run the toolchain build)

SPECS/libxcrypt/libxcrypt.spec
```diff
@@ -142,12 +142,7 @@ Provides:       %{name}-common = %{version}-%{release}
 Obsoletes:      %{name}-compat < %{version}-%{release}
 %endif
 # We need a version of glibc, that doesn't build libcrypt anymore.
-#Requires:       glibc%{?_isa}         >= %{glibc_minver}
-%if %{with override_glibc}
```
PawelWMS, line 146: Could you also remove these few lines from the spec, since they aren't used anymore:
```
# override_glibc and glibcversion are temporary to make libxcrypt install on top of glibc
%define glibcversion 2.38
%bcond_with override_glibc
```

SPECS/libxcrypt/libxcrypt.spec
```diff
@@ -142,12 +142,7 @@ Provides:       %{name}-common = %{version}-%{release}
...
-# Require a specific glibc version so the post macro is compatible.
-BuildRequires:  glibc-devel = %{glibcversion}
-Requires:       glibc = %{glibcversion}
-%endif
+Requires:       glibc%{?_isa}         >= %{glibc_minver}
```
PawelWMS, line 145: Not that it matters much but why such a huge gap before `>=`? The rest of the spec seems to be using a single space, so it caught my eye.

SPECS/libxcrypt/libxcrypt.spec
PawelWMS: I only noticed this right now: since we've been using `%bcond_with override_glibc` in the spec, it looks like `%{with override_glibc}` sections would always be skipped. We don't set any `with X` macros, if I'm not mistaken. I'm fine doing the clean-up, if it's only specific to our spec, but I just wanted to point this out, in case we're expecting any functional changes after this PR is in.

SPECS/libxcrypt/libxcrypt.spec
christopherco: The huge gap is actually in the upstream fedora spec, and is also present in many of the other lines. https://src.fedoraproject.org/rpms/libxcrypt/blob/f31/f/libxcrypt.spec

I suspect some linting was applied to our version of the spec, so the spaces and newlines were condensed/replaced.

I will adjust this line to match the rest of the spec.

SPECS/libxcrypt/libxcrypt.spec
christopherco: Yep will remove those lines.
All this override_glibc logic was added by us to workaround an issue with adding this improved version of `libcrypt` into 1.0 while it was already released. And in 1.0, 2.0, and pre-Nov 2023 3.0, the line was `%bcond_without override_glibc`, but you're right that today `%bcond_with override_glibc` means all these overrides are inert and are diffs from the original spec.

The only real functional change is the requires glibc that you highlighted.

christopherco:
> Looks good, but can we run a toolchain build for extra validation? The buddybuild environment is slightly different. (you might need to push a branch to this repo to run the toolchain build)

Of course! Queued the toolchain build - https://dev.azure.com/mariner-org/mariner/_build/results?buildId=593711&view=results

christopherco:
Toolchain and RPM builds passed:
- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=593718&view=results
- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=593719&view=results

PawelWMS (approved):

PR #9509, opened 2024-06-25
Clear systemd state when creating an image

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

The systemd package creates `/etc/machine-id` when it installs (in m2 it did not create this file). If this file is populated on a vm image, every instance of that vm will have the same id. This will break networking, k8s, etc. To avoid this, we need to clear the state before creating the image. The correct first-boot behavior is to set the id to 'uninitialized' and let the system generate a new one on boot. Clearing the file is similar, systemd will generate a new id on boot, but it will not run the first-boot setup.

We will keep the current behavior with `/etc/machine-id` being empty. A new config flag will allow an image to use the first-boot flow.

As part of https://systemd.io/BUILDING_IMAGES several other files are listed that should be cleared out. Currently we don't seem to have them but for correctness I've also included a check that will delete them if found. 

###### Change Log  

- replace `addMachineId()` with `clearSystemdState()`
  - Set `/etc/machine-id` to `uninitalized\n` if `EnableSystemdFirstboot=true` , otherwise keep the 2.0 behavior of creating an empty file
  - Also clear `/var/lib/systemd/random-seed`, `boot/efi/loader/random-seed`, `/var/lib/systemd/credential.secret` if any are in the image.

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_queries/edit/51856148/?triage=true

###### Test Methodology

- Local builds with basic validation (boots, login possible, systemd seems ok)
  - Local hyper-v test with core-efi (machine-id is generated, firstboot runs)
  - MIC based dm-verity read-only test (machine-id is ephemeral but generates as expected, firstboot does not run since /etc is not writable)
  - Azure marketplace gen2 (same as local hyper-v)

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -873,6 +873,7 @@ func clearSystemdState(installChroot *safechroot.Chroot, enableSystemdFirstboot 
 			return err
 		}
 
+		// Do an explicit check for existence so we can log the file removal.
 		if exists {
 			ReportActionf("Removing systemd state file (%s)", filePath)
 			err = file.RemoveFileIfExists(fullPath)
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -863,7 +863,8 @@ func clearSystemdState(installChroot *safechroot.Chroot, enableSystemdFirstboot 
 		return err
 	}
 
-	// These files should not be present in the image, but we should be thorough and double-check.
+	// These files should not be present in the image, but per https://systemd.io/BUILDING_IMAGES/ we should
+	// be thorough and double-check.
 	for _, filePath := range otherFilesToRemove {
 		fullPath := filepath.Join(installChroot.RootDir(), filePath)
 		exists, err = file.PathExists(fullPath)
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -834,7 +834,7 @@ func clearSystemdState(installChroot *safechroot.Chroot, enableSystemdFirstboot 
 	// create an empty file so that read-only configurations will work as expected. If the user requests that firstboot
 	// be enabled we will set it to "uninitalized" as per option 3).
 
-	ReportAction("Clearing systemd state files for first boot")
+	ReportAction("Configuring systemd state files for first boot")
 
 	// The systemd package will create this file, but if its not installed, we need to create it.
 	exists, err := file.PathExists(filepath.Join(installChroot.RootDir(), machineIDFile))
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -852,10 +852,10 @@ func clearSystemdState(installChroot *safechroot.Chroot, enableSystemdFirstboot 
 	}
 
 	if enableSystemdFirstboot {
-		ReportActionf("Enabling systemd firstboot")
+		ReportAction("Enabling systemd firstboot")
 		err = file.Write(machineIDFirstBootOn, filepath.Join(installChroot.RootDir(), machineIDFile))
 	} else {
-		ReportActionf("Disabling systemd firstboot")
+		ReportAction("Disabling systemd firstboot")
 		err = file.Write(machineIDFirstbootOff, filepath.Join(installChroot.RootDir(), machineIDFile))
 	}
 	if err != nil {
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -542,7 +542,7 @@ func PopulateInstallRoot(installChroot *safechroot.Chroot, packagesToInstall []s
 	}
 
 	// Configure machine-id and other systemd state files
-	err = clearSystemdState(installChroot)
+	err = clearSystemdState(installChroot, config.EnableSystemdFirstboot)
 	if err != nil {
 		err = fmt.Errorf("failed to clean systemd files:\n%w", err)
 		return
@@ -784,12 +784,15 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
 
 // clearSystemdState clears the systemd state files that should be unique to each instance of the image. This is
 // based on https://systemd.io/BUILDING_IMAGES/. Primarily, this function will ensure that /etc/machine-id is configured
-// correctly for first boot, and that random seed and credential files are removed if they exist.
-func clearSystemdState(installChroot *safechroot.Chroot) (err error) {
+// correctly, and that random seed and credential files are removed if they exist.
+// - installChroot is the chroot to modify
+// - enableSystemdFirstboot will set the machine-id file to "uninitialized" if true, and "" if false
+func clearSystemdState(installChroot *safechroot.Chroot, enableSystemdFirstboot bool) (err error) {
 	const (
-		machineIDFile           = "/etc/machine-id"
-		machineIDPresentContent = "uninitialized\n"
-		machineIDFilePerms      = 0444
+		machineIDFile         = "/etc/machine-id"
+		machineIDFirstBootOn  = "uninitialized\n"
+		machineIDFirstbootOff = ""
+		machineIDFilePerms    = 0444
 	)
 
 	// These state files are very unlikely to be present, but we should be thorough and check for them.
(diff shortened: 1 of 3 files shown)


dmcilvaney:
Some interesting data points:
- Ubuntu marketplace does not enable first boot
- https://github.com/ostreedev/ostree/issues/2840 Looks like preset behavior might change... https://fedoraproject.org/wiki/Changes/Preset_All_Systemd_Units_on_First_Boot But I think we use %systemd_post to set stuff so it should be ok.
- Old issue on systemd github about this: https://github.com/systemd/systemd/issues/8268

tldr, first boot seems to be the preferred path in modern systemd distros

dmcilvaney:
Per some discussions, going to keep old behavior.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -781,36 +782,109 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
...
+	//         writable).
+	//     5. If /etc/machine-id already contains a valid machine-id, this is not a first boot.
+	//     If according to the above rules a first boot is detected, units with ConditionFirstBoot=yes will be run and
+	//     systemd will perform additional initialization steps, in particular presetting units.
+	//
+	// However, the detailed guide for creating images (https://systemd.io/BUILDING_IMAGES/) suggests that the
+	// machine-id file should be "uninitialized\n" to trigger first-boot mode. This is because the machine-id file is
+	// used to determine if the image is being booted for the first time. If the file is empty, systemd will assume it
+	// is not the first boot and will not perform first-boot actions.
+	// (see https://www.freedesktop.org/software/systemd/man/latest/systemd-firstboot.html)
```
dmcilvaney, line 836:
```suggestion
	// We will use option 4) by default since AZL has traditionally not used firstboot mechanisms. All configuration
	// that systemd-firstboot would set should have already been configured by the imager tool. It is important to
	// create an empty file so that read-only configurations will work as expected. If the user requests that firstboot
	// be enabled we will set it to "uninitalized" as per option 3).
```

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -781,36 +782,108 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
...
+	//     5. If /etc/machine-id already contains a valid machine-id, this is not a first boot.
+	//     If according to the above rules a first boot is detected, units with ConditionFirstBoot=yes will be run and
+	//     systemd will perform additional initialization steps, in particular presetting units.
+	//
+	// We will use option 4) by default since AZL has traditionally not used firstboot mechanisms. All configuration
+	// that systemd-firstboot would set should have already been configured by the imager tool. It is important to
+	// create an empty file so that read-only configurations will work as expected. If the user requests that firstboot
+	// be enabled we will set it to "uninitalized" as per option 3).
+
+	ReportAction("Clearing systemd state files for first boot")
```
dmcilvaney, line 837:
```suggestion
	ReportAction("Configuring systemd state files for first boot")
```

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -781,36 +782,108 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
...
+
+	var machineIdContent string
+	if enableSystemdFirstboot {
+		ReportActionf("Enabling systemd firstboot")
+		machineIdContent = machineIDFirstBootOn
+	} else {
+		ReportActionf("Disabling systemd firstboot")
+		machineIdContent = machineIDFirstbootOff
+	}
+	err = file.Write(machineIdContent, filepath.Join(installChroot.RootDir(), machineIDFile))
```
dmcilvaney, line 862:
```suggestion
	if enableSystemdFirstboot {
		ReportActionf("Enabling systemd firstboot")
		err = file.Write(machineIDFirstBootOn, filepath.Join(installChroot.RootDir(), machineIDFile))
	} else {
		ReportActionf("Disabling systemd firstboot")
		err = file.Write(machineIDFirstbootOff, filepath.Join(installChroot.RootDir(), machineIDFile))
	}
```

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -781,36 +782,106 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
...
+	} else {
+		ReportActionf("Disabling systemd firstboot")
+		err = file.Write(machineIDFirstbootOff, filepath.Join(installChroot.RootDir(), machineIDFile))
+	}
+	if err != nil {
+		err = fmt.Errorf("failed to write empty machine-id:\n%w", err)
+		return err
+	}
+
+	// These files should not be present in the image, but we should be thorough and double-check.
```
dmcilvaney, line 866:
```suggestion
	// These files should not be present in the image, but per https://systemd.io/BUILDING_IMAGES/ we should
	// be thorough and double-check.
```

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -781,36 +782,107 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
...
-// addMachineID creates the /etc/machine-id file in the installChroot, if it is
-// not already present (systemd package will include it if installed).
-func addMachineID(installChroot *safechroot.Chroot) (err error) {
-	// From https://www.freedesktop.org/software/systemd/man/machine-id.html:
+// clearSystemdState clears the systemd state files that should be unique to each instance of the image. This is
+// based on https://systemd.io/BUILDING_IMAGES/. Primarily, this function will ensure that /etc/machine-id is configured
+// correctly, and that random seed and credential files are removed if they exist.
+// - installChroot is the chroot to modify
+// - enableSystemdFirstboot will set the machine-id file to "uninitialized" if true, and "" if false
+func clearSystemdState(installChroot *safechroot.Chroot, enableSystemdFirstboot bool) (err error) {
```
anphel31, line 790: is the capitalization of the first letter intentionally different from the systemconfig option? `enableSystemdFirstboot` vs `EnableSystemdFirstboot`

dmcilvaney: Go enforces capitalized names for public values in packages and structs, but generally variables should be lower case. We treat our config structures as publicly accessible, hence `config.EnableSystemdFirstboot`, but the parameter for the function is not a "public" element so `enableSystemdFirstboot`.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -781,36 +782,107 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
...
+	if err != nil {
+		err = fmt.Errorf("failed to write empty machine-id:\n%w", err)
+		return err
+	}
+
+	// These files should not be present in the image, but per https://systemd.io/BUILDING_IMAGES/ we should
+	// be thorough and double-check.
+	for _, filePath := range otherFilesToRemove {
+		fullPath := filepath.Join(installChroot.RootDir(), filePath)
+		exists, err = file.PathExists(fullPath)
```
cwize1, line 870: If you are using `file.RemoveFileIfExists`, why do you need this `file.PathExists` check?

cwize1 (approved):

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: I want to leave a breadcrumb in the logs that we removed that file. I actually noticed that on some images we DO remove one of the random seem files... not sure why it only shows up on some images. 

I originally did this pattern for testing, then decided it was useful to keep for post-mortem analysis of a build.

I'll add a comment to that affect explaining.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -781,36 +782,107 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
...
+		if err != nil {
+			err = fmt.Errorf("failed to check if systemd state file (%s) exists:\n%w", filePath, err)
+			return err
+		}
+
+		if exists {
```
dmcilvaney, line 876:
```suggestion
		// Do an explicit check for existence so we can log the file removal.
		if exists {
```

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -781,36 +782,108 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
...
+			return err
+		}
+	}
+
+	if enableSystemdFirstboot {
+		ReportActionf("Enabling systemd firstboot")
+		err = file.Write(machineIDFirstBootOn, filepath.Join(installChroot.RootDir(), machineIDFile))
+	} else {
+		ReportActionf("Disabling systemd firstboot")
+		err = file.Write(machineIDFirstbootOff, filepath.Join(installChroot.RootDir(), machineIDFile))
```
PawelWMS, line 859: Do we even need this line and `machineIDFirstbootOff` if it's an empty string?

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -781,36 +782,108 @@ func calculateTotalPackages(packages []string, installRoot string) (installedPac
...
+			err = fmt.Errorf("failed to create empty machine-id:\n%w", err)
+			return err
+		}
+	}
+
+	if enableSystemdFirstboot {
+		ReportActionf("Enabling systemd firstboot")
+		err = file.Write(machineIDFirstBootOn, filepath.Join(installChroot.RootDir(), machineIDFile))
+	} else {
+		ReportActionf("Disabling systemd firstboot")
```
PawelWMS, line 858: Tiny simplification.

```suggestion
		ReportAction("Enabling systemd firstboot")
		err = file.Write(machineIDFirstBootOn, filepath.Join(installChroot.RootDir(), machineIDFile))
	} else {
		ReportAction("Disabling systemd firstboot")
```

PawelWMS (approved):
A couple clean-up suggestions but looks good.

toolkit/tools/imagegen/installutils/installutils.go
dmcilvaney: Probably not, but I wanted to be explicit on the behavior relative to the systemd docs.

PR #9517, opened 2024-06-26
Changed 'installutils.go' to print out the entire tdnf debug log

Tonisal-byte:
###### Summary 

This one line change causes the build tools to output the entire tdnf log when installing packages in any build process. There were instances when developers did not understand what errors were causing a tdnf install failure, leading to further confusion due to the lack of logs. This change provides more context for the developer and helps in the debugging process.

###### Change Log  
- Added a single line to 'installutils.go' in the TdnfInstallWithProgress function (line 599).

NO

###### Test Methodology

- Used build tools and checked logs to see if the full output was logged.

### DIFF
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -641,6 +641,7 @@ func TdnfInstallWithProgress(packageName, installRoot string, currentPackagesIns
 		const tdnfInstallPrefix = "Installing/Updating: "
 
 		// Only process lines that match tdnfInstallPrefix
+		logger.Log.Debug(line)
 		if !strings.HasPrefix(line, tdnfInstallPrefix) {
 			return
 		}


neha170 (approved):
Change LGTM from tooling perspective

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -641,6 +641,7 @@ func TdnfInstallWithProgress(packageName, installRoot string, currentPackagesIns
 		const tdnfInstallPrefix = "Installing/Updating: "
 
 		// Only process lines that match tdnfInstallPrefix
+		logger.Log.Debug(line)
```
cwize1, line 644: If it is desirable to put all the stdout lines in Debug, then this code can be greatly simplified. Specifically, `LogLevel(logrus.TraceLevel, logrus.WarnLevel)` below can be changed to `LogLevel(logrus.DebugLevel, logrus.WarnLevel)`. And then most of the `onStdout` function can be deleted.

But that was intentionally not done because the tdnf's stdout can be very spammy.

cwize1 (changes requested):

PR #9564, opened 2024-07-02
Osamaesmail/patch cni

osamaesmailmsft:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


SPECS/cni/cni.spec
```diff
@@ -65,13 +66,13 @@ the container is deleted. Because of this focus, CNI has a wide
 range of support and the specification is simple to implement.
 
 %prep
-%setup -q
+%autosetup -N
 cp %{SOURCE2} build.sh
-
-%build
```
PawelWMS, line 71: Good call to move untarring into `%prep`.

SPECS/cni/cni.spec
```diff
@@ -65,13 +66,13 @@ the container is deleted. Because of this focus, CNI has a wide
...
 # create vendor folder from the vendor tarball and set vendor mode
 tar -xf %{SOURCE3} --no-same-owner
+%patch 0 -p1 -d "vendor/golang.org/x/text"
```
PawelWMS, line 73: Since we're already moving thing around here I think we can drop the `tar -xf %{SOURCE3} --no-same-owner` and `%patch` lines completely and modify `%autosetup` instead to use this:
```
%autosetup -a 3 -p1
```

The `-a 3` argument tells `autosetup` to extract `Source3` in addition to `Source0`.

PawelWMS, line 73: Instead of using `-d` I'd suggest modifying the paths in the patch. This way we can use `%autosetup` or `%autopatch` to make adding or removing patches easier.

dmcilvaney: Checkout https://github.com/microsoft/azurelinux/blob/04f7ddc43f8949db24791413eba21172d11b925e/SPECS/coredns/coredns.spec#L46

Looks like the auto patcher has adopted the pattern we used elsewhere (its mostly gone from 3.0, probably because we updated all the packages, but if you look for "# Apply vendor before patching" in 2.0 you can see quite a few).

dmcilvaney: It should be enough to just change the paths in the patch file, or you can do it "properly" and use `git format-patch`
 to update the patch file. For these vendor ones I just tend to manually edit the patch file.

osamaesmailmsft: So, to clarify, should I do something along the lines of 
`%autosetup -a 3 -p1` and modify the patch file, or should I do something like `%autosetup -N` and keep the `tar` / replace `%patch` with `%autopatch`?

PawelWMS: It sounds like the former.

jslobodzian (approved):

PR #9812, opened 2024-07-12
Image Customizer: Improve resolv.conf behavior.

cwize1:
Instead of just deleting the `resolv.conf` file at the end of customization, restore it to its original state that was seen in the base image.

In addition, if the `resolv.conf` didn't exist in the base image and the `systemd-resolved` service is enabled, then add the symlink that would normally be created during first-boot. This is particularly useful for verity images where the filesytem is readonly.

Also, fix the verity example config for Azure Linux 3.0 to have default network settings that enable a DHCP client on all "physical" NICs. This is needed since cloud-init doesn't work well in verity images.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Ran existing UTs.
- Added new UTs.


toolkit/tools/internal/file/file.go
```diff
@@ -196,6 +196,13 @@ func Write(data string, dst string) (err error) {
 	return
 }
 
+func WriteWithPerm(data string, dst string, perm os.FileMode) (err error) {
+	logger.Log.Debugf("Writing to (%s)", dst)
```
PawelWMS, line 200: Might be also worth to mention the used permission in the log.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
```diff
@@ -0,0 +1,142 @@
...
+
+type resolvConfInfo struct {
+	existingType resolvConfType
+	fileContents string
+	symlinkPath  string
+}
+
+const (
+	resolvConfPath        = "/etc/resolv.conf"
+	resolvSystemdStubPath = "../run/systemd/resolve/stub-resolv.conf"
```
PawelWMS, line 35: Is there a reason to not use an absolute path for `resolvSystemdStubPath`? I doubt that `resolvConfPath` will ever change but if not needed, I think it's unnecessary to bind `resolvSystemdStubPath` to `resolvConfPath`. Or at least I think these two paths are independent of each other as in: regardless of what `resolvConfPath` may be, `resolvSystemdStubPath` may change independently.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
```diff
@@ -0,0 +1,142 @@
...
+	return existing, nil
+}
+
+func restoreResolvConf(existing resolvConfInfo, imageChroot *safechroot.Chroot) error {
+	imageResolveConfPath := filepath.Join(imageChroot.RootDir(), resolvConfPath)
+
+	// Delete the overridden resolv.conf file.
+	err := os.RemoveAll(imageResolveConfPath)
+	if err != nil {
+		return fmt.Errorf("failed to delete overridden resolv.conf file: %w", err)
```
PawelWMS, line 91:
```suggestion
		return fmt.Errorf("failed to delete overridden resolv.conf file:\n%w", err)
```

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
```diff
@@ -0,0 +1,142 @@
...
+		err := imageChroot.UnsafeRun(func() error {
+			err := shell.ExecuteLive(true, "systemctl", "is-enabled", "systemd-resolved.service")
+
+			var exitError *exec.ExitError
+			if errors.As(err, &exitError) && exitError.ExitCode() == 1 {
+				resolvedEnabled = false
+				return nil
+			}
+			return err
+		})
```
PawelWMS, line 107: Considering that you already added a comment to explain what's happening in this part, this looks like a good candidate to extract a separate function like `isSystemdResolvedEnabled` or maybe an even more helpful one for the long-run like `systemd.IsServiceEnabled`.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to check if systemd-resolved service is enabled:\n%w", err)
+		}
+
+		if resolvedEnabled {
+			// The systemd-resolved.service is enabled.
+			// So, create the symlink for the resolv.conf file.
+			// While technically this symlink will be automatically created on first boot, doing it now is helpful for
```
PawelWMS, line 115: Nit-pick: this comment line is getting a bit long, so it may be harder for some devs to work with depending on their environment.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
```diff
@@ -0,0 +1,142 @@
...
+	err = file.Copy(resolvConfPath, imageResolveConfPath)
+	if err != nil {
+		return resolvConfInfo{}, fmt.Errorf("failed to override resolv.conf file with host's resolv.conf:\n%w", err)
+	}
+
+	return existing, nil
+}
+
+func restoreResolvConf(existing resolvConfInfo, imageChroot *safechroot.Chroot) error {
+	imageResolveConfPath := filepath.Join(imageChroot.RootDir(), resolvConfPath)
```
PawelWMS, line 86: For symmetry, might be worth adding:
```suggestion
	logger.Log.Infof("Restoring resolv.conf file.")
	
	imageResolveConfPath := filepath.Join(imageChroot.RootDir(), resolvConfPath)
```

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
```diff
@@ -0,0 +1,142 @@
...
+
+	case resolvConfTypeSymlink:
+		// Restore the resolv.conf symlink.
+		err := os.Symlink(existing.symlinkPath, imageResolveConfPath)
+		if err != nil {
+			return fmt.Errorf("failed to restore resolv.conf file:\n%w", err)
+		}
+
+	default:
+		panic(fmt.Sprintf("unexpected imagecustomizerlib.resolvConfType: %#v", existing.existingType))
```
PawelWMS, line 138: Why not this?
```suggestion
		logger.Log.Panicf("unexpected imagecustomizerlib.resolvConfType: %#v", existing.existingType)
```

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
```diff
@@ -0,0 +1,142 @@
...
+			// verity images where the root filesystem is readonly.
+			err := os.Symlink(resolvSystemdStubPath, imageResolveConfPath)
+			if err != nil {
+				return fmt.Errorf("failed to create resolv.conf symlink:\n%w", err)
+			}
+		}
+
+	case resolvConfTypeFile:
+		// Restore the resolv.conf file.
+		err := file.WriteWithPerm(existing.fileContents, imageResolveConfPath, 0o644)
```
PawelWMS, line 125: To be on the safe side, what do you think about also saving the file permissions of the original resolv.conf?

toolkit/tools/pkg/imagecustomizerlib/resolvconf_test.go
```diff
@@ -0,0 +1,170 @@
...
+		return
+	}
+
+	imageConnection, err = connectToCoreEfiImage(buildDir, outImageFilePath)
+	if !assert.NoError(t, err) {
+		return
+	}
+	defer imageConnection.Close()
+
+	// Ensure resolv.conf symlink was set to systemd-resolved.
```
PawelWMS, line 162: I'd suggest splitting these into separate tests with names indicating the precise scenario being tests. This will make it easier to debug faster which case failed. The code for each scenario will also be shorted, making it simpler to understand and debug.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to check if systemd-resolved service is enabled:\n%w", err)
+		}
+
+		if resolvedEnabled {
+			// The systemd-resolved.service is enabled.
+			// So, create the symlink for the resolv.conf file.
+			// While technically this symlink will be automatically created on first boot, doing it now is helpful for
+			// verity images where the root filesystem is readonly.
+			err := os.Symlink(resolvSystemdStubPath, imageResolveConfPath)
```
PawelWMS, line 117: More for my own info: does it matter if the linked file actually exists? As in: should we inform/warn the user in case it's missing or it's controlled by systemd and even if it's not there now, it will get created eventually?

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: This is the value that `systemd-resolved` uses when it sets up the symlink on first boot. I am just copying it.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: For normal images, it is fine if the `resolv.conf` file is missing because `systemd-resolved` will create it on first-boot. But for verity images, `systemd-resolved` cannot create the file since the root filesystem is readonly. And this prevents networking from working.

Another approach would be to leave it up the user to create the `resolv.conf` file in their MIC config for verity images. But this adds extra complexity to the minimal verity config.

In general, for MIC, I generally try to follow the rule of not doing something unless the user explicitly asks for it. And this `resolv.conf` change is technically a violation of that rule. But I reason that since MIC needs to mess with the `resolv.conf` file anyway, it doesn't hurt to do a bit extra.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: Good call.

toolkit/tools/internal/file/file.go
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: I usually line wrap at a 120. But regardless, fixed.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/resolvconf_test.go
cwize1: I think it ended up with more code overall. But fixed.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: This was auto-generated by VSCode and I forgot to change it. Fixed.

amritakohli (approved):

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
```diff
@@ -0,0 +1,143 @@
...
+			// While technically this symlink will be automatically created on first boot, doing
+			// it now is helpful for verity images where the root filesystem is readonly.
+			err := os.Symlink(resolvSystemdStubPath, imageResolveConfPath)
+			if err != nil {
+				return fmt.Errorf("failed to create resolv.conf symlink:\n%w", err)
+			}
+		}
+
+	case resolvConfTypeFile:
+		logger.Log.Debugf("Restoring resolv.conf file")
```
PawelWMS, line 121: Right now, for the `resolvConfTypeNone` scenario, we add an additional info-level log that we're adding the symlink as part of restoring the resolv.conf file. For the other two options we're only doing a debug-level log. I think it would be useful to stay consistent. Personally, I'm leaning towards making the debug-level logs and info and removing the info log at the very beginning of the function, but my main focus is on consistency.

toolkit/tools/pkg/imagecustomizerlib/resolvconf_test.go
PawelWMS: Yes, more code overall but less per test. If there's any common boilerplate code, which is common across all tests, I'm also fine with extracting it to a separate function.

PawelWMS (approved):

toolkit/tools/pkg/imagecustomizerlib/resolvconf.go
cwize1: Firstly, I think it is goodness to have an info log before anything happens. In this case, before the overridden `resolv.conf` file is removed.

Secondly, I think that restoring the previous file/symlink falls under banner of "Restoring resolv.conf". But creating a new symlink probably deserves a special callout. Hence the additional info log.

PR #9845, opened 2024-07-16
Fix Tensorflow Golden Container Smoke test

rikenm1:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Tensorflow Golden container smoke test is failing.
###### Change Log  

- Change: Update Keras to 3.3.3 and add a patch to enable wheel installation. Also use keras build to enable keras API for tensorflow
- Change: Remove D_GLIBCXX_ASSERTIONS from optree build 

NO

###### Test Methodology

- Pipeline build id: [605833](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=605833&view=results)


SPECS/keras/0001-Add-Keras-3.3.3.patch
```diff
@@ -0,0 +1,40 @@
...
++root_path = ""
+ to_copy = ["setup.py", "README.md"]
+ 
+ 
+@@ -86,7 +87,8 @@
+         move_tf_keras_directory()
+         print(os.getcwd())
+ 
+-        from keras.src.version import __version__  # noqa: E402
++        #from keras.src.version import __version__  # noqa: E402
```
rikenm1, line 17: was getting an error so just hardcoded this instead

jslobodzian: What was the error?  Can you comment in the SPEC file.  And if we absolutely need this, can you add something to the %build section that fails the build when somebody upgrades keras to anything other than 3.3.3?

SPECS/python-optree/python-optree.spec
```diff
@@ -30,18 +32,28 @@ A PyTree is a recursive structure that can be an arbitrarily nested Python conta
 
 
 %build
+
+export FFLAGS=$(echo $FFLAGS | sed 's/\(-Wp,-D_GLIBCXX_ASSERTIONS\)//g')
+export CXXFLAGS=$(echo $CXXFLAGS | sed 's/\(-Wp,-D_GLIBCXX_ASSERTIONS\)//g')
+export CFLAGS=$(echo $CFLAGS | sed 's/\(-Wp,-D_GLIBCXX_ASSERTIONS\)//g')
+export RPM_OPT_FLAGS=$(echo $RPM_OPT_FLAGS | sed 's/\(-Wp,-D_GLIBCXX_ASSERTIONS\)//g')
```
jslobodzian, line 39: Likewise for RPM_OPT_FLAGS and FCFLAGS.  I think it is safe to limit the change to just the CXXFLAGS or CFLAGS.  I believe it's using gcc not g++ to compile, so probably the CFLAGS are the only ones that actually matter.

jslobodzian (changes requested):
Can we narrow the change to just the impacted flags.

SPECS/python-optree/python-optree.spec
PawelWMS: +1 to doing only surgical changes to the flags we actually need to change. Otherwise, the specs (or any piece of code, really) gets noisy, it's hard to understand the true intention and which changes we actually need. In general, never include redundant/unnecessary lines. It looks like you're very close now.

SPECS/python-optree/python-optree.spec
```diff
@@ -30,18 +32,28 @@ A PyTree is a recursive structure that can be an arbitrarily nested Python conta
 
 
 %build
+
+export FFLAGS=$(echo $FFLAGS | sed 's/\(-Wp,-D_GLIBCXX_ASSERTIONS\)//g')
```
PawelWMS, line 36: Aren't we're missing the `%check` section? I understood we caught the issue when you were adding a simple tests, which tried to initialize a tree with an empty tuple.

SPECS/keras/0001-Add-Keras-3.3.3.patch
rikenm1: The error was it would complain about missing tensorflow module.  Tensorflow used to build time dep but with Keras 3.0 not anymore. Keras also doesn't use bazel anymore either.  
Currently, TensorFlow is not even runtime nor a build time dependency of keras. Keras is runtime req of tensorflow.  "pip3 install keras" doesn't install tensorflow.  Hence hardcoded this. If not it would have created TensorFlow as a build time req just to get the version number. 
I will add a variable.

SPECS/python-optree/python-optree.spec
rikenm1: We caught it from the smoke test for tensorflow container. when running %pytest, it complains about missing helper files.  Looks like test provided by python-optree doesn't work properly.

SPECS/keras/0001-Add-Keras-3.3.3.patch
rikenm1: here's the stack trace. 
image

SPECS/python-optree/python-optree.spec
rikenm1: updated

SPECS/python-optree/python-optree.spec
```diff
@@ -30,18 +32,24 @@ A PyTree is a recursive structure that can be an arbitrarily nested Python conta
 
 
 %build
+# Remove -D_GLIBCXX_ASSERTIONS because optree throws std lib assertion errors.
```
PawelWMS, line 35: Sounds a bit like we're burying a legitimate error, which is not really what's happening. How about this:
```suggestion
# Remove "-D_GLIBCXX_ASSERTIONS" because optree will otherwise crash
# on an C++ std libs assertion error. This happens when one tries to
# call 'map_tree' on an empty tuple (or possibly all empty collections).
# We currently suspect an upstream bug, as 'map_tree' seems to work
# fine when compiled without the "_GLIBCXX_ASSERTIONS" macro.
```

PawelWMS (approved):

jslobodzian (approved):

PR #9874, opened 2024-07-17
OSModifier: Add support for updating grub

elainezhao96:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Context: On 3.0, the grub.cfg file is generated from /etc/default/grub which is covered by the hash. Thus MIC manually edits /boot/grub2/grub.cfg after generating it to insert the root hash as an extra argument. However, Trident needs to re-generate the grub.cfg and thus needs to reapply the edits. We have been planning on using EMU to perform these edit operations

###### Change Log  

Overall EMU at this stage will be able to update some values required for trident
-  `./osmodifier --update-grub` enable EMU to inspect the grub.cfg and determine what are the **verity/overlay/selinux** settings, then update (if exists) or add (if not exists) these values to the /etc/default/grub, and then run grub2-mkconfig.
-  `./osmodifier --config-file=xxx.yaml` and yaml can have 
```
overlays:
- lowerDir: /etc
  upperDir: /etc/upper_root
  workDir: /etc/work_root
  partition:
    id: ....

selinux:
  mode: enforcing
```
 enable EMU to accept the selinux and overlayfs values from config file, then apply the changes to the /etc/default/grub, and then run grub2-mkconfig. There is a custom overlay to EMU API so that trident can pass their overlay values to EMU (the overlay struct is mostly similar to the old overlay struct)

note: kernel command line is not included in this pr, it needs to be discussed and will be in a future task

**NO**

###### Test Methodology

- Verified locally. Will test in trident locally and in trident  pipeline
- tlocal test description is in related pr: https://dev.azure.com/mariner-org/ECF/_git/trident/pullrequest/19796


toolkit/tools/osmodifier/main.go
```diff
@@ -47,6 +47,15 @@ func main() {
 }
 
 func modifyImage() error {
+	if *configFile == "" {
```
cwize1, line 50: It would probably be better to use [subcommands](https://github.com/alecthomas/kingpin?tab=readme-ov-file#sub-commands) than keying off if the `--config-file` arg is provided or not.

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -17,14 +17,14 @@ type BootCustomizer struct {
 	grubCfgContent string
 
 	// The contents of the /etc/default/grub file.
-	defaultGrubFileContent string
+	DefaultGrubFileContent string
```
cwize1, line 20: Making the field public is breaking the abstraction of this class.

toolkit/tools/pkg/osmodifierlib/modifierutils.go
```diff
@@ -23,3 +29,69 @@ func doModifications(baseConfigPath string, osConfig *imagecustomizerapi.OS) err
...
+		return fmt.Errorf("error getting verity changes:\n%v", err)
+	}
+
+	bootCustomizer, err := imagecustomizerlib.NewBootCustomizer(dummyChroot)
+	if err != nil {
+		return err
+	}
+
+	// Apply the verity changes to /etc/default/grub
+	err = bootCustomizer.ApplyChangesToGrub("GRUB_CMDLINE_LINUX", verityParams)
```
cwize1, line 48: I don't really understand what this change is doing. MIC doesn't write to the `/etc/default/grub` file since doing so would break the hash. So, why is it ok to write to the `/etc/default/grub` file here?

cwize1 (changes requested):

toolkit/tools/pkg/osmodifierlib/modifierutils.go
fintelia: Trident will be running osmodifier with an overlay mounted over /etc, which means that editing /etc/default/grub will impact only the copy in the overlay and not change the hash

toolkit/tools/osmodifier/main.go
elainezhao96: I've tried subcommands, it will somehow fail to read flags (--config-file), so I'll keep them both as flags

cwize1 (changes requested):

toolkit/tools/pkg/osmodifierlib/modifierutils.go
```diff
@@ -21,5 +26,163 @@ func doModifications(baseConfigPath string, osConfig *imagecustomizerapi.OS) err
...
+	// Function to extract the root value from kernelCommandLine.extraCommandLine
+	re := regexp.MustCompile(`root=([^\s]+)`)
+	matches := re.FindStringSubmatch(string(kernelExtraArguments))
+	if len(matches) < 2 {
+		return fmt.Errorf("no root device found in extraCommandLine")
+	}
+
+	rootDeviceValue := matches[1]
+
+	err := bootCustomizer.ApplyChangesToGrub("GRUB_DEVICE", rootDeviceValue)
```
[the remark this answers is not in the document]
elainezhao96: per trident request, modifyDefaultGrub reads the root device value from the existing grub.cfg to update GRUB_DEVICE. Here this update adjusts GRUB_DEVICE based on the root device value provided by osconfig, potentially overriding the values previously set by modifyDefaultGrub

toolkit/tools/pkg/osmodifierlib/modifierutils.go
```diff
@@ -21,5 +26,163 @@ func doModifications(baseConfigPath string, osConfig *imagecustomizerapi.OS) err
...
+		return fmt.Errorf("error getting changes:\n%v", err)
+	}
+
+	bootCustomizer, err := imagecustomizerlib.NewBootCustomizer(dummyChroot)
+	if err != nil {
+		return err
+	}
+
+	// Stamp verity and overlayfs to /etc/default/grub
+	err = bootCustomizer.ApplyChangesToGrub("GRUB_CMDLINE_LINUX", verityAndOverlayValues)
```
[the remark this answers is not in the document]
elainezhao96: ah you're right, it could be problematic to replace entire GRUB_CMDLINE_LINUX. I should only update the certain values like selinux/verity/overlay ... and leave the rest unchanged

toolkit/tools/pkg/osmodifierlib/modifierutils.go
```diff
@@ -21,5 +26,163 @@ func doModifications(baseConfigPath string, osConfig *imagecustomizerapi.OS) err
 		return err
 	}
 
+	if osConfig.Overlays != nil {
```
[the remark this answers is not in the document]
elainezhao96: got reply from lanze: "The new MIC Overlays API will not impact how Trident to enable the overlays. So, with dracut-overlays pkg and the required kernel cmd args, it still work to enable customized overlays."

toolkit/tools/pkg/osmodifierlib/modifierutils.go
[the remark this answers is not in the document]
elainezhao96: add UpdateCmdlineValues  function which will:
1. Adding New Keys: If newValues contains a key that isn't already in GRUB_CMDLINE_LINUX, the function will add that key.
2. Updating Existing Keys: If the key already exists, the function will update its value.
3. Handling Flags: If a key is present without an = (e.g., nosplash), it's treated as a boolean flag and added accordingly.

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -206,3 +214,53 @@ func (b *BootCustomizer) WriteToFile(imageChroot *safechroot.Chroot) error {
 
 	return nil
 }
+
+func (b *BootCustomizer) ApplyChangesToGrub(varName string, newValue string) error {
```
cwize1, line 218: You do not need the `ApplyChangesToGrub` and `UpdateCmdlineValues` functions. Please delete them. The `UpdateKernelCommandLineArgs` function already does what you need.

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -160,37 +160,45 @@ func (b *BootCustomizer) UpdateKernelCommandLineArgs(defaultGrubFileVarName defa
...
 
 	return nil
 }
 
-func (b *BootCustomizer) WriteToFile(imageChroot *safechroot.Chroot) error {
+func (b *BootCustomizer) WriteToFile(imageChroot safechroot.ChrootInterface) error {
 	if b.isGrubMkconfig {
 		// Update /etc/defaukt/grub file.
-		err := writeDefaultGrubFile(b.defaultGrubFileContent, imageChroot)
+		fmt.Println("heheheheheheheh", b.defaultGrubFileContent)
```
cwize1, line 197: Remove.

(If it would be useful, you could add a trace log with `defaultGrubFileContent` contents.)

toolkit/tools/pkg/osmodifierlib/modifierutils.go
```diff
@@ -21,5 +28,100 @@ func doModifications(baseConfigPath string, osConfig *imagecustomizerapi.OS) err
...
+		bootCustomizer, err := imagecustomizerlib.NewBootCustomizer(dummyChroot)
```
cwize1, line 32: The `bootCustomizer` variable here is hiding the `bootCustomizer` in the parent scope because you are using `:=` instead of `=`. So the `if bootCustomizer == nil` check below doesn't actually do anything since the parent's scope `bootCustomizer` variable will always be `nil` at that point.

toolkit/tools/pkg/imagecustomizerlib/defaultgrubutils.go
```diff
@@ -313,14 +313,15 @@ func updateDefaultGrubFileVariable(defaultGrubFileContent string, varName string
 		}
 
 		defaultGrubFileContent = insertDefaultGrubFileVarAssign(defaultGrubFileContent, insertAfter, varName, newValue)
+
```
cwize1, line 316: Remove added newline

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -160,37 +160,45 @@ func (b *BootCustomizer) UpdateKernelCommandLineArgs(defaultGrubFileVarName defa
...
-		defaultGrubFileContent, err = updateDefaultGrubFileVariable(defaultGrubFileContent, "GRUB_DISABLE_RECOVERY",
+		defaultGrubFileContent, err = UpdateDefaultGrubFileVariable(defaultGrubFileContent, "GRUB_DISABLE_RECOVERY",
 			"true")
 		if err != nil {
 			return err
 		}
 
+		// Check if rootDeviceValue is available as a field in BootCustomizer or another way
+		if rootDeviceValue != "" {
+			defaultGrubFileContent, err = UpdateDefaultGrubFileVariable(defaultGrubFileContent, "GRUB_DEVICE", rootDeviceValue)
```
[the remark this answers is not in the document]
elainezhao96: when I test EMU in trident, "/dev/mapper/root" is somehow not set in original /etc/default/grub, so I guess still need to read from grub.cfg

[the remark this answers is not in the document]
cwize1: If you are using an image generated by the existing MIC version, then `GRUB_DEVICE` won't be set in `/etc/default/grub` because your change hasn't gone in yet.

Please remove the `rootDeviceValue` parameter and always set `GRUB_DEVICE` to `"/dev/mapper/root"`.

If you want to continue calling `PrepareForVerity()` in EMU/Trident so that the code works with older images, then that is fine.

cwize1 (changes requested):

cwize1 (approved):

PR #9900, opened 2024-07-22
Image Customizer: Validate HOME and USER env vars.

cwize1:
The gpg command, and by extension, the tdnf command expects the USER and HOME environment variables to be valid for the OS they are running under (including under chroot). Since, the image customizer tool is typically run under `sudo` and since root is always a valid user, this generally isn't a problem. But this isn't true if `sudo -E` is used. And this can result in strange difficult to diganose errors in tdnf. So, this change verifies that `sudo -E` isn't being used.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Manually tested image customizer tool.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
@@ -230,6 +230,11 @@ func CustomizeImage(buildDir string, baseConfigPath string, config *imagecustomi
 		}
 	}()
 
+	err = checkEnvironmentVars()
+	if err != nil {
+		return err
+	}
+
 	// ensure build and output folders are created up front
 	err = os.MkdirAll(imageCustomizerParameters.buildDirAbs, os.ModePerm)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -897,3 +902,29 @@ func humanReadableUnitSizeAndName(size int64) (int64, string) {
...
+		rootHome = "/root"
+		rootUser = "root"
+	)
+
+	envHome := os.Getenv("HOME")
+	envUser := os.Getenv("USER")
+
+	if envHome != rootHome || envUser != rootUser {
+		return fmt.Errorf("tool should be run as root (e.g. by using sudo):\n"+
+			"HOME should be set to '%s' and USER should be set to '%s'",
```
gmileka, line 925: nit: the 'should' here sounds (a little bit) like it's suggesting the user to set these variables, but it's not, correct?

gmileka (approved):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: Fixed.

PR #9932, opened 2024-07-25
Toolkit: Do not give GPT partitions a default label of "primary".

cwize1:
When creating partitions using `parted`, the API is:

- MBR: `parted mkpart <partition-type> <fs-type> ...`
- GPT: `parted mkpart <partition-label> <fs-type> ...`

But currently we are treating the first param as always being the `<partition-type>`, when for GPT it is `<partition-label>`. This has the effect that GPT partitions are given a default label of "primary", which is a tad silly. (If an explicit label is provided by the user, then the label is overridden in a subsequent step.)

This change fixes this behavior so that GPT partitions have a default label of nothing ("").

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Added UT.

### DIFF
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -464,7 +464,7 @@ func CreatePartitions(diskDevPath string, disk configuration.Disk, rootEncryptio
 		partType, partitionNumber := obtainPartitionDetail(idx, usingExtendedPartition)
 		// Insert an extended partition
 		if partType == extendedPartitionType {
-			err = createExtendedPartition(diskDevPath, partitionTableType.String(), disk.Partitions, partIDToFsTypeMap, partDevPathMap)
+			err = createExtendedPartition(diskDevPath, partitionTableType, disk.Partitions, partIDToFsTypeMap, partDevPathMap)
 			if err != nil {
 				return
 			}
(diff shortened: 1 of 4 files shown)


toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -541,21 +543,37 @@ func CreateSinglePartition(diskDevPath string, partitionNumber int, partitionTab
...
+		// Part type.
+		mkpartArgs = append(mkpartArgs, partType)
+
+	case configuration.PartitionTableTypeGpt:
+		// Partition label.
+		if partition.Name == "" {
+			// Since parted is a scripting language, you have to specify "" to represent an empty string.
+			mkpartArgs = append(mkpartArgs, "\"\"")
+		} else {
+			mkpartArgs = append(mkpartArgs, partition.Name)
```
dmcilvaney, line 559: Not 100% sure this works, but reads a bit better to me.
Also replaced "\\"\\"" -> \`""\`
```suggestion
			mkpartArgs = append(mkpartArgs, []string{"--name", `""`}...)
		} else {
			mkpartArgs = append(mkpartArgs, []string{"--name",partition.Name}...)
```

dmcilvaney (approved):

toolkit/tools/imagegen/diskutils/diskutils.go
cwize1: Fixed.

PR #9963, opened 2024-07-29
Explicit toolchain signature validation

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

**NOTE:** Built on top of https://github.com/microsoft/azurelinux/pull/9941

Allow automatic validation of toolchain package signatures during builds without relying on sideband mechanisms. `tdnf`, `dnf`, etc. offer additional validation when interacting with repos (`*.repo` offers `gpgcheck=1` and `repogpgcheck=1`), however by default `rpm` does not offer this functionality, especially when we are using it to create our worker environment out of downloaded packages (we use `rpm --root ...` to build it up).

We also do not want to use the above repo level checks during package caching since the tools will often be working with locally built, unsigned, packages. 

To do validation of toolchain currently either the toolchain must be built in a trusted environment, or each individual package must be validated. We can automate this later process since we have the public part of the GPG keys available in our repo and can use them as we download each toolchain package.

The automated validation is enabled by default when downloading toolchains from the upstream repos (base, preview), but is disabled when building locally, or using the unsigned daily test bits. Switching from one to the other will automatically invalidate the toolchain download directory.

Validation relies on `rpmkeys --checksig` with an explicit `-D "%_pkgverify_level signature"` to force full validation [see rpm macros here](https://github.com/rpm-software-management/rpm/blob/ef87d2503498f65577b5d7af07cd453d622fe02c/macros.in#L667). We can use an empty directory using `--root`, install exactly the keys we care about, then validate the package.

###### Change Log  

- Add `VALIDATE_TOOLCHAIN_GPG` flag to the builds

**YES**

###### Test Methodology

```
# Fails validation (wrong key explicitly set)
 sudo make toolchain REBUILD_TOOLS=y TOOLCHAIN_GPG_VALIDATION_KEYS=../SPECS/azurelinux-repos/MICROSOFT-METADATA-GPG-KEY -j1
# Now works when we change the key
 sudo make toolchain REBUILD_TOOLS=y TOOLCHAIN_GPG_VALIDATION_KEYS=../SPECS/azurelinux-repos/MICROSOFT-RPM-GPG-KEY -j1
```
Also validated that a packaged toolkit works.

- Full: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=614191&view=results
- Buddy: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=614195&view=results


cwize1 (approved):

toolkit/scripts/toolchain/download_toolchain_rpm.sh
```diff
@@ -19,6 +19,8 @@ function usage() {
     echo "           until the RPM is successfully downloaded. Full URL will be <url>/<rpm_name>"
     echo "-c|--certificate: Optional path to a certificate file to use for the download"
     echo "-k|--private-key: Optional path to a private key file to use for the download"
+    echo "-K|--enforce-signatures: Optional flag to enforce RPM signatures"
```
neha170, line 22: I would choose a different letter to avoid confusion since 'k' is already used

dmcilvaney: I was basing on `rpm -K`, but I'll think of a better one.

toolkit/Makefile
```diff
@@ -143,6 +143,20 @@ PACKAGE_URL_LIST   += https://packages.microsoft.com/azurelinux/$(RELEASE_MAJOR_
...
+VALIDATE_TOOLCHAIN_GPG ?= n
+else
+ifneq ($(DAILY_BUILD_ID),)
+VALIDATE_TOOLCHAIN_GPG ?= n
+else
+VALIDATE_TOOLCHAIN_GPG ?= y
+endif
+endif
+
+TOOLCHAIN_GPG_VALIDATION_KEYS ?= $(wildcard $(SPECS_DIR)/azurelinux-repos/MICROSOFT-*-GPG-KEY)$(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)
```
jslobodzian, line 158: space?

jslobodzian (approved):

toolkit/Makefile
dmcilvaney:
```suggestion
TOOLCHAIN_GPG_VALIDATION_KEYS ?= $(wildcard $(SPECS_DIR)/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)
```

toolkit/scripts/toolchain.mk
```diff
@@ -292,7 +292,7 @@ $(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(STATUS_FLAGS_DIR)/toolchain_local_tem
 
 # No archive was selected, so download from online package server instead. All packages must be available for this step to succeed.
 else
-$(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(STATUS_FLAGS_DIR)/daily_build_auto_cleanup.flag $(depend_REBUILD_TOOLCHAIN) $(go-downloader) $(SCRIPTS_DIR)/toolchain/download_toolchain_rpm.sh
+$(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(STATUS_FLAGS_DIR)/toolchain_auto_cleanup.flag $(depend_REBUILD_TOOLCHAIN) $(go-downloader) $(SCRIPTS_DIR)/toolchain/download_toolchain_rpm.sh
```
PawelWMS, line 295: Shouldn't we also depend on `$(TOOLCHAIN_GPG_VALIDATION_KEYS)` for this target?
```suggestion
$(toolchain_rpms): $(TOOLCHAIN_MANIFEST) $(STATUS_FLAGS_DIR)/toolchain_auto_cleanup.flag $(depend_REBUILD_TOOLCHAIN) $(go-downloader) $(SCRIPTS_DIR)/toolchain/download_toolchain_rpm.sh $(TOOLCHAIN_GPG_VALIDATION_KEYS)
```

toolkit/scripts/toolchain/download_toolchain_rpm.sh
```diff
@@ -104,6 +116,13 @@ if [ -n "$key" ]; then
     key="--private-key=$key"
 fi
 
+if $enforce_signatures; then
+    if [ -z "$allowable_gpg_keys" ]; then
```
PawelWMS, line 120: I'd even test that each of the key paths is an existing file with `[ -f ... ]` to catch issues earlier and print a more helpful error message.

toolkit/scripts/toolchain/download_toolchain_rpm.sh
```diff
@@ -19,6 +19,8 @@ function usage() {
     echo "           until the RPM is successfully downloaded. Full URL will be <url>/<rpm_name>"
     echo "-c|--certificate: Optional path to a certificate file to use for the download"
     echo "-k|--private-key: Optional path to a private key file to use for the download"
+    echo "-e|--enforce-signatures: Optional flag to enforce RPM signatures"
+    echo "-g|--allowable-gpg-keys: Optional space separated list of GPG keys to allow for signature validation"
     exit 1
```
PawelWMS, line 24: Nit-pick and not even introduced in this change: I don't think `usage` should always `exit 1`. For instance, running the tool with `-h` shouldn't exit like an error.

toolkit/scripts/toolchain/download_toolchain_rpm.sh
```diff
@@ -139,6 +158,32 @@ function download() {
...
 }
 
+function validate_signatures() {
+    if [ -z "$work_dir" ]; then
+        work_dir=$(mktemp -d)
+        function cleanup() {
+            rm -rf "$work_dir"
+        }
+        trap cleanup EXIT
+    fi
```
PawelWMS, line 168: I understand this is the only place we're using a work directory right now, but I'd still suggest moving the set-up and clean-up to the global prep before we run any other logic. Bash scripts allow you to set-up only one function call for a trap, so it's good to expose all traps in a visible spot, so future changes can append to it. Otherwise, it's easy to accidentally try to set-up a separate trap elsewhere for the same signal and run into problems.

toolkit/scripts/toolchain/download_toolchain_rpm.sh
```diff
@@ -139,6 +158,32 @@ function download() {
...
+    echo "Validating toolchain RPM: $rpm_name" | tee -a "$log_file"
+
+    for key in $allowable_gpg_keys; do
+        echo "Adding key ($key) to empty workdir at ($work_dir)" >> "$log_file"
+        rpmkeys --root "$work_dir" --import "$key" >> "$log_file"
+    done
+
+    if ! rpmkeys --root "$work_dir" --checksig --verbose "$dst_file" -D "%_pkgverify_level signature" >> "$log_file"; then
+        echo "Failed to validate toolchain package $rpm_name signature, aborting." | tee -a "$log_file"
+        exit 1
```
PawelWMS, line 179: I'd suggest for functions to use `return` as much as possible. This way you can leave it up to the caller to decide, if they want to exit immediately, ignore the error, or do some more processing before they are done. Combining that with `set -e`, still allows us to add no additional code, if we want to immediately exit after `return <no-zero code>`.

toolkit/scripts/toolchain.mk
dmcilvaney: I was assuming they would be static, and the `$(wildcard...)` will only report "real" files, but sure, won't hurt.

toolkit/scripts/toolchain/download_toolchain_rpm.sh
dmcilvaney: Sure, I can fix it. This script was added as a pre-cursor to this change.

toolkit/scripts/toolchain/download_toolchain_rpm.sh
```diff
@@ -129,16 +168,39 @@ function download() {
...
+    fi
+    return 0
 }
 
 mkdir -p "$(dirname "$log_file")"
 
-download
+if ! download; then
+    exit 1
+fi
```
PawelWMS, line 200: Unnecessary with `set -e` at the top:
```suggestion
download
```

toolkit/scripts/toolchain/download_toolchain_rpm.sh
```diff
@@ -129,16 +168,39 @@ function download() {
...
+
+if $enforce_signatures; then
+    if ! validate_signatures; then
+        exit 1
+    fi
+fi
```
PawelWMS, line 206: Same here:
```suggestion
if $enforce_signatures; then
    validate_signatures
fi
fi
```

PawelWMS (approved):

PR #10016, opened 2024-08-04
fix: upgrade walinuxagent to 2.11.1.4 and add azurelinux patch

christopherco:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Upgrade WALinuxAgent to the most recent release - 2.11.1.4.

Add patch to include azurelinux support into the agent. The patch should
be removed in the next 2.12 update.

Also because WALinuxAgent will automatically update at runtime, thus
removing the patched-in support for azurelinux, add another patch to
bump the version to 2.11.8.8 so further 2.11 hotfixes are not taken.
Since 2.12 will have azurelinux support, the agent will be able to
automatically upgrade to 2.12 at runtime.

fix: correct changelog date warning

fix: use /usr/bin and /usr/lib/systemd paths 

In the upstream code's setup.py, the bin path and systemd path are set
through calls into the distro's osutil class. In our case, since we pass
--lnx-distro='mariner', the code will use the MarinerOsUtil derived
class for these paths, which are /usr/bin and /usr/lib/systemd
respectively.

In addition, now that the paths are set correctly again, remove the sed
operation to the waagent.service so the waagent path points to the
correct location again.

Also fix the ephemeral-disk-warning.service to use the proper path as
well.

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/52061747

###### Test Methodology

- Pipeline build id: [616327](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=616327&view=results)

### DIFF
--- a/SPECS/WALinuxAgent/0002-fix-bump-version-to-2.11.8.8.patch
+++ b/SPECS/WALinuxAgent/0002-fix-bump-version-to-2.11.8.8.patch
@@ -1,0 +1,30 @@
+From e2f24e98cd7f495d5cf6ef436625d9dad22785eb Mon Sep 17 00:00:00 2001
+From: Chris Co <chrco@microsoft.com>
+Date: Sat, 3 Aug 2024 22:25:04 -0700
+Subject: [PATCH] fix: bump version to 2.11.8.8
+
+In order to prevent further 2.11 agent upgrades, bump the version to
+2.11.8.8 so any upcoming 2.11 hotfixes are not applied. 2.12 will be
+released soon and the agent will take the 2.12 upgrade correctly.
+
+Signed-off-by: Chris Co <chrco@microsoft.com>
+---
+ azurelinuxagent/common/version.py | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/azurelinuxagent/common/version.py b/azurelinuxagent/common/version.py
+index c1715ff7..b4e93e81 100644
+--- a/azurelinuxagent/common/version.py
++++ b/azurelinuxagent/common/version.py
+@@ -209,7 +209,7 @@ AGENT_LONG_NAME = "Azure Linux Agent"
+ #
+ #            When doing a release, be sure to use the actual agent version. Current agent version: 2.4.0.0
+ #
+-AGENT_VERSION = '2.11.1.4'
++AGENT_VERSION = '2.11.8.8'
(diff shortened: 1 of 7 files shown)


SPECS/WALinuxAgent/0002-fix-bump-version-to-2.11.8.8.patch
```diff
@@ -0,0 +1,30 @@
...
+diff --git a/azurelinuxagent/common/version.py b/azurelinuxagent/common/version.py
+index c1715ff7..b4e93e81 100644
+--- a/azurelinuxagent/common/version.py
++++ b/azurelinuxagent/common/version.py
+@@ -209,7 +209,7 @@ AGENT_LONG_NAME = "Azure Linux Agent"
+ #
+ #            When doing a release, be sure to use the actual agent version. Current agent version: 2.4.0.0
+ #
+-AGENT_VERSION = '2.11.1.4'
++AGENT_VERSION = '2.11.8.8'
```
henryli001, line 24: Is `2.11.8` a value that we read from a dev branch or future WALinuxAgent version release plan? Based on my understanding of reading the comment, this version should represent the last minor version of v2.11 to prevent updating to a higher v2.11 version at runtime. Please correct me if misunderstood here.

henryli001:
May not necessarily be an issue for this change, but still would like to call it out here just in case.

Previously, we identified a [bug](https://github.com/Azure/WALinuxAgent/issues/3025) in v2.9.1.1 that affects both Ubuntu and Azure Linux where it fails to assign IP address for IB interface. A [fix](https://github.com/Azure/WALinuxAgent/pull/3073) has been made for this issue but seems like the change is targeting the `develop` branch instead of the `master` branch. Wondering whether this is something we should bring up to the upstream team to see whether to merge this change into the master branch as well since we're doing this version upgrade.

SPECS/WALinuxAgent/0002-fix-bump-version-to-2.11.8.8.patch
christopherco: Yes, your understanding is correct. 2.11.8 will be higher than any agent version hotfix for 2.11. And the WALinuxAgent will be updating to 2.12 in their next release which is upcoming

henryli001: Thanks for the clarification!

henryli001 (approved):

PR #10020, opened 2024-08-05
Ignore bogus case-insensitive provides results from repocloner

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

`tdnf provides <pkgname>` is used to find non-obvious requirements for the build (ie mapping `cap(thing)` to `thing-pkg`). `tdnf` will switch to case-insensitive matching if it can't find a result (`perl(mypkg)` will match with `perl(MyPkg)`), but this isn't actually a valid result. We convert these results into an actual `*.rpm` file path and expect future steps to be able to install this package to satisfy the requirement. During package build, we may never need to install a package so this (i.e. its a runtime requirement of a node which is never itself used in another build). Only at image generation time (or ad-hoc install by a user) will the lack of valid dependency be detected.

`tdnf` prints a message `[ignoring case for 'pkgname']` when it enters this mode, so we can watch for it and ignore those results. This may not be an actual error, future stages of the fetcher may find a match, or a local backage build may provide it, but if we skip adding the bogus result we will get a descriptive error further down the line (either as part of the image fetching process, or as part of the package build process).

###### Change Log  

- have `RpmRepoCloner.WhatProvides()` ignore any results that require case-insensitive matching
- Fix invalid case in `perl-Net-SSLeay` (`perl(Autoloader)` -> `perl(AutoLoader)`)

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/52845569

###### Test Methodology

- Add `Requires: symcrypt` to `words.spec`, add `"Packages": ["words"]` to `core-efi.json`. Run `sudo  make image CONFIG_FILE=./imageconfigs/core-efi.json SRPM_PACK_LIST=words DAILY_BUILD_ID=lkg`
- Check the latest 3.0-dev builds and search for "ignoring case for..." in the fetcher logs (found `perl-Net-SSLeay`).

### DIFF
--- a/SPECS/perl-Net-SSLeay/perl-Net-SSLeay.spec
+++ b/SPECS/perl-Net-SSLeay/perl-Net-SSLeay.spec
@@ -1,7 +1,7 @@
 Summary:        Perl extension for using OpenSSL
 Name:           perl-Net-SSLeay
 Version:        1.92
-Release:        4%{?dist}
+Release:        5%{?dist}
 License:        Artistic 2.0
 Group:          Development/Libraries
 URL:            https://metacpan.org/pod/distribution/Net-SSLeay/lib/Net/SSLeay.pod
@@ -19,7 +19,7 @@ BuildRequires:  perl(English)
 BuildRequires:  perl(ExtUtils::MakeMaker)
 BuildRequires:  perl-generators
 %if 0%{?with_check}
-BuildRequires:  perl(Autoloader)
+BuildRequires:  perl(AutoLoader)
 BuildRequires:  perl(CPAN)
 BuildRequires:  perl(CPAN::Meta)
 BuildRequires:  perl(CPAN::Meta::Requirements)
@@ -81,6 +81,9 @@ make test
 %{_mandir}/man?/*
 
 %changelog
+* Mon Aug 05 2024 Daniel McIlvaney <damcilva@microsoft.com> - 1.92-5
+- Fix bad capitalization of perl(AutoLoader)
+
 * Thu May 30 2024 Andrew Phelps <anphel@microsoft.com> - 1.92-4
 - Add BR on openssl
 
--- a/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
+++ b/toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
@@ -468,11 +468,21 @@ func (r *RpmRepoCloner) WhatProvides(pkgVer *pkgjson.PackageVer) (packageNames [
 				return
 			}
 
+			// Flag the result if we did case-insensitive lookup.
+			lookupIgnoredCase := tdnf.DidCaseInsensitiveMatchRegex.MatchString(stdout)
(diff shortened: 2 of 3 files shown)


toolkit/tools/internal/packagerepo/repocloner/rpmrepocloner/rpmrepocloner.go
```diff
@@ -468,11 +468,21 @@ func (r *RpmRepoCloner) WhatProvides(pkgVer *pkgjson.PackageVer) (packageNames [
...
+			// Flag the result if we did case-insensitive lookup.
+			lookupIgnoredCase := tdnf.DidCaseInsensitiveMatchRegex.MatchString(stdout)
+
 			// MUST keep order of packages printed by TDNF.
 			// TDNF will print the packages starting from the highest version, which allows us to work around an RPM bug:
 			// https://github.com/rpm-software-management/rpm/issues/2359
 			for _, matches := range tdnf.PackageLookupNameMatchRegex.FindAllStringSubmatch(stdout, -1) {
 				packageName := matches[tdnf.PackageNameIndex]
+				if lookupIgnoredCase {
+					logger.Log.Warnf("'%s' was found by case-insensitive lookup of '%s', but this is not a valid (Build)Requires and will be ignored", packageName, pkgVer.Name)
```
PawelWMS, line 480: I would not mention (Build)Requires in a log message. This loosely ties this tool to RPM specs' terminology, while the tool itself can be used in more generic cases to find package providing a capability for other scenarios. Suggestion:
```suggestion
					logger.Log.Warnf("'%s' was found by case-insensitive lookup of '%s', but this is not valid and will be ignored", packageName, pkgVer.Name)
```

SPECS/perl-Net-SSLeay/perl-Net-SSLeay.spec
```diff
@@ -19,7 +19,7 @@ BuildRequires:  perl(English)
 BuildRequires:  perl(ExtUtils::MakeMaker)
 BuildRequires:  perl-generators
 %if 0%{?with_check}
-BuildRequires:  perl(Autoloader)
+BuildRequires:  perl(AutoLoader)
```
PawelWMS, line 22: @dmcilvaney, do you recall what was exactly the failure, which helped us find this TDNF behaviour? I did a quick check running `tdnf install perl(Autoloader)` and `tdnf provides perl(Autoloader)` and both worked fine. DNF wouldn't work, so I think we still should be strict about casing, but I want to understand the failure better in case there's something more lingering there.

dmcilvaney: Was this a M2 or AZL3 version of `tdnf`?

PawelWMS: Good question. As far as I could tell, the behaviour is the same on both. Although 3.0 TDNF is faster.:)

jslobodzian (approved):

Camelron (approved):

PR #10025, opened 2024-08-05
Add priorities to local repos

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

`tdnf` has been updated to support priorities. As part of this it appears to have changed its behavior when selecting packages. To try and get our old behavior back lets use priorities in our local repo definitions. 

Lower number is higher priority, so set the toolchain lowest, followed by local repo, followed by cache.

###### Change Log  

- Add explicit priorities to `local.repo` for package builds.

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/53064337

###### Test Methodology

- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=616897&view=results

### DIFF
--- a/toolkit/resources/manifests/package/local.repo
+++ b/toolkit/resources/manifests/package/local.repo
@@ -5,6 +5,7 @@ enabled=1
 gpgcheck=0
 skip_if_unavailable=1
 sslverify=0
+priority=1
 
 [local-repo]
 name=Local Build Repo (out/RPMS)
@@ -13,6 +14,7 @@ enabled=1
 gpgcheck=0
 skip_if_unavailable=1
 sslverify=0
+priority=2
 
 [upstream-cache-repo]
 name=Cache Repo for upstream RPMs (build/rpm_cache/cache)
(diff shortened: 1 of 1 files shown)


toolkit/resources/manifests/package/local.repo
```diff
@@ -5,6 +5,7 @@ enabled=1
 gpgcheck=0
 skip_if_unavailable=1
 sslverify=0
+priority=1
```
PawelWMS, line 8: While we're here, maybe we can also add a comment at the top explaining the reasoning behind this order of repositories, so that we remember what we were doing later on. Or for the future devs.

When I think about it now, I believe our design was:
- Start with Azure Linux's toolchain packages on purpose, even if they would overwrite the local ones, to guarantee a stable, known build environment.
- For everything else, prefer the local packages first in case someone wants to build with their version of a package we already have in PMC.
- PMC for the remaining stuff.

anphel31 (approved):

PR #10039, opened 2024-08-06
dracut: add fix for initrd not showing prompt when root device is locked

trungams:
A change to dracut in v102 (https://github.com/dracut-ng/dracut-ng/commit/2339acfaeee60d6bb26a1103db2e53bc8f9cb2d1) made root decryption fail when booting an image with root partition encrypted. Normally, user is prompted for a password to unlock the encrypted partition during boot. With the above change in v102, no prompt is shown during initrd and eventually the boot fails. The commit is already reverted upstream and is supposed to appear in dracut v104. Adding this patch to revert the change in the meantime.

For CVM images, auto unlock of root partition with TPM enrollment also does not work, likely due to this bug.

- Upstream bug: https://github.com/dracut-ng/dracut-ng/issues/492
- https://microsoft.visualstudio.com/OS/_workitems/edit/51090393


Camelron (approved):
lgtm!

christopherco (approved):

PR #10075, opened 2024-08-09
fix: produce debuginfo in symcrypt package

christopherco:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

 Produce debuginfo package for SymCrypt

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/53101074

###### Test Methodology

- Pipeline build id: [625321](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=625321&view=results)


mfrw (approved):

SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -1,8 +1,7 @@
-%define debug_package %{nil}
```
samuel-lee-msft, line 1: I'm a little concerned that if you have not rebuilt SymCrypt-OpenSSL or OpenSSL packages which depend on this, you won't have tested that the produced libsymcrypt.so works with only the current %check in this spec.

When building modules in ADO we test the produced dynamic module with symcryptunittest, rather than just testing that symcryptunittest (with statically linked SymCrypt) is working as expected: https://github.com/microsoft/SymCrypt/blob/171f6973dab9b76f0dc61d966d3e977021325bc1/.pipelines/templates/build-linux.yml#L128-L135

This is independent of needing to do manual validation that the new debug symbols work as expected

SPECS/SymCrypt/SymCrypt.spec
paulmon, line 1: Am I reading this right?
This looks like the commit hash for 103.4.1, and the packages version is 103.4.2
`SYMCRYPT_COMMIT_HASH=a84ffe1`

SPECS/SymCrypt/SymCrypt.spec
christopherco: Does this mean anytime we update one of SymCrypt-OpenSSL, OpenSSL, or SymCrypt, we actually need to rebuild and redistribute all 3 in order for debug symbols to work?

SPECS/openssl/openssl.spec
```diff
@@ -9,7 +9,7 @@
 Summary: Utilities from the general purpose cryptography library with TLS implementation
 Name: openssl
 Version: 3.3.0
-Release: 2%{?dist}
+Release: 3%{?dist}
```
PawelWMS, line 12: I'm not sure I understand this part. OpenSSL doesn't build with SymCrypt as far as I could tell, so the bump shouldn't make a difference.

SPECS/SymCrypt-OpenSSL/SymCrypt-OpenSSL.spec
```diff
@@ -1,7 +1,7 @@
 Summary:        The SymCrypt engine for OpenSSL (SCOSSL) allows the use of OpenSSL with SymCrypt as the provider for core cryptographic operations
 Name:           SymCrypt-OpenSSL
 Version:        1.5.0
-Release:        1%{?dist}
+Release:        2%{?dist}
```
PawelWMS, line 4: Harmless, but is the bump necessary? We're only adding a debuginfo package, so nothing really changed for SymCrypt-OpenSSL.

SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -78,7 +77,10 @@ chmod 755 %{buildroot}%{_libdir}/libsymcrypt.so.%{version}
 %{_includedir}/*
 
 %changelog
-* Wed Jun 25 2024 Maxwell Moyer-McKee <mamckee@microsoft.com> - 103.4.2-1
+* Fri Aug 09 2024 Chris Co <chrco@microsoft.com> - 103.4.2-2
+- Produce debuginfo
+
+* Thu Jun 25 2024 Maxwell Moyer-McKee <mamckee@microsoft.com> - 103.4.2-1
```
PawelWMS, line 83: Thanks for the date fix!:)

SPECS/openssl/openssl.spec
christopherco: This bump is part of this comment - https://github.com/microsoft/azurelinux/pull/10075#discussion_r1712170768

SPECS/SymCrypt-OpenSSL/SymCrypt-OpenSSL.spec
christopherco: This bump is part of this comment - https://github.com/microsoft/azurelinux/pull/10075#discussion_r1712170768

christopherco: Looks like I misunderstood the comment. Will unbump these.

SPECS/openssl/openssl.spec
christopherco: Looks like I misunderstood the comment. Will unbump these.

SPECS/SymCrypt/SymCrypt.spec
samuel-lee-msft: Sorry for the ambiguity here - I just meant that we should run the tests from SymCrypt-OpenSSL + OpenSSL spec files when we update this SymCrypt spec file, as the test section of this build recipe is not as in depth for how callers use the produced libsymcrypt.so.

dmcilvaney (approved):

PawelWMS (approved):

PR #10110, opened 2024-08-12
Add tdnf Virtual Repo Snapshot via patch

sameluch:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

The PR adds a virtual snapshot feature to tdnf via the use of a new command line and config parameter --epochtimelimit
See this comparison for a better representation of the patch changes: https://github.com/sameluch/tdnf/compare/v3.5.6...sameluch:tdnf:add-time-bounded-install

###### Change Log  

- Add patch for tdnf virtual snapshot

**YES**

###### Test Methodology

- AMD64 Toolchain build id: https://dev.azure.com/mariner-org/mariner/_build?definitionId=2118&_a=summary
- ARM64 Toolchain build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=620287&view=results


SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,900 @@
...
+ uint32_t
+diff --git a/client/repo.c b/client/repo.c
+index b6073e2..cc29780 100644
+--- a/client/repo.c
++++ b/client/repo.c
+@@ -36,6 +36,8 @@ TDNFInitRepo(
+     Pool* pPool = NULL;
+     int nUseMetaDataCache = 0;
+     PSOLV_REPO_INFO_INTERNAL pSolvRepoInfo = NULL;
++    PTDNF_CMD_OPT pSetOpt;
```
anphel31, line 67: Set to NULL?

SPECS/tdnf/tdnf.spec
```diff
@@ -25,6 +25,10 @@ Patch3:         tdnf-printf-fix.patch
 # Patch to be removed once we upgrade to a version of tdnf which contains the upstream fix
 # https://github.com/vmware/tdnf/commit/5311b5ed0867a40ceb71b89358d70290bc2d0c51
 Patch4:         tdnf-sqlite-library.patch
+
+# Patch in vitual snapshot
+Patch5:			virtual-repo-snapshot.patch
```
jslobodzian, line 30: tabs/spaces

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,900 @@
...
+     pConf->nOpenMax = TDNF_DEFAULT_OPENMAX;
+ 
+     register_ini(NULL);
+@@ -122,6 +123,10 @@ TDNFReadConfig(
+         {
+             pConf->nInstallOnlyLimit = atoi(cn->value);
+         }
++        else if (strcmp(cn->name, TDNF_CONF_KEY_EPOCH_TIME_LIMIT) == 0)
++        {
++            pConf->nRepoEpochTimeLimit = atol(cn->value); //assumes your system's time_t is typedef long
```
jslobodzian, line 40: This should be strol or similar.  atoi is "non-safe"

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,900 @@
...
+index bb7fba1..160eca7 100644
+--- a/client/prototypes.h
++++ b/client/prototypes.h
+@@ -586,7 +586,8 @@ uint32_t
+ TDNFInitRepoFromMetadata(
+     Repo *pRepo,
+     const char* pszRepoName,
+-    PTDNF_REPO_METADATA pRepoMD
++    PTDNF_REPO_METADATA pRepoMD,
++    time_t pnRepoEpochLimit
```
jslobodzian, line 55: resconsider the "pn" here, definitely not a p and n is ok-ish

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,900 @@
...
+diff --git a/client/repo.c b/client/repo.c
+index b6073e2..cc29780 100644
+--- a/client/repo.c
++++ b/client/repo.c
+@@ -36,6 +36,8 @@ TDNFInitRepo(
+     Pool* pPool = NULL;
+     int nUseMetaDataCache = 0;
+     PSOLV_REPO_INFO_INTERNAL pSolvRepoInfo = NULL;
++    PTDNF_CMD_OPT pSetOpt;
++    time_t pnRepoEpochTimeLimit = NULL;
```
jslobodzian, line 68: ditto on the prefix here

jslobodzian, line 68: time_t should probably be memset to 0 for portability

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,900 @@
...
+ 
+     if (!pTdnf || !pRepoData || !pSack || !pSack->pPool)
+     {
+@@ -43,6 +45,21 @@ TDNFInitRepo(
+         BAIL_ON_TDNF_ERROR(dwError);
+     }
+ 
++    // set local epoch limit if conf or cmd line opt is present
++    if (pTdnf->pConf->nRepoEpochTimeLimit != NULL)
```
jslobodzian, line 77: Validated pConf

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,900 @@
...
++    // set local epoch limit if conf or cmd line opt is present
++    if (pTdnf->pConf->nRepoEpochTimeLimit != NULL) 
++    {
++        pnRepoEpochTimeLimit = pTdnf->pConf->nRepoEpochTimeLimit;
++    }
++
++    // take command line over config if both are present
++    for (pSetOpt = pTdnf->pArgs->pSetOpt; pSetOpt; pSetOpt = pSetOpt->pNext) 
++    {
++        if(strcmp(pSetOpt->pszOptName, "epochtimelimit") == 0)
```
jslobodzian, line 85: consider strncmp

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,900 @@
...
+ 
+diff --git a/common/config.h b/common/config.h
+index 222a448..1370c0d 100644
+--- a/common/config.h
++++ b/common/config.h
+@@ -19,6 +19,7 @@
+ //Conf file key names
+ #define TDNF_CONF_KEY_GPGCHECK            "gpgcheck"
+ #define TDNF_CONF_KEY_INSTALLONLY_LIMIT   "installonly_limit"
++#define TDNF_CONF_KEY_EPOCH_TIME_LIMIT    "epochtimelimit"
```
jslobodzian, line 160: use this constant above

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,900 @@
...
+diff --git a/include/tdnferror.h b/include/tdnferror.h
+index c9349a0..c207264 100644
+--- a/include/tdnferror.h
++++ b/include/tdnferror.h
+@@ -187,6 +187,15 @@ extern "C" {
+ #define ERROR_TDNF_HISTORY_ERROR 1801
+ #define ERROR_TDNF_HISTORY_NODB 1802
+ 
++#define ERROR_TDNF_FILTER_BASE 1900
++// filter MEMORY
```
jslobodzian, line 186: consider squeezing "TIME" between TDNF and FILTER or EPOCH or something to  allow for other future "FILTERS"

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,900 @@
...
+ // libsolv
+@@ -44,4 +43,10 @@
+ #include "../history/history.h"
+ #include "prototypes.h"
+ 
++#include <expat.h>
++#include <stdlib.h>
++#include <time.h>
++#include <string.h>
++#include <zlib.h>
```
jslobodzian, line 274: Is this still needed?

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: done

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: added

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
+diff --git a/solv/tdnfrepo.c b/solv/tdnfrepo.c
+index c27b907..aef6c61 100644
+--- a/solv/tdnfrepo.c
++++ b/solv/tdnfrepo.c
+@@ -9,6 +9,522 @@
+ 
+ #include "includes.h"
+ 
++// #### XML FILTER CODE ####
++void checkAndResizeBuffer(char ** pszCharBuffer, int * nBufferMaxLen, int nLengthToAdd) {
```
jslobodzian, line 299: It looks like nBufferMaxLen is not updated, but is passed as a pointer.

So...either null check the pointer...or pass nBufferMaxLen by value.

Also probably want to check that it's positive.

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    uint32_t dwError = 0;
++    int nBufferContentLen = strlen(*pszCharBuffer);
++    while (nBufferContentLen + nLengthToAdd >= *nBufferMaxLen)
++    {
++        *nBufferMaxLen *= 2;
++        char * pszTempCharBuffer = NULL;
++        pszTempCharBuffer = realloc(*pszCharBuffer, *nBufferMaxLen);
```
jslobodzian, line 306: This realloc is called in a loop.  I don't think you want to do that.

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    int nPrevElement; // enum 0 -> start, 1 -> data, 2 -> end
++
++    //set and forget on creation
++    time_t nSearchTime;
++    FILE * pbOutfile;
++} TDNFFilterData;
++
++#define TDNF_MAX_FILTER_INPUT_THRESHOLD 10000000
++
++#define BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError) \
```
jslobodzian, line 245: Is there another similar macro that does the same thing?  Do we need a new one here?

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    int nBufferContentLen = strlen(*pszCharBuffer);
++    while (nBufferContentLen + nLengthToAdd >= *nBufferMaxLen)
++    {
++        *nBufferMaxLen *= 2;
++        char * pszTempCharBuffer = NULL;
++        pszTempCharBuffer = realloc(*pszCharBuffer, *nBufferMaxLen);
++        if (!pszTempCharBuffer)
++        {
++            dwError = ERROR_TDNF_TIME_FILTER_MEMORY;
++            BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
```
jslobodzian, line 310: why not just return dwError?

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++        //set expanded char buffer 
++        *pszCharBuffer = pszTempCharBuffer;
++    }
++cleanup:
++    return;
++error:
++    goto cleanup;
++}
++
++char * descEscapeCharLinter(char * desc) {
```
jslobodzian, line 321: should be const char *

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    uint32_t dwError = 0;
++    char * amp = "&amp;";
```
jslobodzian, line 323: These should all be const char *

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    char * gt = "&gt;";
++    char * lt = "&lt;";
++
++    int nDescLen = strlen(desc) * sizeof(char) + 1;
```
jslobodzian, line 327: not sure if they are using UTF8 here.  If so, +1 won't work.  Also, why +1?

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++                pszCharToAdd = lt;
++                break;
++        }
++
++        //resize buffer if needed
++        if (pszCharToAdd != NULL)
++        {
++            nAddStrlen = strlen(pszCharToAdd);
++        }
++        checkAndResizeBuffer(&pszLintedDesc, &nLintedSize, nAddStrlen);
```
jslobodzian, line 361: If realloc failed, it's not checked here.

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
+     return dwError;
+ 
+ error:
+@@ -195,10 +713,14 @@ SolvReadYumRepo(
+     const char *pszPrimary,
+     const char *pszFilelists,
+     const char *pszUpdateinfo,
+-    const char *pszOther
++    const char *pszOther,
++    char * pszRepoEpochLimit
```
jslobodzian, line 839: const char

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++// #### END XML FILTER CODE ####
++
+ uint32_t
+ SolvLoadRepomd(
+     Repo* pRepo,
+@@ -53,6 +569,7 @@ SolvLoadRepomdPrimary(
+ {
+     uint32_t dwError = 0;
+     FILE *fp = NULL;
++
```
jslobodzian, line 821: looks like only newlines added here and a few lines below.  We should undo that change

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    {
++        free(pszCharData);
++    }
++    return;
++error:
++    goto cleanup;
++}
++
++char * 
++TDNFFilterFile(const char * pszInFilePath, char * pszEpochTime) {
```
jslobodzian, line 658: const char * on the pszEpochTime argument

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    XML_Parser bParser;
++    bzero(&bParser, sizeof(XML_Parser));
++    char pszTimeExtension[100];
++    char * pszOutFilePath = NULL;
++    time_t nEpochTime;
++    bzero(&nEpochTime, sizeof(time_t));
++    char * pszEpochEnd = NULL;
++    nEpochTime = strtoll(pszEpochTime, &pszEpochEnd, 10);
++    
++    if (nEpochTime == 0 && strcmp(pszEpochTime, "0") != 0) {
```
jslobodzian, line 673: you need to check "errno"...and pszEpochTime will equal pszEpochTime  if the parsing had an issue

errno = 0;
nEpochTime = strtoll(...)

if (errno !=0) || (pszEpochEnd == pszEpochTime  } {
   //there is an error

}

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++        printf("Error converting --epochtimelimit: %s\n", pszEpochTime);
++        dwError = ERROR_TDNF_TIME_FILTER_PARSE;
++        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
++    }
++    if (pszEpochEnd != NULL && strcmp(pszEpochEnd, "") != 0) {
```
jslobodzian, line 678: I don't think you'll need this check.  See previous comment.

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++        printf("The time provided to --epochtimelimit is misformatted: \'%s\' should be an epoch time\n", pszEpochTime);
++        dwError = ERROR_TDNF_TIME_FILTER_PARSE;
++        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
++    }
++
++    sprintf(pszTimeExtension, "-%ld.xml", nEpochTime);
++    int nOutFileLen = strlen(pszInFilePath) + strlen(pszTimeExtension);
++
++    pszOutFilePath = malloc(nOutFileLen * sizeof(char));
```
jslobodzian, line 687: Didn't leave space for the null-character

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    if (!pszOutFilePath)
++    {
++        dwError = ERROR_TDNF_TIME_FILTER_MEMORY;
++        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
++    }
++    bzero(pszOutFilePath, nOutFileLen * sizeof(char));
++
++    // find total extension length
++    int nFileExtLen = 4; // len of ".xml"
```
jslobodzian, line 696: no need to add space for the extension...because I think you already included that in pszTimeExtension above in the sprintf

jslobodzian: OR.... you need to remove the .xml from the sprintf above

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
++    }
++    pbOutFile = fopen(pszOutFilePath, "w");
++    if (!pbOutFile)
++    {
++        dwError = ERROR_TDNF_TIME_FILTER_IO;
++        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
++    }
++
++    pData = (TDNFFilterData *) malloc(sizeof(TDNFFilterData));
```
jslobodzian, line 721: Does this need to be allocated on the Heap?

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    if (pszElementBuffer)
++    {
++        free(pszElementBuffer);
++    }
++    return;
++error:
++    goto cleanup;
++}
++
++void TDNFFilterCharDataHandler(void * userData, const char * content, int length) {
```
jslobodzian, line 609: Why take void* here if the data type is known to be TDNFFilterData*

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    uint32_t dwError = 0;
++    // load tracking data
++    TDNFFilterData * pTracking = (TDNFFilterData *)userData;
```
jslobodzian, line 612: can userData be null?

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: Not inherently required. This was an artifact of the prototype design prior to using heap space for the element buffer. Will change to use a stack var.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: This is still required, nFileExtLen is used to remove the arbitrary length file extension from the original input file, unrelated to the sprintf above. The file extension at a baseline will have a ".xml" but could be ".xml.gz" or ".xml.xz" or similar. This is unknown up front, so we just start with a base of 4 and check for any compression before copying the file path into our output variable.

sameluch: That said the variable name is not quite clear, I will update to use `nInFileExtLen` instead to make this more clear to the reader what it is used for.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: Since the InFilePath does not have the file extension + compression cleaved off it will technically have enough space, but we can do better/be more exact. I have moved this calculation down and added 1 for the null char specifically.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: Yes this realloc is called within a loop. This is a way minimize the number of calls to realloc in scenarios where this function will be called back to back (character linting or adding large character chunks to the cached package element). This allows the buffer to be overestimated, rather than being realloced exactly each time potentially increasing the number of required calls to the function.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: There are similar macros... but stylistically (and in other tdnf code), adding another bail specifically for the time filter seemed appropriate (see BAIL_ON_TDNF_LIBSOLV_ERROR() for example). We could likely use `BAIL_ON_TDNF_ERROR` if that seems more appropriate.

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++        {
++            dwError = ERROR_TDNF_TIME_FILTER_MEMORY;
++            BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
++        }
++        //set expanded char buffer 
++        *pszCharBuffer = pszTempCharBuffer;
++    }
++cleanup:
++    return;
++error:
```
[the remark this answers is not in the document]
sameluch: Error handling for each function within the tdnf codebase is done using the BAIL_ON_ERROR/goto error structure. Each function expecting errors has an `error` and `cleanup` tag within. To go along with the existing method of handling errors I have added these tags to each of my functions and appropriately had the errors bail on each particular issue. 

That said, with this being brought into question and the return value above, I did convert most functions (sans the expat handler functions which are specified void function pointers) to return and bail on these dwError codes for consistency rather than having some be `void`, or NULL check a `char *`, etc. With there error printouts being placed in the top level sections on where the error can cascade up to using `pr_err` as done within other parts of the codebase.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: See below discussion on use of goto for error handling.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: done

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: Good catch, I have moved the +1 to happen prior to the multiplication of size.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: added

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: added

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -0,0 +1,906 @@
...
++    {
++        void * pszXMLParseBuffer = XML_GetBuffer(bParser, BUFSIZ);
++        if (!pszXMLParseBuffer)
++        {
++            fprintf(stderr, "Couldn't allocate memory for buffer\n");
++            dwError = ERROR_TDNF_TIME_FILTER_PARSE;
++            BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
++        }
++
++        const size_t len = fread(pszXMLParseBuffer, 1, BUFSIZ - 1, pbInFile);
```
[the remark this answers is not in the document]
sameluch: added error checking using `ferror` as an error in the stream is not explicitly reported via fread.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: This is the function handle/format used for expat's character parsing. That is why the first thing done in each handler is to cast it to TDNFFilterData. Unfortunately, I don't think we can change this easily...

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: userData cannot be null afaik. It will be a pointer to the data given to the expat parser upon creation, guaranteed by the library.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: Updated to reflect suggestion.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: Error checking using dwError has now been standardized across functions and added to each call to checkAndResizeBuffer among other functions, including this call.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: inside of the loop, prior to the realloc the value is update. But yes I should check the value is positive :)

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: no, removed.

SPECS/tdnf/tdnf.spec
sameluch: fixed

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: converted the instances of `atoi` added by my code to use `strtoll` instead as suggested.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: converted to char * to put the validation closer to use of the parameter.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: changed over to using strncmp with the epochtimelimit constant length as the bounding factor.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: fixed

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: If the realloc fails, the function will simply bail as if any other realloc would bail, causing the program to fail.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: yup, removed.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: added

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: removed.

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: added

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: validated

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: per some more discussion on a call, the realloc has been moved outside of the loop and only the size recalculation is done inside of the loop, to reduce the number of calls to realloc even further.

jslobodzian (approved):

PR #10189, opened 2024-08-20
Coal/uki

SeanDougherty:
## Changelist: 
- Adds a UKI to Steamboat's bootflow
- Adds use of the Discoverable Partition Specification to Steamboat (for automounting of `/`)
- Changes steamboat's bootloader from `grub` to `systemd-boot` as the bootloader for steamboat
- Removes read-only and overlays (temporarily) from the image because of the new use of the Discoverable Partition Specification (DPS). 
- Adds a fix to dracut from 3.0
- Reverts azure-init back to the stable 0.1.1 version
- Alphabetizes the coal-packages json


SPECS/azure-init/generate-vendor-tarball.sh
```diff
@@ -3,28 +3,28 @@
...
-# cp 0001-add-Azure-Linux-support.patch $workdir/
+cp 0001-add-Azure-Linux-support.patch $workdir/
 
 pushd $workdir
 wget $SOURCE_URL
-tar -xzvf "${SOURCE_VERSION}.tar.gz"
+tar -xzvf "${SOURCE_VERSION_V}.tar.gz"
 pushd "azure-init-${SOURCE_VERSION}"
-# patch -p1 < ../0001-add-Azure-Linux-support.patch
+patch -p1 < ../0001-add-Azure-Linux-support.patch
```
hbeberman, line 22: Curious, why are we applying a patch file within this script instead of in the spec?

SPECS/azure-init/azure-init.spec
```diff
@@ -34,7 +34,7 @@ cargo build --all
 mkdir -p %{buildroot}%{_sharedstatedir}/azure-init
 mkdir -p %{buildroot}%{_unitdir}
 install -m 0755 target/debug/azure-init %{buildroot}%{_sharedstatedir}/azure-init/azure-init
-install -m 0644 config/azure-init.service %{buildroot}%{_unitdir}/azure-init.service
+install -m 0644 config/azure-provisioning-agent.service %{buildroot}%{_unitdir}/azure-provisioning-agent.service
```
hbeberman, line 37: At this point we might as well install target this to `azure-init.service` since future version of the package will be using that anyway.

SPECS/kernel/kernel-uki-dracut.conf
```diff
@@ -31,8 +34,8 @@ drivers+=" virtio-blk virtio-scsi "      # qemu-kvm
 drivers+=" hv-storvsc "                  # hyperv
 drivers+=" xen-blkfront "                # xen
 
-# root encryption
-drivers+=" dm_crypt "
```
hbeberman, line 35: does dm_crypt need to be removed here?

toolkit/docs/steamboat/building.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Dependencies
+
+- [Standard AzL 3.0 dependencies](/toolkit/docs/building/building.md#install-prerequisites)
+- parted >=3.6
+
+_Note: If you are on Mariner 2.0, you will need to build Parted 3.6 from source. You should be able to build and install using [this srpm](https://packages.microsoft.com/azurelinux/3.0/prod/base/srpms/Packages/p/parted-3.6-1.azl3.src.rpm) from PMC_
+
+## Building an image
+Set up an AzureLinux 3 build environment with coal/experimental. When iterating on packages or images, call `sudo make clean-imagegen` b/w builds to ensure the imager's local repo stays up-to-date.
```
hbeberman, line 11: "b/w" -> "between"

SPECS/azure-init/generate-vendor-tarball.sh
SeanDougherty: I hvae a habit of applying patches to the source tarballs in case the patch makes a change to the list of vendored dependencies

SPECS/kernel/kernel-uki-dracut.conf
SeanDougherty: hmmm, good catch. I removed this early in testing, before I found the Parted fix. Let me add it back and verify.

toolkit/imageconfigs/additionalconfigs/coal.sh
```diff
@@ -24,13 +25,60 @@ echo "/data/containerd /var/lib/containerd none bind 0 0" >> /etc/fstab
...
+# sed -i '/\/ ext4/d' /etc/fstab
+# sed -i '/\/efi vfat/d' /etc/fstab
+echo > /etc/fstab
+
+# echo "fstab-after:"
+# cat /etc/fstab
+
+# copy UKI into the ESP
+mkdir -p /efi/EFI/Linux
+get_kernel_version
```
hbeberman, line 66: This function is only called once and is implicitly setting an env var for its caller. Just call `KERNEL_VERSION="$(ls /usr/lib/modules)"` directly here.

SPECS/kernel/kernel-uki-dracut.conf
SeanDougherty: Re-added and everything is working as expected. Fixed in "round 1" of addressing comments 👍

toolkit/docs/steamboat/building.md
SeanDougherty: Fixed!

toolkit/imageconfigs/additionalconfigs/coal.sh
SeanDougherty: Fixed in "round 2" of addressing comments!

hbeberman (approved):

PR #10199, opened 2024-08-20
Image Customizer: Be robust to lsblk and fdisk output ordering.

cwize1:
It seems that in most cases, the lsblk and fdisk commands output partitions in the correct order. But this isn't actually guaranteed and there have been user reports of unusual errors that are the result of assuming this is the case. This change fixes the code so that it is agnostic to the ordering of the partitions.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Ran UTs.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
@@ -34,9 +34,8 @@ func shrinkFilesystems(imageLoopDevice string, verityHashPartition *imagecustomi
 
 	// Get the start sectors of all partitions
 	startSectors, err := getStartSectors(imageLoopDevice, len(diskPartitions)-1)
-	// Number of partitions is len(diskPartitions)-1 as diskPartitions[0] refers to the loop device for the image itself
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get partitions start sectors:\n%w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
```diff
@@ -57,8 +67,18 @@ func shrinkFilesystems(imageLoopDevice string, verityHashPartition *imagecustomi
 
 		logger.Log.Infof("Shrinking partition (%s)", partitionLoopDevice)
 
+		startSector, foundStartSector := startSectors[partitionLoopDevice]
+		if !foundStartSector {
+			return fmt.Errorf("failed to find start sector for partition (%s)", partitionLoopDevice)
```
amritakohli, line 72: the errors from getStartSectors() don't seem to bubble up - is that intentional?

cwize1: I don't understand the question. The `err` returned by `getStartSectors()` is checked.

`getStartSectors()` is now returning a mapping from dev-path to start sector. The check on this line is ensuring the partition we are interested in, is actually contained within the map. This check is probably very unlikely to fail. But I like to be as robust as possible.

amritakohli: Ah never mind got it, I see it on line 38. Ignore

PR #10200, opened 2024-08-20
Suggesting test cases for build verification.

PawelWMS:
Each numbered subfolder is a slightly different scenario with multiple test cases defined in the `Readme.txt` files.

I'm also including a `bogus-generator` spec in case it may come in handy.

### DIFF
--- a/toolkit/tools/build_test_cases/1/A.spec
+++ b/toolkit/tools/build_test_cases/1/A.spec
@@ -1,0 +1,30 @@
+Summary:        Test package A
+Name:           A
+Epoch:          1
+Version:        1.0.0
+Release:        1%{?dist}
+License:        MIT
+URL:            https://test.com
+Group:          Test
+Vendor:         Microsoft
+Distribution:   Azure Linux
+
+BuildRequires:  B
+
+%description
+Test spec A.
+
+%prep
+
+%build
+
+%install
+
+%check
+
+%files
+%defattr(-,root,root)
+
+%changelog
+* Tue Aug 20 2024 Pawel Winogrodzki <pawelwi@microsoft.com> - 1.0.0-1
+- Spec created.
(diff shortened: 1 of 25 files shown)


toolkit/tools/build_test_cases/1/A.spec
```diff
@@ -0,0 +1,30 @@
...
+Epoch:          1
+Version:        1.0.0
+Release:        1%{?dist}
+License:        MIT
+URL:            https://test.com
+Group:          Test
+Vendor:         Microsoft
+Distribution:   Azure Linux
+
+BuildRequires:  B
```
PawelWMS, line 12:
```suggestion
BuildRequires:  A
```

toolkit/tools/build_test_cases/4/Readme.txt
```diff
@@ -0,0 +1,34 @@
...
+- Version 2:
+    - A and B build in parallel. A builds with upstream B.
+    - A rebuilds with the local B, since its DynDep(B) dependency is available from a newer B.
+
+
+Test case III:
+- B absent locally and upstream.
+
+Expectation:
+- A fails to build due to missing dependency.
```
dmcilvaney, line 34: We could have an option to do a fallback build of "everything" here, but agreed probably best to print a useful error to the user suggesting they trigger the package build manually once.

toolkit/tools/build_test_cases/4/Readme.txt
```diff
@@ -0,0 +1,34 @@
+Dependency on a dynamically-generated provides. Similar to tests suite 2.
```
dmcilvaney, line 1: Permutation of test 4+6:
Local copy of A dynamic BRs C
Upstream copy of A is unchanged, but does not list BR C (maybe macro environment changed?)
We have changed C locally

If we have nothing locally, I'd expect us to re-use A and avoid rebuilding
If we have a local copy of the new A, I'd expect us to now rebuild C -> A

PR #10269, opened 2024-08-27
fix: use static search path for toolchain GPG keys during validation

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

The new toolchain validation step relies on `$(TOOLCHAIN_GPG_VALIDATION_KEYS)` to check the toolchain rpms against. This variable is by default a list of gpg keys from `./SPECS/azurelinux-repos/`. 

When using `make package-toolkit` these keys are included in `./toolkit/repos`. The search is `$(wildcard $(PROJECT_ROOT)/SPECS/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)`.

This works well for common use cases, but if the user selects a custom `SPECS_DIR=...` with the core repo the validation will fail. Instead use `$(PROJECT_ROOT)/SPECS` which is a stable directory (similar to https://github.com/microsoft/azurelinux/blob/9f8d0fad669392e8c53e7334da1f481f5306e673/toolkit/scripts/preview.mk#L16).

The two supported use-cases for the toolkit are via core repo checkout, and `make package-toolkit` -> `tar -xzvf toolkit.tar.gz`. The packaged flow will always include the GPG keys in the toolkit. For the core checkout, we can always assume that the toolkit will be present inside the repo, so it should be safe to assume the GPG keys are also present. Even if the user is not using the `./SPECS` from the repo the keys are still valid.

###### Change Log  

- Use GPG keys for toolchain validation from `$(PROJECT_ROOT)/SPECS` instead of `$(SPECS_DIR)`

**YES** (Only in so much as GPG validation for some less common local build flows was broken)

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/53478906

###### Test Methodology

```
mkdir ./SPECS_TEST
cp ./SPECS/words ./SPECS_TEST/ -r
cd toolkit
git checkout 3.0-stable
sudo make clean
sudo make build-packages SPECS_DIR=../SPECS_TEST/ USE_PREVIEW_REPO=n QUICK_REBUILD_PACKAGES=y PRECACHE=n CONFIG_FILE="" REBUILD_TOOLS=y
```
Also
```
make printvar-TOOLCHAIN_GPG_VALIDATION_KEYS SPECS_DIR=not/a/real/dir
# vs
make printvar-TOOLCHAIN_GPG_VALIDATION_KEYS
```

### DIFF
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -155,7 +155,7 @@ VALIDATE_TOOLCHAIN_GPG ?= y
 endif
 endif
 
-TOOLCHAIN_GPG_VALIDATION_KEYS ?= $(wildcard $(SPECS_DIR)/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)
+TOOLCHAIN_GPG_VALIDATION_KEYS ?= $(wildcard $(PROJECT_ROOT)/SPECS/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)
 
 ######## COMMON MAKEFILE UTILITIES ########
 


PawelWMS (approved):

toolkit/Makefile
```diff
@@ -155,7 +155,7 @@ VALIDATE_TOOLCHAIN_GPG ?= y
 endif
 endif
 
-TOOLCHAIN_GPG_VALIDATION_KEYS ?= $(wildcard $(SPECS_DIR)/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)
+TOOLCHAIN_GPG_VALIDATION_KEYS ?= $(wildcard $(PROJECT_ROOT)/SPECS/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)
```
jslobodzian, line 158: What does this do if you are a derivative building outside the tree?

jslobodzian (approved):

toolkit/Makefile
dmcilvaney: `$(wildcard ...)` will return the empty string if the directory doesn't exist, and the second call to ` $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)` should return the packaged copies.

PR #10307, opened 2024-08-30
add exclude snapshot repo option to virtual snapshot

sameluch:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

After discussing snapshot functionality slightly deeper, the ability to exclude a repo from the snapshot is desired and useful. The patch has added --excludesnapshot option for the command line of tdnf

###### Change Log  

- Add --excludesnapshot patch to tdnf

**YES**

###### Test Methodology

- Local testing


SPECS/tdnf/tdnf.spec
```diff
@@ -219,6 +219,9 @@ fi
 /%{_lib}/systemd/system/tdnf*
 
 %changelog
+* Thu Aug 29 2024 Sam Meluch <sammeluch@microsoft.com> - 3.5.6-3
+- Add virtual repo snapshot exclude
```
PawelWMS, line 223: Is there a reason we want a separate argument for the snapshot instead of relying on the existing `--disablerepo` and `--enablerepo`? I'm worried that having two may be confusing when using the snapshot.

sameluch: The reason to add a snapshot exclude option is to allow for a repo containing new packages locally built to still be included when building an image, but not from upstream. This is also additional flexibility within the feature for tdnf, not just applicable to our use case for allowing local RPMs from within the toolkit.

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -1133,3 +1133,212 @@ index 0558611..18c84fc 100644
...
+ 
++uint32_t
++TDNFExcludeFromSnapshot(
++    PTDNF_REPO_DATA pRepos,
++    const char* pszId
++    )
++{
++    uint32_t dwError = 0;
++    int nIsGlob = 0;
++    if(!pRepos && IsNullOrEmptyString(pszId))
```
jslobodzian, line 1220: && or || here?

SPECS/tdnf/virtual-repo-snapshot.patch
```diff
@@ -1133,3 +1133,212 @@ index 0558611..18c84fc 100644
...
+index 86c34e7..4097203 100644
+--- a/tools/cli/lib/help.c
++++ b/tools/cli/lib/help.c
+@@ -23,8 +23,8 @@ static const char *help_msg =
+  "           [--downloadonly]\n"
+  "           [--enablerepo=<repoid>]\n"
+  "           [--enableplugin=<plugin_name>]\n"
+- "           [--snapshottime=<POSIX_time>]\n"
+  "           [--exclude [file1,file2,...]]\n"
++ "           [--excludesnapshot=<repoid>]\n"
```
jslobodzian, line 1309: can we rename this to something like --snapshotexclusionrepos so that it sorts with snapshottime.  And helps clarify that the snapshot exclusion applies to a repo.

Also...how big of a stretch from here to take a comma separated list of repos (I know we discussed supporting only 1)

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: Good catch, this is actually copy paste from upstream, we can likely contribute this small fix upstream if not already present

SPECS/tdnf/virtual-repo-snapshot.patch
sameluch: Updated to `--snapshotexcluderepos` to make the names come together.

I also added the ability to specify multiple repos in one go as a comma separated list in line.

sameluch: tested by running the following commands to check if the second repo was snapshot when using the updated tdnf:
`tdnf install valgrind --snapshottime=1724119509 --snapshotexcluderepos=azurelinux-official-base`
`tdnf install valgrind --snapshottime=1724119509 --snapshotexcluderepos=azurelinux-official-base,azurelinux-official-base-debug`
image

jslobodzian (approved):

PR #10345, opened 2024-09-04
libsolv: enable zstd support to match createrepo_c

reubeno:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

In 3.0, `createrepo_c`'s default behavior is to create repodata files compressed using zstd; prior to this change, `libsolv` didn't support `zstd`, though, so `tdnf`/`dnf`/et al. couldn't actually handle those repos. As a workaround, the toolkit and manual invocations of `createrepo_c` for 3.0 have needed to use the `--compatibility` flag. Standard OSS tooling doesn't know to do this (nor should it have to), though; the right thing to do is to add support for zstd to `libsolv`, as other distros have done.

###### Change Log  

Enable zstd support in `libsolv` at compile time.

Yes.

###### Test Methodology

Local build + pipeline build through and including toolchain (phase 1).

### DIFF
--- a/SPECS/libsolv/libsolv.spec
+++ b/SPECS/libsolv/libsolv.spec
@@ -10,7 +10,7 @@ Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 BuildRequires:  cmake
 BuildRequires:  rpm-devel
-BuildRequires:  libzstd-devel
+BuildRequires:  zstd-devel
 Requires:       expat-libs
 
 %description
--- a/SPECS/libsolv/libsolv.spec
+++ b/SPECS/libsolv/libsolv.spec
@@ -1,7 +1,7 @@
 Summary:        A free package dependency solver
 Name:           libsolv
 Version:        0.7.28
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        BSD
 URL:            https://github.com/openSUSE/libsolv
 Source0:        https://github.com/openSUSE/libsolv/archive/refs/tags/%{version}.tar.gz#/%{name}-%{version}.tar.gz
@@ -10,6 +10,7 @@ Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 BuildRequires:  cmake
 BuildRequires:  rpm-devel
+BuildRequires:  libzstd-devel
 Requires:       expat-libs
 
 %description
@@ -47,7 +48,8 @@ Requires:       xz
     -DENABLE_RPMDB_BYRPMHEADER=ON       \
     -DENABLE_RPMDB_LIBRPM=ON            \
     -DENABLE_RPMMD=ON                   \
-    -DENABLE_COMPS=ON
+    -DENABLE_COMPS=ON                   \
+    -DENABLE_ZSTD_COMPRESSION=ON
 %make_build
 
 %install
@@ -77,6 +79,9 @@ find %{buildroot} -type f -name "*.la" -delete -print
 %{_mandir}/man1/*
 
 %changelog
+* Wed Sep 04 2024 Reuben Olinsky <reubeno@microsoft.com> - 0.7.28-2
+- Enable zstd support to match createrepo_c.
+
 * Wed Feb 07 2024 Alberto David Perez Guevara <aperezguevar@microsoft.com> - 0.7.28-1
 - Upgrade to version 0.7.28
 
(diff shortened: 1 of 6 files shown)


SPECS/libsolv/libsolv.spec
```diff
@@ -10,6 +10,7 @@ Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 BuildRequires:  cmake
 BuildRequires:  rpm-devel
+BuildRequires:  libzstd-devel
```
cwize1, line 13: I suspect you might need to update the [build_official_toolchain_rpms.sh](https://github.com/microsoft/azurelinux/blob/3.0/toolkit/scripts/toolchain/build_official_toolchain_rpms.sh) file to handle the new dependency between `libsolv` and `libzstd-devel`.

cwize1, line 13: Does this increase the size of the base image? If so, by how much?

reubeno: Good question! I still need to build the artifacts to confirm, but `rpm` already has a runtime `Requires` on `libzstd.so.1()(64bit)` and that's exactly what `libsolv` ends up requiring too with this change. It seems safe to say that anywhere `libsolv` is meaningfully used in our system that `rpm` would be expected to be there too, so I would not expect to see an image size increase.

reubeno: Thanks for the call-out; I'm not familiar with this script and how dependencies are encoded there.

I see that it already has `build_rpm_in_chroot_no_install zstd` (right near similar for `gzip`) but it's not clear to me if that's sufficient. @cwize1 would you mind taking a quick look and provide some guidance?

cwize1: That script is basically a manually specified topological ordering of the packages that make up the toolkit. Every package in the toolchain needs to have a `build_rpm_in_chroot_no_install` line. However, for packages that need to be installed to build other packages (i.e. `BuildRequires`), then you need to add a `chroot_and_install_rpms` line for that package as well. We also typically add comments describing what the build dependencies are.

For example, currently there is a line that says:

`# libsolv needs cmake`

This should probably be updated to:

`# libsolv needs cmake, libzstd`

And the following line should be added:

`chroot_and_install_rpms libzstd`

Once you think you have it right, use the `[NoArch-0-OneBranch]-Dev-PrepareBuild` pipeline to test it.

PawelWMS: And for local build debugging (if the pipeline is going to fail for some reason), you can also download the 3.0 raw toolchain, put it under `build/toolchain/toolchain_from_container.tar.gz`, then run `touch` on it to let `make` know it's fresh and run a local `make toolchain QUICK_REBUILD=y`. Alternatively, you can create this PR against `fasttrack/3.0` and that'll trigger the PR check, which builds the toolchain as the first step anyway and publishes the built RPMs.

toolkit/scripts/toolchain/build_official_toolchain_rpms.sh
```diff
@@ -541,8 +541,9 @@ build_rpm_in_chroot_no_install libxslt
 chroot_and_install_rpms pam
 build_rpm_in_chroot_no_install docbook-style-xsl
 
-# libsolv needs cmake
+# libsolv needs cmake, zstd-libs
 chroot_and_install_rpms cmake
+chroot_and_install_rpms zstd-libs
```
PawelWMS, line 546:
```suggestion
chroot_and_install_rpms zstd
```

toolkit/scripts/toolchain/build_official_toolchain_rpms.sh
```diff
@@ -541,8 +541,9 @@ build_rpm_in_chroot_no_install libxslt
 chroot_and_install_rpms pam
 build_rpm_in_chroot_no_install docbook-style-xsl
 
-# libsolv needs cmake
+# libsolv needs cmake, zstd-libs
```
PawelWMS, line 544: In the spec you've got a BR on `libzstd-devel`, which translates to `zstd-devel`, so I think either this comment or the spec is inaccurate.

reubeno: Thanks for the feedback. I've updated the comment and arguments for consistency.

SPECS/libsolv/libsolv.spec
cwize1: If `[AMD64-1-OneBranch]-Dev-BuildToolchain` succeeds, then you should be good.

PawelWMS (approved):

anphel31 (approved):

reubeno:
@PawelWMS -- I forgot to explicitly mention, but I just wanted to confirm that the phase 1 `Dev-BuildToolchain` pipeline succeeds with the latest changes: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=648440&view=results (with `Force re-building raw toolchain` *not* selected).

PR #10347, opened 2024-09-04
mock: upgrade and port from extended to core (including dependencies)

reubeno:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Upgrades `mock` to a usable version that will run on Azure Linux 3.0 and upgrades it and its required dependencies to core from extended. Since the 2.0 -> 3.0 upgrade, the version of `mock` present in extended hasn't been able to run correctly.

###### Change Log  
- Upgrade `mock` and move to core
- Upgrade `mock-core-configs` and move to core (part of `mock`)
- Upgrade `distribution-gpg-keys` and move to core (required by `mock`)
- Upgrade `libuser` and move to core (required by `mock`)
- Add `python-backoff` (required by `mock`)
- Upgrade `python-pyroute2` and move to core (required by `mock`)
- Add `python-rpmautospec-core` (required by `mock`)
- Upgrade `python-templated-dictionary` and move to core (required by `mock`)
- Upgrade `usermode` and move to core (required by `mock`)

No.

###### Test Methodology
Local build + buddy build.


SPECS/distribution-gpg-keys/distribution-gpg-keys.spec
```diff
@@ -1,7 +1,9 @@
 Summary:	GPG keys of various Linux distributions
 Name:		distribution-gpg-keys
-Version:	1.104
+Version:	1.105
 Release:	1%{?dist}
+Vendor:     Microsoft Corporation
+Distribution: Azure Linux
```
PawelWMS, line 6: Small request: could I ask you to align the spacing in the new tags?

SPECS/distribution-gpg-keys/distribution-gpg-keys.spec
```diff
@@ -1,7 +1,9 @@
...
 License:	CC0
 URL:		https://github.com/rpm-software-management/distribution-gpg-keys
 # Sources can be obtained by
```
PawelWMS, line 9: This looks a bit ugly. We try to hold a higher bar for core packages, where the source tarball is taken directly from an upstream or we have a `generate_source_tarball.sh` script for hand-made tarballs. Script's name is important, as it's ran by our auto-upgrade pipelines.

SPECS/libuser/libuser.spec
```diff
@@ -149,6 +146,9 @@ export PYTHONPATH
 %endif
 
 %changelog
+* Wed Aug 28 2024 Reuben Olinsky <reubeno@microsoft.com> - 0.64-1
```
PawelWMS, line 149: Couldn't add the comment in the `%check` section, adding it here:

The way it's detected if a ptest have failed or succeeded in our builds, is from the exit code of the last command. Not ideal, and there's some history behind that, I can get into details offline, if you're curious. Bottom line right now is that the way `%check` is currently written, a failing `%make_build check` run will not mark the tests as failing. Considering that, I'd like to ask you to reorganize `%check` a little.

One way we've worked around this elsewhere is by keeping a `test_ok` variable initialized to `true` and tracking return codes for all ran test commands and running `$test_ok` as the last command. Every time we see an error, we'd do `test_ok=false`. That's not required, though, it's just one idea. Example in `edk2.spec`.

SPECS/mock-core-configs/mock-core-configs.spec
```diff
@@ -1,54 +1,76 @@
...
 Release:    1%{?dist}
-License:    GPLv2+
+Vendor:     Microsoft Corporation
+Distribution: Azure Linux
+Summary:    Mock core config files basic chroots
+
+License:    GPL-2.0-or-later
 URL:        https://github.com/rpm-software-management/mock/
 # Source is created by
 # git clone https://github.com/rpm-software-management/mock.git
```
PawelWMS, line 15: Similar ask regarding the source tarball: can we use the upstream directly or create a script?

SPECS/mock-core-configs/mock-core-configs.spec
```diff
@@ -1,54 +1,76 @@
...
 # cd mock/mock-core-configs
 # git reset --hard %%{name}-%%{version}
 # tito build --tgz
 Source:     https://github.com/rpm-software-management/mock/archive/refs/tags/%{name}-%{version}-1/%{name}-%{version}-1.tar.gz#/%{name}-%{version}.tar.gz
+BuildArch:  noarch
+
+# The mock.rpm requires this.  Other packages may provide this if they tend to
+# replace the mock-core-configs.rpm functionality.
+Provides: mock-configs
```
PawelWMS, line 24: Should this `Provide` be versionless?

SPECS/mock-core-configs/mock-core-configs.spec
```diff
@@ -62,15 +84,64 @@ fi
...
 fi
-mock_arch=$(python -c "import rpmUtils.arch; baseArch = rpmUtils.arch.getBaseArch(); print baseArch")
-cfg=%{?fedora:fedora}%{?rhel:epel}%{?mageia:mageia}-$ver-${mock_arch}.cfg
+
+cfg=unknown-distro
+%if 0%{?fedora}
+cfg=fedora-$ver-$mock_arch.cfg
+%endif
+%if 0%{?azl}
+cfg=azurelinux-3.0-$mock_arch.cfg
```
PawelWMS, line 118: We shouldn't hard-code a version in the spec:
```suggestion
cfg=azurelinux-%{azl}.0-$mock_arch.cfg
```

PawelWMS: Or, since the `ver` variable seems to be derived for this exact purpose:
```suggestion
cfg=azurelinux-%ver-$mock_arch.cfg
```

SPECS/mock/mock.spec
```diff
@@ -1,55 +1,136 @@
...
 %global __python %{__python3}
 %global python_sitelib %{python3_sitelib}
 
-Summary:        Builds packages inside chroots
-Name:           mock
-Version:        2.16
-Release:        2%{?dist}
-License:        GPLv2+
+Summary: Builds packages inside chroots
+Name: mock
```
PawelWMS, line 11: Can you bring back the alignment in the tags?

SPECS/mock/mock.spec
```diff
@@ -1,55 +1,136 @@
...
+Version: 5.6
+Release: 1%{?dist}
+Vendor:  Microsoft Corporation
+Distribution: Azure Linux
+License: GPL-2.0-or-later
 # Source is created by
 # git clone https://github.com/rpm-software-management/mock.git
```
PawelWMS, line 18: Similar sources tarball request.

SPECS/mock/mock.spec
```diff
@@ -1,55 +1,136 @@
...
+%else
+Requires: usermode
+%endif
+Requires: createrepo_c
+
+# We know that the current version of mock isn't compatible with older variants,
+# and we want to enforce automatic upgrades.
+Conflicts: mock-core-configs < 33
+
+# Requires 'mock-core-configs', or replacement (GitHub PR#544).
```
PawelWMS, line 38: Is this our GitHub PR number? If not, it may be confusing in the future.

SPECS/python-pyroute2/python-pyroute2.spec
```diff
@@ -1,90 +1,60 @@
...
-PyRoute2 provides several levels of API to work with Netlink
-protocols, such as Generic Netlink, RTNL, TaskStats, NFNetlink,
-IPQ.
+Name: python-%{srcname}
+Version: 0.7.12
+Release: 1%{?dist}
+Vendor: Microsoft Corporation
+Distribution: Azure Linux
+Summary: Pure Python netlink library
+License: GPL-2.0-or-later OR Apache-2.0
```
PawelWMS, line 14: This reminded me we were trying to use SPDX short identifiers for licenses at some point. I've got no problem here but now I'm not sure, if this is written down anywhere, so it might be worth clarifying for any future plans we may have.

SPECS/python-pyroute2/python-pyroute2.spec
```diff
@@ -1,90 +1,60 @@
...
-tar xzvf %{SOURCE2}
-tar xzvf %{SOURCE3}
-tar xzvf %{SOURCE4}
-tar xzvf %{SOURCE5}
-tar xzvf %{SOURCE6}
-tar xzvf %{SOURCE7}
+%autosetup -n %{srcname}-%{version}
+
+%generate_buildrequires
+%pyproject_buildrequires
```
PawelWMS, line 36: Since this doesn't work for us are we sure we have the right build time dependencies? I'm guessing the package builds, but I'm worried we may be missing some features disabled during build time due to missing Python modules.

SPECS/python-rpmautospec-core/python-rpmautospec-core.spec
```diff
@@ -0,0 +1,101 @@
...
+%py3_install
+echo '%{python3_sitelib}/%{srcname}*' > %{pyproject_files}
+%endif
+
+%check
+%pytest
+
+%files -n python3-%{canonicalname} -f %{pyproject_files}
+%doc README.md
+%if %{without pyproject_build}
```
PawelWMS, line 94: Why is the license included conditionally? Or is the install logic when we're building with `pyproject_build` already taking care of the license?

SPECS/python-templated-dictionary/python-templated-dictionary.spec
```diff
@@ -1,44 +1,63 @@
...
+Release:    5%{?dist}
+Vendor:     Microsoft Corporation
+Distribution: Azure Linux
+Summary:    Dictionary with Jinja2 expansion
+
+License:    GPL-2.0-or-later
+URL:        https://github.com/xsuchy/templated-dictionary
+
+# Source is created by:
+# git clone https://github.com/xsuchy/templated-dictionary && cd templated-dictionary
```
PawelWMS, line 19: Another request to use upstream or a script for the sources.

SPECS/mock-core-configs/mock-core-configs.spec
reubeno: Ah, good catch. These comments came from the Fedora version of the spec; I had the same concern and was able to directly use the upstream source archive directly. I'll appropriately update the comments. (The URLs of the archives listed in cgmanifest is accurate.)

SPECS/distribution-gpg-keys/distribution-gpg-keys.spec
reubeno: Good call-out. The comment is erroneous, inherited from the Fedora version of this spec. I've removed the comment to reflect that it's really just taken from the listed upstream source.

SPECS/mock/mock.spec
reubeno: Yep! See above, please.

SPECS/python-pyroute2/python-pyroute2.spec
reubeno: I needed to add the `BuildRequires` listed above because we don't have support for the dynamic dependencies. I will remove these lines to avoid confusion.

SPECS/mock-core-configs/mock-core-configs.spec
reubeno: This was taken directly from the Fedora spec. I can add `= %{version}-%{release}` if that's preferred.

SPECS/mock-core-configs/mock-core-configs.spec
reubeno: FYI, `$ver` doesn't work as-is because of the additional `.0`, but I'd agree `$ver.0` is still cleaner. I'll go with that unless you feel strongly otherwise.

SPECS/python-pyroute2/python-pyroute2.spec
reubeno: Happy to chat more about this, but I'll assume this is fine as-is for the purposes of this PR.

SPECS/python-rpmautospec-core/python-rpmautospec-core.spec
reubeno: Yes, my read of the spec matches your supposition. When building with `pyproject_build`, `%{pyproject_files}` ends up including `LICENSE` listed in it, and the `-f %{pyproject_files}` takes care of including that file appropriately. (Note the `sed` logic used to specifically annotate that entry with `%license`.) When building without `pyproject_build`, the `LICENSE` needs to be explicitly added because `%{pyrpject_files}` is manually generated during `%install`, and presumably without the `LICENSE` file being included.

SPECS/libuser/libuser.spec
reubeno: Yes, would be great to learn more about that history sometime; but thanks for the context regardless. I've taken your suggestion and applied it.

SPECS/mock-core-configs/mock-core-configs.spec
PawelWMS: I'm never fully sure about `Provides`, so I'm not going to pretend int's preferred. I low-key hoped you may simply know more here.:)

SPECS/mock-core-configs/mock-core-configs.spec
PawelWMS: No, this is what I was thinking about, just didn't notice `$ver` is just the major version number.

PawelWMS (approved):

reubeno:
@PawelWMS thanks for all the feedback on this PR--definitely some good learning for me along the way.

When you get a chance, would you please look at the updates I made to see if they address your questions and concerns? I know there's a conflict with the license map right now, but I'm inclined to hold off on (re-)resolving that / (re-)sync'ing until getting a general thumbs up from you on the meaningful changes herein.

(As a side note, is there something we could do to decrease conflicts on LICENSES-MAP.md? Even splitting up the giant table cells into multiple rows on multiple textual lines would likely decrease conflicts and enable auto-merging without sacrificing the actual contents. Would there be support for doing that?)

PawelWMS (approved):

PR #10348, opened 2024-09-04
kernel: change space/tab combination into the right amount of spaces

ddstreet:
Non functional change; only converts a single tab char to the correct number of spaces.


rlmenge:
I addressed this in https://github.com/microsoft/azurelinux/pull/10365/files :)

PR #10389, opened 2024-09-09
Fixed toolchain tests blocking non-toolchain packages' tests.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

We have a bug in the PR check where a change to a toolchain package prevents ptest from being run for all non-toolchain packages. This change moves toolchain ptests to a separate job, so the original `BuildAndTest` will build and test only non-toolchain packages.

Requires #10392 to be merged first.

No.

###### Test Methodology

- PR check.


.pipelines/templates/PackageBuild.yml
```diff
@@ -37,6 +37,14 @@ parameters:
     # - name: build-artifacts
     #   rpmsTarball: cache.tar.gz
 
+  - name: isAllowToolchainRebuilds
+    type: string
+    default: "default"
```
anphel31, line 42: Why not default to true or false? and how is the value of "default" handled? (it seems to be skipped by the if/elif case below)

anphel31 (approved):

.pipelines/templates/PackageBuild.yml
PawelWMS: With "default" I can pass no arguments to the toolkit and let toolkit's defaults take over. If I forced "true" or "false" here, then I'd need to keep up with every update of the defaults in the toolkit.

PR #10404, opened 2024-09-10
Image Customizer: Add support for 'vfat' filesystem.

cwize1:
The FAT filesystem is a tad confusing. Technically, FAT12, FAT16, and FAT32 refer to a fat filesystem with 12, 16, and 32 bit addressing respectively. However, you can't just use FAT32 for everything since each variant has different min and max partition sizes. So, instead you have to dynamically select the correct one based on the partition size. However, a lot of partioning tools (including Windows) hide this complexity and just use 'fat32' to mean either FAT12, FAT16, or FAT32.

There is also VFAT which refers to any FAT filesystem variants where long filename support is enabled. Nowadays it is pretty much always enabled. So, it isn't unusual for tools (including the Linux kernel) to use 'vfat' as a generic term to refer to any of the FAT filesystem variants.

The image customizer tool currently supports 'fat32' to mean dynamically selecting between FAT12, FAT16, or FAT32. This change adds 'vfat' to mean the same thing.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Ran tests.

### DIFF
--- a/toolkit/tools/imagecustomizerapi/storage_test.go
+++ b/toolkit/tools/imagecustomizerapi/storage_test.go
@@ -123,7 +123,7 @@ func TestStorageIsValidBadEspFsType(t *testing.T) {
 
 	err := storage.IsValid()
 	assert.Error(t, err)
-	assert.ErrorContains(t, err, "ESP partition must have 'fat32' filesystem type")
+	assert.ErrorContains(t, err, "ESP partition must have 'fat32' or 'vfat' filesystem type")
 }
 
 func TestStorageIsValidBadBiosBootFsType(t *testing.T) {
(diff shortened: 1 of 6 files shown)


toolkit/tools/imagecustomizerapi/storage_test.go
```diff
@@ -123,7 +123,7 @@ func TestStorageIsValidBadEspFsType(t *testing.T) {
 
 	err := storage.IsValid()
 	assert.Error(t, err)
-	assert.ErrorContains(t, err, "ESP partition must have 'fat32' filesystem type")
+	assert.ErrorContains(t, err, "ESP partition must have 'fat32' or 'vfat' filesystem type")
```
PawelWMS, line 126: Might be also worth adding a test case specifically for VFAT.

PawelWMS (approved):

toolkit/tools/imagecustomizerapi/storage_test.go
cwize1: This test actually is setting the partition to `ext4` to trigger this error. So, it doesn't make sense to create a second one for vfat.

However, I did add a no-error test that uses vfat.

PawelWMS (approved):

PR #10452, opened 2024-09-13
Revert to 1.19.4, add epoch and add patch for CVE-2024-37371 and CVE-2024-37370

nicogbg:
Revert to 1.19.4, add epoch and add patch for CVE-2024-37371 and CVE-2024-37370

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Minor upgrade of krb5 to 1.21.3 caused regression in OS. This PR is reverting back to 1.19.4 and adding patches for two important cves [CVE-2024-37370](https://github.com/advisories/GHSA-wvrw-2fv8-cjvx) & [CVE-2024-37371](https://github.com/advisories/GHSA-8wpj-v5qv-3wf4). 

The revert will use epoch so that the 1.19.4 version is recognized as the "latest" version. The motivation behind this revert was that 1.21 upgrade broke a downstream partner team that uses krb5 for Powershell Session connecting from AZL to Windows.

In terms of the patch:  The CVE-3730.patch file that is being used from the upstream commit, but we are not including three of the new package tests: `test_cfx_altered_ec, test_cfx_short_plaintext and test_cfx_large_ec`  because they require a 1.20+ version of internal function logic that was heavily modified in this upstream PR: [Always use platform PRNG by greghudson · Pull Request #1219 · krb5/krb5 (github.com)](https://github.com/krb5/krb5/pull/1219/files). 

This PR also introduces a patch to include the k5-der.h file as it is needed for the CVE-3730.patch. 

###### Change Log  

- Change
- Change
- Change

**YES**

###### Associated issues  

- #xxxx

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2024-37370
- https://nvd.nist.gov/vuln/detail/CVE-2024-37371

###### Test Methodology

- Pipeline build id: xxxx


PawelWMS (approved):

SPECS/krb5/krb5.spec
```diff
@@ -3,16 +3,20 @@
 
 Summary:        The Kerberos newtork authentication system
 Name:           krb5
-Version:        1.21.3
-Release:        2%{?dist}
+Epoch:          1
```
PawelWMS, line 6: The `%{name} = %{version}-%{release}` dependency in the `devel` and `lang` subpackages would now also need to be changed to `%{name} = %{epoch}:%{version}-%{release}`. Same thing applies for all other packages with a dependency on a specific version of `krb5`.

aditjha-msft: Updated the devel, lang, and lib lines. Please let me know if I missed anything.

PawelWMS: This spec looks good. Please also check other specs for dependencies (both `Requires` and `BuildRequires`) on a specific version of `krb5` and its packages. I think I haven't seen any but please verify my findings.

SPECS/krb5/CVE-2024-37370.patch
```diff
@@ -26,8 +26,8 @@ version_fixed: 1.21.3
  src/lib/gssapi/krb5/k5sealv3.c    |   5 +
  src/lib/gssapi/krb5/k5sealv3iov.c |   3 +-
  src/lib/gssapi/krb5/k5unsealiov.c |  80 +++++++++-
- src/tests/gssapi/t_invalid.c      | 233 +++++++++++++++++++++++++-----
- 4 files changed, 275 insertions(+), 46 deletions(-)
```
PawelWMS: Why did you remove the tests? They seem to be present in the [original CVE patch](https://github.com/krb5/krb5/commit/55fbf435edbe2e92dd8101669b1ce7144bc96fef#diff-39f0036d8943271a8f7510ab2dafef5a3a4b8f35d36a3395a65447d9bf1c630b).

PawelWMS (changes requested):
Need more details about the removed tests.

SPECS/krb5/krb5.spec
aditjha-msft: I also checked in the codebase and no dependency is specifying specifically a krb5 version, only `>= {krb5-version}` so I think we're good on that.

SPECS/krb5/CVE-2024-37370.patch
aditjha-msft: Replied in email thread as well but adding here: The new tests that are added as part of the patch cause the t_invalid.c tests to fail as there is an abort() being called on one of the return values of an internal function. I These specific tests work fine on 1.21.3 but given that they are having issues in 1.19.4, I believe there has to be some changes in between the versions in their codebase that is impacting the behavior of some of these internal functions.

jslobodzian (approved):

SPECS/krb5/krb5.spec
```diff
@@ -3,21 +3,25 @@
...
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 Group:          System Environment/Security
 URL:            https://web.mit.edu/kerberos/
 Source0:        https://kerberos.org/dist/%{name}/%{maj_version}/%{name}-%{version}.tar.gz
 Source1:        krb5.conf
-Patch0:         CVE-2024-26461.patch
+Patch0:         CVE-2023-36054.patch
+Patch1:         CVE-2024-26461.patch
+Patch2:         k5-der-header.patch
```
PawelWMS, line 18: Can you add a comment explaining why this particular patch is needed?

SPECS/krb5/krb5.spec
```diff
@@ -3,21 +3,26 @@
...
 Distribution:   Mariner
 Group:          System Environment/Security
 URL:            https://web.mit.edu/kerberos/
 Source0:        https://kerberos.org/dist/%{name}/%{maj_version}/%{name}-%{version}.tar.gz
 Source1:        krb5.conf
-Patch0:         CVE-2024-26461.patch
+Patch0:         CVE-2023-36054.patch
+Patch1:         CVE-2024-26461.patch
+# The k5-der.h file is needed to for the functions in CVE-2024-37370.patch to work.
+Patch2:         k5-der-header.patch
```
PawelWMS, line 19: Just had a short conversation with Jon and since this is done only for the CVE fix, please just bundle it inside `CVE-2024-37370.patch`. That way it's clearer they go together. Btw. why was this header patch needed as a separate change? Is it due to the CVE fix being a backport from a newer version?

aditjha-msft: that is correct. the newer versions have this small header file and the main cve patch for 3730 and 3731 is including this header file in order to work. I will try to bundle it inside the other CVE patch

PR #10464, opened 2024-09-17
Upgrade tdnf to version 3.5.8 and Fix the ptests

sameluch:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Currently the ptests for tdnf don't run, also we are taking a minor upgrade for tdnf to go alongside the snapshot work. PR stops skipping tests, additionally makes them compatible with the chroot environment.

###### Change Log  

- Upgrade tdnf to version 3.5.8
- Fix the ptests to run for azure linux

**YES**

###### Test Methodology

- Local Testing
- sudo make build-packages REBUILD_TOOLS=y CONFIG_FILE= RUN_CHECK=y TEST_RERUN_LIST="tdnf" SRPM_PACK_LIST="tdnf" DAILY_BUILD_ID=lkg REBUILD_TOOLCHAIN=n -j100
image
- full build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=641602&view=results


dmcilvaney (approved):

SPECS/tdnf/fix-tests-for-azl.patch
```diff
@@ -0,0 +1,157 @@
...
++gpgkey=file:///etc/pki/rpm-gpg/MICROSOFT-RPM-GPG-KEY
+ gpgcheck=0
+ enabled=0
+ EOF
+@@ -146,7 +146,7 @@ cat << EOF > ${TEST_REPO_DIR}/yum.repos.d/photon-test-src.repo
+ [photon-test-src]
+ name=basic
+ baseurl=http://localhost:8080/photon-test-src
+-gpgkey=file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY
++gpgkey=file:///etc/pki/rpm-gpg/MICROSOFT-RPM-GPG-KEY
```
PawelWMS, line 65: For the key paths, it might be easier to instead link/copy them in the `%check` section to match the expectations from the tests. Specs tend to be more friendly to view if we ever change the key name/path and my hope is it'll be harder to miss this workaround at that time.

SPECS/tdnf/tdnf.spec
```diff
@@ -121,23 +129,30 @@ systemd services for periodic automatic update
...
+%endif
+
+mkdir -p build && cd build
 %cmake \
     -DCMAKE_BUILD_TYPE=Debug \
     -DBUILD_SHARED_LIBS=OFF \
     -DCMAKE_INSTALL_LIBDIR:PATH=%{_libdir} \
-    -DHISTORY_DB_DIR=%{_tdnf_history_db_dir}
+    -DHISTORY_DB_DIR=%{_tdnf_history_db_dir} \
+	..
```
PawelWMS, line 143: Please use spaces instead of tabs.

SPECS/tdnf/tdnf.spec
```diff
@@ -121,23 +129,30 @@ systemd services for periodic automatic update
 %autosetup -p1
 
 %build
+%if 0%{?with_check}
+# remove problematic test file when running tests from within the rpm build directory
+rm pytests/tests/test_srpms.py
```
PawelWMS, line 134: Two questions here:
- What is wrong with the tests? I'm reluctant to remove tests without a good reason. If it's something on our end, we should fix it. If it's a different case, please add some details as to why it's OK to remove these tests.
- Is this really ran from the `%build` stage? It would be ideal, if we could keep all test steps inside the `%check` section, although I'm aware that's not always possible.

SPECS/tdnf/tdnf.spec
```diff
@@ -121,23 +129,30 @@ systemd services for periodic automatic update
...
-%cmake_build
-
-cd %{__cmake_builddir}
+%make_build
 %make_build python
 
 %check
-pip3 install pytest requests pyOpenSSL
-cd build && make %{?_smp_mflags} check
+pip3 install pytest pyOpenSSL flake8
```
PawelWMS, line 149: Two things:
- Since we're here, could you freeze the versions of the external Python modules to keep the tests stable?
- We do have `python-flake8` as a package. Can we use that instead of the external one?

SPECS/tdnf/tdnf.spec
PawelWMS:
```suggestion
    ..
```

SPECS/tdnf/tdnf.spec
sameluch: The tests count and check the srpms present within the build expected RPM build directory. We don't necessarily know what exactly will be in this directory at the time of build, so these tests did not make much sense when this is the case and when being executed within the build directory. Additionally, the test's expected cleanup is to delete the entire directory. By doing so, the entire source directory and actively running tests are removed. Thus I've opted to remove thei file as to not be included prior to being built/included by cmake within the build directory.

Alternative to this, a skip directive could be added to each test, and the cleanup function could be removed as to no blow away the source. Though removing this in the spec `%build` or `%prep` section seemed much cleaner. I opted to include it within `%build` as it seemed more relevant to building rather than prepping to build.

SPECS/tdnf/tdnf.spec
sameluch: I went ahead and added version specifications here for the packages specified. We offer flake8 in extended and I would hesitate to migrate it forward into core (and upgrade it) for this PR as the purpose of a minor upgrade. Alternatively I have locked the version inline on the install command.

SPECS/tdnf/tdnf.spec
sameluch: fixed.

SPECS/tdnf/tdnf.spec
PawelWMS: That was dumb of me - I missed the fact that `flake8` is an extended package. Sorry for the trouble.

SPECS/tdnf/fix-tests-for-azl.patch
PawelWMS: Yup, or just add a symbolic link before we run the tests.

SPECS/tdnf/tdnf.spec
PawelWMS: Got it. In that case, could I ask you to add that explanation you just gave here about why we need to remove the tests? A shorted version is fine, though.:)

And can you check, if moving this to the `%check` section is still fine? That way it's clearer this is a test-only change.

sameluch: I will check if its possible to remove during the check section, though I suspect it may have to remain in build or prep.

sameluch: Explanation added, also moved removal to check section. Since cmake duplicates the file in the build directory, it also is removed there for safety.

SPECS/tdnf/fix-tests-for-azl.patch
sameluch: Symbolic link added in during check section.

PawelWMS (approved):

PR #10475, opened 2024-09-17
Update openssl to 3.3.2 under cloud-hypervisor-cvm in order to address CVE-2024-6119

jiria:
Update openssl to 3.3.2 under cloud-hypervisor-cvm in order to address CVE-2024-6119

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This PR updates `openssl` dependency of `cloud-hypervisor-cvm` to version `3.3.2`, as that is the version that contains a patch for CVE-2024-6119. The patch has been included into the vendored archive and is also part of this change.

###### Change Log  
- Cargo.lock patch of the openssl version and updated vendored archive name

**NO**

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2024-6119

###### Test Methodology

- Local build with RUN_CHECK=y

### DIFF
--- a/SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
+++ b/SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
@@ -5,7 +5,7 @@
 Name:           cloud-hypervisor-cvm
 Summary:        Cloud Hypervisor CVM is an open source Virtual Machine Monitor (VMM) that enables running SEV SNP enabled VMs on top of MSHV using the IGVM file format as payload.
 Version:        38.0.72.2
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        ASL 2.0 OR BSD-3-clause
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
@@ -15,14 +15,22 @@ Source0:        https://github.com/microsoft/cloud-hypervisor/archive/refs/tags/
 %if 0%{?using_vendored_crates}
 # Note: the %%{name}-%%{version}-cargo.tar.gz file contains a cache created by capturing the contents downloaded into $CARGO_HOME.
 # To update the cache and config.toml run:
-#   tar -xf %{name}-%{version}.tar.gz
-#   cd %{name}-%{version}
+#   tar -xf %%{name}-%%{version}.tar.gz
+#   cd %%{name}-%%{version}
+#   patch -u -p0 < ../upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
 #   cargo vendor > config.toml
-#   tar -czf %{name}-%{version}-cargo.tar.gz vendor/
-# rename the tarball to %{name}-%{version}-cargo.tar.gz when updating version
-Source1:        %{name}-%{version}-cargo.tar.gz
+#   tar -czf %%{name}-%%{version}-cargo.tar.gz vendor/
+# rename the tarball to %%{name}-%%{version}-2-cargo.tar.gz when updating version
+# (feel free to drop -2 and this comment on version change)
+Source1:        %{name}-%{version}-2-cargo.tar.gz
 Source2:        config.toml
 %endif
+# Generated using:
+#   tar -xf %%{name}-%%{version}.tar.gz
+#   cd %%{name}-%%{version}
+#   cargo update -p openssl-src --precise 300.3.2+3.3.2
+#   diff -u ../cloud-hypervisor-msft-v38.0.72.2.backup/Cargo.lock Cargo.lock > ../upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
+Patch0:         upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
 
 BuildRequires:  binutils
 BuildRequires:  gcc
@@ -75,6 +83,9 @@ tar xf %{SOURCE1}
 mkdir -p .cargo
 cp %{SOURCE2} .cargo/
 %endif
+# The vendored archive has been populated based on the patch, so we need to
+# repatch here as well in order to use the same versions
+%autopatch -p0
 
 %install
 install -d %{buildroot}%{_bindir}
@@ -136,6 +147,9 @@ cargo build --release --target=%{rust_musl_target} %{cargo_pkg_feature_opts} %{c
 %license LICENSE-BSD-3-Clause
 
 %changelog
+* Tue Sep 17 2024 Jiri Appl <jiria@microsoft.com> - 38.0.72.2-2
+- Patch openssl in the vendored archive to 3.3.2 to address CVE-2024-6119
+
 * Fri Jul 12 2024 Archana Choudhary <archana1@microsoft.com> - 38.0.72.2-1
 - Upgrade to v38.0.72.2
 - Fixes CVE-2023-45853, CVE-2018-25032, CVE-2023-5363, CVE-2023-5678, CVE-2023-6129, CVE-2023-6237, CVE-2024-0727, CVE-2024-4603
--- a/SPECS/cloud-hypervisor-cvm/upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
+++ b/SPECS/cloud-hypervisor-cvm/upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
@@ -1,0 +1,14 @@
+--- ../cloud-hypervisor-msft-v38.0.72.2.backup/Cargo.lock	2024-09-17 12:55:41.269905595 -0700
++++ Cargo.lock	2024-09-17 13:49:15.579003678 -0700
+@@ -1421,9 +1421,9 @@
+ 
+ [[package]]
+ name = "openssl-src"
+-version = "300.3.1+3.3.1"
++version = "300.3.2+3.3.2"
+ source = "registry+https://github.com/rust-lang/crates.io-index"
+-checksum = "7259953d42a81bf137fbbd73bd30a8e1914d6dce43c2b90ed575783a22608b91"
++checksum = "a211a18d945ef7e648cc6e0058f4c548ee46aab922ea203e0d30e966ea23647b"
+ dependencies = [
+  "cc",
+ ]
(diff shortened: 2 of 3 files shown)


SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -17,12 +17,20 @@ Source0:        https://github.com/microsoft/cloud-hypervisor/archive/refs/tags/
...
+# rename the tarball to %{name}-%{version}-2-cargo.tar.gz when updating version
+# (feel free to drop -2 and this comment on version change)
+Source1:        %{name}-%{version}-2-cargo.tar.gz
 Source2:        config.toml
 %endif
+# Generated using:
+#   tar -xf %{name}-%{version}.tar.gz
+#   cd %{name}-%{version}
+#   cargo update -p openssl-src --precise 300.3.2+3.3.2
+#   diff -u ../cloud-hypervisor-msft-v38.0.72.2.backup/Cargo.lock Cargo.lock > ../upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
```
PawelWMS, line 32: Please use `%%` in the comments. Otherwise `rpmbuild` will keep printing a bunch of warnings about macros expanded in comments, which generate noise and tend to confuse people when a build failures needs to be debugged.
```suggestion
#   tar -xf %%{name}-%%{version}.tar.gz
#   cd %%{name}-%%{version}
#   cargo update -p openssl-src --precise 300.3.2+3.3.2
#   diff -u ../cloud-hypervisor-msft-v38.0.72.2.backup/Cargo.lock Cargo.lock > ../upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
```

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -17,12 +17,20 @@ Source0:        https://github.com/microsoft/cloud-hypervisor/archive/refs/tags/
 # To update the cache and config.toml run:
 #   tar -xf %{name}-%{version}.tar.gz
 #   cd %{name}-%{version}
+#   patch -u -p0 < ../upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
```
PawelWMS, line 20: Please apply the patch in the `%prep` section. This helps to keep the flow consistent:
1. Extract unmodified source tarballs.
2. Patch their contents.

Also, this way we don't need to generate and upload a whole new tarball to the blob store.

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -70,6 +78,7 @@ Cloud Hypervisor is an open source Virtual Machine Monitor (VMM) that runs on to
 %prep
 
 %setup -q -n cloud-hypervisor-msft-v%{version}
+%autopatch -p0
```
PawelWMS, line 81: Just making sure I understand: does this mean the patch applies both to the original sources as well as the vendored sources?

jiria: It is my understanding the vendored package only includes the /vendor directory. The patch is against Cargo.lock, so needs to be applied in both cases. And needs to be applied as part of the creating the vendor archive as well, to ensure that the updated version of openssl dependency gets cached.

SPECS/cloud-hypervisor-cvm/upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
```diff
@@ -0,0 +1,14 @@
+--- ../cloud-hypervisor-msft-v38.0.72.2.backup/Cargo.lock	2024-09-17 12:55:41.269905595 -0700
++++ Cargo.lock	2024-09-17 13:49:15.579003678 -0700
+@@ -1421,9 +1421,9 @@
+ 
+ [[package]]
+ name = "openssl-src"
+-version = "300.3.1+3.3.1"
++version = "300.3.2+3.3.2"
```
PawelWMS, line 8: I'm not sure I get this change. What's exactly happening? Is 300.3.2 == 3.3.2 mentioned in the PR's summary? And does this mean `cloud-hypervisor-cvm` is using its own version of OpenSSL instead of the one provided by the `openssl` package (currently on version 3.3.0)?

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
jiria: I dont think we can do that. My understanding is that we need to patch the Cargo.lock before cargo vendor, in order to cache the updated bits of the openssl. 

So I think we should leave it in the comment, upload the new vendor archive. But I will move it from %setup to %prep. 

Do you see any concern with this approach?

SPECS/cloud-hypervisor-cvm/upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
jiria: https://github.com/alexcrichton/openssl-src-rs

So yes, 300.3.2 is 3.3.2. And yes, they carry their own. I recall there was a way to make it link what OS carries instead. Should we file a work item to investigate if this can be moved to using the OS openssl instead of a vendored one? I would rather not deal with that as part of the CVE, but can if you think that is the better approach.

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
jiria: Prep does not work as is, so will move back to setup.

SPECS/cloud-hypervisor-cvm/upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
jiria: Paco looked at it further:

the openssl and openssl-sys crates can use either the system's or a vendored version of openssl. That is selected through cargo features. Looking at the cargo files of openssl, openssl-sys and openssl-src, the latter should only be included when the vendored feature is enabled. SO something pulls it in with that feature enabled.
Cargo.lock
microsoft/cloud-hypervisor
 
Looking at the graph in  https://github.com/microsoft/cloud-hypervisor/blob/msft-main/Cargo.lock... looks like
openssl-src <- openssl-sys <- libssh2-sys <- ssh2 <- test_infra
which takes us here:
https://github.com/microsoft/cloud-hypervisor/blob/msft-main/test_infra/Cargo.toml
and the offender:
ssh2 = { version = "0.9.4", features = ["vendored-openssl"] }
https://github.com/microsoft/cloud-hypervisor/blob/a586445cb3c9f878e8d1a3a76278d34cbacc4d7b/test_infra/Cargo.toml#L15C1-L15C62
 
Cargo.lock
microsoft/cloud-hypervisor
 
interestingly test-infra is only pulled as a dev dependency, so it is possible that you could dispute the CVE

PawelWMS: Discussed offline: out of scope for this PR (we're not introducing this dependency, just updating the version) but we'll need to get this sorted out through a separate work item.

PawelWMS (approved):

SPECS/cloud-hypervisor-cvm/upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
jiria: jirka@jiria-linux:~/repos/AzureLinux/SPECS/cloud-hypervisor-cvm/cloud-hypervisor-msft-v38.0.72.2$ cargo tree -i openssl-src
  Downloaded openssl-src v300.3.1+3.3.1
  Downloaded 1 crate (9.8 MB) in 1.02s
openssl-src v300.3.1+3.3.1
[build-dependencies]
└── openssl-sys v0.9.99
    └── libssh2-sys v0.3.0
        └── ssh2 v0.9.4
            └── test_infra v0.1.0 (/home/jirka/repos/AzureLinux/SPECS/cloud-hypervisor-cvm/cloud-hypervisor-msft-v38.0.72.2/test_infra)
                [dev-dependencies]
                └── cloud-hypervisor v38.0.0 (/home/jirka/repos/AzureLinux/SPECS/cloud-hypervisor-cvm/cloud-hypervisor-msft-v38.0.72.2)

SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.spec
```diff
@@ -15,14 +15,22 @@ Source0:        https://github.com/microsoft/cloud-hypervisor/archive/refs/tags/
...
+# (feel free to drop -2 and this comment on version change)
+Source1:        %{name}-%{version}-2-cargo.tar.gz
 Source2:        config.toml
 %endif
+# Generated using:
+#   tar -xf %%{name}-%%{version}.tar.gz
+#   cd %%{name}-%%{version}
+#   cargo update -p openssl-src --precise 300.3.2+3.3.2
+#   diff -u ../cloud-hypervisor-msft-v38.0.72.2.backup/Cargo.lock Cargo.lock > ../upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
+Patch0:         upgrade-openssl-to-3.3.2-to-address-CVE-2024-6119.patch
```
jslobodzian, line 33: In theory we should be able to introduce openssh-devel and the rust build should use the system version of openssh and break the dependency on openssl.  I *think* and I think that might be a better solution.

jiria: Is that something that you want to do as part of high sev CVE as opposed to merging as is and addressing separately?

jslobodzian: I am ok taking it this way, but nobody will follow up on that until the next CVE issue.  If my suggestion is correct, then it will be essentially fix this issue permanently...and I think that would be a better solution.

ms-mahuber:
Is this not also relevant for the main branch? https://github.com/microsoft/azurelinux/blob/main/SPECS/cloud-hypervisor-cvm/cloud-hypervisor-cvm.signatures.json

ms-mahuber (approved):

jslobodzian (approved):

PR #10480, opened 2024-09-18
Fix bad interactions between timeouts and build retires

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When we queue a package to build (or test), we set a timeout (by default 8h). If the build has not finished by then we forcibly stop the build and mark it as failed.

We also support `PACKAGE_BUILD_RETRIES` and `CHECK_BUILD_RETRIES`, which will cause failed builds to re-run.

However, each time the retry was triggered the timeout would reset. For example in the buddy builds this means that a stuck package test could take 4x8=32h to build, which would exceed pipeline time limits. We want to exit gracefully with an error state so that we can generate and publish logs correctly. If the pipeline forces the timeout, it can be difficult to debug.

Instead of resetting the timeout with each retry, have all attempts share a single timeout. If the timeout is exceeded stop retrying (use `RunWithLinearBackoff()` which will take a `ctx` configured with a timeout, so we can break out early).

As part of this fix, I also noticed that the timeout handling was not cleaning up the build chroot correctly. We should not be using anything related to `panic()` for error handling, instead use `logger.Log.Fatal*()` which gives the logging library a chance to run its registered cleanup functions (ie final chroot cleanup) before exiting "gracefully".

###### Change Log  

- Package build timeout shared by all retry attempts, each invocation of `BuildAgent.BuildPacakge()` now takes a `time.Duration` instead of using the value from `BuildAgentConfig`.
- Properly clean up build chroot on timeout
  - Handle timeout logic inside the chroot.Run so we correctly exit the chroot before leaving the function, otherwise the chroot cleanup code will run from within the chroot itself and the paths will be wrong.
  - Add a new `StopAllChildProcesses()` which is like `PermanentlyStopAllChildProcesses()` but does not set the disable flag (so we can run the gpg-agent cleanup still on exit).

**NO**

###### Associated issues  

- #xxxx

###### Test Methodology

(Added custom %check to words with `sleep 9h`)
- https://dev.azure.com/mariner-org/mariner/_build/results?buildId=642266&view=results

### DIFF
--- a/toolkit/tools/pkgworker/pkgworker.go
+++ b/toolkit/tools/pkgworker/pkgworker.go
@@ -215,6 +215,7 @@ func buildSRPMInChroot(chrootDir, rpmDirPath, toolchainDirPath, workerTar, srpmF
 
 	err = chroot.Initialize(workerTar, extraDirs, mountPoints, true)
 	if err != nil {
+		err = fmt.Errorf("failed to initialize chroot:\n%w", err)
 		return
 	}
 	defer chroot.Close(noCleanup)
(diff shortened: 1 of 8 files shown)


toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -238,6 +239,7 @@ func buildSRPMInChroot(chrootDir, rpmDirPath, toolchainDirPath, workerTar, srpmF
 	case <-time.After(timeout):
 		logger.Log.Errorf("Timeout after %v: killing all processes in chroot...", timeout)
 		shell.PermanentlyStopAllChildProcesses(unix.SIGKILL)
+		time.Sleep(5 * time.Second) // If we don't wait, the chroot mounts may still be busy
```
cwize1, line 242: This is a very unreliable way of doing this. Please try to find an alternative signal to wait on.

cwize1 (changes requested):

toolkit/tools/pkgworker/pkgworker.go
dmcilvaney: This is an edge-case of an edge-case type thing, but I'll see if there is a simple way to handle this better.

I'm going to say we just keep this hack if there isn't an easy fix. This failure case is very uncommon, and worst case just leaves a folder in place on the system.

I'm going to look at the gpg key service handing code, it might be failing since we call `PermanentlyStopAllChildProcesses()` and this code wasn't considered when we added the gpg stuff since its so rare.

dmcilvaney: Ok, figured out the underlying issue here.

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -214,34 +215,39 @@ func buildSRPMInChroot(chrootDir, rpmDirPath, toolchainDirPath, workerTar, srpmF
...
-		shell.PermanentlyStopAllChildProcesses(unix.SIGKILL)
-		err = fmt.Errorf("build timed out after %s", timeout)
-	}
+	err = chroot.Run(func() (err error) {
+		go func() {
+			results <- buildRPMFromSRPMInChroot(srpmFileInChroot, outArch, runCheck, defines, packagesToInstall, isCCacheEnabled(ccacheManager))
+		}()
+
+		var chrootErr error = nil
+		select {
```
cwize1, line 238: Why was this moved under `chroot.Run`?

dmcilvaney: Added a quick blurb to the description, but more details:

So, we rely on `chroot.Run()` to both setup and tear down the chroot (ie restore the root dir). If we do the timeout outside the `chroot.Run()`, we will exit the function and begin teardown immediately. Unfortunately... we won't have restored the correct root yet, so all the cleanup code will be running relative to the chroot's root dir.

This means we will try to unmount stuff, but the path doesn't actually exist, so it fails.

cwize1: It would be good to add a comment in the code.

cwize1 (approved):

toolkit/tools/pkgworker/pkgworker.go
dmcilvaney: Done

cwize1 (approved):

PR #10518, opened 2024-09-20
OSModifier: Read root device from grub.cfg

elainezhao96:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

In trident e2e pipeline there are deployments not using verity test images, and in those images, root device path is not /dev/mapper/root, thus it cause the failure: `/dev/mapper/root does not exist`. It is incorrect for EMU to always set root device to /dev/mapper/root, and it should read it from grub.cfg,  for example, root=UUID=2303cd9a-ab00-4e39-9b5a-8bf8da5a6118, we need to pass the real root device either via UUID= or PARTUUID=. With this change, EMU should be able to handle all test images in that pipeline. 

**NO**

###### Test Methodology

- tested locally and will publish new EMU and rerun trident pipeline

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
@@ -212,10 +212,13 @@ func (b *BootCustomizer) WriteToFile(imageChroot safechroot.ChrootInterface) err
 	return nil
 }
 
-func (b *BootCustomizer) GetDefaultGrubFileContent() string {
-	return b.defaultGrubFileContent
-}
+func (b *BootCustomizer) SetRootDevice(rootDevice string) error {
+	updatedGrubFileContent, err := UpdateDefaultGrubFileVariable(b.defaultGrubFileContent, "GRUB_DEVICE", rootDevice)
+	if err != nil {
+		return err
+	}
 
-func (b *BootCustomizer) SetDefaultGrubFileContent(content string) {
-	b.defaultGrubFileContent = content
+	b.defaultGrubFileContent = updatedGrubFileContent
+
+	return nil
 }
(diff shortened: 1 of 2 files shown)


toolkit/tools/pkg/osmodifierlib/modifydefaultgrub.go
```diff
@@ -26,7 +27,7 @@ var grubArgs = []string{
 func modifyDefaultGrub() error {
 	var dummyChroot safechroot.ChrootInterface = &safechroot.DummyChroot{}
 	// Get verity, selinux, overlayfs, and root device values from /boot/grub2/grub.cfg
-	values, err := extractValuesFromGrubConfig(dummyChroot)
+	values, rootDevice, err := extractValuesFromGrubConfig(dummyChroot)
```
fintelia, line 30: I'm concerned that this approach will prevent Trident from having any control over what root device is set in the kernel command line. I think we should consider having Trident pass the desired root device to EMU instead

elainezhao96: I see. Created task https://dev.azure.com/mariner-org/ECF/_workitems/edit/9150 for this and I think I will make it a follow up PR, for now I'd like to first confirm the pipeline works with this change

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -211,3 +211,11 @@ func (b *BootCustomizer) WriteToFile(imageChroot safechroot.ChrootInterface) err
 
 	return nil
 }
+
+func (b *BootCustomizer) GetDefaultGrubFileContent() string {
```
cwize1, line 215: The `GetDefaultGrubFileContent` and `SetDefaultGrubFileContent` functions are breaking the abstraction of `BootCustomizer`. Add a `SetRootDevice` function instead.

cwize1 (changes requested):

cwize1 (approved):

PR #10584, opened 2024-09-30
OSModifier: Extend EMU API to update verity and root device

elainezhao1:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Add verity, rootdevice to EMU API (already integrated [here](https://dev.azure.com/mariner-org/ECF/_git/trident?path=/src/engine/boot/grub.rs&version=GBmain&line=220&lineEnd=220&lineStartColumn=9&lineEndColumn=46&lineStyle=plain&_a=contents)
- Due to the bug https://dev.azure.com/mariner-org/ECF/_workitems/edit/9648/ , use enforcing=0 for permissive mode for EMU

**NO**

###### Test Methodology

-  Already published and tested in trident pipeline,


toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
```diff
@@ -591,6 +591,28 @@ func selinuxModeToArgs(selinuxMode imagecustomizerapi.SELinuxMode) ([]string, er
...
+	switch selinuxMode {
+	case imagecustomizerapi.SELinuxModeDisabled:
+		newSELinuxArgs = []string{installutils.CmdlineSELinuxDisabledArg}
+
+	case imagecustomizerapi.SELinuxModeForceEnforcing:
+		newSELinuxArgs = []string{installutils.CmdlineSELinuxSecurityArg, installutils.CmdlineSELinuxEnabledArg,
+			installutils.CmdlineSELinuxEnforcingArg}
+
+	case imagecustomizerapi.SELinuxModePermissive, imagecustomizerapi.SELinuxModeEnforcing:
+		newSELinuxArgs = []string{installutils.CmdlineSELinuxSecurityArg, installutils.CmdlineSELinuxEnabledArg,
```
cwize1, line 606: Isn't this forcing the SELinux mode to be permissive? So, I don't think this is correct settings for `SELinuxModeEnforcing`.

toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
```diff
@@ -591,6 +591,28 @@ func selinuxModeToArgs(selinuxMode imagecustomizerapi.SELinuxMode) ([]string, er
...
+	newSELinuxArgs := []string(nil)
+	switch selinuxMode {
+	case imagecustomizerapi.SELinuxModeDisabled:
+		newSELinuxArgs = []string{installutils.CmdlineSELinuxDisabledArg}
+
+	case imagecustomizerapi.SELinuxModeForceEnforcing:
+		newSELinuxArgs = []string{installutils.CmdlineSELinuxSecurityArg, installutils.CmdlineSELinuxEnabledArg,
+			installutils.CmdlineSELinuxEnforcingArg}
+
+	case imagecustomizerapi.SELinuxModePermissive, imagecustomizerapi.SELinuxModeEnforcing:
```
cwize1, line 605: Would it make sense to create a new mode called `SELinuxModeForcePermissive` to MIC? That way, we can have a consistent API between MIC and EMU.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -55,6 +55,9 @@ const (
 	// CmdlineSELinuxEnforcingArg is the arg required for forcing SELinux to be in enforcing mode.
 	CmdlineSELinuxEnforcingArg = "enforcing=1"
 
+	// CmdlineSELinuxPermissiveArg is the arg for SELinux to be in force-permissive mode.
+	CmdlineSELinuxPermissiveArg = "enforcing=0"
```
cwize1, line 59: Since this change is outside the MIC directories, it will need to be approved by shiproom. I recommend making the change that shiproom needs to review as small as possible. So, it might be good to handle the SELinux stuff as a separate PR.

toolkit/tools/osmodifierapi/os.go
```diff
@@ -13,10 +13,12 @@ import (
...
 type OS struct {
-	Hostname string                     `yaml:"hostname"`
-	SELinux  imagecustomizerapi.SELinux `yaml:"selinux"`
-	Users    []imagecustomizerapi.User  `yaml:"users"`
-	Overlays *[]Overlay                 `yaml:"overlays"`
+	Hostname   string                     `yaml:"hostname"`
+	SELinux    imagecustomizerapi.SELinux `yaml:"selinux"`
+	Users      []imagecustomizerapi.User  `yaml:"users"`
+	Overlays   *[]Overlay                 `yaml:"overlays"`
+	Verity     *imagecustomizerapi.Verity `yaml:"verity"`
```
cwize1, line 20: The way you are using the fields under `imagecustomizerapi.Verity` is very different from how they are used in MIC. Therefore, you should create a new verity API type that is designed for EMU.

toolkit/tools/pkg/osmodifierlib/modifierutils.go
```diff
@@ -61,6 +61,67 @@ func doModifications(baseConfigPath string, osConfig *osmodifierapi.OS) error {
 		}
 	}
 
+	if osConfig.Verity != nil {
+
+		bootCustomizer, err := imagecustomizerlib.NewBootCustomizer(dummyChroot)
```
cwize1, line 66: The `NewBootCustomizer` and `WriteToFile` only need to be called once and can be shared with both `Verity` and `RootDevice`.

toolkit/tools/pkg/osmodifierlib/modifierutils.go
```diff
@@ -61,6 +61,67 @@ func doModifications(baseConfigPath string, osConfig *osmodifierapi.OS) error {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func updateDefaultGrubForVerity(verity *imagecustomizerapi.Verity, bootCustomizer *imagecustomizerlib.BootCustomizer) error {
+
```
cwize1, line 104: nit: Don't add empty lines after a `{`.

Similar comment elsewhere.

cwize1 (changes requested):

toolkit/tools/pkg/osmodifierlib/modifierutils.go
elainezhao96: addressed in pr in new repo

toolkit/tools/pkg/osmodifierlib/modifierutils.go
elainezhao96: updated in pr in new repo

toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
elainezhao96: I'm not sure if it makes sense to add a force-permissive mode in MIC, the trident will pass value `permissive` not `force-permissive` to EMU. The difference between EMU' permissive and MIC's permissive is EMU needs to add enforicing=0 and not modify etc/selinux/config

toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
elainezhao96: fixed it the pr in new repo

toolkit/tools/osmodifierapi/os.go
elainezhao96: fixed in the pr in new repo

PR #10608, opened 2024-10-02
ImageCustomizer - Service and Overlay recommendations for Verity-enabled images.

liulanze:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Service and Overlay recommendations for Verity-enabled images.
Clear services down errors when booting Verity enabled images.

**NO**

###### Test Methodology

- Local build


toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -550,6 +546,23 @@ os:
     corruptionOption: panic
 ```
 
+When enabling the Verity feature, it is recommended that overlays are applied to
```
cwize1, line 549: I suspect that over time the recommendations for verity images could end up being quite long. There are lots of different services/programs available on Azure Linux and a number of them may need special handling under verity. 

In addition, there may be multiple options for a service/program with different tradeoffs. Some users may want a service/program configuration option changeable for convenience. While other users may want it immutable for security.

As such, I recommend splitting off the verity recommendations into its own doc.

cwize1 (changes requested):

toolkit/tools/imagecustomizer/docs/configuration.md
romoh: Sounds reasonable to me.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -550,6 +546,27 @@ os:
     corruptionOption: panic
 ```
 
+The Verity-enabled root filesystem is always mounted as read-only. Its root hash
+and hash tree are computed at build time and verified by systemd during the
+initramfs phase on each boot. When enabling the Verity feature, it is
+recommended to create a writable persistent partition (e.g., /var) for any
+directories that require write access. Critical files and directories, such as
+SSH host keys or logs, can be redirected to the writable partition using
+symlinks or similar methods. See [Verity Filesystem Layout
```
romoh, line 555: why not call out `overlays` here as well?

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -550,6 +546,27 @@ os:
...
+Recommendations](./verity-filesystem-layout-recommendations.md) for details.
```
romoh, line 556: nit - too long, how about `verity.md`?

romoh (approved):
Left some suggestions

toolkit/tools/imagecustomizer/docs/configuration.md
liulanze: Fixed, thanks.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -511,10 +511,6 @@ Example: `noatime,nodiratime`
 Specifies the configuration for dm-verity root integrity verification. Please
 execute `sudo modprobe nbd` before building the image with verity enablement.
```
cwize1, line 512: While you're in the area, it might be goodness to remove the reference to `nbd`.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -550,6 +546,27 @@ os:
     corruptionOption: panic
 ```
 
+The Verity-enabled root filesystem is always mounted as read-only. Its root hash
```
cwize1, line 549: I would move all of this prose to the `verity.md` file. And then in this file, only say something like: `There are multiple ways to configure a verity enabled image. For recommendations see, [Verity Image Recommendations](./verity.md)`.

toolkit/tools/imagecustomizer/docs/verity.md
```diff
@@ -0,0 +1,16 @@
+# Verity Filesystem Layout Recommendations
```
cwize1, line 1: I would probably title this document 'Verity Image Recommendations`.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -550,6 +546,27 @@ os:
...
+symlinks or similar methods. See [Verity Filesystem Layout
+Recommendations](./verity.md) for details.
+
+Please note that some services and programs on Azure Linux may require specific
+handling when using Verity. Depending on user needs, there are different
+configuration options that offer tradeoffs between convenience and security.
+Some configurations can be made flexible to allow changes, while others may be
+set as immutable for enhanced security. 
+
+For example, if your image includes the `systemd-growfs-root.service`, which
```
cwize1, line 564: Instead of using prose, let's try to make the recommendations more structured. Also, including example configs would be goodness.

For example:

```Markdown
## systemd-growfs-root

This service attempts to resize the root filesystem, which fails since verity makes the root filesystem readonly and a fixed size.

### Solution 1: Do nothing

Since the root filesystem is readonly, the `systemd-growfs-root` service will fail.
However, the only impact will be an error in the boot logs.

### Solution 2: Disable service

Disabling the service removes the error from the boot logs.

`` `yaml
os:
  services:
    disable:
    - systemd-growfs-root
`` `
```

Do something similar for other services (e.g. cloud-init, ssh, auditd, docker, etc.).

cwize1 (changes requested):

toolkit/tools/imagecustomizer/docs/configuration.md
romoh: My interpretation of Chris' comment was that it shouldn't be the sole recommendation but IMO we should include it as one possible option.

liulanze: It makes sense to me.

toolkit/tools/imagecustomizer/docs/verity.md
```diff
@@ -0,0 +1,141 @@
...
+require write access. Critical files and directories can be redirected to the
+writable partition using symlinks or similar methods.
+
+Please also note that some services and programs on Azure Linux may require
+specific handling when using Verity. Depending on user needs, there are
+different configuration options that offer tradeoffs between convenience and
+security. Some configurations can be made flexible to allow changes, while
+others may be set as immutable for enhanced security.
+
+## systemd-growfs-root
```
cwize1, line 16: It might be goodness to create a dedicated section for the var partition, since so many partitions need it to be writable. That way, you can avoid sections pointing to the `cloud-init` section.

toolkit/tools/imagecustomizer/docs/verity.md
```diff
@@ -0,0 +1,141 @@
...
+Disabling the service removes the error from the boot logs.
+
+```yaml
+os:
+  services:
+    disable:
+    - systemd-growfs-root
+```
+
+## cloud-init-local, cloud-config, cloud-final
```
cwize1, line 37: nit: Just `## cloud-init`

While there are multiple systemd services included with cloud-init, they are all part of the same application.

cwize1, line 37: `cloud-init` is an odd one. cloud-init has a bunch of different features that can be used to configure the system. (e.g. user accounts, networking, etc.) But a lot of those features typically require the /etc directory to be writable.

I think the general recommendation should be to disable cloud-init and not use it.

Though it is possible that some cloud-init features can be used under verity. For those, we may end up giving them dedicated sections in this doc.

toolkit/tools/imagecustomizer/docs/verity.md
```diff
@@ -0,0 +1,141 @@
...
+filesystem.
+
+Refer to the YAML layout under the
+[cloud-init](#cloud-init-local-cloud-config-cloud-final) section for the
+partition configuration.
+
+The following example script can be added to the `postCustomization` scripts to
+automatically move the SSH host keys and update the `sshd_config` file:
+
+```bash
```
cwize1, line 132: You should provide a sample config, not just the script.

Note: Scripts can now be specified inline using `content` instead of `path`.

cwize1, line 132: The `sshd-keygen.service` file also need to be updated.

Note: Additional files now support `content` instead of `source`.

toolkit/tools/imagecustomizer/docs/verity.md
cwize1, line 16: You should add a section for networking.

In Mariner 3.0, there are no longer any default network settings. In non-verity images, this doesn't matter too much since cloud-init will provide default networking settings. But cloud-init fails to provision the network in verity images since `/etc` is not writable. Hence, it is a good idea for verity images to specify network settings. The [test verity config](https://github.com/microsoft/azurelinux/blob/3.0/toolkit/tools/pkg/imagecustomizerlib/testdata/verity-config.yaml#L78) does this.

cwize1 (changes requested):

toolkit/tools/imagecustomizer/docs/verity.md
liulanze: Fixed. Thanks.

liulanze: Added, please review.

toolkit/tools/imagecustomizer/docs/verity.md
liulanze: Fixed. Thanks.

liulanze: I have re-write the cloud-init part, however, the cloud-init services I faced so far (cloud-init-local, cloud-config, cloud-final), they all need write access to `/etc`, so I think for the section of cloud-init features that can be used under verity, the doc can be updated when users meet them.

toolkit/tools/imagecustomizer/docs/verity.md
liulanze: Fixed, thanks.

liulanze: Fixed, thanks.

cwize1 (approved):

PR #10657, opened 2024-10-08
Fix `pytest` so that it's actually usable.

surfacepatterns:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

When I attempted to build and install packages updated in [this PR](https://github.com/microsoft/azurelinux/pull/10347), I noticed that, while the packages built successfully, tests would fail in packages that had `pytest` fixtures.  I investigated the issue, and noticed that the current `pytest` package in Azure Linux simply doesn't work:
```
[danderson: ~]
> python3
Python 3.12.3 (main, Aug 24 2024, 02:10:35) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pytest
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/site-packages/pytest/__init__.py", line 6, in <module>
    from _pytest.assertion import register_assert_rewrite
  File "/usr/lib/python3.12/site-packages/_pytest/assertion/__init__.py", line 9, in <module>
    from _pytest.assertion import rewrite
  File "/usr/lib/python3.12/site-packages/_pytest/assertion/rewrite.py", line 34, in <module>
    from _pytest.assertion import util
  File "/usr/lib/python3.12/site-packages/_pytest/assertion/util.py", line 20, in <module>
    from _pytest.config import Config
  File "/usr/lib/python3.12/site-packages/_pytest/config/__init__.py", line 47, in <module>
    from .findpaths import determine_setup
  File "/usr/lib/python3.12/site-packages/_pytest/config/findpaths.py", line 13, in <module>
    import iniconfig
ModuleNotFoundError: No module named 'iniconfig'
```
This PR adds a new `python-iniconfig` package, and makes `pytest` depend on that package.  When attempting to build and install the packages from the aforementioned PR with the `pytest` changes in this PR, the tests run successfully.

@microsoft/cbl-mariner-devs: Note that an Azure Linux developer will need to upload [the source archive for `python-iniconfig`](https://files.pythonhosted.org/packages/23/a2/97899f6bd0e873fed3a7e67ae8d3a08b21799430fb4da15cfedf10d6e2c2/iniconfig-1.1.1.tar.gz) to the source server.

###### Change Log  

- Add `python-iniconfig` package, a runtime dependency of `pytest`.
- Update `pytest` spec to explicitly depend on `python-iniconfig`.
- Update Azure Linux license and manifest info.

NO

###### Test Methodology

- local build


surfacepatterns:
AFAICT, the source signature check failure is wrong.  The tarball signature is correct:
```
[danderson: ~/projects/azurelinux/SPECS/python-iniconfig]
> python3
Python 3.12.3 (main, Aug 24 2024, 02:10:35) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from hashlib import sha256, file_digest
>>> fp = open("iniconfig-1.1.1.tar.gz", "rb")
>>> d = file_digest(fp, sha256).digest()
>>> fp.close()
>>> d.hex()
'bc3af051d7d14b2ee5ef9969666def0cd1a000e121eaea580d4a313df4b37f32'
```
I'll figure out the circular dependency check failure after I sleep some. :)

surfacepatterns:
@microsoft/cbl-mariner-devs: This PR is ready for review.

As mentioned above, I believe the source signature that the `Source Signature Check` is complaining about is correct.  I'm guessing the source server has a different tarball associated with `iniconfig-1.1.1`, even though the package isn't currently in Azure Linux.  Perhaps someone else tried adding such a package at some point.

reubeno:
Thanks for working on this! 

The existing published source tarball looks more like a full git archive, rather than a copy of the distributable archive published to pypi/pythonhosted.

One question -- I see that `iniconfig` 2.0 was released at the beginning of 2023. Is there a specific reason you were recommending going with `1.1.1` instead?

surfacepatterns:
> Thanks for working on this!

I'm happy to help. :)

> The existing published source tarball looks more like a full git archive, rather than a copy of the distributable archive published to pypi/pythonhosted.

Interesting.  I don't know the Azure Linux policy on such things, but my assumption is that having a static tarball that __should__ always yield the same signature when fetched from the URL in the spec is preferable to having a dynamically generated `git` snapshot.

> One question -- I see that `iniconfig` 2.0 was released at the beginning of 2023. Is there a specific reason you were recommending going with `1.1.1` instead?

Two reasons:
1. The upstream Fedora version is 1.1.1, and re-using their spec is easy.
2. The 2.0 tarball hosted on pypi doesn't include tests for whatever reason.  Even though tests aren't run in the default build due to a circular dependency with `pytest`, it's possible to run tests if you already have a working `pytest` package and supply the `tests` flag when building the `iniparse` package.  I figured that having a build with tests is preferable, even if tests aren't run be default.

jslobodzian (approved):

reubeno:
@surfacepatterns Thanks for your patience as we were sorting out the source archives on the backend. We're working on uploading the version of the archive that you reference in your PR (with sha256 hash `bc3af051d7d14b2ee5ef9969666def0cd1a000e121eaea580d4a313df4b37f32`).

Once that goes through, I'll kick off some additional build validation and report back. In addition to ensuring successful build/test of the affected components, I'd also like to make sure that some of the other packages currently `pip`-installing `iniconfig` will still continue to pass their tests. Once this change is in, we can go back and remove those workarounds.

reubeno:
The validation build passed, and confirms this resolves the original issue.

PR #10706, opened 2024-10-11
Enable Dracut's livenet rootfs handling when systemd-networkd is in use.

gmileka:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Dracut provides a livenet module which allows the downloading of a liveos ISO image from a PXE server at boot time. However, this functionality is not implemented for systemd-networkd in Dracut 102 and earlier (see [bug](https://github.com/dracut-ng/dracut-ng/issues/719) and [fix](https://github.com/dracut-ng/dracut-ng/pull/388)).

It is desired to produce Azure Linux liveos ISO images that are PXE-ready. To enable this scenario, we need to ensure that Dracut's livenet module is functional on Azure Linux images.

There are two options:
1. Upgrade dracut version to 103+.
2. Patch the current Dracut version.

Given that Azure Linux has GA'ed, and given that Dracut introduced refactoring around the networking supported between 102 and 103, we opted to minimize the risk of upgrading at this time and go with patching the current Dracut version.

This change augments Dracut's livenet module with a daemon to be started when the network is up. The new daemon will then call the livenet module download script.

###### Regression Risk

Minimal.

To avoid side effects or conflicts with existing user configuration, this flow is only activated if the new `rd.live.azldownloader` kernel parameter is present and set to 'enabled' - otherwise, it is not activated.

###### Change Log  

- [Update dracut livenet module with an iso downloader script to enable PXE booting for liveos ISOs.](https://github.com/microsoft/azurelinux/pull/10706/commits/b22bb77a8c3e094e9711878e724ac4a27de116bb)

- **no**

###### Associated issues  

- 9741

###### Links to CVEs  
- n/a

###### Test Methodology

Unit Tests
- Ran MIC liveos ISO tests (tests include coverage for new scenario).

Functional Tests
- Deploy a PXE server on the network.
- Use MIC to generate a liveos iso (vhdx->iso) and specify the new `--output-pxe-artifacts-dir` parameter.
  - Attach the iso to a hyper-v VM
    - Verify the VM boots, and the user can log-in.
  - Deploy the iso artifacts to a PXE server and boot a PXE client on the same network.
    - Verify the PXE client boots, and that the user can log-in.
- Use MIC to generate a liveos iso (iso->iso) and specify the new `--output-pxe-artifacts-dir` parameter.
  - Attach the iso to a hyper-v VM
    - Verify the VM boots, and the user can log-in.
  - Deploy the iso artifacts to a PXE server and boot a PXE client on the same network.
    - Verify the PXE client boots, and that the user can log-in.
    - Verify user settings from the first run (vhdx-iso) have been retained.

### DIFF
--- a/SPECS/dracut/90livenet/liveos-artifacts-download.service
+++ b/SPECS/dracut/90livenet/liveos-artifacts-download.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=LiveOS Artifacts Download Daemon
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+ExecStart=/sbin/liveos-artifacts-download
+Type=oneshot
+
+[Install]
+WantedBy=multi-user.target
--- a/SPECS/dracut/add-livenet-download-service.patch
+++ b/SPECS/dracut/add-livenet-download-service.patch
@@ -1,0 +1,56 @@
+From 3055f6f900c5718d93bb68a4cdf0798ac0f55c14 Mon Sep 17 00:00:00 2001
+From: George Mileka <gmileka@microsoft.com>
+Date: Fri, 30 Aug 2024 18:17:00 -0700
+Subject: [PATCH] Enable netroot for systemd-networkd - temporary fix till
+ upgrading to 103+.
+
+Dracut version 102 does not implement the code path that handles livenet
+rootfs download (i.e. by calling /sbin/livenetroot) when systemd-networkd
+is the underlying networking manager.
+
+This has been implemented in 103 (see https://github.com/dracut-ng/dracut-ng/pull/388)
+
+As a mitigation for 102, this patch allows Dracut's livenet module to schedule
+a daemon to run after the network stack is up and running. The daemon will
+then invoke the usual livenet rootfs handling code to download and pivot to
+the new rootfs.
+---
+ modules.d/90livenet/module-setup.sh  | 2 ++
+ modules.d/90livenet/parse-livenet.sh | 8 ++++++++
+ 2 files changed, 10 insertions(+)
+
+diff --git a/modules.d/90livenet/module-setup.sh b/modules.d/90livenet/module-setup.sh
+index db0def50..74bacd6c 100755
+--- a/modules.d/90livenet/module-setup.sh
++++ b/modules.d/90livenet/module-setup.sh
+@@ -17,6 +17,8 @@ install() {
+     inst_hook cmdline 29 "$moddir/parse-livenet.sh"
+     inst_hook initqueue/online 95 "$moddir/fetch-liveupdate.sh"
+     inst_script "$moddir/livenetroot.sh" "/sbin/livenetroot"
++    inst_simple "$moddir/liveos-artifacts-download.service" "/etc/systemd/system/liveos-artifacts-download.service"
++    inst_script "$moddir/liveos-artifacts-download.sh" "/sbin/liveos-artifacts-download"
+     if dracut_module_included "systemd-initrd"; then
+         inst_script "$moddir/livenet-generator.sh" "$systemdutildir"/system-generators/dracut-livenet-generator
+     fi
+diff --git a/modules.d/90livenet/parse-livenet.sh b/modules.d/90livenet/parse-livenet.sh
+index a1d14a8f..996f5210 100755
+--- a/modules.d/90livenet/parse-livenet.sh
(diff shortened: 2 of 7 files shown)


dmcilvaney (approved):

SPECS/dracut/add-livenet-download-service.patch
```diff
@@ -0,0 +1,56 @@
...
+@@ -27,6 +27,14 @@ if get_url_handler "$liveurl" > /dev/null; then
+     root="livenet" # quiet complaints from init
+     # shellcheck disable=SC2034
+     rootok=1
++
++    enableAzureLinuxDownloader=$(getarg rd.live.azldownloader=)
++
++    if [[ "$enableAzureLinuxDownloader" == "enable" ]]; then
++        systemctl enable liveos-artifacts-download
++        systemctl start liveos-artifacts-download &
```
cwize1, line 48: Why are you running the download in a systemd service? Why not run `liveos-artifacts-download.sh` inline?

Camelron (approved):

SPECS/dracut/90livenet/liveos-artifacts-download.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=LiveOS Artifacts Download Daemon
+After=network-online.target
```
cwize1, line 3: If the service is started manually using `systemctl start` then there isn't much point to adding `After` and `Wants`.

SPECS/dracut/add-livenet-download-service.patch
cwize1, line 48: You typically shouldn't manually start systemd services. Instead, you should add a `WantedBy` to the `.service` file and then enable the service. Then the service will be scheduled automatically by systemd.

To enable a service, in the `module-setup.sh` file run `$SYSTEMCTL -q --root "$initdir" enable <name>.service`

For the `WantedBy`, you'll want to check the [dracut.bootup](https://man7.org/linux/man-pages/man7/dracut.bootup.7.html). But I think you may want `WantedBy=sysinit.target`.

In terms of `getarg rd.live.azldownloader=`, you can probably check that in `liveos-artifacts-download.sh` and return early if it isn't present.

gmileka: This script, parse-livenet.sh, is invoked before udev and network bring-up, so, at this point, we can't invoke the download script. However, we can schedule it such that it's gated on network daemon up and running.

SPECS/dracut/90livenet/liveos-artifacts-download.service
gmileka: We start it early before udev and the network stack is up, and then it'll block until 'After' `network-online.target`.

SPECS/dracut/90livenet/liveos-artifacts-download.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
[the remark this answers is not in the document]
gmileka: pending.

SPECS/dracut/add-livenet-download-service.patch
gmileka: We actually don't want the service enabled all the time at initrd generation time. Instead, we want it enabled only when it's a live netroot setup - hence, the enablement in the parse file and not in the module-setup.sh.

The problem with always having it enabled is that the user could have a network manager other than systemd-networkd enabled, and then specifies `root=live:http:...`. This will result in triggering both handlers (livenet and ours) at the same time if our new daemon is always enabled.

For the condition, we can't move to the script because we need it at parse time to enable the daemon.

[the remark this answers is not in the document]
gmileka: That works too. I'll switch to `--no-block` because it is more explicit.

SPECS/dracut/90livenet/liveos-artifacts-download.service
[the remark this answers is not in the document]
gmileka: updated.

SPECS/dracut/add-livenet-download-service.patch
cwize1: Fair enough. But it might be good to add a comment.

cwize1 (approved):

romoh (approved):

PR #10723, opened 2024-10-14
add SymCrypt-debuginfo package

tobiasb-ms:
- [x] Ready to merge

---

###### Summary 
Creates `SymCrypt-debuginfo` package.

The normal process for creating `debuginfo` doesn't quite work for `SymCrypt`, so we had to be a little creative. There are two fundamental issues.

1. It requires that the stripped binary retain certain symbols. In its normal build, `SymCrypt` uses `strip --keep-symbol` to do that, but the `find-debuginfo` script that strips and generates files for `debuginfo` packages uses `eu-strip` which doesn't have an equivalent feature.
2. It does post-processing on the stripped binary which, due to FIPS self-tests must be done after any modification of the binary.

To work around this, we:
1. Apply a couple patches to `SymCrypt`'s build files. These allow us to skip the stripping and post-processing the binary, along with parameterizing where the post-process script stores temporary files. Both of these are in discussion for upstream acceptance.
2. Copied and modified the `find-debuginfo` script to use `strip` instead of `eu-strip` and provide a `--keep-symbol` flag, then had `SymCrypt.spec` use that.

###### Change Log  
- Created `SymCrypt-debuginfo` package; see general notes for details.

**NO**

###### Test Methodology
- Built and tested in local container.
- Tested RPMs from container in local hyper-v image.
- Buddy Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=657346&view=results
- Big Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=657345&view=results


SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -40,26 +42,47 @@ A core cryptographic library written by Microsoft
...
+# Override the default to allow us to do custom fips post-processing after debug info/stripping is done.
+# The post-processing script writes the modified file to the same location as the original file, which
+# is subject to default permissions, so we need to set permissions manually after the script.
+%define __spec_install_post \
+    %{?__debug_package:%{__debug_install_post}} \
+    %{__arch_install_post} \
+    %{__os_install_post} \
+    mkdir -p "bin/module/generic/processing" \
+    python3 "scripts/process_fips_module.py" "%{buildroot}%{_libdir}/libsymcrypt.so.%{version}" --processing-dir "bin/module/generic/processing" --debug \
+    chmod 755 "%{buildroot}%{_libdir}/libsymcrypt.so.%{version}" \
```
dmcilvaney, line 83: I should probably know this, but do we need a `;` or `&&` to split these shell commands? (ie `mkdir`, `python3`, `chmod`)

dmcilvaney: nvm, found my answer. tldr `\` it consumed by the macro parser and just tells it to keep reading, it takes the newline character verbatim. `\\\` is needed if you want traditional shell behaviour of `\\`.

dmcilvaney (approved):

christopherco (approved):

SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -1,15 +1,17 @@
...
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 Group:          System/Libraries
 URL:            https://github.com/microsoft/SymCrypt
 Source0:        https://github.com/microsoft/SymCrypt/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
 Source1:        https://github.com/smuellerDD/jitterentropy-library/archive/v3.3.1.tar.gz#/jitterentropy-library-3.3.1.tar.gz
+Source2:        find-debuginfo
+Patch1:         0001-add-build-flags-to-prevent-stripping-and-post-proces.patch
+Patch2:         0001-add-parameter-to-process_fips_module-to-specify-the-.patch
 BuildRequires:  cmake
```
PawelWMS, line 15: Nit-pick: could you only modify the patch names a bit, remove the noise like "0001" or make them concise enough, so that they are not truncated like the "the-" ending of the second one? Nothing big, just so it's easier for the reader.

SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -1,15 +1,17 @@
...
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        MIT
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 Group:          System/Libraries
 URL:            https://github.com/microsoft/SymCrypt
 Source0:        https://github.com/microsoft/SymCrypt/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
 Source1:        https://github.com/smuellerDD/jitterentropy-library/archive/v3.3.1.tar.gz#/jitterentropy-library-3.3.1.tar.gz
+Source2:        find-debuginfo
```
PawelWMS, line 12: Where did we get the modified `find-debuginfo` script from? If it's an updated version of the one from the `debugedit` package, then I think it might even better to update that script in `debugedit.spec` itself. This way it's available everywhere.

SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -40,26 +42,47 @@ A core cryptographic library written by Microsoft
...
+# The post-processing script writes the modified file to the same location as the original file, which
+# is subject to default permissions, so we need to set permissions manually after the script.
+%define __spec_install_post \
+    %{?__debug_package:%{__debug_install_post}} \
+    %{__arch_install_post} \
+    %{__os_install_post} \
+    mkdir -p "bin/module/generic/processing" \
+    python3 "scripts/process_fips_module.py" "%{buildroot}%{_libdir}/libsymcrypt.so.%{version}" --processing-dir "bin/module/generic/processing" --debug \
+    chmod 755 "%{buildroot}%{_libdir}/libsymcrypt.so.%{version}" \
+%{nil}
```
PawelWMS, line 84: Small request to try something out. Since we're only appending to the original `__spec_install_post`, I'm curious if we could do this to make it clearer what is the new part and that we're not skipping any original steps:
```
%define __spec_install_post \
    %{?__spec_install_post} \
    mkdir -p "bin/module/generic/processing" \
    python3 "scripts/process_fips_module.py" "%{buildroot}%{_libdir}/libsymcrypt.so.%{version}" --processing-dir "bin/module/generic/processing" --debug \
    chmod 755 "%{buildroot}%{_libdir}/libsymcrypt.so.%{version}" \
tobiasb-ms marked this conversation as resolved.
%{nil}
```

That way, if `__spec_install_post` gets updated, we also won't miss out on the update.

SPECS/SymCrypt/SymCrypt.spec
tobiasb-ms: I feel pretty strongly that we should *not* do that -- this is currently `SymCrypt`-specific and we don't want to potentially cause problems elsewhere.

If at some point `eu-strip` supported `--keep-symbol`, I'd be all for piping that argument through `find-debuginfo` everywhere, but that isn't true right now.

SPECS/SymCrypt/SymCrypt.spec
tobiasb-ms: That's a great idea, assuming it works. I'll experiment.

tobiasb-ms: I tried this and it gets an infinite recursion issue. Doesn't seem like it will work. Too bad.

SPECS/SymCrypt/SymCrypt.spec
tobiasb-ms: I at least mostly agree. If I end up making a different change before this is merged, I'll change them. If not, it's not the end of the world: These changes have been accepted upstream so at some point we'll get rid of these patches altogether.

SPECS/SymCrypt/find-debuginfo
```diff
@@ -0,0 +1,790 @@
...
+  #     eu-strip --remove-comment $r $g ${keep_remove_args} ${keep_symbol_args} -f "$1" "$2" || exit
+  #
+  # Always remove the comment section
+  keep_remove_args="${keep_remove_args} --remove-section .comment"
+
+  # We don't set $r and $g in my testing. If we ever do set those, then this script
+  # will need to be modified to either find the equivalent in strip, or intentionally
+  # decide to not support these flags.
+  if [ -n "$r" ] || [ -n "$g" ]; then
+    echo >&2 "*** ERROR: strip_to_debug: r ('${r}') or g ('${r}') is set, but this script does not support it"
```
PawelWMS, line 348:
```suggestion
    echo >&2 "*** ERROR: strip_to_debug: r ('${r}') or g ('${g}') is set, but this script does not support it"
```

SPECS/SymCrypt/find-debuginfo
```diff
@@ -0,0 +1,790 @@
...
+  --keep-section)
+    keep_remove_args="${keep_remove_args} --keep-section $2"
+    shift
+    ;;
+  --remove-section)
+    keep_remove_args="${keep_remove_args} --remove-section $2"
+    shift
+    ;;
+  --keep-symbol)
+    keep_symbol_args="${keep_symbol_args} --keep-symbol $2"
```
PawelWMS, line 238: Idea: maybe we can merge `keep_symbol_args` into `keep_remove_args`.

PawelWMS (approved):

SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -40,26 +42,47 @@ A core cryptographic library written by Microsoft
...
         -DCMAKE_CXX_COMPILER=%{symcrypt_cxx} \
         -DCMAKE_C_FLAGS="%{symcrypt_c_flags}" \
         -DCMAKE_CXX_FLAGS="-Wno-unused-but-set-variable"
 
 cmake --build bin
 
+# Override the default find-debuginfo script to our own custom one, which is modified
+# to allow us to keep symbols.
+# Also add custom options to the call to find-debuginfo.
+%define __find_debuginfo ./find-debuginfo
```
PawelWMS, line 71: Maybe?

```suggestion
%define __find_debuginfo "%{SOURCE2}"
```

SPECS/SymCrypt/SymCrypt.spec
```diff
@@ -40,26 +42,47 @@ A core cryptographic library written by Microsoft
...
+%define _find_debuginfo_opts --keep-symbol SymCryptVolatileFipsHmacKey --keep-symbol SymCryptVolatileFipsHmacKeyRva --keep-symbol SymCryptVolatileFipsBoundaryOffset --keep-symbol SymCryptVolatileFipsHmacDigest
```
PawelWMS, line 72:
```suggestion
%define _find_debuginfo_opts \
    --keep-symbol SymCryptVolatileFipsHmacKey \
    --keep-symbol SymCryptVolatileFipsHmacKeyRva \
    --keep-symbol SymCryptVolatileFipsBoundaryOffset \
    --keep-symbol SymCryptVolatileFipsHmacDigest
```

anphel31 (approved):

PawelWMS (approved):

anphel31:
The only fasttrack PR failures are ptests for `bpftrace` and `libguestfs`, which are both known failures

PR #10789, opened 2024-10-18
Image Customizer: Make verity API a list.

cwize1:
Change the verity config from a single item to a list of items. This is being done so that it is easier to add support for other verity partitions (e.g. /usr) in the future. However, this change restricts the verity API to only the root partition (`/`).

In addition, move the verity config from `.os` to `.storage`. This is being done for alignment with the Trident API. But is also probably a more morally correct place for verity to be placed.

As a side effect, this change removes support for enabling verity on a base image that somehow had all the correct partitions to support verity but didn't actually have verity enabled. None of our base images are like this. So, it is expected that no user ever made use of this functionality. This functionality could be re-added in the future. It was omitted from this change to avoid adding additional complexity.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Ran UTs.


toolkit/tools/imagecustomizerapi/verity.go
```diff
@@ -5,25 +5,55 @@ package imagecustomizerapi
...
+
+var (
+	verityNameRegex = regexp.MustCompile("^[a-z]+$")
 )
 
 type Verity struct {
-	DataPartition    IdentifiedPartition `yaml:"dataPartition"`
-	HashPartition    IdentifiedPartition `yaml:"hashPartition"`
-	CorruptionOption CorruptionOption    `yaml:"corruptionOption"`
+	// ID is used to correlate `Verity` objects with `FileSystem` objects.
```
liulanze, line 21: You forgot aligning struct columns, perhaps:

```
Id                      string               `yaml:"id"`
Name                    string               `yaml:"name"`
DataDeviceId            string               `yaml:"dataDeviceId"`
DataDeviceMountIdType   MountIdentifierType  `yaml:"dataDeviceMountIdType"`
HashDeviceId            string               `yaml:"hashDeviceId"`
HashDeviceMountIdType   MountIdentifierType  `yaml:"hashDeviceMountIdType"`
CorruptionOption        CorruptionOption     `yaml:"corruptionOption"`
```

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
```diff
@@ -80,7 +80,7 @@ func doOsCustomizations(buildDir string, baseConfigPath string, config *imagecus
 		return err
 	}
 
-	verityUpdated, err := enableVerityPartition(config.OS.Verity, imageChroot)
+	verityUpdated, err := enableVerityPartition(config.Storage.Verity, imageChroot)
```
liulanze, line 83: Since the Verity config has been moved to `.storage`, does it still make sense to put it under the `customizeos`? Though I agree here is still the best sequence place.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -63,7 +63,7 @@ func updateFstabForVerity(imageChroot *safechroot.Chroot) error {
 	for _, entry := range fstabEntries {
 		if entry.Target == "/" {
 			// Replace existing root partition line with the Verity target.
-			entry.Source = "/dev/mapper/root"
+			entry.Source = imagecustomizerapi.VerityRootDevicePath
```
liulanze, line 66: Perhaps we can also pass `rootfsVerity` as parameter of function `updateFstabForVerity`, then give `entry.Source = "dev/mapper/" + rootfsVerity.Name`. This way it can support `usr` Verity as well though for now we only support `rootfs` Verity, thoughts?

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -162,66 +165,50 @@ func updateGrubConfigForVerity(verity *imagecustomizerapi.Verity, rootHash strin
...
-		return true, fmt.Errorf("invalid idType provided (%s)", string(partitionId.IdType))
-	}
+) bool {
+	partUuid := partIdToPartUuid[configDeviceId]
+	return partition.PartUuid == partUuid
 }
 
 // systemdFormatPartitionId formats the partition ID based on the ID type following systemd dm-verity style.
-func systemdFormatPartitionId(partition imagecustomizerapi.IdentifiedPartition, partIdToPartUuid map[string]string,
+func systemdFormatPartitionId(configDeviceId string, mountIdType imagecustomizerapi.MountIdentifierType,
```
liulanze, line 190: This `MountIdentifierType` use is impressive. Thanks.

toolkit/tools/imagecustomizerapi/verity.go
```diff
@@ -5,25 +5,55 @@ package imagecustomizerapi
 
 import (
 	"fmt"
+	"regexp"
+)
+
+const (
+	VerityRootDevicePath = "/dev/mapper/root"
+	VerityRootDeviceName = "root"
```
liulanze, line 13: I feel like this might be duplicated with `rootfsVerity.Name`, but if without this constant `VerityRootDeviceName`, we need to force the `rootfsVerity.Name` to be `root` when the rootfs Verity.

liulanze (approved):

toolkit/tools/imagecustomizerapi/verity.go
cwize1: The code has been formatted by the Go formatter. When you add comments in between the fields, then Go no longer aligns the values vertically.

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
cwize1: We are still making changes to the OS to prepare for verity. So, I think it makes sense.

toolkit/tools/imagecustomizerapi/verity.go
cwize1: I've removed `VerityRootDevicePath`.

`VerityRootDeviceName` has to stick around since we need to use it in API validation.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
cwize1: Fixed.

liulanze (approved):
Thanks for the revision, looks good!

PR #10804, opened 2024-10-21
Toolkit: Add missing `flock` calls.

cwize1:
When making changes to partitions or filesystems, it is recommended to take a file lock over the disk block device as this informs the host OS that you are making changes and that it should avoid scanning or changing the device until you are done. While most of the relevant operations are covered, there a few places that are missing the lock. For example, when calling `mkfs` or `resize2fs`.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Ran image customizer UTs.

### DIFF
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -492,7 +492,7 @@ func CreatePartitions(diskDevPath string, disk configuration.Disk, rootEncryptio
 			return partDevPathMap, partIDToFsTypeMap, encryptedRoot, readOnlyRoot, err
 		}
 
-		partFsType, err := FormatSinglePartition(partDevPath, partition)
+		partFsType, err := formatSinglePartition(diskDevPath, partDevPath, partition)
 		if err != nil {
 			err = fmt.Errorf("failed to format partition:\n%w", err)
 			return partDevPathMap, partIDToFsTypeMap, encryptedRoot, readOnlyRoot, err
@@ -782,12 +782,13 @@ func setGptPartitionType(partition configuration.Partition, timeoutInSeconds, di
 	return
 }
 
-// FormatSinglePartition formats the given partition to the type specified in the partition configuration
-func FormatSinglePartition(partDevPath string, partition configuration.Partition,
+// formatSinglePartition formats the given partition to the type specified in the partition configuration
+func formatSinglePartition(diskDevPath string, partDevPath string, partition configuration.Partition,
 ) (fsType string, err error) {
 	const (
-		totalAttempts = 5
-		retryDuration = time.Second
+		totalAttempts    = 5
+		retryDuration    = time.Second
+		timeoutInSeconds = "5"
 	)
 
 	fsType = partition.FsType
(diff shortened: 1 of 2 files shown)


toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -803,12 +804,12 @@ func FormatSinglePartition(partDevPath string, partition configuration.Partition
 			fsType = "vfat"
 		}
 
-		mkfsArgs := []string{"-t", fsType}
+		mkfsArgs := []string{"--timeout", timeoutInSeconds, diskDevPath, "mkfs", "-t", fsType}
```
dmcilvaney, line 807: Doesn't need to be here, but I wonder if its worth having a dedicated flock function that wraps shell.Execute (or maybe an extra bit of logic on execute?).

dmcilvaney (approved):

toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
```diff
@@ -85,7 +85,8 @@ func shrinkFilesystems(imageLoopDevice string, verityHashPartition *imagecustomi
 		}
 
 		// Shrink the file system with resize2fs -M
-		stdout, stderr, err := shell.Execute("resize2fs", "-M", partitionLoopDevice)
+		stdout, stderr, err := shell.Execute("flock", "--timeout", "5", imageLoopDevice,
```
dmcilvaney, line 88: For consistency maybe use the same `timeoutInSeconds = "5"` pattern (or again, maybe this just lives in a dedicated spot).

cwize1: The Image Customizer codebase doesn't use that code pattern as much. Hence the inconsistency.

toolkit/tools/imagegen/diskutils/diskutils.go
cwize1: I have thought about potentially creating a helper type that calls the file lock API directly, instead of using `flock`. Though such a change would be rather involved. So, I am not sure how worth it it would be, seeing that using `flock` doesn't seem to cause any problems other than making the code slightly harder to read.

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -803,12 +804,12 @@ func FormatSinglePartition(partDevPath string, partition configuration.Partition
...
 		mkfsArgs = append(mkfsArgs, mkfsOptions...)
 		mkfsArgs = append(mkfsArgs, partDevPath)
 
 		err = retry.Run(func() error {
-			_, stderr, err := shell.Execute("mkfs", mkfsArgs...)
+			_, stderr, err := shell.Execute("flock", mkfsArgs...)
 			if err != nil {
 				logger.Log.Warnf("Failed to format partition using mkfs: %v", stderr)
```
PawelWMS, line 814: Could we also update the log messages to mention that now either `flock` or `mkfs` could have failed. I think it would be a useful hint for debugging purposes.

cwize1: The string `"Failed to format partition using mkfs"` is unique within the codebase. So, if we observe it in the logs, we will know the code failed at this location.

romoh (approved):

toolkit/tools/imagegen/diskutils/diskutils.go
PawelWMS: My comment was more about helping the dev more quickly know what was run from the log itself without having to look at the code. Especially since not everyone is familiar with it or needs to be. If the error message is not `mkfs`-specific, it will get a bit confusing and you'll need to dive into the code, to figure out what's happening.

cwize1: Ok, fixed.

PawelWMS (approved):

PR #10850, opened 2024-10-25
Port fasttrack PR check to PRs that target dev branches

PawelWMS:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This is porting our FastTrack PR build check into PRs that target dev branches, except they also use daily repo to grab packages from

###### Change Log  

- Added FastTrack style PR check for used for dev PRs, uses daily repo for packages to be downloaded
- Moved out common PR check code into template file

**NO**

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=823650&view=results


.pipelines/templates/PackageBuild.yml
```diff
@@ -226,6 +231,12 @@ steps:
         max_cascading_rebuilds_arg="MAX_CASCADING_REBUILDS=${{ parameters.maxCascadingRebuilds }}"
       fi
 
+      additional_args=""
+      if [[ ${{ parameters.dailyBuildID }} == "lkg" ]]; then
+        echo "##[debug]Building toolchain for latest daily build without using daily build manifests"
+        additional_args="DAILY_BUILD_ID=${{ parameters.dailyBuildID }} DAILY_BUILD_ID_UPDATE_MANIFESTS=n"
+      fi
```
PawelWMS, line 238: The `DAILY_BUILD_ID` argument is separate from `DAILY_BUILD_ID_UPDATE_MANIFEST`. Don't bundle them together. This is a generic template for building packages - it should have no assumptions about how it's used.

.pipelines/templates/ToolchainBuild.yml
```diff
@@ -27,7 +32,16 @@ steps:
     parameters:
       buildRepoRoot: ${{ parameters.buildRepoRoot }}
 
-  - bash: sudo make -C "${{ parameters.buildRepoRoot }}/toolkit" "-j$(nproc)" toolchain QUICK_REBUILD=y
+  - bash: |
+      additional_args=""
```
PawelWMS, line 36: Nit-pick: could we re-name `additional_args` to something more specific? For instance `daily_build_args`. I don't want to encourage bloating this variable with any future args, we may want to add here.

.pipelines/templates/PackageBuild.yml
mbykhovtsev-ms: Split them into their own arguments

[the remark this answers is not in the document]
mbykhovtsev-ms: Removed echo

.pipelines/templates/ToolchainBuild.yml
mbykhovtsev-ms: Renamed it to be explicit and also split them up

.pipelines/templates/PackageBuild.yml
```diff
@@ -244,7 +255,8 @@ steps:
         $quick_rebuild_packages_arg \
         $run_check_arg \
         $toolchain_archive_arg \
-        $use_ccache_arg
+        $use_ccache_arg \
+        $additional_args
```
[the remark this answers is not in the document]
mbykhovtsev-ms: sorted

.pipelines/templates/PackageBuild.yml
```diff
@@ -226,6 +231,14 @@ steps:
         max_cascading_rebuilds_arg="MAX_CASCADING_REBUILDS=${{ parameters.maxCascadingRebuilds }}"
       fi
 
+      if [[ ${{ parameters.dailyBuildID }} == "lkg" ]]; then
+        daily_build_id_arg="DAILY_BUILD_ID=${{ parameters.dailyBuildID }}"
+        daily_build_update_manifests_arg="DAILY_BUILD_ID_UPDATE_MANIFESTS=n"
+      elif [[ -n "${{ parameters.dailyBuildID }}" ]]; then
+        daily_build_id_arg=""
+        daily_build_update_manifests_arg=""
+      fi
```
PawelWMS: We're getting closer but this is not exactly it. The `PackageBuild.yml` template does not make any guesses as to what the caller wants to do, so hard-coding `DAILY_BUILD_ID_UPDATE_MANIFESTS=n` is not the way to go. It would be confusing behaviour, if someone used that template to build packages from a dev branch and with `dailyBuildID=lkg` and got a build error, because the manifests were not updated. From the point of view of this template, these two arguments are independent and should be treated as such. The caller is the one, who knows how they want to build the packages and will thus set them accordingly.

.pipelines/templates/ToolchainBuild.yml
```diff
@@ -27,7 +32,19 @@ steps:
...
 
-  - bash: sudo make -C "${{ parameters.buildRepoRoot }}/toolkit" "-j$(nproc)" toolchain QUICK_REBUILD=y
+  - bash: |
+      if [[ ${{ parameters.dailyBuildID }} == "lkg" ]]; then
+        daily_build_id_arg="DAILY_BUILD_ID=${{ parameters.dailyBuildID }}"
+        daily_build_update_manifests_arg=DAILY_BUILD_ID_UPDATE_MANIFESTS=n"
+      elif [[ -n "${{ parameters.dailyBuildID }}" ]]; then
+        daily_build_id_arg=""
+        daily_build_update_manifests_arg=""
+      fi
```
PawelWMS: Similar request as for the package builds.

.pipelines/templates/PackageBuild.yml
mbykhovtsev-ms: The thing is that `DAILY_BUILD_ID_UPDATE_MANIFESTS` depends on `DAILY_BUILD_ID` being set. So someone passing down `dailyBuildID` as empty string parameter to this template and then something like `dailyBuildUpdateManifests` parameter as `true`/`y` would not make sense and result in no different behavior than just passing alone `dailyBuildID` parameter as empty string https://github.com/microsoft/azurelinux/blob/3.0/toolkit/scripts/daily_build.mk#L23

I can still change it to be parameter based value, but I don't see `DAILY_BUILD_ID_UPDATE_MANIFESTS` behavior being decouples from the `DAILY_BUILD_ID` being set, as in order for it to update the manifests and pull them, it has to know the `DAILY_BUILD_ID` to use https://github.com/microsoft/azurelinux/blob/3.0/toolkit/scripts/manifests.mk#L14

mbykhovtsev-ms: Updated

.pipelines/templates/ToolchainBuild.yml
mbykhovtsev-ms: Updated

.pipelines/templates/PackageBuildPRCheck.yml
```diff
@@ -10,6 +10,13 @@ parameters:
     default: ""
     displayName: "Daily build ID"
 
+  - name: isDailyBuildUpdateManifests
```
PawelWMS: On this level, I think we don't need to be so detailed on the other hand. What we care about during PR checks, if we should use the published public RPMs or a daily build ID, so I think we can simplify this to one parameter. The caller shouldn't need to worry about implementation details like the manifest updates.

.pipelines/templates/ToolchainBuild.yml
```diff
@@ -11,6 +11,13 @@ parameters:
     default: ""
     displayName: "Daily build ID"
 
+  - name: isDailyBuildUpdateManifests
+    type: string
+    default: "true"
+    values:
+      - "true"
```
PawelWMS: Please also add the "default" option (`default: "default"`) like other string parameters pretending to be booleans. By forcing a specific value right now we're overwriting the toolkit's defaults, so if they ever change, we don't follow them. Thus the additional "default" string value.

.pipelines/templates/PackageBuild.yml
```diff
@@ -18,6 +18,18 @@ parameters:
     type: string
     default: ""
 
+  - name: dailyBuildID
+    type: string
+    default: ""
+    displayName: "Daily build ID"
+
+  - name: isDailyBuildUpdateManifests
```
PawelWMS: Please keep parameters in alphabetical order as well. Thanks!

.pipelines/templates/ToolchainBuild.yml
```diff
@@ -27,7 +39,24 @@ steps:
     parameters:
       buildRepoRoot: ${{ parameters.buildRepoRoot }}
 
-  - bash: sudo make -C "${{ parameters.buildRepoRoot }}/toolkit" "-j$(nproc)" toolchain QUICK_REBUILD=y
+  - bash: |
+      if [[ ${{ parameters.dailyBuildID }} == "lkg" ]]; then
+        daily_build_id_arg="DAILY_BUILD_ID=${{ parameters.dailyBuildID }}"
+      elif [[ -n "${{ parameters.dailyBuildID }}" ]]; then
+        daily_build_id_arg=""
+      fi
```
PawelWMS: Why do we want to set the arg to an empty string if `dailyBuildID` is not an empty string but it's different than "lkg"? Also no need to set an empty string - uninitialized variables are empty by default in Bash. Unless you use `set -u`, in which case using them is considered an error.:)

PawelWMS: Also, I think instead of `if [[ ${{ parameters.dailyBuildID }} == "lkg" ]];` we're more interested in `if [[ -n "${{ parameters.dailyBuildID }}" ]];`.

.pipelines/templates/PackageBuild.yml
```diff
@@ -18,6 +18,18 @@ parameters:
...
+    type: string
+    default: "true"
+    values:
+      - "true"
+      - "false"
```
PawelWMS:
```suggestion
  - name: isDailyBuildUpdateManifests
    type: string
    default: "default"
    values:
      - "default"
      - "false"
      - "true"
```

.pipelines/templates/PackageBuild.yml
```diff
@@ -226,6 +239,16 @@ steps:
         max_cascading_rebuilds_arg="MAX_CASCADING_REBUILDS=${{ parameters.maxCascadingRebuilds }}"
       fi
 
+      if [[ -n "${{ parameters.dailyBuildID }}" ]]; then
+        daily_build_id_arg="DAILY_BUILD_ID=${{ parameters.dailyBuildID }}"
+      fi
+
+      if [[ ${{ parameters.isDailyBuildUpdateManifests }} == "true" ]]; then
```
PawelWMS, line 246: Still needs an update to handle "default" like other parameters.

.pipelines/templates/ToolchainBuild.yml
```diff
@@ -27,7 +40,22 @@ steps:
     parameters:
       buildRepoRoot: ${{ parameters.buildRepoRoot }}
 
-  - bash: sudo make -C "${{ parameters.buildRepoRoot }}/toolkit" "-j$(nproc)" toolchain QUICK_REBUILD=y
+  - bash: |
+      if [[ -n ${{ parameters.dailyBuildID }} ]]; then
+        daily_build_id_arg="DAILY_BUILD_ID=${{ parameters.dailyBuildID }}"
+      fi
+
+      if [[ ${{ parameters.isDailyBuildUpdateManifests }} == "true" ]]; then
```
PawelWMS, line 48: Some comment about "default" here as well.

.pipelines/templates/PackageBuildPRCheck.yml
mbykhovtsev-ms: Updated

.pipelines/templates/ToolchainBuild.yml
mbykhovtsev-ms: Updated

.pipelines/templates/PackageBuild.yml
mbykhovtsev-ms: Updated

.pipelines/templates/ToolchainBuild.yml
mbykhovtsev-ms: Updated

.pipelines/templates/PackageBuild.yml
mbykhovtsev-ms: Updated

.pipelines/templates/PackageBuild.yml
mbykhovtsev-ms: Sorry, I don't understand. The way its made right now is to be just like other parameters, for example `isDeltaBuild`. If this parameter is set to `default`, based on the conditional here it will make `daily_build_update_manifests_arg` set to nothing which is what default behavior will be

PawelWMS: You're right, sorry.

PawelWMS:
@CBL-Mariner-Stable-Maintainers: approved, but it's my own PR, so I can't do it.

jslobodzian (approved):

PR #10880, opened 2024-10-29
Image Customizer: Set VHDX block-size to 2 MiB.

cwize1:
Currently, the size of outputted VHDX files can be substantially larger than VHD or qcow2 files. This behavior occurs because `qemu-img` dynamically sets the block-size based on the size of the disk. And larger blocks sizes means the file format is less space efficient, which results in a larger image file.

This change sets the block-size to 2 MiB for VHDX files. This matches the `qemu-img` defaults for VHD and QCOW2.

For reference, Windows seems to use a default block-size of 1 MiB for VHDX files.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

- Manually ran image customizer.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
@@ -466,6 +466,12 @@ func toQemuImageFormat(imageFormat string) (string, string) {
 	case ImageFormatVhdFixed:
 		return QemuFormatVpc, "subformat=fixed,force_size"
 
+	case ImageFormatVhdx:
+		// For VHDX, qemu-img dynamically picks the block-size based on the size of the disk.
+		// However, this can result in a significantly larger file size than other formats.
+		// So, use a fixed block-size of 2 MiB to match the block-sizes used for qcow2 and VHD.
+		return ImageFormatVhdx, "block_size=2097152"
+
 	default:
 		return imageFormat, ""
 	}


toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -466,6 +466,12 @@ func toQemuImageFormat(imageFormat string) (string, string) {
 	case ImageFormatVhdFixed:
 		return QemuFormatVpc, "subformat=fixed,force_size"
 
+	case ImageFormatVhdx:
+		// For VHDX, qemu-img dynamically picks the block-size based on the size of the disk.
+		// However, this can result in a significantly larger file size than other formats.
+		// So, use a fixed block-size of 2 MiB to match the block-sizes used for qcow2 and VHD.
+		return ImageFormatVhdx, "block_size=2097152"
```
gmileka, line 473: It sounds like 2MB per block is pretty large still given that there probably many files that are much smaller than 2MBs (all configuration files, scripts, etc).

gmileka (approved):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: This doesn't affect the block-size of the filesystem or the underlying physical disk. It just affects the block lookup table in the VHDX file.

For reference, a comparison of block-sizes vs. disk sizes for core-efi image:

- 1 MiB: 537 MiB
- 2 MiB: 583 MiB

Note: 1 MiB is the smallest block size possible for VHDX.

So, I guess there is a little bit of savings to be had by dropping down to 1 MiB (~10%). But not a lot.

gmileka: I see - thx Chris for explaining. Yep, it has minimal effect. I'm fine merged as-is.

PR #10995, opened 2024-11-10
Update shim to v15.8

christopherco:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Update the Azure Linux 3.0 shim to the latest version v15.8.
This change also adds MOK Manager (mm) support for allowing users to enroll their own keys into the system.
We do build fallback.efi as well, but for Azure Linux 3.0, we will not be utilizing it in order to keep consistency with our 3.0 preview and previous offerings. We will re-evaluate bringing in fallback.efi in the next major OS release.
Additionally, this new shim embeds our updated Azure Linux Trusted Base CA certificate.
This update also includes the signed shim and mm binaries into the "shim" package for wider use.
We also deprecate the previous "shim-unsigned" package as it is no longer needed, and we update the references in the ARM64 image definitions to the "shim" package.
Finally, the change bumps the grub package to ensure it gets signed with the newer key and the RPM dependencies are set such that this shim and grub should update together.

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/48576481
- Supersedes https://github.com/microsoft/azurelinux/pull/10903 and https://github.com/microsoft/azurelinux/pull/8406

###### Test Methodology

- Pipeline build id: [672965](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=672965&view=results)
- Image Test build: [673763](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=673763&view=results)
   - shim boots as expected when Microsoft UEFI CA is selected as certificate. Expected for grub binary to not boot since it is not signed in test build. On ISO, MOK does not boot as expected since toolkit changes are needed to add mm.efi to the ISO. On standard images, MOK does get verified correctly and boots to allow enrolling keys and hashes.
   - images and ISO boots (and installs in case of ISO) successfully when secure boot is disabled.


SPECS/shim-unsigned-aarch64/shim-unsigned-aarch64.spec
```diff
@@ -1,65 +1,222 @@
...
+
+# For prereleases, % global prerelease rc2, and downpatch Makefile
+%if %{defined prerelease}
+%global dashpre -%{prerelease}
+%global dotpre .%{prerelease}
+%global tildepre ~%{prerelease}
+%global zdpd 0%{dotpre}.
+%endif
+
+%global efidir azurelinux
```
ms-mahuber, line 12: Just a q: Here, for the aarch64 case, we intentionally want azurelinux?

SPECS/shim-unsigned-aarch64/shim-unsigned-aarch64.spec
```diff
@@ -1,65 +1,222 @@
...
+Vendor:		Microsoft Corporation
+Distribution:	Azure Linux
+URL:		https://github.com/rhboot/shim
+Source0:	https://github.com/rhboot/shim/releases/download/%{version}%{?dashpre}/shim-%{version}%{?dotpre}.tar.bz2
+Source1:	azurelinux-ca-20230216.der
+%if 0%{?dbxfile}
+Source2:	%{dbxfile}
+%endif
+Source3:	sbat.azurelinux.csv
+Source4:	shim.patches
```
ms-mahuber, line 48: A generally new pattern to list .patch files we add?

SPECS/shim-unsigned-x64/shim-unsigned-x64.spec
```diff
@@ -1,80 +1,284 @@
...
+Vendor:		Microsoft Corporation
+Distribution:	Azure Linux
+URL:		https://github.com/rhboot/shim
+Source0:	https://github.com/rhboot/shim/releases/download/%{version}%{?dashpre}/shim-%{version}%{?dotpre}.tar.bz2
+Source1:	azurelinux-ca-20230216.der
+%if 0%{?dbxfile}
+Source2:	%{dbxfile}
+%endif
+Source3:	sbat.azurelinux.csv
+Source4:	shim.patches
```
ms-mahuber, line 52: Renders empty to me. Probably want to list the patch files. I know the PR is not ready for review. Just pre-emptively iterating of it.

SPECS/shim-unsigned-aarch64/shim-unsigned-aarch64.spec
rlmenge: Both shim-unsigned-x64.spec and shim-unsigned-aarch64.spec use `%global efidir azurelinux`. Unclear if this is intentional as it matches how other distros offer shim. 

Our actual signed shim rpm (from shim.spec) still uses the old file paths `/boot/efi/EFI/BOOT/boot{arch}.efi` as described in commit https://github.com/microsoft/azurelinux/pull/10995/commits/93c34aaf82b73dba7d3187629c7eb21888cf6210

SPECS/shim-unsigned-x64/shim-unsigned-x64.spec
rlmenge: looks to be an artifact from using the fedora spec :) https://github.com/microsoft/azurelinux/pull/10995/commits/73c0c0065cac90bca75ea7a5fc30981e1f917651

jslobodzian (approved):

PR #11019, opened 2024-11-12
[DRAFT / RFC] Add PR checks for build/test/install

reubeno:
_(Draft change -- requesting feedback/comments)_

This is an initial attempt at adding PR checks that detect changed specs, build + test them, and try to install/uninstall the built RPMs. For an example run, see this [private run](https://github.com/reubeno/azurelinux/actions/runs/11787998425).

Note that the install/uninstall script is an as-is copy from existing ADO pipeline scripts.

The changes to the toolkit golang code needs the most feedback (e.g., the emitting of a tests results JSON). Specifically note the proposal to allow a .spec to use a specialized comment in this line:

```
#!AZL expected-test-failure
```

to denote a baselined ptest failure.

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -1,0 +1,318 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+#
+# NOTES:
+#
+# This is a starting point for a validation workflow that can detect the specs that
+# have changed in a given git commit, build the RPMs for those specs, and run ptests
+# for those.
+#
+# Known limitations/concerns:
+#
+# * We build + test the packages in an Ubuntu VM runner. This is easiest in a GitHub
+#   action, since it's the only Linux OS supported running in the VM; in time,
+#   it may be worthwhile looking into running the package build itself in an
+#   Azure Linux 3.0 environment.
+#
+# * We take a specific dependency on daily repos in Azure Linux 3.0+, as well as the
+#   Azure Linux 3.0 container. To extend support to 2.0, some work will be required
+#   to make these dependencies conditional.
+#
+# * Packages without %check sections will have failed test runs.
+#
+# * This workflow doesn't yet support building or running tests for "EXTENDED" specs.
+#
+
+name: "Spec Build/Test"
+
+on:
+  # N.B. For now we enable this for 3.0-dev, but expect that the logic below will work
+  # for other branches as well.
+  push:
+    branches: [3.0-dev]
+  pull_request:
+    branches: [3.0-dev]
+  workflow_dispatch:
+    inputs:
+      comparison_base_ref:
+        description: 'The base ref to compare changest against.'
+        required: true
+        default: 'refs/heads/3.0-dev'
+
+env:
+  # For consistency, we use the same major/minor version of the latest version of Python that
+  # Azure Linux ships.
+  PYTHON_VERSION: "3.12"
+
+jobs:
+  detect-changes:
+    name: 'Detect spec changes'
+    runs-on: ubuntu-latest
+
+    outputs:
+      updated_specs: ${{ steps.export.outputs.UPDATED_SPECS }}
+      updated_extended_specs: ${{ steps.export.outputs.UPDATED_EXTENDED_SPECS }}
+      target_ref: ${{ steps.export.outputs.TARGET_REF }}
+      azl_extra_make_args: ${{ steps.export.outputs.AZL_EXTRA_MAKE_ARGS }}
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Setup Python ${{ env.PYTHON_VERSION }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Install Python dependencies for tools
+        run: python3 -m pip install -r toolkit/scripts/requirements.txt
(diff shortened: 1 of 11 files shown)


toolkit/scripts/pkggen.mk
```diff
@@ -38,6 +38,7 @@ cached_file              = $(PKGBUILD_DIR)/cached_graph.dot
 preprocessed_file        = $(PKGBUILD_DIR)/preprocessed_graph.dot
 built_file               = $(PKGBUILD_DIR)/built_graph.dot
 output_csv_file          = $(PKGBUILD_DIR)/build_state.csv
+test_results_file		 = $(PKGBUILD_DIR)/test_results.json
```
0xba1a, line 41: Need tabs to be converted to spaces. In other places too.

toolkit/tools/pkg/specreaderutils/specreaderutil.go
```diff
@@ -704,6 +713,49 @@ func readBuildRequires(specFile, sourceDir, arch string, defines map[string]stri
 	return
 }
 
+func querySpecForTestFailureExpectation(specFile string) (result bool, err error) {
```
0xba1a, line 716: IMHO it is little over complication. If the spec writer knows that the PTest is expected to fail, they can simply disable the check section. So, instead of imposing everyone to add the special comment, we can ask them to disable the check.

0xba1a (commented):
Thanks @reubeno for taking it forward!

I've made quick observation about introducing hints in the spec-file. I could be wrong. Please let me know if I miss any advantage in running test and ignoring the results over not running the tests at all.

I think this PR can be split into multiple chunks.
1. Capture test output into a machine readable JSon format
2. Introduction of run_ptests.py scripts and necessary parameters
3. Introduction of install/uninstall tests
4. Build-check workflow for GitHub

Because, splitting the PR will let us to avoid unintentional inter-dependency.

I'm continuing to review the PR. I'll add more comments the more I understand the changes.

toolkit/tools/pkg/specreaderutils/specreaderutil.go
reubeno: Right now we find ourselves with a large number of failing tests. We're going to need to make a bulk change to baseline those known-failing tests. I'm definitely open to other approaches for capturing it if you have alternate recommendations.

And note I intentionally *don't* want to disable the check section. If a test is failing, IMHO we should still run it but not report the failure as blocking.

reubeno:
> I think this PR can be split into multiple chunks.

100% agreed. I'd still like to get feedback on the approach before doing that, though.

.github/workflows/check-build.yml
```diff
@@ -0,0 +1,321 @@
...
+        run: |
+          DAILY_BUILD_ID=$(toolkit/scripts/get_lkg_id.sh /tmp)
+          echo "DAILY_BUILD_ID=${DAILY_BUILD_ID}" >>$GITHUB_OUTPUT
+          echo "DAILY_BUILD_ARCH=x86-64" >>$GITHUB_OUTPUT
+
+      - name: "Build modified specs (base)"
+        id: build-specs
+        if: ${{ env.updated_specs != '' }}
+        run: |
+          sudo make -C toolkit \
```
0xba1a, line 169: @reubeno Does This workflow assume that the source-tarball is already uploaded to the source cache?

toolkit/scripts/detect_changes.py
```diff
@@ -0,0 +1,160 @@
...
+    return get_closest_git_ancestor(target_ref, well_known_refs)
+
+def is_well_known_branch_ref(branch_ref: str) -> bool:
+    branch_pieces = branch_ref.split("/")
+    if not branch_pieces:
+        return False
+
+    if branch_ref.startswith("refs/heads/"):
+        # Make sure there's no prefix on the branch name.
+        if len(branch_pieces) != 3:
```
0xba1a, line 62: I think this logic doesn't consider `fasttrack` branches.

.github/workflows/check-build.yml
reubeno: Yes, it does. Per our previous discussion that will need to be addressed in follow-up work.

toolkit/scripts/detect_changes.py
```diff
@@ -0,0 +1,160 @@
...
+args = parser.parse_args()
+
+if args.verbose:
+    log_level = logging.DEBUG
+else:
+    log_level = logging.INFO
+
+logging.basicConfig(format="%(levelname)s: %(message)s", level=log_level)
+
+def autodetect_compare_ref(target_ref: str) -> str:
```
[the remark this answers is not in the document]
reubeno: The goal was for this script to be runnable in an ADO build context, a GitHub action context, and locally within a developer's environment.

If you have some specific suggestions on what APIs you have in mind, I'd be eager to look into them.

toolkit/scripts/pkggen.mk
reubeno: Thank for catching this, I've fixed this one. Not sure where the others were you referring to?

toolkit/scripts/detect_changes.py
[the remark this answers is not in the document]
reubeno: Just looked back at this. The script *does* use GitHub PR metadata to find the comparison ref. I found the autodetection logic useful for local dev scenarios; it was otherwise ignored in the PR check.

PR #11038, opened 2024-11-12
Image Customizer: Use fixed xfs filesystem options.

cwize1:
Currently, we have a fixed set of options that are used when formatting with the ext4 filesystem type. However, this is currently missing for xfs. This means that if an Azure Linux 2.0 image is customized in Azure Linux 3.0, then the OS won't be able to boot since the xfs partition will have some features enabled that aren't supported on the older kernel/grub versions.

This change sets the xfs options so that they are consistent regardless of the OS being used to build/customize the image.

---

###### Merge Checklist  

- [x] Ready to merge

NO

###### Test Methodology

### DIFF
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -20,6 +20,7 @@ import (
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/retry"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/versioncompare"
 	"github.com/sirupsen/logrus"
 )
 
@@ -36,6 +37,10 @@ var (
 	}
 
 	partedVersionRegex = regexp.MustCompile(`^parted \(GNU parted\) (\d+)\.(\d+)`)
+
+	// Example:
+	//   mkfs.xfs version 6.5.0
+	mkfsXfsVersionRegex = regexp.MustCompile(`^mkfs\.xfs version (\d+\.\d+\.\d+)$`)
 
 	// The default partition name used when the version of `parted` is too old (<3.5).
 	LegacyDefaultParitionName = "primary"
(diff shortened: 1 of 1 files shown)


toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -840,20 +851,66 @@ func FormatSinglePartition(partDevPath string, partition configuration.Partition
...
+		mkfsXfsVersion, err := getMkfsXfsVersion()
+		if err != nil {
+			return nil, err
+		}
+
+		// Feature  xfsprogs   kernel
+		// nrext64  v5.19      v5.19
+		if mkfsXfsVersion.Compare(versioncompare.New("5.19")) >= 0 {
+			// Disable feature that is not supported on older kernel.
+			options := []string{"-i", "nrext64=0"}
```
reubeno, line 886: The PR summary says that we're trying to change the xfs options so they are consistent regardless of the host OS. Are we getting `mkfs.xfs` from the host system? If so, doesn't this version-specific logic mean that we *aren't* going to consistently create the filesystem? And if not, why are we checking versions?

What am I missing?

dmcilvaney: This threw me for a loop looking at the compare code....

I think our comment is just wrong in versioncompare.go (or I'm crazy...).

The compare returns `1` if `mkfsXfsVersion` is >= "5.91".

Soo if I'm understanding this right, xfsprogs <= 5.19 did not have the `nrext64` option at all, so if we see our version if higher than 5.19 (ie >= 0) we tell it to never use that feature so that the produced images can still run with the old tools in the produced image.

reubeno: Forgive me for being slow here, but how does the kernel relate here? Are there really *2* problems, i.e., (1) different versions of `mkfs.xfs` create the filesystem with different options, *and* (2) newer versions of `mkfs.xfs` create the filesystem with options that may not be supported on certain kernel versions? If so, which kernel versions are we talking about? The kernel version of the image being generated, or the version of the kernel running on the host?

cwize1: You are correct. In fact, I had a [version of this change](https://github.com/microsoft/azurelinux/tree/user/chrisgun/micXfsFeatures) that took into account: the OS kernel version, the build host kernel version, the version of the mkfs tools, and the version of grub. I wasn't a fan of the complexity of the change. Hence, I created this simpler version.

cwize1:
Moving to [azure-linux-image-tools](https://github.com/microsoft/azure-linux-image-tools) repo.

PR #11039, opened 2024-11-12
Toolkit: Use systemd-detect-virt instead of /.dockerenv to detect container builds.

dmcilvaney:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

There have been several issues with a mismatch between the build environment and the detected state by the toolkit. When running in docker, the chroots generally need to be configured externally with their mounts and re-used. This can be done via setting `CHROOT_DIR=/path/to/reusable/chroots`, and if the toolkit thinks it's in a container, it will switch modes.

See https://github.com/microsoft/azurelinux-tutorials/tree/main/build-in-container for more details.

Some builds are currently failing because what is ostensibly a container environment does not have `/.dockerenv` present. 

In the opposite direction, there are also situations where WSL images (which should work fine as a normal build) are reporting as docker because they have a `/.dockerenv` file present.

`systemd` has a tool (`systemd-detect-virt`) which is designed to detect what sort of virtualization is being used to run the current environment. Instead of designing a new system to re-implement this behavior, we can just use this tool. We already have an implicit build dependency on `systemd` (we run the docker service etc.) so adding it as an explicit requirement shouldn't change anything.

Also, to help people self-diagnose, sanity check the configurations and warn the user:
- If the CHROOT_DIR is set, but the tool thinks it's in a normal environment, print a warning. A fatal error will follow immediately after if this is actually broken.
- If the `systemd-detect-virt` tool is not present, print a warning but fallback to the old behavior.

To validate this we will need to add a new testcase to the toolkit sanity test pipeline that ensures the chroots keep working.

**We will need to back-port this to 2.0 as well.**

###### Change Log  

- Prefer `systemd-detect-virt` over `/.dockerenv` for container detection
- Print warnings if misconfiguration is detected

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/55008881
- https://microsoft.visualstudio.com/OS/_workitems/edit/54578711

###### Test Methodology

- New sanity test (~pipeline PR soon~): https://dev.azure.com/mariner-org/mariner/_build/results?buildId=675755&view=results
  - PR for new pipeline: https://dev.azure.com/mariner-org/mariner/_git/CBL-Mariner-Pipelines/pullrequest/21071
- Full build: ~https://dev.azure.com/mariner-org/mariner/_build/results?buildId=676056&view=results~ https://dev.azure.com/mariner-org/mariner/_build/results?buildId=678948&view=results

### DIFF
--- a/toolkit/tools/internal/buildpipeline/buildpipeline.go
+++ b/toolkit/tools/internal/buildpipeline/buildpipeline.go
@@ -8,28 +8,144 @@ package buildpipeline
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/file"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
-
-	"golang.org/x/sys/unix"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
 )
 
 const (
-	rootBaseDirEnv = "CHROOT_DIR"
-	chrootLock     = "chroot-pool.lock"
-	chrootUse      = "chroot-used"
+	rootBaseDirEnv        = "CHROOT_DIR"
+	chrootLock            = "chroot-pool.lock"
+	chrootUse             = "chroot-used"
+	systemdDetectVirtTool = "systemd-detect-virt"
 )
+
+var isRegularBuildCached *bool
+
+// checkIfContainerDockerEnvFile checks if the tool is running in a Docker container by checking if /.dockerenv exists. This
+// check may not be reliable in all environments, so it is recommended to use systemd-detect-virt if available.
+func checkIfContainerDockerEnvFile() bool {
+	exists, err := file.PathExists("/.dockerenv")
+	if err != nil {
+		logger.Log.Errorf("Error checking /.dockerenv: %v", err)
+		return false
+	}
+	return exists
+}
+
+// checkIfContainerIgnoreDockerEnvFile checks if the user has placed a file in the root directory to ignore the Docker
+// environment check.
+func checkIfContainerIgnoreDockerEnvFile() bool {
+	ignoreDockerEnvExists, err := file.PathExists("/.mariner-toolkit-ignore-dockerenv")
+	if err != nil {
+		logger.Log.Errorf("Failed to check if /.mariner-toolkit-ignore-dockerenv exists: %s", err)
+		return false
+	}
+	return ignoreDockerEnvExists
+}
+
+// checkIfContainerChrootDirEnv checks if the user has set the CHROOT_DIR environment variable, which is a requirement for
+// Docker-based builds. If the variable exists, it is likely that the tool is running in a Docker container.
+func checkIfContainerChrootDirEnv() bool {
+	_, exists := os.LookupEnv(rootBaseDirEnv)
+	return exists
+}
+
+// checkIfContainerSystemdDetectVirt uses systemd-detect-virt, a tool that can be used to detect if the system is running
+// in a virtualized environment. More specifically, using '-c' flag will detect container-based virtualization only.
+func checkIfContainerSystemdDetectVirt() (bool, error) {
+	// We should have the systemd-detect-virt command available in the environment, but check for it just in case since it
+	// was previously not explicitly required for the toolkit.
+	_, err := exec.LookPath(systemdDetectVirtTool)
+	if err != nil {
+		err = fmt.Errorf("failed to find %s in the PATH: %w", systemdDetectVirtTool, err)
+		return false, err
+	}
+
+	// The tool will return error code 1 based on detection, we only care about the stdout so ignore the return code.
+	stdout, _, _ := shell.Execute(systemdDetectVirtTool, "-c")
+
+	// There are several possible outputs from systemd-detect-virt we care about:
+	// - none: Not running in a virtualized environment, easy
+	// - wsl: Reports as a container, but we don't want to treat it as such. It should be able to handle regular builds
+	// - anything else: We'll assume it's a container
+	stdout = strings.TrimSpace(stdout)
+	switch stdout {
+	case "none":
+		logger.Log.Debugf("Tool is not running in a container, systemd-detect-virt reports: '%s'", stdout)
+		return false, nil
+	case "wsl":
+		logger.Log.Debugf("Tool is running in WSL, treating as a non-container environment, systemd-detect-virt reports: '%s'", stdout)
+		return false, nil
+	default:
+		logger.Log.Debugf("Tool is running in a container, systemd-detect-virt reports: '%s'", stdout)
+		return true, nil
+	}
+}
 
 // IsRegularBuild indicates if it is a regular build (without using docker)
 func IsRegularBuild() bool {
-	// some specific build pipeline builds Azure Linux from a Docker container and
-	// consequently have special requirements with regards to chroot
-	// check if .dockerenv file exist to disambiguate build pipeline
-	dockerEnvExists, _ := file.PathExists("/.dockerenv")
-	ignoreDockerEnvExists, _ := file.PathExists("/.mariner-toolkit-ignore-dockerenv")
-	return ignoreDockerEnvExists || !dockerEnvExists
+	if isRegularBuildCached != nil {
+		return *isRegularBuildCached
+	}
+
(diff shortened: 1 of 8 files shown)


toolkit/scripts/prerequisites.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash -e
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+prereq_md_path=""
+prereq_json_path=""
+
+function help {
+    echo "Pre-requisite installer. Installs azurelinux pre-requisites for the toolkit."
```
reubeno, line 9: Is the aspiration for this script to install the prerequisites, or just scrape out the list of them (for a separate command to install)? (For what it's worth, I'm thrilled to see someone put together an actual honest-to-goodness machine-parseable list of the prerequisites and am supportive of both goals :))

.github/workflows/check-prerequisites.yml
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) Microsoft Corporation.
```
reubeno, line 1: Woohoo, more GitHub PR checks 😄

toolkit/scripts/prerequisites.sh
dmcilvaney: This is going to go in another PR, need to polish it up some more. I'm actually going to mark this as a draft for now while I work to stabilize the pipeline side.

.github/workflows/check-prerequisites.yml
dmcilvaney: Split off to here https://github.com/microsoft/azurelinux/pull/11081

toolkit/tools/internal/buildpipeline/buildpipeline.go
```diff
@@ -8,28 +8,144 @@ package buildpipeline
...
+	chrootLock            = "chroot-pool.lock"
+	chrootUse             = "chroot-used"
+	systemdDetectVirtTool = "systemd-detect-virt"
 )
 
+var isRegularBuildCached *bool
+
+// checkIfContainerDockerEnvFile checks if the tool is running in a Docker container by checking if /.dockerenv exists. This
+// check may not be reliable in all environments, so it is recommended to use systemd-detect-virt if available.
+func checkIfContainerDockerEnvFile() bool {
```
cwize1, line 34: For helper functions like these, it is probably better to return the `error` and let the caller decide if they want to ignore the error or not. This helps to keep things generic.

It would also ensure `checkIfContainerIgnoreDockerEnvFile` and `checkIfContainerChrootDirEnv` are consistent with `checkIfContainerSystemdDetectVirt`.

toolkit/tools/internal/buildpipeline/buildpipeline.go
```diff
@@ -8,28 +8,144 @@ package buildpipeline
...
+}
+
+// checkIfContainerSystemdDetectVirt uses systemd-detect-virt, a tool that can be used to detect if the system is running
+// in a virtualized environment. More specifically, using '-c' flag will detect container-based virtualization only.
+func checkIfContainerSystemdDetectVirt() (bool, error) {
+	// We should have the systemd-detect-virt command available in the environment, but check for it just in case since it
+	// was previously not explicitly required for the toolkit.
+	_, err := exec.LookPath(systemdDetectVirtTool)
+	if err != nil {
+		err = fmt.Errorf("failed to find %s in the PATH: %w", systemdDetectVirtTool, err)
```
cwize1, line 68: nit: `\n%w`

toolkit/tools/internal/buildpipeline/buildpipeline.go
dmcilvaney: Good point, done.

nicogbg (approved):

dmcilvaney:
Pipe line to validate this keeps working: https://dev.azure.com/mariner-org/mariner/_git/CBL-Mariner-Pipelines/pullrequest/21071

toolkit/tools/pkggen/worker/create_worker_chroot.sh
```diff
@@ -10,12 +10,13 @@ set -o pipefail
 # $3 path to find RPMs. May be in PATH/<arch>/*.rpm
 # $4 path to log directory
 
-[ -n "$1" ] && [ -n "$2" ] && [ -n "$3" ] && [ -n "$4" ] || { echo "Usage: create_worker.sh <./worker_base_folder> <rpms_to_install.txt> <./path_to_rpms> <./log_dir>"; exit; }
+[ -n "$1" ] && [ -n "$2" ] && [ -n "$3" ] && [ -n "$4" ] && [ -n "$4" ] || { echo "Usage: create_worker.sh <./worker_base_folder> <rpms_to_install.txt> <./path_to_rpms> <./containercheck> <./log_dir>"; exit; }
```
jslobodzian, line 13: Did you intend to repeat the [-n "$4" ] after the previous $4 clause?

toolkit/tools/internal/buildpipeline/buildpipeline.go
```diff
@@ -8,28 +8,154 @@ package buildpipeline
...
+	isRegularBuild := true
+	isDockerContainer, err := checkIfContainerSystemdDetectVirt()
+	if err != nil {
+		isContainerBuild, err := checkIfContainerDockerEnvFile()
+		if err != nil {
+			// Log the error, but continue with the check.
+			logger.Log.Warnf("Failed to check if /.dockerenv exists: %s", err)
+		}
+		isRegularBuild = !isContainerBuild
+		message := []string{
```
jslobodzian, line 125: Note sure what "PrintMessageBox" is doing, but it seems to me the code would be a little simpler here if we just emitted the "Failed to detect" message" first, then composed the result

toolkit/tools/pkggen/worker/create_worker_chroot.sh
dmcilvaney: Nope, good catch thanks
```suggestion
[ -n "$1" ] && [ -n "$2" ] && [ -n "$3" ] && [ -n "$4" ] && [ -n "$5" ] || { echo "Usage: create_worker.sh <./worker_base_folder> <rpms_to_install.txt> <./path_to_rpms> <./containercheck> <./log_dir>"; exit; }
```

toolkit/tools/internal/buildpipeline/buildpipeline.go
dmcilvaney: PrintMessageBox is a really obvious warning to the user with big borders. I try to add these when I think there is a non-obvious thing that the user should fix that might cause confusing errors later. I think I see what you mean with the ordering though, I'll give it a go.

cwize1 (approved):

toolkit/tools/internal/buildpipeline/buildpipeline.go
```diff
@@ -8,28 +8,156 @@ package buildpipeline
...
+		isDockerContainer, err = checkIfContainerDockerEnvFile()
+		if err != nil {
+			// Log the error, but continue with the check.
+			logger.Log.Warnf("Failed to check if /.dockerenv exists: %s", err)
+		} else {
+			isRegularBuild = !isDockerContainer
+		}
+		message := []string{
+			"Failed to detect if the system is running in a container using systemd-detect-virt.",
+			err.Error(),
```
dmcilvaney, line 134: New flow can overwrite the err with nil, explicitly grab the error string.
```suggestion
		// Fallback if systemd-detect-virt isn't available.
		systemdErrMsg := err.Error()
		isDockerContainer, err = checkIfContainerDockerEnvFile()
		if err != nil {
			// Log the error, but continue with the check.
			logger.Log.Warnf("Failed to check if /.dockerenv exists: %s", err)
		} else {
			isRegularBuild = !isDockerContainer
		}
		message := []string{
			"Failed to detect if the system is running in a container using systemd-detect-virt.",
			systemdErrMsg ,
```

toolkit/tools/internal/buildpipeline/buildpipeline.go
```diff
@@ -8,28 +8,157 @@ package buildpipeline
...
+		isDockerContainer, err = checkIfContainerDockerEnvFile()
+		if err != nil {
+			// Log the error, but continue with the check.
+			logger.Log.Warnf("Failed to check if /.dockerenv exists: %s", err)
+		} else {
+			isRegularBuild = !isDockerContainer
+		}
+		message := []string{
+			"Failed to detect if the system is running in a container using systemd-detect-virt.",
+			systemdErrMsg ,
```
dmcilvaney, line 135: Fix stupid formatting from hand edited suggestion.
```suggestion
			systemdErrMsg,
```

jslobodzian (approved):

PR #11196, opened 2024-11-22
ImageCustomizer:  Add Support for Configuring User Sudoers File

elainezhao1:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- 

**NO**

###### Associated issues  

- #xxxx

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/toolkit/tools/imagecustomizerapi/user.go
+++ b/toolkit/tools/imagecustomizerapi/user.go
@@ -20,6 +20,7 @@ type User struct {
 	SecondaryGroups     []string  `yaml:"secondaryGroups"`
 	StartupCommand      string    `yaml:"startupCommand"`
 	HomeDirectory       string    `yaml:"homeDirectory"`
+	Sudo                bool      `yaml:"sudo"`
 }
 
 func (u *User) IsValid() error {
--- a/toolkit/tools/pkg/imagecustomizerlib/customizefiles.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizefiles.go
@@ -17,8 +17,8 @@ const (
 	defaultFilePermissions = 0o755
 )
 
-func copyAdditionalFiles(baseConfigPath string, additionalFiles imagecustomizerapi.AdditionalFileList,
-	imageChroot *safechroot.Chroot,
+func CopyAdditionalFiles(baseConfigPath string, additionalFiles imagecustomizerapi.AdditionalFileList,
+	imageChroot safechroot.ChrootInterface,
 ) error {
 	for _, additionalFile := range additionalFiles {
 		logger.Log.Infof("Copying: %s", additionalFile.Destination)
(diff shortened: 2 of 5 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeusers.go
```diff
@@ -137,5 +137,40 @@ func addOrUpdateUser(user imagecustomizerapi.User, baseConfigPath string, imageC
...
+	// Set user's sudo access.
+	err = UpdateSudoAccess(imageChroot, user.Name, user.Sudo)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func UpdateSudoAccess(installChroot safechroot.ChrootInterface, username string, enableSudo bool) error {
```
elainezhao96, line 149: Was trying to put this function inside userutils.py but will create cyclic dependency issue.

toolkit/tools/imagecustomizerapi/user.go
```diff
@@ -20,6 +20,7 @@ type User struct {
 	SecondaryGroups     []string  `yaml:"secondaryGroups"`
 	StartupCommand      string    `yaml:"startupCommand"`
 	HomeDirectory       string    `yaml:"homeDirectory"`
+	Sudo                bool      `yaml:"sudo"`
```
reubeno, line 23: Is this setting specifically enabling *no-password* sudo? If that's the case, then could we make the setting name appropriately specific to that? Given that this is a security-impacting setting, I think we should be as clear as we can about what the image owner is opening up.

toolkit/tools/pkg/imagecustomizerlib/customizeusers.go
```diff
@@ -137,5 +137,40 @@ func addOrUpdateUser(user imagecustomizerapi.User, baseConfigPath string, imageC
...
+		// Create the AdditionalFile object
+		additionalFiles := imagecustomizerapi.AdditionalFileList{
+			imagecustomizerapi.AdditionalFile{
+				Destination: fmt.Sprintf("/etc/sudoers.d/%s", username),
+				Content:     &content,
+				Permissions: &permissions,
+			},
+		}
+
+		err := CopyAdditionalFiles(sudoersFile, additionalFiles, installChroot)
```
cwize1, line 170: You don't need to use `CopyAdditionalFiles`. Just write to the file using [file.Write](https://github.com/microsoft/azurelinux/blob/3.0-dev/toolkit/tools/internal/file/file.go#L191). (You can use `installChroot .RootDir()` to get the rootfs directory.)

toolkit/tools/pkg/imagecustomizerlib/customizefiles.go
```diff
@@ -17,8 +17,8 @@ const (
 	defaultFilePermissions = 0o755
 )
 
-func copyAdditionalFiles(baseConfigPath string, additionalFiles imagecustomizerapi.AdditionalFileList,
-	imageChroot *safechroot.Chroot,
+func CopyAdditionalFiles(baseConfigPath string, additionalFiles imagecustomizerapi.AdditionalFileList,
```
cwize1, line 20: You don't need to make a function public if you are accessing it from the same package/directory.

PR #11247, opened 2024-11-27
ImageCustomizer: Refresh partitions after veritysetup so that hash partition's UUID is accurate

fintelia:
###### Summary 

This change corrects the value set for the `systemd.verity_root_hash=` kernel cmdline parameter when `hashDeviceMountIdType`  in the configuration is set to `uuid`.
 
In particular, the veritysetup command earlier in the function generates the UUID for the verity hash partition. But since the diskPartitions array is queried before that call, the partition would be marked as having an empty filesystem UUID.

An alternative would be to edit only the specific partition's UUID in the diskPartitions array, either by passing a manually generated UUID to veritysetup (via the `--uuid=UUID` flag) or by querying the chosen UUID after veritysetup had run.

**NO**

###### Test Methodology

- local build


liulanze (approved):

romoh (changes requested):
Please move PR to new repo [https://github.com/[microsoft/azure-linux-image-tools](https://github.com/microsoft/azure-linux-image-tools)

PR #11296, opened 2024-12-03
Add (manual) GitHub workflow to build changed specs

reubeno:
###### Summary
This is the first step toward "buddy build" validation workflows that run via GitHub actions. At present it is a manual workflow, executed via `workflow_dispatch`, and only builds changed specs and attempts to install the built packages. After its gotten some broader testing and use, I expect that the next step will be to enable it all pull requests into `3.0-dev`. Beyond that, some work may be needed to adapt it for other branches or events.

The draft PR that this came from (#11019) continues beyond this and runs and reports on package tests as well, also handling expected failures.

###### Change Log 
* Adds new 'Spec Build/Test' action (not run by default).
* Adds new helper script `detect_changes.py`, based on inline logic available in other GitHub PR checks. I further extended the logic so that it could be usable in local dev environments, both for testing and for similar functionality.
* Ports existing `test_package_install.sh` script from private repos, enabling reuse of logic to test install/uninstall of the built packages.

n/a

###### Test Methodology
Validated in my GitHub fork of this repo. For a sample run:

https://github.com/reubeno/azurelinux/actions/runs/12146627411

image

###### Merge Checklist
- [X] Ready to merge

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -1,0 +1,269 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+#
+# NOTES:
+#
+# This is a starting point for a validation workflow that can detect the specs that
+# have changed in a given git commit, build the RPMs for those specs, and run ptests
+# for those. For now, it builds the packages and attempts to validate that they
+# install correctly.
+#
+# Known limitations/concerns:
+#
+# * We build + test the packages in an Ubuntu VM runner. This is easiest in a GitHub
+#   action, since it's the only Linux OS supported running in the VM; in time,
+#   it may be worthwhile looking into running the package build itself in an
+#   Azure Linux 3.0 environment.
+#
+# * We take a specific dependency on daily repos in Azure Linux 3.0+, as well as the
+#   Azure Linux 3.0 container. To extend support to 2.0, some work will be required
+#   to make these dependencies conditional.
+#
+
+name: "Spec Build/Test"
+
+on:
+  # N.B. For now this workflow is only manually invoked. In the future we plan to enable
+  # it for pull requests and pushes.
+  workflow_dispatch:
+    inputs:
+      comparison_base_ref:
+        description: 'The base ref to compare changes against.'
+        required: true
+        default: '3.0-dev'
+
+env:
+  # For consistency, we use the same major/minor version of the latest version of Python that
+  # Azure Linux ships.
+  PYTHON_VERSION: "3.12"
+
+jobs:
+  detect-changes:
+    name: 'Detect spec changes'
+    runs-on: ubuntu-latest
+
+    outputs:
+      updated_specs: ${{ steps.export.outputs.UPDATED_SPECS }}
+      updated_extended_specs: ${{ steps.export.outputs.UPDATED_EXTENDED_SPECS }}
+      target_ref: ${{ steps.export.outputs.TARGET_REF }}
+      azl_extra_make_args: ${{ steps.export.outputs.AZL_EXTRA_MAKE_ARGS }}
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Setup Python ${{ env.PYTHON_VERSION }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Install Python dependencies for tools
+        run: python3 -m pip install -r toolkit/scripts/requirements.txt
+
+      #
+      # We first get oriented, figuring out which version of the code we're building
+      # and which version of the code we should be comparing against.
+      #
+      # We define the following:
+      #   - base_sha: The commit hash of the commit we're comparing against.
+      #   - target_ref: The ref we're building.
+      #
+
+      - name: "Find base commit (PRs)"
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          set -x
+          # In a PR workflow, 'github.base_ref' is the *target* branch of the PR.
+          git fetch origin ${{ github.base_ref }}
+          echo "base_sha=$(git rev-parse origin/${{ github.base_ref }})" >>$GITHUB_ENV
+          echo "target_ref=${{ github.base_ref }}" >>$GITHUB_ENV
+          echo "Merging ${{ github.sha }} into ${{ github.base_ref }}"
+
+      - name: "Find base commit (pushes)"
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          set -x
+          # In a push workflow, 'github.event.before' is the SHA hash of the commit
+          # immediately preceding the commit being pushed. 
+          git fetch origin ${{ github.event.before }}
+          echo "base_sha=${{ github.event.before }}" >>$GITHUB_ENV
+          echo "target_ref=${{ github.ref }}" >>$GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/check-build.yml
```diff
@@ -0,0 +1,269 @@
...
+      #   - base_sha: The commit hash of the commit we're comparing against.
+      #   - target_ref: The ref we're building.
+      #
+
+      - name: "Find base commit (PRs)"
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          set -x
+          # In a PR workflow, 'github.base_ref' is the *target* branch of the PR.
+          git fetch origin ${{ github.base_ref }}
```
0xba1a, line 78: Will it be a problem when the PR is not rebased? Shouldn't you use something like `${{ github.event.pull_request.base.sha }}` ?

.github/workflows/check-build.yml
```diff
@@ -0,0 +1,269 @@
...
+          echo "base_sha=$(git rev-parse origin/${{ github.base_ref }})" >>$GITHUB_ENV
+          echo "target_ref=${{ github.base_ref }}" >>$GITHUB_ENV
+          echo "Merging ${{ github.sha }} into ${{ github.base_ref }}"
+
+      - name: "Find base commit (pushes)"
+        if: ${{ github.event_name == 'push' }}
```
0xba1a, line 84: IMO, we don't need this trigger. There is a chance of false-positive (or the reviewer will overlook) that this action will display green, but it is only for the recent push but not for the entire PR.

.github/workflows/check-build.yml
```diff
@@ -0,0 +1,269 @@
...
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Setup Python ${{ env.PYTHON_VERSION }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Install Python dependencies for tools
+        run: python3 -m pip install -r toolkit/scripts/requirements.txt
```
0xba1a, line 62: This step is not required here as you are only detecting the changes in this job.

toolkit/scripts/test_package_install.sh
```diff
@@ -0,0 +1,199 @@
...
+# Add built packages as a local repo for tdnf
+sudo createrepo --compatibility "$RPM_DIRECTORY"
+
+repo_content=$(cat << EOF
+[builtpackages]
+name=Local Built Packages($ARCHITECTURE)
+baseurl=file://$RPM_DIRECTORY
+enabled=1
+priority=0
+skip_if_unavailable=True
```
0xba1a, line 102: Shouldn't it be `false`? Isn't there a possibility of the package being pulled from the remote?

toolkit/scripts/test_package_install.sh
```diff
@@ -0,0 +1,199 @@
...
+sudo tdnf makecache
+
+sudo tdnf repolist --refresh
+
+PACKAGES_FAIL_INSTALL=()
+PACKAGES_FAIL_UNINSTALL=()
+
+for rpm_to_install in "${RPMS_TO_INSTALL[@]}"; do
+    path_to_rpm=$(find $RPM_DIRECTORY -name "$rpm_to_install")
+    package_name_to_install=$(rpm -qp --queryformat "%{name}" "$path_to_rpm" 2>/dev/null)
```
0xba1a, line 147: Skipping the version-release_no combined with `skip_if_unavailable=True` may bring the package from remote! It's a false positive.

toolkit/scripts/test_package_install.sh
```diff
@@ -0,0 +1,199 @@
...
+done
+
+echo ""
+echo "------------------------------------------------"
+echo "-- Printing install + uninstall error details --"
+echo "------------------------------------------------"
+echo ""
+
+if [ "${#PACKAGES_FAIL_INSTALL[@]}" -gt 0 ]; then
+    echo "##vso[task.logissue type=error]Error, ${#PACKAGES_FAIL_INSTALL[@]} packages failed to install, check logs to see the list"
```
0xba1a, line 185: I think its specific to pipeline. Hope it will not cause any confusions in Actions!

.github/workflows/check-build.yml
```diff
@@ -0,0 +1,269 @@
...
+          fi
+
+      - name: Advertise build logs
+        if: always() && steps.upload-build-logs.outcome == 'success'
+        run: |
+          echo "_Build logs available [for download](${{ steps.upload-build-logs.outputs.artifact-url }})_" >>${GITHUB_STEP_SUMMARY}
+
+  install:
+    name: 'Test installing RPM'
+    needs: [build, detect-changes]
```
suresh-thelkar, line 216: Do we need to add detect-changes here since build already depends on detect-changes ?

.github/workflows/check-build.yml
```diff
@@ -0,0 +1,269 @@
...
+
+      - name: Install Python dependencies for tools
+        run: python3 -m pip install -r toolkit/scripts/requirements.txt
+
+      - name: Find daily repo
+        id: find-daily-repo
+        run: |
+          DAILY_BUILD_ID=$(toolkit/scripts/get_lkg_id.sh /tmp)
+          echo "DAILY_BUILD_ID=${DAILY_BUILD_ID}" >>$GITHUB_OUTPUT
+          echo "DAILY_BUILD_ARCH=x86-64" >>$GITHUB_OUTPUT
```
suresh-thelkar, line 161: Are we only running the build for x86-64 ?

.github/workflows/check-build.yml
Sumynwa: It will be needed in my opinion for line number 217 since we need the output from that job, which is not available from build for updated_specs/updated_extended_specs.

.github/workflows/check-build.yml
reubeno: That implies an assumption that the Python script called below has no extra requirements. It would seem more future-proof to leave this step in.

.github/workflows/check-build.yml
reubeno: There is no trigger on push for now. In the future we may choose to have a trigger on push to a tracked branch (like 3.0-dev, 3.0, etc.) and we would need this logic for then. I agree with your assessment that it would be noisy to trigger on push to the PR *source* branch.

.github/workflows/check-build.yml
reubeno: Initially, yes, but just to scope the initial change. I'd be happy for someone to contribute an extension of this to add `aarch64` support in parallel.

toolkit/scripts/detect_changes.py
```diff
@@ -0,0 +1,172 @@
...
+        if result.returncode == 0:
+            fork_point_ref = result.stdout.decode("utf-8").strip()
+            fork_points[ref] = fork_point_ref
+
+    best_distance = None
+    best_candidate_ref = None
+
+    for candidate_ancestor, fork_point in fork_points.items():
+        # Find the distance in the chain.
+        result = subprocess.run(["git", "rev-list", "--count", f"{fork_point}..{target_ref}"], capture_output=True, check=True)
```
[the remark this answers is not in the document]
reubeno: Why would there be an invalid git reference, @suresh-thelkar ? I'm sure there's something I may not have considered but an example could help.

toolkit/scripts/test_package_install.sh
reubeno: It shouldn't, will just create noise in the output.

PR #11424, opened 2024-12-12
Added distrusted CAs to the cert bundles.

PawelWMS:
###### Merge Checklist  

- [X] Ready to merge

---

###### Summary 

Windows has recently distrusted a leaf node, which is a new scenario for AZL. Adding the logic to handle that in our system.

No.

###### Test Methodology

- PR check ADO build.


SPECS/ca-certificates/ca-certificates.spec
```diff
@@ -307,6 +315,9 @@ rm -f %{pkidir}/tls/certs/*.{0,pem}
 %dir %{pkidir}/tls
 %dir %{pkidir}/tls/certs
 
+# Distrusted CAs
+%{_datadir}/pki/ca-trust-source/%{p11_format_distrusted_bundle}
```
jslobodzian, line 319: Do we need to do anything special here regarding file permissions.  We definitely would not want the "distrusted list" to be deleted by some rogue actors

PawelWMS: Good question. The whole `%{_datadir}/pki` directory and its children are read-only for non-root users. Having said that, this is probably mostly due to the fact, that we build the RPMs as root. Let me see, if I can make this more explicit.

PawelWMS: I've now explicitly set the file ownership to `root:root`, so regular users won't be able to modify these files even if we change how we build our RPMs.

jslobodzian (approved):

PR #11442, opened 2024-12-13
Fix issue when version pinning packages that are checked by the configvalidator tool

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Several features require specific packages also be installed to function (selinux, fips, etc.). One of these is `shadow-utils` which is needed for user changes.

We also have an undocumented feature where you can pin packages to specific versions in a package list via "pkg-name = 1.2.3-4". These two features conflict with each other since the validator assumed only basic package names would be used.

Also flesh out the unit tests a bit for the validator and rely less on using the built-in configs and instead use a dedicated test config file.

There is a limitation however, we cannot distinguish between `pkg-long-name` and `pkg-version-release` even though tdnf will accept that, so that will still be passed to the check verbatim and will fail. The form `pkg = version-release` should be used instead.

###### Change Log  

- Support version pinning in image configs for validated packages (selinux, fips, shadow-utils).
- Fix up the validator unit tests

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_queries/edit/55399616/?triage=true

###### Test Methodology

- local testing, including against customer's config
- pipeline https://dev.azure.com/mariner-org/mariner/_build/results?buildId=695490&view=results

### DIFF
--- a/toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
+++ b/toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
@@ -15,6 +15,7 @@ import (
 	"github.com/microsoft/azurelinux/toolkit/tools/imagegen/installutils"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/exe"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/pkgjson"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/timestamp"
 	"github.com/microsoft/azurelinux/toolkit/tools/pkg/profile"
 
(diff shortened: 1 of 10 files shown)


toolkit/tools/imageconfigvalidator/imageconfigvalidator.go
```diff
@@ -134,16 +136,24 @@ func validatePackages(config configuration.Config) (err error) {
 		}
 
 		for _, pkg := range packageList {
-			if pkg == "kernel" {
+			// The installer tools have an undocumented feature which can support both "pkg-name" and "pkg-name=version" formats.
+			// This is in use, so we need to handle pinned versions in this check. Technically, 'tdnf' also supports "pkg-name-version" format,
+			// but it is not easily distinguishable from "long-package-name" format so it will not be supported here.
+			pkgVer, err := pkgjson.PackageStringToPackageVer(pkg)
+			if err != nil {
+				return fmt.Errorf("%s: %w", validateError, err)
```
dmcilvaney, line 144: Validator has always used old-form error wrapping, just keeping consistent here.

reubeno:
Changes seem reasonable to me; have you been able to confirm that they do, in fact, now permit the config used by the bug finder?

Also, thanks for taking the time to improve the tests while you were at it. Always love to see us leaving the code better than we found it!

dmcilvaney:
> Changes seem reasonable to me; have you been able to confirm that they do, in fact, now permit the config used by the bug finder?
> 
> Also, thanks for taking the time to improve the tests while you were at it. Always love to see us leaving the code better than we found it!

The specific branch that broke is gone, but I was able to dig up a similar config. It looks like it should pass fine now.

dmcilvaney:
Queued a delta build to check that all the built-in images build ok: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=695490&view=results

jslobodzian (approved):

PR #11479, opened 2024-12-17
Bphilip/add mofed and dependencies

binujp:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR brings in MOFED and dependent packages required to build our AI/ML distribution story. Immediately HPC images require infiniband for GPU-clustering. MOFED is thus a core requirement. There are first and third party requirments for MOFED drivers directly and dependent packages.

CUDA and GdrCopy drivers need MOFED as a build-time and runtime requirements. This is not satisfied today and CUDA driver is built without MOFED support. These packages getting integrated into core will enable that work stream to be unblocked as well.

###### Change Log  
There are 18 new packages being added by this change to the SPECS directory. There are 8 new SIGNED-SPECS being added too. They are either directly required to build out MOFED and support infrastructure or are tools required to provide a complete infiniband solution like tools to configure the hw/driver and debug issues.

fwctl
ibarr
ibsim
iser
isert
knem
mft_kernel
mlnx-ethtool
mlnx-iproute2
mlnx-nfsrdma
mlnx-ofa_kernel
mlx-steering-dump
multiperf                                                                                                                                                                                                       rshim
sockperf
srp
xpmem-lib
xpmem

No

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=709280&view=results

The RPMs from a local build of these specs were installed on A100, V100 and T4 GPU based VMs and functionality verified.


LICENSES-AND-NOTICES/SPECS/data/licenses.json
```diff
@@ -2453,12 +2453,43 @@
...
+                "fwctl-signed",
+                "ibarr",
+                "ibsim",
+                "iser",
+                "iser-signed",
+                "isert",
+                "isert-signed",
+                "knem",
+                "knem-modules-signed",
                 "libnvidia-container",
```
ellie-di, line 2466: is this introduced in this PR?

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,143 @@
...
+%global target_kernel_release %(/bin/rpm -q --queryformat '%{RPMTAG_RELEASE}' $(/bin/rpm -q --whatprovides kernel-headers) | /bin/cut -d . -f 1)
+
+%global KVERSION %{target_kernel_version_full}
+%global K_SRC /lib/modules/%{target_kernel_version_full}/build
+
+# KMP is disabled by default
+%{!?KMP: %global KMP 0}
+
+# take kernel version or default to uname -r
+# %{!?KVERSION: %global KVERSION %(uname -r)}
```
gjswalling, line 48: if we are not getting kernel version from uname, can we drop this commented-out line and clarify the comment to something like, "take kernel version from this Azl build"?

SPECS/ibsim/ibsim.spec
```diff
@@ -0,0 +1,54 @@
...
+%{_bindir}/ibsim-run
+%doc README COPYING TODO net-examples scripts
+
+%changelog
+* Tue Dec  17 2024 Binu Jose Philip <bphilip@microsoft.com>
+- Moving to core from azlinux-ai-ml repo
+- Initial Azure Linux import from NVIDIA (license: GPLv2)
+- License verified
+* Thu Nov 07 2024 Suresh Babu Chalamalasetty <schalam@microsoft.com>
+- Initial version Azure Linux
```
PawelWMS, line 54: Please use only one source attribution "Initial..." line. Right now it's unclear, where we got the spec from: Nvidia or did Suresh write these himself.

SPECS/iser/iser.spec
```diff
@@ -0,0 +1,253 @@
...
+#    copy of which is available from the Open Source Initiative, see
+#    http://www.opensource.org/licenses/gpl-license.php.
+#
+# Licensee has the right to choose one of the above licenses.
+#
+# Redistributions of source code must retain the above copyright
+# notice and one of the license notices.
+#
+# Redistributions in binary form must reproduce both the above copyright
+# notice, one of the license notices in the documentation
```
PawelWMS, line 24: I'm not sure how to interpret this licensing comment and the source attributions (we mention both Nvidia and claim we've done this ourselves at the same time). I think we need more clarify on the licensing of the new specs.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,143 @@
...
+fwctl signed kernel modules
+
+%prep
+
+%build
+rpm2cpio %{Source0} | cpio -idmv -D %{buildroot}
+
+%install
+cp %{Source1} %{buildroot}/lib/modules/%{KVERSION}/updates/fwctl/fwctl.ko
+cp %{Source2} %{buildroot}/lib/modules/%{KVERSION}/updates/fwctl/mlx5/mlx5_fwctl.ko
```
gjswalling, line 118: kernel modules are generally installed with 644 (execute bit off), for example in [kernel](https://github.com/microsoft/azurelinux/blob/2a17e481fe1b06e3e7980d5d4d32d8a15ad97879/SPECS/kernel/kernel.spec#L285), so you could consider using `install -D -m 644 %{Source2} %{buildroot}/lib/modules/%{KVERSION}/updates/fwctl/mlx5/mlx5_fwctl.ko

SPECS/fwctl/fwctl.spec
```diff
@@ -0,0 +1,256 @@
...
+%if "%{_vendor}" == "suse"
+%debug_package
+%endif
+
+%if 0%{?anolis} == 8
+%global __find_requires %{nil}
+%endif
+
+# set modules dir
+%if "%{_vendor}" == "redhat" || ("%{_vendor}" == "openEuler")
```
gjswalling, line 150: seems like there are several vendor-specific lines in this spec.  Since we are setting vendor to Microsoft, can we remove the irrelevant condtions and code to simplify?

SPECS/iser/iser.spec
binujp: I took this line 
> Licensee has the right to choose one of the above licenses.
as the guidance. Does that work?

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
binujp: Sure, that makes sense. In general there is spec cleanup that can be done.

SPECS/ibsim/ibsim.spec
binujp: Got it, I will clean that up in all the specs.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
binujp: Yes, I will fix that here and check if that applies to any of the other modules.

SPECS/fwctl/fwctl.spec
binujp: Sure, some more spec cleanup which applies to few other specs as well. I had tried to keep the spec as similar to what Nvidia had published. It is already diverging so I will make these changes.

gjswalling: easier to maintain if we leave these in case we pull an update from fedora in the future

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
gjswalling: Suresh explained that default is 644, so cp is probably fine

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: fixed.

SPECS/ibsim/ibsim.spec
chalamalasetty: fixed

gjswalling (approved):

SPECS/iser/iser.spec
chalamalasetty: Added license copyright files as needed for SPECS fwctl, iser, mofed kernel, srp etc.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
...
+%else
+%global _release1 %{_release}
+%endif
+%global _kmp_rel %{_release1}%{?_kmp_build_num}%{?_dist}
+
+Summary:	 %{_name} Driver
+Name:		 %{_name}
+Version:	 %{_version}
+Release:	 1%{?dist}
+License:	 GPLv2
```
jslobodzian, line 66: Just to be sure, the SPEC file itself has it's own license, and I believe the header comment in this file says we have a choice between GPLv2 OR BSD.  However, I want to be sure that doesn't extend to the fwctl source itself.  If so, then the License: section should explicitly say GPLv2 OR BSD

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
...
+%description
+fwctl signed kernel modules
+
+%prep
+
+%build
+rpm2cpio %{Source0} | cpio -idmv -D %{buildroot}
+
+%install
+cp %{Source1} %{buildroot}/lib/modules/%{KVERSION}/updates/fwctl/fwctl.ko
```
jslobodzian, line 116: where is %license for this?

And...if we get to choose the license, we should be choosing BSD.

SPECS-SIGNED/iser-signed/iser-signed.spec
```diff
@@ -0,0 +1,138 @@
...
+%else
+%global _release1 %{_release}
+%endif
+%global _kmp_rel %{_release1}%{?_kmp_build_num}%{?_dist}
+
+Summary:	 %{_name} Driver
+Name:		 %{_name}
+Version:	 %{_version}
+Release:	 1%{?dist}
+License:	 GPLv2
```
jslobodzian, line 64: comment above suggests we get to choose the license.  I think we would want to prefer BSD.
(Similar comment as earlier one...I can't tell if this LICENSE header in the commented section applies to the code all up or the SPEC file only...I believe it's all-up)

SPECS-SIGNED/iser-signed/iser-signed.spec
```diff
@@ -0,0 +1,138 @@
...
+%post
+if [ $1 -ge 1 ]; then # 1 : This package is being installed or reinstalled
+  /sbin/depmod %{KVERSION}
+fi # 1 : closed
+# END of post
+
+%postun
+/sbin/depmod %{KVERSION}
+
+%if "%{KMP}" != "1"
```
jslobodzian, line 127: where's the %license?

SPECS-SIGNED/isert-signed/isert-signed.spec
```diff
@@ -0,0 +1,138 @@
...
+%else
+%global _release1 %{_release}
+%endif
+%global _kmp_rel %{_release1}%{?_kmp_build_num}%{?_dist}
+
+Summary:	 %{_name} Driver
+Name:		 %{_name}
+Version:	 %{_version}
+Release:	 1%{?dist}
+License:	 GPLv2
```
jslobodzian, line 64: Same comment.  We should prefer BSD

SPECS-SIGNED/isert-signed/isert-signed.spec
```diff
@@ -0,0 +1,138 @@
...
+%postun
+/sbin/depmod %{KVERSION}
+
+%if "%{KMP}" != "1"
+%files
+%defattr(-,root,root,-)
+/lib/modules/%{KVERSION}/%{install_mod_dir}/
+%config(noreplace) %{_sysconfdir}/depmod.d/zz02-%{name}-*.conf
+%endif
+
```
jslobodzian, line 133: where is %license?

SPECS-SIGNED/knem-signed/knem-modules-signed.spec
```diff
@@ -0,0 +1,127 @@
...
+%postun
+if [ $1 = 0 ]; then  # 1 : Erase, not upgrade
+    depmod %{KVERSION} -a
+fi
+%endif # end KMP=1
+
+%if "%{KMP}" != "1"
+%files
+/lib/modules/%{KVERSION}/%{install_mod_dir}/
+%endif
```
jslobodzian, line 121: where is %license

jslobodzian (approved):

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
...
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+# BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+# ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+
+%{!?_name: %define _name fwctl}
+%{!?_version: %define _version 24.10}
+%{!?_release: %define _release OFED.24.10.0.6.7.1}
```
PawelWMS, line 35: Why define these `_version` separately? Won't the `%{version}` work? Similar question for `_name`. And what's the logic behind `_release`? I see it trickles down to `_kmp_rel` for instance but I don't see `_kmp_rel` used anywhere?

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
...
+# define release version
+%{!?src_release: %global src_release %{_release}_%{krelver}}
+%if "%{KMP}" != "1"
+%global _release1 %{src_release}
+%else
+%global _release1 %{_release}
+%endif
+%global _kmp_rel %{_release1}%{?_kmp_build_num}%{?_dist}
+
+Summary:	 %{_name} Driver
```
PawelWMS, line 62: With the specs done, please also update the `version_release_matching_groups` inside `toolkit/scripts/check_entangled_specs.py`. This is part of the PR check, which makes sure "entangled" specs get updated in lockstep within one PR. In case of (regular, signed) spec pairs, if you only update the regular one and forget about this one here, we'd end up either publishing a newer version without the signed `.ko` files or not publishing the new version at all (I'd need to check the details).

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: This version is defined from MOFED SPECS for alignment. Hence keeping it same for furture updates.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: Sign pipeline changes are coming from another PR. Is this can be done with signing pipeline changes for signed SPECs?

toolkit/scripts/spec_source_attributions.py
```diff
@@ -17,6 +17,7 @@
     "Fedora (ISC)":                   r'\n-\s+Initial (CBL-Mariner|Azure Linux) import from Fedora \d+ \(license: ISC\)(\.|\n|$)',
     "Magnus Edenhill Open Source":    r'\n-\s+Initial (CBL-Mariner|Azure Linux) import from Magnus Edenhill Open Source \(license: BSD\)(\.|\n|$)',
     "NVIDIA":                         r'\n-\s+Initial (CBL-Mariner|Azure Linux) import from NVIDIA \(license: (ASL 2\.0|GPLv2)\)(\.|\n|$)',
+    "NVIDIA (BSD)":                   r'\n-\s+Initial (CBL-Mariner|Azure Linux) import from NVIDIA \(BSD\) \(license: (BSD)\)(\.|\n|$)',
```
PawelWMS, line 20: No need for "CBL-Mariner"` in the new source attributions. The other ones are to support the old specs. And there's no need to repeat "BSD", I think.
```suggestion
    "NVIDIA (BSD)":                   r'\n-\s+Initial Azure Linux import from NVIDIA \(license: (BSD)\)(\.|\n|$)',
```

A bigger question, perhaps: why add a new entry specially for BSD? I see the older one for Nvidia already covers two licenses. Is there a reason we don't want to just update that one?

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
...
+
+BuildRequires:  gcc
+BuildRequires:  make
+BuildRequires:  kernel-devel = %{target_kernel_version_full}
+BuildRequires:  kernel-headers = %{target_kernel_version_full}
+BuildRequires:  binutils
+BuildRequires:  systemd
+BuildRequires:  kmod
+BuildRequires:  mlnx-ofa_kernel-devel = %{_version}
+BuildRequires:  mlnx-ofa_kernel-source = %{_version}
```
PawelWMS, line 100: General comment for all signed specs: they don't need most `BuildRequires`, since they only expand already built RPMs and replace a few files. The packages providing `rpm2cpio` and `cpio` are part of the default build chroot but to be on the safe side, I'd add them to `BuildRequires`.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
...
+
+Requires:       mlnx-ofa_kernel = %{_version}
+Requires:       mlnx-ofa_kernel-modules  = %{_version}
+Requires:       kernel = %{target_kernel_version_full}
+Requires:       kmod
+
+%description
+fwctl signed kernel modules
```
PawelWMS, line 108: This is the description, that the published package will have, so I'd suggest something, that gives users more information about the package.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
...
+License:	 GPLv2
+Url:		 http://nvidia.com
+Group:		 System Environment/Base
+
+# This package's "version" and "release" must reflect the unsigned version that
+# was signed.
+# An important consequence is that when making a change to this package, the
+# unsigned version/release must be increased to keep the two versions consistent.
+# Ideally though, this spec will not change much or at all, so the version will
+# just track the unsigned package's version/release.
```
PawelWMS, line 75: No need for the comment - the specs entanglement PR check will handle that for you. Just make sure to update the set of spec pairs, which need to be kept in sync.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
...
+#   1. Build the unsigned packages as normal
+#   2. Sign the desired binary
+#   3. Place the unsigned package and signed binary in this spec's folder
+#   4. Build this spec
+
+Source0:        %{name}-%{version}-%{release}.%{_arch}.rpm
+Source1:        fwctl.ko
+Source2:        mlx5_fwctl.ko
+
+BuildRoot:	/var/tmp/%{name}-%{version}-build
```
PawelWMS, line 87: Is there a need to define `BuildRoot` explicitly here instead of using the default?

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
...
+%post
+if [ $1 -ge 1 ]; then # 1 : This package is being installed or reinstalled
+  /sbin/depmod %{KVERSION}
+fi # 1 : closed
+# END of post
+
+%postun
+/sbin/depmod %{KVERSION}
+
+%if "%{KMP}" != "1"
```
PawelWMS, line 131: I'm not sure `KMP` makes much sense in the signed spec, since its only purpose is to repackage the unsigned RPM. Also, with the way it's written right now, if `KMP` ever was equal to 1, we'd end up with an invalid spec file (no `%files` section).

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -0,0 +1,142 @@
+#
```
PawelWMS, line 1: I've added a bunch of comments into this spec but they apply to most/all other signed spec files.

SPECS/fwctl/fwctl.spec
```diff
@@ -0,0 +1,255 @@
...
+%global _kmp_rel %{_release1}%{?_kmp_build_num}%{?_dist}
+
+Summary:	 %{_name} Driver
+Name:		 fwctl
+Version:	 24.10
+Release:	 1%{?dist}
+License:	 GPLv2
+Url:		 http://nvidia.com
+Group:		 System Environment/Base
+Source0:         https://linux.mellanox.com/public/repo/mlnx_ofed/24.10-0.7.0.0/SRPMS/fwctl-24.10.tgz#/%{_name}-%{_version}.tgz
```
PawelWMS, line 72: Please also use macros to parameterize the URL, so that it updates automatically with the macros.

SPECS/knem/knem.spec
```diff
@@ -0,0 +1,288 @@
...
+
+%if "%{KMP}" != "1"
+%files -n %{non_kmp_pname}
+/lib/modules/%{KVERSION}/%{install_mod_dir}/
+%if %{IS_RHEL_VENDOR}
+%if ! 0%{?fedora}
+%config(noreplace) %{_sysconfdir}/depmod.d/%{_name}.conf
+%endif
+%endif
+%endif
```
PawelWMS, line 279: Missing `%license`. Not strictly necessary when there's a signed spec counterpart to this one but consistency helps.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: Fixed.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: Added version_release_matching_groups for MOFED unsigned and signed SPECS.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: Fixed

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: Description is copied from unsigned SPECS. Both unsigned and signed SPECS have same package description.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: Removed.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: Fixed .

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: Fixed.

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
chalamalasetty: Fixed for all signed SPECS.

SPECS/knem/knem.spec
chalamalasetty: Added license file.

toolkit/scripts/spec_source_attributions.py
chalamalasetty: Fixed. Added BSD to default Nvidia entry.

SPECS/xpmem/xpmem.spec
```diff
@@ -0,0 +1,249 @@
...
+
+%if ! %{with kernel_only}
+%files -n libxpmem
+%{_libdir}/libxpmem.so.*
+
+%files -n libxpmem-devel
+%{_prefix}/include/xpmem.h
+%{_libdir}/libxpmem.a
+%{_libdir}/libxpmem.so
+%{_libdir}/pkgconfig/cray-xpmem.pc
```
PawelWMS, line 236: These two subpackages would need `%license` as well. It's because they can be installed separately from the default `xpmem` package (e.g. they don't `Require: %{name} = %{version}-%{release}`), so it means that it's possible to install a package from this spec without a license file.

To clarify: it is completely fine for a subpackage to not require the default package inside one spec. I'm only making sure we end up with license info regardless of which of the packages gets installed.

Please also verify if the subpackages from other specs don't have a similar problem.

chalamalasetty: fixed and pushed changes.

PawelWMS (approved):

PR #11986, opened 2025-01-17
Dracut: Update Dracut overlay module to correctly locate 'chcon'

gmileka:
###### Merge Checklist  

- [n/a] The toolchain has been rebuilt successfully (or no changes were made to it)
- [n/a] The toolchain/worker package manifests are up-to-date
- [ok] Any updated packages successfully build (or no packages were changed)
- [n/a] Packages depending on static components modified in this PR (Golang, `*-static` subpackages, etc.) have had their `Release` tag incremented.
- [ok] Package tests (%check section) have been verified with RUN_CHECK=y for existing SPEC files, or added to new SPEC files
- [ok] All package sources are available
- [ok] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`, `.github/workflows/cgmanifest.json`)
- [ok] LICENSE-MAP files are up-to-date (`./LICENSES-AND-NOTICES/SPECS/data/licenses.json`, `./LICENSES-AND-NOTICES/SPECS/LICENSES-MAP.md`, `./LICENSES-AND-NOTICES/SPECS/LICENSE-EXCEPTIONS.PHOTON`)
- [ok] All source files have up-to-date hashes in the `*.signatures.json` files
- [ok] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [n/a] Documentation has been updated to match any changes to the build system
- [ok] Ready to merge

---

###### Summary 

LiveOS ISO utilizes an overlay file system that consists of a read-only full OS rootfs (lower dir) and an in-memory read-write file system (upper dir).

When selinux is enabled, the final root (`/`) must be labeled `root_t`. Given that the final root (`/`) is constructed from the lower dir and the upper dir, both should also be marked as such (i.e. with `root_t`). The read-only full OS rootfs is already labeled as such during build time. The in-memory file system is new, however, and still need to be labeled.

A fix was merged in December to achieve that by calling selinux `chcon` utility. At the time, the labeling was done correctly, and selinux did not report any violations in our private testing.

However, this bug report alerted us that `chcon` is not being found on the path as tested by Brian, and as a result, the labeling is not taking place, and the liveos iso does not boot when selinux is enabled.

This fix changes the dracut module responsible for the labeling to explicitly specify the full path of the `chcon` utility. This guarantees that regardless of the contents of the initrd, the command always succeeds.

When doing root cause analysis for the regression, the most likely reason it did not not show in the private testing is that the initrd had a copy of `chcon`, and it was on the path. This hid the problem that would arise if the contents of the initrd changed.

###### Change Log  

- [Unblock SELinux for ISO - Update Dracut module to correctly locate 'chcon'](https://github.com/microsoft/azurelinux/pull/11986/commits/f482913d595ee3cb3a7a5b4d10075ff2a06d3c82)

- NO

###### Associated issues  

- n/a

###### Links to CVEs  
- n/a

###### Test Methodology

**Validation**
- Locally, on dev box:
  - New enlistment of Azure Linux.
  - Build the dracut package.
  - Build the baremela image.
  - Build Prism.
  - Use Prism to build a LiveOS ISO.
  - Create a VM with the generated ISO.
  - Boot the VM.
  - Log-in.
  - Verify that no selinux denials exist in the journal logs.
    
**Automation**

- Enabled the VM tests suite in the Prism build/test pipelines.
  - Download varius Azure Linux 2.0/3.0 images.
  - Customize using Prism.
  - Ensure they all boot, get an IP, and the test can ssh and inspect the contents.
 
**Note** While we've put in automation to catch this kind of failures in the future, the downloaded images in the test are the public one which do not have the fixed Dracut package. However, once the Dracut package is published, the test will automatically pick it up and verify it boots.

### DIFF
--- a/SPECS/dracut/90overlayfs/azl-configure-selinux.sh
+++ b/SPECS/dracut/90overlayfs/azl-configure-selinux.sh
@@ -28,7 +28,8 @@ if [ -n "$overlayfs" ]; then
     echo "root folder label  : ($contextType)"
 
     # Set the type on the target folders
-    [ -e /sysroot ] && chcon -t $contextType /sysroot
-    [ -e /run/overlayfs ] && chcon -t $contextType /run/overlayfs
-    [ -e /run/ovlwork ] && chcon -t $contextType /run/ovlwork
+    CHCON_TOOL=$NEWROOT/usr/bin/chcon
+    [ -e /sysroot ] && $CHCON_TOOL -t $contextType /sysroot
+    [ -e /run/overlayfs ] && $CHCON_TOOL -t $contextType /run/overlayfs
+    [ -e /run/ovlwork ] && $CHCON_TOOL -t $contextType /run/ovlwork
 fi
--- a/SPECS/dracut/dracut.signatures.json
+++ b/SPECS/dracut/dracut.signatures.json
@@ -6,7 +6,7 @@
   "00-vrf.conf": "e2885a4b090d8ca3771e60ce6dcd8b849e28ce5002a5c7b71ff796a92deb2810",
   "00-xen.conf": "8b7a89b7716cb40a9c0d681caed6994d81ff4dfad4fe50cea15cd47b885dc5a6",
   "50-noxattr.conf": "61d95f05890ac6ee3355d0a386dd5645d82b7a4202d90305d997fd18c6d139dd",
-  "azl-configure-selinux.sh": "5f526509910fccdc2dffad4ef5070740847195510e3faefff39b831c9d28a439",
+  "azl-configure-selinux.sh": "a32da2e920348a5f991d6c418873a0a761f6e4bd926bb17a43cdb2c28b4f21d6",
   "azl-liveos-artifacts-download.service": "888be8c82297cccd510d7f963611c2360ae67559826b2b474da6d9935237de64",
   "azl-liveos-artifacts-download.sh": "f21dc68de8c81d8a8128e7a9d7be45d25978f0b5e47a4cf1a2d97b1e171ec045",
   "dracut-102.tar.gz": "601b175cbf4d2ee902bb7bda3af8826ae2ca060c1af880f6da5a833413f4ec70",
--- a/SPECS/dracut/dracut.spec
+++ b/SPECS/dracut/dracut.spec
@@ -4,7 +4,7 @@
 Summary:        dracut to create initramfs
 Name:           dracut
 Version:        102
-Release:        10%{?dist}
+Release:        11%{?dist}
 # The entire source code is GPLv2+
 # except install/* which is LGPLv2+
 License:        GPLv2+ AND LGPLv2+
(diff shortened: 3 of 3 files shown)


SPECS/selinux-policy/0038-enable-liveos-iso-flow.patch
```diff
@@ -131,6 +132,29 @@ index ea08036b6..59c1afeaf 100644
...
++
+ manage_dirs_pattern(sshd_t, sshd_tmp_t, sshd_tmp_t)
+ manage_files_pattern(sshd_t, sshd_tmp_t, sshd_tmp_t)
+ manage_sock_files_pattern(sshd_t, sshd_tmp_t, sshd_tmp_t)
+@@ -343,6 +345,9 @@ allow ssh_keygen_t self:unix_stream_socket create_stream_socket_perms;
+ allow ssh_keygen_t sshd_key_t:file manage_file_perms;
+ files_etc_filetrans(ssh_keygen_t, sshd_key_t, file)
+ 
++allow ssh_keygen_t var_t:dir { write add_name remove_name };
++allow ssh_keygen_t var_t:file { getattr setattr create open read write rename };
```
pebenito, line 153: `var_t` typically isn't used for files, since it is extremely generic, policy-wise. What dir(s)/file(s) are created?

gmileka: Looked into it further, it turns out we're moving the ssh key file location from /etc/ to /var/etc:
```
/var/etc/ssh/ssh_host_rsa_key.pub
/var/etc/ssh/ssh_host_rsa_key
/var/etc/ssh/ssh_host_ecdsa_key.pub
/var/etc/ssh/ssh_host_ecdsa_key
/var/etc/ssh/ssh_host_ed25519_key.pub
/var/etc/ssh/ssh_host_ed25519_key
```

This change is necessary for when verity is enabled and the etc location is read-only with a volatile overlay.

I'll drop the policy changes from the PR based on that. Thanks Chris.

SPECS/dracut/90overlayfs/azl-configure-selinux.sh
```diff
@@ -28,7 +28,8 @@ if [ -n "$overlayfs" ]; then
     echo "root folder label  : ($contextType)"
 
     # Set the type on the target folders
-    [ -e /sysroot ] && chcon -t $contextType /sysroot
-    [ -e /run/overlayfs ] && chcon -t $contextType /run/overlayfs
-    [ -e /run/ovlwork ] && chcon -t $contextType /run/ovlwork
+    CHCON_TOOL=$NEWROOT/usr/bin/chcon
```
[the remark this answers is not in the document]
gmileka: The setup in this flow is that:
- initrd is the first (minimal) file system that gets attached to the kernel to run the essential services.
  - so, if we run `chcon` now, it'll look-up the path, and if it's in initrd, it'll succeed.
- `$NEWROOT` points to the final full OS file system mount that will become the lower layer of the overlay fs. At this point, this file system is not yet the root - and is just mounted at $NEWROOT. Later, the kernel will be told to 'pivot' to it.

Given that SELinux is enabled on the kernel command-line (the script checks for that earlier in the code @line 10), then, the full file system should have the minimal set of selinux tools - including `chcon`.

Camelron (approved):

romoh (approved):
LGTM

PR #12842, opened 2025-03-07
bug: fix remote build dependency variability

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

**NOTE:** This is targeting 3.0 intentionally for testing, will re-direct to 3.0-dev before merging.

Previous fix caused issues in the scheduler (with a more constrained matching, we were more likely to pick different versions of a package for each of the build nodes for a SRPM, leading to weird double builds etc.).

Now, we require an exact match for remote dependencies. This will lead to more remote nodes in the graph (for example `golang`,`golang<1.23.0`,`golang>=1.17` etc.) In theory `tdnf` should de-duplicate many of these for us, but this will make the fetcher slightly slower overall. By preserving all these nodes, we can ensure two things:
1. Each package build gets build dependencies that are guaranteed to meet its requirements
2. Every build node for a given SRPM will have identical dependencies, so they will unblock together.

~Fix a bug in the version satisfaction algorithm which tries to match requested dependencies to already existing nodes in the graph. In this case, it was incorrectly deciding that the set `[1,5]` could always satisfy the new dependency `[2,3]`. While initially this seems reasonable, if the only package available turns out to be `version=4`, the second condition would be violated, but since the node was de-duplicated this knowledge is lost until an inevitable build failure much later.~

###### Change Log  

- Change `grapher` to always use exact version matching for remote dependencies.
- ~Replace `interval.LowerBound.Compare(<VER>) == 0` with `interval.LowerBound.EQ(<VER>)))` etc. to simplify the code and make understanding the algorithm a bit easier.~
- ~Make `version.Compare()` private (use the `GT(), LT(), EQ()...` functions instead)~
- ~`interval.versionInInterval()` and `interval.Satisfies()` are replaced with simplified functions:~
  - ~`interval.Satisfies()`~
  - ~`interval.GuaranteedSatisfies()`~
  - ~`interval.intersects()`~
  - ~`interval.contains()`~
- ~Remove `CompareWithConditional()` which was dead code~
- ~Add fuzzer and 2nd implementation of the satisfaction test to try and gain confidence with interval function~

**YES/NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_queries/edit/56521875/?triage=true

###### Test Methodology

- ~Full build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=755615&view=results~
- ~Dummy prod build with same config that broke prior: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=755649&view=results~
- dummy prod: ~https://dev.azure.com/mariner-org/mariner/_build/results?buildId=758625&view=results~ https://dev.azure.com/mariner-org/mariner/_build/results?buildId=761159&view=results
- full: ~https://dev.azure.com/mariner-org/mariner/_build/results?buildId=758626&view=results~ https://dev.azure.com/mariner-org/mariner/_build/results?buildId=761160&view=results
- full no delta: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=762983&view=results


dmcilvaney:
Thoughts on the rework of the compare? I did it for my own sanity reading the code, but its easy to revert back to the old stuff now that the fix is done.

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -339,78 +338,56 @@ func (interval *PackageVerInterval) Compare(other *PackageVerInterval) (result i
...
+// GuaranteedSatisfies returns true if every version in the interval is guaranteed to satisfy the query. A pair of
+// intervals may overlap, but not guarantee that the query interval is satisfied.
+// i.e., interval: [1, 3], query: [2, 4] will overlap, but the valid version '1' does not satisfy the query.
+func (interval *PackageVerInterval) GuaranteedSatisfies(queryInterval *PackageVerInterval) (valid bool) {
+	return queryInterval.contains(interval)
+}
 
-	return queryUpperValid || queryLowerValid || superset
+// Satisfies returns true if there exists some version in the interval that satisfies the query interval.
+func (interval *PackageVerInterval) Satisfies(queryInterval *PackageVerInterval) (valid bool) {
```
PawelWMS, line 389: The `Satisfies` vs `GuaranteedSatisfies` naming feel a bit confusing. The mathematical terms you're using in the private functions seem more descriptive. What do you think about using `Contains` and `Intersects`/`Overlaps` directly? In the former case I guess we'd need to reverse the order of how we call `GuaranteedSatisfies` now but it feel like a small price to pay for clarity. Especially since this piece of code needs extra precision to avoid making a mistake.

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -600,11 +620,20 @@ func TestConditionalLookupBasic(t *testing.T) {
...
 
-	// Best is D2, which is ver: <=2
+	// Best is D2, which is ver: <=2, BUT since we can't predict the future its not safe to use it
+	lu, err = g.FindBestPkgNode(&pkgjson.PackageVer{Name: "D", Version: "1"})
+	assert.NoError(t, err)
+	assert.Nil(t, lu)
+
+	// If we add a new special exact version of D, we should get that one
+	pkgDExactly1 := pkgjson.PackageVer{Name: "D", Version: "1", Condition: "="}
+	pkgpkgDExactly1Unresolved := buildUnresolvedNodeHelper(&pkgDExactly1)
```
PawelWMS, line 630: Small request since we're here: can we remove "Helper" from all helper functions? I keep thinking we're building helpers instead of nodes and I think that's unnecessarily confusing.

PawelWMS, line 630: To be consistent with how this test is written, I think we should make this new node part of `allNodes`. Despite how I dislike how this test is written in general.:)

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -600,11 +620,20 @@ func TestConditionalLookupBasic(t *testing.T) {
 	assert.NotNil(t, lu)
 	assert.True(t, lu.RunNode.Equal(pkgCRun))
 
-	// Best is D2, which is ver: <=2
+	// Best is D2, which is ver: <=2, BUT since we can't predict the future its not safe to use it
```
PawelWMS, line 623: What is wrong with picking `D2` in this case? I didn't understand the bit about predicting the future, I think.

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -600,11 +620,20 @@ func TestConditionalLookupBasic(t *testing.T) {
...
+	lu, err = g.FindBestPkgNode(&pkgjson.PackageVer{Name: "D", Version: "1"})
+	assert.NoError(t, err)
+	assert.Nil(t, lu)
+
+	// If we add a new special exact version of D, we should get that one
+	pkgDExactly1 := pkgjson.PackageVer{Name: "D", Version: "1", Condition: "="}
+	pkgpkgDExactly1Unresolved := buildUnresolvedNodeHelper(&pkgDExactly1)
+	require.NoError(t, addNodesHelper(g, []*PkgNode{pkgpkgDExactly1Unresolved}))
```
PawelWMS, line 631: Suggestion to distinguish version from package representations:
```suggestion
	pkgDVerExactly1 := pkgjson.PackageVer{Name: "D", Version: "1", Condition: "="}
	pkgDExactly1Unresolved := buildUnresolvedNodeHelper(&pkgDVerExactly1)
	require.NoError(t, addNodesHelper(g, []*PkgNode{pkgDExactly1Unresolved}))
```

toolkit/tools/internal/pkggraph/pkggraph_test.go
PawelWMS, line 623: Typo nit-pick:
```suggestion
	// Best is D2, which is ver: <=2, BUT since we can't predict the future it's not safe to use it
```

toolkit/tools/internal/pkggraph/pkggraph_test.go
```diff
@@ -615,35 +644,37 @@ func TestConditionalLookupBasic(t *testing.T) {
 func TestConditionalLookupMulti(t *testing.T) {
 	g := NewPkgGraph()
 	err := addNodesHelper(g, unresolvedNodes)
-	assert.NoError(t, err)
-	assert.NotNil(t, g)
+	require.NoError(t, err)
+	require.NotNil(t, g)
 
-	// Should get D2 (<= 2)
+	// Should get nothing, closest is D2 (<= 2), but we have no garantuee that it will be >=1 when resolved
```
PawelWMS, line 650: I realized I'm not sure what a package with version `<=` means. Is it the "provides" from this package? Or maybe it's just invalid to have a package with a version anything else than `=`? If it's a "provides", then `<= 2` would mean that it guarantees to provide version 2 of the package, so this should be satisfactory for a `>= 1` requirement.

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -153,7 +153,7 @@ func (pkgVer *PackageVer) Interval() (interval PackageVerInterval, err error) {
 		}
 	case pkgVer.Version != "" && pkgVer.SVersion != "":
 		// Explicit version information for both (duplicate version data is handled above)
-		if v1.Compare(v2) == versioncompare.LessThan {
+		if v1.LT(v2) {
```
PawelWMS, line 156: Praise you for getting rid of the `== versioncompare.X` nonsense!

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -339,78 +338,56 @@ func (interval *PackageVerInterval) Compare(other *PackageVerInterval) (result i
...
+// Contains checks if the interval fully contains the query range
+func (interval *PackageVerInterval) contains(queryInterval *PackageVerInterval) (contains bool) {
+	lowerBoundValid := false
+	upperBoundValid := false
+
+	// Check if the each bound of the query is within the interval, or if it is the same as the bound of the interval
+	// and the bound is inclusive.
+	if interval.LowerBound.LT(queryInterval.LowerBound) ||
+		(interval.LowerBound.EQ(queryInterval.LowerBound) && interval.LowerInclusive) ||
+		(interval.LowerBound.EQ(queryInterval.LowerBound) && (interval.LowerInclusive == queryInterval.LowerInclusive)) {
```
PawelWMS, line 350: Since this is package is so complex and needs a lot of precision, I'm proposing this: let's extract these complex `if` statements into separate functions and write tests for each of these exported functions. I think that may give us more certainty we've got every one of them right.

PawelWMS, line 350: Potential simplification (assuming I didn't mess something up):
```suggestion
		(interval.LowerBound.EQ(queryInterval.LowerBound) && (interval.LowerInclusive || !queryInterval.LowerInclusive)) {
```

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -339,78 +338,56 @@ func (interval *PackageVerInterval) Compare(other *PackageVerInterval) (result i
...
-}
 
-// Contains checks if the interval fully contains the query range
-func (interval *PackageVerInterval) Contains(queryInterval *PackageVerInterval) (contains bool) {
-	// Check the lower bound
-	lowerBoundValid := interval.versionInInterval(queryInterval.LowerBound)
-	upperBoundValid := interval.versionInInterval(queryInterval.UpperBound)
+	if interval.UpperBound.GT(queryInterval.UpperBound) ||
+		(interval.UpperBound.EQ(queryInterval.UpperBound) && interval.UpperInclusive) ||
+		(interval.UpperBound.EQ(queryInterval.UpperBound) && (interval.UpperInclusive == queryInterval.UpperInclusive)) {
```
PawelWMS, line 356: Similar simplification idea:
```suggestion
		(interval.UpperBound.EQ(queryInterval.UpperBound) && (interval.UpperInclusive || !queryInterval.UpperInclusive)) {
```

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -339,78 +338,56 @@ func (interval *PackageVerInterval) Compare(other *PackageVerInterval) (result i
...
 }
 
-// Satisfies returns true the query interval overlaps the current interval at any point (ie it provides a version of a package
-// which satisfied the query)
-func (interval *PackageVerInterval) Satisfies(queryInterval *PackageVerInterval) (valid bool) {
-	var (
-		queryUpperValid, queryLowerValid, superset bool
-	)
-	if interval.LowerBound.Compare(queryInterval.UpperBound) > 0 || interval.UpperBound.Compare(queryInterval.LowerBound) < 0 {
+func (interval *PackageVerInterval) intersects(queryInterval *PackageVerInterval) (intersects bool) {
```
PawelWMS, line 363: Minor consistency suggestion: since we have a comment for `contains`, we could add one for `intersects` as well.

toolkit/tools/internal/pkgjson/pkgjson_test.go
```diff
@@ -257,25 +284,116 @@ func TestSubsetValidOutside(t *testing.T) {
...
+	testInputs := []struct {
+		name                                                                 string
+		existingPkgVer1, existingPkgCond1, existingPkgVer2, existingPkgCond2 string
+		queryPkgVer1, queryPkgCond1, queryPkgVer2, queryPkgCond2             string
+		expected                                                             bool
+	}{
+		{
+			name:            "p < q, unbounded",
+			existingPkgVer1: "1", existingPkgCond1: ">",
+			queryPkgVer1: "3", queryPkgCond1: ">",
```
PawelWMS, line 315: Why not use `PackageVer` immediately, since we use the strings only to create `PackageVer` objects anyway:
```suggestion
		existingPkg PackageVer
		queryPkg PackageVer
		expected                                                             bool
	}{
		{
			name:            "p < q, unbounded",
			existingPkgVer1: PackageVer{Version: "1", Condition: "<"},
			queryPkgVer1: PackageVer{Version: "2", Condition: ">"},
```

toolkit/tools/internal/pkgjson/pkgjson_test.go
```diff
@@ -257,25 +284,116 @@ func TestSubsetValidOutside(t *testing.T) {
...
+			name:            "subset",
+			existingPkgVer1: "1", existingPkgCond1: ">", existingPkgVer2: "3", existingPkgCond2: "<",
+			queryPkgVer1: "1", queryPkgCond1: ">=", queryPkgVer2: "5", queryPkgCond2: "<=",
+			expected: true,
+		},
+	}
+	for _, test := range testInputs {
+		t.Run(test.name, func(t *testing.T) {
+			p1 := &PackageVer{Version: test.existingPkgVer1, Condition: test.existingPkgCond1, SVersion: test.existingPkgVer2, SCondition: test.existingPkgCond2}
+			p2 := &PackageVer{Version: test.queryPkgVer1, Condition: test.queryPkgCond1, SVersion: test.queryPkgVer2, SCondition: test.queryPkgCond2}
```
PawelWMS, line 388: Could we stick to one naming scheme, so it's easier to follow the logic? Right now we have `existing` vs `query`, `p1` vs `p2`, and `p` vs `q` all at once and it's making this a bit confusing.

toolkit/tools/internal/pkgjson/pkgjson_test.go
```diff
@@ -887,3 +1005,80 @@ func TestShouldFailToConvertPackageListEntryWithWhitespacesInVersion(t *testing.
...
+				Provides: &PackageVer{Name: "pkgA", Version: "1.0.0"},
+			},
+		},
+	}
+
+	SortPackageList(pkgRepo.Repo)
+	assert.Equal(t, "pkgA", pkgRepo.Repo[0].Provides.Name)
+	assert.Equal(t, "pkgB", pkgRepo.Repo[1].Provides.Name)
+
+}
```
PawelWMS, line 1025: Do we need `PackageRepo`?
```suggestion
	pkgRepo := []*Package{
		{
			Provides: &PackageVer{Name: "pkgB", Version: "1.0.0"},
		},
		{
			Provides: &PackageVer{Name: "pkgA", Version: "1.0.0"},
		},
	}

	SortPackageList(pkgRepo)
	assert.Equal(t, "pkgA", pkgRepo[0].Provides.Name)
	assert.Equal(t, "pkgB", pkgRepo[1].Provides.Name)
}
```

I think we're also not using `PackageRepo` below.

toolkit/tools/internal/pkgjson/pkgjson_test.go
```diff
@@ -887,3 +1005,80 @@ func TestShouldFailToConvertPackageListEntryWithWhitespacesInVersion(t *testing.
 
 	assert.Error(t, err)
 }
+
+func TestSimpleSort(t *testing.T) {
```
PawelWMS, line 1009: Could you make the test name more descriptive? For example `TestSort2PackagesSameVersionDifferentName`.

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -73,11 +74,54 @@ type Package struct {
...
+		return true
+	} else if a.Name == b.Name {
+		v1 := versioncompare.New(a.Version)
+		v2 := versioncompare.New(b.Version)
+		if v1.LT(v2) {
+			return true
+		}
+	}
+
+	return false
```
PawelWMS, line 90: Simplification idea:
```suggestion
	if a.Name == b.Name {
		v1 := versioncompare.New(a.Version)
		v2 := versioncompare.New(b.Version)
		return v1.LT(v2)
	}

	return a.Name < b.Name
```

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -73,11 +74,54 @@ type Package struct {
 	RunTests      bool          `json:"RunTests"`      // Should we run tests for this package.
 }
 
+//cmp func(a, b E) int
+
+func PackageLess(a, b PackageVer) bool {
```
PawelWMS, line 79: For consistency, I'd consider creating a set of `func (pkgVer *PackageVer) LT|GT|LE|GE|EQ|NEQ` functions like what you did for intervals.

dmcilvaney: Done

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -73,11 +74,54 @@ type Package struct {
 	RunTests      bool          `json:"RunTests"`      // Should we run tests for this package.
 }
 
+//cmp func(a, b E) int
+
```
dmcilvaney, line 78:
```suggestion
```

toolkit/tools/internal/pkgjson/pkgjson_test.go
dmcilvaney: Fixed. This used to be a member function of the repo but I realized that wasn't useful in the grapher but didn't update the tests fully.

toolkit/tools/internal/pkgjson/pkgjson.go
dmcilvaney: Done

toolkit/tools/grapher/grapher.go
```diff
@@ -205,6 +208,36 @@ func addNodesForPackage(g *pkggraph.PkgGraph, pkg *pkgjson.Package) (foundDuplic
...
+			return nil, err
+		}
+		logger.Log.Debugf("Added new node: '%s' for dependency %+v", reslovedNode.FriendlyName(), dependency)
+	} else {
+		// This exact dependency is already in the graph, so reuse it.
+		reslovedNode = existingRemoteNode.RunNode
+		logger.Log.Debugf("Found existing exact remote node: '%s' for dependency %+v", reslovedNode.FriendlyName(), dependency)
+	}
+
+	return reslovedNode, nil
```
dmcilvaney, line 238:
```suggestion
		resolvedNode, err = addUnresolvedPackage(g, dependency)
		if err != nil {
			err = fmt.Errorf("failed to add a remote node (%s):\n%w", dependency.Name, err)
			return nil, err
		}
		logger.Log.Debugf("Added new node: '%s' for dependency %+v", resolvedNode.FriendlyName(), dependency)
	} else {
		// This exact dependency is already in the graph, so reuse it.
		resolvedNode= existingRemoteNode.RunNode
		logger.Log.Debugf("Found existing exact remote node: '%s' for dependency %+v", resolvedNode.FriendlyName(), dependency)
	}

	return resolvedNode, nil
```

toolkit/tools/grapher/grapher.go
```diff
@@ -205,6 +208,36 @@ func addNodesForPackage(g *pkggraph.PkgGraph, pkg *pkgjson.Package) (foundDuplic
 	return
 }
 
+// handleRemoteDependency ensures that a remote dependency is available in the graph for a given dependency.
```
dmcilvaney, line 211:
```suggestion
// handleRemoteDependency ensures that a remote node is available in the graph for every unresolved dependency.
```

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -73,11 +74,54 @@ type Package struct {
 	RunTests      bool          `json:"RunTests"`      // Should we run tests for this package.
 }
 
+//cmp func(a, b E) int
```
PawelWMS: Copilot comment leftover?

toolkit/tools/grapher/grapher.go
```diff
@@ -122,7 +123,9 @@ func addUnresolvedPackage(g *pkggraph.PkgGraph, pkgVer *pkgjson.PackageVer) (new
 		return
 	}
 
-	nodes, err := g.FindBestPkgNode(pkgVer)
+	// Double check that the package is not already in the graph. The graph should have already used that node before
+	// this point.
```
PawelWMS, line 127: I understand the first sentence but not really the second one. If I understand the `nodes != nil` check correctly, the node should **NOT** be present in the graph at this point. Could you rephrase the comment a bit?

toolkit/tools/grapher/grapher.go
```diff
@@ -217,15 +250,16 @@ func addSingleDependency(g *pkggraph.PkgGraph, packageNode *pkggraph.PkgNode, de
...
 
-	if nodes == nil {
-		dependentNode, err = addUnresolvedPackage(g, dependency)
+	// If we can't find the dependency in the graph, or it is a remote dependency, we need to do a bit of extra validation.
+	if nodes == nil || nodes.RunNode.Type != pkggraph.TypeLocalRun {
+		dependentNode, err = handleRemoteDependency(g, dependency)
 		if err != nil {
-			err = fmt.Errorf("failed to add a package (%s):\n%w", dependency.Name, err)
-			return err
+			err = fmt.Errorf("failed to handle remote dependency from %+v to %+v:\n%w", packageNode.VersionedPkg, dependency, err)
```
PawelWMS, line 257: Shouldn't we return the `err` here?
```suggestion
			err = fmt.Errorf("failed to handle remote dependency from %+v to %+v:\n%w", packageNode.VersionedPkg, dependency, err)
			return err
```

toolkit/tools/grapher/grapher.go
```diff
@@ -205,6 +208,36 @@ func addNodesForPackage(g *pkggraph.PkgGraph, pkg *pkgjson.Package) (foundDuplic
 	return
 }
 
+// handleRemoteDependency ensures that a remote node is available in the graph for every unresolved dependency.
+// 1. Check if the exact dependency is already in the graph. If it is, reuse it.
+// 2. If it is not, create a new unresolved node for the dependency.
+// It is important that we only match on the exact dependency name and version. If we don't, we may end up with
+// unpredictable behavior in the scheduler. If two different remote dependencies are added to two different build
+// nodes of a single SRPM, then the scheduler may queue that node twice.
+func handleRemoteDependency(g *pkggraph.PkgGraph, dependency *pkgjson.PackageVer) (resolvedNode *pkggraph.PkgNode, err error) {
```
PawelWMS, line 217: Naming suggestion: "handle" feel rather vague. Since we're already making a distinction between "Find**Best**" and "Find**Exact**" and we're calling `handleRemoteDependency` from inside `addSingleDependency`, what do you think about calling `handleRemoteDependency` `addExactRemoteDependency`. This way it's clearer that the effect of the function is an optional addition of a new node and that it's a subset of `addSingleDependency`, because we're focusing explicitly on exact-matched remote dependencies.

toolkit/tools/grapher/grapher.go
```diff
@@ -205,6 +208,36 @@ func addNodesForPackage(g *pkggraph.PkgGraph, pkg *pkgjson.Package) (foundDuplic
...
+	existingRemoteNode, err := g.FindExactPkgNodeFromPkg(dependency)
+	if err != nil {
+		err = fmt.Errorf("failed to check lookup list for exact remote %+v:\n%w", dependency, err)
+		return nil, err
+	}
+
+	if existingRemoteNode == nil {
+		// No exact match, add a new one.
+		resolvedNode, err = addUnresolvedPackage(g, dependency)
```
PawelWMS, line 226: I think the word "resolved" is overloaded in this scope. Could we use something else than `resolvedNode` so we don't confuse it with resolved/unresolved nodes in the graph? If I understand correctly, we are actually creating a new **unresolved** node, so I'd be very careful about the wording.

toolkit/tools/grapher/grapher.go
```diff
@@ -335,10 +369,14 @@ func populateGraph(graph *pkggraph.PkgGraph, repo *pkgjson.PackageRepo) (err err
 	timestamp.StopEvent(nil) // add package nodes
 	timestamp.StartEvent("add dependencies", nil)
 
+	// Sort the map to ensure the order is deterministic
```
PawelWMS, line 372: For my information: this has (to your knowledge) no functional impact and you're doing this to make debugging and easier and runs more deterministic, right? In other words: things should still build even if we didn't sort the packages.

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -73,11 +74,48 @@ type Package struct {
...
+		v1 := versioncompare.New(a.Version)
+		v2 := versioncompare.New(b.Version)
+		return v1.Compare(v2) < 0
+	}
+
+	return a.Name < b.Name
+}
+
+func SortPackageList(packages []*Package) {
+
```
PawelWMS, line 88: NIt-pick:
```suggestion
```

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -73,11 +74,48 @@ type Package struct {
...
+	sort.Slice(packages, func(i, j int) bool {
+		return packageLess(*packages[i].Provides, *packages[j].Provides)
+	})
+
+	// For each package, also sort the package lists
```
PawelWMS, line 93: Since we're already adding a comment, I think "dependencies lists" would be more precise and give more info to the reader. Or "its dependencies". What do you think?
```suggestion
	// For each package, also sort its dependencies
```

toolkit/tools/internal/pkgjson/pkgjson_test.go
```diff
@@ -887,3 +887,91 @@ func TestShouldFailToConvertPackageListEntryWithWhitespacesInVersion(t *testing.
 
 	assert.Error(t, err)
 }
+
+func TestSortPackageListByName(t *testing.T) {
+	Repo := []*Package{
```
PawelWMS, line 892: Convention nit-pick: start variables from small caps.

toolkit/tools/grapher/grapher.go
```diff
@@ -122,7 +123,9 @@ func addUnresolvedPackage(g *pkggraph.PkgGraph, pkgVer *pkgjson.PackageVer) (new
 		return
 	}
 
-	nodes, err := g.FindBestPkgNode(pkgVer)
+	// Double check that the package is not already in the graph. The graph should have already used that node before
```
PawelWMS, line 126: ### Unrelated clean-up asks

Could we also fix these small issues:
- Remove line 32. The `usePMCtoResolveCycles` variable is not used anywhere anymore.
- Similar story for line 47-48 with `depGraph`.
- Typos in "packged" and "PackgetVer" on line 117.

toolkit/tools/grapher/grapher_test.go
```diff
@@ -0,0 +1,149 @@
...
+	require.Equal(t, 2, len(depGraph.AllBuildNodes()))
+}
+
+// Load a specially crafted repo containing several sub-packages of a single SRPM, and two malicious
+// packages that try to pollute the available remote dependencies. The test will check if the various build
+// nodes of the sub-package SRPM are all using the same set of dependencies. The other packages will add
+// new remote nodes to the graph since they may not be satisfied by the already existing nodes. Once
+// this happens, the sub-package build nodes may start using the "better" dependencys, which is not what we want.
+func TestScenarioMultiRemoteProvides(t *testing.T) {
+	anyVer := []*pkgjson.PackageVer{
```
PawelWMS, line 72: Could you add some info, that these are versions of build-time dependencies. For instance, call this `anyVerBR` or instead of "dep1" use "BuildRequires1". Something along these lines, just to make it clearer what we're looking at.

toolkit/tools/grapher/grapher_test.go
```diff
@@ -0,0 +1,149 @@
...
+	// Validate that the graph is not empty
+	require.Equal(t, 2, len(depGraph.AllRunNodes()))
+	require.Equal(t, 2, len(depGraph.AllBuildNodes()))
+}
+
+// Load a specially crafted repo containing several sub-packages of a single SRPM, and two malicious
+// packages that try to pollute the available remote dependencies. The test will check if the various build
+// nodes of the sub-package SRPM are all using the same set of dependencies. The other packages will add
+// new remote nodes to the graph since they may not be satisfied by the already existing nodes. Once
+// this happens, the sub-package build nodes may start using the "better" dependencys, which is not what we want.
```
PawelWMS, line 70: Typo nit-pick:
```suggestion
// this happens, the sub-package build nodes may start using the "better" dependencies, which is not what we want.
```

toolkit/tools/grapher/grapher_test.go
```diff
@@ -0,0 +1,149 @@
...
+		SrpmPath:     pkgName + ".src.rpm",
+		RpmPath:      pkgName + ".rpm",
+		SourceDir:    pkgName + "-src",
+		SpecPath:     pkgName + ".spec",
+		Architecture: "x86_64",
+		IsToolchain:  false,
+		RunTests:     true,
+	}
+
+	pkg.BuildRequires = buildRequires
```
PawelWMS, line 43: Irrelevant but just curious: why is `BuildRequires` set separately from all other members?

toolkit/tools/grapher/grapher.go
dmcilvaney: Correct, this shouldn't matter functionally at all. Maps are always non-deterministically ordered so 'sorted' order is just a subset of those.

toolkit/tools/grapher/grapher_test.go
dmcilvaney: Just cruft from a previous implementation

dmcilvaney:
Build after rebase onto 3.0-dev: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=761231&view=results

toolkit/tools/internal/pkgjson/pkgjson.go
```diff
@@ -73,11 +74,49 @@ type Package struct {
...
+		v1 := versioncompare.New(a.Version)
+		v2 := versioncompare.New(b.Version)
+		return v1.Compare(v2) < 0
+	}
+
+	return a.Name < b.Name
+}
+
+// SortPackageList ensures the package list is sorted in a deterministic way. It will primarily sort by name, then
+// version. It will also recursively sort the package lists (requires, buildRequires, testRequires) of each package.
```
PawelWMS, line 88: I'd drop the word "recursively", since we don't go down each (build|test)requires and sort their (build|test)requires and then their (build|test)requires and then their (build|test)requires and... :)
```suggestion
// version. It will also sort the package lists (requires, buildRequires, testRequires) of each package.
```

PawelWMS (approved):
One more comment nit-pick but looks good.

sameluch (approved):

anphel31 (approved):

reubeno:
Changes look good to me; I did an "over-the-shoulder" code review with @dmcilvaney.

dmcilvaney:
Rerunning https://dev.azure.com/mariner-org/mariner/_build/results?buildId=766135&view=results to check image build failure. Initial assessment is its unrelated to these changes.

jslobodzian (approved):

reubeno (approved):

PR #12928, opened 2025-03-13
Add Prism as an RPM

elainezhao1:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This PR introduces [Prism ](https://github.com/microsoft/azure-linux-image-tools/tree/main/toolkit/tools/imagecustomizer)(azure linux image customizer) as an RPM, ensuring it can be built, installed, and used in Azure Linux environments.

###### Change Log  

- Creates a prism.spec file to build prism(imagecustomizer) from source.

**NO**

###### Test Methodology

tested locally , will have a pipeline run after prism GA
- Pipeline build id: xxxx


SPECS/prism/prism.spec
```diff
@@ -0,0 +1,63 @@
...
+Name:           prism
+Version:        0.12.0
+Release:        1%{?dist}
+License:        MIT
+Vendor:         Microsoft Corporation
+Distribution:   Azure Linux
+URL:            https://github.com/microsoft/azure-linux-image-tools
+Source0:        https://github.com/microsoft/azure-linux-image-tools/archive/refs/tags/v%{version}.tar.gz#/azure-linux-image-tools-%{version}.tar.gz
+BuildArch: x86_64
+BuildRequires: golang >= 1.21
```
romoh, line 14: We are using `1.23.6` now in prism.

SPECS/prism/prism.spec
```diff
@@ -0,0 +1,63 @@
...
+Requires: cdrkit
+Requires: parted
+Requires: e2fsprogs
+Requires: dosfstools
+Requires: xfsprogs
+Requires: zstd
+Requires: veritysetup
+Requires: grub2
+Requires: grub2-pc
+Requires: systemd-ukify
```
romoh, line 33: systemd-ukify is run on host so I don't think it is needed as a dependency. Lanze can confirm.

SPECS/prism/prism.spec
```diff
@@ -0,0 +1,63 @@
...
+
+%description
```
romoh, line 35: We may need to update this to `Prism` once it GAs

SPECS/prism/prism.spec
```diff
@@ -0,0 +1,63 @@
...
+Requires: squashfs-tools
+Requires: cdrkit
+Requires: parted
+Requires: e2fsprogs
+Requires: dosfstools
+Requires: xfsprogs
+Requires: zstd
+Requires: veritysetup
+Requires: grub2
+Requires: grub2-pc
```
romoh, line 32: grub2-pc is only available on amd64 and not on arm. so we may need to handle this dependency per arch.

romoh (changes requested):
left comments

vinceaperri (changes requested):

SPECS/prism/prism.spec
```diff
@@ -0,0 +1,63 @@
...
+Distribution:   Azure Linux
+URL:            https://github.com/microsoft/azure-linux-image-tools
+Source0:        https://github.com/microsoft/azure-linux-image-tools/archive/refs/tags/v%{version}.tar.gz#/azure-linux-image-tools-%{version}.tar.gz
+BuildArch: x86_64
+BuildRequires: golang >= 1.21
+Requires: qemu-img
+Requires: rpm
+Requires: coreutils
+Requires: util-linux
+Requires: /usr/bin/udevadm
```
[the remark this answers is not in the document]
cwize1: This was my suggestion. Mainly because I could easily see `udevadm` being split off into its own package some day.

That being said, it might be goodness to change all the other `Requires` to focus on specific tools as well.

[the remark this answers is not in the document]
elainezhao1: I'll use package-level Requires to align with the Prism documentation's tdnf install command. I’ll also add a comment. Happy to revisit this if we decide to switch to tool-level Requires across the board.

SPECS/prism/prism.spec
```diff
@@ -29,35 +28,58 @@ Requires: xfsprogs
...
+  (e.g. using cloud-init).
 
 %prep
 %autosetup -n azure-linux-image-tools-%{version}
 
 %build
 export GOPATH=%{our_gopath}
-cd toolkit/tools/imagecustomizer
-go build -o %{_builddir}/imagecustomizer .
+make -C ./toolkit go-imagecustomizer REBUILD_TOOLS=y
```
reubeno, line 69: FYI -- specs are encouraged to use official RPM macros for invoking standard build tools like `make`. Reference: https://docs.fedoraproject.org/en-US/packaging-guidelines/#_parallel_make

SPECS/prism/prism.spec
```diff
@@ -3,20 +3,19 @@
...
 Summary:        Azure Linux image customization tool
 Name:           prism
-Version:        0.12.0
+Version:        0.13.0
 Release:        1%{?dist}
 License:        MIT
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 URL:            https://github.com/microsoft/azure-linux-image-tools
 Source0:        https://github.com/microsoft/azure-linux-image-tools/archive/refs/tags/v%{version}.tar.gz#/azure-linux-image-tools-%{version}.tar.gz
```
reubeno, line 12: Is there a plan to make source drops available as part of upstream releases? This certainly isn't the only spec that uses github-provided source archives as sources, but there have been [previous events](https://github.blog/open-source/git/update-on-the-future-stability-of-source-code-archives-and-hashes/) where GitHub has changed the contents of the archives over time, and they seem to strongly recommend use of release archives if there's a desire to be able to confirm that the contents of the release haven't changed.

PR #13084, opened 2025-03-24
Set unique names for subpackages in kernel-rt spec

harshitgupta1337:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Currently, the subpackages `python3-perf` and `bpftools` are being built both by the kernel-rt and kernel specs. However, they have the same name - irrespective of which spec file they are built from. This causes the `python3-perf` RPM built against kernel-rt source-code to override the RPM built against kernel source-code (same for `bpftools` RPM).

###### Change Log  
- Rename the `bpftool` and `python3-perf` subpackages to be kernel specific in the `kernel-rt` spec file.
- Since `tuned` is the only package that uses `python3-perf`, change its Requires field to use either `python3-perf` (built from kernel spec) or `python3-perf-rt` (built from kernel-rt spec).

**NO**

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/SPECS-EXTENDED/kernel-rt/kernel-rt.spec
+++ b/SPECS-EXTENDED/kernel-rt/kernel-rt.spec
@@ -4,6 +4,7 @@
 %define uname_r %{version}-%{rt_version}-%{release}
 %define mariner_version 3
 %define version_upstream %(echo %{version} | rev | cut -d'.' -f2- | rev)
+%define short_name rt
 
 # find_debuginfo.sh arguments are set by default in rpm's macros.
 # The default arguments regenerate the build-id for vmlinux in the
@@ -24,7 +25,7 @@
 Summary:        Realtime Linux Kernel
 Name:           kernel-rt
 Version:        6.6.77.1
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        GPLv2
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
@@ -144,17 +145,19 @@ Requires:       audit
 %description tools
 This package contains the 'perf' performance analysis tools for Linux kernel.
 
-%package -n     python3-perf
+%package -n     python3-perf-%{short_name}
 Summary:        Python 3 extension for perf tools
+Requires:       %{name} = %{version}-%{release}
 Requires:       python3
 
-%description -n python3-perf
+%description -n python3-perf-%{short_name}
 This package contains the Python 3 extension for the 'perf' performance analysis tools for Linux kernel.
 
-%package -n     bpftool
+%package -n     bpftool-%{short_name}
 Summary:        Inspection and simple manipulation of eBPF programs and maps
+Requires:       %{name} = %{version}-%{release}
 
-%description -n bpftool
+%description -n bpftool-%{short_name}
 This package contains the bpftool, which allows inspection and simple
 manipulation of eBPF programs and maps.
 
@@ -406,14 +409,17 @@ ln -sf linux-%{uname_r}.cfg /boot/mariner.cfg
 %{_unitdir}/cpupower.service
 %config(noreplace) %{_sysconfdir}/sysconfig/cpupower
 
-%files -n python3-perf
+%files -n python3-perf-%{short_name}
 %{python3_sitearch}/*
 
-%files -n bpftool
+%files -n bpftool-%{short_name}
 %{_sbindir}/bpftool
 %{_sysconfdir}/bash_completion.d/bpftool
 
 %changelog
+* Thu Mar 20 2025 Harshit Gupta <guptaharshit@microsoft.com> - 6.6.77.1-2
+- Rename bpftool and python3-perf to be kernel specific
+
 * Mon Feb 24 2025 CBL-Mariner Servicing Account <cblmargh@microsoft.com> - 6.6.77.1-1
 - Auto-upgrade to 6.6.77.1
 
--- a/SPECS/tuned/tuned.spec
+++ b/SPECS/tuned/tuned.spec
@@ -7,7 +7,7 @@
 Summary:      A dynamic adaptive system tuning daemon
 Name:         tuned
 Version:      2.21.0
-Release:      2%{?dist}
+Release:      3%{?dist}
 License:      GPLv2+
 Vendor:       Microsoft Corporation
 Distribution:   Azure Linux
@@ -36,7 +36,7 @@ Requires: python3-configobj
 Requires: python3-dbus 
 Requires: python3-decorator
 Requires: python3-linux-procfs
-Requires: python3-perf
+Requires: (python3-perf or python3-perf-rt)
 Requires: python3-gobject
 Requires: python3-pyudev
 Requires: python3-schedutils
@@ -429,6 +429,9 @@ fi
 %{_mandir}/man7/tuned-profiles-openshift.7*
 
 %changelog
+* Thu Mar 20 2025 Harshit Gupta <guptaharshit@microsoft.com> 2.21.0-3
+- Set boolean dependency on python3-perf or python3-perf-rt
+
 * Wed Dec 11 2024 Sandeep Karambelkar <skarambelkar@microsoft.com> 2.21.0-2
 - Fix CVEs - CVE-2024-52336 CVE-2024-52337
 


SPECS/tuned/tuned.spec
```diff
@@ -36,7 +36,7 @@ Requires: python3-configobj
 Requires: python3-dbus 
 Requires: python3-decorator
 Requires: python3-linux-procfs
-Requires: python3-perf
+Requires: (python3-perf or python3-perf-rt)
```
christopherco, line 39: @dmcilvaney what is our graph's behavior for this kind of "or" for one package built in specs and another built in specs-extended, but we're only building our core specs?

dmcilvaney: I'll double check, but I believe it takes the pessimistic approach and ensures both are built. I'll need to figure out what the scheduler picks though.

dmcilvaney: I queued https://dev.azure.com/mariner-org/mariner/_build/results?buildId=773396&view=results to double check experimentally as well.

christopherco: I also queued off a buddy build for `tuned` and it does have an "unresolved dependency" failure on python3-perf-rt missing, even though `tuned` went ahead and was built.

https://dev.azure.com/mariner-org/mariner/_build/results?buildId=773389&view=logs&j=bac295d9-f2e0-5165-e63e-e76f383e1a27&t=11f6dde0-e75e-52b2-fbf0-d83a00678fa3

christopherco (changes requested):
Change currently causes a build failure. Evaluating alternatives

SPECS/tuned/tuned.spec
christopherco, line 39: @rlmenge @harshitgupta1337 have you evaluated if we could keep this as `Requires: python3-perf` and instead do a `Provides: python3-perf` from within the various kernel specs?

If that method works, it would be more sustainable than always needing to keep this `Requires: (<package> or <package2> ...)` method for each kernel package added or removed

harshitgupta1337: I agree that the method you propose would be a more sustainable method of managing this dependency of tuned.
However, if we do that - and then try installing `tuned` from a system running `kernel-rt`, how would we enforce the selection of `python3-perf-rt` (part of `kernel-rt.spec`) and not `python3-perf` (part of `kernel.spec`) ?
In the PR created by @rlmenge  (#5897 ), each kernel spec has a conflict with every other kernel spec - which could be the solution to the above.

SPECS-EXTENDED/kernel-rt/kernel-rt.spec
```diff
@@ -144,17 +145,19 @@ Requires:       audit
...
+%description -n python3-perf-%{short_name}
 This package contains the Python 3 extension for the 'perf' performance analysis tools for Linux kernel.
 
-%package -n     bpftool
+%package -n     bpftool-%{short_name}
 Summary:        Inspection and simple manipulation of eBPF programs and maps
+Requires:       %{name} = %{version}-%{release}
 
-%description -n bpftool
+%description -n bpftool-%{short_name}
```
PawelWMS, line 160: Why do we even need to build a separate version of these tools for each kernel? Do they have some compiled-in values, which would prevent the packages built in `kernel.spec` from working with other kernels?

harshitgupta1337: That is, indeed, a good question. I have been following the same approach as the other kernels. Do we know why we added similar `bpftools-*` and `python3-perf-*` packages for `kernel-64k` ?
https://github.com/microsoft/azurelinux/blob/5244dca9c0af53dfdd9a1ae715a39c1f6cef3db1/SPECS/kernel-64k/kernel-64k.spec#L157 ?

PawelWMS: I'm going to pass this question to @rlmenge.:)

rlmenge: I looked into this previously and couldn't conclusively decide. Especially since some kernels have had different sources. This should be something investigated for 4.0

SPECS/tuned/tuned.spec
harshitgupta1337: Hi @christopherco , I have tried the same approach that @rlmenge used in #5897 and tested it with an image build using AzureLinux toolkit with default kernel set to one of `kernel`, `kernel-rt`, and `kernel-ipe`. In each case, the `tuned` package was installed, which installed the correct type of `python3-perf` package based on the kernel type.

harshitgupta1337:
> Change currently causes a build failure. Evaluating alternatives

@christopherco The build failure was likely due to the boolean OR expression in the `Requires` directive of `tuned`. With the updated approach, `tuned` requires simply `python3-perf`, and each kernel variant produces a subpackage (e.g., `python3-perf-ipe`, `python3-perf-64k` that provides `python3-perf`. The `Conflict` directives and the dependency of the `python3-perf` subpackage on its kernel ensure that only the current kernel's matching subpackage is installed to satisfy `tuned`.

The RPMs build successfully, as evidenced by the following pipeline runs:
- [AMD64-2-OneBranch]-Dev-BuildRpms](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=833448&view=results)
- [OneBranch-Unified-Buddy-Build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=834792&view=results)

Furthermore, I have validated that upon installing `tuned`, the right subpackage is installed (based on the current kernel).  A description of the testing process is in this PR's description.

harshitgupta1337:
@christopherco @rlmenge I had to push some new commits due to a rebase. The PR is ready for review.
Please check [this comment](https://github.com/microsoft/azurelinux/pull/13084#issuecomment-2970908878) and the PR description for details about the approach and testing process.

rlmenge (approved):

harshitgupta1337:
Testing after recent rebase:
- BuildRpms pipeline succeeds https://dev.azure.com/mariner-org/mariner/_build/results?buildId=838575&view=results
- Building `core-efi` image with `tuned` package and different kernel variants results in the correct variant of `python3-perf` subpackage being installed (based on the kernel variant).

dmcilvaney (approved):
I think this approach *should* work. Might want to double check on a VM/Container that it doesn't swap the kernel or something, but otherwise lgtm.

harshitgupta1337:
> I think this approach _should_ work. Might want to double check on a VM/Container that it doesn't swap the kernel or something, but otherwise lgtm.

@dmcilvaney Thanks for this very valid concern. I have tested and verified that the installation attempt of a different python3-perf sub-package variant or `tuned` package does not result in a different kernel being installed. Here is the testing methodology I followed:

- Created an AzLinux 3.0 VM with the new `kernel` installed.
- Attempts to install `python3-perf-rt`, `python3-perf-ipe` and `python3-perf-lpg-innovate` fail due to conflict with current kernel.
```bash
mariner_user [ ~ ]$ sudo tdnf install python3-perf-rt
1. package python3-perf-rt-6.6.85.1-2.azl3.x86_64 requires kernel-rt = 6.6.85.1-2.azl3, but none of the providers can be installed
Found 1 problem(s) while resolving
Error(1301) : Solv general runtime error
mariner_user [ ~ ]$
mariner_user [ ~ ]$ sudo tdnf install python3-perf-ipe
[sudo] password for mariner_user:
1. package python3-perf-ipe-6.6.92.2-3.azl3.x86_64 requires kernel-ipe = 6.6.92.2-3.azl3, but none of the providers can be installed
Found 1 problem(s) while resolving
Error(1301) : Solv general runtime error
mariner_user [ ~ ]$ sudo tdnf install python3-perf-lpg-innovate
1. package python3-perf-lpg-innovate-6.6.89.2-1002.azl3.x86_64 requires kernel-lpg-innovate = 6.6.89.2-1002.azl3, but none of the providers can be installed
Found 1 problem(s) while resolving
Error(1301) : Solv general runtime error
```
- Installation of `tuned` package results in the installation of `python3-perf`, which is a sub-package of the currently installed `kernel`. `tuned` can be successfully installed.

harshitgupta1337:
Testing after yesterday's rebase.

- BuildRpms pipeline succeeds: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=846250&view=results
- Building `core-efi` image with `tuned` package and different kernel variants results in the correct variant of python3-perf subpackage being installed (based on the kernel variant).

rlmenge (approved):
Approving after rebase

allenpais (approved):

PR #13299, opened 2025-04-08
dnf: update dnf.conf to add installonlypkgs to match entry in tdnf.conf

ddstreet:
update dnf.conf to add installonlypkgs to match entry in tdnf.conf

NO


ddstreet:
@rlmenge

PR #13567, opened 2025-04-23
kernel-mshv: Introduce headers subpackage

weltling:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This PR adds `kernel-mshv-headers`, which is required by userspace tools at the build time. 

###### Change Log  

- Add kernel-mshv-headers to kernel-mshv

**NO**

###### Associated issues  

- N/A

###### Links to CVEs  
- N/A

###### Test Methodology

- Pipeline build id: 800287

### DIFF
--- a/SPECS/kernel-mshv/kernel-mshv.spec
+++ b/SPECS/kernel-mshv/kernel-mshv.spec
@@ -11,7 +11,7 @@
 Summary:        Mariner kernel that has MSHV Host support
 Name:           kernel-mshv
 Version:        5.15.157.mshv1
-Release:        3%{?dist}
+Release:        4%{?dist}
 License:        GPLv2
 Group:          Development/Tools
 Vendor:         Microsoft Corporation
@@ -76,6 +76,15 @@ Requires:       audit
 
 %description tools
 This package contains the 'perf' performance analysis tools for MSHV kernel.
+
+%package headers
+Summary:        Linux API header files
+Group:          System Environment/Kernel
+Conflicts:      kernel-headers
+
+%description headers
+The Linux API Headers expose the kernel's API for use by Glibc. This version
+also includes mshv specific headers
 
 %prep
 %autosetup -p1
@@ -168,6 +177,12 @@ make -C tools JOBS=1 DESTDIR=%{buildroot} prefix=%{_prefix} perf_install
 # Remove trace (symlink to perf). This file causes duplicate identical debug symbols
 rm -vf %{buildroot}%{_bindir}/trace
 
+# Prepare headers
+make headers
+find usr/include* \( -name ".*" -o -name "Makefile" \) -delete
+mkdir -p /%{buildroot}%{_includedir}
+cp -rv usr/include/* /%{buildroot}%{_includedir}
+
 %triggerin -- initramfs
 mkdir -p %{_localstatedir}/lib/rpm-state/initramfs/pending
 touch %{_localstatedir}/lib/rpm-state/initramfs/pending/%{uname_r}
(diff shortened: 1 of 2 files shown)


manuelh-dev (approved):

SPECS/kernel-mshv/kernel-mshv.spec
```diff
@@ -77,6 +77,14 @@ Requires:       audit
 %description tools
 This package contains the 'perf' performance analysis tools for MSHV kernel.
 
+%package headers
+Summary:        Linux API header files
+Group:          System Environment/Kernel
```
Camelron, line 82: I think we're going to want to add a Provides: kernel-headers or exclusive: kernel-headers here to prevent file conflicts with the standard kernel-headers subpackage. 

cc: @rlmenge

Camelron (changes requested):

SPECS/kernel-mshv/kernel-mshv.spec
ms-mahuber: Can you explain why kernel-mshv-headers would clash with kernel-headers? This change here is intentional

Camelron: The headers files are taken directly from the kernel tree, occupy the same space. For example, in a lsg-virt-mariner image consumed internally, the kernel-headers-mshv looks like so: 

image

The header files from a standard kernel-headers package look like so:

image

Basically every file would overlap in this case except the mshv.h file, unique to kernel-headers-mshv 

image

So maybe what we can do here @weltling is merely package the mshv header files in this package and require the base kernel-headers? But this carries the risk of the kernel-mshv source drifting apart from the standard kernel shipped in Mariner...

weltling: @Camelron yep, packaging just MSHV headers would be too complicated in the maintanence, I'd say. We're moving to the point, where the tarbal of the standard `kernel` would contain the same source as `kernel-mshv`, but most likely a different version. Thus, `kernel-mshv-headers` are principally same as `kernel-headers`, but would serve better the projects that consume `kernel-mshv` and move on their own pace. It would be expected, that `kernel` moves faster than `kernel-mshv`, but packages that depend on `kernel-mshv` and headers need to be kept compatible and buildable.

Of course, the paths in both `kernel-headers` and `kernel-mshv-headers` will collide. Having this flagged on the RPM level would be of course helpful. Perhaps a more explicit solution would be to use `Conflicts: kernel-headers`, thus giving the user a clear hind these 2 options cannot coexist.

Thanks

Camelron: Yea conflicts: seems right at a glance, let me get back to you in a bit if another approach is better.

Camelron: @weltling I synced up with @rlmenge, we agree that Conflicts is probably all we need to do here. Can you add the Conflicts: to this subpackage and I'll go and build/test some RPM transactions end-to-end with it to confirm we have the desired behavior? 

Thanks

weltling: Done. Thanks for the additional checks!

weltling: An attempt to install one of the headers after another shows an expected conflict. This.

Camelron (approved):
Thanks for addressing the kernel-headers conflict.

weltling:
After further considerations, the MSHV headers are better to be present in the existing `kernel-headers` package. This will happen, once the MSHV feature branches are integrated into the default Azure Linux kernel.

Thanks

PR #13735, opened 2025-05-09
SymCrypt-OpenSSL -- Update mechanism for creating keysinuse logging directory.

tobiasb-ms:
###### Merge Checklist  
- [x] Ready to merge

---

###### Summary 
Change the management of the `/var/log/keysinuse` directory in the `SymCrypt-OpenSSL` spec from shell commands in the `%post` and `%postun` scriptlets to a `%dir` directive in the `%files` section. This has a few advantages:
1. It removes the implicit dependency on `/bin/sh` that the scriptlets pull in to run the shell commands. This dependency is undesirable on distroless containers and this package is installed by default everywhere, including distroless.
2. We no longer delete the log files on uninstall. According [Fedora's draft guidance](https://fedoraproject.org/wiki/PackagingDrafts/Logfiles), it is more correct to leave them.
3. It informs `rpm` that this package owns `/var/log/keysinuse`. For example, `rpm -q --whatprovides /var/log/keysinuse` will return this package rather than nothing.

###### Change Log
- Changed mechanism for creating `/var/log/keysinuse`.

**NO**

###### Associated issues
- ADO: https://microsoft.visualstudio.com/OS/_queries/edit/57424422

###### Test Methodology
- Buddy Build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=805730&view=results

On both an amd64 and arm64 vm:
1. Validated that with the new version, `openssl speed`, which does a bunch of crypto operations, worked properly
2. Confirmed that with a clean install and an upgrade, `/var/log/keysinuse` has the same permissions as before.
3. Enabled the `keysinuse` logging and confirmed the log files are created in the same way as before.

For both the amd64 and arm64 rpms:
1. Confirmed that `rpm -qp --requires --recommends /path/to/rpm` was exactly the same as the `1.7.0-1` rpm (and specifically does not contain `/bin/sh`).
2. Confirmed that `tdnf install -y --releasever=3.0 --installroot /path/to/temporary/installroot /path/to/rpm` does not install `/bin/sh` and that the tree structure is identical to the `1.7.0-1` version.

### DIFF
--- a/SPECS/SymCrypt-OpenSSL/SymCrypt-OpenSSL.spec
+++ b/SPECS/SymCrypt-OpenSSL/SymCrypt-OpenSSL.spec
@@ -49,6 +49,7 @@ mkdir -p %{buildroot}%{_libdir}/engines-3/
 mkdir -p %{buildroot}%{_libdir}/ossl-modules/
 mkdir -p %{buildroot}%{_includedir}
 mkdir -p %{buildroot}%{_sysconfdir}/pki/tls/
+mkdir -p %{buildroot}%{_localstatedir}/log/keysinuse/
 
 # We still install the engine for backwards compatibility with legacy applications. Callers must
 # explicitly load the engine to use it. It will be removed in a future release.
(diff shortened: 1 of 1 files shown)


jslobodzian (approved):

reubeno (approved):

SPECS/SymCrypt-OpenSSL/SymCrypt-OpenSSL.spec
```diff
@@ -76,7 +68,21 @@ fi
 %{_includedir}/e_scossl.h
 %{_sysconfdir}/pki/tls/symcrypt_prov.cnf
 
+# The log directory for certsinuse logging has permissions set to 1733.
+# These permissions are a result of a security review to mitigate potential risks:
+# - Group and others are denied read access to prevent user-level code from inferring
```
PawelWMS, line 73: Just curious about the group permissions: why prevent "root" group read access to the folder? I'm wondering, if this wouldn't prevent legitimate actions from being executed.

PawelWMS (approved):

jslobodzian (approved):

shaharmor:
I'm not sure if this issue is related, but the latest release on containers now fails due to an existing /var/log/keysinuse folder.
See this: https://github.com/microsoft/SymCrypt-OpenSSL/commit/d07487deb9e3e83764b36165c0e936c89a34aa88#r156778518

And this: https://github.com/microsoft/KeysInUse-OpenSSL/pull/14

shaharmor:
I don’t know, didn’t dig through but when we try it now with 1.8.0-2 it fails for sure.

we fixed it internally by manually removing the folder until my PR is merged

PR #13810, opened 2025-05-15
[2.0] bmake: fix test issue

anphel31:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Improve reliability of `bmake` package build by skipping a test that is known to be flaky (fails every week or two in daily builds).  
The unit tests currently run during the %install section - I've changed this, so the tests run in a %check section.

###### Change Log  

- Change bmake to skip failing test

**NO**

###### Associated issues  

- https://microsoft.visualstudio.com/OS/_workitems/edit/57304584

###### Test Methodology

- Buddy build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=810848&view=results

### DIFF
--- a/SPECS/bmake/bmake.spec
+++ b/SPECS/bmake/bmake.spec
@@ -11,6 +11,7 @@ Source0:       %{url}/bmake-%{version}.tar.gz
 # on pipeline machines. Disabling with this patch for now, and
 # tracking this bug in workitem 38644519
 Patch0:        remove-inconsistent-time-tests.patch
+Patch1:        do-not-run-tests-on-install.patch
 Requires:      mk-files
 
 BuildRequires: gcc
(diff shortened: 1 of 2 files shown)


sameluch (approved):
Approved, but is there a way to only run the tests as a part of a %check section instead of on install?

SPECS/bmake/bmake.spec
```diff
@@ -48,6 +48,8 @@ sh ./make-bootstrap.sh
 
 %install
 export STRIP=/bin/true # Make sure binary is not stripped
+# skip 'job-output-null' which randomly fails in build pipelines
+export BROKEN_TESTS=job-output-null
```
PawelWMS, line 52: I think I've seen some logic inside the build files for `bmake`, where this test is already supposed to be skipped when the build runs in Bash (this should be the case here). Have you tried checking, why that doesn't trigger during our builds? Fixing that may be more future-proof.

Having said that, since 2.0 is getting deprecated, I think this is also good enough. I doubt we'll be updating this package.

PawelWMS (approved):

anphel31:
> Approved, but is there a way to only run the tests as a part of a %check section instead of on install?

Good idea. Yes, I've added a patch to skip calling the unit tests during install, and I've moved them to a new %check section.

SPECS/bmake/bmake.spec
anphel31: This test is automatically skipped for the `SCO_SV` OS but that's not triggered for AZL.

jslobodzian (approved):

PR #13820, opened 2025-05-16
Patch javapackages-bootstrap for CVE-2024-25710 [HIGH]

realsdx:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Fix CVE-2024-25710 in javapackages-bootstrap

###### Change Log  

- CVE-2024-25710

**NO**

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2024-25710

###### Test Methodology

- Pipeline build id: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=816266&view=results


kgodara912 (commented):
There is a typo in PR title, 2204 --> 2024. Also please update "Links to CVEs" in pr section. You removed this file, `generate-bundled-provides.sh`, is this because it is no longer required? Other than that patch matches with reference in meaningful sections, and no test included. Hopefully full build will not fail after this change.

SPECS/javapackages-bootstrap/CVE-2024-25710.patch
```diff
@@ -0,0 +1,52 @@
+From 74d2bf8a7f2ad282ebd0055c0f41ed2e6d1f5ea5 Mon Sep 17 00:00:00 2001
+From: Sudipta Pandit <sudpandit@microsoft.com>
+Date: Fri, 16 May 2025 17:10:24 +0530
+Subject: [PATCH] Backport patch for CVE-2024-25710
+
+Upstream Reference: https://github.com/apache/commons-compress/commit/8a9a5847c04ae39a1d45b365f8bb82022466067d
```
PawelWMS, line 6: I couldn't find any indication that this commit fixes this CVE. Could you add details about how these two are related to each other in the PR's summary? What I find surprising, is that the commit was from November 2023, while the CVE is first [mentioned by upstream](https://lists.apache.org/thread/cz8qkcwphy4cx8gltn932ln51cbtq6kf) in February 2024, thus I'd like to make sure we're applying the correct patch.

SPECS/javapackages-bootstrap/javapackages-bootstrap.spec
```diff
@@ -316,6 +318,10 @@ do
   tar -xf "${source}"
 done
 
+pushd "downstream/commons-compress"
+%patch -P 0 -p1
+popd
```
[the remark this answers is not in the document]
realsdx: I tried to keep it as similar as the 2.0 version we have. Also, it made sense cause inside the `downstream/` directory each folder is a different repo/source. We could change that, but I think for this type of packages this is a good exception.

SPECS/javapackages-bootstrap/CVE-2024-25710.patch
realsdx: MSRC triaged this with that reference commit. Also found another reference [here](https://security-tracker.debian.org/tracker/CVE-2024-25710). I am not entirely sure about what exactly was the vulnerability (CVE pages also shows very little information on it) I tried to keep the core logic of the upstream commit which is mentioned as a fix. The Patch just adds extra conditions to throw early exception/returns.

realsdx:
> There is a typo in PR title, 2204 --> 2024. Also please update "Links to CVEs" in pr section. You removed this file, `generate-bundled-provides.sh`, is this because it is no longer required? Other than that patch matches with reference in meaningful sections, and no test included. Hopefully full build will not fail after this change.

Avoided adding binary files for patch, as the patch adds new early return/exception scenarios, doesn't remove or alter any existing cases hopefully it would be fine. Although the check section is not running any tests currently (?):
```
%check
%{buildroot}%{launchersPath}/xmvn --version
```

SPECS/javapackages-bootstrap/CVE-2024-25710.patch
PawelWMS: I've now chatted with the MSRC researcher and the summary is:
- We could find only one mention linking the CVE to the commit on [Debian's pages](https://security-tracker.debian.org/tracker/CVE-2024-25710).
- The potential fix from the link was first present in version 1.26, which was also the first version released after the CVE was published, so this commit <-> CVE date difference makes more sense now.
- The researcher mentioned there is no POC showing how the CVE was exploited, so it's unclear how to verify the fix.

In short: we don't know a 100%, if the commit fixes the CVE. Alternative would be to jump from version 1.23 to 1.26, where the CVE is said to have been fixed, which carries a larger potential for regressions.

I'll do one last check but I think this is the best we can do right now.

SPECS/javapackages-bootstrap/javapackages-bootstrap.spec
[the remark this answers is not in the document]
PawelWMS: Got it. I wouldn't necessarily tak the 2.0 version as good reference but I'm also not going to push strongly for `%autopatch` in case of this spec, since I see your point about being able to show more clearly, which source is getting patched.

SPECS/javapackages-bootstrap/CVE-2024-25710.patch
realsdx: Thank you! Let me know if we need to change anything for this CVE

kgodara912 (approved):
Comments addressed. All checks passed. LGTM

0xba1a:
@PawelWMS @kgodara912 Do we need to patch bootstrap packages?

realsdx:
@0xba1a in 2.0 branch we have multiple patches for bootstrap packages

PawelWMS:
> @PawelWMS @kgodara912 Do we need to patch bootstrap packages?

I asked @jslobodzian this question and the summary to me sounded along the lines of "it's not mandatory but time allowing it's better to have this patched than not just in case".

PawelWMS:
@realsdx, we discussed this change with @jslobodzian now and think we could use a larger test build, because it's a critical package. Could you run a full dev build and a [fast-track dev build](https://dev.azure.com/mariner-org/mariner/_build?definitionId=2214), and verify everything still builds correctly?

realsdx:
Full build trigged: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=821884&view=results

PawelWMS (approved):

jslobodzian (approved):

PR #13881, opened 2025-05-22
containerd2: add updated multi-snapshotters support patch

miz060:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Add multi-snapshotter support patch to containerd2 enabling different runtimes to use different snapshotters simultaneously.

###### Change Log  

- Achieved by passing snapshotter parameter through the image pull flow and checking runtime handlers. This patch also enhance image store tracking to prevent unnecessary re-pulls by tracking which snapshotters already have snapshots for each image.

**NO**

Work item: https://microsoft.visualstudio.com/OS/_workitems/edit/57752493

###### Test Methodology

- buddybuild: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=816725&view=results
- imagebuild: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=820145&view=results

AKS tests:
- core conformance: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=820258&view=results
- core performance: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=820259&view=results
- pre-release: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=820260&view=results
  - containerd2 critest passed: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=820260&view=logs&j=4e017cd2-ead8-5a1f-f558-356f794173bb&t=03a2e994-9556-52b9-0ab8-ae96f929af5b&l=69
- aks-e2e: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=820324&view=results
  - aks-e2ev3 networkisolated scenario: https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=125652550&view=results - CreateManagedCluster passed which was failing for the previous version of this patch.

BVT:
- ContainerBase-Test-Docker: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=821719&view=results
- ContainerBase-Test-WSL2: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=821791&view=results


SPECS/containerd2/multi-snapshotters-support.patch
```diff
@@ -0,0 +1,279 @@
...
+-	if s == nil || s.Annotations == nil {
+-		return snapshotter, nil
+-	}
+ 
+-	// TODO(kiashok): honor the new CRI runtime handler field added to v0.29.0
+-	// for image pull per runtime class support.
+-	runtimeHandler, ok := s.Annotations[annotations.RuntimeHandler]
+-	if !ok {
+-		return snapshotter, nil
++	if runtimeHandler == "" {
```
aadhar-agarwal, line 93: Could you help explain the significance of this `if runtimeHandler == "" {` check

SPECS/containerd2/multi-snapshotters-support.patch
```diff
@@ -0,0 +1,279 @@
...
+ 
+-func (c *Controller) ensureImageExists(ctx context.Context, ref string, config *runtime.PodSandboxConfig, runtimeHandler string) (*imagestore.Image, error) {
++func (c *Controller) ensureImageExists(ctx context.Context, ref string, config *runtime.PodSandboxConfig, runtimeHandler string, snapshotter string) (*imagestore.Image, error) {
+ 	image, err := c.imageService.LocalResolve(ref)
+ 	if err != nil && !errdefs.IsNotFound(err) {
+ 		return nil, fmt.Errorf("failed to get image %q: %w", ref, err)
+ 	}
+ 	if err == nil {
+-		return &image, nil
++		if _, ok := image.Snapshotters[snapshotter]; ok || len(image.Snapshotters) == 0 {
```
aadhar-agarwal, line 192: What does `len(image.Snapshotters) == 0` indicate here? I guess in general, what is the importance of these 2 checks? Can we ever enter this `if err == nil` and then not satisfy the new if condition; and if so, what happens then?

SPECS/containerd2/multi-snapshotters-support.patch
```diff
@@ -0,0 +1,279 @@
...
++	info, err := s.provider.Info(ctx, desc.Digest)
++	if err != nil {
++		return nil, fmt.Errorf("get content store config info: %w", err)
++	}
++
++	snapshotters := make(map[string]struct{})
++	for label := range info.Labels {
++		const Prefix = "containerd.io/gc.ref.snapshot."
++		if strings.HasPrefix(label, Prefix) {
++			snapshotters[label[len(Prefix):]] = struct{}{}
```
aadhar-agarwal, line 250: Do we need to update this map anywhere else? Do we need to remove the keys (snapshotter) for example at any point?

SPECS/containerd2/multi-snapshotters-support.patch
miz060: With this patch's change in PullImage's call of snapshotterFromPodSandboxConfig, we are now adding r.GetImage().GetRuntimeHandler() for runtime information, which will return an empty string if no record is found. The if runtimeHandler == "" block preserves the behavior when runtime information was not passed in the snapshotterFromPodSandboxConfig call before this patch.

aadhar-agarwal:
Talked with @miz060 offline wrt to the comments above. Approving

aadhar-agarwal (approved):

SPECS/containerd2/multi-snapshotters-support.patch
miz060: This is inside ensureImageExists, which will be called to ensure the required pause image can be used for starting the sandbox pod. Inside this function, before this condition, the c.imageService.LocalResolve(ref) call will return true for two scenarios:
1. If there is a pause image blob in the containerd store.
2. If there is a metadata record of a snapshot in containerd.

The two check conditions were added to solve two edge cases:
1. overlayfs started a sandbox pod, pulled the pause image, and saved the overlayfs snapshot in metadata. Then the pause image gets garbage collected. When we try to start another sandbox pod with tardev-snapshotter, localResolve will return true. However, if _, ok := image.Snapshotters[snapshotter]; ok || len(image.Snapshotters) == 0 will be false because we have never saved a record in the map for tardev-snapshotter (only the overlayfs record is saved, and thus the length of the map is 1). Then we proceed to pull the image as it's needed.
2. A local copy of the pause image is cached on the container host image, and the node has no outward network connection. Then localResolve will return true because the image is in the containerd store. image.Snapshotters[snapshotter]; ok || len(image.Snapshotters) == 0 will be true because no snapshot record has been recorded in the map as the image hasn't been unpacked yet. Then we do not pull the image.

SPECS/containerd2/multi-snapshotters-support.patch
miz060: No, this function gets called during the Unpack Image process, which ensures all snapshot records are added correctly. Since it always overwrites, we don't need to clean the map itself. If the snapshot gets garbage collected, localResolve in ensureImageExists (where this map actually gets referred) will return false, and we won't go into the logic to check this map.

christopherco (approved):

PR #13903, opened 2025-05-28
Fix failures in running firewalld service

Sumynwa:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 
This commit introduces changes to fix firewalld issues in AzureLinux 3.0.
The list of changes is as below:
(a) Bump dbus-python version to 1.3.2.
- This fixes the failure to start firewalld service due to error in dbus-python:
             File "/usr/lib/python3.12/site-packages/dbus/decorators.py", line 330, in decorator
             args = inspect.getargspec(func)[0]
             ^^^^^^^^^^^^^^^^^^
Looking at the [upstream release notes](https://dbus.freedesktop.org/doc/dbus-python/news.html#dbus-python-bindings-1-3-0-2022-09-06), there is no breaking API/compatibility changes wrt python 3.12.

(b) Disable `IPv6_rpfilter` in configuration. The present kernel config does not support running firewalld daemon with this configuration. If left enabled, firewalld daemon fails to start.
 
(c) Fix testsuite provided by the firewalld-test sub-package based on the firewalld configuration provided in Azurelinux 3.0.
- disable rpfilter tests, due to missing kernel config
- add MDNS routes data and rules
- add FedoraServer & FedoraWorkStation zone info
- handle error code 255 instead of 1 for ebtables
- remove webcache/tcp from tests 
NOTE: It becomes a loop to add a check section inside the firewalld spec. This is because during build, the chroot environment running the test does not install the final configuration. As a result, the package test passing during the build will fail when running the testsuite after installing the test sub-package due to change in configuration.
Neither the [upstream SPEC](https://github.com/firewalld/firewalld/blob/d99076a59692af306e4dd6c909351a69e80b082a/firewalld.spec), nor fedora specs adds a check section for this package.

More details are mentioned in this [Bug](https://microsoft.visualstudio.com/OS/_workitems/edit/57610770).

###### Change Log  

- Bump dbus-python sources to version 1.3.2.
- Disable `IPv6_rpfilter` in configuration
- Add firewalld patch to fix tests under `src/tests`

NO

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Local Build
- [Full Build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=820733&view=results)
- [Buddy Build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=846924&view=results)

Follow below steps to run firewalld testsuite:
1. Install the built firewalld and dbus-python package
`sudo tdnf install -y python3-dbus firewalld ebtables firewalld-test`

2. Adjust below configs
- Disable ipv6 rpfilter configuration. As the kernel does not support it, the firewalld service will fail to start, causing test failures.
`sudo sed -i 's/^IPv6_rpfilter.*/IPv6_rpfilter=no/' /etc/firewalld/firewalld.conf`

- Switch iptables to legacy mode for testing `backend = iptables` configuration
`sudo update-alternatives --set ebtables /usr/sbin/ebtables-legacy`
`sudo update-alternatives --set iptables /usr/sbin/iptables-legacy`
`sudo update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy`

3. Run the test
`sudo /usr/share/firewalld/testsuite/testsuite -j8`
As of now, there are 15 known failures which occur due to disabling rpfilter in configuration.
List of failing test number: 61 107 120 124 192 193 194 195 196 197 200 240 252 306 324
firewalld_testsuite


Sumynwa:
PR check failures are not related to the changes. 
The failures are due to failing package tests.

AMD64: PTest Failure in systemd & valkey
ARM64: PTest failure in systemd

SPECS/firewalld/firewalld_fix_testsuite.patch
```diff
@@ -0,0 +1,870 @@
...
+ m4_include([features/ipset.at])
+diff --git a/src/tests/features/helpers_custom.at b/src/tests/features/helpers_custom.at
+index f51557a..788d8eb 100644
+--- a/src/tests/features/helpers_custom.at
++++ b/src/tests/features/helpers_custom.at
+@@ -36,6 +36,8 @@ NFT_LIST_RULES([inet], [filter_IN_public_allow], 0, [dnl
+     table inet firewalld {
+         chain filter_IN_public_allow {
+             tcp dport 22 accept
++	    ip daddr 224.0.0.251 udp dport 5353 accept
```
mfrw, line 178: what is the significance of this specific address ?

mfrw (approved):
LGTM :)
Would request a few more eyes to take a look though.

SPECS/firewalld/firewalld_fix_testsuite.patch
Sumynwa: This ipv4/ipv6 address and port is part of the [MDNS configuration](https://github.com/firewalld/firewalld/blob/d99076a59692af306e4dd6c909351a69e80b082a/config/services/mdns.xml#L6).

We patch `public/work` zone configs to add mdns service. These configs are installed under `/usr/lib/firewalld`.
As such, running firewalld testsuite adds additional rules in zone specific chains under firewalld table.
Since the testsuite binary compiled from sources do not include these rules in their `expected result`, they fail due to mismatch.

mfrw: thank you :)
makes sense

Sumynwa:
Looked into [3.0 Ptest run](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=847396&view=results).
The failures exist:
systemd-255-21.azl3 - Same 9 failures as in this PR check
valkey-8.0.3-3.azl3 - Different failure (no dependency install of python3-dbus or firewalld-*)

SPECS/firewalld/firewalld.spec
```diff
@@ -56,7 +56,7 @@ firewall with a D-Bus interface.
 %{?python_provide:%python_provide python3-firewall}
 Summary:        Python3 bindings for firewalld
 
-Requires:       python3-dbus
+Requires:       python3-dbus >= 1.3.2
```
mfrw, line 59: +1
Although, I think we only ship only one version of `python3-dbus` at a given point of time :)

kgodara912: @mfrw, this was done because on the systems where python3-bus is already installed, `firewalld` will not request it to upgrade and will try to start `firewalld` service which will fail. The above change will ensure that before `firewalld` install, `python3-dbus` has correct version so `firewalld` can start correctly.

SPECS/firewalld/firewalld_fix_testsuite.patch
```diff
@@ -0,0 +1,870 @@
...
+
+diff --git a/src/tests/cli/firewall-cmd.at b/src/tests/cli/firewall-cmd.at
+index 5363667..8ed25ab 100644
+--- a/src/tests/cli/firewall-cmd.at
++++ b/src/tests/cli/firewall-cmd.at
+@@ -1469,6 +1469,8 @@ FWD_START_TEST([rich rules priority])
+         table inet firewalld {
+         chain filter_IN_public_allow {
+         tcp dport 22 accept
++	ip daddr 224.0.0.251 udp dport 5353 accept
```
kgodara912, line 37: nit: used tab instead of spaces. Few other places as well.

kgodara912 (approved):
[Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=852731&view=results). Package builds and installs fine. Firewalld service starts fine. python3-dbus changes are on-par with Fedora 41 spec and works fine for firewalld.
```
$ sudo systemctl start firewalld
$ systemctl status firewalld
● firewalld.service - firewalld - dynamic firewall daemon
     Loaded: loaded (/usr/lib/systemd/system/firewalld.service; disabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: active (running) since Mon 2025-06-30 05:39:04 UTC; 37s ago
```

LeoMar4 (approved):

PR #14245, opened 2025-07-09
Enable arm64 build for mlnx-ofa_kernel

ellie-di:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This component is required to build the CUDA for arm64.

###### Change Log  

- Dropped ExcusiveArch
- Build the spec for arm64 against kernel-64k

**NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/SPECS/mlnx-ofa_kernel/mlnx-ofa_kernel.spec
+++ b/SPECS/mlnx-ofa_kernel/mlnx-ofa_kernel.spec
@@ -106,7 +106,7 @@
 Summary:	 Infiniband HCA Driver
 Name:		 mlnx-ofa_kernel
 Version:	 24.10
-Release:	 20%{release_suffix}%{?dist}
+Release:	 21%{release_suffix}%{?dist}
 License:	 GPLv2
 Url:		 http://www.mellanox.com/
 Group:		 System Environment/Base
@@ -116,7 +116,6 @@ Patch0:          001-fix-module-init-for-ibt.patch
 BuildRoot:	 /var/tmp/%{name}-%{version}-build
 Vendor:          Microsoft Corporation
 Distribution:    Azure Linux
-ExclusiveArch:   x86_64
 
 Obsoletes: kernel-ib
 Obsoletes: mlnx-en
@@ -130,7 +129,11 @@ Obsoletes: mlnx-en-doc
 Obsoletes: mlnx-en-debuginfo
 Obsoletes: mlnx-en-sources
 
+%ifarch aarch64
+BuildRequires:  kernel-64k-devel = %{target_kernel_version_full}
+%else
 BuildRequires:  kernel-devel = %{target_kernel_version_full}
+%endif
 BuildRequires:  kernel-headers = %{target_kernel_version_full}
 BuildRequires:  binutils
 BuildRequires:  kmod
@@ -138,7 +141,11 @@ BuildRequires:  libstdc++-devel
 BuildRequires:  libunwind-devel
 BuildRequires:  pkgconfig
 
+%ifarch aarch64
+Requires: kernel-64k = %{target_kernel_version_full}
+%else
 Requires: kernel = %{target_kernel_version_full}
+%endif
 Requires: kmod
 Requires: libstdc++
 Requires: libunwind
@@ -741,6 +748,9 @@ update-alternatives --remove \
 %{_prefix}/src/mlnx-ofa_kernel-%version
 
 %changelog
+* Wed July 09 2025 Elaheh Dehghani <edehghani@microsoft.com> - 24.10-21
+- Enabled aarch64 (ARM64) build by removing ExclusiveArch
+
 * Thu May 29 2025 Nicolas Guibourge <nicolasg@microsoft.com> - 24.10-20
 - Add kernel version and release nb into release nb
 


SPECS-SIGNED/mlnx-ofa_kernel-modules-signed/mlnx-ofa_kernel-modules-signed.spec
```diff
@@ -192,6 +194,9 @@ fi
 %license %{_datadir}/licenses/%{_name}/copyright
 
 %changelog
+* Wed July 09 2025 Elaheh Dehghani <edehghani@microsoft.com> - 24.10-21
```
anphel31: Can you change `July` to `Jul` ?

SPECS/mlnx-ofa_kernel/mlnx-ofa_kernel.spec
```diff
@@ -741,6 +748,9 @@ update-alternatives --remove \
 %{_prefix}/src/mlnx-ofa_kernel-%version
 
 %changelog
+* Wed July 09 2025 Elaheh Dehghani <edehghani@microsoft.com> - 24.10-21
```
anphel31: Can you change `July` to `Jul` ?

SPECS-SIGNED/fwctl-signed/fwctl-signed.spec
```diff
@@ -119,6 +133,10 @@ fi # 1 : closed
 
 
 %changelog
+* Wed July 09 2025 Elaheh Dehghani <edehghani@microsoft.com> - 24.10-21
```
anphel31: Can you use `Jul` instead of `July`

SPECS/xpmem/xpmem.spec
```diff
@@ -70,8 +76,13 @@ ExclusiveArch:   x86_64
 
 BuildRequires:  gcc
 BuildRequires:  make
+%ifarch aarch64
+BuildRequires:  kernel-hwe-devel = %{target_kernel_version_full}
+#BuildRequires:  kernel-hwe-headers = %{target_kernel_version_full}
```
anphel31: Are the kernel-hwe-headers needed? Wondering why is this commented out. Can you please clarify?

ellie-di: I believe no. This was causing conflict during the build with kernel-hwe-headers as chroot already contains kernel-headers. I just kept it in consistency with x86_64 part, but I think I can go ahead and remove it from both?

ellie-di: removed them.

SPECS/isert/isert.spec
```diff
@@ -66,20 +73,22 @@
...
 Source0:         https://linux.mellanox.com/public/repo/mlnx_ofed/24.10-0.7.0.0/SRPMS/isert-24.10.tgz#/isert-%{_version}.tgz
 BuildRoot:	 /var/tmp/%{name}-%{version}-build
 Vendor:          Microsoft Corporation
 Distribution:    Azure Linux
-ExclusiveArch:   x86_64
 
 BuildRequires:  gcc
 BuildRequires:  make
+%ifarch aarch64
+BuildRequires:  kernel-hwe-devel = %{target_kernel_version_full}
```
PawelWMS: I don't think we should mix building with different kernel versions inside one spec. We're just lucky for now that ARM64 is building with `kernel-hwe`, while AMD64 with the "default" `kernel` but if that ever changes (we want to build AMD64 against `kernel-hwe` for instance, or other kernel with a different version), then this won't work. We don't have good support for this scenario at the moment, unfortunately. My initial hunch is to create a separate spec for each "kernel group", where all kernels in a group update their version in lockstep.

ellie-di: I tried to go through that road but that's challenging in a sense that the source name and version is similar and handling them in two different specs with two different name is confusing and error prone. I think we should go with these changes as they are and wait for the solution that comes to support this scenario.

SPECS/mlnx-ofa_kernel-hwe/mlnx-ofa_kernel-hwe.spec
```diff
@@ -0,0 +1,831 @@
...
+# notice, one of the license notices in the documentation
+# and/or other materials provided with the distribution.
+#
+#
+
+%if 0%{azl}
+# hard code versions due to ADO bug:58993948
+%global target_kernel_version_full 6.12.40.1-1.azl3
+%global target_azl_build_kernel_version 6.12.40.1
+%global release_suffix _6.12.40.1.1
```
PawelWMS: Looking at spec like `iser.spec`, I think we actually need two things:
- kernel version
- kernel release number

And then we can combine them to build other macros. Here's my suggestion to limit the number of hard-coded macros, so it's harder to make a typo of forget to update one of them:

```suggestion
%global target_azl_build_kernel_version 6.12.40.1
%global target_kernel_release 1
%global target_kernel_version_full %{target_azl_build_kernel_version}-%{target_kernel_release}.%{?dist}
%global release_suffix _%{target_azl_build_kernel_version}.%{target_kernel_release}
```

SPECS/xpmem-hwe/xpmem-hwe.spec
```diff
@@ -0,0 +1,312 @@
...
+%description
+XPMEM is a Linux kernel module that enables a process to map the
+memory of another process into its virtual address space. Source code
+can be obtained by cloning the Git repository, original Mercurial
+repository or by downloading a tarball from the link above.
+
+This package includes helper tools for the kernel module.
+
+%if ! %{with kernel_only}
+%package -n libxpmem
```
ellie-di: @PawelWMS should I rename this to `libxpmem-hwe`?

SPECS/mlnx-ofa_kernel-hwe/mlnx-ofa_kernel-hwe.spec
ellie-di: Done. Thanks for the suggestion

SPECS/xpmem-hwe/xpmem-hwe.spec
PawelWMS: Yes, please. Good catch btw., I missed it. We shouldn't have any same-named packages coming from different spec file.

PawelWMS: It might be helpful to define a macro to re-use within the spec like `%global kernel_suffix -hwe` or something like that to make it harder to do a typo. And it'll make it easier in the future to adjust these specs, when the toolkit is more flexible to handle the different kernel we have.

SPECS/xpmem-hwe/xpmem-hwe.spec
```diff
@@ -0,0 +1,316 @@
...
+mkdir -p $RPM_BUILD_ROOT/lib/firmware
+%endif
+
+%clean
+rm -rf $RPM_BUILD_ROOT
+
+%post modules
+depmod %{KVERSION} -a
+/sbin/modprobe -r xpmem > /dev/null 2>&1
+/sbin/modprobe xpmem > /dev/null 2>&1
```
PawelWMS, line 214: This `%post` section made me realize: the signed version doesn't have any post-install scripts like these, so it doesn't automatically insert the module upon installation. The default kernel version behaves the same way, so it may be intentional, but wanted to point it out, in case that's just something that was missed when the first `xpmem` was introduced.

SPECS-SIGNED/srp-hwe-signed/srp-hwe-signed.spec
```diff
@@ -0,0 +1,174 @@
...
+%endif
+
+%global KVERSION %{target_kernel_version_full}
+
+%define _name srp-hwe
+%{!?_mofed_full_version: %define _mofed_full_version 24.10-20%{release_suffix}%{?dist}}
+
+Summary:	 srp driver
+Name:		 %{_name}-signed
+Version:	 24.10
```
PawelWMS, line 50: One more file to update: `toolkit/scripts/check_entangled_specs.py`. This is to make sure the signed and unsigned spec pairs are updated together. You'll need to update the `version_release_matching_groups` array in that file. There are examples there for the current OOT module specs already, so I think this will only require some copy-pasting and mild edits to add the `-hwe` pairs.

SPECS/xpmem-hwe/xpmem-hwe.spec
ellie-di: Done. Thanks!

SPECS-SIGNED/xpmem-hwe-modules-signed/xpmem-hwe-modules-signed.spec
```diff
@@ -84,6 +84,10 @@ popd
 /lib/modules/%{KVERSION}/updates/xpmem.ko
 %{_datadir}/licenses
 
+%post
+depmod %{KVERSION} -a
+/sbin/modprobe -r xpmem > /dev/null 2>&1
+/sbin/modprobe xpmem > /dev/null 2>&1
```
PawelWMS: Should we also unload the module and re-run `depmod` in `%preun` or `%postun` to properly clean-up when the package is uninstalled?

ellie-di: updated the `%postun` section to match with what is in the unsigned spec.

PawelWMS (approved):

PR #14360, opened 2025-07-21
Initial azure-linux-image-tools as a new rpm from SPEC.

liulanze:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Here's a shorter, cleaner version:

## What does the PR accomplish, why was it needed?

Adds RPM spec file to package Azure Linux Image Tools, which is a required step for building the Golden Container.

This transforms a complex manual installation process into a standard package installation, significantly improving accessibility and enterprise adoption.

**NO**

###### Test Methodology

- Pipeline build id: buddy build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=875096&view=results


SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,94 @@
...
+#   3. cd %%{name}-%%{version}
+#   4. go mod tidy
+#   5. go mod vendor
+#   6. tar  --sort=name \
+#           --mtime="2025-07-18 00:00Z" \
+#           --owner=0 --group=0 --numeric-owner \
+#           --pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime \
+#           -cf %%{name}-%%{version}-vendor.tar.gz vendor
+#
+Source1:        %{name}-%{version}-vendor-1.tar.gz
```
liulanze: Will update Source1 to `%{name}-%{version}-vendor.tar.gz` once released to `v0.16.0`

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,94 @@
...
+Requires: binutils
+Requires: lsof
+Requires: python3
+Requires: python3-pip
+Requires: jq
+%ifarch x86_64
+Requires: grub2-pc
+Requires: systemd-ukify
+%endif
+# Requires: oras
```
liulanze: The `oras` is not available in Azl3.

SPECS/azure-linux-image-tools/skip-license-scan.patch
```diff
@@ -0,0 +1,66 @@
+diff --git a/toolkit/Makefile b/toolkit/Makefile
```
liulanze: The patch can be removed once released to `v0.16.0`

.pipelines/containerSourceData/azurelinuximagetools/Dockerfile-azure-linux-image-tools
```diff
@@ -0,0 +1,19 @@
...
+ARG BASE_IMAGE
+
+FROM $BASE_IMAGE
+
+@INCLUDE_MAIN_RUN_INSTRUCTION@
+
+RUN tdnf update -y && \
+   tdnf install -y azurelinux-repos-cloud-native && \
+   tdnf update -y && \
+   tdnf install -y qemu-img rpm coreutils util-linux systemd openssl \
```
rikenm1: Why are we installing these packages again from here? We usually just want pkgs installation to be done using the azurelinuximagetools.pkg file. Packages mentioned in that file will be installed at line 8 "@INCLUDE_MAIN_RUN_INSTRUCTIONS".

SPECS/azure-linux-image-tools/azure-linux-image-tools.signatures.json
```diff
@@ -0,0 +1,6 @@
+{
+    "Signatures": {
+        "azure-linux-image-tools-0.16.0.tar.gz": "2af99782b9aaccdaa162a24db56f4520be8b6701c6f5cf6c6e33960e90917b8e",
+        "azure-linux-image-tools-0.16.0-vendor-test.tar.gz": "44ca4f97d2fe8f41d2ae0116ae7835063be60338394493a22cf648cbae1f43cf"
```
liulanze: The `-test` is for testing purpose temporarily, will remove.

SPECS/azure-linux-image-tools/generate-source-tarball.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
vinceaperri: This script needs to be renamed to generate_source_tarball.sh and support "--srcTarball" option for the auto-patcher to support it. You can use the script in the documentation for this: https://github.com/microsoft/azurelinux/blob/3.0/toolkit/docs/building/add-package.md#getting-the-sources

vinceaperri: You can probably still `wget` the tarball if this option is not provided.

.pipelines/containerSourceData/imagecustomizer/imagecustomizer.pkg
```diff
@@ -0,0 +1,22 @@
+qemu-img
```
vinceaperri, line 1: This is missing grub2-pc for amd64 containers as well as systemd-ukify for amd64 3.0 containers. Any way to configure packages specifically for them, or conditionally install them in the Dockerfile?

vinceaperri: See https://github.com/microsoft/azure-linux-image-tools/blob/main/docs/imagecustomizer/quick-start/quick-start-binary.md#instructions

vinceaperri: In the internal PR I see we won't be building for 2.0. So grub2-pc and systemd-ukify can be treated in the same way: if amd64, then install, else ignore.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,102 @@
...
+BuildRequires: systemd-udev
+Requires: imagecustomizer = %{version}-%{release}
+
+%description
+Azure Linux Image Tools. This package provides the Azure Linux Image Customizer tool
+and its dependencies for customizing Azure Linux images.
+
+%package -n imagecustomizer
+Summary: Image Customizer
+Requires: qemu-img
```
cwize1: In general, it is better to declare the dependency on the executable (path) instead of the package name. That allows, the package that provides the executable to change without it affecting this package definition.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,102 @@
...
+#   6. tar  --sort=name \
+#           --mtime="2025-07-18 00:00Z" \
+#           --owner=0 --group=0 --numeric-owner \
+#           --pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime \
+#           -cf %%{name}-%%{version}-vendor.tar.gz vendor
+#
+Source1:        %{name}-%{version}-vendor-test.tar.gz
+BuildRequires: golang >= 1.24.1
+BuildRequires: systemd-udev
+Requires: imagecustomizer = %{version}-%{release}
```
cwize1: Why does the `Requires: imagecustomizer = %{version}-%{release}` exist?

cwize1 (changes requested):

.pipelines/containerSourceData/imagecustomizer/imagecustomizer.pkg
liulanze: I plan to try using build argument `TARGETARCH` that is available automatically inside the Dockerfile and can be leveraged by the processes running as part of the build. Please have a review.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
liulanze: The reason is I want when user run `tdnf install azure-linux-image-tools`, it will help to install `imagecustomizer` tool as well. 

If I put `%{_bindir}/imagecustomizer` not only under `%files -n imagecustomizer` but also under `%files`, it will show conflict error. Unless we want the main package `azure-linux-image-tools` to be empty when gets installed. Or there is any other better ways.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
liulanze: Good point. But I notice most Azure Linux spec files use package names for dependencies, should we maintain consistency here?

SPECS/azure-linux-image-tools/generate-source-tarball.sh
liulanze: Fixed, thanks.

.pipelines/containerSourceData/imagecustomizer/Dockerfile-imagecustomizer
```diff
@@ -2,15 +2,36 @@
 # Licensed under the MIT License.
 
 ARG BASE_IMAGE
+ARG TARGETARCH
+ARG AZ_MON_CONN_STR
 
 FROM $BASE_IMAGE
 
+ENV AZURE_MONITOR_CONNECTION_STRING=${AZ_MON_CONN_STR}
```
vinceaperri, line 10: Does it have to be `AZ_MON_CONN_STR` or can you just use the same variable name `AZURE_MONITOR_CONNECTION_STRING`?

.pipelines/containerSourceData/imagecustomizer/Dockerfile-imagecustomizer
```diff
@@ -2,15 +2,36 @@
...
    tdnf clean all
 
+# Create virtual environment and install Python dependencies for telemetry
+RUN python3 -m venv /opt/telemetry-venv
+COPY telemetry-requirements.txt /telemetry-requirements.txt
+RUN /opt/telemetry-venv/bin/pip install --no-cache-dir -r /telemetry-requirements.txt
+RUN rm -rf /telemetry-requirements.txt
+
+# Copy all necessary files
+COPY .mariner-toolkit-ignore-dockerenv /
```
vinceaperri: You've added the empty file .mariner-toolkit-ignore-dockerenv to this PR, but none of the files under usr/. Is that intentional?

.pipelines/containerSourceData/imagecustomizer/telemetry-requirements.txt
```diff
@@ -0,0 +1,5 @@
+azure-monitor-opentelemetry-exporter==1.0.0b37
```
vinceaperri: There seems to be a lot of file duplication between this directory and azure-linux-image-tools repository. Is there any way to dynamically use the files from the commit of the version declared in the SPEC? Or at the very least document what files need to be updated (and how) during a version upgrade of the SPEC?

SPECS/azure-linux-image-tools/generate_source_tarball.sh
```diff
@@ -0,0 +1,111 @@
...
+SRC_TARBALL=""
+OUT_FOLDER="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# parameters:
+#
+# --srcTarball    : src tarball file
+#                   this file contains the 'initial' source code of the component
+#                   and should be replaced with the new/modified src code
+# --outFolder     : folder where to copy the new tarball(s)
+# --pkgVersion    : package version
```
vinceaperri: Any reason not to include `--vendorVersion` and not to set the vendor tarball name to `$OUT_FOLDER/$NAME_VER-govendor-v$VENDOR_VERSION.tar.gz`?

SPECS/azure-linux-image-tools/generate_source_tarball.sh
```diff
@@ -0,0 +1,111 @@
...
+    SOURCE_FILE="$NAME_VER.tar.gz"
+    wget https://github.com/microsoft/azure-linux-image-tools/archive/refs/tags/v$PKG_VERSION.tar.gz -O "$SOURCE_FILE"
+fi
+
+echo "Unpacking source tarball..."
+tar -xf "$SOURCE_FILE"
+cd "$NAME_VER/toolkit/tools"
+
+echo "Generate vendored modules tarball"
+go mod tidy
```
vinceaperri: Did you find you needed to `go mod tidy` the released source code? Could be removed as I would expect the released source code to already be "tidy". It is not in the documented generate_source_tarball.sh script.

.pipelines/containerSourceData/imagecustomizer/Dockerfile-imagecustomizer
```diff
@@ -0,0 +1,37 @@
...
+ENV AZURE_MONITOR_CONNECTION_STRING=${AZ_MON_CONN_STR}
+
+@INCLUDE_MAIN_RUN_INSTRUCTION@
+
+# Install core packages and architecture-specific packages
+RUN tdnf update -y && \
+   tdnf install -y azurelinux-repos-cloud-native && \
+   tdnf update -y && \
+   tdnf install -y oras && \
+   if [ "$TARGETARCH" = "amd64" ]; then \
```
vinceaperri, line 18: Can we create a follow-up PR in the azure-linux-image-tools repository to use the same logic: https://github.com/microsoft/azure-linux-image-tools/blob/292220e17fd7840722e6d3a8eac93ce64f091db5/toolkit/tools/imagecustomizer/container/build-container.sh#L99 ?

SPECS/azure-linux-image-tools/generate_source_tarball.sh
liulanze: We can, but does the vendor version has use cases? I notice that the `$NAME_VER` already contains the package name for recognition.

SPECS/azure-linux-image-tools/generate_source_tarball.sh
liulanze: I found this error log when I build without `go mod tidy`, but fixed with `go mod tidy`, please check.

```
go: inconsistent vendoring in /usr/src/azl/BUILD/azure-linux-image-tools-0.15.0/toolkit/tools:
    <many dependencies> are explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
```

.pipelines/containerSourceData/imagecustomizer/Dockerfile-imagecustomizer
liulanze: The dependent files / scripts have been ignored by git.

.pipelines/containerSourceData/imagecustomizer/Dockerfile-imagecustomizer
liulanze: I was trying to make the golden container Dockerfile as similar process as the one we have today: https://github.com/microsoft/azure-linux-image-tools/blob/main/toolkit/tools/imagecustomizer/container/imagecustomizer.Dockerfile

.pipelines/containerSourceData/imagecustomizer/Dockerfile-imagecustomizer
liulanze: And as for the empty file `.mariner-toolkit-ignore-dockerenv`, I was trying to mimic the existing `build-container.sh`: https://github.com/microsoft/azure-linux-image-tools/blob/382ec88313eeffeef5dcc95536741cf3506f86a4/toolkit/tools/imagecustomizer/container/build-container.sh#L96

It seems this file gets `touch` but no other ops.

.pipelines/containerSourceData/imagecustomizer/telemetry-requirements.txt
gmileka: I wonder if we are keep both. For example, do we plan on continuing to build the container both as a golden container and as an individual container? Would be good to have one build system building it... Can azure linux golden container build produce contaienrs in a dev acr for internal consumption?
As to where the docker file live, I think they can live in azurelinux with the rest of the golder container - or be pulled into azurelinux repo through a git module form azure-linux-image-tools?
[you guys might have discussed these options already - trying to catch up]

.pipelines/containerSourceData/imagecustomizer/imagecustomizer.pkg
```diff
@@ -0,0 +1,23 @@
...
+rpm
```
gmileka, line 2: Where is this file being used? same for imagecustomizer.name?

SPECS/azure-linux-image-tools/generate_source_tarball.sh
```diff
@@ -0,0 +1,111 @@
...
+go mod vendor
+
+echo "Tar vendored modules"
+tar  --sort=name \
+     --mtime="2025-07-18 00:00Z" \
```
gmileka: curious, why do we need `mtime`? Also, does this specific value matters?

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,102 @@
...
+Summary:        Azure Linux Image Tools
+Name:           azure-linux-image-tools
+Version:        0.16.0
+Release:        1%{?dist}
+License:        MIT
+URL:            https://github.com/microsoft/azure-linux-image-tools/
+Group:          Applications/System
+Vendor:         Microsoft Corporation
+Distribution:   Azure Linux
+Source0:        https://github.com/microsoft/azure-linux-image-tools/archive/refs/tags/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
```
gmileka: Are we going to also store the tar.gz on our internal storage?

SPECS/azure-linux-image-tools/generate_source_tarball.sh
jslobodzian: I think this is required to handle some issues with consistent hash checking the tar ball.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,102 @@
...
+Source1:        %{name}-%{version}-vendor-test.tar.gz
+BuildRequires: golang >= 1.24.1
+BuildRequires: systemd-udev
+Requires: imagecustomizer = %{version}-%{release}
+
+%description
+Azure Linux Image Tools. This package provides the Azure Linux Image Customizer tool
+and its dependencies for customizing Azure Linux images.
+
+%package -n imagecustomizer
```
jslobodzian: Is there a reason to have the main package just install the license and have a second subpackage that is just the binaries?  Consider making the package imagecustomizer or perhaps azlimagecustomizer and drop the -n.  It's not mandatory, but it's not clear why we are splitting these here.

.pipelines/containerSourceData/imagecustomizer/imagecustomizer.pkg
liulanze: Right, please check the internal PR `Adding new container image imagecustomizer` under `CBL-Mariner-Pipelines`.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
liulanze: Yes, the tarball for source0 and source1 have been uploaded to the blob storage.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,102 @@
...
+#   2. tar -xf %%{name}-%%{version}.tar.gz
+#   3. cd %%{name}-%%{version}
+#   4. go mod tidy
+#   5. go mod vendor
+#   6. tar  --sort=name \
+#           --mtime="2025-07-18 00:00Z" \
+#           --owner=0 --group=0 --numeric-owner \
+#           --pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime \
+#           -cf %%{name}-%%{version}-vendor.tar.gz vendor
+#
```
PawelWMS: We started with these elaborate comments before the introduction of the `generate_source_tarball.sh` scripts. Right now, it should be enough for the comment to explain how to use the script locally, if needed.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,102 @@
+%global debug_package %{nil}
```
PawelWMS: Why no debug package?

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
PawelWMS: +1 on asking if we need a separate `imagecustomizer` package.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,102 @@
...
+%description -n imagecustomizer
+The Azure Linux Image Customizer is a tool that can take an
+existing generic Azure Linux image and modify it to be suited for a particular
+scenario. By providing an Azure Linux base image, users can also supply a config
+file specifying how they want the image to be customized. For example, this
+could include the installation of certain RPMs, updating the SELinux mode, and
+enabling DM-Verity.
+
+%prep
+%autosetup -n %{name}-%{version} -p1
```
PawelWMS: I think `-n %{name}-%{version}` shouldn't be necessary - that's the default, if I remember correctly.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
```diff
@@ -0,0 +1,102 @@
...
+rm -rf toolkit/tools/vendor
```
PawelWMS: Why the `rm`?

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
PawelWMS: I'd say both are OK. Technically, it's also possible that a file path will change but both are unlikely and in case of a package re-name we're expected to make sure the old name still works through the use of the `Provides` tag.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
liulanze: Removed.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
liulanze: Fixed, thanks.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
liulanze: Fixed, thanks. The original repo does not contain the vendor folder.

SPECS/azure-linux-image-tools/azure-linux-image-tools.spec
Britel: The plan was to add additional image tools as sub packages. Things like ImageCreator - which share the same source and build system.

PawelWMS: Is the idea for the default `azure-linux-image-tools` package to install ALL subpackages like `imagecustomizer` and the future subpackage for the ImageCreator while at the same time the subpackages are to there to allow the user to install a smaller package, if they don't need all of them? If that's the case, it makes sense to me.

I'd only consider keeping the same package prefix to make it clear all of these tools are from the same `azure-linux-image-tools` "umbrella". This boils down to dropping the `-n` switch in `%package` and `%files`.

While on this topic, I'd also consider re-naming "azure-linux" to "azurelinux" to follow the naming we have for other AZL packages like `azurelinux-repos` or `azurelinux-release`.

Oh, and technically, since `azure-linux-image-tools` doesn't install any binaries and at the same time the subpackages install their licenses, we can drop the `%license LICENSE` line from the `%files` section for `azure-linux-image-tools`.

liulanze: I have updated the package name to `azurelinux-image-tools-*` and remove unnecessary `-n`.

SPECS/azure-linux-image-tools/azure-linux-image-tools.signatures.json
liulanze: `-test` been removed, comment closed.

.pipelines/containerSourceData/imagecustomizer/usr/local/bin/entrypoint.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
vinceaperri: What processes are in force that will prevent these files from becoming out of date? Would it be possible to add a PR check to ensure that the corresponding file in v0.17 of azure-linux-image-tools matches this file exactly?

.pipelines/containerSourceData/imagecustomizer/usr/local/bin/run.sh
```diff
@@ -0,0 +1,136 @@
+#!/bin/bash
+
+set -euo pipefail
+shopt -s nullglob
+
+exit_with_usage() {
+    local error_msg="$1"
```
vinceaperri: This is not equivalent to the run.sh in v0.17 release.

vinceaperri: It is missing the bug I fixed here https://github.com/microsoft/azure-linux-image-tools/pull/350

.pipelines/containerSourceData/imagecustomizer/Dockerfile-imagecustomizer
```diff
@@ -0,0 +1,37 @@
...
+
+ENV AZURE_MONITOR_CONNECTION_STRING=${AZ_MON_CONN_STR}
+
+@INCLUDE_MAIN_RUN_INSTRUCTION@
+
+# Install core packages and architecture-specific packages
+RUN tdnf update -y && \
+   tdnf install -y azurelinux-repos-cloud-native && \
+   tdnf update -y && \
+   tdnf install -y oras && \
```
vinceaperri, line 17: This is missing `qemu-img rpm coreutils util-linux systemd openssl sed createrepo_c squashfs-tools cdrkit parted e2fsprogs dosfstools xfsprogs zstd veritysetup grub2 binutils lsof python3 python3-pip jq` as installed [here](https://github.com/microsoft/azure-linux-image-tools/blob/1ca3fbf0b68e36a2c12008466a4433ae129bf87c/toolkit/tools/imagecustomizer/container/imagecustomizer.Dockerfile#L15)

SPECS/azurelinux-image-tools/generate_source_tarball.sh
```diff
@@ -0,0 +1,113 @@
...
+tmpdir=$(mktemp -d)
+function cleanup {
+    echo "+++ cleanup -> remove $tmpdir"
+    rm -rf $tmpdir
+}
+trap cleanup EXIT
+
+pushd $tmpdir > /dev/null
+
+PKG_NAME="azure-linux-image-tools"
```
gmileka, line 80: azure-linux or azuelinux?

SPECS/azurelinux-image-tools/generate_source_tarball.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
```
gmileka, line 1: is that a helper script to run when uploading a new source tarball?

.pipelines/containerSourceData/imagecustomizer/Dockerfile-imagecustomizer
liulanze: These dependencies will be installed through `imagecustomizer.pkg`, see below. This is one of the golden container build features. AFAIK, the `@INCLUDE_MAIN_RUN_INSTRUCTION@` will help to do so.

SPECS/azurelinux-image-tools/generate_source_tarball.sh
liulanze: Yes, but locally. Once the source tarball and vendor tarball get generated, these files should be uploaded to the blob.

SPECS/azurelinux-image-tools/generate_source_tarball.sh
liulanze: Here should be `azure-linux`, the original repo name decides this. But thanks for the checking here.

.pipelines/containerSourceData/imagecustomizer/usr/local/bin/entrypoint.sh
liulanze: TBD - Mark here

.pipelines/containerSourceData/imagecustomizer/usr/local/bin/run.sh
liulanze: TODO - sync up container files.

SPECS/azurelinux-image-tools/azurelinux-image-tools.spec
```diff
@@ -0,0 +1,105 @@
...
+go test -C toolkit/tools ./...
+
+%files
+
+%files imagecustomizer
+%license LICENSE
+%{_bindir}/imagecustomizer
+# Container support files - placed in container filesystem paths with imagecustomizer- prefix
+/usr/local/bin/imagecustomizer-entrypoint.sh
+/usr/local/bin/imagecustomizer-run.sh
```
vinceaperri, line 97: Not blocking this PR but please be aware that renaming these scripts will break existing functionality:

- Renaming [entrypoint.sh](https://github.com/search?q=repo%3Amicrosoft%2Fazure-linux-image-tools+entrypoint.sh&type=code) seems to be fine but may break the tests.
- Renaming [run.sh](https://github.com/search?q=repo%3Amicrosoft%2Fazure-linux-image-tools+run.sh&type=code) also seems fine but will change things for end-users who use that script.
- Renaming [telemetry_hopper.py](https://github.com/search?q=repo%3Amicrosoft%2Fazure-linux-image-tools+telemetry_hopper.py&type=code) will break the calls in entrypoint.sh and run.sh and needs updating.

.pipelines/containerSourceData/imagecustomizer/Dockerfile-imagecustomizer
```diff
@@ -0,0 +1,33 @@
...
+       echo "Skipping AMD64-specific packages for $TARGETARCH"; \
+   fi && \
+   tdnf clean all
+
+# Create virtual environment and install Python dependencies for telemetry
+RUN python3 -m venv /opt/telemetry-venv
+RUN /opt/telemetry-venv/bin/pip install --no-cache-dir -r /imagecustomizer-telemetry-requirements.txt
+
+# Copy remaining necessary files
+COPY .mariner-toolkit-ignore-dockerenv /
```
cwize1: FYI: This file was removed in https://github.com/microsoft/azure-linux-image-tools/pull/372

SPECS/azurelinux-image-tools/azurelinux-image-tools.spec
liulanze: I have kept the name `telemetry_hopper.py` as is rather than `imagecustomizer-telemetry_hopper.py`, since this script will be called in `run.sh` and `entrypoint.sh`, while in both of these files the name has remained `telemetry_hopper.py`.

PR #14494, opened 2025-08-12
toolkit: enable downloads using managed identities

ms-mahuber:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

This change introduces a fallback method for the toolkit's download functionality. When a download fails and an Azure client ID was supplied, a fallback attempt is executed if the download URI is pointing to an Azure Blob Storage. A new parameter is introduced to supply the Azure client ID. This new parameter is used for the SRPM packacking functionality, meaning that the fallback path can be executed for SRPM downloads. With this, fetching tarballs from authenticated blob storages (where HTTP GET is expected to fail) is enabled via the fallback path.

Example output when HTTP GET fails:
```
INFO[0004][srpmpacker] Packing 1/1 SPECs 
INFO[0004][srpmpacker] Attempting Azure SDK download for blob: ***/***/kata-containers-3.18.0.kata0-cargo.tar.gz from storage account: ***
INFO[0004][srpmpacker] Using ManagedIdentityCredential with supplied client ID for managed identity access: *** 
INFO[0006][srpmpacker] download time: 1.666248922s 
INFO[0006][srpmpacker] Azure SDK fallback succeeded for URL: https://***.blob.core.windows.net/***/***/kata-containers-3.18.0.kata0-cargo.tar.gz 
```

###### Change Log  

- Add new AZURE_CLIENT_ID variable to `Makefile`
- Use new variable in `srpm_pack.mk` and `srpmpacker.go`
- Implement fallback path using the new variable in `downloader.go`, `network.go`, `azureblobstorage.go`
- Add test to `network_test.go` showing that the fallback path is being attempted in case HTTP GET fails
- Add documentation of the new parameter to `building.md`

**YES/NO**

###### Associated issues  

- N/A

###### Links to CVEs  
- N/A

###### Test Methodology

- Pipeline build id: 893051 (a buddy build run using a storage account where anonymous access is disabled)
- Manual execution of `network_test.go`, manual inspection of pipeline runs
- Pipeline build id: 894094 (end-to-end run of toolkit capabilities, chaining up to id 894456)
- Pipeline build id: 891685 (a buddy build run using a public storage account; shows this path is not regressed/changed)


PawelWMS (commented):
This looks like a big change to the toolkit. What is the scenario we're trying to satisfy and what is currently not working without it?

toolkit/tools/internal/network/network.go
```diff
@@ -117,11 +133,49 @@ func DownloadFileWithRetry(ctx context.Context, srcUrl, dstFile string, caCerts
...
 
 // DownloadFile downloads `url` into `dst`. `caCerts` may be nil. If there is an error `dst` will be removed.
-func DownloadFile(ctx context.Context, url, dst string, caCerts *x509.CertPool, tlsCerts []tls.Certificate) (err error) {
+// First attempts HTTP download, then falls back to Azure SDK if the URL is an Azure Blob Storage URL.
+func DownloadFile(
+	ctx context.Context,
+	url, dst string,
+	caCerts *x509.CertPool,
+	tlsCerts []tls.Certificate,
+	azureClientID string,
```
PawelWMS: Please don't bundle up blob store access with a simple HTTP download into this function. This function is meant to handle one simple case. If we need to include Azure blob store integration, then we should do this on a higher level code.

toolkit/tools/internal/network/network.go
```diff
@@ -117,11 +133,49 @@ func DownloadFileWithRetry(ctx context.Context, srcUrl, dstFile string, caCerts
...
+	if azureErr != nil {
+		return buildResponseErrorSDKFallback(err, azureErr)
+	}
+
+	logger.Log.Infof("Azure SDK fallback succeeded for URL: %s", url)
+	return nil
+}
+
+// downloadFileHttp attempts to download a file using HTTP with optional TLS configuration.
+func downloadFileHttp(
```
PawelWMS:
```suggestion
func downloadFileHTTP(
```

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
```diff
@@ -136,7 +187,20 @@ func Create(tenantId string, userName string, password string, storageAccount st
 
 	} else if authenticationType == ManagedIdentityAccess {
 
-		credential, err := azidentity.NewDefaultAzureCredential(nil)
+		var credential azcore.TokenCredential
+
+		if azureClientID == "" {
+			logger.Log.Infof("Using DefaultAzureCredential for managed identity access")
+			credential, err = azidentity.NewDefaultAzureCredential(nil)
```
PawelWMS: Heads-up that usage of `NewDefaultAzureCredential` is forbidden according to latest [S360s](https://vnext.s360.msftcloudes.com/blades/security?blade=AssignedTo:All~KPI:a0f0ce42-3063-5d3b-3b47-1ff3143abdc9~SLA:2~Forums:All~Program:68556099-a3e2-472e-8da0-e1d1b000eda4;65a010a5-1e3d-4777-bb89-f149470a507d~waves:All~Tab:Summary~_loc:Security&peopleBasedNodes=joslobo_team;mthotad_team;chrco_team;japerrin_team&global=4:2b6a31bb-e65d-4c82-b632-a5a4ae663c82) we've been getting (see: [this PR](https://github.com/microsoft/azurelinux/pull/14477) for instance). Why do we even need to pass a specific client ID, though? Would `azidentity.NewAzureCLICredential(nil)` combined with logging into Azure before running the toolkit work? It's what we do across our pipelines and seems to be even the case for AZ CLI commands.

toolkit/tools/internal/network/network.go
```diff
@@ -117,11 +133,49 @@ func DownloadFileWithRetry(ctx context.Context, srcUrl, dstFile string, caCerts
...
+	err = downloadFileHttp(ctx, url, dst, caCerts, tlsCerts)
+	if err == nil {
+		return nil // HTTP download succeeded
+	}
+
+	// HTTP download failed, attempt to parse as Azure Blob Storage URL
+	blobInfo, parseErr := azureblobstorage.ParseAzureBlobStorageURL(url)
+	if parseErr != nil {
+		return buildResponseErrorSDKFallback(err, parseErr)
+	}
```
PawelWMS: Instead of picking the download method this way, what do you think about adding a new config option like `SOURCE_STORE_TYPE`, which would upfront and clearly indicate how we're expecting to be able to download the sources. Right now, I'm guessing that would have two options: one for the current simple "file server" and another for an Azure blob store requiring authentication.

manuelh-dev:
> fetching tarballs from authenticated blob storages (where HTTP GET is expected to fail)

Per description, this provides the capability of *fetching tarballs from authenticated blob storages (where HTTP GET is expected to fail)*. Please let me know if I should open a GitHub issue or clarify the PR description.

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
manuelh-dev: I am not introducing calls to `NewDefaultAzureCredential`. This code has been present before my change. This change introduces the capability to call `NewAzureCLICredential`.

Using `NewAzureCLICredential` w/o specifying a client ID results in the following error:
```
failed to authenticate a system assigned identity. The endpoint responded with {"error":"invalid_request","error_description":"Multiple user assigned identities exist, please specify the clientId / resourceId of the identity in the token request"})
```

Similar, when calling `NewDefaultAzureCredential`, I can't reliably acquire a token, see:
```
DefaultAzureCredential: failed to acquire a token. 
Attempted credentials: 
EnvironmentCredential: missing environment variable AZURE_TENANT_ID 
WorkloadIdentityCredential: no client ID specified. Check pod configuration or set ClientID in the options 
ManagedIdentityCredential: failed to authenticate a system assigned identity. The endpoint responded with {"error":"invalid_request","error_description":"Multiple user assigned identities exist, please specify the clientId / resourceId of the identity in the token request"} 
AzureCLICredential: Azure CLI not found on path 
AzureDeveloperCLICredential: Azure Developer CLI not found on path.
```

Hence, we will want to provide the client ID explicitly.

toolkit/tools/internal/network/network.go
manuelh-dev: Per prior discussions, the goal was to reduce risk to existing logic by introducing a fallback path this way. If this has changed and we want to provide a `SOURCE_STORE_TYPE` instead, happy to do this.

toolkit/tools/internal/network/network.go
manuelh-dev: What is the proper higher level code primitive you were thinking of?

manuelh-dev: Decision to not implement fallback but to add a second download method instead

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
manuelh-dev: Decision to use Azure CLI credentials and to map the credential into chroot environment. Based this on top of new 3.0-dev tip which enables use of NewAzureCLICredential call

toolkit/tools/internal/network/network.go
manuelh-dev: We decided to add a new download method. We can refine nomenclature from here on.

manuelh-dev:
> > fetching tarballs from authenticated blob storages (where HTTP GET is expected to fail)
> 
> Per description, this provides the capability of _fetching tarballs from authenticated blob storages (where HTTP GET is expected to fail)_. Please let me know if I should open a GitHub issue or clarify the PR description.

After receiving feedback, this PR has been reworked and the description updated.

toolkit/Makefile
```diff
@@ -25,6 +25,8 @@ PACKAGE_REBUILD_LIST    ?=
 PACKAGE_IGNORE_LIST     ?=
 ##help:var:SRPM_PACK_LIST:<spec_list>=List of space-separated spec folders inside "SPECS_DIR" to analyze for the build. If empty, all items inside the "SPECS_DIR" will be analyzed. Example: SRPM_PACK_LIST="kernel go which".
 SRPM_PACK_LIST          ?=
+##help:var:SRPM_DOWNLOAD_MODE:<mode>=Mode to use for downloading SRPMs. Modes are define in the SRPM packager functionality.
+SRPM_DOWNLOAD_MODE      ?=
```
PawelWMS: This feels more closely connected with `SOURCE_URL` further down in the file. What do you think about re-naming it to `SOURCE_AUTH_MODE`?

toolkit/docs/building/building.md
```diff
@@ -370,12 +370,19 @@ Daily build packages are available via `DAILY_BUILD_ID`. Use `DAILY_BUILD_ID=lkg
...
 
-If supplying custom endpoints for source/SRPM/package servers, accessing these resources may require keys and certificates. The keys and certificates can be set using:
+If supplying custom endpoints for source/SRPM/package servers, accessing these resources may require authentication.
+Keys and certificates for TLS based authentication can be set using:
 
 ```bash
 sudo make image CONFIG_FILE="./imageconfigs/core-efi.json" CA_CERT=/path/to/rootca.crt TLS_CERT=/path/to/user.crt TLS_KEY=/path/to/user.key
 ```
 
+For SRPM packing (i.e., for retrieving package sources), an Azure CLI based identity can be used to access authenticated Azure blob storages:
```
PawelWMS: Clarity nit-pick:

```suggestion
For SRPM packing (i.e., for retrieving package sources), an Azure CLI login can be used to access Azure blob storages, which do not support anonymous access:
```

toolkit/docs/building/building.md
```diff
@@ -370,12 +370,19 @@ Daily build packages are available via `DAILY_BUILD_ID`. Use `DAILY_BUILD_ID=lkg
...
+```bash
+sudo make build-packages SRPM_DOWNLOAD_MODE="azurecli"
+```
+Using this ID requires prior `az login` with your managed identity ID.
+
 ## Building Everything From Scratch
 
 **NOTE: Source files must be made available for all packages. They can be placed manually in the corresponding SPEC/\* folders, `SOURCE_URL=<YOUR_SOURCE_SERVER>` may be provided, or DOWNLOAD_SRPMS=y may be used to use pre-packages sources. Core Azure Linux source packages are available at `SOURCE_URL=https://azurelinuxsrcstorage.blob.core.windows.net/sources/core`**
```
PawelWMS:
```suggestion
**NOTE: Source files must be made available for all packages. They can be placed manually in the corresponding SPEC/\* folders, `SOURCE_URL=<YOUR_SOURCE_SERVER>` may be provided, or DOWNLOAD_SRPMS=y may be used to use pre-packages sources. Core Azure Linux source packages are available at `SOURCE_URL=https://azurelinuxsrcstorage.blob.core.windows.net/sources/core` and support anonymous access.**
```

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -297,6 +325,7 @@ func createChroot(workerTar, buildDir, outDir, specsDir string) (chroot *safechr
 		outMountPoint    = "/output"
 		specsMountPoint  = "/specs"
 		buildDirInChroot = "/build"
+		azureMountPoint  = "/root/.azure"
```
PawelWMS: Please use the value of the env var `AZURE_CONFIG_DIR` with a default `~/.azure` if it's not set.

toolkit/tools/internal/network/network.go
```diff
@@ -172,6 +206,39 @@ func DownloadFile(ctx context.Context, url, dst string, caCerts *x509.CertPool,
 	return
 }
 
+// downloadFileWithAzureSDK attempts to download a file from an Azure Blob Storage using the Azure SDK for Go
+func downloadFileWithAzureSDK(
```
PawelWMS: I'd prefer to keep `network.go` single-focused on direct HTTP GET downloads instead of making it a package for all kinds of downloads happening through the one `DownloadFileWithRetry` API. This will turn it into a monster function with a lot of complexity and arguments, where most of them will be empty/nil. Since there's already the `azureblobstorage` package, I'd use that directly in the SRPM packer if we determine we need to use this download method. An extra benefit would be that we can run `azureblobstorage.Create` only once, which would save us time on whatever set-up that call ends up doing.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -306,6 +335,18 @@ func createChroot(workerTar, buildDir, outDir, specsDir string) (chroot *safechr
...
+	// Adding the .azure mount ensures the chroot environment can access CLI credentials
+	if useAzureCliAuth {
+		homeDir, homeErr := os.UserHomeDir()
+		if homeErr == nil {
+			azureDir := filepath.Join(homeDir, ".azure")
+			extraMountPoints = append(extraMountPoints, safechroot.NewMountPoint(azureDir, azureMountPoint, "", safechroot.BindMountPointFlags, ""))
+		} else {
+			err = fmt.Errorf("Could not determine user home directory for .azure mount: %v", homeErr)
+			return
+		}
```
PawelWMS: Could you extract this logic into a separate function as well? Similar to what you did with `installAzureCliPackage`.

toolkit/tools/internal/network/network.go
manuelh-dev: Agree'd. I will add a function`DownloadFileFromURL` to `azureblobstorage.go`.

Question 1: In `network.go`, we have `DownloadFileWithRetries`. In `azureblobstorage.go`'s new `DownloadFileFromURL` function, do you also want to download with retries, or would you say retries is only a thing for `HTTP GET`?

Question 2: In order to ensure we call `azureblobstorage.Create` only once, I would move this call right in front of the `for` loop at the top of `hydrateFromRemoteSource`. Does that sound good?

toolkit/docs/building/building.md
```diff
@@ -840,6 +847,7 @@ To reproduce an ISO build, run the same make invocation as before, but set:
 | CA_CERT                       |                                                                                                          | CA cert to access the above resources, in addition to the system certificate store
 | TLS_CERT                      |                                                                                                          | TLS cert to access the above resources
 | TLS_KEY                       |                                                                                                          | TLS key to access the above resources
+| SOURCE_AUTH_MODE              |                                                                                                          | Enables managed identity based authentication for accessing package sources from protected Azure Blob Storages. This is useful when package sources are located in storage accounts prohibiting public access.
```
PawelWMS: I'd also describe each of the available options.

toolkit/Makefile
```diff
@@ -23,6 +23,8 @@ PACKAGE_BUILD_LIST      ?=
 PACKAGE_REBUILD_LIST    ?=
 ##help:var:PACKAGE_IGNORE_LIST:<spec_list>=List of space-separated spec folders to ignore during the build. Must not overlap with "PACKAGE_REBUILD_LIST", may overlap with "PACKAGE_BUILD_LIST". Example: PACKAGE_IGNORE_LIST="zlib".
 PACKAGE_IGNORE_LIST     ?=
+##help:var:SOURCE_AUTH_MODE:<mode>=Mode to use for downloading SRPMs. Modes are define in the SRPM packager functionality.
```
PawelWMS: I'm not sure if the comment is correct. We're downloading source files, not SRPMs themselves. And I think it would be worth to mention the valid options here as well.

toolkit/docs/building/building.md
```diff
@@ -370,15 +370,22 @@ Daily build packages are available via `DAILY_BUILD_ID`. Use `DAILY_BUILD_ID=lkg
...
 
 ```bash
 sudo make image CONFIG_FILE="./imageconfigs/core-efi.json" CA_CERT=/path/to/rootca.crt TLS_CERT=/path/to/user.crt TLS_KEY=/path/to/user.key
 ```
 
+For SRPM packing (i.e., for retrieving package sources), Azure CLI login can be used to access authenticated Azure blob storages, which do not support anonymous access:
+```bash
+sudo make build-packages SOURCE_AUTH_MODE="azurecli"
+```
+Using this ID requires prior `az login` with your managed identity ID.
```
PawelWMS: Maybe?

```suggestion
Using this method requires prior `az login` with your managed identity ID.
```

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
```diff
@@ -22,6 +25,47 @@ const (
...
+// Example for blob (similar response and code for container case):
+// RESPONSE 404: 404 The specified blob does not exist.
+// ERROR CODE: BlobNotFound
+var ErrDownloadInvalidResponse404 = errors.New("RESPONSE 404")
+
+// ParseAzureBlobStorageURL parses an Azure Blob Storage URL and extracts storage account, container, and optionally blob information.
+func ParseAzureBlobStorageURL(urlStr string) (storageAccountName, containerName, blobName string, err error) {
+	parsedURL, err := url.Parse(urlStr)
+	if err != nil {
+		return "", "", "", fmt.Errorf("failed to parse URL: %w", err)
```
PawelWMS:
```suggestion
		return "", "", "", fmt.Errorf("failed to parse URL:\n%w", err)
```

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
```diff
@@ -22,6 +25,47 @@ const (
...
+	}
+
+	if !strings.HasSuffix(parsedURL.Host, ".blob.core.windows.net") {
+		return "", "", "", fmt.Errorf("not a common Azure Blob Storage URL format " +
+			"(expected <storage_account>.blob.core.windows.net)")
+	}
+
+	// Extract storage account from hostname (e.g., "mystorageaccount.blob.core.windows.net")
+	hostParts := strings.Split(parsedURL.Host, ".")
+	if len(hostParts) < 4 {
```
PawelWMS: Shouldn't there be at least 5 parts after the split:
- [storage name],
- "blob",
- "core",
- "windows",
- "net".

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
```diff
@@ -148,3 +192,88 @@ func Create(tenantId string, userName string, password string, storageAccount st
...
+
+	logger.Log.Infof("Attempting Azure SDK download for blob: %s/%s", containerName, blobName)
+
+	retryNum := 1
+	errorWas404 := false
+	wasCancelled, err = retry.RunWithDefaultDownloadBackoff(ctx, func() error {
+		netErr := azureBlobStorage.Download(ctx, containerName, blobName, dstFile)
+		if netErr != nil {
+			// Check if the error is a 404-like condition (blob or container not found)
+			if errors.Is(netErr, ErrDownloadInvalidResponse404) {
```
PawelWMS: When do we set an error to `ErrDownloadInvalidResponse404` or wrap it in another error?

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -354,9 +391,78 @@ func createChroot(workerTar, buildDir, outDir, specsDir string) (chroot *safechr
...
 
+// addAzureConfigMountPoint appends a mount point for the Azure CLI config directory.
+func addAzureConfigMountPoint(extraMountPoints []*safechroot.MountPoint) ([]*safechroot.MountPoint, error) {
+	const (
+		chrootAzureConfigMountPoint = "/root/.azure"
+	)
+
+	// The variable is propagated into chroot, if set
+	azureConfigDir := os.Getenv("AZURE_CONFIG_DIR")
+	mountPoint := azureConfigDir
```
PawelWMS, line 414: Does this work if `AZURE_CONFIG_DIR` is set? It seems that then `mountPoint` has the same value, so we mount the directory inside the chroot under the home path of the host user, that probably doesn't exist inside the chroot.

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
```diff
@@ -22,6 +25,47 @@ const (
...
+
+	// Extract storage account from hostname (e.g., "mystorageaccount.blob.core.windows.net")
+	hostParts := strings.Split(parsedURL.Host, ".")
+	if len(hostParts) < 4 {
+		return "", "", "", fmt.Errorf("invalid Azure Blob Storage hostname format")
+	}
+	storageAccountName = hostParts[0]
+
+	// Extract container name and optionally blob name from path (e.g., "/container" or "/container/folder/blob")
+	pathParts := strings.Split(strings.Trim(parsedURL.Path, "/"), "/")
```
PawelWMS: Suggestion to move the validation and extraction of URL parts into a regular expression. This way the extraction logic should be shorter and any future changes would most likely be limited to one or two lines where we tweak the regular expression.

See full suggestion in #14612.

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
manuelh-dev: Changed via: https://github.com/microsoft/azurelinux/pull/14612

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
manuelh-dev: Changed via: https://github.com/microsoft/azurelinux/pull/14612

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
manuelh-dev: Thank you for this suggestion - merged.

toolkit/tools/internal/azureblobstorage/azureblobstorage.go
manuelh-dev: Thanks for catching this. I have modified the code back to what I originally intended to use. For some reason I was under the assumption that `errors.Is` evaluates to true if the message of the `ErrDownloadInvalidResponse404` error would be included in the error message received in `netErr`. This is not the case (internal build ID: 918820), hence this condition does not evaluate to true in a 404-like condition and iterate on the retry loop until the end.

With the fixed version, we properly detect the 404-like condition and break out of the loop (internal build ID: 91882).

Please let me know if you envision a better solution.

toolkit/tools/srpmpacker/srpmpacker.go
manuelh-dev: It works. The environment variable is propagated into the chroot environment, e.g, the value `home/cloudtest/.azure` when invoking the toolkit using `sudo` as the `cloudtest` user.

This folder does not exist in the chroot environment (running as `/root`), but the chroot initialization logic executes `osMkdir.All` in `createMountPoints` in `safechroot.go`, ensuring the folder exists and that creating the mount point succeeds.

PawelWMS: I think that another piece of the puzzle, which makes this work, is that the `AZURE_CONFIG_DIR` env variable may be set in the chroot as well, so since the host root's env value is now always set as well, this info is propagated into the chroot.

PawelWMS (approved):

PR #14612, opened 2025-09-04
Proposed tweaks.

PawelWMS:
Just some proposed tweaks instead of comments in the original PR.


manuelh-dev:
Thank you! I ran some internal validation (IDs 918813, 918814), looking good, merging.

PR #14614, opened 2025-09-04
[High] Upgrade reaper for CVE-2025-9288

durgajagadeesh:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

[High] Upgrade reaper for CVE-2025-9288
Astrolabe reference:
Upstream:

###### Change Log  

- Modified file;
- Modified file;
- Modified file;

**NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2025-9288

###### Test Methodology

- Upgrade for CVE fix has done cleanly


durgajagadeesh:
Hi,
This PR is ready for review, built successful in my local VM.
image

I have removed several outdated patches after confirming that their changes are already present in the newly generated tarball.
image

All patches have been applied cleanly.
image

And the CVE-2025-9288 fix has already been included in the latest tarball generated by the reaper_build_script.sh script. 

Note: I have the below reaper.signature.json changes while this got success!
image

And build log from the local VM: 
[reaper-3.1.1-20.cm2.src.rpm-10-09-2025-success_build.log](https://github.com/user-attachments/files/22254463/reaper-3.1.1-20.cm2.src.rpm-10-09-2025-success_build.log)

-Thank you!

durgajagadeesh:
Hi,
I have completed the tests for reaper using the resultant rpm binary(reaper-3.1.1-20.cm2.x86_64.rpm).
Please find the reference screenshot below.

image

image
image
image

Note: Reaper web interface at http://localhost:8080 is running successfully.
-Thank you!

Malateshk007:
@realsdx, gentle reminder for review and sign-off if all good.!

realsdx:
On a fresh azurelinux system the script fails with this:
Screenshot 2025-09-10 114047

Please include `nodejs-npm` as well as the dependency in the script

Also, since you have improved the script, please copy all the tar files in the current working directory instead of copying to `reaper_caches`

durgajagadeesh:
> On a fresh azurelinux system the script fails with this: Screenshot 2025-09-10 114047
> 
> Please include `nodejs-npm` as well as the dependency in the script
> 
> Also, since you have improved the script, please copy all the tar files in the current working directory instead of copying to `reaper_caches`

Hi @realsdx ,
    I have implemented the updates as you suggested, and the script is functioning correctly with these modifications.
    
image

If there’s anything else that needs to be updated, please let me know. 
-Thank you!

SPECS/reaper/reaper.signatures.json
```diff
@@ -2,9 +2,9 @@
  "Signatures": {
   "cassandra-reaper-3.1.1.tar.gz": "6efe52195ad4a3c3b7a6f928bafa60d3df011709d9bc918e717033bf86d724d8",
   "reaper-bower-components-3.1.1-1.tar.gz": "51f5b03b3f56966f5fbfe28a13e0a74003cf33372ff4ba13fd82c6fe79092033",
-  "reaper-local-lib-node-modules-3.1.1.tar.gz": "8daf9a8726a85ca31b024a5bab60a357fe927f670908955cdd9b106bf9c6bd60",
+  "reaper-local-lib-node-modules-3.1.1.tar.gz": "043dec92e8d1d07bbcb920f0f10a148d63c600f2101935f3c39f4654d470135d",
```
PawelWMS, line 5: This looks fishy - we can't be overwriting sources in our blob store. We must use new file names instead.

SPECS/reaper/reaper_build_caches.sh
```diff
@@ -61,22 +61,45 @@ function checkInternet {
...
+	env | grep -i npm
+
+	# Install and use Node.js v14.18.0 with n
 	npm install -g n
 	export N_PREFIX="$homeCacheDir/.npm-global"
 	n 14.18.0
-	export XDG_CACHE_HOME=$homeCacheDir/.cache
+
+	# Ensure the shell uses the new node and npm
+	export PATH="$homeCacheDir/.npm-global/bin:$PATH"
```
PawelWMS: Why the second export of `PATH`?

SPECS/reaper/reaper_build_caches.sh
```diff
@@ -147,8 +170,7 @@ buildReaperSources
 
 createCacheTars
 
-mkdir "$HOME/reaper_caches"
 
-cp -a ${reaperCacheDir} "$HOME/reaper_caches"
+cp -a ${reaperCacheDir}/*.tar.gz .
 
-echo "Copied cache tars to $HOME/reaper_caches/ .Exiting."
+echo "Copied cache tars to SPECS/reaper/ directory"
```
PawelWMS: I wouldn't change the output directories. I'm no expert on our auto-upgrade pipelines but they may this script might be used by them and they'd likely expect the output in the same place as it's been so far.

SPECS/reaper/reaper_build_caches.sh
```diff
@@ -61,22 +61,45 @@ function checkInternet {
...
-	# Backward compatible with node@v14.18.0
-	# When installing modules via npm to default prefix='/usr/local' in mariner, the permissions for 'others'
-	# is incoorectly set that causes 'which' to still point to older path, as access/newfstatat fail with -ENOPERM
-	# Setting a new global npm folder for fixing permission issues.
-	# (works well with id=0, but reaper build will fail.)
-	mkdir --mode 0777 $homeCacheDir/.npm-global
+	sudo tdnf install -y nodejs npm
+
+	#Verify installation
+	echo "Verifying npm installation..."
```
PawelWMS: Could you also explain in the summary why we needed to modify this script? I'm guessing it might have not been working on AZL 3.0. If that's the case, could you check that the updated version still works on 2.0? Since it's a 2.0 package, it's possible someone or some automation is still using it on a 2.0 image.

durgajagadeesh: Hi @PawelWMS ,
Initially, the upstream patch included a dependency on the to-buffer package. After consulting with Sumedh, I was advised to run the reaper_build_caches.sh script to generate the new cache tarball, which would include the to-buffer dependency package. I proceeded to run the script, but I am encountering an error.: EACCESS: permission denied, 'make /usr/etc'.  
image

I followed up with Sumedh regarding the error mentioned above, and he suggested that we needed to address the script issue. I worked on the script and fixed it by changing the directory to home instead of creating it in /usr/etc.


The script is working as expected on the 3.0 image. Please see the attached reference screenshot below.
image


I have also verified its operation on the 2.0 image, and the reference screenshot is attached below.
image

-Thank you!

SPECS/reaper/reaper.signatures.json
durgajagadeesh: Is that okay if I add release number(20) along with version(3.1.1). I mean the final name will be like reaper-local-lib-node-modules-3.1.1-20.tar.gz"

SPECS/reaper/reaper_build_caches.sh
durgajagadeesh: Previously, the script generated a new tarball only when manually triggered, and we would run it solely when planning to update the source code. I confirmed this process with Sumedh.

Additionally, The tarball directory has now been changed from $HOME/reaper_caches/ to SPECS/reaper/ in response to following Pandit's @realsdx's request. I updated the script so the tarball is created in this directory.

  Please let me know if you require any further modifications to revert to the previous directory path.

SPECS/reaper/reaper_build_caches.sh
durgajagadeesh: Hi @PawelWMS,
The second export PATH is essential because it ensures that the shell uses the correct versions of node and npm installed in the custom location ($homeCacheDir/.npm-global/bin), rather than any system-wide versions that may be outdated or missing.
-Thank you!

SPECS/reaper/reaper_build_caches.sh
PawelWMS: Thanks for the explanation, @durgajagadeesh.

@Sumynwa, can you confirm this script is not used in any of our automation (example: the [auto-upgrade workflow](https://dev.azure.com/mariner-org/mariner/_build?definitionId=1770))? Or if it is, that it's going to be able to handle the new output path?

SPECS/reaper/reaper_build_caches.sh
PawelWMS: Thank you, that's all I needed to know.:)

SPECS/reaper/reaper_build_caches.sh
PawelWMS: Sorry, I still don't understand. I may be missing something or don't know enough about Bash. The first `export PATH` on line 79 above already prepends `$homeCacheDir/.npm-global/bin` to the `PATH` variable. The export here seems to be prepending that path again and that's why I think we're not changing anything at this point.

durgajagadeesh: Hi @PawelWMS,
 The **second export** of PATH is necessary because, after installing **Node.js v14.18.0** with **n,** the binaries for **node** and **npm** in $homeCacheDir/.npm-global/bin may be updated or replaced. By re-exporting PATH, we ensure that our shell session recognizes and uses these new binaries for all subsequent commands, preventing version conflicts and guaranteeing that tools like bower and npm work as intended.
 
  Please let me know if I need to revert it back, However it May use old or system node/npm instead of newly installed node/npm by n ( v14.18.0, may have changed binaries).
  
  
image

-Thank you!

durgajagadeesh: > Sorry, I still don't understand. I may be missing something or don't know enough about Bash. The first `export PATH` on line 79 above already prepends `$homeCacheDir/.npm-global/bin` to the `PATH` variable. The export here seems to be prepending that path again and that's why I think we're not changing anything at this point.

Hi @PawelWMS,
  If you would like a summary of the requirements for these two export commands, please see the information below.
  
      The first export command (around line 79) is necessary, but it’s not always sufficient by itself. Here’s why:

What Happens After the **First Export**?

   The first export PATH="$homeCacheDir/.npm-global/bin:$PATH" ensures your shell uses the custom-installed node and npm binaries for all npm operations and global installs up to that point.

Why Isn’t It Enough?

After this, The script runs "n 14.18.0" in line number 78, which downloads and installs a new version of Node.js and npm into $homeCacheDir/.npm-global/bin.
The contents of that directory (the actual binaries) change at this step.
Some shells may cache the location of executables, and even with hash -r, if the PATH isn’t re-exported, the shell might not pick up the new binaries immediately.


The **Second Export** Ensures:
    The shell session is “reminded” to look in the right place for the freshly installed Node.js and npm.
All subsequent commands (like bower, npm, etc.) use the correct, updated versions.

**In summary:**
The first export sets up the environment for the initial npm operations.
The second export is a safety step after installing a new Node.js version, guaranteeing the shell uses the new binaries just installed.

Best practice: Keep both exports to avoid subtle bugs or version mismatches, especially in scripts that install or switch Node.js versions dynamically.

-Thank you!

durgajagadeesh:
Hi @Sumynwa, 
I have added an additional medium severity patch, CVE-2021-23445, to this PR at Kanishk's suggestion. This patch has been backported and I have successfully tested it on my local VM.

image

image 

I have also successfully tested the binary RPM on the cbl-mariner OS.
image

Please let me know if there is anything else I need to do.
Thank you!

SPECS/reaper/reaper_build_caches.sh
PawelWMS: Wow, I was not aware of the cache maintained by Bash. Thanks for teaching me @durgajagadeesh. I couldn't find a case where `hash -r` wouldn't be sufficient to make sure we re-search `$PATH` but I think it doesn't matter too much anymore.

SPECS/reaper/reaper_build_caches.sh
Sumynwa: @PawelWMS Reviewed the `reaper_build_caches.sh` change and we don't require them, apart from 2 changes which addresses the PATHs to fix the permission error.
Also, we have now reverted the changes to reaper caches path to keep the original.

durgajagadeesh:
Hi @Sumynwa,
I've updated the script based on your suggestion and tested it successfully. Please see the screenshots below.

I have tested without any issues.

1) The script has been working with those changes.
image

 The reaper script changes are provided as below.
image
image

2. Build is success, patches are applied cleanly.

image
image
image

3.Tested in 2.0 image
image

image

Please let me know if any changes need to be incorporated.
-Thank you!.

SPECS/reaper/reaper.signatures.json
Sumynwa: @durgajagadeesh Please follow [similar](https://github.com/microsoft/azurelinux/blob/c04e7d064a27cefd378bde61a0ef10e7e84d2edf/SPECS/reaper/reaper.spec#L4) to introduce something like `%define local_lib_modules_release 1`

SPECS/reaper/CVE-2021-23445.patch
```diff
@@ -0,0 +1,395 @@
...
+ 
+  This source file is free software, available under the following license:
+    MIT license - http://datatables.net/license
+@@ -9,170 +9,178 @@
+  or FITNESS FOR A PARTICULAR PURPOSE. See the license files for details.
+ 
+  For details please refer to: http://www.datatables.net
+- DataTables 1.10.21
+- ?2008-2020 SpryMedia Ltd - datatables.net/license
++ DataTables 1.11.2
```
Sumynwa, line 1872: @durgajagadeesh I don't have much js expertise, but how did we validate that we only need to update this single file's version, while keeping other sources to `1.10.21`?

durgajagadeesh: > @durgajagadeesh I don't have much js expertise, but how did we validate that we only need to update this single file's version, while keeping other sources to `1.10.21`?

Hi @Sumynwa,
   I have updated the latest CVE-2021-23445.patch[Medium] to include the upgraded DataTables source (version 1.11.2) and applied the CVE fixes. I have also successfully validated the repair tests on the CBL-Mariner 2.0 image.
 
1) Build passes in local VM:image

And patch applies cleanly
image


2) reaper service enabled in cbl-mariner with the generated reaper binary rpm.image

And the reaper tests were passed as shown in below.
image

Please let me know if any changes need to be incorporated.
-Thank you!.

SPECS/reaper/reaper_build_caches.sh
durgajagadeesh: > @PawelWMS Reviewed the `reaper_build_caches.sh` change and we don't require them, apart from 2 changes which addresses the PATHs to fix the permission error. Also, we have now reverted the changes to reaper caches path to keep the original.

Hi @Sumynwa,
I have implemented the script changes as you suggested.
Thank you for your input.

SPECS/reaper/reaper.signatures.json
durgajagadeesh: Hi @Sumynwa,

Could you please confirm whether the changes outlined below are acceptable?

image

-Thank you!

Sumynwa:
For information, in `src/server/pom.xml`, dependency `dropwizard` snapshot cannot be downloaded any more, which means the `reaper_build_caches.sh` is now broken:

`[ERROR] Failed to execute goal on project cassandra-reaper: Could not resolve dependencies for project io.cassandrareaper:cassandra-reaper:jar:3.1.1: Could not find artifact org.secnod.dropwizard:dropwizard-shiro:jar:0.3.0-SNAPSHOT in dependency-snapshots-repo (https://oss.sonatype.org/content/repositories/snapshots) -> [Help 1]`

Sumynwa:
Triggered [Buddy Build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=938676&view=results)

Sumynwa:
Signing off patch changes

Upstream reference
CVE-2025-9288: [Commit](https://github.com/browserify/sha.js/commit/f2a258e9f2d0fcd113bfbaa49706e1ac0d979ba5)
CVE-2021-23445: [Commit](https://github.com/DataTables/Dist-DataTables/commit/59a8d3f8a3c1138ab08704e783bc52bfe88d7c9b)

Toolchain Package: No

Exclusive_Arch: x86_64

Patch Backported: YES
Details:
CVE-2025-9288
- For this CVE, the cached source `reaper-srcui-node-modules` are updated using script `reaper_build_caches.sh`.
- With this update, the patch files patching `src/ui/node_modules` which were no longer needed (fixed in updated sources) have been removed

CVE-2021-23445
- Bumps DataTables from 1.10.21 -> 1.11.2
- Applies upstream patch on updated files

Patch Applies Cleanly: YES
x86_64
image

Sumynwa (approved):

SPECS/reaper/reaper_build_caches.sh
```diff
@@ -62,6 +62,7 @@ function checkInternet {
 function installNodeModules {
 	echo "Installing node modules."
 	sudo tdnf install -y nodejs
+	 npm config set prefix "$homeCacheDir/.npm"
```
PawelWMS: Nit-pick:

```suggestion
	npm config set prefix "$homeCacheDir/.npm"
```

SPECS/reaper/CVE-2021-23445.patch
```diff
@@ -0,0 +1,2214 @@
...
+ 2 files changed, 732 insertions(+), 777 deletions(-)
+
+diff --git a/src/ui/bower_components/datatables/media/js/jquery.dataTables.js b/src/ui/bower_components/datatables/media/js/jquery.dataTables.js
+index 6800c2d..e156f33 100644
+--- a/src/ui/bower_components/datatables/media/js/jquery.dataTables.js
++++ b/src/ui/bower_components/datatables/media/js/jquery.dataTables.js
+@@ -1,15 +1,15 @@
+-/*! DataTables 1.10.21
+- * ©2008-2020 SpryMedia Ltd - datatables.net/license
++/*! DataTables 1.11.2
```
sameluch, line 18: Can we do this version upgrade within the reaper_build_caches.sh instead of creating such a large patch file for the vendor tarball? We could update the version minimum before taring up the components to avoid having to apply the same upgrade from within a patch.

SPECS/reaper/CVE-2021-23445.patch
```diff
@@ -0,0 +1,2214 @@
...
+ .../datatables/media/js/jquery.dataTables.js  | 1165 ++++++++---------
+ .../media/js/jquery.dataTables.min.js         |  344 ++---
+ 2 files changed, 732 insertions(+), 777 deletions(-)
+
+diff --git a/src/ui/bower_components/datatables/media/js/jquery.dataTables.js b/src/ui/bower_components/datatables/media/js/jquery.dataTables.js
+index 6800c2d..e156f33 100644
+--- a/src/ui/bower_components/datatables/media/js/jquery.dataTables.js
++++ b/src/ui/bower_components/datatables/media/js/jquery.dataTables.js
+@@ -1,15 +1,15 @@
+-/*! DataTables 1.10.21
```
sameluch, line 16: I see on the CVE that the fix has a released version of 1.11.3 from the datatables upstream. Can we upgrade directly to 1.11.3 instead of applying the patch to the 1.11.2? This would make tracking the actual fix here a bit easier.

https://github.com/DataTables/Dist-DataTables/commit/59a8d3f8a3c1138ab08704e783bc52bfe88d7c9b

SPECS/reaper/reaper_build_caches.sh
durgajagadeesh: Hi @PawelWMS,
I have implemented the changes as per your suggestions.
Please let me know if any further modifications are required.
Thank you.

SPECS/reaper/CVE-2021-23445.patch
durgajagadeesh: Hi @Sumynwa,
The script is currently not functioning as expected. I will investigate the issue and provide you with an update as soon as possible.
Please, find the reference screenshot from the image 2.0 below.
image

Thank you.

SPECS/reaper/CVE-2021-23445.patch
durgajagadeesh: Hi @sameluch,
I have attempted to upgrade based on your suggestions.
However, as we discussed earlier, the reaper_build_caches.sh script isn’t working as expected.
The error is a Snapshot Dependency Issue:
Maven can’t find io.cassandrareaper.dropwizard:dropwizard-shiro:jar:0.3.0-SNAPSHOT in any repository. Since this is a snapshot (development) version.

Recent Breakage:
Since my script worked 10 days ago, it’s likely that the snapshot artifact was available then, but has since been deleted, moved, or replaced in the upstream repository. This is common with snapshot dependencies, which are not guaranteed to be stable or persistent.

Note: I checked the reaper issues page at https://github.com/thelastpickle/cassandra-reaper/issues?page=8 for a solution to this error, but it appears to be a newly introduced issue and does not exist in the list.

@sameluch, It would be helpful if you could also take a look and share any suggestions for alternative solutions.
-Thank you!

SPECS/reaper/CVE-2021-23445.patch
durgajagadeesh: > Hi @Sumynwa, The script is currently not functioning as expected. I will investigate the issue and provide you with an update as soon as possible. Please, find the reference screenshot from the image 2.0 below. image
> 
> Thank you.

Hi @sameluch,
I have attempted to upgrade within the reaper_build_caches.sh instead of creating such a large patch based on your suggestions.
However, as we discussed earlier, the reaper_build_caches.sh script isn’t working as expected.
The error is a Snapshot Dependency Issue:
Maven can’t find io.cassandrareaper.dropwizard:dropwizard-shiro:jar:0.3.0-SNAPSHOT in any repository. Since this is a snapshot (development) version.
I have tried to resolve the the script error However this is broken with snapshot version.
If  you could find any alternative solutions, then please let us know.
-Thank you!

PR #15357, opened 2025-12-19
Revert InfluxDB change.  PTEST regressed 

jslobodzian:
…65637 [HIGH] (#15267)"

This reverts commit 34f1332dfc8fde944c38b0339bfb57c7b1f23068.

PR #15370, opened 2025-12-22
Cherry-Pick revert of influxdb CVE-2025-65637 (#15357)

jslobodzian:
A fix was submitted to the fasttrack/2.0 branch for influxdb that had two back-to-back ptest failures.  The [original change](https://github.com/microsoft/azurelinux/pull/15253) was automatically cherry-picked to main [here](https://github.com/microsoft/azurelinux/pull/15320).

The change was reverted on fasttrack [here](https://github.com/microsoft/azurelinux/pull/15357) but the revert to main had merge conflicts.  This PR resolves the merge conflicts by bumping the release number and removing the bad patch.

### DIFF
--- a/SPECS/influxdb/influxdb.spec
+++ b/SPECS/influxdb/influxdb.spec
@@ -154,7 +154,7 @@ go test ./...
 
 %changelog
 * Mon Dec 22 2025 Jon Slobodzian <joslobo@microsoft.com> - 2.6.1-27
-- Revert Patch for CVE-2025-10543 as it breaks selftest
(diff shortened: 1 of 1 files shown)


SPECS/influxdb/influxdb.spec
PawelWMS, line 162: Since we removed this 2.6.1-25 changelog entry in the fast-track revert, I'd remove it here as well and drop the new one added in this PR. Otherwise, the changelogs will go out of sync between these two branches.

jslobodzian: Preferring this approach for cve scanning continuity.

PR #15499, opened 2026-01-14
fix: Add useful debug message when files are missing from config

dmcilvaney:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Improve error messages when image config files reference missing files (PackageLists, scripts, AdditionalFiles, etc.).

Previously, if a referenced file was missing, Make would fail with a confusing message about having no rule to build the validation flag. If the missing file was in a non-existent directory, `realpath` would also print unhelpful errors before the actual failure.

This was caused by Make's pattern rule matching behavior. The validation rules is `$(STATUS_FLAGS_DIR)/validate-image-config%.flag:`, which matches the variable `validate-config = $(STATUS_FLAGS_DIR)/validate-image-config-$(config_name).flag`

The rule had `config_other_files = $(if $(CONFIG_FILE),$(call shell_real_build_only, $(SCRIPTS_DIR)/get_config_deps.sh $(CONFIG_FILE)),)` as a dependency. If one of those calculated files was missing, then make would consider the pattern matched rule un-usable, and would try to find an alternate rule that did have all its dependencies. Unfortunately, we only have one pattern match rule, so we must ensure that all the dependencies we list are either real files, or valid make targets, so make won't skip it.

###### Change Log  

- Add early validation in `imggen.mk` to detect missing config dependencies and display a clear error message listing all missing files
- Use `realpath -m` in `get_config_deps.sh` to handle paths where parent directories don't exist
- Skip the missing file check for `printvar-*` targets so users can still debug

**NO**

###### Test Methodology

- Local testing with intentionally broken config files referencing missing files in both existing and non-existing directories
- pipeline: ~https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1025897&view=results~
- pipeline: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1028007&view=results

### DIFF
--- a/toolkit/scripts/utils.mk
+++ b/toolkit/scripts/utils.mk
@@ -15,6 +15,12 @@ build_arch := $(shell uname -m)
 
 no_repo_acl = $(STATUS_FLAGS_DIR)/no_repo_acl.flag
 
+# Define newline for use in error messages and output formatting
+define newline
+
+
+endef
+
 ######## MISC. MAKEFILE Functions ########
 
 # Creates a folder if it doesn't exist. Also sets the timestamp to 0 if it is
--- a/toolkit/scripts/get_config_deps.sh
+++ b/toolkit/scripts/get_config_deps.sh
@@ -44,6 +44,8 @@ do
 	then
 		echo "$filename"
 	else
-		echo $(realpath "$config_base_dir/$filename")
+		# Use -m to canonicalize paths even if they don't exist
+		# This allows the Makefile to detect missing files and provide a helpful error
+		echo $(realpath -m "$config_base_dir/$filename")
 	fi
 done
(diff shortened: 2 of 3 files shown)


toolkit/scripts/utils.mk
```diff
@@ -15,6 +15,12 @@ build_arch := $(shell uname -m)
 
 no_repo_acl = $(STATUS_FLAGS_DIR)/no_repo_acl.flag
 
+# Define newline for use in error messages and output formatting
+define newline
+
+
```
anphel31, line 21: does this intentionally include 2 newlines?

anphel31 (approved):

toolkit/scripts/utils.mk
dmcilvaney: Yes, make trims the last one off, so to get one newline, you need two.

PR #15578, opened 2026-01-26
Add support for kernel flavor versioning

mbykhovtsev-ms:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Modifies toolkit to allow dynamic versioning for the different kernel versions by creating a large macro file containing all of the release and versions for all SPEC files. 

###### Change Log  

- Modifies toolkit to allow dynamic versioning for the different kernel versions by creating a large macro file containing all of the release and versions for all SPEC files.

**NO**

###### Associated issues  

- #xxxx

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/toolkit/scripts/toolkit.mk
+++ b/toolkit/scripts/toolkit.mk
@@ -76,7 +76,7 @@ $(toolkit_archive_versioned_compressed): $(toolkit_archive) $(rpms_snapshot) $(d
 	tar --update -f $(toolkit_archive_versioned) -C $(toolkit_build_dir) $(toolkit_release_file_relative_path) $(toolkit_rpms_snapshot_file_relative_path) && \
 	$(ARCHIVE_TOOL) --best -c $(toolkit_archive_versioned) > $(toolkit_archive_versioned_compressed)
 
-$(toolkit_archive): $(go_tool_targets) $(mariner_repos_files) $(toolkit_component_extra_files) $(toolkit_root_files)
+$(toolkit_archive): $(go_tool_targets) $(mariner_repos_files) $(toolkit_component_extra_files) $(toolkit_root_files) $(rel_versions_macro_file)
 	rm -rf $(toolkit_prep_dir) && \
 	mkdir -p $(toolkit_prep_dir) && \
 	mkdir -p $(toolkit_repos_dir) && \


.github/workflows/validate-cg-manifest.sh
```diff
@@ -124,6 +124,9 @@ prepare_chroot_environment() {
     sudo cp -v "$macro_file" "$chroot_rpm_macros_dir_path"
   done
 
+  echo "Copying the version/release macros file to the chroot."
+  sudo cp -v "build/pkg_artifacts/macros.releaseversions" "$chroot_rpm_macros_dir_path"
```
PawelWMS: Could you pass the path to the macros file through an argument instead of hard-coding it in the script? Same as what we're doing with the path to the chroot tarball right now.

.github/workflows/check-source-signatures.yml
```diff
@@ -81,15 +81,26 @@ jobs:
...
       - name: Check for invalid source signatures
         run: |
           if [ -z "${{ env.changed_pkgs }}" ]; then
             echo "No package changes detected in '${{ matrix.specs-dir }}''."
             exit 0
           fi
 
+          MACROS_RELEASEVERSIONS_PATH=$(find . -type f -path "*/build/pkg_artifacts/macros.releaseversions" -print -quit)
+          if [ -z "$MACROS_RELEASEVERSIONS_PATH" ]; then
+            echo "build/pkg_artifacts/macros.releaseversions not found."
```
PawelWMS: Nit-pick: I'd either assume the macros file is expected in one specific place, hard-code the path here and drop the use of `find` or make the error message more generic, since we're not sure where the file is until we run `find`.

.github/workflows/check-source-signatures.yml
```diff
@@ -81,15 +81,26 @@ jobs:
 
           sudo make -C toolkit -j$(nproc) chroot-tools REBUILD_TOOLS=y DAILY_BUILD_ID=${LKG_BUILD_ID}
 
+          sudo make -C toolkit -j$(nproc) generate-versions-macros-file REBUILD_TOOLS=y DAILY_BUILD_ID=${LKG_BUILD_ID}
```
PawelWMS: I'd also add a comment explaining that this needs to be done for the sake of specs directories other than `SPECS` because all of them depend on the macros generated from the specs inside `SPECS`.

toolkit/Makefile
```diff
@@ -33,6 +33,8 @@ TEST_RUN_LIST           ?=
 TEST_RERUN_LIST         ?=
 ##help:var:TEST_IGNORE_LIST:<spec_list>=List of space-separated spec folders to ignore for package tests. Must not overlap with "TEST_RERUN_LIST", may overlap with "TEST_RUN_LIST". Example: TEST_IGNORE_LIST="acl".
 TEST_IGNORE_LIST        ?=
+##help:var:EXTRA_MACROS_FILES:<file1> <file2>=Space separated list of additional files whose contents will be appended to the versions macro file used during package builds. Example: EXTRA_MACROS_FILES="file1 file2".
+EXTRA_MACROS_FILES ?=
```
PawelWMS, line 37: Can you also add docs for this argument into `toolkit/docs/building/building.md`?

toolkit/scripts/tools.mk
```diff
@@ -62,6 +62,7 @@ go_tool_list = \
 	scheduler \
 	specarchchecker \
 	specreader \
+	versionsprocessor \
```
PawelWMS, line 65: Can you also add info about this new step in packages generation and what the new macros inside the specs are for into `toolkit/docs/how_it_works/3_package_building.md`?

One thing I'd make sure to mentioned in the docs: the `azl_<package_name>_release` macro does **NOT** contain the distro suffix. I understand this is by design, it's just something that we should be explicit about, since we're very inconsistent with what "release" means across our tooling and specs.

PawelWMS, line 65: Please also update the specs using the default kernel to use the new approach. The way of querying the `kernel-headers` for the kernel version is one of the workarounds we want to remove with this change and I'd prefer not to keep it around, since we have a better way now (prevents confusion). Also, some of the specs for default kernel OOTs also have a dependency on the default kernel `mlnx-ofa_kernel` and right now that's a hard-coded version inside these specs, so we should clean that up as well.

Examples:
- `mlnx-ofa_kernel.spec` (queries `kernel-headers`),
- `iser.spec` (queries `kernel-headers` **AND** has a hard-coded version of `mlnx-ofa_kernel`).

If you search for `--whatprovides kernel-headers` and `_mofed_full_version %{_version}` you should find all of them.

SPECS-SIGNED/xpmem-hwe-modules-signed/xpmem-hwe-modules-signed.spec
```diff
@@ -93,6 +95,9 @@ if [ $1 = 0 ]; then  # 1 : Erase, not upgrade
 fi
 
 %changelog
+* Mon Feb 10 2026 Mykhailo Bykhovtsev <mbykhovtsev@microsoft.com> - 2.7.4-29_6.12.57.1.2
```
PawelWMS: Nit-pick about the date: February 10<sup>th</sup> was a Tuesday.

toolkit/scripts/pkggen.mk
```diff
@@ -89,10 +92,29 @@ analyze-built-graph: $(go-graphanalytics)
...
+	$(go-versionsprocessor) \
+		--dir $(SPECS_DIR) \
+		--dist-tag $(DIST_TAG) \
+		$(logging_command) \
+		--build-dir $(parse_working_dir) \
+		--worker-tar $(chroot_worker) \
+		--cpu-prof-file=$(PROFILE_DIR)/versionsprocessor.cpu.pprof \
+		--mem-prof-file=$(PROFILE_DIR)/versionsprocessor.mem.pprof \
+		--trace-file=$(PROFILE_DIR)/versionsprocessor.trace \
+		$(if $(EXTRA_MACROS_FILES),$(foreach file,$(EXTRA_MACROS_FILES),--extra-macros-files=$(file))) \
```
PawelWMS: Argument name nit-pick: I understand it's one file per argument, so I'd drop the "s" and go with `extra-macros-file`. This will make it a bit more clear.

toolkit/scripts/pkggen.mk
```diff
@@ -315,6 +338,7 @@ $(STATUS_FLAGS_DIR)/build-rpms.flag: $(no_repo_acl) $(preprocessed_file) $(chroo
 		--distro-release-version="$(RELEASE_VERSION)" \
 		--distro-build-number="$(BUILD_NUMBER)" \
 		--rpmmacros-file="$(TOOLCHAIN_MANIFESTS_DIR)/macros.override" \
+		--versions-macro-file="$(rel_versions_macro_file)" \
```
PawelWMS, line 341: The `$(rpms_snapshot_per_specs)` target (and the `rpmssnapshot` tool itself) inside `toolkit.mk` will need the macros file as well. If I remember something more, I'll let you know, but in general, I'd look for all tools using the chroot and check, if they do any parsing or queries of the specs, because that's when the macros would need to be set.

toolkit/tools/pkg/specreaderutils/specreaderutil.go
```diff
@@ -167,6 +169,35 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
+
+		copyErr := file.Copy(releaseVersionMacrosFile, macrosDestFile)
+		if copyErr != nil {
+			logger.Log.Errorf("Failed to copy release version macros file into chroot (%s -> %s): %s", releaseVersionMacrosFile, macrosDestFile, copyErr)
+			return
+		}
+
+		logger.Log.Infof("Copied release version macros file into chroot (%s -> %s)", releaseVersionMacrosFile, macrosDestFile)
+	}
+
```
PawelWMS: Few things about this section:
- Very similar (same?) code is repeated in `pkgworker.go` and `srpmpacker.go`. Maybe other places as well - those are the 2 I saw right now. Please move this to one re-usable function. It could be a new public function for the `ChrootInterface` interface like `AddRPMMacrosFile(macrosFilePath string)`.
- The `rpm.GetMacroDir()` call is happening on the host, while we're adding the macros file inside the chroot. In that case, I'd call this function inside the chroot as well with `chroot.Run()`, so that the host's RPM config doesn't influence adding the macros under the right path.
- No need to manually create the directories and calling `file.Copy`. The `chroot.AddFiles()` function is already designed exactly for that.

toolkit/tools/scheduler/buildagents/definition.go
```diff
@@ -26,6 +26,7 @@ type BuildAgentConfig struct {
 	DistroReleaseVersion string
 	DistroBuildNumber    string
 	RpmmacrosFile        string
+	VersionsMacroFile    string
```
PawelWMS, line 29: In this case, I'd update `RpmmacrosFile` to `RPMMacrosFiles`, since it's already a mechanism for passing the macros files to the chroot. It just wasn't designed to handle more than one but I'd go down this route, instead of adding a new option.

toolkit/tools/srpmpacker/srpmpacker.go
```diff
@@ -125,10 +125,11 @@ var (
...
 
-	workers          = app.Flag("workers", "Number of concurrent goroutines to parse with.").Default(defaultWorkerCount).Uint()
-	concurrentNetOps = app.Flag("concurrent-net-ops", "Number of concurrent network operations to perform.").Default(defaultNetOpsCount).Uint()
-	repackAll        = app.Flag("repack", "Rebuild all SRPMs, even if already built.").Bool()
-	nestedSourcesDir = app.Flag("nested-sources", "Set if for a given SPEC, its sources are contained in a SOURCES directory next to the SPEC file.").Bool()
+	workers           = app.Flag("workers", "Number of concurrent goroutines to parse with.").Default(defaultWorkerCount).Uint()
+	concurrentNetOps  = app.Flag("concurrent-net-ops", "Number of concurrent network operations to perform.").Default(defaultNetOpsCount).Uint()
+	repackAll         = app.Flag("repack", "Rebuild all SRPMs, even if already built.").Bool()
+	nestedSourcesDir  = app.Flag("nested-sources", "Set if for a given SPEC, its sources are contained in a SOURCES directory next to the SPEC file.").Bool()
+	versionsMacroFile = app.Flag("versions-macro-file", "File containing release and version macros for all SPECS to use while packing SRPMs.").ExistingFile()
```
PawelWMS: My comment in `definition.go` made me realize that we might as well use a more generic `rpm-macros-file` or `extra-macros-files` (like you have in the versions processor) to handle multiple macros files instead of forcing a single file with a name, which suggests a very narrow purpose. This would be more flexible if we ever need to pass other macros files.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+	targetArch    = app.Flag("target-arch", "The architecture of the machine the RPM binaries run on").String()
+	buildDir      = app.Flag("build-dir", "Directory to store temporary files while parsing.").String()
+	logFlags      = exe.SetupLogFlags(app)
+	profFlags     = exe.SetupProfileFlags(app)
+	workerTar     = app.Flag("worker-tar", "Full path to worker_chroot.tar.gz.  If this argument is empty, specs will be parsed in the host environment.").ExistingFile()
+	timestampFile = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
+	extraFiles    = app.Flag("extra-macros-files", "Additional files whose contents will be appended to the output; may be specified multiple times.").ExistingFiles()
+)
+
+func main() {
```
PawelWMS, line 47: This is a very long function. Could you break it down into smaller, single-purpose functions?

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+	prof, err := profile.StartProfiling(profFlags)
+	if err != nil {
+		logger.Log.Warnf("Could not start profiling: %s", err)
+	}
+	defer prof.StopProfiler()
+
+	timestamp.BeginTiming("versionsprocessor", *timestampFile)
+	defer timestamp.CompleteTiming()
+
+	logger.PanicOnError(err)
```
PawelWMS: This looks weird. We get `err` from `profile.StartProfiling`, so why panic so late?

Btw., I'd also take the error output from `BeginTiming`, even though it seems we're ignoring it elsewhere. Not sure why that's the case but looks incorrect.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+	var buildArch string = *targetArch
+
+	if *targetArch == "" {
+		buildArch, err = rpm.GetRpmArch(runtime.GOARCH)
+		if err != nil {
+			return
+		}
+	}
+
+	if *workerTar != "" {
```
PawelWMS: Do we even want to continue, if there's no worker tarball?

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+		const leaveFilesOnDisk = false
+		chroot, err = createChroot(*workerTar, *buildDir, *specsDir)
```
PawelWMS: Shouldn't we use `extraFiles` for setting up the chroot? How are the versions from non-core packages extracted without it, if their specs depend on versions of core packages?

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+func main() {
+	const (
+		querySrpm             = `%{NAME}-%{VERSION}-%{RELEASE}.src.rpm`
+		queryProvidedPackages = `rpm %{ARCH}/%{nvra}.rpm\n[provides %{PROVIDENEVRS}\n][requires %{REQUIRENEVRS}\n][arch %{ARCH}\n]`
+		prefix                = "azl"
+	)
+
+	var (
+		chroot        *safechroot.Chroot
+		macros_output []byte
```
PawelWMS: Naming nit-pick: `macrosOutput`.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+	profFlags     = exe.SetupProfileFlags(app)
+	workerTar     = app.Flag("worker-tar", "Full path to worker_chroot.tar.gz.  If this argument is empty, specs will be parsed in the host environment.").ExistingFile()
+	timestampFile = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
+	extraFiles    = app.Flag("extra-macros-files", "Additional files whose contents will be appended to the output; may be specified multiple times.").ExistingFiles()
+)
+
+func main() {
+	const (
+		querySrpm             = `%{NAME}-%{VERSION}-%{RELEASE}.src.rpm`
+		queryProvidedPackages = `rpm %{ARCH}/%{nvra}.rpm\n[provides %{PROVIDENEVRS}\n][requires %{REQUIRENEVRS}\n][arch %{ARCH}\n]`
```
PawelWMS: Look unused.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+			releaseVerSplit := strings.Split(versionRelease[0], "-")
+
+			if len(releaseVerSplit) < 2 {
+				logger.Log.Errorf("Invalid version-release format retrieved from spec file (%s): %s", specFileName, versionRelease[0])
+				continue
+			}
+
+			version := releaseVerSplit[0]
+			release := releaseVerSplit[1]
+			releaseClean := strings.Replace(release, ".azl3", "", 1) // targetting azl3 specifically since this won't go into above 3.0 toolkit
```
PawelWMS: Why a hard-coded distro tag instead of using `distTag`? Even if we're not expecting to use it beyond 3.0, I wouldn't hard-code that on this level.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+			sourceDir := filepath.Dir(specFile)
+			noCheckDefines := rpm.DefaultDistroDefines(false, *distTag)
+
+			versionRelease, err := rpm.QuerySPEC(specFile, sourceDir, `%{VERSION}-%{RELEASE}`, buildArch, noCheckDefines, rpm.QueryHeaderArgument)
+			if err != nil {
+				logger.Log.Errorf("Failed to query spec file (%s). Error: %s", specFileName, err)
+				continue
+			}
+
+			logger.PanicOnError(err)
```
PawelWMS: Since we're inside a function, I'd rely on `return err`. We also try not to use `PanicOnError` anywhere outside `main`, so the caller can decide what to do with the error. Also, I think we never hit this code considering the `continue` above.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+
+			if len(versionRelease) == 0 {
+				logger.Log.Errorf("Invalid version-release retrieved from spec file (%s): %s", specFileName, versionRelease)
```
PawelWMS: Nit-pick: since we're explicitly checking for an empty string, I'd make the error message also more explicit. For instance:

```suggestion
				logger.Log.Errorf("Empty version-release retrieved from spec file (%s)", specFileName)
```

Alternatively, I'd check for `len(versionRelease) != 1`, since we don't accept any other outcome.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+				continue
+			}
+
+			releaseVerSplit := strings.Split(versionRelease[0], "-")
+
+			if len(releaseVerSplit) < 2 {
```
PawelWMS: I think `> 2` would also be an unexpected and invalid result.

```suggestion
			if len(releaseVerSplit) != 2 {
```

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+		for _, specFile := range allSpecFiles {
+
+			// Get spec file version-release
+
+			specFileName := filepath.Base(specFile)
+
+			sourceDir := filepath.Dir(specFile)
+			noCheckDefines := rpm.DefaultDistroDefines(false, *distTag)
+
+			versionRelease, err := rpm.QuerySPEC(specFile, sourceDir, `%{VERSION}-%{RELEASE}`, buildArch, noCheckDefines, rpm.QueryHeaderArgument)
```
PawelWMS: This looks fishy. We need versions and releases of all packages built from a given spec file. There could be multiple and their names will differ from the name of the spec file itself. Check out `perl.spec` for an extreme example with 188 subpackages in total with many having their own version, release, and even epoch different from the default package's one.

I'd suggest looking at `QuerySPECForBuiltRPMs`, which prints all package names along with their versions and releases one per line. You could either use it directly and then parse the results with a regex or keep using `QuerySPEC` with a modified version of the `%{nevra}\n` query format from `QuerySPECForBuiltRPMs` and the `rpm.QueryBuiltRPMHeadersArgument` arguments. For instance, `%{NAME}: %{evr}\n` will give you results along these lines (taken from `ca-certificates.spec`):
```bash
ca-certificates: 1:3.0.0-14.azl3
ca-certificates-shared: 1:3.0.0-14.azl3
ca-certificates-base: 1:3.0.0-14.azl3
ca-certificates-tools: 1:3.0.0-14.azl3
ca-certificates-legacy: 1:3.0.0-14.azl3
```

I'd personally just use `QuerySPECForBuiltRPMs` and a regex to split the whole thing in one go. Manual parsing tends to get lengthy and in case there are issues with it, updates tend to be tricky. As a bonus, you can write the reges in such way that it cuts out the dist tag for you, so you don't need `strings.Replace`.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+
+		// Process all specs files
+		for _, specFile := range allSpecFiles {
+
+			// Get spec file version-release
+
+			specFileName := filepath.Base(specFile)
+
+			sourceDir := filepath.Dir(specFile)
+			noCheckDefines := rpm.DefaultDistroDefines(false, *distTag)
```
PawelWMS: These aren't just disabled checks and in general, what the defaults are may change, so I'd use a more generic name. A simple `defines` seems to be what we've been using so far.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+			}
+
+			version := releaseVerSplit[0]
+			release := releaseVerSplit[1]
+			releaseClean := strings.Replace(release, ".azl3", "", 1) // targetting azl3 specifically since this won't go into above 3.0 toolkit
+
+			// strip out the .spec suffix and replace '-' with '_' as RPM macros cannot have '-'
+			specFileNameMacroFormat := strings.Replace(specFileName, ".spec", "", 1)
+			specFileNameMacroFormat = strings.ReplaceAll(specFileNameMacroFormat, "-", "_")
+			specFileNameMacroFormat = strings.ToLower(specFileNameMacroFormat)
```
PawelWMS: Why `ToLower`? Package names (or Linux in general) are case sensitive.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+		allSpecFiles, err := specreaderutils.FindSpecFiles(*specsDir, nil)
+		if err != nil {
+			logger.Log.Panicf("Error finding spec files: %s", err)
+			return err
+		}
+
+		logger.Log.Infof("Processing version and release for %d spec files into %s", len(allSpecFiles), *output)
+
+		// Process all specs files
+		for _, specFile := range allSpecFiles {
```
PawelWMS, line 104: Can you add a longer comment somewhere in this file explaining what we're exactly doing and what results we want to achieve? It's clear to us now but a month from now, that may no longer be the case.;)

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+			versionMacroString := prefix + "_" + specFileNameMacroFormat + "_version"
+			releaseMacroString := prefix + "_" + specFileNameMacroFormat + "_release"
+
+			// Generate RPM macro definitions instead of modifying spec files directly.
+			macros := fmt.Sprintf("%%%s %s\n%%%s %s\n",
+				versionMacroString, version,
+				releaseMacroString, releaseClean,
+			)
+
+			macros_output = append(macros_output, []byte(macros)...)
```
PawelWMS: I think defining `macros_output` as `byte[]` forces us to do casting back and forth unnecessarily. What do you think about making it a `string[]` with one line per entry? You could then do this:

```suggestion
			macrosOutput = append(macrosOutput, fmt.Sprintf("%%%s %s", versionMacroString, version))
			macrosOutput = append(macrosOutput, fmt.Sprintf("%%%s %s", releaseMacroString, releaseClean))
```

Later in the code you could then use `file.ReadLines()` and `file.WriteLines()` instead of the plain `file.Read()` and `file.Write`. This way you can avoid manually working with `\n` as well.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+
+	err = file.Write(string(macros_output), *output)
+	if err != nil {
+		logger.Log.Errorf("Failed to write file (%s)", *output)
+		return
+	}
+}
+
+// createChroot creates a chroot to parse SPECs inside of.
+func createChroot(workerTar, buildDir, specsDir string) (chroot *safechroot.Chroot, err error) {
```
PawelWMS: The purpose of this function seems the same as `specreaderutil.createChroot` - we're preparing a chroot environment to parse the specs inside it. Considering that, I'd consider making that function public and re-using it here.

toolkit/tools/pkg/specreaderutils/specreaderutil.go
```diff
@@ -126,7 +127,8 @@ func ParseSPECsWrapper(buildDir, specsDir, rpmsDir, srpmsDir, toolchainDir, dist
 }
 
 // createChroot creates a chroot to parse SPECs inside of.
-func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safechroot.Chroot, err error) {
+// If releaseVersionMacrosFile is non-empty, its parent directory is also made available inside the chroot.
```
PawelWMS: This comment seems incorrect. We're not exposing the whole directory of `releaseVersionMacrosFile` inside the spec, right?

toolkit/tools/pkg/specreaderutils/specreaderutil.go
```diff
@@ -167,6 +169,35 @@ func createChroot(workerTar, buildDir, specsDir, srpmsDir string) (chroot *safec
...
+			return
+		}
+
+		copyErr := file.Copy(releaseVersionMacrosFile, macrosDestFile)
+		if copyErr != nil {
+			logger.Log.Errorf("Failed to copy release version macros file into chroot (%s -> %s): %s", releaseVersionMacrosFile, macrosDestFile, copyErr)
+			return
+		}
+
+		logger.Log.Infof("Copied release version macros file into chroot (%s -> %s)", releaseVersionMacrosFile, macrosDestFile)
```
PawelWMS: All of the `*Err` errors are never returned from this function in case of an error, because `err` is never updated. Also, I'd just keep re-using `err` instead of adding a new error variable for each function call.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
...
+			// Generate RPM macro definitions instead of modifying spec files directly.
+			macros := fmt.Sprintf("%%%s %s\n%%%s %s\n",
+				versionMacroString, version,
+				releaseMacroString, releaseClean,
+			)
+
+			macros_output = append(macros_output, []byte(macros)...)
+		}
+
+		return parseError
```
PawelWMS: We never assign any errors to `parseError` and the `continue` statements lose any values assigned to `err`. To solve that, you could `errors.Join(parseError, err)` every time `err != nil`. This way, we won't lose any of the errors and can still try to parse all specs before we exit.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,225 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// specreader is a tool to parse spec files into a JSON structure
```
PawelWMS: Invalid comment.

toolkit/scripts/srpm_pack.mk
```diff
@@ -13,8 +13,10 @@
 # update  - Check signatures and updating any mismatches in the signatures file
 SRPM_FILE_SIGNATURE_HANDLING ?= enforce
 
-SRPM_BUILD_CHROOT_DIR = $(BUILD_DIR)/SRPM_packaging
-SRPM_BUILD_LOGS_DIR = $(LOGS_DIR)/pkggen/srpms
+SRPM_BUILD_CHROOT_DIR   = $(BUILD_DIR)/SRPM_packaging
+SRPM_BUILD_LOGS_DIR     = $(LOGS_DIR)/pkggen/srpms
+rel_versions_macro_file = $(PKGBUILD_DIR)/macros.releaseversions
+
```
PawelWMS:
```suggestion
```

SPECS/iser-hwe/iser-hwe.spec
```diff
@@ -251,6 +252,9 @@ fi # 1 : closed
 %endif
 
 %changelog
+* Mon Feb 10 2026 Mykhailo Bykhovtsev <mbykhovtsev@microsoft.com> - 25.07-5_6.12.57.1.1
```
PawelWMS: Can you check, if we can use macros for the changelog as well? Otherwise, this version won't update automatically. Alternatively, we can simplify it and drop the `_*` suffix (I'd check with Jon for his opinion about this solution, though).

toolkit/scripts/tools.mk
mbykhovtsev-ms: I have talked with Jon about this. We agree to add this, but as separate PR

toolkit/scripts/pkggen.mk
mbykhovtsev-ms: I was able to provide it a string with files such as "macros.foo, macros.bar" which worked

toolkit/scripts/srpm_pack.mk
mbykhovtsev-ms: updated

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Updated

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Updated

toolkit/tools/pkg/specreaderutils/specreaderutil.go
mbykhovtsev-ms: Updated

toolkit/tools/pkg/specreaderutils/specreaderutil.go
mbykhovtsev-ms: Removed the comment

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Made it public

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Changed it to string array

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Added comment to the top of the file

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Removed it to now be case sensitive

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Updated

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Changed to use the `%{NAME}: %{evr}\n` with a regex

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Changed it

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Updated

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: I am not sure. I copied this boilerplate from the specreader

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Changed to use the distTag

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Removed

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: renamed

toolkit/Makefile
mbykhovtsev-ms: Added

.github/workflows/check-source-signatures.yml
mbykhovtsev-ms: Added comment

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Split it up a bit more

toolkit/docs/building/building.md
```diff
@@ -814,6 +814,7 @@ To reproduce an ISO build, run the same make invocation as before, but set:
...
 | TEST_IGNORE_LIST              |                                                                                                        | Ignore testing these packages. Ignoring and forcing the same test re-run is invalid and will fail the build. The argument accepts both spec and package names. Example: for `python-werkzeug.spec`, which builds the `python3-werkzeug` package both `python-werkzeug` and `python3-werkzeug` are correct.
+| EXTRA_MACROS_FILES              |                                                                                                        | Space separated list of additional `<macros.name>` files whose content will be appended to the main `macros.releaseversions` file used during package build. Used to resolve versions of kernel Out of Tree Modules.
```
PawelWMS: The doc is user-facing and `macros.releaseversions` is an internal implementation detail. Considering that, I wouldn't mention it here and just limit the explanation to say, that the `EXTRA_MACROS_FILES` are files containing additional RPM macros, which will be available to the build and as an example say, that these can be used to resolve kernel versions for specs with OOT modules.

toolkit/tools/internal/safechroot/dummychroot.go
```diff
@@ -23,3 +23,7 @@ func (d *DummyChroot) UnsafeRun(toRun func() error) (err error) {
 func (d *DummyChroot) AddFiles(filesToCopy ...FileToCopy) (err error) {
 	return AddFilesToDestination(d.RootDir(), filesToCopy...)
 }
+
+func (d *DummyChroot) AddRPMMacrosFile(macrosFilePath string) error {
+	return d.AddRPMMacrosFile(macrosFilePath)
+}
```
PawelWMS, line 29: Isn't this creating an infinite loop?

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -160,6 +162,48 @@ func NewOverlayMountPoint(chrootDir, source, target, lowerDir, upperDir, workDir
 	return
 }
 
+func (c *Chroot) AddRPMMacrosFile(macrosFilePath string) (err error) {
+
+	var (
+		macroDir string
+	)
```
PawelWMS:
```suggestion
	macroDir string
```

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -160,6 +162,48 @@ func NewOverlayMountPoint(chrootDir, source, target, lowerDir, upperDir, workDir
...
+
+	doParse := func() error {
```
PawelWMS: Nit-pick about the `doParse` variable name: we're not really parsing anything in that function, I think.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -160,6 +162,48 @@ func NewOverlayMountPoint(chrootDir, source, target, lowerDir, upperDir, workDir
...
+	macrosDestFile := filepath.Join(macrosDestDir, filepath.Base(macrosFilePath))
+
+	// Ensure destination directory exists and copy the file.
+	mkdirErr := directory.EnsureDirExists(macrosDestDir)
+	if mkdirErr != nil {
+		logger.Log.Errorf("Failed to create macros directory inside chroot (%s): %s", macrosDestDir, mkdirErr)
+		return mkdirErr
+	}
+
+	copyErr := file.Copy(macrosFilePath, macrosDestFile)
```
PawelWMS, line 208: I'd also add a check, that the file doesn't already exist there. If it does, we may be overriding some other macros and cause issues, which may be annoying to debug. I'd either error out if a file exists or print a warning.

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -74,7 +74,7 @@ func New(buildDirPath, workerTarPath, rpmDirPath, nameFilePath, exceptionFilePat
 		jobSemaphore:     make(chan struct{}, runtime.NumCPU()*2),
 	}
 
-	err = newLicenseChecker.simpleToolChroot.InitializeChroot(buildDirPath, chrootName, workerTarPath, rpmDirPath)
+	err = newLicenseChecker.simpleToolChroot.InitializeChroot(buildDirPath, chrootName, workerTarPath, rpmDirPath, "")
```
PawelWMS: Suggestion: instead of an "anonymous" empty string, I'd assign it to a constant with a name explaining its purpose here. For instance `noExtraMacrosFile = ""`.

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -160,6 +162,48 @@ func NewOverlayMountPoint(chrootDir, source, target, lowerDir, upperDir, workDir
...
+	if copyErr != nil {
+		logger.Log.Errorf("Failed to copy release version macros file into chroot (%s -> %s): %s", macrosFilePath, macrosDestFile, copyErr)
+		return mkdirErr
+	}
```
PawelWMS: I think `c.AddFiles()` can take care of creating the directory for you, copying the file, and printing relevant error logs. I think it only needs the source path and the destination path (destination relative to the chroot's root, `macrosDestFile  := filepath.Join(macroDir, filepath.Base(macrosFilePath))`.

toolkit/tools/scheduler/buildagents/chrootagent.go
```diff
@@ -97,8 +97,12 @@ func serializeChrootBuildAgentConfig(config *BuildAgentConfig, basePackageName,
...
 	}
 
-	if config.RpmmacrosFile != "" {
-		serializedArgs = append(serializedArgs, fmt.Sprintf("--rpmmacros-file=%s", config.RpmmacrosFile))
+	if config.RPMMacrosFiles != "" {
+		serializedArgs = append(serializedArgs, fmt.Sprintf("--rpmmacros-file=%s", config.RPMMacrosFiles))
+	}
+
+	if config.VersionsMacroFile != "" {
+		serializedArgs = append(serializedArgs, fmt.Sprintf("--versions-macro-file=%s", config.VersionsMacroFile))
```
PawelWMS, line 105: Why not re-use `--rpmmacros-file`? This is in relation to my comment in `definition.go`.


```suggestion
	if config.RPMMacrosFiles != "" {
		serializedArgs = append(serializedArgs, fmt.Sprintf("--rpmmacros-file=%s", config.RPMMacrosFiles))
	}

	if config.VersionsMacroFile != "" {
		serializedArgs = append(serializedArgs, fmt.Sprintf("--rpmmacros-file=%s", config.VersionsMacroFile))
```

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+	logFlags      = exe.SetupLogFlags(app)
+	profFlags     = exe.SetupProfileFlags(app)
+	workerTar     = app.Flag("worker-tar", "Full path to worker_chroot.tar.gz.  If this argument is empty, specs will be parsed in the host environment.").ExistingFile()
+	timestampFile = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
+	extraFiles    = app.Flag("extra-macros-file", "Additional files whose contents will be appended to the output; may be specified multiple times.").ExistingFiles()
+)
+
+func main() {
+	const (
+		prefix = "azl"
```
PawelWMS: Seem to be used only in one internal function, if I'm seeing it right. If that's the case, I'd just move the declaration there, so there's no need to pass it through all the function arguments.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+	var buildArch string = *targetArch
+
+	if *targetArch == "" {
+		buildArch, err = rpm.GetRpmArch(runtime.GOARCH)
+		if err != nil {
+			logger.PanicOnError(err)
+		}
+		if err != nil {
+			logger.Log.Errorf("Failed to determine RPM architecture for runtime architecture %q: %v", runtime.GOARCH, err)
+		}
```
PawelWMS, line 73: `PanicOnError` will return from the process, the second `err != nil` is unnecessary and unreachable.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	if *workerTar == "" {
+		logger.Log.Error("No worker tar provided, parsing specs in host environment. This may cause issues if the host environment is different from the target build environment.")
+		return
```
PawelWMS: Inconsistent with the `PanicOnError` earlier. Please do only one or the other in the `main()` function. Also, since it looks like we're quitting when there's no worker tar, I'd suggest to make the error simply state that it's required. No need for extra comments about what might have happened if we chosen to do something else. We didn't.:)

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	const leaveFilesOnDisk = false
+	chroot, err = specreaderutils.CreateChroot("versionprocessor_chroot", *workerTar, *buildDir, *specsDir, "", "")
```
PawelWMS: The multiple empty strings passed to `CreateChroot` (there are other places where there's only one but still) look like a code smell. If we want to make this function flexible to build chroot with various configurations, please try applying the [functional options pattern](https://golang.cafe/blog/golang-functional-options-pattern). The end result should be a set of mandatory arguments plus a 0+ number of optional ones set through public functions like `CreateChroot("chroot_name", "tarball/path", "build/dir/path", WithMountPoint("specs/path"), WithExtraMacrosFile("macros/file/path")` - the `WithX` functions are defined next to `CreateChroot` and they are varargs.

I'd ask Copilot in plan mode to refactor that to this pattern - I think it should know how to do it right and adjust all places where `CreateChroot` is used..

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+			if err != nil {
+				logger.Log.Errorf("Error processing spec file (%s): %s", specFile, err)
+				continue
+			}
+		}
+
+		return err
+	}
+
+	if chroot != nil {
```
PawelWMS: Is this check necessary? Is it possible to not have a chroot?

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+		macrosOutput = append(macrosOutput, macros)
+	}
+
+	return macrosOutput, nil
+}
+
+func processPackageVersionString(packageVersionString string, specFileName string, prefix string) (macros string, err error) {
+	// the output of the above query is in the format of "packagename: version-release",
+	// so split by ": " to get the version-release portion we want the second part
+	releaseVerSplit := regexp.MustCompile(`^[^:]+: (.+)-(.+)$`).FindStringSubmatch(packageVersionString)[1:]
```
PawelWMS: I'd make `regexp.MustCompile(`^[^:]+: (.+)-(.+)$`)` a global to avoid re-compiling the regex on every entry to the function.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+
+func processSpecFile(specFile string, buildArch string, prefix string, macrosOutput []string) (newMacrosOutput []string, err error) {
+	// Get spec file version-release
+
+	specFileName := filepath.Base(specFile)
+
+	sourceDir := filepath.Dir(specFile)
+	defines := rpm.DefaultDistroDefines(false, *distTag)
+
+	packages, err := rpm.QuerySPEC(specFile, sourceDir, `%{NAME}: %{evr}\n`, buildArch, defines, rpm.QueryHeaderArgument)
```
PawelWMS, line 150: `%{evr}` will return epoch numbers as well **when epoch > 0**, so things like `5:1.0.0-1.azl3`. With the current parsing logic, the version macro will end up with `5:1.0.0`, which is not really the version but a combination of the epoch and the version, so that's not what we promise with the macros. I'd suggest to always print the epoch for consistency and adding a separate macro for it.

I though that `%{evr}` will always give us the epoch but it looks like we may need an explicit `%{EPOCH}`. So `%{evr}` -> `%{EPOCH}:%{VERSION}-%{RELEASE}`. Or anything else, if it makes parsing easier.:)

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+
+		logger.Log.Infof("Processing version and release for %d spec files into %s", len(allSpecFiles), *output)
+
+		// Process all specs files
+		for _, specFile := range allSpecFiles {
+
+			// Get spec file version-release
+			macrosOutput, err = processSpecFile(specFile, buildArch, prefix, macrosOutput)
+
```
PawelWMS, line 116: Consistency nit-pick:

```suggestion
			// Get spec file version-release
			macrosOutput, err = processSpecFile(specFile, buildArch, prefix, macrosOutput)
```

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+			if err != nil {
+				logger.Log.Errorf("Error processing spec file (%s): %s", specFile, err)
+				continue
```
PawelWMS: Why do we want to continue?

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		logger.PanicOnError("Failed to create chroot")
+	}
+	defer chroot.Close(leaveFilesOnDisk)
+
+	doParse := func() error {
+		// Find all spec files
+		allSpecFiles, err := specreaderutils.FindSpecFiles(*specsDir, nil)
+		if err != nil {
+			logger.Log.Panicf("Error finding spec files: %s", err)
```
PawelWMS: `Panicf` will exit the process. Since this is an internal function, I'd leave that decision to the caller, so log an error and return it.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+
+		return err
+	}
+
+	if chroot != nil {
+		err = chroot.Run(doParse)
+	} else {
+		err = doParse()
+	}
```
PawelWMS: Missing error check.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+	if len(releaseVerSplit) != 2 {
+		err = fmt.Errorf("Empty version-release format retrieved from spec file (%s)", specFileName)
+		logger.Log.Errorf("Empty version-release format retrieved from spec file (%s)", specFileName)
+
+		return "", err
+	}
+
+	version := releaseVerSplit[0]
+	release := releaseVerSplit[1]
+	releaseClean := strings.Replace(release, *distTag, "", 1) // targetting azl3 specifically since this won't go into above 3.0 toolkit
```
PawelWMS: `azl3` comment no longer correct. I'd explain only why we're removing the dist tag.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// strip out the .spec suffix and replace '-' with '_' as RPM macros cannot have '-'
+	specFileNameMacroFormat := strings.Replace(specFileName, ".spec", "", 1)
+	specFileNameMacroFormat = strings.ReplaceAll(specFileNameMacroFormat, "-", "_")
```
PawelWMS: This is incorrect - we don't want to print the name of the spec file but the name of the packages built from the spec file. My previous comment around this area had more details but I believe the `%{NAME}` tag from the RPM query may have that package name we're after.

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,215 @@
...
+
+	versionMacroString := prefix + "_" + specFileNameMacroFormat + "_version"
+	releaseMacroString := prefix + "_" + specFileNameMacroFormat + "_release"
+
+	// Generate RPM macro definitions instead of modifying spec files directly.
+	macros = fmt.Sprintf("%%%s %s\n%%%s %s",
+		versionMacroString, version,
+		releaseMacroString, releaseClean,
+	)
```
PawelWMS: Even if this works, I'd strongly suggest just adding one line per array entry instead of pushing two lines concatenated with `\n`. During debugging this may be a red herring where one will need to check how `file.WriteLines` handle a string with `\n` inside it (does it split it into 2 lines vs prints the literal `\n` in one line).

toolkit/tools/internal/safechroot/dummychroot.go
mbykhovtsev-ms: You are right, fixed it

toolkit/tools/internal/safechroot/safechroot.go
mbykhovtsev-ms: I have attempted to use `AddFiles` helper function, but it regressed and was not copying them in the destination I want. I will keep it like this

toolkit/tools/pkg/licensecheck/licensecheck.go
mbykhovtsev-ms: Updated

toolkit/tools/internal/safechroot/safechroot.go
mbykhovtsev-ms: Added

toolkit/tools/internal/safechroot/safechroot.go
mbykhovtsev-ms: Renamed

toolkit/tools/internal/safechroot/safechroot.go
mbykhovtsev-ms: Changed

toolkit/docs/building/building.md
mbykhovtsev-ms: Changed it

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Updated

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Updated

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Updated

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Removed it

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: `%{evr}` still works as it returns an empty string for epoch if it does not exists. Updated code to support capturing epochs and storing them in macros

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Fixed

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Updated

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Added

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: removed

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: updated

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Updated to return array of strings instead

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Changed to return

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Moved it

toolkit/tools/versionsprocessor/versionsprocessor.go
mbykhovtsev-ms: Changed it to be global variable, also modified to filter out epochs

toolkit/docs/how_it_works/3_package_building.md
```diff
@@ -117,6 +117,10 @@ The files can be ingested into the `graphviz` tools to visualize them, although
 dot -Tpng -o visualized.png < graph.dot
 ```
 
+### Dynamic versioning
+We have a versionsprocessor tool that iterates over all Specs and writes their release and versions into a macro file in a format of
+`azl_<package_name>_release`, `azl_<package_name>_version`, note that the `<package_name>` needs any `-` are replaced with `_` due to macros not allowing `-`.
```
dmcilvaney:
```suggestion
`azl_<package_name>_release`, `azl_<package_name>_version`, note that the `<package_name>` needs any `-` replaced with `_` due to macros not allowing `-`.
```

toolkit/tools/pkg/licensecheck/licensecheck.go
```diff
@@ -74,7 +74,9 @@ func New(buildDirPath, workerTarPath, rpmDirPath, nameFilePath, exceptionFilePat
 		jobSemaphore:     make(chan struct{}, runtime.NumCPU()*2),
 	}
 
-	err = newLicenseChecker.simpleToolChroot.InitializeChroot(buildDirPath, chrootName, workerTarPath, rpmDirPath)
+	noExtraMacrosFile := ""
+
+	err = newLicenseChecker.simpleToolChroot.InitializeChroot(buildDirPath, chrootName, workerTarPath, rpmDirPath, noExtraMacrosFile)
```
dmcilvaney, line 79: will there be any cases where specs can't parse at all without the macros? ie do they all have fallback default values in the spec files?

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -160,6 +162,60 @@ func NewOverlayMountPoint(chrootDir, source, target, lowerDir, upperDir, workDir
 	return
 }
 
+func (c *Chroot) AddRPMMacrosFile(macrosFilePath string) error {
+	return AddRPMMacrosFile(c, macrosFilePath)
+}
+
+func AddRPMMacrosFile(c ChrootInterface, macrosFilePath string) (err error) {
```
dmcilvaney, line 169: have some overlap here with the `customizationmacros` package, maybe check if there is enough overlap to be worth using it, but if not thats fine (I'm leaning towards not).

toolkit/tools/pkgworker/pkgworker.go
```diff
@@ -218,6 +219,15 @@ func buildSRPMInChroot(chrootDir, rpmDirPath, toolchainDirPath, workerTar, srpmF
 		err = fmt.Errorf("failed to initialize chroot:\n%w", err)
 		return
 	}
+
+	// If a release version macros file is provided, copy it into the default RPM macros directory
+	// inside the chroot so rpmspec/rpmbuild pick it up automatically.
+	if releaseVersionMacrosFile != "" {
```
dmcilvaney: This pattern seems to repeat everywhere, would it be better to add a param to either NewChroot() or Initailize()?

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,217 @@
+// Copyright (c) Microsoft Corporation.
```
dmcilvaney, line 1: This shares a lot of overlap with rpmsnappshot, like above, probably not enough common code to bother de-duplicating, but the snapshot tool has a bunch of good features (like parallel processing) that is probably important to grab. If I understand right, this tool is now in the critical path for a lot of build steps.

toolkit/scripts/toolkit.mk
```diff
@@ -25,6 +25,7 @@ rpms_snapshot_dir_name = rpms_snapshots
 rpms_snapshot_build_dir = $(BUILD_DIR)/$(rpms_snapshot_dir_name)
 rpms_snapshot_logs_path = $(LOGS_DIR)/$(rpms_snapshot_dir_name)/rpms_snapshot.log
 rpms_snapshot_per_specs = $(rpms_snapshot_build_dir)/$(specs_dir_name)_$(rpms_snapshot_name)
+rel_versions_macro_file = $(PKGBUILD_DIR)/macros.releaseversions
```
dmcilvaney, line 28: Won't we want this file included in the packaged toolkit so we can use it for other builds that aren't in the SPECS folder? Or is that out of scope?

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,217 @@
...
+		return err
+	}
+
+	err = chroot.Run(doParse)
+
+	if err != nil {
+		logger.Log.Errorf("Error processing spec files: %s", err)
+	}
+
+	err = writeExtraFilesToOutput(*extraFiles, macrosOutput, *output)
```
dmcilvaney, line 135: Don't write the file unless we have valid data. This will break the makefile's dependency tracking. If the tool fails halfway through, it still creates a file and will not update it again since nothing else changed.

Maybe make it logger.log.Fatalf?

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -0,0 +1,217 @@
...
+	packageFileNameMacroFormat := strings.Replace(specFileName, ".spec", "", 1)
+	packageFileNameMacroFormat = strings.ReplaceAll(packageFileNameMacroFormat, "-", "_")
+
+	epochReleaseString := prefix + "_" + packageFileNameMacroFormat + "_epoch"
+	versionMacroString := prefix + "_" + packageFileNameMacroFormat + "_version"
+	releaseMacroString := prefix + "_" + packageFileNameMacroFormat + "_release"
+
+	// Generate RPM macro definitions instead of modifying spec files directly.
+	macros = []string{
+		fmt.Sprintf("%%%s %s", epochReleaseString, epoch),
```
dmcilvaney: Only add the epoch if its present, otherwise we get >14mil lines of log output like:
`warning: /usr/lib/rpm/macros.d/macros.releaseversions: line 841: Macro %azl_golang_1.24_epoch has empty body`

toolkit/tools/versionsprocessor/versionsprocessor.go
dmcilvaney: From a quick test, snapshot is about 3-4x faster.

toolkit/Makefile
dmcilvaney, line 37: Unrelated to your direct changes, but during testing I noticed a makefile bug that makes these new steps slow. Can you fix while you are here?

```diff
diff --git a/toolkit/Makefile b/toolkit/Makefile
index 6690db8bc..cf129ae82 100644
--- a/toolkit/Makefile
+++ b/toolkit/Makefile
@@ -170,7 +170,7 @@ endif
 VALIDATE_IMAGE_GPG ?= n
 
 # Default GPG keys for package GPG validation, used with VALIDATE_TOOLCHAIN_GPG and VALIDATE_IMAGE_GPG
-default_gpg_keys := $(wildcard $(PROJECT_ROOT)/SPECS/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)
+default_gpg_keys := $(strip $(wildcard $(PROJECT_ROOT)/SPECS/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY))
 TOOLCHAIN_GPG_VALIDATION_KEYS ?= $(default_gpg_keys)
 IMAGE_GPG_VALIDATION_KEYS ?= $(default_gpg_keys)
 ```

File: toolkit/tools/pkg/licensecheck/licensecheck.go (hunk 57 above)
Comment (mbykhovtsev-ms) [reply toolkit/tools/pkg/licensecheck/licensecheck.go:79]: No, there is no fallback mechanism for this. There was an idea of adding potentially a fallback, however I believe that if we do that we still need to maintain some sort of hard-coded values even with the fallback mechanism?

File: toolkit/docs/how_it_works/3_package_building.md (hunk 56 above)
Comment (mbykhovtsev-ms) [reply toolkit/docs/how_it_works/3_package_building.md]: Updated

File: toolkit/tools/pkgworker/pkgworker.go (hunk 59 above)
Comment (mbykhovtsev-ms) [reply toolkit/tools/pkgworker/pkgworker.go]: Updated by adding new parameter to Initialize

File: toolkit/tools/versionsprocessor/versionsprocessor.go (hunk 60 above)
Comment (mbykhovtsev-ms) [reply toolkit/tools/versionsprocessor/versionsprocessor.go:1]: I see. I was just thinking the other day if I could parallelize the processing. Applied the pattern from the rpmsnappshot.

File: toolkit/scripts/toolkit.mk (hunk 61 above)
Comment (mbykhovtsev-ms) [reply toolkit/scripts/toolkit.mk:28]: Added it

File: toolkit/tools/versionsprocessor/versionsprocessor.go (hunk 62 above)
Comment (mbykhovtsev-ms) [reply toolkit/tools/versionsprocessor/versionsprocessor.go:135]: Added the fatalf suggestion

File: toolkit/tools/versionsprocessor/versionsprocessor.go (hunk 63 above)
Comment (mbykhovtsev-ms) [reply toolkit/tools/versionsprocessor/versionsprocessor.go]: Adjusted to only print if we have epoch

File: toolkit/Makefile (hunk 4 above)
Comment (mbykhovtsev-ms) [reply toolkit/Makefile:37]: Sure, applied it

File: toolkit/tools/pkg/licensecheck/licensecheck.go (hunk 57 above)
Comment (dmcilvaney) [reply toolkit/tools/pkg/licensecheck/licensecheck.go:79]: ok, make sure the license checker still works then, it also needs to parse the files I believe.

File: toolkit/scripts/toolkit.mk [hunk 64]
```diff
@@ -83,6 +84,7 @@ $(toolkit_archive): $(go_tool_targets) $(mariner_repos_files) $(toolkit_componen
 	cp -r $(toolkit_root_files) $(toolkit_prep_dir) && \
 	cp $(mariner_repos_files) $(toolkit_repos_dir) && \
 	cp $(toolkit_component_extra_files) $(toolkit_prep_dir) && \
+	cp $(rel_versions_macro_file) $(toolkit_prep_dir) && \
```
dmcilvaney, line 87: Need to add it as a build dependency I think, unless its a transitive one I don't see.

toolkit/Makefile
```diff
@@ -170,7 +170,7 @@ endif
 VALIDATE_IMAGE_GPG ?= n
 
 # Default GPG keys for package GPG validation, used with VALIDATE_TOOLCHAIN_GPG and VALIDATE_IMAGE_GPG
-default_gpg_keys := $(wildcard $(PROJECT_ROOT)/SPECS/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY)
+default_gpg_keys := $(strip $(wildcard $(PROJECT_ROOT)/SPECS/azurelinux-repos/MICROSOFT-*-GPG-KEY) $(wildcard $(toolkit_root)/repos/MICROSOFT-*-GPG-KEY))
```
dmcilvaney, line 173: thanks, this should make stuff run much faster for local builds

toolkit/tools/pkg/licensecheck/licensecheck.go
mbykhovtsev-ms: It does work. When I had spec changes as part of the PR the license check PR check was passing

toolkit/scripts/toolkit.mk
mbykhovtsev-ms: Updated

PR #15599, opened 2026-01-28
Add network overrides and build a qemu image

binujp:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

We have to make sure systemd-networkd is managing all interfaces. Add a package to deliver the override.

Brought in logic to build a vm-image with the current package selection. This is based on Nan Liu's changes which accomplished the same.

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


base/comps/azurelinux-overrides/azurelinux-overrides.spec
```diff
@@ -0,0 +1,28 @@
+Summary:        Azure Linux systemd network overrides
```
reubeno: If this is really specific to network config, can we please change the name to match? Something like `azurelinux-config-network` or some variant thereof?

You could do something like:

```
/base/comps/azurelinux-config/
    * azurelinux-config-network.spec
    * azurelinux-config-somethingelse.spec
```

Alternatively, if you want to have them in one spec, we could at least start with multiple sub-packages.

The main thinking here is that general overrides packages can quickly become dumping grounds.

base/comps/azurelinux-overrides/azurelinux-overrides.spec
```diff
@@ -0,0 +1,28 @@
...
+Name:           azurelinux-overrides
+Version:        4.0
+Release:        0.1
```
reubeno: Should include the dist tag; and might as well start at `1` like most specs.

base/comps/azurelinux-overrides/azurelinux-overrides.spec
```diff
@@ -0,0 +1,28 @@
...
+
+%install
+install -d %{buildroot}%{_sysconfdir}/systemd/network
+install -m 644 %{_sourcedir}/50-default.network %{buildroot}%{_sysconfdir}/systemd/network/50-default.network
+
+%files
+%config(noreplace) /etc/systemd/network/50-default.network
+
+%changelog
+* Wed Jan 21 2026 Binu Philip <bphilip@microsoft.com> - 4.0-0.1
```
reubeno: Reminder for us to talk about auto-changelog and auto-release later.

base/images/vm-base/vm-base.kiwi
```diff
@@ -114,9 +115,8 @@
         <package name="system-release" />
     </packages>
 
-    <!--
     <users>
-        <user name="root" password="INSERT-PASSWORD-HERE" groups="root" />
+          <user name="root" password="INSERT-PASSWORD-HERE" groups="root" />
```
reubeno: Absolutely no checked in passwords.

scripts/demo-build.sh
```diff
@@ -13,29 +13,78 @@ if [ ! -f azldev.toml ]; then
...
+if [[ "$#" -eq 1 && "$1" == "vm" ]] ; then
+    image_type="vm"
+    sudo rm -rf "$TARGET_DIR"
+fi
+
+echo "Building a '$image_type' image"
+
 # Build azurelinux-rpm-config to generate system macros, etc.
-azldev comp build azurelinux-rpm-config && createrepo_c ./base/out
+azldev comp build azurelinux-rpm-config; createrepo_c ./base/out
```
reubeno: The `&&` was there intentionally; why remove it? Ditto for below.

scripts/demo-build.sh
```diff
@@ -13,29 +13,78 @@ if [ ! -f azldev.toml ]; then
...
+	 --kiwi-file vm-base.kiwi \
+	 system build  \
+	 --description ./base/images/vm-base \
+	 --target-dir "$TARGET_DIR" \
+	 --add-repo="file:///$PWD/base/out,rpm-md,azl,1"
+
+    # boot QEMU VM, C-a x to quit. VM port 22 is forwarded to host:2222
+    echo "Launching azl4 VM, type 'C-a x' to quit."
+    echo "To login as azureuser: ssh -p 2222 azureuser@localhost"
+    sudo qemu-system-x86_64 -enable-kvm -m 2048 -cpu host \
```
reubeno: Booting sounds like it should be a separate script or at least an option. Ultimately a script's not awesome, but you're right -- it gets things to move forward.

reubeno (changes requested):

base/comps/azurelinux-config/azurelinux-config.spec
```diff
@@ -0,0 +1,37 @@
...
+# This is a reverse dependency to make systemd-networkd pull-in this package.
+# It is still experimental. Consider this as a placeholder.
+Supplements:    systemd-networkd
+
+%description systemd-networkd
+Provides systemd-networkd configuration installed at /etc/systemd/network which asks systemd to manage all interfaces on the system.
+
+%install
+install -d %{buildroot}%{_sysconfdir}/systemd/network
+install -m 644 %{_sourcedir}/50-default.network %{buildroot}%{_sysconfdir}/systemd/network/50-default.network
```
reubeno: The Fedora Packaging Guidelines say not to use %{_sourcedir} based paths; see https://docs.fedoraproject.org/en-US/packaging-guidelines/RPM_Source_Dir/

base/comps/azurelinux-config/50-default.network
```diff
@@ -0,0 +1,6 @@
+# We want systemd to manage all interfaces
+[Match]
+Name=*
+
+[Network]
+DHCP=yes
```
reubeno: Saw your comment on the other PR. Totally understand the need for systemd to manage all interfaces. I think my question was more around appropriateness of hard-coding DHCP enablement on all network interfaces by default--or whether there's another way that should be configured by the VM provisioning process / cloud-init.

These also seem like unrelated settings.

base/images/vm-base/vm-base.kiwi
```diff
@@ -86,7 +86,7 @@
         <package name="sudo" />
         <package name="systemd-boot" />
         <package name="systemd-networkd" />
-        <package name="systemd-resolved" />
+	<package name="azurelinux-config-systemd-networkd" />
```
reubeno: Nit: spacing looks odd? Not worried about that for now, but mental note we should probably send these through an auto-formatter for consistency at some pt.

base/images/vm-base/vm-base.kiwi
```diff
@@ -114,9 +114,8 @@
         <package name="system-release" />
     </packages>
 
-    <!--
     <users>
-        <user name="root" password="INSERT-PASSWORD-HERE" groups="root" />
+          <user name="root" password="INSERT-PASSWORD-HERE" groups="root" />
```
reubeno: I still see hard-coded passwords and user accounts here. Shouldn't we be using cloud-init instead on both accounts? If `azldev image boot` doesn't get you what you need, then let's improve it.

base/images/images.toml
```diff
@@ -1,6 +1,6 @@
-# [images.vm-base]
```
reubeno, line 1: Did you rebase against upstream? The comments should all be gone.

scripts/demo-build.sh
```diff
@@ -1,9 +1,14 @@
 #!/bin/bash
-set -euxo pipefail
```
reubeno, line 2: FYI -- Tobias has some conflicting changes to this script; yours look more extensive.

scripts/demo-build.sh
```diff
@@ -13,29 +18,75 @@ if [ ! -f azldev.toml ]; then
...
+    # boot QEMU VM, C-a x to quit. VM port 22 is forwarded to host:2222
+    echo "Launching VM from image 'vm-base', type 'C-a x' to quit."
+    echo "Login as user/pass: 'test/${1}': ssh -p 8888 test@localhost"
+    azldev image boot vm-base --test-password "${1}"
+}
+
+if [[ "$image_type" == "container" ]] ; then
+    build_container_image
+elif [[ "$image_type" == "vm" || "$image_type" == "vm-boot" ]] ; then
+    # You may need this cleanup until azldev is fixed:
```
reubeno: If there's a bug, can you make sure it's filed? What "fixing" is needed?

base/comps/azurelinux-config/50-default.network
binujp: I went with the thought that if we do not enable DHCP on all interfaces there is no way to know which one is available and hence where to pull in an IP to. But doing this via cloud-init makes the whole dhcp discussion a no-op. Let me see if there are recommendations on this before I make that change.

base/images/vm-base/vm-base.kiwi
binujp: I guess my space-tab setting is screwed again.

base/images/vm-base/vm-base.kiwi
binujp: I assumed since this is a hash which will never match it is fine. But I understand your comment about the optics of having any password at all.

base/images/images.toml
binujp: This got lost in the stacked diff I had. I did rebase but the change still got pushed as comments I added!Fixed.

scripts/demo-build.sh
binujp: Ok, I will be on the lookout for that.

scripts/demo-build.sh
binujp: I need to remove that comment. This was from when selinux was enforced and creating a rpm lock file was getting denied. I was trying to defend against it. Not worth it.

base/comps/azurelinux-config/50-default.network
binujp: Cloud init does take care of enabling dhcp on the most probable interface. https://cloudinit.readthedocs.io/en/latest/reference/network-config.html#fallback-network-configuration. I will comment it out and add the reasoning.

base/comps/azurelinux-config/50-default.network
```diff
@@ -0,0 +1,7 @@
+# We want systemd to manage all interfaces
+[Match]
+Name=*
```
ddstreetmicrosoft, line 3: This is definitely wrong, for 2 reasons: 1) we almost certainly don't want to force systemd-networkd configuaration onto all interfaces like this, and 2) this instructs systemd-networkd to take control of all interfaces but not actually start dhcp on them (as it's commented below)

base/comps/azurelinux-config/azurelinux-config.spec
```diff
@@ -0,0 +1,37 @@
...
+Provides AZL specifc configuration overrides via sub-packages. These sub-packages will be pulled in by the relevant owning packages via reverse dependencies.
+
+
+%package systemd-networkd
+Summary:        Azure Linux specific configuration for systemd-networkd.
+Requires:       systemd-networkd
+
+# This is a reverse dependency to make systemd-networkd pull-in this package.
+# It is still experimental. Consider this as a placeholder.
+Supplements:    systemd-networkd
```
ddstreetmicrosoft, line 24: The `systemd-networkd` package shouldn't require or supplement this package

base/comps/azurelinux-config/azurelinux-config.spec
```diff
@@ -0,0 +1,37 @@
...
+
+%description systemd-networkd
+Provides systemd-networkd configuration installed at /etc/systemd/network which asks systemd to manage all interfaces on the system.
+
+%install
+install -d %{buildroot}%{_sysconfdir}/systemd/network
+install -m 644 %{SOURCE1} %{buildroot}%{_sysconfdir}/systemd/network/50-default.network
```
ddstreetmicrosoft, line 31: `/etc` is for users/admins to use; distros should put stuff into `/usr/lib`

base/images/vm-base/vm-base.kiwi
```diff
@@ -114,9 +114,9 @@
         <package name="system-release" />
     </packages>
 
-    <!--
+    <!-- No users added by default, let cloud-init take care of it.
     <users>
-        <user name="root" password="INSERT-PASSWORD-HERE" groups="root" />
+    <user name="root" password="" groups="root" />
```
ddstreetmicrosoft, line 119: I don't know kiwi, but I find it hard to believe this is correct...`root` should be user 0, and this certainly looks like a definition for a regular (non-system) user, which would be quite wrong...are you sure we actually have to define the root user in kiwi config?

base/comps/azurelinux-config/50-default.network
ddstreetmicrosoft: To clarify, `systemd-networkd` will only use *one* `network` file for any specific interface, and this doesn't seem like what we want that network file to look like for any interface

ddstreetmicrosoft:
I split out this PR into what i think are the logical parts (after rebasing to drop some merged bits), [here](https://github.com/ddstreet/azurelinux/commits/network-override/)

to address each logical commit:

[update gitignore](https://github.com/ddstreet/azurelinux/commit/a792eaa72fd5a4003fcf7f4dcc31a5b4bc97ce81)

this is a trivial one, though i'm not sure if it's correct/needed; i think we should drop/ignore it, and later if anyone wants they can open a simple PR to update the .gitignore file with explanation of what should be ignored

[change azurelinux.distro.toml rhel version from 10->11](https://github.com/ddstreet/azurelinux/commit/f8bed49bba89c923a03fd691e7e3c3f58f53e528)

I'm not sure what the reasoning behind this is; I can open a new PR to do this but we should include the reason why

[update kiwi image boot partition size 2g->4g](https://github.com/ddstreet/azurelinux/commit/67758769eb40706ba9107fa0e4ccdfe3beae3de0)

same as above; why are we increasing it? (also, we should probably revisit if we need a boot partition at all, which we probably do not)

[add new azurelinux-config package](https://github.com/ddstreet/azurelinux/commit/c3926cd5cdf09a93bc46601a4f9bd730b56db5d2)

I don't think we need this at all, but this is really the core of what this PR is about. I think for network configuration, we should rely on existing mechanisms which are specific to the deployment; cloud-init for most or all cases.

[update demo-build.sh](https://github.com/ddstreet/azurelinux/commit/72762c1d39957cce517b92c06277ac165af83c5b)

I don't think this file should exist in the repo, but in any case as it says, "Please think ~~twice~~ thrice before you consider adding anything to it." So I think we should drop this commit.

base/comps/azurelinux-config/50-default.network
binujp: Here is the thought behind above. We keep this configuration as generic as possible to let any kind of NIC be managed by systemd-networkd. If the environment is special and one or more NICs need to be managed by something else, it up to the administrator to exclude that interface specifically. Given that we are choosing for azure vm and baremetal instances, this seemed like a safe choice.

We assume that all VMs are brought up by cloud-init. That takes care of enabling dhcp on the most viable interface. In a mult-nic machine we want only one interface to get an address to make it accessible. The rest of the setup can proceed in whatever the user chooses.

What were your thoughts on alternatives to this? This is by no means etched in stone.

base/comps/azurelinux-config/azurelinux-config.spec
binujp: I am not married to this approach. Question is, do we install azl configs irrespective of what settings they are overriding? If yes, then all we care about is delivering configuration as we want. If not, we have to make sure whichever component this override is applied to is installed.

base/comps/azurelinux-config/azurelinux-config.spec
binujp: I am all for it.

base/images/vm-base/vm-base.kiwi
binujp: We do not have to. This was the result of monkey-see-monkey-do approach with the confidence that we are not in production. I thought the root user had gone away. We can drop this. Cloud-init will take care of setting up users and privileges.

christopherco:
> [update kiwi image boot partition size 2g->4g](https://github.com/ddstreet/azurelinux/commit/67758769eb40706ba9107fa0e4ccdfe3beae3de0)
> 
> same as above; why are we increasing it? (also, we should probably revisit if we need a boot partition at all, which we probably do not)

Wouldn't we need a /boot partition for disk encryption scenarios (say through Azure Disk Encryption VM Extension)? I think that's the primary situation I could see needing a /boot partition.

ddstreet:
> > [update kiwi image boot partition size 2g->4g](https://github.com/ddstreet/azurelinux/commit/67758769eb40706ba9107fa0e4ccdfe3beae3de0)
> > same as above; why are we increasing it? (also, we should probably revisit if we need a boot partition at all, which we probably do not)
> 
> Wouldn't we need a /boot partition for disk encryption scenarios (say through Azure Disk Encryption VM Extension)? I think that's the primary situation I could see needing a /boot partition.

For ADE and Confidential Computing encryption, you're right that it would need a separate unencrypted partition for the bootloader and kernel/UKI. For CC, I believe that's the UEFI system partition, but for ADE it seems like it would be the /boot partiton. It does look like [ADE is being deprecated by 2028](https://learn.microsoft.com/en-us/azure/virtual-machines/disk-encryption-overview) though, in favor of host-side encryption, which I think does the decryption before the VM boots (and so no separate /boot would be needed).

I still think that, where possible, it would be good to remove the separate /boot partition.

ddstreetmicrosoft:
I'm closing this PR as we do not want to provide pre-existing systemd-networkd config files; network config should instead be accomplished using cloud-init (or other user-controlled mechanism, e.g. install iso might prompt user).

PR #15628, opened 2026-01-29
prd(ai tools): Add draft PRD for AI repo tooling

dmcilvaney:
Initial draft PRD for AI tooling support in the distro repo


docs/prds/agent-instructions.prd.md
```diff
@@ -0,0 +1,608 @@
...
+   - Note: azldev is under active development; command syntax may change
+
+4. **Repository Hygiene Rules**
+   - Components defined as more than a single line with just the name SHOULD have their own `.comp.toml` file (guideline, not strict rule)
+   - Overlays MUST include a `description` field
+   - Component names should match upstream where possible
+
+5. **Common Workflows** (summary, link to skills for details)
+
+### File-Type Instructions
```
reubeno, line 214: Should add *.kiwi

docs/prds/agent-instructions.prd.md
```diff
@@ -0,0 +1,608 @@
...
+   - Azure Linux is a TOML-based distro definition
+   - Components are defined in `*.comp.toml` files
+   - Overlays modify upstream specs without forking
+
+2. **Key Concepts**
+   - Components vs packages
+   - Upstream distros (Fedora) and how specs are sourced
+   - The overlay system and when to use it
+
+3. **azldev Tool Reference**
```
reubeno, line 202: I'm open to considering it more, but I don't think the azldev reference nor the TOML reference belongs in this repo. It should be linked from here; but it belongs with the azldev project itself.

There will be other project repos that equally need this information too.

prd/agent-instructions.prd.md
```diff
@@ -0,0 +1,608 @@
...
+
+- Querying current component state before making changes
+- Identifying update type (version bump, dependency change, overlay modification)
+- Making changes to component definitions
+- When to migrate from inline to dedicated file
+- Validation and test build workflow
+
+**Versioning and Compatibility Considerations:**
+
+When updating packages, agents MUST be cautious about version changes:
```
reubeno: I think the update policy should be separated out and will end up being differentiated between different segments of the repo.

docs/prds/agent-instructions.prd.md
```diff
@@ -0,0 +1,608 @@
...
+When updating packages, agents MUST be cautious about version changes:
+
+- **Major version bumps** (e.g., 1.x → 2.x): High risk. May break API/ABI compatibility. Require explicit user approval and should note potential downstream impact.
+- **Minor version bumps** (e.g., 1.2 → 1.3): Medium risk. Generally safe but may introduce new dependencies or deprecations.
+- **Patch version bumps** (e.g., 1.2.3 → 1.2.4): Low risk. Usually safe for security fixes and bug fixes.
+- `.toml` files will document the version change policy for each component.
+
+Agents should:
+
+1. **Never assume backwards compatibility** - Check upstream release notes for breaking changes
```
reubeno, line 291: This stuff is great. We should separate "tell me what may change between these versions" from "is it okay to take this upgrade".

prd/agent-instructions.prd.md
```diff
@@ -0,0 +1,608 @@
...
+**Concepts to cover:**
+
+- How to preserve build state on failure for inspection
+- Entering the mock shell environment for interactive debugging
+- Key locations where build artifacts and logs are stored
+- Iterative workflow: inspect → hypothesize → fix → retry
+
+**Key debugging locations:**
+
+- `~/rpmbuild/BUILD/` - Extracted sources and build artifacts
```
reubeno: Not sure if this is reality.

prd/agent-instructions.prd.md
```diff
@@ -0,0 +1,608 @@
...
+
+Both may cover similar topics but serve different purposes. Skills answer "how does this work?" while prompts answer "help me do this now."
+
+#### `add-package.prompt.md`
+
+**Purpose:** Guided workflow for adding a new package to Azure Linux
+
+**Parameters:**
+
+- `package_name` (required) - Name of the package to add
```
reubeno: We should align on terms between packages and components.

prd/agent-instructions.prd.md
```diff
@@ -0,0 +1,608 @@
...
+5. **`component add`** - Adding new packages
+6. **`component mock-shell`** - Interactive debugging (may need special handling for terminal access)
+
+### MCP Integration Success Criteria
+
+- Agents can complete add-package workflow entirely via MCP (no shell commands)
+- Agents can diagnose overlay failures using MCP tool output
+- Error messages from MCP tools are structured and machine-parseable
+
+## azldev Enhancements for Agent Support
```
reubeno: This (and work on this) doesn't belong here to me -- it belongs in work on the azldev repo. Totally happy to collect *feedback* from work there, though.

prd/agent-instructions.prd.md
```diff
@@ -0,0 +1,608 @@
...
+- Interactive debugging via mock shell
+
+**Project operations:**
+
+- Querying project-level configuration
+- Managing distro definitions
+
+## Appendix: Overlay Quick Reference
+
+Overlays modify upstream specs without forking. Below are illustrative examples; see `external/schemas/azldev.schema.json` for the complete overlay type definitions and required fields.
```
reubeno: There's an overlays.toml in azldev repo that may be helpful as input.

docs/prds/agent-instructions.prd.md
```diff
@@ -0,0 +1,608 @@
+# PRD: Agent Instructions and Skills for Azure Linux
+
+Title: PRD: Agent Instructions and Skills for Azure Linux
```
reubeno, line 3: Thanks for getting the ball rolling with this!

reubeno:
Can we place any PRDs for this repo under `docs/`?

docs/prds/agent-instructions.prd.md
dmcilvaney: My thinking was have the azldev tool able to emit these files and regenerate/add them to repos. The reality is if want the ai to understand these files we need the instructions files etc. They can either be repo specific, or global for the user, but I'm not a fan of bloating the agents' contexts for ALL projects just to support that. Much better to just have them where they are actually used.

prd/agent-instructions.prd.md
dmcilvaney: My hope was it would be encoded as part of the azldev component, which we can just have the agents query. Doesn't necessarily need to be in any specific spot in the repo.

prd/agent-instructions.prd.md
dmcilvaney: ugh... thought I got all the super specific bits out and replaced with general directions

prd/agent-instructions.prd.md
dmcilvaney: I didn't want to make it as hard as 'requirements', but I do feel its equivalent. Where should I track these? I feel like PRDs should generally be self-contained where possible

dmcilvaney: ok, filed these as an issue in the azldev repo

dmcilvaney:
Add section on using control tower?

docs/prds/agent-instructions.prd.md
reubeno: Discussed offline; what's important to me is that the *source of truth* for the docs be in the same repo as `azldev`. There will be multiple repos with TOML files that use azldev and the same schema. I care less about duplication than I do about there being divergent versions of the docs that are independently edited.

Provided there's a way for us to accomplish this, I'm okay with them getting checked in under a dir that's clearly marked as generated/cached content if it helps the AIs.

prd/agent-instructions.prd.md
reubeno: I *think* we're on the same page: policy should be encoded in TOML files. (I didn't want to encode the policy in agent instructions.) Agents should be able to *understand* that encoded policy, of course.

dmcilvaney:
explicitly version the configs (generate them via azldev tool?)

PR #15648, opened 2026-02-02
Add firewalld configuration override specific to azl

binujp:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

Firewalld configuration is driven of a top level config files and then zones tied to some or all interfaces and the zone configuration itself has the rules for the net filters. We are using nftables, an azurelinux zone policy which opens only ssh and optionally dhcpv6-client port. This change adds a sub-package to azurelinux-config which enables all that.

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


base/comps/azurelinux-config/10-azurelinux-firewalld.preset
```diff
@@ -0,0 +1 @@
+enable firewalld.service
```
reubeno: Have you looked at the 90-default.preset in azurelinux-release; it's already enabled. I would expect presets go there.

base/comps/azurelinux-config/50-default.network
```diff
@@ -0,0 +1,6 @@
+# We want systemd to manage all interfaces
```
reubeno, line 1: @ddstreetmicrosoft I believe you had some input on the right approach here, what other distros typically do, etc.?

base/images/vm-base/vm-base.kiwi
```diff
@@ -114,9 +116,8 @@
         <package name="system-release" />
     </packages>
 
-    <!--
     <users>
-        <user name="root" password="INSERT-PASSWORD-HERE" groups="root" />
+          <user name="root" password="INSERT-PASSWORD-HERE" groups="root" />
```
reubeno, line 120: Please remove these changes; you should be able to use cloud-init config to inject user account configs / credentials (e.g., via `azldev image boot`).

I'm happy to chat separately to get you unblocked with that.

base/images/images.toml
reubeno, line 1: Are you up to date with the target branch? These comments were removed last week.

scripts/demo-build.sh
```diff
@@ -1,9 +1,14 @@
...
+exec 3> ./demo-build.trc
+BASH_XTRACEFD=3
+set -x
+
+echo -e "\e[0;33mbash trace, -x output is in ./demo-build.trc\e[0m"
 
 #
 # NOTE: This script is a throwaway script. Please think ~~twice~~ thrice before you
-# consider adding anything to it.
+# consider adding anything to it. Let us push all dev-tooling into azldev.
```
reubeno, line 11: Agreed!

scripts/demo-build.sh
```diff
@@ -13,29 +18,75 @@ if [ ! -f azldev.toml ]; then
...
+         --description ./base/images/container-base \
+         --target-dir ./base/out/images \
+         --add-repo="file:///$PWD/base/out,rpm-md,azl,1"
+
+    # Run a command in the container to verify.
+    xzcat ./base/out/images/azl4-container-base.x86_64-0.1.docker.tar.xz | docker load
+    docker run -it --rm microsoft/azurelinux/base/core:4.0 cat /etc/os-release
+}
+
+function build_vm_image() {
```
reubeno, line 69: When you have a moment, try using `azldev image build` and `azldev image boot`. Holler if you have difficulty.

base/images/vm-base/vm-base.kiwi
```diff
@@ -86,7 +88,7 @@
         <package name="sudo" />
         <package name="systemd-boot" />
         <package name="systemd-networkd" />
-        <package name="systemd-resolved" />
+        <package name="azurelinux-config-systemd-networkd" />
```
reubeno, line 91: Thanks for keeping these near each other; probably a good idea to keep this list grouped/orderly over time.

base/comps/azurelinux-config/azurelinux-config.spec
```diff
@@ -0,0 +1,63 @@
...
+%install
+install -d %{buildroot}%{_sysconfdir}/systemd/network
+install -m 644 %{_sourcedir}/50-default.network %{buildroot}%{_sysconfdir}/systemd/network/50-default.network
+install -d %{buildroot}%{_sysconfdir}/systemd/system-preset
+install -m 644 %{_sourcedir}/10-azurelinux-firewalld.preset %{buildroot}%{_sysconfdir}/systemd/system-preset/10-azurelinux-firewalld.preset
+install -d %{buildroot}%{_sysconfdir}/firewalld/zones
+install -m 644 %{_sourcedir}/azurelinux-firewalld-zone.xml %{buildroot}%{_sysconfdir}/firewalld/zones/azurelinux.xml
+install -m 644 %{_sourcedir}/firewalld-azurelinux.conf %{buildroot}%{_sysconfdir}/firewalld/firewalld-azurelinux.conf
+
+%post firewalld
```
reubeno, line 46: Why is this being installed via %post? Can the symlinks be included directly in %files?

base/comps/azurelinux-config/firewalld-azurelinux.conf
```diff
@@ -0,0 +1,121 @@
...
+# default zone
+# The default zone used if an empty zone string is used.
+# Default: public
+DefaultZone=azurelinux
+
+# Clean up on exit
+# If set to no or false the firewall configuration will not get cleaned up
+# on exit or stop of firewalld.
+# Default: yes
+CleanupOnExit=yes
```
reubeno, line 12: Naive question: is it desirable to remove firewall config on firewalld stop? What if the service is unexpectedly stopped -- will that leave the network more exposed than it ought to be?

base/comps/azurelinux-config/azurelinux-firewalld-zone.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<zone>
```
reubeno, line 2: Naive question: will this zone apply to all non-loopback network interfaces by default?

reubeno, line 2: Other naive question: I see there's a `target` you can set on the `<zone/>`. If left unspecified, will DROP be the default action for unmatched packets?

base/comps/azurelinux-config/azurelinux-firewalld-zone.xml
```diff
@@ -0,0 +1,10 @@
...
+  <short>azurelinux</short>
+  <description>For use in azure cloud vms with zero trust policy. Only ssh is allowed by default. If we need dynamic IPv6 addressing dhcpv6-client will have to be enabled.</description>
+  <service name="ssh"/>
+  <!--
+  <service name="dhcpv6-client"/>
+  -->
+  <forward/>
```
reubeno, line 9: Why is the `<forward/>` here? Do we need to support this by default?

base/comps/azurelinux-config/azurelinux-config.spec
```diff
@@ -0,0 +1,63 @@
...
+Summary:        Azure Linux firewall policies
+Requires:       firewalld
+Supplements:    firewalld
+
+%description firewalld
+Provides Azure Linux specific policies and zones for firewalld. 
+
+%install
+install -d %{buildroot}%{_sysconfdir}/systemd/network
+install -m 644 %{_sourcedir}/50-default.network %{buildroot}%{_sysconfdir}/systemd/network/50-default.network
```
reubeno: Please see https://docs.fedoraproject.org/en-US/packaging-guidelines/RPM_Source_Dir/

>Packages which use files itemized as Source# files, must refer to those files by their %{SOURCE#} macro name, and must not use $RPM_SOURCE_DIR or %{_sourcedir} to refer to those files.

>This is done to ensure that Fedora SRPMS are properly generated. If a Source# item is renamed, a spec which refers to its old name may succeed locally (because the file is still in %{_sourcedir} along with the new file), but the proper file will not be included in the SRPM.

base/comps/azurelinux-config/azurelinux-config.spec
```diff
@@ -0,0 +1,63 @@
...
+
+Source1:        50-default.network
+Source2:        10-azurelinux-firewalld.preset
+Source3:        azurelinux-firewalld-zone.xml
+Source4:        firewalld-azurelinux.conf
+
+BuildArch:      noarch
+
+%description
+Provides AZL specifc configuration overrides via sub-packages. These sub-packages will be pulled in by the relevant owning packages via reverse dependencies.
```
reubeno: Typo: specifc

reubeno (changes requested):

base/comps/azurelinux-config/10-azurelinux-firewalld.preset
binujp: Yes, I saw that presets all go there. My thought was unless azurelinux configuration is available should we enable and start firewalld. If we are sure that is the pattern we want to follow that sounds good.

base/comps/azurelinux-config/50-default.network
binujp: The most robust approach is to let one framework manage all interfaces. Given we are considering only azure cloud deployments I would rather not complicate things. I see there may be cases where the machine is in a DMZ zone and can have NICs in different networks. Let us consider those as special cases to be handled by the customer.

base/comps/azurelinux-config/azurelinux-config.spec
binujp: TIL something, thanks Reuben!

base/comps/azurelinux-config/azurelinux-config.spec
binujp: I assumed since there will be an existing symlink to the default conf, install may fail. I will verify and if it works do it via %files.

base/comps/azurelinux-config/azurelinux-firewalld-zone.xml
binujp: All interfaces get assigned to the default zone unless explicitly assigned. That means all of them should come under azurelinux zone.

binujp: It will apply the default action on any unmatched packet. That is reject. I suppose there is a way to change the default, I have not tried that.

base/comps/azurelinux-config/azurelinux-firewalld-zone.xml
binujp: This allows all interfaces within that zone to forward packets to each other. Load-balancing or NIC failover will need this. I do not have a strong opinion about whether we need this by default. If customers are allowed to spin-up VMs with multiple NICs it is better to have this default.

base/comps/azurelinux-config/firewalld-azurelinux.conf
binujp: Yes; if we err on the side of paranoid and leave iptables rules in place after firewalld is shutdown, that can cause even more pain and upsets. We should rely on systemd to restart firewalld when required.

base/images/vm-base/vm-base.kiwi
binujp: So many ways to do this from alphabetical to functional and domains and ...

base/images/vm-base/vm-base.kiwi
binujp: Oops!

base/images/images.toml
binujp: I did rebase, this change was on top of my other PR https://github.com/microsoft/azurelinux/pull/15599. Maybe that messed this up? I will check.

scripts/demo-build.sh
binujp: That is the invocation here. Did you mean to say get rid of the function wrapper?

base/comps/azurelinux-config/10-azurelinux-firewalld.preset
reubeno: Those presets were mostly just inherited from Fedora; my recollection is that many services are preset-enabled in base Fedora, independent of product. Presumably the thinking is that a customer who doesn't want to use it can arrange not to install the package (or to remove it).

base/comps/azurelinux-config/azurelinux-firewalld-zone.xml
reubeno: As long as it rejects if unmatched--that was my primary question.

base/comps/azurelinux-config/azurelinux-config.spec
reubeno: Ah, I didn't realize another package owns this .conf file.

Interestingly the Fedora guide uses firewalld.conf as a specific example here: https://docs.fedoraproject.org/en-US/packaging-guidelines/Per-Product_Configuration/ -- with some guidance on per-product configs. That sounds applicable?

reubeno: Okay, now that I'm reading through the upstream `firewalld.spec` -- this seems like a case where they expect the base `firewalld` package to carry product-specific .conf files and to switch on VARIANT_ID to select the right one.

There's an interesting question here as to what our VARIANT_ID will be and whether we should go with an overlay as opposed to the layered config package that re-symlinks.

If both `firewalld` and our config package are both trying to symlink -- is there a risk that they'll thrash? i.e., if `firewalld` happens to get reinstalled/upgraded, will it "steal the symlink back"?

base/comps/azurelinux-config/50-default.network
reubeno: Ah, I should have clarified. My question was more around blanket DHCP enablement -- not having systemd-networkd manage the network interfaces.

ddstreetmicrosoft: This is [definitely wrong](https://github.com/microsoft/azurelinux/pull/15599/changes/771d9249a8c05f222d40ca5ac38b5fff1cd07bc9#r2759582097)

base/comps/azurelinux-config/azurelinux-config.spec
```diff
@@ -0,0 +1,60 @@
...
+
+%package firewalld
+Summary:        Azure Linux firewall policies
+Requires:       firewalld
+Supplements:    firewalld
+
+%description firewalld
+Provides Azure Linux specific policies and zones for firewalld. 
+
+%install
```
ddstreetmicrosoft, line 38: as with the other PR, distro-provided files should go into `/usr/lib`; the `/etc` dir is generally for users/admins

base/comps/azurelinux-config/firewalld-azurelinux.conf
```diff
@@ -0,0 +1,121 @@
+# firewalld config file
```
ddstreetmicrosoft, line 1: I would greatly like to see a PR with *individual commits* for every single configuration parameter added here, with specific explanation for why we need to add the config

base/comps/components.toml
```diff
@@ -57,6 +57,7 @@ spec = { type = "upstream", upstream-distro = { name = "fedora", version = "rawh
 [components.file]
 [components.filesystem]
 [components.findutils]
+[components.firewalld]
```
ddstreetmicrosoft, line 60: Is this PR both adding the `firewalld` package itself, *and* adding azl-specific config in a separate package? IMHO simply adding `firewalld` should be completely separate - that should be simple to review/commit.

base/images/vm-base/vm-base.kiwi
```diff
@@ -18,14 +18,14 @@
             format="vhdx"
             initrd_system="dracut"
             filesystem="ext4"
-            kernelcmdline="console=ttyS0 rd.shell=0 systemd.getty_auto=false"
```
ddstreetmicrosoft, line 21: what do these changes have to do with firewalld configuration overrides?

ddstreet:
Similiar to [my comment](https://github.com/microsoft/azurelinux/pull/15599#issuecomment-3844155465) in the [networkd pr](https://github.com/microsoft/azurelinux/pull/15599), I set up [a branch](https://github.com/ddstreet/azurelinux/commits/firewalld-config/) that I think logically breaks up this pr into parts, specifically:

[feat: add firewalld package to kiwi vm-base definition](https://github.com/ddstreet/azurelinux/commit/2039468a5c13f0b918a7a5f516dfd759dc3efb8f)
[ix: remove duplicate systemd-resolved entry from kiwi vm-base definition](https://github.com/ddstreet/azurelinux/commit/22ef9d53353f25ad39ceb8617b3369a7769218f9)

i put both of these in [a separate pr](https://github.com/microsoft/azurelinux/pull/15726)

[update gitignore file](https://github.com/ddstreet/azurelinux/commit/45a90d7a0eafa0638bcfb518f00532bd17e7216f)
[update azurelinux.distro.toml to change rhel from 10->11](https://github.com/ddstreet/azurelinux/commit/d26947592b7698ae8a2d6633b37e2bb7cb940bff)

these 2 are also in [the network pr](https://github.com/microsoft/azurelinux/pull/15599); i think the gitignore change can wait for @binujp to get back, and the rhel version bump needs explanation

[uncomment kiwi vm xml to add definition for root user](https://github.com/ddstreet/azurelinux/commit/35eaa76e6c7dfa13f4666a88b28ace1d56ee5535)

this shouldn't be merged, in fact i think the root user definition (even though its commented out) needs to be removed from the file

[add azurelinux-config spec file with a definition for azurelinux-config-firewalld](https://github.com/ddstreet/azurelinux/commit/1daba66d987ab1045bf6e51138b1ef59d299a352)

this defines the `azurelinux-config` rpm (spec and firewall subpackage, plus config files). I do not think this is how we should deliver firewalld configuration; I think we should look at how the firewalld package currently provides distro-provided config, and do something similar.

[add systemd-networkd config to azurelinux-config package](https://github.com/ddstreet/azurelinux/commit/e2e288df2de444284f731c50e5995abc526d2312)

this seems to be carried over from [the networkd pr](https://github.com/microsoft/azurelinux/pull/15599), and doesn't belong in this pr

[feat: add azurelinux-config packages to kiwi vm-base definition](https://github.com/ddstreet/azurelinux/commit/f08b8df96f328d3c63d211c30d16a13688b35d79)

since I don't think we should create the `azurelinux-config` package, we don't need to add its rpms to the kiwi vm-base image definition.

[update scripts/demo-build.sh](https://github.com/ddstreet/azurelinux/commit/a67b007cf3eef858f3784bd617016efe7461b536)

same opinion as the other pr; don't update this script, as it should be removed

ddstreet:
I rebased my branch so the links in the previous comment are stale, but you can go direction to my branch to see the commits:
https://github.com/ddstreet/azurelinux/commits/firewalld-config/

ddstreet:
Ok, after comparing the `firewalld-azurelinux.conf` and `azurelinux-firewalld-zone.xml` to what's provided out of the box, there are *no* differences in the `.conf` file (besides default zone name pointing to the azurelinux zone), and the azurelinux zone only differs from the standard default zone (public) by closing the `mdns` and `dhcpv6-client` ports.

I think we should close this PR and go with the firewalld defaults as currently provided. Closing the `dhcpv6-client` port in particular will very likely lead to confused and complaining users when their cloud ipv6 configuration doesn't work properly. I don't know how much `mdns` is used in the cloud, but we can decide to close that later if really needed.

ddstreetmicrosoft:
I opened pr #15740 to remove mdns from the public zone, and am closing this pr

PR #15665, opened 2026-02-03
Update kernel component to build without rhel macro

rlmenge:
Summary
Updates the kernel component configuration to build successfully on Azure Linux by setting rhel=0.

Changes
Added defines = { rhel = "0" } to the kernel component in components.toml
Problem
The upstream Fedora kernel spec file checks for the %rhel macro and, when set (e.g., rhel=10), adds BuildRequires for RHEL-specific certificate packages that don't exist in Azure Linux or Fedora. This caused the kernel build to fail with missing dependency errors.

Solution
Explicitly set rhel=0 to disable RHEL-specific code paths in the kernel spec, allowing the build to complete using standard Fedora dependencies.

Testing
Built kernel successfully with azldev comp build kernel
Built VM image with the custom kernel using azldev image build vm-base
Verified VM boots in QEMU

### DIFF
--- a/base/comps/components.toml
+++ b/base/comps/components.toml
@@ -96,7 +96,6 @@ spec = { type = "upstream", upstream-distro = { name = "fedora", version = "rawh
 [components.jq]
 [components.json-c]
 [components.kbd]
-[components.kernel]
 [components.kernel-srpm-macros]
 [components.keyutils]
 [components.kmod]
--- a/base/comps/kernel/kernel.comp.toml
+++ b/base/comps/kernel/kernel.comp.toml
@@ -1,0 +1,6 @@
+[components.kernel]
+
+# Set rhel to 0 to avoid deps on non-existent cert packages
+# (rhel=10 would engage RHEL-specific BuildRequires)
+[components.kernel.build]
+defines = { rhel = "0" }


base/comps/components.toml
```diff
@@ -96,7 +96,13 @@ spec = { type = "upstream", upstream-distro = { name = "fedora", version = "rawh
 [components.jq]
 [components.json-c]
 [components.kbd]
+
 [components.kernel]
```
reubeno: It's still getting ratified, but the current thinking is that we will move a component off into its own dir + .comp.toml file when it becomes more than an unmodified import from Fedora using defaults -- so this file stays relatively readable and just a simple list of imported "as-is" components.

Given that and given that I expect we may have more TOML config for the kernel over time, would you mind doing that now? i.e. -- remove the kernel entirely from this file, create a `kernel` subdir, and a `kernel.comp.toml` file under that with the `[components.kernel]` section and the new `.build` subsection. (`rpm` should be an example to follow.)

rlmenge: Can do!

base/comps/components.toml
```diff
@@ -96,7 +96,13 @@ spec = { type = "upstream", upstream-distro = { name = "fedora", version = "rawh
...
+# Set rhel to 0 to avoid deps on non-existent cert packages
+# (rhel=10 would engage RHEL-specific BuildRequires)
+[components.kernel.build]
+defines = { rhel = "0" }
+
```
tobiasb-ms: suggestion: I'd prefer this be put in its own folder and component file.

tobiasb-ms (changes requested):

base/comps/components.toml
rlmenge: @reubeno recommended the same :) . Doing so now

PR #15711, opened 2026-02-03
increase vm-base image size from 2g->4g

ddstreet:
The larger size is needed to accomodate increasing number of packages. Can be revisited/adjusted later as well.

Cc @binujp


reubeno:
Okay for now. For the future, we should make sure we're aligning with what other base images do.

PR #15714, opened 2026-02-04
Add minimal package list for marketplace image

liunan-ms:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This PR trims the marketplace image package list from 340 to 84 by removing the dependency packages.
This marketplace image package list contains all the packages that are intentionally included in marketplace image.
Validated in both QEMU and azl4 azure VM that the total count of installed packages is 407. 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/base/images/vm-base/vm-base.kiwi
+++ b/base/images/vm-base/vm-base.kiwi
@@ -128,7 +128,6 @@
         <package name="grub2-efi-x64-modules" />
         <package name="grub2-efi-x64" />
         <package name="shim" arch="x86_64" />
-        <package name="system-release" />
     </packages>
 
     <!--


package-list/marketplace-image-packages.txt
```diff
@@ -295,46 +99,15 @@ gdb
...
-libcurl
-krb5-libs
-basesystem
 dracut-network
-filesystem
 grub2-efi-x64-modules
 grub2-efi-x64
-shim
-system-release
-auoms
```
binujp, line 319: We need this. It has to come from azurelinux-official-ms-non-oss repo. Can we add a comment to that effect to enable it when appropriate?

package-list/marketplace-image-packages.txt
```diff
@@ -219,71 +73,21 @@ python3-packaging
...
-libuv
 autogen-libopts
 libtool
-make
 vim
 rsyslog
-cronie
-cronie-anacron
 tini
 postgresql-libs
```
binujp: I am not sure fedora provides this. It is the client and libraries for postgres provided by postgreql and libpq5 packages. In azl3 rsyslog needs it. I am not sure the same holds true for Fedora.

package-list/marketplace-image-packages.txt
```diff
@@ -1,201 +1,58 @@
...
-hypervvssd
-dracut
 dracut-kiwi-oem-repart
-grub2-tools-minimal
 grubby
 kernel
-kernel-modules
-kmod
 grub2
 dnf
```
binujp: I would make it explicit that this is dnf5 which provides dnf.

package-list/marketplace-image-packages.txt
```diff
@@ -1,201 +1,58 @@
...
-cyrus-sasl-lib
-gnutls
-gpgme
-libacl
-libattr
-libgcrypt
-libgpg-error
-libselinux
-openssh
 openssh-clients
```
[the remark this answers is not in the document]
tobiasb-ms: I agree -- we probably need `oenssh-server`

package-list/marketplace-image-packages.txt
```diff
@@ -1,201 +1,58 @@
...
-openssh-server
-openssl
-openssl-libs
```
tobiasb-ms, line 34: Do we really not need `oenssl-libs`?

package-list/marketplace-image-packages.txt
liunan-ms: will double check in Fedora.

package-list/marketplace-image-packages.txt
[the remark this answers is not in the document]
liunan-ms: For `openssh-server`, it is automatically installed as a dependency of `WALinuxAgent`. 
The current list only has head packages. Nothing gets added here if it is a dependency. For example, `systemd` and `bash` are not included in the list as they are both dependencies of some other packages. 
But we eventually want it to be a list of packages that are intentionally needed, explicitly include intentional packages no matter if they're dependencies of others. 

@tobiasb-ms May I have your review of all of them and suggest the ones that we need to explicitly include in the marketplace image? I'll collect the feedback and integrate the next version of the list. Thanks!

package-list/marketplace-image-packages.txt
liunan-ms: Removed postgresql-libs from the list. If any package needs it as a dependency, it'll be installed but not explicitly included in this list.

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
+        <package name="cracklib-dicts" />
+        <package name="cronie" />
         <package name="cryptsetup" />
-        <package name="dnf-plugins-core" />
-        <package name="dnf" />
-        <package name="dracut-config-generic" />
-        <package name="dracut-kiwi-oem-repart"/>
-        <package name="glibc-langpack-en" />
-        <package name="glibc-locale-source" />
-        <package name="glibc-minimal-langpack" />
```
reubeno, line 62: Not for this PR, but we should follow up by looking into `glibc-minimal-langpack` and how it relates to `glibc-langpack-en`.

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
+        <package name="cyrus-sasl" />
+        <package name="device-mapper-event" />
+        <package name="dnf5" />
+        <package name="dnf5-plugins" />
+        <package name="dracut-kiwi-oem-repart" />
+        <package name="elfutils" />
+        <package name="elfutils-default-yama-scope" />
+        <package name="file" />
+        <package name="firewalld" />
+        <package name="glib" />
```
reubeno: We should remove `glib`; it's a library that will get pulled in if needed.

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
         <package name="glibc" />
+        <package name="glibc-langpack-en" />
+        <package name="grub2" />
         <package name="grubby" />
-        <package name="gzip" />
```
reubeno, line 65: Looking at what was here before, we should probably leave these 4 packages in the list:

```
gzip
hostname
iproute
iputils
```

The networking packages in particular are very helpful to have preinstalled to debug networking issues  (in which case dnf can't be used yet to install them).

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
-        <package name="hostname" />
-        <package name="iproute" />
-        <package name="iputils" />
         <package name="irqbalance" />
         <package name="kernel" />
-        <package name="less" />
+        <package name="kernel-headers" />
```
reubeno: Are `kernel-headers` required?

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
+        <package name="kernel-tools" />
+        <package name="lvm2" />
+        <package name="lz4" />
         <package name="man-db" />
-        <package name="ncurses" />
+        <package name="nano" />
+        <package name="ncurses-term" />
+        <package name="net-tools" />
+        <package name="newt" />
```
reubeno: I wasn't familiar with this (`newt`), but it looks like a library; I'd recommend omitting from the list.

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
+        <package name="nftables" />
+        <package name="nghttp2" />
```
reubeno: Looked at this (`nghttp2`) briefly too -- it doesn't seem like it's required for base images.

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
         <package name="openssh-server" />
-        <package name="passwd" />
-        <package name="pciutils" />
-        <package name="policycoreutils" />
-        <package name="procps-ng" />
-        <package name="psmisc" />
-        <package name="rng-tools" />
-        <package name="rpm-plugin-systemd-inhibit" />
-        <package name="rpm" />
+        <package name="pcre2-tools" />
```
reubeno: This (`pcre2-tools`) seems a bit obscure; I'm not sure why it's on the list? Some of the pcre libraries it depends on will get pulled in if needed, but I don't think they need to be listed here.

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
+        <package name="pinentry" />
```
reubeno: Ditto; I'm not familiar with this tool (`pinentry`), but I wouldn't expect it to be needed in the base image.

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
+        <package name="rsync" />
+        <package name="rsyslog" />
         <package name="selinux-policy-targeted" />
         <package name="setup" />
-        <package name="shadow-utils" />
+        <package name="shadow-utils-subid" />
```
reubeno, line 100: We should also explicitly list `shadow-utils`, as its executables are ones we intentionally do want (e.g., `useradd`).

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
         <package name="bash" />
         <package name="ca-certificates" />
         <package name="coreutils" />
-        <package name="system-release" />
 
         <!-- Minimal group -->
+        <package name="acl" />
+        <package name="attr" />
         <package name="audit" />
+        <package name="autogen-libopts" />
```
reubeno: Looks like `autogen-libopts` is a library-only package (just carries an `.so`). Recommend omit

base/images/vm-base/vm-base.kiwi
```diff
@@ -39,63 +39,83 @@
...
+        <package name="bash-completion" />
+        <package name="bc" />
         <package name="binutils" />
+        <package name="bind-license" />
```
reubeno: I don't think we need to explicitly list `bind-license`; it's provided by `bind-libs`, which is pulled in by `bind-utils`. Leaving `bind-utils` is a good idea; it's intentional for us to include `nslookup` et al.

package-list/marketplace-image-packages.txt
```diff
@@ -1,340 +1,84 @@
...
-pam
-readline
-sed
-shadow-utils
-shadow-utils-subid
-systemd
-systemd-libs
-util-linux
-which
+acl
```
reubeno: Now that you're directly updating `vm-base.kiwi`, I don't think you need to have this file as a copy; feel free to delete. We can always programmatically extract the flat list.

base/images/vm-base/vm-base.kiwi
liunan-ms: `glibc-minimal-langpack` is only sufficient if using a minimal locale like `C.UTF-8`, but it typically does not guarantee `en_US` is available.

Since we have kiwi definition:
```
<locale>en_US</locale>
```
we install `glibc-langpack-en` so that `en_US / en_US.UTF-8` locale data actually exists in the image, so drop `glibc-minimal-langpack` from the list.

base/images/vm-base/vm-base.kiwi
liunan-ms: I double checked the azl 3.0 azure VM where `kernel-headers` is not installed by default, and all other distros don't have this package installed by default, so I removed it from this list.

PR #15726, opened 2026-02-04
Vm base definition updates

ddstreet:
2 minor updates to the kiwi `vm-base` image definition:

1) add `firewalld` package
2) remove duplicate `systemd-resolved` entry


ddstreet:
oops, seems I already had one of these patches in #15709

PR #15733, opened 2026-02-05
ci: adopt zizmor for workflow validation

reubeno:
Adopts https://github.com/zizmorcore/zizmor for GitHub workflow linting. This is something we've enabled in other repositories to good effect.

### DIFF
--- a/.github/workflows/check-workflows.yml
+++ b/.github/workflows/check-workflows.yml
@@ -1,0 +1,47 @@
+name: "Check workflows"
+on:
+  # Only trigger for changes that touch workflow files.
+  push:
+    branches: [ "tomls/base/main" ]
+    paths: [ ".github/workflows/**" ]
+  pull_request:
+    branches: [ "tomls/base/main" ]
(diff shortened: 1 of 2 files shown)


.github/workflows/check-workflows.yml
```diff
@@ -0,0 +1,40 @@
+name: "Check workflows"
+on:
+  push:
+    branches: [ main ]
```
ddstreetmicrosoft: Is `main` the correct branch here? Or should it be `tomls/base/main`?

.github/workflows/check-workflows.yml
```diff
@@ -0,0 +1,40 @@
...
+  pull_request:
+    branches: [ main ]
```
ddstreetmicrosoft: here also re: branch name

.github/workflows/check-workflows.yml
reubeno: Oof -- good call! That needs correction; copy-paste error for sure. I'll fix these.

PR #15734, opened 2026-02-05
ci: add workflow to enforce conventional commits

reubeno:
We're planning to adopt Conventional Commits in this branch; this workflow helps guide contributors to following the guidelines.

### DIFF
--- a/.github/workflows/check-pr-title.yml
+++ b/.github/workflows/check-pr-title.yml
@@ -37,7 +37,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - uses: marocchino/sticky-pull-request-comment@773744901bac0e8cbb5a0dc842800d45e9b2b405 # v2.9.4
-        # When the previous steps fails, the workflow would stop. By adding this
+        # When the previous step fails, the workflow would stop. By adding this
         # condition you can continue the execution with the populated error message.
         if: always() && (steps.lint_pr_title.outputs.error_message != null)
         with:
@@ -56,7 +56,7 @@ jobs:
             ```
 
       # Delete a previous comment when the issue has been resolved
-      - if: ${{ steps.lint_pr_title.outputs.error_message == null }}
+      - if: steps.lint_pr_title.outputs.error_message == null
         uses: marocchino/sticky-pull-request-comment@773744901bac0e8cbb5a0dc842800d45e9b2b405 # v2.9.4
         with:
           header: pr-title-lint-error


.github/workflows/check-pr-title.yml
```diff
@@ -0,0 +1,61 @@
...
+# Workflows that run on 'pull_request_target' trigger need to be carefully
+# reviewed since they run in the context of the PR target and consume unvalidated
+# input controlled by a PR submitter. We've reviewed this workflow and
+# allow-listed it via the 'zizmor' comment below.
+#
+
+name: "Check PR Title"
+
+on:
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
```
ddstreetmicrosoft, line 14: do we need to restrict this to only PRs against `tomls/base/main`? We don't want this running for *all* PRs against the entire azl repo, right?

.github/workflows/check-pr-title.yml
```diff
@@ -0,0 +1,61 @@
...
+permissions: {}
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write # Needed to post comments on PR
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
```
ddstreetmicrosoft: is this creating a dependency on an external repo? if so, we should probably clone and use our own repo/branch right?

.github/workflows/check-pr-title.yml
```diff
@@ -0,0 +1,61 @@
...
+        id: lint_pr_title
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: marocchino/sticky-pull-request-comment@v2
```
ddstreetmicrosoft: same question as above about external repo use?

.github/workflows/check-pr-title.yml
```diff
@@ -0,0 +1,61 @@
...
+
+            Details:
+
+            ```
+            ${{ steps.lint_pr_title.outputs.error_message }}
+            ```
+
+      # Delete a previous comment when the issue has been resolved
+      - if: ${{ steps.lint_pr_title.outputs.error_message == null }}
+        uses: marocchino/sticky-pull-request-comment@v2
```
ddstreetmicrosoft: same question as above about external repo use?

.github/workflows/check-pr-title.yml
```diff
@@ -0,0 +1,61 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: marocchino/sticky-pull-request-comment@v2
+        # When the previous steps fails, the workflow would stop. By adding this
+        # condition you can continue the execution with the populated error message.
+        if: always() && (steps.lint_pr_title.outputs.error_message != null)
+        with:
+          header: pr-title-lint-error
+          message: |
```
ddstreetmicrosoft, line 45: Can we also include a TL;DR on common formatting? Probably 99% of commits will want to use just a few common formats, e.g. `feat:`, `fix:`, etc.

On the `conventialcommits.org` page, even the `Quick Summary` isn't a good reference (IMHO) on the most common uses, it takes a bit of reading to see what options there are

.github/workflows/check-pr-title.yml
reubeno: Good call!

.github/workflows/check-pr-title.yml
reubeno: It's a dependency, but it's an expected one for GitHub workflows -- and our OSS office's guidance allows use of them, provided we vet them, look for other projects in our org that use them, etc. These two are ones we've been using for quite some time elsewhere and I see various other projects in this GitHub org using.

The one thing that this PR was missing was hash-pinning the precise commit of the external repo to avoid supply-chain attacks that mutate the tags with injected changes.

.github/workflows/check-pr-title.yml
reubeno: _See above._

.github/workflows/check-pr-title.yml
reubeno: See above.

.github/workflows/check-pr-title.yml
reubeno: I'll add this to the help text -- good suggestion.

PR #15736, opened 2026-02-05
fix: do not pin to older gcc

reubeno:
gcc14 is only present for compatibility purpose; we should be using `gcc`.


ddstreetmicrosoft (approved):
agreed

PR #15737, opened 2026-02-05
fix: correct python version to latest

reubeno:
The latest stable Python3 release is `3.14`--that's what we should be using instead of the older `3.9`.

### DIFF
--- a/base/comps/components.toml
+++ b/base/comps/components.toml
@@ -259,7 +259,7 @@ spec = { type = "upstream", upstream-distro = { name = "fedora", version = "rawh
 [components.python-systemd]
 [components.python-urllib3]
 [components.python-wcwidth]
-[components.'python3.9']
+[components.'python3.14']
 [components.pytz]
 [components.PyYAML]
 [components.quota]


base/comps/components.toml
```diff
@@ -259,7 +259,7 @@ spec = { type = "upstream", upstream-distro = { name = "fedora", version = "rawh
 [components.python-systemd]
 [components.python-urllib3]
 [components.python-wcwidth]
-[components.'python3.9']
+[components.'python3.14']
```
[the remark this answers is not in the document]
ddstreetmicrosoft: this is wrong, f43 definitely does [provide 3.14](https://src.fedoraproject.org/rpms/python3.14)

ddstreetmicrosoft, line 262: Since f43's default/main python is 3.14, the package is actually `python3` not `python3.14`. I don't think this changes for the lifetime of the specific release, so we should probably leave it as `python3`?

https://src.fedoraproject.org/rpms/python3.14/blob/rawhide/f/python3.14.spec#_60

ddstreetmicrosoft: oh wait, this is the source package, not the binary package; ok yeah this should be 3.14 then

PR #15738, opened 2026-02-05
feat: enable wget back-compat package for wget2

tobiasb-ms:
The `wget2` spec provides a way to build a back-compat package that will install symlinks so users can rely on `wget`. It is disabled if `rhel` is set to `<11` and we use `rhel = 10`.

This change provides the back-compat package by always setting the `as_wget` `%bcond` during build.

### DIFF
--- a/base/comps/wget2.comp.toml
+++ b/base/comps/wget2.comp.toml
@@ -1,5 +1,5 @@
 [components.wget2]
 
 [components.wget2.build]
-# Supply the wget compatility package.
+# Supply the wget compatibility package.
 with = ["as_wget"]


base/comps/wget2.comp.toml
```diff
@@ -0,0 +1,12 @@
+[components.wget2]
+
+# Supply the wget compatility package.
+# This could also be done by setting "rhel" to 11 but that can have other side-effects.
+[[components.wget2.overlays]]
```
reubeno: (Chatted offline; let's see if we can toggle via with/without. If that works, then this will be a slam dunk :smile:)

reubeno (approved):

PR #15740, opened 2026-02-05
firewalld: close mdns port in public (default) zone

ddstreet:
This closes the `mdns` port in the `public` zone, which is the default zone.

### DIFF
--- a/base/comps/firewalld/firewalld.comp.toml
+++ b/base/comps/firewalld/firewalld.comp.toml
@@ -1,1 +1,11 @@
 [components.firewalld]
+
+[[components.firewalld.overlays]]
+type = "spec-remove-tag"
+description = "Remove Fedora patch that opens mdns port in public (default) zone"
+tag = "Patch0"
+
+[[components.firewalld.overlays]]
+type = "file-remove"
+description = "Remove Fedora patch that opens mdns port in public (default) zone"
+file = "fedora-only-MDNS-default.patch"


base/comps/firewalld/0001-close-mdns-port-in-public-zone.patch
```diff
@@ -0,0 +1,11 @@
+diff -urpN a/config/zones/public.xml firewalld-2.4.0/config/zones/public.xml
```
reubeno: Remind me -- is the base Fedora version of this XML file a loose file in the dist-git repo for `firewalld`, or does it come upstream-upstream from the repo housing `firewalld`'s code base? (Or are they one and the same?)

If it's a loose file in the package (and not buried in a tarball), we have 2 choices:

* Apply a patch like this.
* Use an overlay to directly edit the file (e.g., textually remove the mdns entry).

If it's not, this is really the only option.

base/comps/firewalld/firewalld.comp.toml
```diff
@@ -0,0 +1,12 @@
+[components.firewalld]
+
+[[components.firewalld.overlays]]
+type = "spec-add-tag"
+description = "Close mdns port in public (i.e. default) zone"
```
reubeno: Something that I think is important is documenting the why/rationale for changes made via overlays or patches so that the current state of this repo is sufficient to understand the deviations made from upstream.

What do you think would be a good pattern for doing this? Do you think we should add a required `reason` field?

base/comps/firewalld/0001-close-mdns-port-in-public-zone.patch
ddstreet: ok clearly I didn't spend enough time actually looking at what I was doing here...as upstream firewalld actually *doesn't* include `mdns` in the public zone, it's added by fedora:
https://src.fedoraproject.org/rpms/firewalld/blob/rawhide/f/fedora-only-MDNS-default.patch

So if we *do* decide to remove it, we should simply remove fedora's patch, instead of adding our own.

It is good to see that fedora provided analysis/justification for including `mdns` in their patch; I'm inclined to follow their lead on this one, unless someone with more security expertise on our team thinks we should close the `mdns` port.

base/comps/firewalld/firewalld.comp.toml
ddstreet: I think the best place for a more detailed description of whatever change we are making is the actual commit message in git. I don't think putting it into the toml file is the best place for it.

ddstreet: the exception is an added patch file; i think all patch files we add should ideally be full `git format-patch` format (preferably actually generated by git) and should include the commit description (which should be a complete description/rationale in every patch)

reubeno: If it's an added patch, I 100% agree that the detailed description belongs in the patch itself, with a short hint in the TOML file.

Do you know if there are any scripts/tools out there that can validate if a patch is a well-formatted git format-patch style patch? That seems like a linter we'd want to enable -- or at least something we'd want AI code reviewing to "watch out for".

ddstreet: > Do you know if there are any scripts/tools out there that can validate if a patch is a well-formatted git format-patch style patch?

I am (yet again) only really familiar with the deb-world tooling for this, and over there patches are supposed to conform to [DEP-3](https://dep-team.pages.debian.net/deps/dep3/), and there is a [patchedit](https://manpages.debian.org/trixie/pkg-perl-tools/patchedit.1.en.html) tool that can check them (as well as some of the build tooling including built-in checking, IIRC).

ddstreet:
Ok, I updated this to remove the fedora patch, instead of adding our own patch. The git commit has a bit of reasoning on why.

base/comps/firewalld/firewalld.comp.toml
```diff
@@ -1 +1,11 @@
 [components.firewalld]
+
+[[components.firewalld.overlays]]
+type = "spec-remove-tag"
+description = "Remove Fedora patch that opens mdns port in public (default) zone"
+tag = "Patch0"
+
+[[components.firewalld.overlays]]
+type = "file-remove"
```
ddstreet, line 9: this is potentially a policy choice...when we "remove" existing upstream patches, should we simply remove their `PatchN:` line in the spec and leave the actual patch file? Or remove the patch file as well?

Also, this policy decision could be moved into azldev if a `spec-remove-patch` command type is added.

base/comps/firewalld/firewalld.comp.toml
reubeno: Good to know. I see https://docs.fedoraproject.org/en-US/packaging-guidelines/PatchUpstreamStatus/ but couldn't quickly find guidelines on patch formats, nor could I find clear evidence as to whether `rpmlint` inspects patches. (It does look some at how the patches are referenced in .spec files.) Possibly an opportunity for contribution?

base/comps/firewalld/firewalld.comp.toml
reubeno: Agreed on `spec-remove-patch` abstracting that. Would you mind filing an issue on the azldev repo to request that?

It could also be less brittle -- being able to, say, match a patch by filename pattern. That would be resilient against the patch being renamed to, say, `Patch1`.

reubeno (approved):
I think this is a reasonable way to disable the patch for now, with some clear follow-ups to extend `azldev` to simplify adding/removing patches.

PR #15741, opened 2026-02-06
feat: add components to cover new VHDX contents and remove basesystem…

liunan-ms:
This PR updates the `base/comps/components.toml` file to cover the new VHDX contents introduced by #15714 .

Changes:

* Added more components to match the package list for marketplace image.
* Added a comment instructing contributors to keep the list sorted alphabetically for readability in `components.toml`.
* Removed the `basesystem` package from the bootstrap list in `vm-base.kiwi`, as it's deprecated since fc43 and now provided by `filesystem` which's explicitly included in the kiwi definition.


base/comps/components.toml
```diff
@@ -242,26 +279,28 @@ includes = ["**/*.comp.toml"]
...
 [components.python-systemd]
 [components.python-urllib3]
 [components.python-wcwidth]
-[components.'python3.14']
 [components.pytz]
 [components.PyYAML]
 [components.quota]
 [components.rdma-core]
 [components.readline]
+[components.redhat-rpm-config]
```
reubeno: I had included it too, but I realize we shouldn't include this; `azurelinux-rpm-config` is our distro's version and supplants it.

liunan-ms: Removed it from the components.

reubeno (approved):

base/comps/components.toml
reubeno: Thanks! This was useful as it helped catch a bug in `azurelinux-rpm-config`; per our previous discussion, I'll look forward to a follow-up PR from you to resolve that.

PR #15747, opened 2026-02-06
Create toml overlay for kernel-headers

rlmenge:
Introduces the kernel-headers component for Azure Linux 4.0, built from the CBL-Mariner-Linux-Kernel 6.18.5.1 source.

Uses the Fedora kernel-headers.spec as a base with overlays to:

1. Switch from Fedora's pre-built headers tarball to generating headers from Azure Linux kernel source
2. Set version to 6.18.5.1 targeting CBL-Mariner rolling-lts/mariner-3
3. Add build dependencies (make, gcc, perl, rsync)
4. Generate both native x86_64 and cross-compiled arm64 headers

Produces `kernel-headers` and `kernel-cross-headers`subpackages.

### DIFF
--- a/base/comps/kernel-headers/kernel-headers.comp.toml
+++ b/base/comps/kernel-headers/kernel-headers.comp.toml
@@ -74,23 +74,23 @@ replacement = "%setup -q -n CBL-Mariner-Linux-Kernel-rolling-lts-mariner-%{azure
 [[components.kernel-headers.overlays]]
 type = "spec-append-lines"
 section = "%build"
-lines = [
-    "make mrproper",
-    "make headers_install INSTALL_HDR_PATH=headers-native",
-    "",
-    "# Determine native kernel arch and cross-compile for the other",
-    "native_karch=$(uname -m | sed 's/x86_64/x86/;s/aarch64/arm64/')",
-    "for cross_arch in arm64 x86; do",
-    "    [ \"$cross_arch\" = \"$native_karch\" ] && continue",
-    "    make ARCH=$cross_arch headers_install INSTALL_HDR_PATH=headers-$cross_arch",
-    "done",
-]
+lines = ["""
+make mrproper
+make headers_install INSTALL_HDR_PATH=headers-native
+
+# Determine native kernel arch and cross-compile for the other
+native_karch=$(uname -m | sed 's/x86_64/x86/;s/aarch64/arm64/')
+for cross_arch in arm64 x86; do
+    [ "$cross_arch" = "$native_karch" ] && continue
+    make ARCH=$cross_arch headers_install INSTALL_HDR_PATH=headers-$cross_arch
+done"""]
 
 # Install: copy headers_install output into RPM buildroot, replacing Fedora's copy logic.
+# The exit 0 at the end short-circuits the original Fedora install logic.
 [[components.kernel-headers.overlays]]
-type = "spec-search-replace"
-regex = "# List of architectures we support and want to copy their headers"
-replacement = """
+type = "spec-prepend-lines"
+section = "%install"
+lines = ["""
 native_karch=$(uname -m | sed 's/x86_64/x86/;s/aarch64/arm64/')
 
 mkdir -p $RPM_BUILD_ROOT%{_includedir}
@@ -103,7 +103,4 @@ for cross_arch in arm64 x86; do
     cp -rv headers-$cross_arch/include/* $cross_arch_includedir
 done
 
-exit 0
-
-# Original Fedora logic (skipped):
-# List of architectures we support and want to copy their headers"""
+exit 0"""]


base/comps/kernel-headers/kernel-headers.comp.toml
```diff
@@ -1 +1,93 @@
...
+# With type = "upstream", azldev fetches sources before applying overlays, so
+# the renamed tarball would not be available. Once azldev supports fetching
+# overlay-modified sources, this can switch to type = "upstream".
+# To get a copy of the source use
+# curl -sL -o kernel-6.18.5.1.tar.gz https://github.com/microsoft/CBL-Mariner-Linux-Kernel/archive/rolling-lts/mariner-3/6.18.5.1.tar.gz
+spec = { type = "local", path = "kernel-headers.spec" }
+
+# Build deps (Fedora's spec has none since it uses pre-built headers).
+[[components.kernel-headers.overlays]]
+type = "spec-search-replace"
```
reubeno, line 20: You should use the `spec-add-tag` overlay, e.g.:

```toml
[[components.mypackage.overlays]]
type = "spec-add-tag"
description = "Include a description here; it will become visible in some logging"
tag = "BuildRequires"
value = "some-devel-package"
```

There's an `overlays.md` file with examples like this, but it's not available on public GitHub yet. I can point you at it offline.

base/comps/kernel-headers/kernel-headers.comp.toml
```diff
@@ -1 +1,93 @@
...
+regex = "%global debug_package %{nil}"
+replacement = "%global debug_package %{nil}\n\nBuildRequires: make\nBuildRequires: gcc\nBuildRequires: perl\nBuildRequires: rsync"
+
+# Version updates: Fedora 6.19 -> Azure Linux 6.18.5.1
+[[components.kernel-headers.overlays]]
+type = "spec-search-replace"
+regex = "%global released_kernel 0"
+replacement = "%global released_kernel 0\n\n%define mariner_version 3"
```
reubeno: Should we use %{azurelinux_version} in place of %{mariner_version} while we're at it?

base/comps/kernel-headers/kernel-headers.comp.toml
```diff
@@ -1 +1,93 @@
...
+
+# Build deps (Fedora's spec has none since it uses pre-built headers).
+[[components.kernel-headers.overlays]]
+type = "spec-search-replace"
+regex = "%global debug_package %{nil}"
+replacement = "%global debug_package %{nil}\n\nBuildRequires: make\nBuildRequires: gcc\nBuildRequires: perl\nBuildRequires: rsync"
+
+# Version updates: Fedora 6.19 -> Azure Linux 6.18.5.1
+[[components.kernel-headers.overlays]]
+type = "spec-search-replace"
```
reubeno, line 25: There are a few options for defining a macro like this. I'd actually recommend something like the following at the top of the file:

```toml
[components.kernel-headers.build]
defines = { "kernel_azurelinux_version" = "3" }
```

That will generate a macro definition without mutating the spec.

And ditto for the other `%define`s below.

base/comps/kernel-headers/kernel-headers.comp.toml
```diff
@@ -1 +1,93 @@
...
+    "make headers",
+    "",
+    "# Determine native kernel arch and cross-compile for the other",
+    "native_karch=$(uname -m | sed 's/x86_64/x86/;s/aarch64/arm64/')",
+    "for cross_arch in arm64 x86; do",
+    "    [ \"$cross_arch\" = \"$native_karch\" ] && continue",
+    "    make ARCH=$cross_arch O=usr/include-$cross_arch headers",
+    "done",
+    "",
+    "find usr/include* \\( -name \".*\" -o -name \"Makefile\" \\) -delete",
```
reubeno: It's a toss-up, but you might consider using single-quoted TOML strings for the strings where there's content that would need escaping in double quotes. (And double-quoted strings like you've done above where there would need to be single-quoted content inside :smile: )

base/comps/kernel-headers/kernel-headers.comp.toml
```diff
@@ -1 +1,93 @@
...
+]
+
+# Install: copy built headers into RPM buildroot, replacing Fedora's copy logic.
+[[components.kernel-headers.overlays]]
+type = "spec-search-replace"
+regex = "# List of architectures we support and want to copy their headers"
```
reubeno: It looks like this is just at the beginning of the `%install` section; did I get that right? (The regex replacements run the risk of being brittle in case the text being searched for changes upstream so I consider them a less-preferred option.)

If that's the case, then a `spec-prepend-lines` overlay could work here? e.g.,

```toml
[[components.kernel-headers.overlays]]
type = "spec-prepend-lines"
section = "%install"
lines = [
    ...
]
```

If you'd prefer not to have the lines chopped up, I think you can use a `"""`-style string within a list. (That may be a readability/maintainability improvement for the other multi-line change above too, come to think of it.)

base/comps/kernel-headers/kernel-headers.comp.toml
```diff
@@ -1 +1,93 @@
...
+    "native_karch=$(uname -m | sed 's/x86_64/x86/;s/aarch64/arm64/')",
+    "for cross_arch in arm64 x86; do",
+    "    [ \"$cross_arch\" = \"$native_karch\" ] && continue",
+    "    make ARCH=$cross_arch O=usr/include-$cross_arch headers",
+    "done",
+    "",
+    "find usr/include* \\( -name \".*\" -o -name \"Makefile\" \\) -delete",
+]
+
+# Install: copy built headers into RPM buildroot, replacing Fedora's copy logic.
```
reubeno: Have we checked to see if the resulting RPM installs roughly the same files + dir structures as the upstream Fedora one, with these changes? Even if we take a different approach for composing it, I'd presume we would want to match the structure of the output.

reubeno (commented):
It looks like we're effectively gutting the upstream `kernel-headers.spec` (i.e., replacing all of %build and %install).

Given that, I could see an argument both ways for whether to overlay the upstream spec, or whether to just fully fork it and make it our own. I'm interested to get your take on it, but have left some comments throughout the .toml file regardless as general suggestions for the current approach.

base/comps/kernel-headers/kernel-headers.comp.toml
rlmenge: Updated the BuildRequires to use spec-add-tag

base/comps/kernel-headers/kernel-headers.comp.toml
rlmenge: Updated the define name

base/comps/kernel-headers/kernel-headers.comp.toml
rlmenge: Updated the defines method! I still had to use the search and replace for the defines already present in the spec.

base/comps/kernel-headers/kernel-headers.comp.toml
rlmenge: Using prepend for both the build and install sections

base/comps/kernel-headers/kernel-headers.comp.toml
rlmenge: Diffs look good! Here are the diffs:

For `kernel-headers`, found azurelinux includes `headers_check.pl` because we use `make headers` but that can be resolved using `make headers_install` which I have updated the spec to use (this is different from AzL3)

For `kernel-cross-headers` there are a lot of differences since we support only arm64 for cross-headers. Additionally it looks like Fedora includes some headers which are more relevant for x86
```
Architecture                Fedora      AZL
------------------------- -------- --------
arm-linux-gnu                 1082      0
arm64-linux-gnu               1082     1053
loongarch-linux-gnu           1082      0
powerpc-linux-gnu             1082      0
riscv-linux-gnu               1082      0
s390-linux-gnu                1082      0
x86-linux-gnu                 1082      0
=== arm64-linux-gnu: Only in Fedora (31 files) ===
/usr/arm64-linux-gnu/include/asm/amd_hsmp.h
/usr/arm64-linux-gnu/include/asm/a.out.h
/usr/arm64-linux-gnu/include/asm/boot.h
/usr/arm64-linux-gnu/include/asm/bootparam.h
/usr/arm64-linux-gnu/include/asm/debugreg.h
/usr/arm64-linux-gnu/include/asm/e820.h
/usr/arm64-linux-gnu/include/asm/elf.h
/usr/arm64-linux-gnu/include/asm/hw_breakpoint.h
/usr/arm64-linux-gnu/include/asm/hwcap2.h
/usr/arm64-linux-gnu/include/asm/ist.h
/usr/arm64-linux-gnu/include/asm/kvm_perf.h
/usr/arm64-linux-gnu/include/asm/ldt.h
/usr/arm64-linux-gnu/include/asm/mce.h
/usr/arm64-linux-gnu/include/asm/msr.h
/usr/arm64-linux-gnu/include/asm/mtrr.h
/usr/arm64-linux-gnu/include/asm/posix_types_32.h
/usr/arm64-linux-gnu/include/asm/posix_types_64.h
/usr/arm64-linux-gnu/include/asm/posix_types_x32.h
/usr/arm64-linux-gnu/include/asm/prctl.h
/usr/arm64-linux-gnu/include/asm/processor-flags.h
/usr/arm64-linux-gnu/include/asm/ptrace-abi.h
/usr/arm64-linux-gnu/include/asm/setup_data.h
/usr/arm64-linux-gnu/include/asm/sgx.h
/usr/arm64-linux-gnu/include/asm/sigcontext32.h
/usr/arm64-linux-gnu/include/asm/svm.h
/usr/arm64-linux-gnu/include/asm/unistd_32.h
/usr/arm64-linux-gnu/include/asm/unistd_x32.h
/usr/arm64-linux-gnu/include/asm/vm86.h
/usr/arm64-linux-gnu/include/asm/vmx.h
/usr/arm64-linux-gnu/include/asm/vsyscall.h
/usr/arm64-linux-gnu/include/linux/a.out.h

=== arm64-linux-gnu: Only in Azure Linux (2 files) ===
/usr/arm64-linux-gnu/include/asm/hwcap.h
/usr/arm64-linux-gnu/include/asm/sve_context.h

```

PR #15753, opened 2026-02-06
feat: Add script for adding components from image packages

liunan-ms:
This PR introduces a new utility script to automate the process of adding missing components from a kiwi image's `.packages` file into the `components.toml` configuration. This streamlines the workflow for ensuring all required components are tracked and properly ordered.
`scripts/add-missing-components-from-image.sh` detects missing components from a kiwi image's `.packages` file and inserts them into `base/comps/components.toml`, maintaining alphabetical order.

Usage: `./scripts/add-missing-components-from-image.sh [packages-file]`


scripts/add-missing-components-from-image.sh
```diff
@@ -0,0 +1,105 @@
...
+    exit 0
+fi
+
+echo "Missing components found:"
+echo "$MISSING_COMPONENTS"
+echo ""
+
+# Add each missing component to components.toml
+for comp in $MISSING_COMPONENTS; do
+    # Check if component already exists in the file
```
reubeno: This won't be correct, as the components may be defined in different files; that's why my original command line I'd shared uses `azldev comp list` with `-O json` and a `| jq ...` pipe to properly query. That would also obviate the need for the more complex parsing below.

scripts/add-missing-components-from-image.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+# Script to find missing components from a kiwi image's .packages file and add them to components.toml
```
reubeno: Thanks for sharing this!

ddstreet:
I think this script would probably be simpler if done in python instead of bash...

ddstreetmicrosoft:
can you rebase this into one commit with a commit message describing what/why/how?

PR #15795, opened 2026-02-11
ci(pr): Explicitly pass copilot token, add debug info for spec review workflow

dmcilvaney:
Attempt at fix for auth issues, with additional debug info in case it doesn't work.

### DIFF
--- a/.github/workflows/scripts/spec_review.sh
+++ b/.github/workflows/scripts/spec_review.sh
@@ -298,7 +298,6 @@ for GIT_SOURCE in "${GIT_SOURCES[@]}"; do
     if ! GIT_CURL_VERBOSE=1 GIT_TRACE=1 git clone --depth 1 --single-branch "$GIT_SOURCE" "$dest" 2>"$clone_log"; then
         echo "Error: Failed to clone $GIT_SOURCE" >&2
         cat "$clone_log" >&2
-        rm -f "$clone_log"
         exit 1
     fi
     rm -f "$clone_log"


.github/workflows/scripts/spec_review.sh
```diff
@@ -402,7 +416,7 @@ for run in {1..4}; do
             "${allow_url_args[@]}" \
             -p "$retry_prompt"
     else
-        copilot --agent "spec-review" \
+        GH_TOKEN="${COPILOT_TOKEN:-$GH_TOKEN}" copilot --agent "spec-review" \
```
reubeno, line 418: Just making sure -- copilot consumes from `GH_TOKEN`?

.github/workflows/scripts/spec_review.sh
```diff
@@ -291,8 +291,18 @@ trap 'rm -rf "$temp_sources_dir"' EXIT SIGHUP SIGINT SIGTERM
...
-    git clone --depth 1 --single-branch "$GIT_SOURCE" "$temp_sources_dir/$(basename "$GIT_SOURCE" .git)"
-    git_folders+=("$temp_sources_dir/$(basename "$GIT_SOURCE" .git)")
+    dest="$temp_sources_dir/$(basename "$GIT_SOURCE" .git)"
+    echo "Cloning $GIT_SOURCE -> $dest"
+    # Capture verbose output but only print on failure to avoid log noise
+    clone_log="$(mktemp)"
+    if ! GIT_CURL_VERBOSE=1 GIT_TRACE=1 git clone --depth 1 --single-branch "$GIT_SOURCE" "$dest" 2>"$clone_log"; then
+        echo "Error: Failed to clone $GIT_SOURCE" >&2
+        cat "$clone_log" >&2
+        rm -f "$clone_log"
```
reubeno: Why remove?

.github/workflows/scripts/spec_review.sh
dmcilvaney: https://docs.github.com/en/copilot/how-tos/copilot-cli/install-copilot-cli#authenticating-with-copilot-cli

.github/workflows/scripts/spec_review.sh
dmcilvaney:
```suggestion
```

PR #15907, opened 2026-02-18
fix(beakerlib): disable conflicting python-platform.patch overlay

dmcilvaney:
The upstream Fedora spec for beakerlib applies python3.patch (Patch2) when %fedora is defined and python-platform.patch (Patch3) when %rhel > 7, expecting these conditions to be mutually exclusive. Azure Linux defines both %fedora=43 (from the Fedora 43 build chroot) and %rhel=10 (from distro build defines), causing both patches to apply. Since both modify the same Python shebang lines (#!/usr/bin/env python), Patch3 fails because Patch2 has already changed them.

Add a spec-search-replace overlay to comment out %patch -P 3 in %prep. Azure Linux uses python3 (Fedora-style), not /usr/libexec/platform-python (RHEL-specific), so Patch2 alone provides the correct behavior.

Move beakerlib from inline definition in components-full.toml to a dedicated beakerlib/beakerlib.comp.toml to hold the overlay.

### DIFF
--- a/base/comps/beakerlib/beakerlib.comp.toml
+++ b/base/comps/beakerlib/beakerlib.comp.toml
@@ -1,0 +1,12 @@
+[components.beakerlib]
+
+# Azure Linux defines both %fedora (from the Fedora 43 chroot) and %rhel=10
+# (from distro build defines). The upstream spec applies python3.patch on Fedora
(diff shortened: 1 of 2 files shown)


christopherco (approved):
Approving this fix. We're still sorting out how best to handle `%fedora` and `%rhel` macros being defined. Ideally, we should only define one.

PR #15908, opened 2026-02-19
fix(fcode-utils): fix GCC 15/C23 build failure with -std=gnu17 overlay

dmcilvaney:
fcode-utils 1.0.3 fails to build with GCC 15's default C23 mode due to:
  1. typedef to 'bool' (a C23 keyword) in shared/types.h
  2. K&R-style empty-parameter function pointers treated as zero-arg in C23
  3. -Wincompatible-pointer-types promoted to error

The upstream project is dormant (last release 2014). A partial fix exists as an unmerged PR: https://github.com/openbios/fcode-utils/pull/32

Fedora is also affected — fcode-utils has failed all mass rebuilds since fc42 (fc42-4, fc43-5, fc44-6 all failed in Fedora Koji). The last successful Fedora build was 1.0.3-3.fc42, before the C23 transition.

Add a spec-search-replace overlay to force -std=gnu17 in CFLAGS, restoring pre-C23 semantics. Move the component from an inline entry in components-full.toml to a dedicated comp.toml to hold the overlay.

Validated: prep-sources clean, build succeeds, smoke-tested toke/detok/ romheaders in mock chroot.

### DIFF
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -534,7 +534,6 @@
 [components.fakeroot]
 [components.fast_float]
 [components.fcgi]
-[components.fcode-utils]
 [components.fdk-aac-free]
 [components.fdupes]
 [components.fedora-iot-config]
--- a/base/comps/fcode-utils/fcode-utils.comp.toml
+++ b/base/comps/fcode-utils/fcode-utils.comp.toml
@@ -1,0 +1,16 @@
+[components.fcode-utils]
+
+# Upstream project is dormant (last release v1.0.3 in 2014). The codebase uses
+# K&R-style function pointer declarations (e.g. void (*funct)()) and redefines
+# bool via typedef, which are incompatible with C23 defaults in GCC 15:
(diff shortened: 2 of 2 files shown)


ddstreetmicrosoft:
> The only thing that requires this seems to be `openbios`, and the only thing that requires `openbios` seems to be `qemu-system` for PPC and Sparc...can we look at dropping `fcode-utils` and `openbios` instread of merging this?

@trungams can you double-check that i got the package deps right ^

trungams:
> > The only thing that requires this seems to be `openbios`, and the only thing that requires `openbios` seems to be `qemu-system` for PPC and Sparc...can we look at dropping `fcode-utils` and `openbios` instread of merging this?
> 
> @trungams can you double-check that i got the package deps right ^

Yep. That's what I got as well. Looks like we shouldn't need this package

christopherco:
Thanks @ddstreetmicrosoft and @trungams ! I'm closing this PR and we'll handle it at the package list/dependency level

PR #15910, opened 2026-02-19
fix(apache-sshd): work around maven incompatibility and add maven complier release…

liunan-ms:
This pull request updates the configuration for the `apache-sshd` component to address build issues and compatibility problems. 

1. Dependency chain:
apache-sshd → maven-remote-resources-plugin → mvn(org.apache.maven.shared:maven-filtering) → mvn(org.codehaus.plexus:plexus-xml)  → mvn(org.apache.maven:maven-xml:4.0.0-rc-4)   (Maven 4 API jar)
`plexus-xml` was compiled with a bridge `Xpp3Dom(XmlNode)` constructor that references Maven 4's `XmlNode` type. At runtime, Fedora 43 has `maven 3.9.11` installed, the old plexus-utils Xpp3Dom wins the classloading (SelfFirstStrategy) and doesn't have that constructor → NoSuchMethodError. So the issue is specifically Fedora's `maven-filtering` → `plexus-xml` dependency introducing Maven 4 types into a Maven 3 runtime.

2. Another issue is `xmvn-mojo:javadoc` reads `<maven.compiler.release>` (resolves to null because the POM never defines it) instead of compiler plugin configuration where correct `<release>${java.sdk.version}</release>` is specified. So `xmvn` falls back to pass `-source 1.8` to javadoc, which JDK 21 rejects.

Component configuration updates:

* Added a new `[components.apache-sshd]` section to `apache-sshd.comp.toml` with overlays to:
  * Remove the `maven-remote-resources-plugin` to work around an API incompatibility between plugin v3.1.0 and plexus-xml/maven4 jars in Fedora 43 chroot.
  * Insert a `<maven.compiler.release>8</maven.compiler.release>` line into `pom.xml` to ensure correct Javadoc generation with JDK 21, addressing issues with the Maven plugin configuration.

Component list maintenance:

* Removed the `[components.apache-sshd]` section from `components-full.toml`, likely to avoid duplication or to reflect the new configuration structure.

### DIFF
--- a/base/comps/apache-sshd/apache-sshd.comp.toml
+++ b/base/comps/apache-sshd/apache-sshd.comp.toml
@@ -1,10 +1,25 @@
 [components.apache-sshd]
-overlays = [
-    # Remove maven-remote-resources-plugin to work around API incompatibility between
-    # plugin v3.1.0 and plexus-xml/maven4 jars in Fedora 43 chroot (NoSuchMethodError on Xpp3Dom).
-    { type = "spec-append-lines", section = "%prep", lines = ["%pom_remove_plugin :maven-remote-resources-plugin"] },
 
-    # POM defines <release> only in maven-compiler-plugin config, not as a property.
-    # xmvn-mojo:javadoc reads ${maven.compiler.release} (null) and falls back to ${maven.compiler.source} = 1.8, which JDK 21 javadoc rejects.
-    { type = "spec-append-lines", section = "%prep", lines = ["sed -i '/<maven.compiler.target>/a\\    <maven.compiler.release>8</maven.compiler.release>' pom.xml"] },
-]
+[[components.apache-sshd.overlays]]
+description = "Remove maven-remote-resources-plugin (Xpp3Dom classloading clash with plexus-xml/maven4)"
+type = "spec-append-lines"
+section = "%prep"
+lines = ["%pom_remove_plugin :maven-remote-resources-plugin"]
+
+[[components.apache-sshd.overlays]]
+description = "Add patch for maven.compiler.release POM property (xmvn javadoc fix)"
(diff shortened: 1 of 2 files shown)


base/comps/apache-sshd/apache-sshd.comp.toml
```diff
@@ -0,0 +1,10 @@
+[components.apache-sshd]
+overlays = [
+    # Remove maven-remote-resources-plugin to work around API incompatibility between
+    # plugin v3.1.0 and plexus-xml/maven4 jars in Fedora 43 chroot (NoSuchMethodError on Xpp3Dom).
+    { type = "spec-append-lines", section = "%prep", lines = ["%pom_remove_plugin :maven-remote-resources-plugin"] },
+
+    # POM defines <release> only in maven-compiler-plugin config, not as a property.
+    # xmvn-mojo:javadoc reads ${maven.compiler.release} (null) and falls back to ${maven.compiler.source} = 1.8, which JDK 21 javadoc rejects.
+    { type = "spec-append-lines", section = "%prep", lines = ["sed -i '/<maven.compiler.target>/a\\    <maven.compiler.release>8</maven.compiler.release>' pom.xml"] },
```
ddstreetmicrosoft: any changes to the actual package files should be done with a patch, not with a sed action in the spec file

base/comps/apache-sshd/apache-sshd.comp.toml
```diff
@@ -0,0 +1,10 @@
+[components.apache-sshd]
+overlays = [
+    # Remove maven-remote-resources-plugin to work around API incompatibility between
```
ddstreetmicrosoft: please change the comments into text in a description element

PR #15911, opened 2026-02-19
fix(glade): add missing BuildRequires for gobject-introspection-devel

dmcilvaney:
The glade build fails during meson setup because gobject-introspection-1.0 pkg-config is not found:

  gladeui/meson.build:242:8: ERROR: Dependency "gobject-introspection-1.0"
  not found, tried pkgconfig

The Fedora f43 spec uses %meson which passes --auto-features=enabled, making GObject introspection support mandatory, but does not list gobject-introspection-devel as a BuildRequires.

This was fixed in Fedora rawhide by Yaakov Selkowitz: https://src.fedoraproject.org/rpms/glade/c/478c15c

Add a spec-add-tag overlay to backport the fix until it lands in f43.

### DIFF
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -799,7 +799,6 @@
 [components.git]
 [components.gjs]
 [components.gl2ps]
-[components.glade]
 [components.glew]
 [components.glib-networking]
 [components.'glibmm2.4']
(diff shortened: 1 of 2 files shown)


christopherco:
Approving this to handle build break. Separately, we will identify the dependency chain for `glade` and ensure we really need to be building it.

PR #15912, opened 2026-02-19
feat(kernel): Remove debug kernel

rlmenge:
Disable the debug kernel build by adding --without debug to the kernel component build config. As it is unnecessary overhead.

The Fedora kernel spec builds both base and debug variants by default. With debugbuildsenabled=1 (the default), this means compiling the entire kernel twice. Passing --without debug skips the debug variant, roughly halving wall-clock build time.

Unmodified taskID=94143 built x86 kernel in 2:59:23 total time
This branch on taskID=100493 built x86 kernel in 1:00:01 total time

### DIFF
--- a/base/comps/kernel/kernel.comp.toml
+++ b/base/comps/kernel/kernel.comp.toml
@@ -2,6 +2,7 @@
 
 # Set rhel to 0 to avoid deps on non-existent cert packages
 # (rhel=10 would engage RHEL-specific BuildRequires)
+# Add without = ["debug"] to skip debug kernel variant and reduce build time
 [components.kernel.build]
 defines = { rhel = "0" }
 without = ["debug"]


ddstreetmicrosoft:
> Disable the debug kernel build by adding --without debug to the kernel component build config. As it is unnecessary overhead.
> 
> The Fedora kernel spec builds both base and debug variants by default. With debugbuildsenabled=1 (the default), this means compiling the entire kernel twice. Passing --without debug skips the debug variant, roughly halving wall-clock build time.
> 
> Unmodified taskID=94143 built x86 kernel in 2:59:23 total time This branch on taskID=100493 built x86 kernel in 1:00:01 total time

can you put this description comment into the actual commit message please?

base/comps/kernel/kernel.comp.toml
```diff
@@ -2,6 +2,7 @@
 
 # Set rhel to 0 to avoid deps on non-existent cert packages
 # (rhel=10 would engage RHEL-specific BuildRequires)
+# Add without = ["debug"] to skip debug kernel variant and reduce build time
```
ddstreetmicrosoft, line 5: can you change this comment into text in a description element (like copilot is asking for)?

ddstreetmicrosoft (commented):
one more minor change please

base/comps/kernel/kernel.comp.toml
rlmenge: "While the 'without' field has a schema description, the codebase convention is to add comments explaining the specific rationale for build" 

I interpretted this as it asking for a comment. But I can use the description schema if preferred. Will check and update.

ddstreetmicrosoft: hmm yeah i guess i don't really know what our actual policy/process is here...

ddstreetmicrosoft: ok let's leave it as a comment isntead of description.

PR #15934, opened 2026-02-19
fix(gnutls): disable bundled_gmp to fix SRPM build failure

dmcilvaney:
The upstream Fedora gnutls spec conditionally enables bundled_gmp when rhel >= 9 and fips is enabled:

    %if 0%{?rhel} >= 9 && %{with fips}
    %bcond_without bundled_gmp

Azure Linux defines rhel=10 and fips is on by default, which activates this path. However, the bundled source (gmp-6.2.1.tar.xz) only exists in the RHEL lookaside cache, not Fedora's, causing rpmbuild to fail:

    error: Bad file: /builddir/build/SOURCES/gmp-6.2.1.tar.xz: No such file or directory

Fix by adding build.without = ["bundled_gmp"] so gnutls links against the system gmp package instead. This matches Fedora's own behavior. The gnutls-fips sub-package correctly pins the system gmp version at build time via the %{fips_requires} macro.

Moved gnutls from inline components.toml to a dedicated comp.toml since it now has build configuration.

Tested: prep-sources succeeds, comp build produces all expected RPMs (gnutls, gnutls-fips, gnutls-utils, gnutls-devel, etc.), and RPM inspection confirms Requires: libgmp.so.10 with no bundled(gmp).


christopherco:
Superseded by #15940

PR #15935, opened 2026-02-20
fix: Python3.14 Build Requirement Rename

Tonisal-byte:
Our definition of `RHEL` has caused the python3.14 package to rely on some versioned python dependencies only found in special RHEL repos. To build with the non-versioned dependencies we are simply renaming the prefix impacted by RHEL.


anphel31 (approved):

base/comps/python3.14/python3.14.comp.toml
```diff
@@ -0,0 +1,10 @@
+[components.'python3.14']
+
+# Override the wheel package prefix to use the Fedora-style unversioned name.
+# Azure Linux sets %rhel, which causes python-rpm-macros to expand
+# %python_wheel_pkg_prefix to "python3" (RHEL-style), requiring packages like
+# python3-pip-wheel / python3-setuptools-wheel. Fedora repos only provide the
+# unversioned python-pip-wheel / python-setuptools-wheel. This define forces
+# the Fedora convention so wheel dependencies resolve correctly.
```
ddstreetmicrosoft, line 8: This sounds like a bootstrap issue, and I don't think it's a good idea for us to mix-and-match between fedora and rhel; if we're going to inherit from rhel, we should stick with its conventions; if we're going to inherit from fedora, go that way.
To me it seems like the correct thing to do is bootstrap a build that produces `python3-pip-wheel` and then the package should build normally without any overlay needed.
It does not seem like changing to fedora style naming, instead of rhel style naming, is a good idea.

ddstreetmicrosoft (changes requested):
I don't think this approach is correct.
cc @reubeno @christopherco

christopherco:
Superseded with #15940 to undefine %rhel. Closing this PR.

PR #15940, opened 2026-02-20
chore: do not define rhel, making us pure fedora by default

tobiasb-ms:
Removed the line in `azurelinux.distro.toml` that sets `rhel` to 10.
Removed existing workaround for the fact that we defined both `fedora` and `rhel`.

### DIFF
--- a/base/comps/kernel/kernel.comp.toml
+++ b/base/comps/kernel/kernel.comp.toml
@@ -1,8 +1,5 @@
 [components.kernel]
 
-# Set rhel to 0 to avoid deps on non-existent cert packages
-# (rhel=10 would engage RHEL-specific BuildRequires)
 # Add without = ["debug"] to skip debug kernel variant and reduce build time
 [components.kernel.build]
-# defines = { rhel = "0" }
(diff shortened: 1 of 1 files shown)


base/comps/kernel/kernel.comp.toml
```diff
@@ -4,5 +4,5 @@
 # (rhel=10 would engage RHEL-specific BuildRequires)
 # Add without = ["debug"] to skip debug kernel variant and reduce build time
 [components.kernel.build]
-defines = { rhel = "0" }
+# defines = { rhel = "0" }
```
christopherco: question: Is there a reason to keep this commented rather than removing the line altogether?

tobiasb-ms: I had meant to remove that. Don't know how I missed it.

tobiasb-ms: Fixed in https://github.com/microsoft/azurelinux/pull/15940/changes/367a242ee024454a483cd498b83f3dc3d066bcb5.

PR #15943, opened 2026-02-20
fix(golang-github-clbanning-mxj): enable %check by removing broken example tests

dmcilvaney:
The upstream example_test.go at v2.5.5 references v1 API identifiers (HandleXmlReader, HandleJsonReader, Map, etc.) that were removed in v2. This causes Go test compilation failures:

  ./example_test.go:19:1: ExampleHandleXmlReader refers to unknown identifier: HandleXmlReader
  ./example_test.go:151:1: ExampleMap_Struct refers to unknown identifier: Map

Fedora's spec (f43, rawhide) ships v2.5.5 with %check enabled but doesn't exclude the broken tests. The issue is fixed upstream in v2.7.0 where the example functions were updated to use v2 imports and commented out: https://github.com/clbanning/mxj/commit/458e5dc01578622621dd1997cf19edbaa7972b00

Rather than skipping %check entirely, add an overlay to remove only the broken example_test.go during %prep, allowing the real unit tests to run.

Tested: Built successfully with %check passing, verified package installs and Go source files are present in /usr/share/gocode/src/github.com/clbanning/mxj/

### DIFF
--- a/base/comps/golang-github-clbanning-mxj/golang-github-clbanning-mxj.comp.toml
+++ b/base/comps/golang-github-clbanning-mxj/golang-github-clbanning-mxj.comp.toml
@@ -1,4 +1,12 @@
 [components.golang-github-clbanning-mxj]
 
-[components.golang-github-clbanning-mxj.build]
-check = { skip = true, skip_reason = "Disabling checks for initial set of failures." }
+# Upstream example_test.go references v1 API identifiers (HandleXmlReader, Map, etc.)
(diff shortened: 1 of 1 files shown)


dmcilvaney:
Superseded by https://github.com/microsoft/azurelinux/commit/df13ad057#diff-90caf72995c8612b1429aa399582f728158b9c53334db43feaa1f822e917bbf8L1 which removes this package completely.

PR #15953, opened 2026-02-23
Add scripts for building, publishing images and deploying azure VM

liunan-ms:
This PR includes a couple of dev scripts. One can change the variable names in `common.sh` and use `build-vm-images.sh` to build VM images locally and publish the image to SIG via `publish-sig-image.sh` and deploy an Azure VM with `deploy-azure-vm.sh` for testing purposes.

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -24,7 +24,7 @@ ARCH="<test-vm-architecture>" # e.g., "x86_64" or "aarch64"
 if [ "$ARCH" = "aarch64" ]; then
     export TEST_VM_SIZE="Standard_D4ps_v6"
 else
-    export TEST_VM_SIZE="Standard_D4s_v7"
+    export TEST_VM_SIZE="Standard_D4s_v5"
 fi
 
 # Set local image path


scripts/common.sh
```diff
@@ -0,0 +1,23 @@
...
+export SSH_PUBLIC_KEY_PATH="/home/liunan/.ssh/id_rsa.pub"
+export TEST_VM_SIZE="Standard_D2_v4"
+
+# Set local image path
+export IMAGE_PATH="./base/out/images/vm-base/azl4-vm-base.x86_64-0.1.vhd"
+
+# Set gallery configuration
+export GALLERY_NAME="basevmimagegallery"
+export GALLERY_IMAGE_DEFINITION="basevmimage"
+export VM_NAME="azl4-vm-1"
```
binujp: comment(non-blocking): It would be good to tag the default name with $USER. Even better would be to add user and a date string or some such

scripts/azure-gallery-image-base.bicep
```diff
@@ -0,0 +1,42 @@
...
+          uri: sourceDiskUrl
+        }
+      }
+    }
+    safetyProfile: {
+      allowDeletionOfReplicatedLocations: allowDeletionOfReplicatedLocations
+    }
+  }
+  tags: {}
+}
```
christopherco: issue: please update your text editor configurations to end files with newlines. It's not an issue with this file in particular, but can be an issue if newlines are missed at the end of unix config files

scripts/build-vm-images.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -euxo pipefail
+TARGET_DIR="/var/tmp/azl-vm-images"
```
christopherco: issue: This creates a target directory in the user's /var/tmp namespace, which is not guaranteed to be safe to manipulate.

Instead, consider reworking to utilize `mktemp -d` invocation so you get a unique temp directory. This command also has the possible benefit, depending on how the system is configured, to create the directory in a tmpfs, which is backed by RAM for faster I/O transactions than a directory backed by a storage disk media

scripts/build-vm-images.sh
```diff
@@ -0,0 +1,29 @@
...
+# Build vm-base image using azldev
+azldev image build vm-base --local-repo ./base/out --remote-repo "$REMOTE_KOJI_REPO_URL"
+
+# Convert VHDX to VPC fixed size
+qemu-img convert -f vhdx -O vpc -o subformat=fixed,force_size \
+    ./base/out/images/vm-base/azl4-vm-base.x86_64-0.1.vhdx \
+    ./base/out/images/vm-base/azl4-vm-base.x86_64-0.1.vhd
+
+# Copy the resulting VHD to Windows Downloads folder
+cp ./base/out/images/vm-base/azl4-vm-base.x86_64-0.1.vhd "$HOME/Downloads/azl4-vm-base.x86_64-0.1.vhd"
```
christopherco: issue: We don't own the user's $HOME/Downloads directory so we shouldn't copy files into this path

I suggest to remove this line

scripts/deploy-azure-vm.sh
```diff
@@ -0,0 +1,17 @@
...
+set -euxo pipefail
+# Find the absolute path of the directory containing this script
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+. "$SCRIPTS_DIR/common.sh"
+
+az vm create \
+        --resource-group "$RESOURCE_GROUP_NAME" \
+        --name "$VM_NAME" \
+        --size "$TEST_VM_SIZE" \
+        --os-disk-size-gb 60 \
```
christopherco: question: do we need to set this os disk size to 60GB? Can we successfully deploy with the default size?

scripts/publish-sig-image.sh
```diff
@@ -0,0 +1,98 @@
...
+# https://basevmimage.blob.core.windows.net/vhds/azl4-vm-base.x86_64-0.1.vhd
+# Find the absolute path of the directory containing this script
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+. "$SCRIPTS_DIR/common.sh"
+
+storage_account_url="https://$STORAGE_ACCOUNT_NAME.blob.core.windows.net"
+storage_account_resource_id="/subscriptions/$SUBSCRIPTION_ID/resourceGroups/$RESOURCE_GROUP_NAME/providers/Microsoft.Storage/storageAccounts/$STORAGE_ACCOUNT_NAME"
+
+replicationMode="Shallow"
+image_version=$VERSION
```
christopherco: suggestion (non-blocking): Consider a future change to automatically update the image_version value so users can keep uploading new images to the same SIG, but with different increasing version numbers.

scripts/publish-sig-image.sh
```diff
@@ -0,0 +1,98 @@
...
+containerExists=$(az storage container exists --account-name "$STORAGE_ACCOUNT_NAME" --name "$STORAGE_CONTAINER_NAME" --auth-mode login | jq .exists)
+if [[ $containerExists != "true" ]]; then
+    echo "Could not find container \"$STORAGE_CONTAINER_NAME\". Creating container \"$STORAGE_CONTAINER_NAME\" in storage account \"$STORAGE_ACCOUNT_NAME\"..."
+    az storage container create \
+        --account-name "$STORAGE_ACCOUNT_NAME" \
+        --name "$STORAGE_CONTAINER_NAME" \
+        --auth-mode login
+fi
+
+# Upload the image artifact to Steamboat Storage Account
```
christopherco: chore: remove references to steamboat

scripts/build-vm-images.sh
liunan-ms: I removed the TARGET_DIR which was used for hosting the vhd outside my WSL for experiments.

scripts/build-vm-images.sh
```diff
@@ -0,0 +1,16 @@
...
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+. "$SCRIPTS_DIR/common.sh"
+
+# Build vm-base image using azldev
+azldev image build vm-base --local-repo ./base/out --remote-repo "$REMOTE_KOJI_REPO_URL"
+
+# Convert VHDX to VPC fixed size
+qemu-img convert -f vhdx -O vpc -o subformat=fixed,force_size \
+    ./base/out/images/vm-base/azl4-vm-base.x86_64-0.1.vhdx \
+    ./base/out/images/vm-base/azl4-vm-base.x86_64-0.1.vhd
```
binujp: Suggestion: I have the same observation as copilot. An alternative is to find git root `git rev-parse --show-toplevel` and base paths from that. It would also help to build out common prefix paths, that is a coding pattern.

scripts/common.sh
```diff
@@ -0,0 +1,68 @@
...
+export SUBSCRIPTION_ID="<your-subscription-id>"
+
+# Set resource configuration
+export RESOURCE_GROUP_NAME="<your-resource-group-name>"
+export LOCATION="westus3"
+export STORAGE_ACCOUNT_NAME="<your-storage-account-name>"
+export STORAGE_CONTAINER_NAME="<your-storage-container-name>"
+export PUBLISHER="<your-publisher-name>"
+export OFFER="<your-offer-name>"
+export storage_blob_name="azl4-vm-base.x86_64-$(date +%Y%m%d-%H%M%S).vhd"
```
binujp: suggestion (refactor): The time string tag is used elsewhere too. You could make this available via a variable and use it for all artifacts that need a unique-id. VM name, or dir-name and such.

scripts/common.sh
```diff
@@ -0,0 +1,68 @@
...
+    az sig image-version list \
+        --resource-group "$RESOURCE_GROUP_NAME" \
+        --gallery-name "$GALLERY_NAME" \
+        --gallery-image-name "$GALLERY_IMAGE_DEFINITION" --query '[].name' -o tsv |
+        sort -t "." -k1,1n -k2,2n -k3,3n |
+        tail -1
+}
+
+function get-image-version() {
+    local OP="${1:-}"
```
binujp: nit (non-blocking): A getter should not have side effects even at the cost of more code. I would strongly recommend splitting this into get-image-version() and a separate increment-version() function. You already have all of the required code in hand.

scripts/deploy-azure-vm.sh
binujp, line 1: nit (design, non-blocking, opinion): This script can be stand alone and maybe is meant to be used that way. It does not take any command line arguments but expects environment variables to be used. While that pattern works well there is a locality issue for arguments and context. i.e. any of the environment variables can be set at any moment prior to the invocation. That applies to interactive use and use in a script.

This kind of usage is very common in pipeline scripts I have seen. Unless there is a precedent or coding guideline which recommends this pattern, I would rather use explicit command line arguments for all passed-in parameters.

scripts/publish-sig-image.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
```
binujp, line 1: Here and other scripts, you may want '-e' even though most commands seems to be in a conditional. Safer to do that.

scripts/publish-sig-image.sh
```diff
@@ -0,0 +1,98 @@
...
+REGIONS_JSON=$(echo "$LOCATION" | awk -F, '{
+  printf "[";
+  for(i=1;i<=NF;i++) {
+    printf "\"%s\"", $i;
+    if(i<NF) printf ",";
+  }
+  printf "]";
+}')
+# Create Image Version from storage account blob
+az deployment group create \
```
christopherco, line 87: issue: this deployment is failing due to "The source blob is not a page blob" error

christopherco (changes requested):

scripts/publish-sig-image.sh
christopherco: I was able to fix it by updating the `azcopy` command to add the --blob-type=PageBlob option

```
azcopy copy "$IMAGE_PATH" "$storage_blob_endpoint" --blob-type=PageBlob
```

scripts/common.sh
```diff
@@ -0,0 +1,57 @@
...
+export STORAGE_ACCOUNT_NAME="<your-storage-account-name>"
+export STORAGE_CONTAINER_NAME="<your-storage-container-name>"
+export PUBLISHER="<your-publisher-name>"
+export OFFER="<your-offer-name>"
+export TIME_TAG="$(date +%Y%m%d-%H%M%S)"
+export STORAGE_BLOB_NAME="azl4-vm-base.x86_64-${TIME_TAG}.vhdfixed"
+export VM_NAME="${USER}-azl-vm-${TIME_TAG}"
+export SSH_USER="<your-ssh-username>"
+export SSH_PUBLIC_KEY_PATH="<path-to-your-ssh-public-key>"
+export TEST_VM_SIZE="<your-test-vm-size>"
```
trungams: suggestion(non-blocking): can this variable take some default value, personally I often have a hard time remembering available SKUs for different image types, and will have to search up docs/notes to figure it out

christopherco: +1 on this suggestion. Maybe we can use "Standard_D4s_v7" as a default x86_64 VM Size and "Standard_D4ps_v6" for ARM64 VM size

PR #15961, opened 2026-02-24
feat(kernel): Use CBL-Mariner-Linux-Kernel Source

rlmenge:
Adds a custom Azure Linux kernel based on the Fedora 43 upstream kernel spec, using CBL-Mariner-Linux-Kernel 6.18.5.1 sources. This replaces the previous minimal kernel stub with a fully overlayed component that builds kernel 6.18.5.1 with Azure Linux-specific configs.

Changes
Initial kernel overlay (a8b9e29db)

Replaced the minimal kernel.comp.toml stub with a full overlay-based component targeting the Fedora 43 upstream kernel spec
Added CBL-Mariner-Linux-Kernel 6.18.5.1 tarball as an external source with SHA512 verification
Added Azure Linux kernel configs for x86_64 (8,078 lines) and aarch64 (12,555 lines)

Overlays include:
1. specrpmversion / specversion / tarfile_release / pkgrelease / specrelease set to 6.18.5.1
2. Source0 repointed to the CBL-Mariner tarball with extraction path fix
3. Azure Linux config files (not yet used)
4. kextraversion = ".1" prepended to EXTRAVERSION so uname -r = 6.18.5.1-1.azl4.x86_64 (matches module path)
5. Redhat patch declaration and application disabled

Testing
- Built successfully with azldev comp build -p kernel — produces 13 RPMs including kernel-6.18.5.1-1.azl4.x86_64.rpm
- Built VM image with azldev image build vm-base --local-repo ./base/out
- Booted in QEMU (vm-base) — uname -r = 6.18.5.1-1.azl4.x86_64 confirmed

```
[test@azurelinux-vm ~]$ uname -a
Linux azurelinux-vm 6.18.5.1-1.azl4.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Feb 24 18:44:34 UTC 2026 x86_64 GNU/Linux
[test@azurelinux-vm ~]$ cat /etc/os-release
NAME="Azure Linux"
VERSION="4.0 (Evergreen Prerelease)"
RELEASE_TYPE=development
...
[test@azurelinux-vm ~]$ 
```

- Deployed to Azure VM (Standard_D4s_v3, westus3) — uname -r = 6.18.5.1-1.azl4.x86_64 confirmed via SSH
- built in koji taskID=241090


base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
...
+
+[components."kernel".build]
+# debugbuildsenabled: set to "0" to skip debug kernel variant (faster builds)
+# azl_pkgrelease: RPM release number for the Azure Linux kernel package
+# kextraversion: 4th version component appended to EXTRAVERSION so uname -r
+# matches source version (6.18.5.1).
+defines = { debugbuildsenabled = "0", azl_pkgrelease = "1", kextraversion = ".1" }
+without = ["debug", "configchecks"]
+
+[[components."kernel".source-files]]
```
rlmenge: This will change to use our copy of the source rather than the GitHub hosted.

base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
...
+[[components.kernel.overlays]]
+description = "Declare Azure Linux aarch64 config source"
+type = "spec-add-tag"
+tag = "Source10001"
+value = "6.18-aarch64-azl.config"
+
+[[components.kernel.overlays]]
+description = "Install Azure Linux configs and extra sources during %prep"
+type = "spec-search-replace"
+regex = "find . -name .gitignore -delete >/dev/null"
```
rlmenge: Last line in the kernel spec. Considering appending to the %prep section instead

base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
...
+# Standard Azure Linux kernel
+
+[components."kernel"]
+spec = { type = "upstream", upstream-name = "kernel" }
+
+[components."kernel".build]
+# debugbuildsenabled: set to "0" to skip debug kernel variant (faster builds)
+# azl_pkgrelease: RPM release number for the Azure Linux kernel package
+# kextraversion: 4th version component appended to EXTRAVERSION so uname -r
+# matches source version (6.18.5.1).
```
rlmenge: Lost comment for the without. Updating

base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
-[components.kernel]
+###############################################################################
+# AUTO-GENERATED — do not edit manually.
+# Source: kernel-matrix.toml    variant: kernel
+# Regenerate: python3 scripts/generate-kernels.py
```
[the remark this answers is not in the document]
rlmenge: A remenant of an approach to generate multiple tomls from a single "template" toml. Will update

base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
...
+description = "Disable redhat patch declaration"
+type = "spec-search-replace"
+regex = "Patch1: patch-%{patchversion}-redhat.patch"
+replacement = "# Patch1: patch-%{patchversion}-redhat.patch (disabled for AzureLinux)"
+
+[[components.kernel.overlays]]
+description = "Disable redhat patch application"
+type = "spec-search-replace"
+regex = "ApplyOptionalPatch patch-%{patchversion}-redhat.patch"
+replacement = "# ApplyOptionalPatch patch-%{patchversion}-redhat.patch (disabled for AzureLinux)"
```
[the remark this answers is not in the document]
rlmenge: Was reverted due to a bad build enviornment

base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
...
+# Regenerate: python3 scripts/generate-kernels.py
+###############################################################################
 
-# Add without = ["debug"] to skip debug kernel variant and reduce build time
-[components.kernel.build]
-without = ["debug"]
+# Standard Azure Linux kernel
+
+[components."kernel"]
+spec = { type = "upstream", upstream-name = "kernel" }
```
reubeno: Since this is the default, it can be omitted. Also a note that `kernel` doesn't need to be in quotes in the `[components."kernel"]` syntax; the quotes are generally only needed when component names include periods or plus signs or other punctuation that is meaningful in TOML files.

base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
...
+
+[components."kernel".build]
+# debugbuildsenabled: set to "0" to skip debug kernel variant (faster builds)
+# azl_pkgrelease: RPM release number for the Azure Linux kernel package
+# kextraversion: 4th version component appended to EXTRAVERSION so uname -r
+# matches source version (6.18.5.1).
+defines = { debugbuildsenabled = "0", azl_pkgrelease = "1", kextraversion = ".1" }
```
reubeno: Minor style suggestion. If you want to, you can break out the `defines` into their own section and then have comments go with the individual items, e.g.:

```toml
[components.kernel.build.defines]
# set to "0" to skip debug kernel variant (faster builds)
debugbuildsenabled = "0"
# RPM release number for the Azure Linux kernel package
azl_pkgrelease = "1"
# 4th version component appended to EXTRAVERSION so uname -r matches source version (6.18.5.1).
kextraversion = ".1"
```

That's a matter of taste and up to you as the maintainer of this component.

base/comps/kernel/kernel.comp.toml
reubeno: That's a good idea! I think that will be easier to read/maintain.

base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
...
+cp %{SOURCE10001} kernel-%{specrpmversion}-aarch64.config"""
+
+[[components.kernel.overlays]]
+description = "Set specrpmversion to 6.18.5.1"
+type = "spec-search-replace"
+regex = "%define specrpmversion .+"
+replacement = """%define azurelinux_version 3
+%define specrpmversion 6.18.5.1"""
+
+[[components.kernel.overlays]]
```
reubeno, line 29: Since you have so many overlays, consider whether you'd find it easier to read like this or formatted like we did [here](https://github.com/microsoft/azurelinux/blob/tomls/base/main/base/comps/azurelinux-rpm-config/azurelinux-rpm-config.comp.toml).

That's also purely a matter of style and taste for you to decide.

base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
...
+replacement = "CBL-Mariner-Linux-Kernel-rolling-lts-mariner-%{azurelinux_version}-%{specrpmversion}"
+
+[[components.kernel.overlays]]
+description = "Prepend kextraversion (.1) to EXTRAVERSION so uname -r = 6.18.5.1-release (matches module path)"
+type = "spec-search-replace"
+regex = 'EXTRAVERSION = -\$\{ShortRel\}'
+replacement = 'EXTRAVERSION = %{kextraversion}-${ShortRel}'
+
+[[components.kernel.overlays]]
+description = "Disable redhat patch declaration"
```
reubeno: There's a `patch-remove` overlay that we added based on @ddstreet's feedback; you should consider whether that will work for you.

base/comps/kernel/kernel.comp.toml
```diff
@@ -1,5 +1,123 @@
...
+[[components."kernel".source-files]]
+filename = "kernel-6.18.5.1.tar.gz"
+hash = "0d8c755e02857704eb4c4396657ad30b22b474296233b8681365aeaa2fc16926afe31f551e80f565e0d73318f2b7941051ec44f85ccbabaa28ba5a56022f9f4a"
+hash-type = "SHA512"
+origin = { type = "download", uri = "https://github.com/microsoft/CBL-Mariner-Linux-Kernel/archive/rolling-lts/mariner-3/6.18.5.1.tar.gz" }
+
+[[components.kernel.overlays]]
+description = "Add Azure Linux x86_64 config file"
+type = "file-add"
+file = "6.18-x86_64-azl.config"
```
reubeno: This is going to be a complicated one, necessarily. Please do give us feedback on maintainability.

For example, we haven't thought about variable setting/replacement in TOML files but I see that you need to replicate `6.18` or `6.18.5.1` or similar in a number of places. The ideal would be to refer to spec macros for that and `%define` that in one place, but some of the references here are TOML-specific ones that don't get RPM spec macro-expanded. We *could* consider allowing something like:

```toml
[components.kernel.variables]
kernel-minor = "6.18"

[[components.kernel.overlays]]
type = "spec-search-replace"
...
replacement = "something something something {kernel-minor} something"
expand-vars = true
```

...but we'll definitely want to make sure it's worth the complexity to add it before we jump to doing something like that. Your feedback will help :smile:

base/comps/kernel/kernel.comp.toml
rlmenge: This was an artifact of my kernel variant generation scripts as some of them were accounting for supporting multiple versions of the kernel (I.e. "kernel6.18"). Will remove for now though :)

base/comps/kernel/kernel.comp.toml
rlmenge: It looks like our current suggestion is not to use the overlay array? https://github.com/microsoft/azurelinux/blob/tomls/base/main/.github/instructions/comp-toml.instructions.md

base/comps/kernel/kernel.comp.toml
rlmenge: Looks like patch-remove doesn't work for the kernel's redhat patch.

AI summary from my tests:

1. Macro form (file = "patch-%{patchversion}-redhat.patch") — patch-remove doesn't match because it parses the Patch1: tag structurally rather than matching raw text, and fails to handle %{patchversion} as a literal string.
2. Resolved form (file = "patch-6.18-redhat.patch") — also fails, suggesting the parser doesn't fully resolve RPM macros either. 

Neither form finds a match. Sticking with replace for now but also trying to update the patch remove to handle macros

base/comps/kernel/kernel.comp.toml
rlmenge: Delaying the config overlay for another PR. Resolving this thread for now

base/comps/kernel/kernel.comp.toml
rlmenge: I prefer this style. Have updated

ddstreetmicrosoft:
The kernel is uniquely complicated; I think maintaining a full patched spec and corresponding build files should be seriously considered. Will that be easier to understand and maintain than the toml?

rlmenge:
> The kernel is uniquely complicated; I think maintaining a full patched spec and corresponding build files should be seriously considered. Will that be easier to understand and maintain than the toml?

Agreed. I have a seperate workstream where I am evaluating making a simplified version of the kernel.spec. This will likely be presented as a new PR but in the effort of getting _something_ I am moving with Fedora's kernel (as well as giving myself time to stumble upon all the side effects of how Fedora formats their modules, uname etc)

base/comps/kernel/kernel.comp.toml
reubeno: That was AI generated and not something I'd agree with. I think it's a developer taste preference for readability right now.

rlmenge:
> > > The kernel is uniquely complicated; I think maintaining a full patched spec and corresponding build files should be seriously considered. Will that be easier to understand and maintain than the toml?
> > 
> > 
> > Agreed. I have a seperate workstream where I am evaluating making a simplified version of the kernel.spec. This will likely be presented as a new PR but in the effort of getting _something_ I am moving with Fedora's kernel (as well as giving myself time to stumble upon all the side effects of how Fedora formats their modules, uname etc)
> 
> I'm good with that, but I think this patchset is complex enough that it needs a logical rebase to be able to properly review it.

I have rebased and restructred the commits to be a more logical flow. The last commit (https://github.com/microsoft/azurelinux/pull/15961/changes/584177f5af9c77ab56b99f6997bbe2148e8bfbdd) is still large due to many defines for the kernel version within the spec. They are all needed to ensure a proper build.

base/comps/kernel/kernel.comp.toml
```diff
@@ -4,13 +4,72 @@
...
+[[components.kernel.overlays]]
+description = "Use AzureLinux kernel source URL"
+type = "spec-update-tag"
+tag = "Source0"
+value = "kernel-%{specrpmversion}.tar.gz"
+
+[[components.kernel.overlays]]
+description = "Fix tarball extraction path for AzureLinux"
+type = "spec-search-replace"
+regex = "linux-%{tarfile_release}"
```
ddstreetmicrosoft, line 63: be careful here, the regex `linux-%{tarfile_release}` actually matches twice in the spec:
> Source0: linux-%{tarfile_release}.tar.xz                                                                                                                                                      
> mv linux-%{tarfile_release} linux-%{KVERREL}                                                                                                                                                  
I *think* this is safe because the previous overlay updated the `Source0:` line, but still I'm a bit concerned about  all this regex-replace.

base/comps/kernel/kernel.comp.toml
```diff
@@ -4,13 +4,72 @@
...
+replacement = "CBL-Mariner-Linux-Kernel-rolling-lts-mariner-%{azurelinux_version}-%{specrpmversion}"
+
+# Disable redhat-specific sources and patches since they are not applicable to AzureLinux
+[[components.kernel.overlays]]
+description = "Disable copy of Makefile.rhelver (Source1) — not applicable to AzureLinux"
+type = "spec-search-replace"
+regex = "cp -a %{SOURCE1} ."
+replacement = "# cp -a %{SOURCE1} . (disabled for AzureLinux — Makefile.rhelver is Red Hat-specific)"
```
ddstreetmicrosoft, line 71: Is this really needed? I don't see anywhere in the spec that actually refers to any `Makefile.rhelver` (besides dropping it into the build dir), so I assume the only place that references it is inside fedora's kernel source, which looks like only [here](https://gitlab.com/cki-project/kernel-ark/-/blob/fedora-6.18/Makefile?ref_type=heads#L1318). So it should be safe to just leave this unchanged right?

ddstreetmicrosoft (approved):
A couple minor comments for your future consideration, in general LGTM

base/comps/kernel/kernel.comp.toml
rlmenge: Originally tried this route but I had to add extra Extraversion logic. It  was unhappy with the LSG extraversion otherwise

PR #15969, opened 2026-02-24
fix(python-jaraco-context): add missing build dep and disable checks

dmcilvaney:
python-jaraco-context 6.1.0 fails to build because coherent.licensed is
listed in pyproject.toml build-system.requires but the corresponding
package python-coherent-licensed was not available in AZL.

Add python-coherent-licensed as a new component to satisfy the build
dependency. Disable check suite for initial bringup — tests pass
(14/14) but require python-jaraco-test >= 5.6.0 (rawhide), while AZL
currently ships 5.5.0 (f43). Re-enable once python-jaraco-test is
updated.

### DIFF
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -2872,8 +2872,6 @@
 [components.python-cligj]
 [components.python-cloudpickle]
 [components.python-cmd2]
-# Required by python-jaraco-context: coherent.licensed is listed in its
-# pyproject.toml build-system.requires (Koji task 158990).
 [components.python-coherent-licensed]
 [components.python-coincidence]
 [components.python-coloredlogs]


base/comps/python-jaraco-test/python-jaraco-test.comp.toml
```diff
@@ -0,0 +1,4 @@
+[components.python-jaraco-test]
+# Pin to rawhide for 5.6.0 — needed by python-jaraco-context's test suite
+# (jaraco-test >= 5.6). f43 is still at 5.5.1. Re-align once f43 catches up.
+spec = { type = "upstream", upstream-distro = { name = "fedora", version = "rawhide" } }
```
dmcilvaney: Just disable tests for now

base/comps/components-full.toml
```diff
@@ -2872,6 +2872,9 @@
 [components.python-cligj]
 [components.python-cloudpickle]
 [components.python-cmd2]
+# Required by python-jaraco-context: coherent.licensed is listed in its
+# pyproject.toml build-system.requires (Koji task 158990).
+[components.python-coherent-licensed]
```
dmcilvaney:
```suggestion
[components.python-coherent-licensed]
```

PR #15972, opened 2026-02-24
fix(zbar): drop GIR typelib/gir file listings from %files sections

dmcilvaney:
The zbar f43 spec lists ZBar-1.0.typelib and ZBar-1.0.gir in the zbar-gtk and zbar-gtk-devel subpackages, but those files are only generated when g-ir-scanner is present at build time. Since GLib 2.80, GObject Introspection tools were merged into the glib2 source tree, but the legacy g-ir-scanner binary still ships in the separate gobject-introspection package. The zbar spec never explicitly BuildRequires it — it only worked by accident via transitive deps that no longer pull it in.

Fedora hit the same FTBFS and removed the GIR file listings in rawhide: https://src.fedoraproject.org/rpms/zbar/c/a1e97927fd862a8ca777462155b54c4ee4176923

Apply four spec-search-replace overlays to remove the four affected lines, aligning with the upstream fix.

Tested: built locally, zbar-gtk RPM no longer lists typelib, zbar-gtk-devel no longer lists .gir, zbarimg --version works in mock.

### DIFF
--- a/base/comps/zbar/zbar.comp.toml
+++ b/base/comps/zbar/zbar.comp.toml
@@ -10,33 +10,17 @@
 # Ref: https://src.fedoraproject.org/rpms/zbar/c/a1e97927fd862a8ca777462155b54c4ee4176923 ("Fix FTBFS")
 
 [[components.zbar.overlays]]
-description = "Remove ZBar typelib directory from zbar-gtk %files (g-ir-scanner not available, aligns with Fedora rawhide)"
+description = "Remove GIR typelib entries from zbar-gtk %files (g-ir-scanner not available, aligns with Fedora rawhide)"
 type = "spec-search-replace"
 section = "%files"
 package = "gtk"
-regex = '%dir %\{_libdir\}/girepository-1\.0'
+regex = '%.*girepository-1\.0.*'
 replacement = ""
 
 [[components.zbar.overlays]]
-description = "Remove ZBar typelib from zbar-gtk %files (g-ir-scanner not available, aligns with Fedora rawhide)"
-type = "spec-search-replace"
-section = "%files"
-package = "gtk"
-regex = '%\{_libdir\}/girepository-1\.0/ZBar-1\.0\.typelib'
-replacement = ""
-
-[[components.zbar.overlays]]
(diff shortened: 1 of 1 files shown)


base/comps/zbar/zbar.comp.toml
```diff
@@ -0,0 +1,42 @@
...
+[[components.zbar.overlays]]
+description = "Remove ZBar typelib from zbar-gtk %files (g-ir-scanner not available, aligns with Fedora rawhide)"
+type = "spec-search-replace"
+section = "%files"
+package = "gtk"
+regex = '%\{_libdir\}/girepository-1\.0/ZBar-1\.0\.typelib'
+replacement = ""
+
+[[components.zbar.overlays]]
+description = "Remove ZBar .gir directory from zbar-gtk-devel %files (g-ir-scanner not available, aligns with Fedora rawhide)"
```
reubeno: Couldn't you combine this overlay and the one after it to blanket remove any reference to `%{_datadir}/gir-1.0` under %files? That seems to be the *intent*.

base/comps/zbar/zbar.comp.toml
```diff
@@ -0,0 +1,42 @@
...
+# We do the same: remove the typelib/gir entries from %files so the build doesn't
+# fail when g-ir-scanner isn't present.
+# Ref: https://src.fedoraproject.org/rpms/zbar/c/a1e97927fd862a8ca777462155b54c4ee4176923 ("Fix FTBFS")
+
+[[components.zbar.overlays]]
+description = "Remove ZBar typelib directory from zbar-gtk %files (g-ir-scanner not available, aligns with Fedora rawhide)"
+type = "spec-search-replace"
+section = "%files"
+package = "gtk"
+regex = '%dir %\{_libdir\}/girepository-1\.0'
```
reubeno: See other comment; what about merging the changes that affect this dir or things underneath it? That makes the change more concise. If you wanted to be fancier with regex you could probably even combine this all in a single search-replace against `%files`, although I'll acknowledge merging to 1 vs. 2 regex replacements may minorly decrease readability.

reubeno (approved):

PR #15974, opened 2026-02-24
fix(opencv): remove CVE-2025-53644 patch targeting deleted bundled openjpeg

christopherco:
The upstream Fedora spec (f7431cf) added a patch for CVE-2025-53644 that modifies 3rdparty/openjpeg/openjp2/jp2.c, but the spec's %prep deletes the bundled 3rdparty/openjpeg directory since it uses system openjpeg2-devel instead. This causes the build to fail with "can't find file to patch".

Add overlays to remove the Patch17 declaration and its %patch application. The CVE fix should be tracked in the system openjpeg package.

Also remove stale opencv entries from components-full.toml that conflict with their dedicated comp.toml files.

### DIFF
--- a/base/comps/opencv/opencv.comp.toml
+++ b/base/comps/opencv/opencv.comp.toml
@@ -1,0 +1,18 @@
+[components.opencv]
+
+# CVE-2025-53644 patch targets bundled 3rdparty/openjpeg, which Fedora's spec
+# deletes during %prep (it uses system openjpeg2-devel instead). The patch
+# can't find the file and fails. Remove both the patch declaration and its
+# application. The CVE fix belongs in the system openjpeg package, not here.
+# Upstream Fedora offending commit: https://src.fedoraproject.org/rpms/opencv/c/f7431cf
+
+[[components.opencv.overlays]]
+description = "Remove CVE-2025-53644 patch — targets bundled openjpeg that is deleted in %prep; fix belongs in system openjpeg package"
+type = "patch-remove"
+file = "**/a39db41390de546d18962ee1278bd6dbb715f466.patch"
+
+[[components.opencv.overlays]]
+description = "Remove application of CVE-2025-53644 patch (see above)"
+type = "spec-search-replace"
+regex = '%patch.*cve2025-53644'
+replacement = ""
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -1144,7 +1144,6 @@
 [components.openbox]
 [components.opencl-headers]
 [components.opencore-amr]
-[components.opencv]
 [components.openexr2]
 [components.openfec]
 [components.openjade]


dmcilvaney (approved):

base/comps/opencv/opencv.comp.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[[components.opencv.overlays]]
+description = "Remove CVE-2025-53644 patch — targets bundled openjpeg that is deleted in %prep; fix belongs in system openjpeg package"
+type = "spec-remove-tag"
+tag = "Patch17"
+
+[[components.opencv.overlays]]
+description = "Remove application of CVE-2025-53644 patch (see above)"
+type = "spec-search-replace"
+regex = '%patch -P 17 -p1 -b \.cve2025-53644'
```
reubeno: Is the full line needed in the regex? Why not just target:

```
%patch .*cve2025-53644
```

base/comps/opencv/opencv.comp.toml
```diff
@@ -0,0 +1,18 @@
...
+# CVE-2025-53644 patch targets bundled 3rdparty/openjpeg, which Fedora's spec
+# deletes during %prep (it uses system openjpeg2-devel instead). The patch
+# can't find the file and fails. Remove both the patch declaration and its
+# application. The CVE fix belongs in the system openjpeg package, not here.
+# Upstream Fedora offending commit: https://src.fedoraproject.org/rpms/opencv/c/f7431cf
+
+[[components.opencv.overlays]]
+description = "Remove CVE-2025-53644 patch — targets bundled openjpeg that is deleted in %prep; fix belongs in system openjpeg package"
+type = "spec-remove-tag"
+tag = "Patch17"
```
reubeno: Do we need a feature to be able to remove a patch by patch name so the overlay doesn't need to be coupled to the current NNN value of PatchNNN?

ddstreetmicrosoft: I think there is already the `patch-remove` overlay command that should cover this case?

base/comps/opencv/opencv.comp.toml
christopherco: Full line isn't needed. Will loosen the regex.

christopherco: Pushed a new change. The important part of the regex here is `%patch` and `-P 17` which indicates to patch using Patch17. The `-b <suffix>` is just for creating a backup file that was patched (using the provided suffix as the backup file extension).

base/comps/opencv/opencv.comp.toml
christopherco: Updated to use `patch-remove` overlay for this case

base/comps/opencv/opencv.comp.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[[components.opencv.overlays]]
+description = "Remove CVE-2025-53644 patch — targets bundled openjpeg that is deleted in %prep; fix belongs in system openjpeg package"
+type = "patch-remove"
+file = "**/a39db41390de546d18962ee1278bd6dbb715f466.patch"
+
+[[components.opencv.overlays]]
+description = "Remove application of CVE-2025-53644 patch (see above)"
+type = "spec-search-replace"
+regex = '%patch\b.*-P\s*17\b.*'
```
reubeno: I know you've already iterated on this, but...

What we really care about is removing the patch for the CVE. Is there a way to have the targeted regex focused on *that*, rather than focused on "patch17"?

e.g.,:

```
regex = '%patch.*cve2025-53644'
```

The above is nice and simple, and will start failing to apply when the upstream spec removes a patch for this CVE.

As the PR currently stands, if the patch were swapped out with a different "Patch17", we would unwittingly remove the wrong `%patch` directive.

christopherco: Fundamentally "cve2025-53644" in that line is purely optional and easily removable/different across versions of the base package as well. That is just a "backup file extension" string for the %patch command to apply a suffix to the backup file. It is not directly an indicator of the actual patch file itself, and is freely editable.

That being said, I get what you mean about the failure mode when tying the regex to patch17 explicitly. Will update to condition on this optional backup extension suffix instead.

christopherco: Updated to `regex = '%patch.*cve2025-53644'`

PR #15975, opened 2026-02-25
Elide wxGTK test-only BuildRequires, and remove xclock package

ddstreet:
The xclock package doesn't build, but it's gui so we don't want to include it anyway. It's only BuildRequired: by the wxGTK package, and only as a test requirement.

This changes the wxGTK test disabling from `check = { skip = true }` over to setting the spec `tests` macro to default to false, which disables the tests and elides the test-only BuildRequires.

### DIFF
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -2956,7 +2956,6 @@
 [components.xcb-util-renderutil]
 [components.xcb-util-wm]
 [components.xclip]
-[components.xclock]
 [components.xdg-dbus-proxy]
 [components.xdg-user-dirs]
 [components.xdg-utils]


base/comps/wxGTK/wxGTK.comp.toml
```diff
@@ -1,4 +1,7 @@
 [components.wxGTK]
 
-[components.wxGTK.build]
-check = { skip = true, skip_reason = "Disabling checks for initial set of failures." }
+[[components.wxGTK.overlays]]
+description = "Disable checks for initial set of failures. Also elide the test-only BuildRequires as we do not provide all of them."
+type = "spec-search-replace"
+regex = "%bcond_without tests"
```
reubeno: If it's a bcond we have a cleaner way to disable, via the `with`/`without` options.

ddstreetmicrosoft: thanks, updated

reubeno:
@ddstreet I signed off, but looks like you'll need to resolve the merge conflict?

ddstreetmicrosoft:
> @ddstreet I signed off, but looks like you'll need to resolve the merge conflict?

thanks, rebased and should be ready to go now

PR #15980, opened 2026-02-25
Remove apache-ivy support for Pack200 compression, to fix FTBFS

ddstreet:
This package FTBFS because java dropped Pack200 support. This backports a patch to make Pack200 support optional, as well as disabling that optional Pack200 support, so the package builds.

Note this package's upstream source tarball contains files in Windows (CRLF) format, which prevents git patches from cleanly applying; so this PR does more work than should be required simply to add two patch files.

### DIFF
--- a/base/comps/apache-ivy/apache-ivy.comp.toml
+++ b/base/comps/apache-ivy/apache-ivy.comp.toml
@@ -7,4 +7,6 @@ overlays = [
     { type = "spec-add-tag", tag = "Source5", value = "remove-Pack200Packing-java.patch" },
 
     { type = "spec-search-replace", regex = "dos2unix src/java/org/apache/ivy/ant/IvyAntSettings.java", replacement = "dos2unix src/java/org/apache/ivy/ant/IvyAntSettings.java asciidoc/release-notes.adoc ivy.xml optional.patterns src/java/org/apache/ivy/core/pack/PackagingManager.java src/java/org/apache/ivy/core/pack/PackingRegistry.java src/java/org/apache/ivy/core/settings/IvySettings.java src/java/org/apache/ivy/core/pack/ArchivePacking.java src/java/org/apache/ivy/core/pack/OsgiBundlePacking.java src/java/org/apache/ivy/core/pack/Pack200Packing.java src/java/org/apache/ivy/core/pack/StreamPacking.java src/java/org/apache/ivy/core/pack/ZipPacking.java src/java/org/apache/ivy/util/FileUtil.java test/java/org/apache/ivy/core/retrieve/RetrieveTest.java test/java/org/apache/ivy/core/pack/ZipPackingTest.java version.properties" },
+
+    { type = "spec-search-replace", regex = "patch -p1 -l < %{SOURCE3}", replacement = "patch -p1 -l < %{SOURCE3}\npatch -p1 -l < %{SOURCE4}\npatch -p1 -l < %{SOURCE5}" },
 ]


ddstreet:
@reubeno please let me know if there's a better way I could do this with the tooling...it seems like a bit of a special case package that comes to us already doing stuff the wrong way (i.e. patch handling)

base/comps/apache-ivy/apache-ivy.comp.toml
```diff
@@ -0,0 +1,37 @@
...
+[[components.apache-ivy.overlays]]
+description = "Add upstream patch file to make Pack200 support optional."
+type = "file-add"
+file = "IMPROVEMENT-use-Apache-Commons-Compress-for-pack200-.patch"
+source = "IMPROVEMENT-use-Apache-Commons-Compress-for-pack200-.patch"
+
+[[components.apache-ivy.overlays]]
+description = "Add upstream patch file tag to make Pack200 support optional."
+type = "spec-insert-tag"
+tag = "Source4"
```
reubeno: I see that the upstream spec already does some patching via SourceXXX files instead of PatchXXX files... but it also does use `%autosetup`.

Did you look into whether the new patches you're introducing could be applied as normal PatchXXX files? Or is the issue that they need to be applied after the others that are already being applied oddly?

(We do have a `patch-add` overlay, but I'm assuming you didn't go for it because the oddities you mention in the preexisting spec.)

ddstreet: I tried `patch-add` first, but the problem is it'll then get applied by `%autosetup`, and the patch that `git format-patch` creates from the upstream source doesn't apply because of the source tarball's CRLF line endings; it fails with errors like:

> checking file asciidoc/release-notes.adoc
> Hunk #1 FAILED at 55 (different line endings).
> 1 out of 1 hunk FAILED
> checking file ivy.xml
> Hunk #1 FAILED at 32 (different line endings).
> Hunk #2 FAILED at 46 (different line endings).
> Hunk #3 FAILED at 55 (different line endings).
> Hunk #4 FAILED at 67 (different line endings).
> 

I did some brief tests to see if I could unix2dos the patch file but that didn't help, and this is such a rare situation (I've never seen this problem before) I didn't want to spend too much time trying to finesse the patch file into just the right line endings in the right places to apply. I assume the upstream packager came to the same conclusion.

reubeno:
> @reubeno please let me know if there's a better way I could do this with the tooling...it seems like a bit of a special case package that comes to us already doing stuff the wrong way (i.e. patch handling)

Agreed -- it looks like the upstream spec was already complicated. About the only thing that comes to mind is that it *might* be easier to read/understand the `.toml` file if we formatted it like [this one](https://github.com/microsoft/azurelinux/blob/tomls/base/main/base/comps/azurelinux-rpm-config/azurelinux-rpm-config.comp.toml) but I think that's a matter of taste and style. I'd defer to you on that.

One note -- `spec-insert-tag` was added while I was on vacation. It's a highly specialized tag that we *generally* may not want to use. We should review this some. `spec-add-tag` was the one intended for general usage, and it fails if it tries to add a tag that already exists. I think the new one was added due to issues with `%if` conditionals and other special cases.

I think the only other considerations I'd take from here (but not change this PR for) are:

* It *might* be useful to have a `source-add` overlay (like the `patch-add` one) but the trickiness would be on how to figure out which auto-generated `NNN` (for `SourceNNN`) was used -- and that might make it not so useful.

ddstreetmicrosoft:
rebase on main, i'll address comments next

ddstreet:
> > @reubeno please let me know if there's a better way I could do this with the tooling...it seems like a bit of a special case package that comes to us already doing stuff the wrong way (i.e. patch handling)
> 
> Agreed -- it looks like the upstream spec was already complicated. About the only thing that comes to mind is that it _might_ be easier to read/understand the `.toml` file if we formatted it like [this one](https://github.com/microsoft/azurelinux/blob/tomls/base/main/base/comps/azurelinux-rpm-config/azurelinux-rpm-config.comp.toml) but I think that's a matter of taste and style. I'd defer to you on that.

ack, updated

> 
> One note -- `spec-insert-tag` was added while I was on vacation. It's a highly specialized tag that we _generally_ may not want to use. We should review this some. `spec-add-tag` was the one intended for general usage, and it fails if it tries to add a tag that already exists. I think the new one was added due to issues with `%if` conditionals and other special cases.

ack, changed over to use `spec-add-tag`

> 
> I think the only other considerations I'd take from here (but not change this PR for) are:
> 
> * It _might_ be useful to have a `source-add` overlay (like the `patch-add` one) but the trickiness would be on how to figure out which auto-generated `NNN` (for `SourceNNN`) was used -- and that might make it not so useful.

yeah i think this would be a useful addition.

PR #15998, opened 2026-02-25
fix (socat): skip VSOCK_ECHO test (410) in %check and export SHELL env variable in %check

liunan-ms:
This PR:

- Adds a `spec-search-replace` overlay to append test 410 to the existing NUMCOND exclusion list (alongside 146, 478, 528). 
- Adds an overlay to export the SHELL environment variable in the `%check` section to fix the `SHELL variable undefined` error that caused 14 test failures (495-502, 506, 508, 587, 590-593).
This error only happens in Koji build.
Mock's built-in Python defaults ([config.py](https://github.com/rpm-software-management/mock/blob/main/mock/py/mockbuild/config.py#L268)) do set `SHELL=/bin/bash` and `azurelinux-4.0.tpl` does not override `config_opts['environment']`. We will check inside the Koji builders to see if there's any overrides.
- Moves `socat` from inline entry in `components-full.toml` to a dedicated comp.toml.

Validation: 

- Koji scratch build passed: https://52.249.25.247/koji/taskinfo?taskID=241084
- Local build succeeds with `azldev comp build -p socat`, 0 test failures (589 selected, 304 ok)

### DIFF
--- a/base/comps/socat/socat.comp.toml
+++ b/base/comps/socat/socat.comp.toml
@@ -1,6 +1,12 @@
 [components.socat]
 
 [[components.socat.overlays]]
+description = "Export SHELL env var — socat SHELL address type requires it but mock chroot does not set it"
+type = "spec-prepend-lines"
+section = "%check"
+lines = ["export SHELL=/bin/bash"]
+
+[[components.socat.overlays]]
 description = "Exclude VSOCK_ECHO test (410) — /dev/vsock unavailable in mock chroot"
 type = "spec-search-replace"
 section = "%check"


base/comps/socat/socat.comp.toml
```diff
@@ -0,0 +1,8 @@
+[components.socat]
+
+[[components.socat.overlays]]
+description = "Exclude VSOCK_ECHO test (410) — /dev/vsock unavailable in mock chroot"
+type = "spec-search-replace"
+section = "%check"
+regex = '-a \\\\\$N -ne 528'
+replacement = '-a \\$N -ne 410 -a \\$N -ne 528'
```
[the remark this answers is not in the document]
liunan-ms: `replacement = '-a \\$N -ne 410 -a \\$N -ne 528'` produces the literal `\\$N` that is aligned with other excluded tests in the spec: https://src.fedoraproject.org/rpms/socat/blob/f43/f/socat.spec#_66

base/comps/socat/socat.comp.toml
```diff
@@ -1,5 +1,11 @@
 [components.socat]
 
+[[components.socat.overlays]]
+description = "Export SHELL env var — socat SHELL address type requires it but mock chroot does not set it"
+type = "spec-prepend-lines"
+section = "%check"
+lines = ["export SHELL=/bin/bash"]
```
ddstreetmicrosoft, line 7: I updated a mock build spec file like:

```
%check
echo $SHELL
exit 1

```

and it resulted in:

> Executing(%check): /bin/sh -e /var/tmp/rpm-tmp.OOpbK3                                                                                                                                         
> + umask 022                                                                                                                                                                                   
> (...setting lots of *FLAG env vars...)
> + cd /builddir/build/BUILD/socat-1.8.0.3-build                                                                                                                                                
> + LT_SYS_LIBRARY_PATH=/usr/lib64:                                                                                                                                                             
> + export LT_SYS_LIBRARY_PATH                                                                                                                                                                  
> + CC=gcc                                                                                                                                                                                      
> + export CC                                                                                                                                                                                   
> + CXX=g++                                                                                                                                                                                     
> + export CXX                                                                                                                                                                                  
> + cd socat-1.8.0.3                                                                                                                                                                            
> + echo /bin/bash                                                                                                                                                                              
> /bin/bash                                                                                                                                                                                     
> + exit 1                                                                                                                                                                                      
> 

so it looks like $SHELL is already defined inside `%check`, i'm confused about why this is needed?

ddstreetmicrosoft: to clarify, i'm fine with checking this in, but it seems like it's actually a bug in our koji build environment, and i think we should mark it as a workaround so we can go back and dig deeper when we have time. if there is something i'm missing that would indicate this *isn't* a workaround please add the detail here, otherwise please mark this commit as a workaround (e.g. with 'workaround' in the commit subject) and i'm good to merge it after that

liunan-ms: > I updated a mock build spec file like:
> 
> ```
> %check
> echo $SHELL
> exit 1
> ```
> so it looks like $SHELL is already defined inside `%check`, i'm confused about why this is needed?

Is this in local build or Koji build? I didn't hit `SHELL variable undefined` error in local build where SHELL is set, only see it undefined in Koji build. I'll check if there's a different mock config in koji builder. will also mark it as a workaround.

ddstreetmicrosoft (approved):
LGTM, thanks!

PR #16003, opened 2026-02-26
fix(systemd): lower setup version requirement to match AZL's available package

aditjha-msft:
Upstream Fedora 41+ systemd (v258.4) requires setup >= 2.15.0-3 via both BuildRequires and Conflicts tags (added for /usr/lib/sysusers.d ownership in setup-2.15.0-3: https://src.fedoraproject.org/rpms/setup/commits/rawhide).

AZL currently ships setup-2.15.0-1, which causes systemd to conflict at install time -- blocking any image build that includes systemd (i.e., nearly every image).

This adds two spec-search-replace overlays to lower the threshold from 2.15.0-3 to 2.15.0-1, and migrates the systemd component from an inline entry in components.toml to a dedicated systemd.comp.toml file.

Verified overlays apply cleanly via prep-sources diff. Build and smoke-test pending.


reubeno:
> This is a workaround but also needed for the nightly build. I've got this workaround in tracking to undo later.

I'm assuming the key issue we're hitting here is %autorelease not working (yet); because it's not working as expected, many of our built RPMs are reverting to `-1`, causing these issues. We need a general category fix.

PR #16012, opened 2026-02-26
fix(shim): replace 'sources' file to use our shim binaries from our lookaside cache

ddstreet:
The shim binary files are kept in our lookaside cache, and the 'sources' file contains the hashes so the build will download the correct files before building.


ddstreet:
@PawelWMS @reubeno
This may not work currently, if our build processes the `sources` file *before* applying our overlays, but it needs to work that way eventually (i.e. we need to be able to manage the `sources` file normally).

If there's a different way to do it (other than a `file-add` overlay) please let me know, I can't find any documentation on the toml besides the overlays.

ddstreet:
also note this PR isn't strictly needed right now; the previous workaround #15971 causes the locally installed unsigned shim binaries to replace whatever was downloaded from the lookaside cache. it's fine with me if we close this and wait for an update to the azldev tool so it can properly update the `sources` file, and then open a new PR with the new toml command.

ddstreetmicrosoft:
leaving this as draft, since it currently won't work anyway (as noted [above](https://github.com/microsoft/azurelinux/pull/16012#issuecomment-3967339456)).

PR #16020, opened 2026-02-26
Remove fedora pkgs, add fedora 'compat' package

ddstreet:
(Draft/RFC)

This sets up a compat components that tries to create some fedora->azl logic to handle any bootstrap-time dependencies or file/dir naming issues. Open for review and discussion but marked as draft.

### DIFF
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -246,7 +246,6 @@
 [components.fdk-aac-free]
 [components.fdupes]
 [components.fedora-logos]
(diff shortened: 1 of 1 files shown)


ddstreet:
this is probably no longer necessary after #16025, unless we do run into any build breakage due to packages still referencing/requiring anything fedora/redhat, and that isn't covered by existing bootstrapping workarounds

ddstreetmicrosoft:
closing this, as i don't think we will need the compat pkg anymore.

PR #16022, opened 2026-02-26
fix(openpace): add missing $0 placeholder in install-data-local find -exec calls

WithEnoughCoffee:
## Summary

The `install-data-local` target in openpace's `Makefile.am` uses
`find -exec sh -c '...' {} +` without a dummy `$0` placeholder.
Per POSIX, the first argument after `sh -c script` becomes `$0` and is
excluded from `$@`, causing the first file returned by `find` to be
silently dropped on every build.

This causes noarch rpmdiff failures in Koji because different filesystem
types on x86_64 vs aarch64 builders return files in different `find`
order, so different files are dropped on each architecture.

## The Fix

Add the conventional `_` placeholder as `$0` before `{}` on both
`find` calls in `install-data-local`:

```diff
-find $(srcdir)/docs -type d -exec sh -c '...' {} +
+find $(srcdir)/docs -type d -exec sh -c '...' _ {} +
-find $(srcdir)/docs -type f -exec sh -c '...' {} +
+find $(srcdir)/docs -type f -exec sh -c '...' _ {} +
```

## Alternatives considered

- **`rsync -a`** — simplest but adds a build dependency
- **`find | xargs`** — avoids the `sh -c` `$0` trap but introduces quoting differences
- **`cp -a`** — clearer but loses batching optimization

The minimal `_` fix was chosen as the smallest correct change with no
behavior or dependency changes.

## Verification

**Koji scratch build passed** ✅
- Task ID: **222601**
- Task URL: https://52.249.25.247/koji/taskinfo?taskID=222601
- Both x86_64 and aarch64 `buildArch` tasks closed successfully
- No noarch rpmdiff error — doc files now match across architectures

**Local build verified** with `azldev comp build -p openpace`:
- All 6 RPMs built cleanly
- All 177 doc files present in `openpace-doc` noarch RPM (previously 176)

## References

- POSIX `sh -c` specification: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html
- ShellCheck SC2156: https://www.shellcheck.net/wiki/SC2156
- Upstream PR: https://github.com/frankmorgner/openpace/pull/TBD

### DIFF
--- a/base/comps/openpace/openpace.comp.toml
+++ b/base/comps/openpace/openpace.comp.toml
@@ -1,0 +1,25 @@
+[components.openpace]
+
+[[components.openpace.overlays]]
+description = """
+Fix install-data-local silently dropping one documentation file per build.
+
+The upstream Makefile.am uses 'find -exec sh -c ... {} +' without a dummy $0
+placeholder. Per POSIX, the first argument after 'sh -c script' becomes $0 and
+is excluded from $@, causing the first file returned by find to be silently
+dropped. Different filesystem types return files in different find order, so
+different files are dropped on x86_64 vs aarch64 builders, causing noarch
+rpmdiff failures in Koji.
+
+TODO: push change to upstream
+"""
+type = "patch-add"
+source = "fix-install-data-local-missing-dollar0.patch"
+
+[[components.openpace.overlays]]
+description = "Apply the install-data-local fix patch after existing patches"
+type = "spec-search-replace"
+section = "%prep"
+regex = '^%patch 1 -p1 -b \.unsupported-tests$'
+replacement = """%patch 1 -p1 -b .unsupported-tests
+%patch 2 -p1"""
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -1957,7 +1957,6 @@
 [components.openjpeg]
 [components.openjph]
 [components.openmpi]
-[components.openpace]
 [components.openpgm]
 [components.opensc]
 [components.openslide]
--- a/base/comps/openpace/fix-install-data-local-missing-dollar0.patch
+++ b/base/comps/openpace/fix-install-data-local-missing-dollar0.patch
@@ -1,0 +1,13 @@
+--- a/Makefile.am
++++ b/Makefile.am
+@@ -117,8 +117,8 @@
+ 
+ install-data-local:
+ 	$(INSTALL) -d $(DESTDIR)$(htmldir)
+-	find $(srcdir)/docs -type d -exec sh -c 'for f in $$@; do $(MKDIR_P)            "$(DESTDIR)$(htmldir)/$${f##$(srcdir)/docs/}"; done' {} +
+-	find $(srcdir)/docs -type f -exec sh -c 'for f in $$@; do $(INSTALL_DATA) "$$f" "$(DESTDIR)$(htmldir)/$${f##$(srcdir)/docs/}"; done' {} +
++	find $(srcdir)/docs -type d -exec sh -c 'for f in $$@; do $(MKDIR_P)            "$(DESTDIR)$(htmldir)/$${f##$(srcdir)/docs/}"; done' _ {} +
++	find $(srcdir)/docs -type f -exec sh -c 'for f in $$@; do $(INSTALL_DATA) "$$f" "$(DESTDIR)$(htmldir)/$${f##$(srcdir)/docs/}"; done' _ {} +
+ 
+ uninstall-local:
+ 	rm -rf $(DESTDIR)$(htmldir)


base/comps/openpace/openpace.comp.toml
```diff
@@ -0,0 +1,30 @@
...
+"""
+type = "patch-add"
+file = "fix-install-data-local-missing-dollar0.patch"
+source = "fix-install-data-local-missing-dollar0.patch"
+
+[[components.openpace.overlays]]
+description = "Switch to %autosetup to auto-apply all patches including the new one"
+type = "spec-search-replace"
+regex = '%setup -q'
+replacement = '%autosetup -p1'
```
tobiasb-ms: suggestion (non-blocking): Consider adding a `section` element here like:
```
section = "%prep"
```
This would make sure we don't somehow match this in a different section (would be weird, so non-blocking).

base/comps/openpace/openpace.comp.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[[components.openpace.overlays]]
+description = "Remove explicit %patch call now handled by %autosetup"
+type = "spec-search-replace"
```
binujp: usage (blocking): I believe this is a patch-remove overlay type. Can you please confirm?

base/comps/openpace/openpace.comp.toml
```diff
@@ -0,0 +1,30 @@
...
+Upstream PR: https://github.com/frankmorgner/openpace/pull/TBD
+"""
+type = "patch-add"
+file = "fix-install-data-local-missing-dollar0.patch"
+source = "fix-install-data-local-missing-dollar0.patch"
+
+[[components.openpace.overlays]]
+description = "Switch to %autosetup to auto-apply all patches including the new one"
+type = "spec-search-replace"
+regex = '%setup -q'
```
tobiasb-ms: suggestion (non-blocking): This will currently match any `%setup -q`, including if they added new arguments. I'm not sure that's what we want. Consider matching on the entire line with `^%setup -q$`, so we get a failure if that line ever changes. Minor issue though.

base/comps/openpace/openpace.comp.toml
```diff
@@ -0,0 +1,30 @@
...
+replacement = '%autosetup -p1'
+
+[[components.openpace.overlays]]
+description = "Remove explicit %patch call now handled by %autosetup"
+type = "spec-search-replace"
+regex = '%patch 1 -p1 -b .unsupported-tests'
```
tobiasb-ms: suggestion (non-blocking): As above, consider matching on the entire line to prevent some weird case where they change just the line.

base/comps/openpace/openpace.comp.toml
```diff
@@ -0,0 +1,30 @@
...
+Fix install-data-local silently dropping one documentation file per build.
+
+The upstream Makefile.am uses 'find -exec sh -c ... {} +' without a dummy $0
+placeholder. Per POSIX, the first argument after 'sh -c script' becomes $0 and
+is excluded from $@, causing the first file returned by find to be silently
+dropped. Different filesystem types return files in different find order, so
+different files are dropped on x86_64 vs aarch64 builders, causing noarch
+rpmdiff failures in Koji.
+
+Upstream PR: https://github.com/frankmorgner/openpace/pull/TBD
```
binujp: nit (style, non-blocking): This maybe better worded as a TODO: push change to upstrean

base/comps/openpace/openpace.comp.toml
reubeno: Is switching to %autosetup required for adding the patch?

base/comps/openpace/openpace.comp.toml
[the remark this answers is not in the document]
WithEnoughCoffee: changed to TODO

base/comps/openpace/openpace.comp.toml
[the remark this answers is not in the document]
WithEnoughCoffee: fixed new approach has just 1 overlay that keeps everything in place and appends %patch 2 -p1

base/comps/openpace/openpace.comp.toml
WithEnoughCoffee: fixed on line 22

WithEnoughCoffee:
> Good work on this.

Thank you!

base/comps/openpace/openpace.comp.toml
WithEnoughCoffee: This is actually adding a patch file. This is appending %patch 2 -p1 after the existing %patch 1 line.

base/comps/openpace/openpace.comp.toml
tobiasb-ms: I find `%autosetup` preferable to adding another `%patch` below.

base/comps/openpace/openpace.comp.toml
```diff
@@ -0,0 +1,25 @@
...
+type = "patch-add"
+source = "fix-install-data-local-missing-dollar0.patch"
+
+[[components.openpace.overlays]]
+description = "Apply the install-data-local fix patch after existing patches"
+type = "spec-search-replace"
+section = "%prep"
+regex = '^%patch 1 -p1 -b \.unsupported-tests$'
+replacement = """%patch 1 -p1 -b .unsupported-tests
+%patch 2 -p1"""
```
tobiasb-ms, line 25: suggestion (non-blocking): If you're doing it this way, be consistent with the existing style and make backup files, so something like:
```
%patch 2 -p1 -b .fix-install
```

tobiasb-ms (approved):

base/comps/openpace/openpace.comp.toml
WithEnoughCoffee: Is this a blocking comment? and would %patch 2 be preferred or %autosetup?

base/comps/openpace/openpace.comp.toml
WithEnoughCoffee: fixed

base/comps/openpace/openpace.comp.toml
WithEnoughCoffee: I could also make it consistent the way Tobias suggested and do` %patch 2 -p1 -b .fix-install`

If this is blocking I can change it, If not as is has passed builds locally and koji and is ready togo just let me know what you'd prefer.

PR #16034, opened 2026-02-27
workaround(shim): add dist tag to shim packages to workaround NEVR conflicts

reubeno:
`shim-unsigned-{x64,aarch64}` do not include dist tags in their `Release` tags. With our current building approach, this results in cross-tag NEVR conflicts.


base/comps/shim-unsigned-x64/shim-unsigned-x64.comp.toml
```diff
@@ -0,0 +1,9 @@
+[components.shim-unsigned-x64]
+
+# Workaround: add dist tag to the release to avoid NEVR conflicts
+# while we are still rebuilding all packages in new tags nightly.
+[[components.shim-unsigned-x64.overlays]]
+description = "Add a dist tag to avoid NEVR conflicts"
+type = "spec-update-tag"
+tag = "Release"
+value = "2%{?dist}"
```
christopherco: question: is 2 here intentional?

ddstreet:
this also doesn't fix the `shim` package which needs adjustment too

base/comps/shim-unsigned-x64/shim-unsigned-x64.comp.toml
reubeno: good catch!

reubeno:
> this also doesn't fix the `shim` package which needs adjustment too

Thank you! That was something I missed. Should be resolved in latest changes.

anphel31 (approved):

PR #16048, opened 2026-02-27
workaround(azurelinux-release): disable autorelease to avoid parse error in koji

reubeno:
This cleans up some metadata and also works around an issues we're seeing with rpmautospec in koji parsing the spec.

### DIFF
--- a/base/comps/azurelinux-release/azurelinux-release.spec
+++ b/base/comps/azurelinux-release/azurelinux-release.spec
@@ -1,4 +1,4 @@
-%define release_name "Alpha1"
+%define release_name Alpha1
 %define is_evergreen 0
 
 # Define this to 1 for Branched releases prior to RC


base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -1,4 +1,4 @@
-%define release_name "Alpha1"
```
ddstreetmicrosoft, line 1: this really *should* have been rebased back into the previous commit, but let's get this thing building.

next time please rebase into a clean patchset, don't make me get pedantic on you. ;-)

reubeno: Darn it -- I meant to do that but forgot in a hurry. Sorry, @ddstreetmicrosoft !

PR #16054, opened 2026-02-28
build: allow overriding mock baseurl for local stage2 builds

reubeno:
This allowed the following:

```
azldev comp build less --mock-config-opt stage1_repourl=http://KOJI-SERVER-HERE/REPO-PATH-HERE --mock-config-opt chroot_setup_cmd='install bash binutils bzip2 coreutils cpio diffutils dnf findutils gawk glibc-devel grep gzip kernel-headers patch redhat-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz'
```

### DIFF
--- a/distro/mock/azurelinux-4.0.tpl
+++ b/distro/mock/azurelinux-4.0.tpl
@@ -1,14 +1,11 @@
 # Disable bootstrap image for now.
 config_opts['use_bootstrap_image'] = False
 
-# General packages required
-config_opts['chroot_setup_cmd'] = 'install @{% if mirrored %}buildsys-{% endif %}build'
+# Setup buildroot and plugins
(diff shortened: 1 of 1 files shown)


ddstreetmicrosoft:
It might be better for us to just include the `comps.xml` appropriate for our needs into the local repo `azldev` sets up (using `createrepo`).

reubeno:
> It might be better for us to just include the `comps.xml` appropriate for our needs into the local repo `azldev` sets up (using `createrepo`).

Agreed we need something better -- this was just what I was able to accomplish today.

Ultimately the comps.xml or at least its core contents will live in microsoft/azurelinux and be available for local tooling to use, agreed.

PR #16055, opened 2026-02-28
fix(pcsc-lite-asekey): use gcc directly rather than a symlinked version to fix an issue in the aarch64 build

tobiasb-ms:
Currently `aarch64` build fails because it's looking for a `aarch64-koji-linux` which doesn't exist. This fix makes it just use `gcc`. Tested on koji server.

### DIFF
--- a/base/comps/pcsc-lite-asekey/pcsc-lite-asekey.comp.toml
+++ b/base/comps/pcsc-lite-asekey/pcsc-lite-asekey.comp.toml
@@ -1,0 +1,14 @@
+[components.pcsc-lite-asekey]
+
+# The upstream Makefile constructs the compiler name as CC=$(BUILD)-gcc, where
+# BUILD is the GNU triplet from ./configure --build=. For reasons that we still need
+# to investigate, on our Koji builders the triplet vendor is "koji" (e.g. aarch64-koji-linux),
+# for which no gcc symlink exists. This results in calling a symlink that doesn't exist,
+# which fails.
+# We can work around this for now by overriding CC=gcc to call gcc directory.
+[[components.pcsc-lite-asekey.overlays]]
+description = "Use plain gcc instead of triplet-prefixed compiler name to work around build on Koji builders"
+type = "spec-search-replace"
+section = "%build"
+regex = '%\{make_build\}'
+replacement = '%{make_build} CC=gcc'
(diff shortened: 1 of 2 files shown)


reubeno:
Why isn't this an issue upstream?

base/comps/pcsc-lite-asekey/pcsc-lite-asekey.comp.toml
```diff
@@ -0,0 +1,13 @@
+[components.pcsc-lite-asekey]
+
+# The upstream Makefile constructs the compiler name as CC=$(BUILD)-gcc, where
+# BUILD is the GNU triplet from ./configure --build=. On AZL Koji builders the
+# triplet vendor is "koji" (e.g. aarch64-koji-linux), for which no gcc symlink
+# exists, causing "aarch64-koji-linux-gcc: No such file or directory".
```
ddstreetmicrosoft: > On AZL Koji builders the
> triplet vendor is "koji" (e.g. aarch64-koji-linux), for which no gcc symlink
> exists
> 

This managed to build on the Fedora koji system somehow, right? So it doesn't seem like it's the package's problem, it seems like it's our koji setup's problem, unless I'm missing something?

Can you please reword the git commit to `workaround` instead of `fix`, and we can merge this until we have time to look more closely at our koji build environment.

ddstreetmicrosoft (changes requested):

base/comps/pcsc-lite-asekey/pcsc-lite-asekey.comp.toml
reubeno: +1. I'd also like us to spend a few minutes sorting out the environmental issue because it sounds like the sort of thing that could affect other packages as well.

tobiasb-ms: Discussed in parking lot today. Basic issue is that the koji server is setting `%{_vendor}` to 'koji', which makes this package try to use `aarch-koji-linux-gnu`, which of course doesn't exist. Agreed that for now we'll use this workaround, and we'll investigate the fact that the builders are configured incorrectly (in a couple ways) after alpha1.

tobiasb-ms: I reworded the comment and the commit message to reflect the fact that this is workaround rather than a fix.

PR #16057, opened 2026-02-28
feat(rpmautospec): include building python-rpmautospec

ddstreet:
Add the `python-rpmautospec` package into our build, as we need its functionality in our koji builds.


christopherco:
Looks like #16059 went in already which includes python-rpmautospec. Closing this PR

PR #16061, opened 2026-02-28
workaround(rust-podman-sequoia): fix clang-sys runtime feature unification panic

christopherco:
Enable dlwrap's clang-runtime feature to fix a build failure caused by Cargo feature unification. When bindgen enables clang-sys/runtime, the runtime feature propagates to dlwrap's clang dependency, but without clang-runtime enabled, clang_sys::load() is never called — causing a panic: 'a libclang shared library is not loaded on this thread'.

This matches Fedora f43's approach (rust-dlwrap built with -f clang-runtime).

- Move rust-podman-sequoia from inline components-full.toml to dedicated comp.toml with a patch-add overlay
- Add enable-dlwrap-clang-runtime.patch to set features = ["clang-runtime"] on the dlwrap build-dependency in Cargo.toml

This workaround goes away when we move to a newer version of `podman-sequoia` that contains this type of patch. See upstream patch - https://github.com/ueno/podman-sequoia/commit/9bedf62e6d538c1c28b7e7fbece0113c5a151d18

### DIFF
--- a/base/comps/rust-podman-sequoia/rust-podman-sequoia.comp.toml
+++ b/base/comps/rust-podman-sequoia/rust-podman-sequoia.comp.toml
@@ -1,0 +1,25 @@
+[components.rust-podman-sequoia]
+
+# Fix build failure: clang-sys "runtime" feature unification causes panic
+# in dlwrap's build script because libclang is loaded via dlopen but
+# clang_sys::load() is never called.
+#
+# When bindgen (a transitive build-dep) enables clang-sys/runtime, Cargo
+# feature unification applies it to the shared clang-sys instance used by
+# dlwrap's clang dependency. But without dlwrap's clang-runtime feature,
+# the clang crate doesn't know to call load() first — causing a panic:
+#   "a `libclang` shared library is not loaded on this thread"
+#
+# The clang-runtime feature (added in dlwrap 0.3.9) propagates the runtime
+# feature through clang and clang-sys, ensuring load() is called.
+#
+# Fedora hit the same issue with 0.2.0-4.fc44 mass rebuild and resolved
+# it by bumping to 0.3.2 and adding the equivalent patch to enable building 
+# dlwrap with the clang-runtime feature in the podman-sequoia toml:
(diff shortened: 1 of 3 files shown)


reubeno:
>This workaround goes away when we start using our own dlwrap package as a dependency.

@christopherco Something I didn't quite catch; for stage1 builds we're using F43's dlwrap. Does that not have what we need?

christopherco:
> > This workaround goes away when we start using our own dlwrap package as a dependency.
> 
> @christopherco Something I didn't quite catch; for stage1 builds we're using F43's dlwrap. Does that not have what we need?

Apologies, what I wrote there was actually incorrect. We actually need a newer version of `podman-sequoia` which has this same change. See upstream patch - https://github.com/ueno/podman-sequoia/commit/9bedf62e6d538c1c28b7e7fbece0113c5a151d18

This commit is not present in our rust-podman-sequoia 0.2.0 version, which is the same version that F43 is on.

I have corrected my description comment and the comment in the toml overlay to accurately reflect what is needed to remove the workaround.

base/comps/rust-podman-sequoia/rust-podman-sequoia.comp.toml
```diff
@@ -0,0 +1,23 @@
...
+# When bindgen (a transitive build-dep) enables clang-sys/runtime, Cargo
+# feature unification applies it to the shared clang-sys instance used by
+# dlwrap's clang dependency. But without dlwrap's clang-runtime feature,
+# the clang crate doesn't know to call load() first — causing a panic:
+#   "a `libclang` shared library is not loaded on this thread"
+#
+# The clang-runtime feature (added in dlwrap 0.3.9) propagates the runtime
+# feature through clang and clang-sys, ensuring load() is called.
+#
+# Fedora hit the same issue with 0.2.0-4.fc44 mass rebuild and resolved
```
christopherco, line 16: issue: Reword this part of the comment. We need version 0.3.2 AND the upstream patch to enable the feature clang-runtime in the project's toml. Simply building with rust-dlwrap with -f clang-runtime isn't enough.
F44's upgrade to 0.3.2 includes the same change in it, which is why the builds are succeeding.

PR #16064, opened 2026-03-01
workaround(libspf2): disable doxygen call graphs to fix noarch rpmdiff failure

christopherco:
Add a dedicated comp.toml for libspf2 with a patch-add overlay that sets
CALL_GRAPH=NO in Doxyfile.in, plus a spec-append-lines overlay to apply
the patch in %prep. This eliminates nondeterministic SHORT_NAMES-numbered
call graph files (*_cgraph.map/md5/png) that differ between x86_64 and
aarch64 builders, causing Koji's noarch rpmdiff check to fail.

Root cause: doxygen's convertNameToFile() in util.cpp assigns SHORT_NAMES
numbers via a global first-come-first-served counter (g_usedNamesCount).
The call order is influenced by POSIX readdir() nondeterminism in file
discovery and multi-threaded output generation pipelines, making the
numbering host/architecture-dependent. AZL's overlayfs-backed builders
surface this nondeterminism; Fedora's tmpfs-backed builders do not. This
is an upstream doxygen design issue with no known fix.

The inline [components.libspf2] entry is moved from components-full.toml
to a dedicated base/comps/libspf2/libspf2.comp.toml file to house the
overlay and patch.

This overlay is a TEMPORARY WORKAROUND until the upstream doxygen
issue is addressed and the fix makes it into the distribution's doxygen
package.

### DIFF
--- a/base/comps/libspf2/disable-call-graphs.patch
+++ b/base/comps/libspf2/disable-call-graphs.patch
@@ -1,0 +1,23 @@
+Subject: Disable doxygen call graphs for reproducible noarch builds
+
+Doxygen's SHORT_NAMES numbering is non-deterministic across different
+filesystems/hosts due to readdir() ordering affecting the global
+first-come-first-served counter in convertNameToFile(). This causes
+call graph filenames (*_cgraph.map/md5/png) to differ between
+architectures, breaking Koji's noarch rpmdiff check.
+
+Disable CALL_GRAPH to remove the non-deterministic output files.
+The HTML pages, class/file indexes, and all textual content are
+unaffected.
+
+--- a/Doxyfile.in
++++ b/Doxyfile.in
+@@ -1122,7 +1122,7 @@
+ # So in most cases it will be better to enable call graphs for selected 
+ # functions only using the \callgraph command.
+ 
+-CALL_GRAPH             = YES
++CALL_GRAPH             = NO
+ 
+ # If the GRAPHICAL_HIERARCHY and HAVE_DOT tags are set to YES then doxygen 
+ # will graphical hierarchy of all classes instead of a textual one.
--- a/base/comps/libspf2/libspf2.comp.toml
+++ b/base/comps/libspf2/libspf2.comp.toml
@@ -1,0 +1,51 @@
+[components.libspf2]
+
+# Disable doxygen call graphs to fix noarch rpmdiff failure.
+#
+# Root cause: doxygen's SHORT_NAMES numbering (a00001, a00002, ...) is assigned
+# via a global first-come-first-served counter (g_usedNamesCount in util.cpp's
+# convertNameToFile()). The number depends on the order convertNameToFile() is
+# called, which is influenced by:
+#
+#   1. readDir() in doxygen.cpp iterates directories via DirIterator (dir.cpp),
+#      which wraps ghc::filesystem::directory_iterator — i.e. POSIX readdir().
+#      readdir() order is filesystem/host-dependent and not guaranteed by POSIX.
+#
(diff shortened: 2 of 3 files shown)


christopherco:
Pulling back to draft. It still really isn't clear to me why Fedora's builders don't encounter this issue. Even though doxygen's code/design with SHORT_NAMES might be fragile, I'd like to find out why Fedora's builders are getting deterministic results even when readdir() is non-deterministic by POSIX standard.

christopherco:
@reubeno This might be a builder difference in mock configuration to use tmpfs vs not. Fedora is likely using tmpfs enabled while we are not. Without tmpfs enabled, AZL builder defaults to host environment which is backed by an overlayfs and readdir() ordering likely diverges.

base/comps/libspf2/libspf2.comp.toml
```diff
@@ -0,0 +1,45 @@
...
+# This overlay is a TEMPORARY WORKAROUND until the upstream doxygen issue is
+# addressed and the fix makes it into the distribution's doxygen package.
+
+[[components.libspf2.overlays]]
+type = "spec-search-replace"
+section = "%build"
+description = "Disable doxygen call graphs to fix noarch rpmdiff failure caused by readdir() nondeterminism in SHORT_NAMES numbering"
+regex = '/usr/bin/doxygen'
+replacement = """sed -i -e 's/\\(CALL_GRAPH[[:space:]]*=[[:space:]]*\\)YES/\\1NO/' Doxyfile
+sed -i -e 's/\\(CALLER_GRAPH[[:space:]]*=[[:space:]]*\\)YES/\\1NO/' Doxyfile
```
reubeno: Is Doxyfile something that can be changed with a standard patch? Or is it a loose file that could be updated with a file-based overlay?

reubeno (commented):
Let's see if we can avoid a rexeg here.

base/comps/libspf2/libspf2.comp.toml
christopherco: Will take a look now that I think I've found the underlying difference in the Fedora vs AZL builders that are exposing this problem for only us. https://github.com/microsoft/azurelinux/pull/16064#issuecomment-3978906071

reubeno:
> @reubeno This might be a builder difference in mock configuration to use tmpfs vs not. Fedora is likely using tmpfs enabled while we are not. Without tmpfs enabled, AZL builder defaults to host environment which is backed by an overlayfs and readdir() ordering likely diverges.

That's a great insight. We've seen multiple issues with non-deterministic results influenced by fs iteration. While they're all legitimate robustness and reproducibility issues in the upstream specs / projects we should also try to align with Fedora builder configuration to decrease deltas.

We should follow up with the folks operating our koji instance on this.

base/comps/libspf2/disable-call-graphs.patch
```diff
@@ -0,0 +1,24 @@
+From: Azure Linux Team <azurelinux@microsoft.com>
```
christopherco: issue: hallucination. need to remove

base/comps/libspf2/libspf2.comp.toml
christopherco: Updated to use a patch file instead. Much more readable/maintainable than the regex 😄

base/comps/libspf2/disable-call-graphs.patch
reubeno: Sounds good. Change otherwise looks pretty reasonable.

Mental note: we're seeing a number of specs that explicitly apply each patch; we could handle them better.

reubeno (approved):

PR #16069, opened 2026-03-01
feat: add CD ADO pipeline syncing new sources

PawelWMS:
Adding a new pipeline to sync all sources from the TOML config into the temporary lookaside cache. The pipeline is configured to run on every merge to the `tomls/base/main` branch. Current design very simple: checks all components and updates files missing in the cache.

[Sample pipeline run](https://dev.azure.com/mariner-org/azl/_build/results?buildId=1061080&view=logs&j=47c85f2b-56bd-5f36-a580-1cc953dfe176&t=8fc29e5b-e85f-580f-d1cc-468dfe14fba4).

### DIFF
--- a/scripts/sources-sync/README.md
+++ b/scripts/sources-sync/README.md
@@ -1,13 +1,13 @@
 # Source Sync Scripts
 
-This directory contains scripts for downloading, verifying, and uploading
-component sources to the Azure Blob Storage lookaside cache used by the
-Azure Linux build system.
+This directory contains scripts for downloading and uploading component
+sources to the Azure Blob Storage lookaside cache used by the Azure Linux
+build system.
 
 The entry point is **`sync-sources.py`** — a single script that
-combines the download and upload workflows.  An Azure DevOps (ADO)
-pipeline at [`.azure-pipelines/sync-sources.yml`](../.azure-pipelines/sync-sources.yml)
-automates this on every merge to the `tomls/base/main` branch.
+combines the download and upload workflows. An Azure DevOps (ADO)
+pipeline at [`.azure-pipelines/sync-sources.yml`](../../.azure-pipelines/sync-sources.yml)
+automates this on merges to the `tomls/base/main` branch.
 
 ---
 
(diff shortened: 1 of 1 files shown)


scripts/sources-sync/sync-sources.py
```diff
@@ -0,0 +1,891 @@
...
+
+# ---------------------------------------------------------------------------
+# Component discovery
+# ---------------------------------------------------------------------------
+
+
+def discover_components(component_filter: str) -> list[str]:
+    """Auto-discover components by running ``azldev component list``.
+
+    Runs ``azldev component list -a -O markdown`` and filters lines that
```
reubeno, line 257: I know this is supposed to be quick and dirty -- but why aren't we using `-O json` from `azldev`? I see this and multiple other places where this script is doing text scraping while the programmatic data should just be there.

scripts/sources-sync/README.md
```diff
@@ -0,0 +1,298 @@
...
+```
+python scripts/sync-sources.py [OPTIONS]
+```
+
+| Argument | Type | Required | Default | Description |
+|---|---|---|---|---|
+| `--account-url` | `str` | Yes (unless `--verify-only`) | — | Azure Storage Account URL (e.g. `https://<account>.blob.core.windows.net`). |
+| `--container` | `str` | Yes (unless `--verify-only`) | — | Blob container name to upload into. |
+| `--component-filter` | `str` | No | `"Upstream: fedora"` | Substring filter on `azldev component list -a -O markdown` output to select components. |
+| `--components-file` | `Path` | No | — | Read component names from this file (one per line, blank lines and `#` comments ignored) instead of auto-discovering. Accepts the output of `--failed-output`. |
```
reubeno, line 89: What's this used for?

scripts/sources-sync/sync-sources.py
```diff
@@ -0,0 +1,891 @@
+#!/usr/bin/env python3
```
reubeno, line 1: With all this code, I can't help but asking if it would be have taken as much time to AI-code an extension to `azldev` itself to do what we want?

scripts/sources-sync/sync-sources.py
PawelWMS: You're right, there are many improvements we could do in this cript. My summary of all of them: remove the script along with this pipeline.

This script is a throw-away result of compiling more smaller scripts together, which were just helper tools to get some information (example from your comment: text parsing was easier to explain to Copilot, because it began from an in-line `grep` for some other work I was doing). This whole change is throw-away code to temporarily replace the lack of CT and GitHub functionality we've initially had planned for this cycle to automatically upload sources on approved PRs. With that in mind, I'm not sure if it's worth making any fixes to the script at this point. At least if we want to have something working today. I'm happy to hand this off to Carla as well, but I'm just not sure, if it's worth the time and if we won't suffer more from manual uploads in the meantime.

scripts/sources-sync/README.md
PawelWMS: Not sure what you mean but maybe the selected code block has shifted due to later commits. This is just an explanation of the script arguments. In general, the readme is to help anyone, how has to work with this script in my absence.

scripts/sources-sync/sync-sources.py
PawelWMS: Can't say, didn't really think about making this part of `azldev`. This script is a stitching of 4+ smaller helper scripts, which I used when I didn't yet know, we'll need this lookaside repo maintained for longer. I've got nothing against adding this functionality into `azldev` and simplifying the pipeline but can't really do it anymore before I'm gone.

.azure-pipelines/sync-sources.yml
```diff
@@ -0,0 +1,163 @@
...
+#
+# See scripts/README.md for full documentation.
+# ---------------------------------------------------------------------------
+
+# Trigger controlled by ADO
+trigger: none
+
+# Allow manual runs from the ADO UI with parameter overrides.
+pr: none
+
```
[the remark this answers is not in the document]
PawelWMS: It's already mentioned that the trigger is controlled by ADO.

ddstreet:
Personally, I am a HARD nack on all of this. 1) I don't think upstream tarballs should be automatically added to the lookaside cache, it should be done manually (the manual operation could include use of a script/program but I think some level of manual review of each source tarball upload should be required, even if it's just a manual verification of the upstream tarball URL), and 2) *no tools in this repository*

PR #16070, opened 2026-03-02
workaround(java-25): add java-25-openjdk and java-25-openjdk-portable compon…

binujp:
java25 needs sources supplied by java-portable-sources which is version 25.0.20.0.10 in fed43 repo. Fedora build pulls in this version from fed42 repo. Downgrading java25 to version 25.0.0.0.32 via an older snapshot date takes care of this.

For java25 portable build, there is a bootstrap stage which needs N-1 jdk version, 24 in our case. This is not available in fed43 or fed42 and fedora build pulls this from fed41. We have to fix this to get stage2 build going. Downgrade will not work. Instead we override the bootstrap java version to 25 - self bootstrap, which will be built by stage1.

Below is copilot's explanation of the mechanics of the change.

Add dedicated component definitions for java-25-openjdk-portable and java-25-openjdk, migrating them from inline definitions in components-full.toml to dedicated comp.toml files with overlays to fix two build failures:

1. Missing bootstrap JDK: The upstream Fedora spec uses conditional BuildRequires that resolve to java-latest-openjdk-devel, which does not pull in a specific JDK version. Added an explicit BuildRequires for java-25-openjdk-devel and set buildjdkver to 25.

2. ccache symlink rejection: OpenJDK's autoconf configure script detects /usr/lib64/ccache/gcc as a ccache wrapper symlink and refuses to proceed:

     configure: error: /usr/lib64/ccache/gcc is a symbolic link to
     ccache. This is not supported.

   Added exports for CC=/usr/bin/gcc, CXX=/usr/bin/g++, and GCC=/usr/bin/gcc in the %build section to bypass the ccache wrappers in PATH.

Overlays applied to java-25-openjdk-portable:
- spec-search-replace: buildjdkver 24 -> 25
- spec-add-tag: BuildRequires java-25-openjdk-devel
- spec-prepend-lines: export CC/CXX/GCC to real compiler paths

java-25-openjdk uses a separate Fedora 43 snapshot (2025-07-30) to match the portable sources version produced by java-25-openjdk-portable.

Tested: Built java-25-openjdk-portable-25.0.0.0.32-0.1.ea.azl4 successfully, producing 13 RPMs (runtime, devel, docs, sources, static-libs, and debug variants).

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
+++ b/base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
@@ -17,7 +17,7 @@ regex = '^%global buildjdkver 24$'
 replacement = '%global buildjdkver 25'
 
 [[components.java-25-openjdk-portable.overlays]]
-description = "We need JDK 25 to build sicne that is our bootstrap version"
+description = "We need JDK 25 to build since that is our bootstrap version"
 type = "spec-add-tag"
 tag = "BuildRequires"
 value = "java-25-openjdk-devel"


base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
```diff
@@ -0,0 +1,20 @@
...
+replacement = '%global buildjdkver 25'
+
+[[components.java-25-openjdk-portable.overlays]]
+description = "We need JDK 25 to build sicne that is our bootstrap version"
+type = "spec-add-tag"
+tag = "BuildRequires"
+value = "java-25-openjdk-devel"
+
+[[components.java-25-openjdk-portable.overlays]]
+description = "Set GCC and CC environment variables to /usr/bin/gcc to avoid ccache symlink that OpenJDK configure rejects"
```
ddstreetmicrosoft: This isn't a 'fix', this is a workaround because we enable mock ccache (in `distro/mock/azurelinux-4.0.tpl`). Whether or not we should actually do that (especially for production builds) is a debate for another day, but for this PR, please put this overlay into its own commit and label that as a workaround, not a fix.

base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
```diff
@@ -0,0 +1,20 @@
+[components.java-25-openjdk-portable]
+spec = { type = "upstream", upstream-distro = { name = "fedora", version = "43", snapshot = "2025-07-22T00:00:00-08:00" } }
```
ddstreetmicrosoft, line 11: This package builds fine for me in a local build, once I disable mock ccache. What's the reason we need to use the older version?

base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
ddstreetmicrosoft: this is *definitely* not a fix, this is a workaround because we enable the mock ccache plugin (in `distro/mock/azurelinux-4.0.tpl`). Whether using ccache is appropriate/helpful or not, especially for production builds, is a discussion for another day, but for this please label it as a workaround, not a fix.

ddstreetmicrosoft (commented):
I'm not marking this review as 'request changes' because i don't want to block getting everything building, but I added some comments to explain why I would normally block this PR.

base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
ddstreetmicrosoft: correction - it builds fine, but %check fails, if that's the reason for using the older version, shouldn't we just disable %check for now instead?

binujp: This package will not build in stage2 because we do not build jdk24 in stage1. I am also not sure how it built for you without attempting to pull in jdk24! Can you please check what version of jdk it used to bootstrap or did you disable bootstrap build?

base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
binujp: Yes, will do.

base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
ddstreetmicrosoft: this package doesn't require 24:
https://src.fedoraproject.org/rpms/java-25-openjdk-portable/blob/f43/f/java-25-openjdk-portable.spec#_414

base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
ddstreetmicrosoft: I talked to @reubeno and he thinks we should switch the mock cfg to default to ccache disabled, can you include a commit to do that instead of this workaround?

ddstreetmicrosoft:
I built `java-25-openjdk-portable` locally with only 2 changes:
1) set `config_opts['plugin_conf']['ccache_enable'] = False` in `distro/mock/azurelinux-4.0.tpl`
2) add overlay to fix spec file missing `%changelog` section:
```
[[components.java-25-openjdk-portable.overlays]]
type = "spec-search-replace"
regex = "%autochangelog"
replacement = "%changelog\n%autochangelog"

```

The `java-25-openjdk` package does need to build at the older commit for stage1. After we finish bootstrapping we can (and should) remove that.

ddstreetmicrosoft:
We had some discussion and the issue here is a stage1/stage2 problem:

For stage1, the java-25-openjdk-portable package will build fine (with [changes noted](https://github.com/microsoft/azurelinux/pull/16070#issuecomment-3985806372)), but the java-25-openjdk will not build because the version of the java-25-openjdk-portable pacakge it needs has not been published in f43. So an older version of the java-25-opendjk package must be built in stage1, using the older java-25-openjdk-portable published in f43.

Then in stage2, which uses only packages built in stage1 and does not use the f43 repo packages, the java-25-openjdk-portable package can still build (with noted changes), but the java-25-openjdk package will not build, because the older java-25-openjdk-portable package is not available.

This is the easiest way to get both packages building in both stages, i.e. set both to the older version. After our bootstrap phase we can iterate up to the latest version.

One change I would prefer is to not add the workaround for ccache; if we can merge #16081 first, then this PR can remove the ccache workaround.

ddstreetmicrosoft:
i think your rebase might have got messed up, there's a lot of wrong-hash commits in your pr for recent changes to the tomls/base/main branch

binujp:
Whoops, let me take a look.

base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
binujp: Done.

PR #16076, opened 2026-03-02
refactor: add TOML configs for stage1 vs. stage2

reubeno:
Create separate TOML definitions for azl4-stage1 vs. azl4-stage2:

* Create separate mock config files (only used by local development)
* Add not-enabled-by-default .toml config fragment that can be enbled via `azldev --config-file`

Also adds stage1/stage2 compat RPM with symlink workaround. This allows doing something like the following on a machine with VPN access:

```console
$ azldev comp build azurelinux-stage1-compat --local-repo-with-publish base/out
$ azldev comp build vim --config-file ./distro/stage2.toml --local-repo ./base/out
```


ddstreetmicrosoft:
this does work for me, with the stage2 rpms pulling their build env and dep pkgs from the azl koji repo.

however i think this is only applicable to our current bootstrap situation; it should be removed after we have a stable koji repo, right (i.e. setup to use only that stable koji repo)?

PR #16084, opened 2026-03-02
chore: move check disablements for initial bringup into a component config group

tobiasb-ms:
Moved all `check` disables with the reason `Disabling checks for initial set of failures.` into a config group so we have them all in known location and manage them more easily.

### DIFF
--- a/external/schemas/azldev.schema.json
+++ b/external/schemas/azldev.schema.json
@@ -96,6 +96,19 @@
     },
     "ComponentGroupConfig": {
       "properties": {
+        "description": {
+          "type": "string",
+          "title": "Description",
+          "description": "Description of this component group"
+        },
+        "components": {
+          "items": {
+            "type": "string"
+          },
+          "type": "array",
+          "title": "Components",
+          "description": "List of component names that are members of this group"
+        },
         "specs": {
           "items": {
             "type": "string",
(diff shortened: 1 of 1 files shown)


base/comps/components.toml
```diff
@@ -43,6 +44,7 @@ includes = ["**/*.comp.toml", "components-full.toml"]
 [components.dialog]
 [components.diffutils]
 [components.ding-libs]
+[components.dnf]
```
reubeno: Did all of these come from `components.toml` vs. `components-full.toml`? For example, `dnf` isn't actually something we want in the distro even though it got pulled in.

tobiasb-ms: Well spotted. I think I have them properly sorted into the correct files now.

tobiasb-ms:
> Did you confirm at least 1 or 2 of these build via CT/koji? Just want to doubly make sure that koji has a new enough version of azldev and it will work.

I confirmed the version of `azldev` on the server, but you're right -- I'll kick off a few packages and confirm.

tobiasb-ms:
> > Did you confirm at least 1 or 2 of these build via CT/koji? Just want to doubly make sure that koji has a new enough version of azldev and it will work.
> 
> I confirmed the version of `azldev` on the server, but you're right -- I'll kick off a few packages and confirm.

Kicked off a few packages with ones that enable and disable the check section. All of them succeeded and logs look like I want them to.

PR #16117, opened 2026-03-04
chore: update repo file for alpha-1 publish locations

reubeno:
Updates default `.repo` file:

* If the dist tag contains no date in it, uses a hard-coded base URI that does not (yet) exist.
* If the dist tag contains a date in it (e.g., `.azl4~bootstrap.20260303`) extracts out the date and forms a URI to our daily publish location.

Assumes that the publish location will look something like:

`https://stcontroltowerdevjwisitg.blob.core.windows.net/daily-repo-non-prod/20260303`


ddstreetmicrosoft (approved):
LGTM

PR #16122, opened 2026-03-05
feat(kernel): Drop in custom kernel configs for x86 and arm64

rlmenge:
Replace the AZL kernel configs with Ubuntu 6.17.0-1004-azure kernel configs as the new base for both x86_64 and aarch64, then apply minimal AZL-specific customizations on top

New files:
- 6.18-x86_64-azl.config — Ubuntu 6.17 Azure x86_64 config, rebased via olddefconfig against 6.18.5.1, with AZL customizations
- 6.18-aarch64-azl.config — Ubuntu 6.17 Azure ARM64 config, rebased via olddefconfig against 6.18.5.1 on native aarch64, with AZL customizations
- azurelinux-ca-20230216.pem — AZL signing CA certificate for CONFIG_SYSTEM_TRUSTED_KEYS

Modified:
- kernel.comp.toml — New overlays to inject configs and cert into the Fedora kernel spec. **Note because the append logic places the new code right before teh `%build` there is an awkward format issue where the `# build` section comment is above the new appended lines **
- The overlays let the full Fedora config pipeline run first (generate_all_configs.sh, merge.py, process_configs.sh), then append to %prep to:
1. Copy mariner.pem into the kernel source tree
2. Remove all Fedora-generated configs
3. Copy the AZL config for the build arch only (guarded by %ifarch)
4. Re-run process_configs.sh to validate
5. A %ifnarch noarch %nobuildarches guard ensures this block is skipped during the noarch bootstrap sub-build.

Testing
x86_64: Built successfully (27 RPMs, ~55 min on 16 threads). Smoke tested in mock chroot — 1928 modules, vmlinuz PE32+, mariner.pem embedded.
aarch64: Built successfully on native ARM64 VM (27 RPMs, ~32 min on 96 core). Smoke tested in mock chroot — 7310 modules, vmlinuz PE32+ ARM64 EFI executable.
Both booted in QEMU vm
Built : taskID=315107

### DIFF
--- a/base/comps/kernel/kernel.comp.toml
+++ b/base/comps/kernel/kernel.comp.toml
@@ -135,7 +135,7 @@ type = "spec-append-lines"
 section = "%prep"
 lines = [
     "# AZL: Copy mariner.pem signing CA cert into kernel source tree",
-    "%ifnarch %nobuildarches",
+    "%ifnarch noarch %nobuildarches",
     "cp %{SOURCE5002} linux-%{KVERREL}/certs/mariner.pem",
     "# AZL: Overwrite Fedora-generated configs with complete AZL kernel configs",
     '%{log_msg "AZL: Overwrite configs with AZL kernel configs"}',


base/comps/kernel/kernel.comp.toml
```diff
@@ -87,3 +87,70 @@ description = "Disable redhat patch application — not applicable to AzureLinux
...
+description = "Register AZL signing CA certificate as Source5002"
+type = "spec-insert-tag"
+tag = "Source5002"
+value = "azurelinux-ca-20230216.pem"
+
+[[components.kernel.overlays]]
+description = "Overwrite Fedora-generated configs with AZL configs and re-validate with process_configs.sh at the end of %prep — lets the entire Fedora config pipeline run first, then overwrites with our complete configs. Only copies the config matching the build arch to avoid cross-arch make failures."
+type = "spec-append-lines"
+section = "%prep"
+lines = [
```
binujp, line 136: thoughts: Is there a provision to move large multiline overlays like this into a file and include the file instead? I suppose availability of context for the overlay is the same.

binujp:
Looks good to me. Thanks for the comments, they make intention behind the overlays clear. The config does not look heavily customized for azure from a drivers perspective. I am assuming this is deliberate.

ddstreetmicrosoft (approved):
LGTM, though at some point in the future it might be easier to just fork the spec, etc. if the number of toml overlays gets too large and/or complex.

PR #16135, opened 2026-03-06
fix(vm-image): Add Azure-specific configuration from Fedora Cloud SIG

ddstreet:
There are some Azure-cloud-specific configuration settings needed to work with the specifics of the Azure cloud; the Fedora Cloud SIG has already done the work to determine the settings we want to use.

The specific changes include:

Enable openssh-server's keepalive; otherwise idle ssh connections appear to hang and/or are killed.

Enable use of Azure's PTP time source.

Add configuration to correctly handle Azure's non-standard way of providing accelerated networking, using two separate interfaces to represent a single SRIOV interface.

### DIFF
--- a/base/images/vm-base/config.sh
+++ b/base/images/vm-base/config.sh
@@ -1,0 +1,22 @@
+
+# Azure-specific configuration, from Fedora Cloud SIG, f43 branch
+# https://pagure.io/fedora-kiwi-descriptions/blob/d8874b9ee71851b06aaecf43e492d1a6a4f3f164/f/config.sh#_164
+
+cat > /etc/ssh/sshd_config.d/50-client-alive-interval.conf << EOF
+ClientAliveInterval 120
+EOF
+
+cat >> /etc/chrony.conf << EOF
+# Azure's virtual time source:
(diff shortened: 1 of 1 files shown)


ddstreet:
> Add configuration to correctly handle Azure's non-standard way of providing accelerated networking, using two separate interfaces to represent a single SRIOV interface

This actually isn't needed for our base image, as it defaults to using systemd-networkd instead of NetworkManager; the workaround here is only for NetworkManager. With systemd-networkd, instances with accelerated networking do appear to work fine.

reubeno:
@ddstreet If we do need some custom configs, would we not want to add an Azure-specific RPM that adds separate conf.d conf fragments rather than post-installing into the image? I've seen image post-install scripts get significantly abused in the past; they're harder to track, update, share, etc. too.

ddstreetmicrosoft:
> @ddstreet If we do need some custom configs, would we not want to add an Azure-specific RPM that adds separate conf.d conf fragments rather than post-installing into the image? I've seen image post-install scripts get significantly abused in the past; they're harder to track, update, share, etc. too.

that's probably better, yeah, i was just duplicating the approach that the fedora cloud sig took in their image generation.

This should probably go into an identity sub-package like `azurelinux-release-identity-cloud-azure` or something similar, but I'll leave this PR open as draft until we get that set up

PR #16170, opened 2026-03-11
[Manual Cherry-Pick] [AutoPR- Security] Patch cmake for CVE-2025-14524, CVE-2025-10966 [ME…

Kanishk-Bansal:
Manual Cherry Pick of https://github.com/microsoft/azurelinux/pull/15557


SPECS/cmake/cmake.spec
```diff
@@ -41,6 +41,8 @@ Patch26:        CVE-2025-5916.patch
 Patch27:        CVE-2025-5917.patch
 Patch28:        CVE-2025-5918.patch
 Patch29:        CVE-2025-14017.patch
+Patch30:        CVE-2025-10966.patch
```
jslobodzian, line 44: I'm not sure we need this change.  There is a buildrequires against the system curl.  Can we confirm this is really needed?

jslobodzian: nvm,  I see that the vendored curl is included.

PR #16176, opened 2026-03-11
Updated schema and disabled origins for fedora override

Tonisal-byte:
These changes introduce a feature added to `azldev` to explicitly disable preparing sources from outside the configured lookaside caches. You can disable origins by passing in this flag to a distro definition as shown below. This also requires that all components with `source-files` include a hash and hash-type in their definition. 

```toml
[distros.some-distro]
dist-git-base-uri = ...
lookaside-base-uri = ...
disable-origins = true
```

### DIFF
--- a/external/schemas/azldev.schema.json
+++ b/external/schemas/azldev.schema.json
@@ -57,6 +57,43 @@
           "$ref": "#/$defs/CheckConfig",
           "title": "Check configuration",
           "description": "Configuration for the %check section"
+        },
+        "failure": {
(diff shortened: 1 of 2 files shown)


reubeno:
@Tonisal-byte Is this still a change that we should be shepherding in?

Tonisal-byte:
> @Tonisal-byte Is this still a change that we should be shepherding in?

We don't really need this anymore since we will start moving to building from specs instead of TOMLs :) Closing now

PR #16184, opened 2026-03-13
[WIP] Draft proposal for new image tomls

realsdx:
This PR proposes extending Azure Linux image definitions with delivery metadata so they can be consumed by TEE.

As part of that, `images.toml` is changed to a thin include based registry and image definitions are moved into per-image `*.image.toml` files alongside their KIWI definitions. The new manifest shape introduces fields such as name, distro, delivery, and variants so image delivery mechanisms and constraints can be defined closer to the image itself.

Scope: Introduce image delivery definitions 
`*` means this field is intentionally unresolved in the manifest and must be filled in by a later delivery, publish, or fulfillment step. The key is still declared here so downstream tooling knows the field exists and can provide or validate it.

Note: It has not been verified with azldev. Runtime support in azldev is still needed for the new image fields.


base/images/container-base/container-base.image.toml
```diff
@@ -0,0 +1,4 @@
+[images.container-base]
+description = "Container Base Image"
+distro = "azurelinux"
```
reubeno, line 3: Why is `distro` identified here?

base/images/vm-base/vm-base.image.toml
```diff
@@ -0,0 +1,29 @@
+[images.vm-base]
+description = "VM Base Image"
+distro = "azurelinux"
```
reubeno, line 3: Ditto regarding `distro` as well.

base/images/vm-base/vm-base.image.toml
```diff
@@ -0,0 +1,29 @@
...
+definition = { type = "kiwi", path = "vm-base.kiwi" }
+
+[images.vm-base.delivery.marketplace]
+publisher = "MicrosoftCBLMariner"
+offer = "azure-linux-4"
+sku_prefix = "azure-linux-4"
+version = "*"
+
+[images.vm-base.delivery.vhd]
+vhd_path = "*"
```
reubeno, line 13: What's the `*` for? What does this mean?

base/images/vm-base/vm-base.image.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[images.vm-base.delivery.shared_gallery]
```
reubeno, line 15: I've also been working on drafting what publishing info looks like. Since these are going to need to look different across branches of the product, we're going to need to place that in the distro TOMLs.

Consider something like this:

```toml
[distros.azurelinux.versions.azl4-dev-stage1.outputs]
source-lookaside-caches = ["azl4-dev-lookaside-cache"]
package-repos = ["azl4-alpha1-rpms-base"]
image-stores = ["azl4-alpha1-acr", "azl4-alpha1-acg"]

[[resources.alpha1-acg]]
publish-labels = ["base-vm"]
interface-type = "azure-compute-gallery"
backend = "azure-compute-gallery"
resource-urn = "/subscriptions/e4ab81f8-030f-4593-a8f2-3ea2c7630a19/resourceGroups/azl-preview-publishing/providers/Microsoft.Compute/galleries/azlpubStagingGalleryoxz2o4gw"
images = [
    { name = "AzureLinuxAlpha1-x64", arch = "x86_64" },
    { name = "AzureLinuxAlpha1-arm64", arch = "aarch64" }
]
```

The image itself just needs to indicate that it should be published to any targets with the `base-vm` label.

What do you think?

base/images/vm-base/vm-base.image.toml
```diff
@@ -0,0 +1,29 @@
...
+subscription_id = "*"
+resource_group_name = "*"
+image_gallery = "*"
+image_definition = "*"
+image_version = "*"
+
+[images.vm-base.variants]
+architectures = [
+    "amd64",
```
reubeno, line 24: Minor note: we should use x86_64 and aarch64 as the standard arch names.

base/images/vm-base/vm-base.image.toml
```diff
@@ -0,0 +1,29 @@
...
+    "arm64",
+]
+hyperv_generations = [
```
reubeno, line 27: I'd expect gen1 and gen2 to be different kiwi image profiles. How would you anticipate this metadata be used? What's the reasons for/against the gen1 vs. gen2 images just being modeled as different `images`?

base/images/vm-base/vm-base.image.toml
```diff
@@ -0,0 +1,29 @@
+[images.vm-base]
+description = "VM Base Image"
+distro = "azurelinux"
+definition = { type = "kiwi", path = "vm-base.kiwi" }
+
+[images.vm-base.delivery.marketplace]
```
reubeno, line 6: See below; for any publishing targets that need to be branch-differentiated/specific, the metadata needs to be split: some branch-specific distro metadata around publishing targets, and some per-artifact identification of which artifacts should be published.

base/images/vm-base/vm-base.image.toml
realsdx: `*` means value is not hardcoded will be determined at runtime, as this will be a blob url with sas token.

base/images/container-base/container-base.image.toml
realsdx: we needed the distro metadata cause in future we want to support other distros like `flatcar`. If this is not good place we can remove, currently not used.

base/images/vm-base/vm-base.image.toml
realsdx: So, a separate toml with all publishing info? what are image stores? what is `backend` here?

base/images/vm-base/vm-base.image.toml
realsdx: If gen1/gen2 have different kiwi files, we need create different tomls for them in this model with only one gen hardcoded in it

base/images/vm-base/vm-base.image.toml
realsdx: I was thinking from a perspective of test/image owners, having the image definition and its related delivery/publishing methods at the same place would help with code/config locality IMO, what do you think?

realsdx:
Another way if we follow https://github.com/microsoft/azure-linux-dev-tools/pull/38
```toml
[images.vm-base]
description = "Azure Linux 4 VM base image"
distro = "azurelinux"
definition = { type = "kiwi", path = "vm-base.kiwi" }

[images.vm-base.publish.marketplace]
publisher = "MicrosoftCBLMariner"
offer = "azure-linux-4"
sku_prefix = "azure-linux-4"
version = "*"

[images.vm-base.publish.vhd]
vhd_path = "*"

[images.vm-base.publish.shared_gallery]
subscription_id = "*"
resource_group_name = "*"
image_gallery = "*"
image_definition = "Azl4AMD64"
image_version = "*"
replica_regions = ["*"]

[images.vm-base.publish.shared_gallery.capability]
accelerated_networking = false
hibernation_support = false
high_perf_nvme = false
security_type = "TrustedLaunchSupported"

[images.vm-base.variants]
architectures = ["amd64", "arm64"]
hyperv_generations = ["gen2"]
```

and on the CT side, something like this:
```toml
[profiles.dev.shared_gallery]
subscription_id = "00000000-0000-0000-0000-000000000000"
resource_group_name = "rg-azlinux-dev"
image_gallery = "AzureLinuxDev"
replica_regions = ["eastus2"]
storage_account_type = "Standard_LRS"
```
---
But if we do similar to channel based thing
```toml
[images.vm-base]
description = "Azure Linux 4 VM base image"
distro = "azurelinux"
definition = { type = "kiwi", path = "vm-base.kiwi" }

[images.vm-base.publish]
channel = "vm-base-sig"
```

In CT:
```toml
[publish.channels.vm-base-sig]
subscription_id = "00000000-0000-0000-0000-000000000000"
resource_group_name = "rg-azlinux-dev"
image_gallery = "AzureLinuxDev"
replica_regions = ["eastus2"]
storage_account_type = "Standard_LRS"

[publish.channels.vm-base-sig.shared_gallery.capability]
accelerated_networking = false
hibernation_support = false
high_perf_nvme = false
security_type = "TrustedLaunchSupported"
```
But this way we are splitting(loosing) the actual image capability metadata from the azurelinux repo keeping some of it in CT, which might not be ideal. What do you think?

reubeno:
> But this way we are splitting(loosing) the actual image capability metadata from the azurelinux repo keeping some of it in CT, which might not be ideal. What do you think?

I think this separation is by design. We want to be able to define images and identify their publishing requirements, separate from the physical details that help CT publish them to the appropriate destination, given the context of the target branch, etc.

realsdx:
> > But this way we are splitting(loosing) the actual image capability metadata from the azurelinux repo keeping some of it in CT, which might not be ideal. What do you think?
> 
> I think this separation is by design. We want to be able to define images and identify their publishing requirements, separate from the physical details that help CT publish them to the appropriate destination, given the context of the target branch, etc.

If this separation is by design, then metadata itself might not be enough to serve as a contract between services. In that case, we need to treat metadata as "hints" and resolve those hints in CT, and CT will send them (resolved hints) to downstream services. The contract between those services and CT will no longer be driven by this metadata here. We need to define a separate contract for those services.

base/comps/components-vm-tests.toml
```diff
@@ -0,0 +1,13 @@
...
+    "verify_docker_compose_wordpress_app",
+    "verify_docker_dotnet31_app",
+    "verify_docker_dotnet50_app",
+    "verify_docker_java_app",
+    "verify_docker_python_app",
+    "verify_docker_seccomp_profile"
+]
+
+[components.containerd.vm-tests]
+test_lables = ["container"]
```
realsdx, line 13: For package-to-test mapping, we will prefer a label-based approach, similar to package publish labels.
Each package will be tagged with at least one label, with the option to also pin specific individual test cases.
Labels will be resolved on TEE's end, where the label-to-tests (group of tests) mapping will be defined.

As TEE is the sole owner of all test-related data and with all the context, If CT sends a test label that TEE cannot run on the specified image, it will return an error. In future, we can extend TEE to suggest the most suitable image/test-suite for the given tests, allowing CT/VS to select the appropriate one.

PR #16214, opened 2026-03-17
first-cut overlays for  nvidia kmod and driver

ellie-di:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/base/comps/buildsys-build-azurelinux/buildsys-build-azurelinux-kerneldevpkgs.spec
+++ b/base/comps/buildsys-build-azurelinux/buildsys-build-azurelinux-kerneldevpkgs.spec
@@ -1,0 +1,105 @@
+Name:           buildsys-build-azurelinux-kerneldevpkgs
+Version:        4.0
+Release:        7%{?dist}
+Summary:        Kernel dev package list for kmod builds on Azure Linux
+License:        MIT
+BuildArch:      noarch
+
+# The Azure Linux kernel version to build kernel modules against.
+# Update these macros when the kernel component version changes.
+%define azl_kernel_version 6.18.5.1
+%define azl_kernel_release 3%{?dist}
+
+Provides:       buildsys-build-azurelinux-kerneldevpkgs-current-x86_64
+Provides:       buildsys-build-azurelinux-kerneldevpkgs-current-aarch64
+Provides:       buildsys-build-azurelinux-kerneldevpkgs-akmod-x86_64
+Provides:       buildsys-build-azurelinux-kerneldevpkgs-akmod-aarch64
+
+Requires:       kernel-devel = %{azl_kernel_version}-%{azl_kernel_release}
+
+%description
+Meta-package that provides the kerneldevpkgs-current data file and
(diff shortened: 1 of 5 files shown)


ddstreet (commented):
i don't think you should use anything from rpmfusion, so i don't have much specifically to comment on here, other than to say don't do any of this.

is there something you can't figure out how to do when creating your own simpler spec file?

ellie-di:
> i don't think you should use anything from rpmfusion, so i don't have much specifically to comment on here, other than to say don't do any of this.
> 
> is there something you can't figure out how to do when creating your own simpler spec file?

can you elaborate more on "i don't think you should use anything from rpmfusion"? 
I've been looking into the upstream approach used by fedora and rpmfusion that I came across the [Kmods2 ](https://rpmfusion.org/Packaging/KernelModules/Kmods2#kmod_template) standard, and I found it interesting. I know that I could add our own version of the spec, but the goal of this PoC was to see how the upstream does it.
I think adopting a standard like Kmod2 for OOT kernel module development would improve the readability and maintainability of our spec files, particularly if we’re not using rpmfusion specs directly.

ddstreet:
> the goal of this PoC was to see how the upstream does it.

rpmfusion is *not* upstream Fedora.

> I think adopting a standard like Kmod2 for OOT kernel module development would improve the readability and maintainability of our spec files,

I really don't see how that is true. the rpmfusion-specific kmod2 'standard' is *extremely* complicated, and *still* requires you to explicitly maintain the exact kernel verion(s) you want to build the OOT module for. It appears to have been created to fill numerous different needs, as [they said themselves in the faq](https://rpmfusion.org/Packaging/KernelModules/Kmods2#This_standard_is_stupid.), and almost certainly has far, far more flexibility/complexity than we actually need. It probably makes sense in the context of needing a uniform way to prepare and provide OOT kernel modules, for wide variety of kernel flavors, from multiple distros, through the rpmfusion repo; but we aren't building or delivering our OOT kernel modules through rpmfusion, and we have only one distro and (currently) only one kernel flavor to build/deliver for.

However, I don't have to maintain whatever we eventually use, so my opinion is *solely* a suggestion. If kmod2 makes sense to you, I have no doubt that it can be made to work.

ellie-di:
Thanks for the reply. I get your position better now. 

> we have only one distro and (currently) only one kernel flavor to build/deliver for.
> 

At the beginning of a project, it’s easy to say our system is simpler and that we should keep things simple. But in practice, that often turns into something that doesn’t scale well down the road. For example, when kernel‑hwe was introduced in azl3.0, we had to figure out how to correctly query the kernel version, and we ended up hardcoding it in the spec. I’m not saying we need to follow that standard exactly, but having a scalable, policy‑driven model for developing our kmod specs would help avoid these kinds of issues.

PR #16254, opened 2026-03-23
dbus: enable user sessions

ddstreet:
While most distros have migrated over to dbus-broker, we still provide dbus-daemon, and we do not enable user sessions, so each user login does not have any dbus that can be used to communicate with the user session's systemd manager. This leads to frequent error messages while upgrading packages, and at other times, similar to:

Failed to connect to bus: No medium found
Reload daemon failed: Transport endpoint is not connected

Those errors also mean that while the systemd system manager (pid1) can be reloaded or reexeced, systemd user managers will not be reexeced on a package upgrade, nor can any other package that provides user session services cause their user services to reload/reexec on upgrade. This may leave user processes running old code after some package updates.

Note that the user can still communicate with the user session systemd manager, via systemctl, because it will use the private systemd bus. However, the root user (including scripts run during package updates) cannot connect to any user's private systemd bus and so must use the public user dbus, which is not provided by our dbus package currently.

This changes the dbus build to include support for user sessions.

###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

enable user dbus support

###### Change Log  

enables user dbus support

n

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/SPECS/dbus/dbus.spec
+++ b/SPECS/dbus/dbus.spec
@@ -2,7 +2,7 @@
 Summary:        DBus for systemd
 Name:           dbus
 Version:        1.15.8
-Release:        3%{?dist}
+Release:        4%{?dist}
 License:        GPLv2+ OR AFL
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
@@ -51,6 +51,7 @@ autoreconf -i
     --docdir=%{_versioneddocdir}  \
     --enable-libaudit=yes \
     --enable-selinux=yes \
+    --enable-user-session=yes \
     --runstatedir=/run \
     --with-console-auth-dir=/run/console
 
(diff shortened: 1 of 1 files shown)


corvus-callidus:
Before this goes in we should do a sweep and ensure that we are aware of any and patch any CVEs this might expose us to. I have not looked yet myself, but could easily see there being some CVEs that we disputed because we didn't turn on user sessions.

ddstreetmicrosoft:
> Before this goes in we should do a sweep and ensure that we are aware of any and patch any CVEs this might expose us to. I have not looked yet myself, but could easily see there being some CVEs that we disputed because we didn't turn on user sessions.

I checked astrolabe, and all the dbus CVEs were already fixed in the version of dbus we have in azl3. I didn't find any CVEs that applied to the version of dbus currently in azl3.

ddstreetmicrosoft:
closing this as we will keep azl3 as-is for this; user dbus sessions will be enabled for azl4.

PR #16277, opened 2026-03-25
refactor(rust-nix): remove 4-part kernel version workaround

rlmenge:
Now that the kernel uses a 3-part version scheme (6.18.5 + kextraversion in Release), uname -r produces standard semver-compatible output (e.g. 6.18.5-1.4.azl4). The patch that truncated 4-part versions for the semver crate in require_kernel_version! is no longer needed.

The inotify test skip overlay is retained — it addresses a separate containerized build environment limitation (ENOSPC on inotify_init).

Kernel refactor: https://github.com/microsoft/azurelinux/pull/16253
original rust-nix pr: https://github.com/microsoft/azurelinux/pull/16235

TaskID= 770425


christopherco:
We won't be able to remove the workaround until we move our koji builders onto a kernel that has standard 3-part semver.

PR #16443, opened 2026-04-02
fix(python-beartype): pin python-beartype to upstream commit 9c277e0

liunan-ms:
This PR creates dedicated comp.toml for `python-beartype` pinned to f43 upstream commit 9c277e0bb3a1 which fixes build/test issues with Python 3.14.

Validation: waiting for `python-beartype` 0.22.9 tarball is available in koji

### DIFF
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -3883,7 +3883,6 @@
 [components.python-backoff]
 [components.python-banal]
 [components.python-bcrypt]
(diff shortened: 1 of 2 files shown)


christopherco (changes requested):
Sources need to still be uploaded to lookaside cache. Blocking PR until that operation is done

christopherco (approved):
Sources are now uploaded and the scratch build passed

PR #16444, opened 2026-04-02
DO NOT CHECK IN: sample checked-in sources

PawelWMS:
I've checked in all source files, which were NOT listed in the `sources` file, which were produced by `azldev prepare-sources`. The `sources` for `kernel-headers` is a weird one IMO but maybe I don't understand the logic behind `prepare-sources`.


base/comps/kernel-headers/sources
```diff
@@ -1 +1 @@
-SHA512 (kernel-6.18.5.1.tar.gz) = 0d8c755e02857704eb4c4396657ad30b22b474296233b8681365aeaa2fc16926afe31f551e80f565e0d73318f2b7941051ec44f85ccbabaa28ba5a56022f9f4a
+SHA512 (kernel-headers-6.18.3.tar.xz) = b2c285b37be51e314cfafa82447e9ad99a6f488fc09a5f5d3f0c150e2d9fb2234979d26947cd4c5d86be4bc4c91a3a46d815eb53fefa26bb9abee4ba3f65e5c3
```
PawelWMS, line 1: That's an interesting one. Looks like `sources` were already checked into this repo and had the file name and hash for the source file defined in the `source-files` section of the TOML. Not sure why we needed to define the same source info twice this way.

More importantly, `azldev prepare-sources` seems to **NOT** modify the `sources` file according to the TOML config and leaves the upstream original around. Not sure if a bug or something we may need to adjust, if we want to auto-generate post-overlay contents to be checked into the repo.

dmcilvaney:
Can we drop the generated files into `./SPECS/<pkg>/` That will likely be the final location, separate from the tomls.

The source file also looks weird to me...

PR #16458, opened 2026-04-03
fix: address build failures due to missing dependencies

trungams:
Add version pins via commit and workaround for some components that failed due to missing dependencies in build root. This change should be rebase-and-merged.


christopherco (approved):
:shipit:

PR #16461, opened 2026-04-04
workaround rust-nix* test failures due to 4-part kernel version issue

christopherco:
Azure Linux 3.0 kernels, used on our current Koji builders, have 4-part
version strings (e.g. 6.6.121.1-1.azl3) which the semver crate's
Version::parse() rejects. Two tests (gro, gso) panic on unwrap of the
parse error.

Apply the same fix-4part-kernel-version.patch already used by rust-nix:
truncate to 3 numeric components before parsing so the extra part
becomes a pre-release suffix (which is cleared immediately after).

Apply to rust-nix0.23, rust-nix0.28, rust-nix0.29, rust-nix0.30.

This workaround can be undone when the Koji builders move to a
kernel with 3-part version strings, although the current logic will still
continue to function as intended with a standard 3-part kernel version
string.

### DIFF
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -102,7 +102,6 @@
 [components.'rust-nix0.24']
 [components.'rust-nix0.26']
 [components.'rust-nix0.27']
-[components.'rust-nix0.29']
 [components.'rust-object0.36']
 [components.'rust-object0.37']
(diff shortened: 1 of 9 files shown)


binujp:
Looks good to me. We probably will ever know why they did not use a regular expression.

binujp (approved):

PR #16462, opened 2026-04-05
feat(qemu): disable Xen support

christopherco:
Azure Linux targets Hyper-V/KVM, not Xen. Since our builders have the %fedora macro set, the upstream spec enables Xen on x86_64/aarch64 by default, pulling in xen-devel at build time and libxenstore.so.4 at runtime.

Override have_xen back to 0 via overlay. Verified: build succeeds and no output RPMs carry xen dependencies.

Part of fixing: [workitem](https://dev.azure.com/mariner-org/mariner/_workitems/edit/18553)

### DIFF
--- a/base/comps/qemu/qemu.comp.toml
+++ b/base/comps/qemu/qemu.comp.toml
@@ -1,0 +1,10 @@
+[components.qemu]
+
+# Disable Xen support — not applicable to Azure Linux's Hyper-V/KVM environment.
+# The upstream spec enables Xen when %fedora is set (which it is on our builders),
+# so we must explicitly override have_xen back to 0.
+[[components.qemu.overlays]]
+description = "Disable Xen support - not needed for Azure Linux (Hyper-V/KVM only)"
+type = "spec-search-replace"
+regex = '%global have_xen 1'
+replacement = '%global have_xen 0'
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -4739,7 +4739,6 @@
 [components.qcustomplot]
 [components.qdbm]
 [components.qdox]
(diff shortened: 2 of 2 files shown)


base/comps/qemu/qemu.comp.toml
```diff
@@ -0,0 +1,10 @@
+[components.qemu]
+
+# Disable Xen support — not applicable to Azure Linux's Hyper-V/KVM environment.
+# The upstream spec enables Xen when %fedora is set (which it is on our builders),
```
reubeno, line 4: Did you look at whether we should be disabling `fedora` for this spec? What other features are we getting that we wouldn't if we looked more like RHEL?

christopherco: I didn't for this one, but it is a good analysis that can be done. I did a similar evaluation on `edk2`, but ended up going with `fedora`'s version instead of `rhel`'s since there were some valuable parts that the `fedora` `edk2` build (namely 4MB OVMF support and IGVM). I'll do a similar analysis for `qemu`

christopherco:
Bringing back to draft while I analyze `fedora` vs `rhel` qemu builds

christopherco:
I've done the review and there's quite a few differences. While we debate the feature differences, I'm re-opening this PR to get `qemu` building again. Will make a follow-on PR with the result of our feature discussions.

ddstreetmicrosoft (approved):
lgtm, and i think leaving `fedora` and selectively disabling xen is better than changing to `rhel` - i think we generally lean more towards what comes with `fedora` for this package, instead of the mostly-feature-disabled build for `rhel`.

PR #16479, opened 2026-04-07
docs: update overlay instructions with lessons from xen removal

christopherco:
Add spec-remove-section to overlay types table and mark it as a non-exhaustive quick reference pointing to upstream docs at azure-linux-dev-tools as the canonical source.

Add new rows to the choosing-right-type table:
- build.without for bcond toggles
- spec-remove-section for removing subpackages
- file-search-replace for source config files

Add pitfalls learned from practical overlay work:
- spec-search-replace cannot match section tags
- file-search-replace targets source files, not specs
- -n packages need unexpanded macros in package field
- Blank lines from empty replacements break continuations
- Replacement string escaping differs from regex

Add builder environment note to copilot-instructions.md documenting that %fedora is set on builders and its implications for feature gating.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -37,6 +37,8 @@ azldev.toml                          # Root config — includes distro/ and base
 
 **TOML include hierarchy**: `azldev.toml` → `distro/distro.toml` + `base/project.toml` → `base/comps/components.toml` → `**/*.comp.toml` (stitched into single namespace).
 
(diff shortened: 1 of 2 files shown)


.github/copilot-instructions.md
```diff
@@ -37,6 +37,8 @@ azldev.toml                          # Root config — includes distro/ and base
 
 **TOML include hierarchy**: `azldev.toml` → `distro/distro.toml` + `base/project.toml` → `base/comps/components.toml` → `**/*.comp.toml` (stitched into single namespace).
 
+**Builder environment**: Our Koji builders run Fedora, so the `%fedora` macro IS defined during builds. Upstream Fedora specs that gate features on `%{defined fedora}` or `0%{?fedora}` will have those features **enabled** unless explicitly overridden via overlays or `build.defines`. This is the most common reason packages produce unwanted subpackages (e.g., Xen, Flatpak-specific features). When investigating unwanted build output, always check for `%fedora`-gated conditionals in the spec first.
```
reubeno, line 40: Not quite true that it's defined because we run Fedora in the builders. (Will follow up on this offline.)

christopherco: ACK - discussed offline. Will amend this instruction to be more precise about only the `fedora` macro being set in the build environment, without stating that we run Fedora in the builder machines, because that's not true.

PR #16502, opened 2026-04-07
fix(rust-clap): pin to version 4.6.0

trungams:
f43 repos only provide crate(clap_builder) 4.6.0. We need to move the version forward to build rust-clap


binujp:
Looks good, thanks Trung. We may be able to do a publish build since the version was upgraded.

binujp (approved):

PR #16503, opened 2026-04-07
fix(erlang-rebar3): pin erlang-rebar3 to 3.27.0 for hex_core 0.15+ compatx_core 0.15+

liunan-ms:
Pin to f43 commit 24837f9 which updates to `erlang-rebar3 3.27.0`, fixes the `hex_core 0.15+` API incompatibility (request_to_file/6), and has bootstrap mode enabled so rebar3 can build itself without needing the old broken 3.26.0 in the build tag.
Validation: waiting for erlang-rebar3 3.27.0 tarball uploaded to azl lookaside cache.

### DIFF
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -612,7 +612,6 @@
 [components.erlang-proper]
 [components.erlang-providers]
 [components.erlang-rebar3-gpb]
-[components.erlang-rebar3]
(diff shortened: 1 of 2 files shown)


binujp (approved):
Looks good, thanks Nan. We can merge it now, out of abundance of caution let us wait for a koji build.

PR #16507, opened 2026-04-08
fix(python3.14): disable experimental JIT to resolve clang/llvm dependency

christopherco:
The upstream Fedora 43 spec requires clang(major) = 19 / llvm(major) = 19 to build JIT stencils, but AZL ships LLVM 21. Disable JIT entirely, matching the RHEL approach: https://src.fedoraproject.org/rpms/python3.14/pull-request/76

JIT is experimental and off-by-default (requires PYTHON_JIT=1 at runtime).

- Create dedicated python3.14.comp.toml with build.without = ["jit"]
- Remove inline entry from components.toml


binujp (approved):
Looks good.

PR #16512, opened 2026-04-08
feat: Disable systemd udev rule that sets default io scheduler to bfq

ddstreet:
This disables (comments out) the udev rule that sets the default io scheduler to 'bfq'. We do not want that as the kernel default of 'none' (or 'mq-deadline' for single-queue devices) provides better raw performance, especially in the Azure cloud where all drives are virtualized and the underlying hypervisor can better handle io scheduling if needed.


ddstreet:
This does not force all drives to use the 'none' scheduler, instead it removes the explicit udev scheduler setting (to 'bfq') coming from the Fedora-specific udev rules file.

The result is regular drives (with a single queue) will get 'mq-deadline' and nvme drives will get 'none', as those are the kernel defaults.

christopherco (approved):
Ran this locally and built in cloud build. LGTM!

PR #16514, opened 2026-04-08
Configure chrony to use PTP in azure

ddstreet:
This configures chrony to use the `/dev/ptp_hyperv` PTP device present in azure vms.


christopherco:
I've rebased this to handle the recent change that caused merge conflict

christopherco (approved):

PR #16519, opened 2026-04-08
fix(ImageMagick): pin to 7.1.1.47-6 to avoid libuhdr dependency

liunan-ms:
Pin to commit d19ffa4 (1:7.1.1.47-6), before uhdr support was added in ImageMagick as libuhdr is not available in AZL4.
Move ImageMagick from inline components-full.toml to dedicated file.

Validation: [Koji build](https://controltower-dev-jwisitgpr74k6-gjb0fchvgkbnamgp.b01.azurefd.net/workflow/job/f11aeb1e-adbd-4191-f455-08de95919377/koji-summary)


binujp:
Thanks for this change, Nan. It can merge but I have a thought about commit pinning. Given we have an overlay which is potentially version specific it is safer to pin the spec to the version snapshot is getting us. What do you think?

binujp (approved):
Looks good, thanks Nan!

PR #16521, opened 2026-04-08
fix(openblas): add missing BuildRequires on python3

christopherco:
The openblas %build section runs `make ... lapack-test`, which invokes lapack_testing.py — a Python 3 script. Neither the upstream Fedora spec nor AZL declares BuildRequires: python3 explicitly.

In Fedora this works because python3 is pulled in transitively:
  perl-devel → systemtap-sdt-devel → python3
AZL builds perl with `without: ["perl_enables_systemtap"]`, which drops the systemtap-sdt-devel Requires from perl-devel, breaking that chain. The bootstrap build succeeded because it used raw Fedora fc43 packages where this chain was intact; the regular build failed on aarch64 because the AZL-rebuilt perl-devel no longer pulls in python3.

Add BuildRequires: python3 via spec-add-tag overlay. Move the component from inline components-full.toml to a dedicated comp.toml to hold the overlay.

### DIFF
--- a/base/comps/components-full.toml
+++ b/base/comps/components-full.toml
@@ -2615,7 +2615,6 @@
 [components.open-vmdk]
 [components.openal-soft]
 [components.openbios]
-[components.openblas]
 [components.openbox]
 [components.opencl-filesystem]
 [components.opencl-headers]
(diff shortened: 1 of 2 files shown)


reubeno (approved):

ddstreetmicrosoft (approved):
lgtm

PR #16522, opened 2026-04-08
feat: add fully generated comps.xml file and readme

ddstreet:
Add a /base/comps/comps.xml file for azl4, and a
base/comps/README-comps-xml.md file describing how to generate a new comps.xml file in the future.


ddstreet:
I did a quick pass through all the packages listed, and removed all that we don't seem to carry.

I also left translations intact, as it's actually more work to not include translations, but it's easy to remove the translations if we think that's better.

base/comps/comps.xml
```diff
@@ -0,0 +1,428 @@
...
+    <description xml:lang="sk">Tieto nástroje obsahujú základné vývojové nástroje ako sú git a cCVS.</description>
+    <description xml:lang="sv">Dessa verktyg inkluderar allmänna utvecklingsverktyg såsom git och CVS.</description>
+    <description xml:lang="tr">Bu araçlar git ve CVS gibi genel geliştirme araçlarını içerir.</description>
+    <description xml:lang="uk">До списку цих засобів включено загальні інструменти розробки, зокрема git та CVS.</description>
+    <description xml:lang="zh_CN">这些工具包括通用开发工具，如 git 和 CVS。</description>
+    <description xml:lang="zh_TW">這些工具包括一般開發工具，例如 git 與 CVS。</description>
+    <default>false</default>
+    <uservisible>true</uservisible>
+    <packagelist>
+      <packagereq type="mandatory">gettext</packagereq>
```
reubeno, line 411: Okay, interesting these "Development Tools" aren't what I expected at all from our previous conversation. From what Chris said, I was expecting to see something that was the equivalent of "build-essential".

This list doesn't... seem terribly useful.

reubeno (commented):
This looks pretty reasonable, modulo my one comment. Thanks for doing the work to trim this down!

base/comps/comps.xml
christopherco: ooof yeah, I should have actually checked the Fedora package grouping more closely before making that correlation to "Development Tools". These all seem like useful user dev tools, but not the equivalent of `build-essential` I thought it would be. The group that matches the closest is `dnf group info "C Development Tools and Libraries"`

christopherco: @ddstreet and @reubeno thoughts on including "C Development Tools and Libraries" group?

christopherco (approved):

PR #16526, opened 2026-04-09
fix(rust-clap_derive): pin 4.6.0 to match rust-clap and rust-clap_builder

anphel31:
rust-clap and rust-clap_builder were pinned to 4.6.0 (upstream Fedora commit by Fabio Valentini, 2026-03-24), but rust-clap_derive was left at 4.5.55. This caused a cascading dependency failure:

  rust-clap 4.6.0 requires crate(clap_derive/default) = 4.6.0
  → rust-dlwrap requires crate(clap/derive) >= 4.0.0
  → rust-podman-sequoia requires rust-dlwrap+default-devel

Pin rust-clap_derive to upstream commit 4642814e (version 4.6.0, "Update to version 4.6.0; Fixes RHBZ#2447042") to restore the lockstep version alignment with rust-clap and rust-clap_builder.

Tested: built rust-clap_derive 4.6.0 locally, verified RPMs install in mock chroot and provide crate(clap_derive) = 4.6.0 and crate(clap_derive/default) = 4.6.0.


base/comps/rust-clap_derive/rust-clap_derive.comp.toml
```diff
@@ -0,0 +1,4 @@
+# Pin to 4.6.0 to be in sync with rust-clap and rust-clap_builder
```
reubeno: How did this work upstream? I see that f43 does *now* have `rust-clap_derive` at 4.6.0. Is it that our current snapshot is between when the updates happened across crates?

anphel31: I think the issue is just that a [previous commit](https://github.com/microsoft/azurelinux/commit/e9785c388931b8a1aa0e3fefdea7e404c44b4c44) pinned `rust-clap` and `rust-clap_builder` to version 4.6.0, and we also needed `rust-clap_derive` to linue up with these

anphel31: So for upstream f43, all 3 packages were bumped to 4.6.0 simultaneously

base/comps/rust-clap_derive/rust-clap_derive.comp.toml
```diff
@@ -0,0 +1,4 @@
...
+# Remove this pin once the Koji builders resolve these from prior completed builds
```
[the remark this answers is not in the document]
anphel31: fixed

reubeno (approved):

PR #16529, opened 2026-04-09
fix(ci): pin azldev to specific commit, drop schema freshness check

christopherco:
The lint CI job was failing because the checked-in JSON schema (azldev.schema.json) was stale — missing types and fields added to azldev since the initial import.

Changes

1. Regenerated the schema via azldev config generate-schema.
2. Replaced the Rust/taplo toolchain with azldev in CI:
   - `azldev config dump` for strict config validation.
   - `azldev config generate-schema` + `diff` to fail if the checked-in schema drifts.
3. Deleted taplo.toml.

`azldev` is the authoritative parser for these files and validates the full config lifecycle, which is strictly stronger than taplo + a static schema.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -15,25 +15,23 @@ permissions:
 
 jobs:
   lint:
-    name: "Lint TOML files"
+    name: "Validate TOML config"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
-    
-      - name: Set up rust toolchain
-        uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
+
+      - name: Set up Go
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
         with:
-          toolchain: stable
+          go-version: stable
+          cache: false
 
-      - name: Enable cargo cache
-        uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
+      - name: Install azldev
+        run: go install github.com/microsoft/azure-linux-dev-tools/cmd/azldev@0590d38b32c0a414e0969e85aca8542c3d339079
 
-      - name: "Install linter (taplo)"
-        run: cargo install --locked taplo-cli
-
-      - name: "Run linter"
-        run: taplo lint --colors=always
+      - name: "Validate config (strict)"
+        run: azldev config dump > /dev/null
    
--- a/external/schemas/azldev.schema.json
+++ b/external/schemas/azldev.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://github.com/gim-home/azldev-preview/internal/projectconfig/config-file",
+  "$id": "https://github.com/microsoft/azure-linux-dev-tools/internal/projectconfig/config-file",
   "$ref": "#/$defs/ConfigFile",
   "$defs": {
     "CheckConfig": {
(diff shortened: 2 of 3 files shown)


.github/workflows/lint.yaml
```diff
@@ -15,25 +15,31 @@ permissions:
...
+        run: go install github.com/microsoft/azure-linux-dev-tools/cmd/azldev@main
 
-      - name: "Install linter (taplo)"
-        run: cargo install --locked taplo-cli
+      - name: "Validate config (strict)"
+        run: azldev config dump -q > /dev/null
 
-      - name: "Run linter"
-        run: taplo lint --colors=always
+      - name: "Check schema freshness"
```
christopherco: question: do we want this schema freshness check? It would mean any update to the schema in azldev would start failing our CI, even if the utilization is compatible with the change being evaluated.
The checked-in schema is really only for AI to reference more readily than running `azldev`, but I could see dropping this freshness job and making the checked-in schema best-effort.

.github/workflows/lint.yaml
```diff
@@ -15,25 +15,31 @@ permissions:
...
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
         with:
-          toolchain: stable
+          go-version: stable
+          cache: false
 
-      - name: Enable cargo cache
-        uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
+      - name: Install azldev
+        run: go install github.com/microsoft/azure-linux-dev-tools/cmd/azldev@main
```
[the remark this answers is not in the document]
christopherco: @copilot+claude-opus-4.6 apply changes based on [this feedback](https://github.com/microsoft/azurelinux/pull/16529#discussion_r3055921404) pin to 0590d38b32c0a414e0969e85aca8542c3d339079

.github/workflows/lint.yaml
```diff
@@ -15,25 +15,23 @@ permissions:
...
 
-      - name: "Install linter (taplo)"
-        run: cargo install --locked taplo-cli
+      - name: Install azldev
+        run: go install github.com/microsoft/azure-linux-dev-tools/cmd/azldev@0590d38b32c0a414e0969e85aca8542c3d339079
 
-      - name: "Run linter"
-        run: taplo lint --colors=always
+      - name: "Validate config (strict)"
+        run: azldev config dump -q > /dev/null
```
reubeno: Why quiet?

external/schemas/azldev.schema.json
```diff
@@ -1,6 +1,6 @@
 {
```
reubeno, line 1: Do we still need the schema checked in at all now that it's available on GitHub under microsoft/azure-linux-dev-tools?

(I'm okay deferring that evaluation to a later PR.)

reubeno (approved):
Righteous change!

I'm confused by the PR summary, though:

>CI was using go install ...@main for azldev — non-reproducible and a supply-chain risk.

CI wasn't doing *anything* with `azldev`.

christopherco:
> Righteous change!
> 
> I'm confused by the PR summary, though:
> 
> > CI was using go install ...@main for azldev — non-reproducible and a supply-chain risk.
> 
> CI wasn't doing _anything_ with `azldev`.

sooooo I queued off Copilot to fix one comment, and I expected a new commit with the change, but looks like it completely rewrote the PR summary while I was away. Will fix, and noted for the future when using Copilot in the cloud

external/schemas/azldev.schema.json
christopherco: The schema is still referenced in the various skill files in this tree. I agree that we should just link to the schema in azure-linux-dev-tools instead, and remove the schema altogether from this repo. I'll do that in a follow up PR.

.github/workflows/lint.yaml
christopherco: It doesn't need to be quiet, so I'll change it.
I did introspect this since I used Copilot to generate the changes, and looks like it was followed guidance from what we have in copilot-instructions for the agent to utilize '-q' and also in the example commands. https://github.com/microsoft/azurelinux/blob/e61472472c651c933e2a9c6f665f269b918123b3/.github/copilot-instructions.md?plain=1#L44

christopherco: I'll change this in a follow on PR to keep this change focused on fixing the CI.

christopherco:
CI error with `zizmor` is due to other workflow issues. I'll have another PR for those shortly.

christopherco:
> CI error with `zizmor` is due to other workflow issues. I'll have another PR for those shortly.

https://github.com/microsoft/azurelinux/pull/16544

PR #16548, opened 2026-04-09
fix(libplasma): pin to 6.6.3 to fix SONAME mismatch (.so.6 -> .so.7)

liunan-ms:
This pull request addresses a SONAME mismatch issue for the `libplasma` component by moving its configuration to a dedicated file and pinning its version.


base/comps/kscreenlocker/kscreenlocker.comp.toml
christopherco, line 1: nit: please remove the whitespace changes here, so the diff is only for `libplasma`

christopherco (approved):

base/comps/kscreenlocker/kscreenlocker.comp.toml
christopherco: Fixed

PR #16549, opened 2026-04-09
feat(packages): add package publish annotation support

liunan-ms:
This PR introduces a project-wide example configuration for package publishing channels and updates the project to include this configuration. It also adds an example of component-level configuration for `wget2-wget` package. These changes help clarify and demonstrate how packages are grouped and assigned to publishing channels in the project.

**Project-wide package publishing configuration:**

* Added `packages-demo.toml`, an example configuration file that demonstrates how packages are grouped and assigned to publish channels (`rpm-base` and `rpm-build-only`). This file includes sample package groupings and default publishing behaviors.
* Updated `project.toml` to include the new `packages-demo.toml` in the list of configuration includes, ensuring the example package publishing configuration is loaded by the project.

**Component-specific publishing configuration:**

* Added a `publish.channel` setting for the `wget2-wget` package in `wget2.comp.toml`, specifying that it should be published to the `rpm-base` channel.

`azldev package list <package-name>` can be used to view the package publish annotations before build, 
e.g.
image

The annotation will be available in the build output:
image

### DIFF
--- a/base/comps/wget2/wget2.comp.toml
+++ b/base/comps/wget2/wget2.comp.toml
@@ -3,3 +3,7 @@
 [components.wget2.build]
 # Supply the wget compatibility package.
 with = ["as_wget"]
+
+# Component level configuration
+[components.wget2.packages.wget2-wget.publish]
+channel = "rpm-base"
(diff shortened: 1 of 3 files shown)


christopherco (changes requested):
@liunan-ms is making adjustments to `azldev` to support a different channel naming. Once that is ready, this can be updated and we can re-review/merge

PR #16550, opened 2026-04-10
realmd: add --with-distro=redhat overlay

reubeno:
Add a spec-search-replace overlay to inject --with-distro=redhat into the %configure call so realmd knows that Azure Linux is Fedora-like. Without this change, the component fails to build.

_**Validation**: local stage2 build failed without the change, succeeds with it._


christopherco (approved):
:shipit:

PR #16557, opened 2026-04-10
fix(automake): set rhel=10 to skip Fedora-only test deps

christopherco:
Azure Linux does not ship gcc-objc or gcc-objc++ (ObjC support is
disabled in our GCC build). The upstream Fedora spec already skips
these BuildRequires for RHEL. Set rhel=10 via build.defines so the
spec takes the RHEL path, skipping the same set RHEL excludes:
gcc-objc, gcc-objc++, imake, lzip, ncompress.


base/comps/automake/automake.comp.toml
```diff
@@ -0,0 +1,9 @@
+[components.automake]
+
+# Azure Linux does not ship gcc-objc/gcc-objc++ (ObjC support is disabled in our GCC build).
+# The upstream spec already skips these BuildRequires for RHEL — extend the conditional to cover AZL too.
+[[components.automake.overlays]]
+description = "Skip gcc-objc, gcc-objc++, imake, lzip, ncompress BuildRequires (not available in AZL, same as RHEL)"
+type = "spec-search-replace"
+regex = '%if !0%\{[?]rhel:1\}$'
+replacement = "%if !0%{?rhel:1} && !0%{?azl:1}"
```
christopherco: Better to just define `rhel` here

reubeno (approved):

PR #16569, opened 2026-04-10
fcode-utils: force -std=gnu17 to fix GCC 15 build failure

reubeno:
GCC 15 defaults to C23, where `bool` is a keyword. This conflicts with the `typedef enum boolean { FALSE = 0, TRUE = -1 } bool;` declaration in shared/types.h, causing the build to fail.

Force -std=gnu17 via a CFLAGS overlay as a conservative workaround until a new upstream release (> 1.0.3) includes the fix.

Ref: https://github.com/openbios/fcode-utils/issues/20

_**Validation**: confirmed it (originally) reproduced with a local stage2 build; confirmed build now succeeds with this fix._

### DIFF
--- a/base/comps/components.toml
+++ b/base/comps/components.toml
@@ -483,7 +483,7 @@ includes = ["**/*.comp.toml", "component-check-disablement.toml"]
 [components.fast_float]
 [components.fasterxml-oss-parent]
 [components.fcgi]
-[components.fcode-utils]
+
 [components.fcoe-utils]
 [components.fdk-aac-free]
 [components.fdupes]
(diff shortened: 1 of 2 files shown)


ddstreetmicrosoft (approved):
The commit subject line isn't in conventional comments format, we should try to use that when possible.

This does seem to have been [patched upstream](https://github.com/openbios/fcode-utils/pull/25/changes/3fa70fef80276aab948ee6c7d10903b268d5b2d5) for the issue you referenced, which is a larger change, so probably better to just use the smaller/simpler fix here until we do a full version update.

otherwise lgtm.

PR #16570, opened 2026-04-10
jss: enable PQC auth defs unconditionally for NSS >= 3.112

reubeno:
AZL's NSS (3.120+) includes PQC auth defs (ML-DSA) in ssl_auth_size, but the Fedora 43 JSS spec only enables the matching NSS_VERSION_PQC_DEF flag for RHEL >= 10.  This causes a PR_STATIC_ASSERT failure in SSLCipher.c during stage2 builds (where NSS comes from AZL, not Fedora).

Supply the enable_nss_version_pqc_def_flag macro via build defines so the PQC cmake flag is always passed — matching what Fedora rawhide (jss 5.9.0) now does unconditionally.

We alternatively considered trying to reuse the existing conditionally-applied define in the upstream spec we're using, but the only possible overlays were too brittle -- running the risk that we might unwittingly removing the wrong conditionality if upstream spec changes added new RHEL-conditional ifdefs.

_**Validation**: confirmed that local stage2 build fails without this fix; confirmed that local stage2 build succeeds with this fix._


ddstreetmicrosoft (approved):
the commit description is a bit hard to follow, but generally lgtm.

i will mention i'm not a fan of duplicating the commit description in both the git commit as well as toml comment.

PR #16571, opened 2026-04-10
iscsi-initiator-utils: fix libiscsi build failure on AZL (wrong meson build dir)

reubeno:
Fedora's 0104-libiscsi-build-fixes.patch hardcodes -L../redhat-linux-build in the libiscsi Makefile, assuming the meson build directory is named redhat-linux-build (Fedora's %{_vpath_builddir} = %{_vendor}-%{_target_os}-build where _vendor=redhat). On AZL, _vendor=azurelinux, so the meson build directory is azurelinux-linux-build and the linker cannot find libopeniscsiusr.so:

  /usr/bin/ld: cannot find -lopeniscsiusr: No such file or directory

Fix by:
1. Replacing the hardcoded path in the patch with a $(MESON_BUILDDIR) make variable via file-search-replace overlay.
2. Passing MESON_BUILDDIR=%{_vpath_builddir} from the spec's %make_build invocation via spec-search-replace overlay.

This uses the standard RPM macro rather than hardcoding any vendor-specific directory name.

_**Validation**: verified that it failed on a local stage2 build without the fix; verified that a local stage2 build passes with the fix._


ddstreetmicrosoft (approved):
lgtm

PR #16572, opened 2026-04-10
libyang: fix hardcoded redhat-linux-build in %check for AZL

reubeno:
The spec hardcodes 'pushd redhat-linux-build' in %check, assuming the CMake build directory follows Fedora's naming (%{_vendor}=redhat). On AZL, %{_vendor}=azurelinux, so the directory is azurelinux-linux-build and the ctest invocation fails.

Replace the hardcoded path with %{_vpath_builddir}, the standard RPM macro that resolves to the correct vendor-specific build directory on any distro.

_**Validation**: confirmed %check fails on a stage2 build via local build without this fix; confirms that local stage2 build succeeds with the fix._


ddstreetmicrosoft (approved):
lgtm

PR #16573, opened 2026-04-10
onnx: fix hardcoded redhat-linux-build in setup.py patch for AZL

reubeno:
Fedora's 0002-Let-pyproject_wheel-use-binaries-from-cmake_build.patch hardcodes "redhat-linux-build" as the CMake build directory in setup.py. On AZL, %{_vpath_builddir} is azurelinux-linux-build (_vendor=azurelinux), so %pyproject_wheel can't find the CMake build output.

Fix by:
1. Replacing the hardcoded string in the patch with an os.environ.get() call that reads VPATH_BUILDDIR (with redhat-linux-build as fallback for Fedora compatibility).
2. Exporting VPATH_BUILDDIR=%{_vpath_builddir} at the start of %build via spec-prepend-lines overlay.

_**Validation**: confirmed a local stage2 build of this component failed without the fix; confirmed a local stage2 build suceeded with the fix._


ddstreetmicrosoft (approved):
lgtm

PR #16577, opened 2026-04-10
chore: update stage2 TOML URI

reubeno:
This updates the Stage2 TOML URI, which is used by local Stage2 builds.

Also removes the 'cost' value that had mistakenly ended up in the mock config, and which was preventing locally built packages from overriding the prebuilt koji packages.


ddstreetmicrosoft (approved):
lgtm

PR #16583, opened 2026-04-10
Adopt (significantly modified) version of Fedora kiwi xml

ddstreet:
This pulls in the Fedora kiwi xml files and reduces it down the (image-providing) profiles `AzureVM`, `QemuVM`, and `HypervVM`.


ddstreet:
rebase on tomls/base/main

ddstreet:
The list of packages *removed* from the current `vm-base` image is:

> acl      
> add-determinism     
> ansible-srpm-macros      
> attr         
> azurelinux-release
> azurelinux-release-identity-basic
> azurelinux-rpm-config
> bind-libs         
> bind-utils     
> binutils       
> brotli 
> build-reproducibility-srpm-macros
> bzip2             
> cronie            
> cronie-anacron         
> crontabs    
> cyrus-sasl
> dns-root-data
> dracut-network
> dwz   
> efi-srpm-macros
> elfutils
> filesystem-srpm-macros                  
> fonts-srpm-macros
> forge-srpm-macros
> fpc-srpm-macros
> fstrm               
> gap-srpm-macros
> ghc-srpm-macros   
> gnat-srpm-macros
> gnutls-dane      
> go-srpm-macros
> gpgme               
> gpgverify                
> hiredis      
> irqbalance      
> java-srpm-macros
> jq                   
> kernel            
> kernel-srpm-macros
> kernel-tools     
> kernel-tools-libs
> libestr         
> libfastjson       
> libmaxminddb      
> libndp                 
> libuv          
> lua-srpm-macros
> lz4         
> ncurses-term
> net-tools     
> NetworkManager      
> NetworkManager-libnm
> ngtcp2                                  
> ngtcp2-crypto-ossl                                                                             
> numactl-libs                                                                                   
> ocaml-srpm-macros
> oniguruma           
> openblas-srpm-macros     
> package-notes-srpm-macros
> pciutils-libs     
> perl-srpm-macros                 
> protobuf-c           
> pyproject-srpm-macros
> python-srpm-macros
> qt5-srpm-macros
> qt6-srpm-macros
> rsyslog                          
> rust-srpm-macros  
> shadow-utils-subid
> systemd-rpm-macros     
> tree-sitter-srpm-macros
> unbound-libs
> wget2        
> wget2-libs    
> wget2-wget
> zchunk         
> zig-srpm-macros
> zip                                     
> 

the list of packages *added* in the new image is:

> 7zip       
> azurelinux-release-cloud
> azurelinux-release-identity-cloud
> azurelinux-repos-evergreen
> c-ares               
> console-login-helper-messages
> console-login-helper-messages-motdgen
> console-login-helper-messages-profile
> crypto-policies-scripts
> cyrus-sasl-gssapi
> dracut-config-generic  
> dracut-config-rescue
> efibootmgr
> freetype
> gawk-all-langpacks
> graphite2
> grub2-tools-efi
> grub2-tools-extra
> harfbuzz
> initscripts
> initscripts-rename-device
> initscripts-service
> ipset
> ipset-libs
> libbasicobjects
> libcap-ng-python3
> libcollection
> libdhash
> libini_config
> libldb
> libpath_utils
> libpng
> libref_array
> libsecret
> libsss_certmap
> libsss_idmap
> libsss_nss_idmap
> libsss_sudo
> libtalloc
> libtdb
> libtevent
> libxkbcommon
> memstrack
> mkpasswd
> mtools
> nano-default-editor
> ncurses
> pinentry
> pkcs11-provider
> prefixdevname
> python3-pysocks
> python3-urllib3+socks
> python-unversioned-command
> qrencode-libs
> rootfiles
> rpm-plugin-audit
> sssd-client
> sssd-common
> sssd-kcm
> sssd-krb5-common
> sudo-python-plugin
> systemd-oomd-defaults
> tpm2-tools
> tpm2-tss-fapi
> whois-nls
> xkeyboard-config
> zram-generator
> zram-generator-defaults
>

ddstreet:
To comment on the *removed* packages:

> ansible-srpm-macros
> build-reproducibility-srpm-macros
> efi-srpm-macros
> filesystem-srpm-macros
> fonts-srpm-macros
> forge-srpm-macros
> fpc-srpm-macros
> gap-srpm-macros
> ghc-srpm-macros
> gnat-srpm-macros
> go-srpm-macros
> java-srpm-macros
> kernel-srpm-macros
> lua-srpm-macros
> ocaml-srpm-macros
> openblas-srpm-macros
> package-notes-srpm-macros
> perl-srpm-macros
> pyproject-srpm-macros
> python-srpm-macros
> qt5-srpm-macros
> qt6-srpm-macros
> rust-srpm-macros
> systemd-rpm-macros
> tree-sitter-srpm-macros
> zig-srpm-macros
> 

we don't need, and shouldn't include, any *-[s]rpm-macros files in our image. They are needed solely for building packages, not normal operation.

> azurelinux-release
> azurelinux-release-identity-basic

we use azurelinux-release-identity-cloud in the new image, which is what we should be using.

> add-determinism
> binutils
> dwz
> elfutils

this is only needed when building and/or debugging stuff

> brotli
> bzip2
> lz4
> zchunk
> zip

these are cmdline compression tools, i don't think we need these installed by default

> cronie
> cronie-anacron
> crontabs

'cron' has generally been replaced by systemd timers; we do not need these packages

> jq
> libndp
> NetworkManager
> NetworkManager-libnm
> oniguruma
> dracut-network

all these are required *only* due to installing dracut-network, and we don't need the network operational during initrd

> dns-root-data
> gnutls-dane
> gpgme
> hiredis
> ngtcp2
> ngtcp2-crypto-ossl
> unbound-libs
> wget2
> wget2-libs
> wget2-wget

all of these are required *only* due to installing wget2; i don't think we need wget2 installed by default

> libestr
> libfastjson
> rsyslog

these are required *only* due to installing rsyslog, which we don't need - systemd-journald is what we use for system logging by default

> bind-libs
> bind-utils
> fstrm
> libmaxminddb
> libuv

these are required *only* due to installing bind-utils; i don't think we need to have bind-utils installed by default, all it provides are some convenience tools for dns lookups

> ncurses-term

i don't know why we would need this? it provides 'additional terminal descriptions not found in the ncurses-base package'

> net-tools

this contains a collection of deprecated tools and is being phased out of most distros. we don't need it installed by default.

> numactl-libs
> irqbalance

numactl-libs is only needed by irqbalance, and irqbalance is not always needed, and sometimes actively degrades performance. especially for a cloud vm (i.e. virtualized vm) i don't think we should install irqbalance by default.

> pciutils-libs
> kernel-tools
> kernel-tools-libs

these are needed only be kernel-tools; i don't think we should install it by default

> protobuf-c

this is needed by both bind-utils and wget2-wget; if we don't need either of those packages, this won't be installed by default.

> shadow-utils-subid

this is used for maanging subordinate uid/gid ranges; it provides the program 'getsubids' and the library 'libsubid'; i can see this might be used in certain use cases, but i don't think it should be installed by default.

> kernel

this is a metapackage; we don't need it installed by default

> azurelinux-rpm-config
> acl
> attr
> cyrus-sasl
> gpgverify

maybe we need this? I'm not sure.

ddstreet:
to comment on the _added_ packages:

> azurelinux-release-cloud
> azurelinux-release-identity-cloud

we do need these, and set to the cloud edition/variant.

> azurelinux-repos-evergreen

we need this (i manually changed my image build to use evergreen, the actual kiwi xml uses the regular azurelinux-repos)

> c-ares
> cyrus-sasl-gssapi
> libbasicobjects
> libcollection
> libdhash
> libini_config
> libldb
> libpath_utils
> libref_array
> libsss_certmap
> libsss_idmap
> libsss_nss_idmap
> libsss_sudo
> libtalloc
> libtdb
> libtevent
> sssd-client
> sssd-common
> sssd-kcm
> sssd-krb5-common

all these are needed *only* for the sssd-kcm and sssd-common packages; i'm not familiar with 'sssd' and i don't know if we need it installed by default? i suspect we don't?

> console-login-helper-messages
> console-login-helper-messages-motdgen
> console-login-helper-messages-profile

these can provide some messages at the login console - note that they will only be seen on the serial console, printed right before the login prompt is shown.

> dracut-config-generic

this allows dracut to provide a 'generic' (i.e. non-host-specific) initrd. we definitely should use this.

> dracut-config-rescue

this allows dracut to provide a 'rescue' initrd; i'm not familiar with what exactly that means, but it could be useful. maybe we should remove this package until we can investigate in more detail if it's useful to us or not

> efibootmgr

this provides tools to interact with and/or configure the system's UEFI; i think this may be of limited value in the cloud environment. maybe we should drop this?

> freetype
> graphite2
> harfbuzz
> libpng
> mtools
> grub2-tools-extra

these are all needed *only* by grub2-tools-extra, which provides some grub2-related commands that probably won't be useful very often; i think we should drop this

> grub2-tools-efi

similar to above, but it provides grub2-related commands used specifically on a system with UEFI. maybe we can drop this.

> initscripts
> initscripts-rename-device
> initscripts-service

these provide scripts/tools used only for compatibility with the old SysV standard. we _probably_ don't need these, or at least not all of them.

> ipset
> ipset-libs
> libcap-ng-python3

these are recommends from firewalld; we should probably keep them, but they are optional

> libsecret
> pinentry

these are recommended by gnupg2; as above, we should probably keep them, but they are optional

> memstrack

this is recommended by dracut; i do not think we need it

> mkpasswd
> whois-nls

these are recommended by libxcrypt; i do not think we need them

> nano-default-editor

this requires nano, and sets it as the default editor. we should keep it

> ncurses

this provides common cli tools for terminal management, such as 'clear' and 'reset'. we should keep this.

> prefixdevname

udev helper 'that provides network interface naming using user defined prefix' by adding something like 'net.ifnames.prefix=xyz' to the kernel cmdline, after which it will name all interfaces something like 'xyz0, 'xyz1', etc. i don't think this really needs to be installed by default.

> python3-pysocks
> python3-urllib3+socks

these appear to be needed for cloud-init; i'm not sure if they are required or optional.

> python-unversioned-command

this small package provides /usr/bin/python which calls python3. i think we're far enough past python2 EOL that we can assume all python scripts handle python3 and anything calling unversioned 'python' can run under python3, so we should keep this package.

> rootfiles

provides 'skeleton' files for root user. we should probably keep this.

> rpm-plugin-audit

provides audit logging for package operations; we should keep it

> systemd-oomd-defaults

provides default values for systemd-oomd setup. this seems to be what all distros are doing now, so we should probably include it.

> tpm2-tools
> tpm2-tss-fapi

definitely need these so we can interact with vTPMs in azure cloud

> zram-generator
> zram-generator-defaults

i don't know if we need to include these by default, as i don't think users will want to use zram on their cloud system

> 7zip
> crypto-policies-scripts
> gawk-all-langpacks
> pkcs11-provider
> sudo-python-plugin
> libxkbcommon
> xkeyboard-config
> qrencode-libs

I'm not sure if we need this?

ddstreet:
converted to draft as i expect changes to be needed based on feedback (including some of the packages i outlined above as probably not needing)

base/images/common/base.xml
```diff
@@ -0,0 +1,127 @@
...
+		<package name="parted"/>
+		<package name="policycoreutils"/>
+		<package name="prefixdevname"/>
+		<package name="procps-ng"/>
+		<package name="rootfiles"/>
+		<package name="rpm"/>
+		<package name="selinux-policy-targeted"/>
+		<package name="setup"/>
+		<package name="shadow-utils"/>
+		<package name="sssd-common"/>
```
reubeno: Do we want to include `sssd` by default? What's the reasons for doing so?

base/images/common/base.xml
```diff
@@ -0,0 +1,127 @@
...
+		<package name="sssd-kcm"/>
+		<package name="sudo"/>
+		<package name="systemd"/>
+		<package name="systemd-networkd"/>
+		<package name="systemd-oomd-defaults"/>
+		<package name="systemd-resolved"/>
+		<package name="uboot-images-armv8" arch="aarch64"/>
```
reubeno: Why are we including uboot tools? I wouldn't expect uboot is used in Azure VMs.

base/images/common/base.xml
```diff
@@ -0,0 +1,127 @@
...
+		<package name="uboot-tools" arch="aarch64"/>
+		<package name="util-linux"/>
+		<package name="vim-minimal"/>
```
reubeno, line 42: `vim` tends to be full of CVEs. I can see reason to make sure we have *some* editor (nano?) but not necessarily vim.

base/images/common/base.xml
```diff
@@ -0,0 +1,127 @@
...
+		<package name="zram-generator-defaults"/>
+	</packages>
+	<packages type="image" patternType="plusRecommended" profiles="BaseCommon">
+		<package name="abrt-cli"/>
+		<package name="acl"/>
+		<package name="amd-ucode-firmware" arch="x86_64"/>
```
reubeno: Do we include firmware packages generally?

base/images/vm/base.xml
```diff
@@ -0,0 +1,49 @@
...
+		<package name="systemd-oomd-defaults"/>
+	</packages>
+	<packages type="image" patternType="plusRecommended" profiles="CoreHypervVM">
+		<package name="hyperv-daemons"/>
+	</packages>
+	<packages type="image" patternType="plusRecommended" profiles="CoreQemuVM">
+		<package name="qemu-guest-agent"/>
+	</packages>
+	<packages type="image" patternType="plusRecommended" profiles="BaseVM">
+		<ignore name="kernel"/>
```
reubeno, line 41: Remind me what effect ignores take? Is this for ignoring weak dependencies?

base/images/AzureLinux.kiwi
```diff
@@ -0,0 +1,28 @@
...
+		<contact>azurelinux@microsoft.com</contact>
+		<specification>AzureLinux</specification>
+	</description>
+	<preferences>
+		<version>4</version>
+		<packagemanager>dnf5</packagemanager>
+		<locale>en_US</locale>
+		<keytable>us</keytable>
+		<timezone>UTC</timezone>
+		<release-version>4</release-version>
```
reubeno, line 15: We should probably discuss the whole `4` vs. `4.0` thing to make sure we're consistent. (I don't think I have a strong opinion one way or the other.)

base/images/images.toml
```diff
@@ -6,6 +6,18 @@ definition = { type = "kiwi", path = "vm-base/vm-base.kiwi" }
 description = "Container Base Image"
 definition = { type = "kiwi", path = "container-base/container-base.kiwi" }
 
+[images.AzureVM]
```
reubeno, line 9: How do you see AzureVM being different from the `vm-base` we already have? Is the plan to remove the latter?

We should also align on naming within the TOML file (i.e., `images.AzureVM` vs. `images.azure-vm`), independent of the kiwi profile names used.

base/images/common/base.xml
```diff
@@ -0,0 +1,127 @@
+<image>
+	<profiles>
+		<profile name="CoreCommon" description="Core common packages"/>
```
reubeno, line 3: What's the intended scoping of "Core" vs. "Base"? I'm assuming we're inheriting the terms from Fedora's kiwi files, but I don't think I'll have an easy time keeping them straight.

base/images/common/base.xml
```diff
@@ -0,0 +1,127 @@
...
+		<profile name="BaseCommon" description="Base common packages"/>
```
reubeno: Where is BaseCommon used?  I searched through quickly but couldn't find any consuming references.

reubeno (commented):
I think a synchronous discussion is going to help get to consensus on the deltas. What I'd most like to understand are the proposed guiding principles for which are included/excluded in which images.

base/images/common/base.xml
ddstreet: sorry i had meant to remove that, it's created from the comps.xml 'standard' group but not used by any (current) image profile. i'll do a quick removal of it and force-push

ddstreet:
removed BaseCommon profile

base/images/common/base.xml
ddstreet: as you said, it's inherited from Fedora's terminology.

base/images/images.toml
ddstreet: > How do you see AzureVM being different from the vm-base we already have

see the package list comparisons in previous comments

> Is the plan to remove the latter?

i think so, yes

> We should also align on naming

sure that's fine, do you have specific name suggestions?

base/images/AzureLinux.kiwi
ddstreet: i'd prefer '4' since we have not yet ever released a '.1' for any of our releases, and keeping it an integer also aligns with Fedora versioning.

base/images/vm/base.xml
ddstreet: yep, ignore packages unless it was explicitly specified or is a Required: (or in Required: chain) of a explicitly specified package, so this can trim out Recommends: or Suggests: (is my understanding, at least)

base/images/common/base.xml
ddstreet: no, but we aren't using this profile, and it doesn't exist in the pr anymore

base/images/common/base.xml
ddstreet: vim-minimal is already in our vm iamge, nothing new re: this pr.

base/images/common/base.xml
ddstreet: sorry yeah i did mean to remove that as well, thanks. forced-pushed with them removed.

base/images/common/base.xml
ddstreet: check my comments above re: sssd

ddstreet:
removed uboot from CoreCommon

ddstreet:
TL;DR for packages _removed_ from our current image:

I do not think we need _any_ of the packages that are listed above as being present in our current image definition, but not present in our new image definition, besides these packages:

> > azurelinux-rpm-config
> > acl
> > attr
> > cyrus-sasl
> > gpgverify

I am not sure if we need these packages or not. I tend to think we don't, but I am open to any reason(s) we should include them.

ddstreet:
TL;DR for packages _added_ to our new image:

I think we should include _all_ of these packages which are present in our new image definition, but missing from our current image definition:

> > console-login-helper-messages
> > console-login-helper-messages-motdgen
> > console-login-helper-messages-profile
> > dracut-config-generic
> > dracut-config-rescue
> > nano-default-editor
> > ncurses
> > python-unversioned-command
> > rootfiles
> > rpm-plugin-audit
> > systemd-oomd-defaults
> > tpm2-tools
> > tpm2-tss-fapi

These are packages only present due to Recommends: or similar optional dependency relationship. I think we should include them, but we can drop them without losing any other package(s).

> > ipset
> > ipset-libs
> > libcap-ng-python3
> > libsecret
> > pinentry
> > python3-pysocks
> > python3-urllib3+socks

I don't know about these packages. I think we probably can drop them but I'd have to dig more into what they provide and/or what other package(s) require them, if any.

> > 7zip
> > crypto-policies-scripts
> > gawk-all-langpacks
> > pkcs11-provider
> > sudo-python-plugin
> > libxkbcommon
> > xkeyboard-config
> > qrencode-libs

ddstreet:
rebase on tomls/base/main

ddstreetmicrosoft:
sorry, i didn't realize this was still open - closing.

PR #16585, opened 2026-04-10
fix: fix packages with dependency issues

trungams:
updating a few components with dependency issues


binujp (approved):
Looks good, thanks Trung.

PR #16588, opened 2026-04-10
fix(blis): Drop memkind BR on Fedora

liunan-ms:
This PR drops `memkind` BR on Fedora as `memkind-devel` is not available in the AZL build tag. Fedora 44 also removed it because memkind was dropped from Fedora 44+ and RHEL 10.

Validation: [koji build](https://controltower-dev-jwisitgpr74k6-gjb0fchvgkbnamgp.b01.azurefd.net/workflow/job/ab669709-309e-4429-81d3-08de972630f0/koji-summary)


base/comps/blis/blis.comp.toml
```diff
@@ -0,0 +1,7 @@
+[components.blis]
+
+[[components.blis.overlays]]
+description = "Remove %{?fedora} from the arch-limiting condition — memkind is not available in AZL4."
+type = "spec-search-replace"
+regex = '%if 0%\{\?el8\}%\{\?el9\}%\{\?fedora\}'
```
reubeno, line 6: Rather than try to flip the complex conditional, I'd recommend instead going with something that captures the more core intent: using a `spec-remove-tag` overlay and explicitly list `memkind-devel`.

base/comps/blis/blis.comp.toml
```diff
@@ -0,0 +1,7 @@
+[components.blis]
+
+[[components.blis.overlays]]
+description = "Remove %{?fedora} from the arch-limiting condition — memkind is not available in AZL4."
```
reubeno, line 4: Is its exclusion from AZL4 intentional?

PR #16620, opened 2026-04-12
fix(comps): Remove mingw-openexr component

christopherco:
mingw-openexr requires mingw32-libdeflate and mingw64-libdeflate which are not available. No other AZL package depends on it — gdal uses native openexr-devel and has mingw explicitly disabled.


christopherco:
Note: I plan to do a deeper review of `mingw` components and see if we can remove many more. `gdal` is one of the primary reasons `mingw` was pulled in, but since we're not enabling the `gdal` `mingw` path, we can do a thorough review to remove more `mingw` components from the distro.

anphel31 (approved):

PR #16653, opened 2026-04-13
feat: remove vm-base.kiwi package rsyslog

ddstreet:
This isn't needed for logging, which uses systemd-journald; and this package pulls in:

- libestr
- libfastjson


christopherco (dismissed):

christopherco (approved):
We've discussed this and have agreed to remove this to reduce footprint and utility duplication

PR #16655, opened 2026-04-13
feat: remove vm-base.kiwi package kernel-tools

ddstreet:
This contains tools mostly useful for bare-metal systems; we should not include this by default in our base image. It additionally pulled in:

- kernel-tools-libs
- libnl3
- pciutils-libs


reubeno (approved):
I'm supportive of this. It doesn't look like AZL3 pulled these in either.

christopherco (approved):

PR #16656, opened 2026-04-13
feat: add sources upload pipeline

PawelWMS:
Adding a pipeline to trigger CT's workflow for uploading sources for new or updated components. It's meant to be triggered by each merge queue item.

Why an ADO pipeline instead of GitHub Actions: calling CT endpoints would require authentication secrets kept in GitHub and we want don't want that.

**NOTES**:
- `.github/workflows/scripts/check_rendered_specs.py` is temporary until #16656 is merged and I can switch to the version of that script from that PR.
- The pipeline won't fail for the time being until we know they are stable and can be marked as required.
- The pipeline is meant for non-fork PRs **only** - enabling forks require security hardening (out of scope for this PR).

Other work:
- Added AI instructions for how we want the ADO pipelines written.
- Refactored scripts into subdirectories for clearer grouping and to scope down the dependencies for each inside their `requiremenets.txt` files.

ADO task: [AB#18397](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/18397)


.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -0,0 +1,181 @@
...
+              # The config key may not be present on every agent image, so tolerate its absence.
+              git config --unset extensions.worktreeConfig || true
+
+              # Full history is needed for spec rendering to work.
+              git fetch --unshallow
+
+              specs_dir="specs"
+
+              echo "Rendering specs..."
+              azldev component render -q -a -o "$specs_dir" --force
```
dmcilvaney: probably want --clean-stale here, or just use -a with the default config dir (ie don't use -o).
If its config driven, it does all the "right" things automatically.

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -0,0 +1,181 @@
...
+            displayName: "Authenticate pip"
+            inputs:
+              artifactFeeds: "azl/ControlTowerFeed"
+
+          - script: |
+              set -euo pipefail
+
+              echo "##[group]Mock"
+              tdnf install -y mock mock-rpmautospec python3-chardet
+              sudo usermod -aG mock "$(whoami)"
```
dmcilvaney, line 100: tdnf install mock should already setup the group

.github/workflows/ado/templates/sources-upload-stages.yml
PawelWMS: Thanks, I forgot about that. When I was writing this, the config wasn't there, yet. Or it wasn't read by `azldev` - I forgot which one it was.:)

.github/workflows/ado/sources-upload.yml
```diff
@@ -0,0 +1,63 @@
...
+      runOnHost: true
+      EnableCDPxPAT: false
+    stages:
+      - template: /.github/workflows/ado/templates/sources-upload-stages.yml@self
+        parameters:
+          outputDirectory: $(Build.ArtifactStagingDirectory)/output
+          artifactBaseName: prcheck
+          containerImage: mcr.microsoft.com/onebranch/azurelinux/build:3.0
+          poolType: linux
+          serviceConnection: CT-Endpoints-Access-ServiceConnection-DEV
```
dmcilvaney, line 68: Is the mismatch between dev and official template intentional here?

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -0,0 +1,186 @@
...
+
+              cat << EOF
+              NOTE: It's possible more components changed between the source and target commits than displayed below.
+                    This workflow only checks for updates in the 'sources' files and ignores all other changes.
+              EOF
+
+              components=()
+              while IFS= read -r line; do
+                components+=("$(basename "$(dirname "$line")")")
+              done < <(git diff "$SOURCE_COMMIT" "$TARGET_COMMIT" --name-only | grep '/sources$' | sort -u)
```
dmcilvaney: Looks like `...` might be more appropriate here, `..` (or default) diffs ALL changes including those from other PRs. `...` seems to auto-calculate the merge base for the PR.

```suggestion
              done < <(git diff "$TARGET_COMMIT...$SOURCE_COMMIT" --name-only | grep '/sources$' | sort -u)
```

.github/workflows/ado/templates/sources-upload-stages.yml
dmcilvaney: Technically its STILL not there... waiting on https://github.com/microsoft/azurelinux/pull/16674/changes#diff-c887c6bcaf3c91cd09515ad27d4da91292a3ec45f3146344b0d2f6a0e7ad6453R13

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -0,0 +1,186 @@
...
+              # The config key may not be present on every agent image, so tolerate its absence.
+              git config --unset extensions.worktreeConfig || true
+
+              # Full history is needed for spec rendering to work.
+              git fetch --unshallow
+
+              specs_dir="specs"
+
+              echo "Rendering specs..."
+              azldev component render -q -a -o "$specs_dir" --force
```
dmcilvaney: I think I've got a solid flow for isolating mock if this ever needs to handle untrusted input: https://github.com/microsoft/azurelinux/pull/16674/changes#diff-30a9c2fec326355d71a476f939a68c942b218cd7479ea9710f7afd4c5d41695bR105

If we can guarantee this only runs on code we have looked at, might not care.

.github/workflows/scripts/control-tower-prcheck/run_control_tower_prcheck.py
```diff
@@ -0,0 +1,505 @@
...
+    session: requests.Session,
+    base_url: str,
+    credential: DefaultAzureCredential,
+    audience: str,
+    token_holder: _TokenHolder,
+    payload: dict,
+) -> dict:
+    """POST the prcheck request and return the parsed response dict."""
+    url = f"{base_url}/api/Scenario/prcheck"
+    response = _request_with_refresh(
```
dmcilvaney, line 260: Does CT de-duplicate these requests, or handle them in a sane way? Network failure could occur after CT has committed to processing the request, but before we get the success response back.

.github/workflows/scripts/control-tower-prcheck/run_control_tower_prcheck.py
```diff
@@ -0,0 +1,505 @@
...
+    base_url = args.api_base_url.rstrip("/")
+
+    # ── Build payload ────────────────────────────────────────────────
+    payload: dict = {
+        "components": args.components,
+        "buildReason": args.build_reason,
+        "repoUri": args.repo_uri,
+        "sourceCommitSha": args.source_commit,
+    }
+    if args.source_branch is not None:
```
dmcilvaney, line 416: These strings could be "", which probably causes some stuff to blow up. argparse will only guarantee they have some value, not that its an actual string.

.github/workflows/ado/templates/sources-upload-stages.yml
PawelWMS: From what I was able to tell, installation only creates the `mock` group but doesn't add any users to it.

.github/workflows/ado/sources-upload.yml
PawelWMS: Yes, this is a SC to a dev environment but the pipeline itself is considered official, because talking to, and potentially altering, resources in Azure. Also, in the future this will be switched to publish to our production resources and I wanted to handle all errors related to using the `Official` OB template as soon as possible to avoid future surprises. In general, I believe there's nothing wrong with using the `Official` version of the template - it's just more restrictive about what you can do.

.github/workflows/scripts/control-tower-prcheck/run_control_tower_prcheck.py
PawelWMS: I'm intentionally leaving validation of the arguments to the endpoint because that's the source of truth. I figured duplicating them here doesn't may save us milliseconds and would require constant syncing whenever the CT changes.

.github/workflows/ado/templates/sources-upload-stages.yml
PawelWMS: For future reference: took a look at what it would take to secure a PR from forks for cases when the pipeline needs access to the internal service connection (this case). The pipeline would need to be modified in a few critical ways with potential changes to our ADO projects structure. Discussed this briefly with @reubeno and decided to only support non-fork PRs for the time being to unblock at least some source uploads scenarios.

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -82,12 +82,19 @@ stages:
               git config --unset extensions.worktreeConfig || true
 
               # Full history is needed for spec rendering to work.
-              git fetch --unshallow
+              if [ "$(git rev-parse --is-shallow-repository)" = "true" ]; then
+                git fetch --unshallow
+              fi
 
+              # TODO: Replace with using the TOML config to read the specs dir once available.
```
dmcilvaney, line 134: your call, but once the config is in, you don't need any flags at all, just `-a`. Everything else will be configured automatically.

dmcilvaney (approved):

jslobodzian (approved):

.github/workflows/ado/sources-upload.yml
```diff
@@ -0,0 +1,72 @@
...
+          variableGroup: ControlTower-PRCheck
+          # Must exceed the script's --poll-timeout-seconds (default 7200s = 120m)
+          # with enough headroom for setup steps and the final API call.
+          timeoutInMinutes: 150
```
carlapgavilan: why 30 minutes here?

.github/workflows/scripts/control-tower-prcheck/run_control_tower_prcheck.py
```diff
@@ -0,0 +1,506 @@
...
+            suffix = f" | {task_summary}" if task_summary else ""
+            print(
+                f"Job {job_id} status: {transition} (elapsed {elapsed}s){suffix}",
+                flush=True,
+            )
+            previous_status = current_status
+        else:
+            # Heartbeat so the user can see the script is alive and still polling.
+            print(
+                f"[heartbeat] waiting on job {job_id}, status={current_status}, elapsed={elapsed}s",
```
carlapgavilan: Do we want to print this every 10s? the --poll-timeout-seconds default = 7200s and the
--poll-interval-seconds default = 10s so that could be up to 720 heartbeats.

PR #16657, opened 2026-04-13
feat: remove vm-base.kiwi package kernel

ddstreet:
This is only a metapackage that pulls in other packages, and shouldn't be installed by default.

### DIFF
--- a/base/images/vm-base/vm-base.kiwi
+++ b/base/images/vm-base/vm-base.kiwi
@@ -82,7 +82,6 @@
         <package name="iproute" />
         <package name="iputils" />
         <package name="irqbalance" />
-        <package name="kernel" />
         <package name="kernel-tools" />
         <package name="lvm2" />
         <package name="lz4" />


reubeno:
Are we still guaranteed to get `kernel-modules` installed?

ddstreet:
> Are we still guaranteed to get `kernel-modules` installed?

it is [explicitly listed](https://github.com/ddstreet/azurelinux/blob/7dc1ace254a7692a2c5585012cd16fbd72e8a6b7/base/images/vm-base/vm-base.kiwi#L120)

rlmenge:
Agreed that `kernel` is just a meta package and that the kenrel is likely getting pulled by `kernel-modules` but is there a reason we should not have `kernel-core` explicitly?

ddstreet:
> Agreed that `kernel` is just a meta package and that the kenrel is likely getting pulled by `kernel-modules` but is there a reason we should not have `kernel-core` explicitly?

in the PR to switch over to fedora-based kiwi xml, [it's in there](https://github.com/ddstreet/azurelinux/blob/a40b07b150b3121abaf798687528d6a44c1d2292/base/images/common/boot.xml#L18), but this (and the other PRs modifying our `vm-base.kiwi `flat file) don't have it defined

maybe it's better to close out these prs against our kiwi flat file that aren't strictly needed, like this one. I'll close this out.

PR #16658, opened 2026-04-13
feat: Add distroless containers

tobiasb-ms:
This change adds three flavors of distroless containers: minimal, base and debug, modeled heavily on https://mcr.microsoft.com/en-us/artifact/mar/azurelinux/distroless/minimal/about. The implementation relies on profiles, keeping everything in a single kiwi files.

The main implementation is in `config.sh`, which is called by kiwi as part of the image build. It does a few things:
1. Copies necessary certificates from `ca-certificates` -- basically mimicking `prebuilt-ca-certificates` without needing to create that package.
2. For uninstalls `bash` and `ca-certificates`, even if they are dependencies of packages we want. They aren't "real" dependencies for the purposes of a distroless container.
3. Removes all packages that aren't declared as `image` packages for the profile (or their dependencies).
4. Cleans up some files from the disk.

The results are small containers with very little on them, to be used by teams like dotnet and golang to create their purpose-built containers.

### DIFF
--- a/base/images/container-base/config.sh
+++ b/base/images/container-base/config.sh
@@ -29,7 +29,7 @@
 #   "Performing autoremove"  — dnf's own transaction log
 
 test -f /.profile && . /.profile
-set -uo pipefail
+set -xuo pipefail
 
 echo "config.sh: building profile(s): ${kiwi_profiles:-<none>}"
 
--- a/base/images/container-base/config.sh
+++ b/base/images/container-base/config.sh
@@ -27,9 +27,6 @@
 #   "Force-removing ..."     — disallowed packages stripped before autoremove
 #   "All installed packages" — full list before stripping
 #   "Performing autoremove"  — dnf's own transaction log
-#
-# To investigate why a dependency survived, rebuild with the debug
-# profile and use: dnf repoquery --whatrequires <pkg>
 
 test -f /.profile && . /.profile
 set -uo pipefail
--- a/base/images/container-base/config.sh
+++ b/base/images/container-base/config.sh
@@ -1,0 +1,269 @@
+#!/bin/bash
+# config.sh — Kiwi config.sh hook for container image minimization
+#
+# Strips the image down to only the packages declared in the kiwi XML
+# (plus their transitive dependencies), producing a minimal "distroless"
+# container.
+#
+# Strategy:
+#   1. Parse the kiwi XML to collect declared image packages
+#      (shared base + profile-specific).
+#   2. Resolve virtual Provides to real package names
+#      (e.g., system-release → azurelinux-release).
+#   3. Force-remove disallowed packages (bash, ca-certificates) while
+#      preserving generated certificate files via snapshot/restore.
+#   4. Mark all installed packages as auto-dependencies, re-mark
+#      declared packages as user-installed, then dnf autoremove
+#      strips everything else — including dnf and rpm themselves.
+#   5. Post-autoremove cleanup via statically-linked busybox: remove
+#      package manager state, locale stubs, docs, and leftover dirs.
+#
+# Only type="image" packages are considered — bootstrap packages are
+# used by kiwi only during initial chroot setup and are not part of
+# the final image definition.
+#
+# Output sections (for debugging):
+#   "Packages to KEEP"       — resolved names from the kiwi XML
+#   "Force-removing ..."     — disallowed packages stripped before autoremove
+#   "All installed packages" — full list before stripping
+#   "Performing autoremove"  — dnf's own transaction log
+#
+# To investigate why a dependency survived, rebuild with the debug
+# profile and use: dnf repoquery --whatrequires <pkg>
+
+test -f /.profile && . /.profile
+set -uo pipefail
+
+echo "config.sh: building profile(s): ${kiwi_profiles:-<none>}"
+
+# ---------------------------------------------------------------------------
+# Only run for distroless profiles
+# ---------------------------------------------------------------------------
+# kiwi_profiles is a comma-separated list set by kiwi at build time
+# (e.g., "distroless-minimal").  Skip this entire script for non-distroless
+# builds (e.g., "container-base") — they don't need image stripping.
+if [[ ! "${kiwi_profiles:-}" =~ ^distroless ]]; then
+    echo "config.sh: profile '${kiwi_profiles:-}' does not start with 'distroless' — skipping."
+    exit 0
+fi
+
+# ---------------------------------------------------------------------------
+# Helper functions
+# ---------------------------------------------------------------------------
+
+# Print a message to stderr and exit with an error code.
+die() {
+    echo "ERROR: $*" >&2
+    exit 1
+}
+
+# Extract package names from a <packages type="image"> section of the kiwi
+# XML definition.  Returns one package name per line.
+# $1: XPath predicate to further filter the <packages> element
+#     (e.g., 'not(@profiles)' or "@profiles='distroless-debug'")
+get_image_packages() {
+    xmllint --xpath \
+        "//packages[@type='image' and ${1}]/package/@name" \
+        /image/config.xml 2>/dev/null \
+    | grep -oP 'name="\K[^"]+' \
+    || true  # No matches is not an error — the profile may have no packages.
+}
+
+# ---------------------------------------------------------------------------
+# 1. Collect declared image packages from the kiwi XML
+# ---------------------------------------------------------------------------
+
+# Shared packages (no profiles attribute) apply to all profiles.
+KEEP_PKGS=$(get_image_packages 'not(@profiles)')
+
+# Packages scoped to specific profiles.  kiwi_profiles is a comma-separated
+# list of active profiles, set by kiwi at build time (e.g.,
+# "distroless-debug,distroless-core,distroless-minimal").
+IFS=',' read -ra PROFILES <<< "${kiwi_profiles:-}"
+for profile in "${PROFILES[@]}"; do
+    KEEP_PKGS+=$'\n'$(get_image_packages "@profiles='${profile}'")
+done
+
+# Deduplicate.
+KEEP_PKGS=$(echo "${KEEP_PKGS}" | sort -u | grep -v '^$') || true
+
+if [[ -z "${KEEP_PKGS}" ]]; then
+    die "no packages found to keep — aborting to avoid stripping the entire image"
+fi
+
+# ---------------------------------------------------------------------------
+# 2. Resolve virtual Provides to installed package names
+# ---------------------------------------------------------------------------
+# Some kiwi entries reference virtual provides (e.g., "system-release")
+# rather than concrete package names (e.g., "azurelinux-release"). Resolve
+# them so that dnf mark operates on real package names.
+
+RESOLVED_KEEP_PKGS=""
+while IFS= read -r pkg; do
+    if rpm -q "${pkg}" &>/dev/null; then
+        # Concrete package name — already installed.
+        RESOLVED_KEEP_PKGS+="${pkg}"$'\n'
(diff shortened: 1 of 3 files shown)


base/images/container-base/container-base.kiwi
```diff
@@ -6,6 +6,16 @@
         <contact>user@example.com</contact>
         <specification>Azure Linux Container Image Example</specification>
     </description>
+    <profiles>
```
reubeno, line 9: I'm less familiar with our distroless containers from AZL3. Can we take the opportunity to document what the *intended* use / requirements are for the different distroless-* images, even if only briefly?

base/images/container-base/container-base.kiwi
```diff
@@ -26,16 +36,38 @@
...
+    <packages type="image" profiles="distroless-minimal">
         <package name="filesystem" />
         <package name="system-release" />
+        <package name="tzdata" />
+    </packages>
+
+    <packages type="image" profiles="distroless-base">
+        <package name="glibc" />
+        <package name="libgcc" />
+        <package name="openssl-libs" />
```
reubeno, line 57: Why are `openssl-libs` expected to be present?

base/images/container-base/config.sh
```diff
@@ -0,0 +1,269 @@
...
+# Output sections (for debugging):
+#   "Packages to KEEP"       — resolved names from the kiwi XML
+#   "Force-removing ..."     — disallowed packages stripped before autoremove
+#   "All installed packages" — full list before stripping
+#   "Performing autoremove"  — dnf's own transaction log
+#
+# To investigate why a dependency survived, rebuild with the debug
+# profile and use: dnf repoquery --whatrequires <pkg>
+
+test -f /.profile && . /.profile
```
reubeno, line 31: What's this about?

base/images/container-base/config.sh
```diff
@@ -0,0 +1,269 @@
...
+set -uo pipefail
```
reubeno: Is this omission of `-e` intentional? Do we have guards/checks for errors from all commands?

base/images/container-base/config.sh
```diff
@@ -0,0 +1,269 @@
...
+}
+
+# Extract package names from a <packages type="image"> section of the kiwi
+# XML definition.  Returns one package name per line.
+# $1: XPath predicate to further filter the <packages> element
+#     (e.g., 'not(@profiles)' or "@profiles='distroless-debug'")
+get_image_packages() {
+    xmllint --xpath \
+        "//packages[@type='image' and ${1}]/package/@name" \
+        /image/config.xml 2>/dev/null \
```
reubeno, line 64: I take it that `/image/config.xml` is a contractual/documented location that kiwi promises to have the final expanded XML file in?

base/images/container-base/config.sh
```diff
@@ -0,0 +1,269 @@
...
+echo "=== Packages to KEEP (from kiwi XML) ==="
+echo "${RESOLVED_KEEP_PKGS}"
+echo ""
+
+# ---------------------------------------------------------------------------
+# 3a. Force-remove disallowed packages
+# ---------------------------------------------------------------------------
+echo "=== Force-removing disallowed packages ==="
+
+# --- ca-certificates ---
```
reubeno, line 136: How does this logic compare with how we produce `prebuilt-ca-certificates` in AZL3?

base/images/container-base/config.sh
reubeno: Is it worth always including `-x` for debugging purposes when things go wrong, or does stdout not get routed out somewhere useful?

base/images/container-base/config.sh
```diff
@@ -0,0 +1,269 @@
+#!/bin/bash
+# config.sh — Kiwi config.sh hook for container image minimization
```
reubeno, line 2: There's enough complexity here that I think we'll definitely want to lean on static validation of the produced images to ensure that they have the "right results" -- perhaps even checking directly against an expected file list. (Runtime checks too by running things in them, but I know that's another story.)

base/images/images.toml
```diff
@@ -4,7 +4,19 @@ definition = { type = "kiwi", path = "vm-base/vm-base.kiwi" }
 
 [images.container-base]
 description = "Container Base Image"
-definition = { type = "kiwi", path = "container-base/container-base.kiwi" }
+definition = { type = "kiwi", path = "container-base/container-base.kiwi", profile = "core" }
```
reubeno, line 7: Shall we align on *either* `base` or `core` for the image name and have the profile match it? It may be confusing to use different terms.

base/images/images.toml
```diff
@@ -4,7 +4,19 @@ definition = { type = "kiwi", path = "vm-base/vm-base.kiwi" }
...
+
+[images.container-distroless-minimal]
```
reubeno, line 9: Would you be able to post to the PR somewhere the actual resulting file lists for what's contained in each of the images with this composition when built? For bonus points, knowing how that compares with AZL3 would be helpful too.

base/images/container-base/container-base.kiwi
tobiasb-ms: Do you have that in mind for the kiwi file, a wiki, a more public document or something else? That will guide who writes it, when, and whether it's part of this PR.

base/images/container-base/container-base.kiwi
tobiasb-ms: In AZL3 (as described [here](https://mcr.microsoft.com/en-us/artifact/mar/azurelinux/distroless/minimal/about)), we actually have both `openssl-libs` and `openssl` itself. There have been various requests to remove `openssl` as it bloats the image and is really only a command-line tool. However, the users of distroless appear to want `openssl-libs` there already (except for `minimal`).

More generally, we're mostly just following AZL3.

base/images/container-base/config.sh
tobiasb-ms: Not sure which part you're talking about here.

The comments were from an earlier iteration that I missed during self-review. Removed.

The sourcing of `/.profile` gives us access to kiwi-defined variables. You can see something similar in the [fedora kiwi source](https://pagure.io/fedora-kiwi-descriptions/blob/f43/f/config.sh#_9) (we don't seem to need `.kconfig`). See [this doc](https://osinside.github.io/kiwi/concept_and_workflow/shell_scripts.html#profile-environment-variables) for information about the variables in `.profile`.

base/images/container-base/config.sh
tobiasb-ms: It is intentional. There are a few cases where we don't want it -- especially the best-effort cleanup parts.

That said, I went back and forth about using `-e` and then either disabling it or using the standard ways to get around it where we wanted to vs. what I have. I'm honestly happy with either method -- do you have a strong opinion?

base/images/container-base/config.sh
tobiasb-ms: It's a little iffy to be honest. It's referenced in [the docs](https://osinside.github.io/kiwi/concept_and_workflow/shell_scripts.html#profile-environment-variables), and it's certainly there.

According to the [general image process doc](https://osinside.github.io/kiwi/concept_and_workflow/shell_scripts.html#profile-environment-variables), `config.sh` is called at the end of the `prepare` step, which sets up the chroot, including `/image/config.xml`.

That said, there is not an actual guarantee, which does make this fragile. I went with this approach so that the packages we explicitly want to keep are declared in the kiwi file per profile.

An alternate approach would be to have those lists in `config.sh` itself. That's actually technically easier, but it forces us to remember to look in a non-obvious place for that information. On balance, I went with the slightly more fragile version in the interest of making it a little easier to see our intentions.

Thoughts?

base/images/container-base/config.sh
tobiasb-ms: That logic is [here](https://github.com/microsoft/azurelinux/blob/4cc5012da4e09d7ee6d9f879c35f049edd0a5622/SPECS/prebuilt-ca-certificates/prebuilt-ca-certificates.spec#L33-L42).

It's largely the same. The main difference is that Fedora's version of `ca-certificates` includes a directory of cert hashes that is evidently kind of an obscure way to look up certs used in certain `openssl` libraries, and was taking quite a bit of space (I don't currently have the numbers). Given that current users of the AZL3 version aren't using that, I figured it was ok to remove them.

base/images/container-base/config.sh
tobiasb-ms: It's iffy. Because `kiwi` splits `stdout` and `stderr`, following the trace is much harder than it should be.

That said, it's probably a better idea to have **something** rather than nothing. I'll add that.

base/images/container-base/config.sh
tobiasb-ms: I agree. I consider that to be a separate issue than this PR (and likely won't live in this repo). Agree?

base/images/images.toml
tobiasb-ms: Yes we should.

Personally I prefer `core`. So profile is `core`, name is `container-core` and description is `Container Core Image`.

That leaves the question of whether want `distroless-base` (which matches AZL3) or `distroless-core`, which I personally prefer.

Thoughts?

base/images/container-base/container-base.kiwi
```diff
@@ -26,16 +36,38 @@
...
     </repository>
 
-    <packages type="image">
-        <!-- Core group -->
-        <package name="basesystem" />
+    <packages type="image" profiles="core">
+        <package name="filesystem" />
         <package name="bash" />
         <package name="system-release" />
     </packages>
```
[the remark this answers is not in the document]
tobiasb-ms: I'm comfortable requiring a profile.

base/images/container-base/container-base.kiwi
reubeno: I was thinking even just a 1 or 2 line comment in the kiwi file and/or TOML file.

base/images/container-base/config.sh
reubeno: I don't have a strong opinion -- mostly want to make sure you'd thought about it.

base/images/container-base/config.sh
reubeno: I'm okay with the reference. Having a clear check + error if the file can't be found might be helpful, particularly considering we're throwing away stderr here, though.

base/images/container-base/config.sh
reubeno: I think it *will* live in this repo, but can definitely be a separate PR.

base/images/images.toml
reubeno: I don't have a strong preference.

To me `base` implies something that's a foundational layer underneath other things -- and `core` is somewhat overlapping, implying some minimal nucleus.

base/images/container-base/config.sh
christopherco: https://dev.azure.com/mariner-org/mariner/_workitems/edit/18790

base/images/container-base/config.sh
christopherco: https://dev.azure.com/mariner-org/mariner/_workitems/edit/18791

base/images/container-base/container-base.kiwi
christopherco: https://dev.azure.com/mariner-org/mariner/_workitems/edit/18792

base/images/images.toml
christopherco: https://dev.azure.com/mariner-org/mariner/_workitems/edit/18793

base/images/images.toml
christopherco: https://dev.azure.com/mariner-org/mariner/_workitems/edit/18794

christopherco (approved):

PR #16660, opened 2026-04-14
feat: remove vm-base.kiwi explicitly listed compression packages

ddstreet:
This removes the explicit listing of compression packages, with the exception of gzip and bzip2.

### DIFF
--- a/base/images/vm-base/vm-base.kiwi
+++ b/base/images/vm-base/vm-base.kiwi
@@ -55,7 +55,6 @@
         <package name="bc" />
         <package name="binutils" />
         <package name="bind-utils" />
-        <package name="brotli" />
         <package name="bzip2" />
         <package name="chkconfig" />
         <package name="chrony" />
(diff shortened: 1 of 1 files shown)


base/images/vm-base/vm-base.kiwi
```diff
@@ -106,13 +104,10 @@
         <package name="systemd-rpm-macros" />
         <package name="tar" />
         <package name="tpm2-tss" />
-        <package name="unzip" />
```
reubeno, line 109: What's the thinking for removing unzip but not tar?

ddstreet: tar is widely used directly, while unzip isn't. at least in the linux world...

i suppose azl4 may see more 'unzip' usage due to its proximity to the windows world.

reubeno: Yes, that's probably true. VM extensions downloaded by wala are zipped; not sure if it depends on `zip`/`unzip` or uses the libs internally.

Edit: I'm guessing wala may use python's in-built Zip functionality?

Second edit: looks like unzip is, not, in fact in AZL3.0 marketplace image. Several others are, though.

ddstreet:
This isn't important to get merged, a few extra compression packages are not a big deal, so let's close this out.

PR #16661, opened 2026-04-14
feat: remove vm-base.kiwi package file

ddstreet:
The file package will be pulled in by other packages (e.g. grub2-tools, grub2-pc, etc) and should not be explicitly listed.

### DIFF
--- a/base/images/vm-base/vm-base.kiwi
+++ b/base/images/vm-base/vm-base.kiwi
@@ -71,7 +71,6 @@
         <package name="dracut-kiwi-oem-repart" />
         <package name="elfutils" />
         <package name="elfutils-default-yama-scope" />
-        <package name="file" />
         <package name="firewalld" />
         <package name="glibc" />
         <package name="glibc-langpack-en" />


base/images/vm-base/vm-base.kiwi
```diff
@@ -71,7 +71,6 @@
         <package name="dracut-kiwi-oem-repart" />
         <package name="elfutils" />
         <package name="elfutils-default-yama-scope" />
-        <package name="file" />
```
reubeno, line 74: I don't think we should trim redundant runtime dependencies if we would intentionally pull in the package even had it not been pulled in. In other words, I think this list should capture our intent.

Do you think `file` shouldn't be included if it's not otherwise required as a dependency? I'd imagine plenty of scripts assume it's present.

ddstreet: > Do you think file shouldn't be included if it's not otherwise required as a dependency

Right, I don't think it should be included other than as a dependency.

> I'd imagine plenty of scripts assume it's present

I don't think that's true, except for scripts in packages that actually list 'file' as a requirement, but in any case it's hardly worth arguing over, as the package will be pulled in by (at least) the grub2 packages, so it doesn't *currently* make any difference if we list it explicitly or not.

PR #16662, opened 2026-04-14
feat: replace vm-base.kiwi package ncurses-term with ncurses

ddstreet:
The ncurses-term package contains 'additional terminal descriptions' which we don't need to include by default; what we need is the main ncurses package.

### DIFF
--- a/base/images/vm-base/vm-base.kiwi
+++ b/base/images/vm-base/vm-base.kiwi
@@ -88,6 +88,7 @@
         <package name="lz4" />
         <package name="man-db" />
         <package name="nano" />
+        <package name="ncurses" />
         <package name="ncurses-term" />
         <package name="net-tools" />
         <package name="nftables" />


base/images/vm-base/vm-base.kiwi
```diff
@@ -88,7 +88,7 @@
         <package name="lz4" />
         <package name="man-db" />
         <package name="nano" />
-        <package name="ncurses-term" />
+        <package name="ncurses" />
```
reubeno, line 91: I see that `ncurses` pulls in `ncurses-base` (via `ncurses-libs`), and `ncurses-base` includes terminfo for the most prevalent terminal emulators / TERM values (e.g., xterm-256color).

Is the thinking here that any user using something more interesting will need to know to install `ncurses-term`? For example, I see that `ncurses-term` may be required for better experience on `ghostty`, `iTerm2`, et al. I don't know what we've done in the past, but that sounds reasonable to me.

ddstreet: > Is the thinking here 

honestly, my thinking here is that everyone else includes `ncurses`, `ncurses-base`, and `ncurses-libs` in their image, while we are the only one who included `ncurses-term`; and we also did not include `ncurses`. so we don't actually have useful terminal commands like 'clear' and 'reset'.

I don't know why we included `ncurses-term`, but it seems like it was a mistake because `ncurses` is what we actually want.

> may be required for better experience on ghostty, iTerm2, et al. 

This is something we should include by default in our image? Have we had any customers actually ask for this? I'd expect almost all customers to be using either standard linux or windows, and `ncurses-base` does include putty term support.

I guess I don't really have strong feelings about it, but it seems pretty unlikely to actually be needed as a _default_ part of our images.

ddstreet:
updated to leave the `ncurses-term` package in the image

christopherco (approved):
I like the addition here. Will file a bug to evaluate whether to keep or remove `ncurses-term`

base/images/vm-base/vm-base.kiwi
christopherco: https://dev.azure.com/mariner-org/mariner/_workitems/edit/18795

PR #16664, opened 2026-04-14
qemu: Enable user_static builds for providing user mode emulation.

Sumynwa:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This PR introduces changes to enable user_static builds in qemu to build qemu-user-static* arch specific packages for providing user mode emulation. This enables support to run for ex aarch64 bins using qemu-user-static-aarch64 sub-package via binfmt-misc on x86_64 host. By default, the binfmt config is set with 'F = fix binary' flag to preload the interpreter binary to allow running non-native instructions inside the container environment.

The sub-package `qemu-user-binfmt` also provides binfmt config but they are dynamic interpreter configurations.
For ex qemu-aarch64, the interpreter registered is "/usr/bin/qemu-aarch64" which is a dynamically linked shared object.
Also, the interpreter is registered without the F = fix binary flag.
We cannot use this package for container-based builds because:
- there are no entries inside the /proc/sys/fs/binfmt_misc IN the containers mount ns
- when installing arm64 packages, the pre/post scriptlets start to fail when trying to run without the 'F' flag
    - because there is no binfmt_misc registered inside container ns
    - in host, the interpreter is not loaded since without F indicates lazy loading by kernel
    - tries to execute arm64 instructions natively and fails

###### Change Log  

- Enable user_static global compile flag
- configure user_static with 'enable_pie'
- manage systemd-binfmt restarts post install/uninstall to avoid start-limit-hit
- add conflicts between qemu-user-static and qemu-user-binfmt as both
  provide binfmt rules

NO

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: [Buddy Build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1094245&view=results)

### DIFF
--- a/SPECS/qemu/qemu.spec
+++ b/SPECS/qemu/qemu.spec
@@ -1170,10 +1170,6 @@ Requires: %{name}-user = %{version}-%{release}
 Requires(post): systemd-units
 Requires(postun): systemd-units
 # qemu-user-binfmt + qemu-user-static both provide binfmt rules
-# Temporarily disable to get fedora CI working. Re-enable
-# once this CI issue let's us deal with subpackage conflicts:
-# https://pagure.io/fedora-ci/general/issue/184
-#
 # introduce conflicts to not install both binfmt rules.
 Conflicts: qemu-user-static
 %description user-binfmt
--- a/SPECS/qemu/qemu.spec
+++ b/SPECS/qemu/qemu.spec
@@ -66,9 +66,6 @@ Distribution:   Azure Linux
 
 
 %global user_static 1
-%if 0%{?azl}
-%global user_static 0
-%endif
 
 %global have_kvm 0
 %if 0%{?kvm_package:1}
@@ -435,7 +432,7 @@ Obsoletes: sgabios-bin <= 1:0.20180715git-10.fc38
 Summary: QEMU is a FAST! processor emulator
 Name: qemu
 Version: 9.1.0
-Release: 5%{?dist}
+Release: 6%{?dist}
 License: Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND FSFAP AND GPL-1.0-or-later AND GPL-2.0-only AND GPL-2.0-or-later AND GPL-2.0-or-later WITH GCC-exception-2.0 AND LGPL-2.0-only AND LGPL-2.0-or-later AND LGPL-2.1-only AND LGPL-2.1-or-later AND MIT AND LicenseRef-Fedora-Public-Domain AND CC-BY-3.0
 URL: http://www.qemu.org/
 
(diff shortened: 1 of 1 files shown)


SPECS/qemu/qemu.spec
```diff
@@ -1153,7 +1150,9 @@ Requires(postun): systemd-units
 # Temporarily disable to get fedora CI working. Re-enable
```
jiria: Should we remove from Temporarily disable...?

jiria (approved):

SPECS/qemu/qemu.spec
```diff
@@ -1197,8 +1192,10 @@ Requires(postun): systemd-units
...
```
harshitgupta1337: Should we remove these comments?

harshitgupta1337 (commented):
Looks good to me. Just a nit comment.

SPECS/qemu/qemu.spec
Sumynwa: Yes, I forgot them here, removed the earlier snippet. Will remove.

Sumynwa: Fixed.

kgodara912:
[Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1112640&view=results) after recent changes.

SPECS/qemu/qemu.spec
```diff
@@ -2384,96 +2378,190 @@ useradd -r -u 107 -g qemu -G kvm -d / -s /sbin/nologin \
...
+%define try_run_systemd_binfmt() \
+if ! /bin/systemctl is-active --quiet systemd-binfmt.service; then \
+    /bin/systemctl --system restart systemd-binfmt.service &>/dev/null || : \
+else \
+    /bin/systemctl --system try-restart systemd-binfmt.service &>/dev/null || : \
+fi \
+%{nil}
+
+%pretrans user-static
+touch %{setbinfmtonce}
```
kgodara912, line 2394: The parent directory may not be available in some scenarios, for example in containers, should we try to create the directory first, mkdir -p to avoid file exists, and then touch the file so that file creation doesn't fail or we are not going to support containers altogether as PR description says?

kgodara912 (changes requested):
As such changes look fine, just a minor clarification needed.

SPECS/qemu/qemu.spec
Sumynwa: Yes, we should create it just in case.
This works in vm install since it is already created by initramfs.
I will make changes.

Sumynwa: Changes done.

kgodara912 (approved):
Changes are approved by shiproom. Corrections in `BuildRequires` and `%end` placement are good as well as enablement of user_static builds. A race condition may come between %pretran and %posttrans where file is created and installation is interrupted but that is understood for package installations. Buddy build is successful. LGTM.

PR #16669, opened 2026-04-14
feat(dracut): add dracut-hyperv subpackage for Azure VM initramfs

rlmenge:
Summary
Add a dracut-hyperv subpackage to the dracut component via overlays, ensuring all Hyper-V guest kernel modules are included in the initramfs at image build time. This mirrors the approach used in Azure Linux 3.0. (https://github.com/microsoft/azurelinux/pull/9326)

Motivation
Dracut's generic mode (used during image builds) only discovers a subset of Hyper-V modules via its built-in hardware scan:

In initramfs by default: hv_vmbus, hv_storvsc, pci-hyperv, hid-hyperv, hyperv-keyboard
Missing from initramfs: hv_netvsc, hv_utils, hv_balloon, hv_sock
The missing modules are loaded later via udev after switch_root, which is sufficient for basic boot. However, including them in the initramfs ensures early availability of networking, guest services (heartbeat, KVP, shutdown), dynamic memory, and vsock — which is important for Azure cloud-init, provisioning, and guest agent communication. See https://learn.microsoft.com/en-us/troubleshoot/azure/virtual-machines/linux/linux-hyperv-issue

Fedora Cloud gets away without this because those modules load ~12 seconds into boot from the real root filesystem, but for an enterprise distro targeting Azure, early availability seems ideal.

Changes
- Move dracut from inline components.toml to a dedicated dracut.comp.toml
- Add 5 overlays to create the dracut-hyperv subpackage:
- file-add: ships 00-hyperv.conf (dracut config with add_drivers+=)
- spec-insert-tag: adds Source9999: 00-hyperv.conf
- spec-append-lines x3: adds %package hyperv, %install, and %files hyperv sections
- Include dracut-hyperv in vm-base.kiwi image definition

Testing
- azldev comp prep-sources — overlays apply cleanly
- azldev comp build -p dracut — builds successfully, produces dracut-hyperv-107-8.azl4.x86_64.rpm
- Mock chroot smoke test — RPM installs, contains /etc/dracut.conf.d/00-hyperv.conf with correct content
- Full image build (azldev image build vm-base --local-repo) — succeeds
- lsinitrd on built image — confirms all 9 Hyper-V modules present in initramfs (6 from 00-hyperv.conf + 3 already caught by generic mode)


rlmenge:
These are not strictly needed in the initramfs therefore closing

PR #16674, opened 2026-04-14
feat(spec render): Add supporting infrastructure for rendered specs

dmcilvaney:
- CI pipeline that validates the state of the files
- Update agent instructions to understand rendered files
- Add instructions about writing workflows


.github/workflows/check-rendered-specs-stub.yml
```diff
@@ -0,0 +1,43 @@
...
+# It delegates all real work to the reusable template on tomls/base/main.
+#
+# This two-stage design lets fork PRs trigger the check safely: the stub runs in the
+# context of the default branch (with write token), but the reusable workflow checks out
+# the PR's data files (TOML configs, specs) into a separate directory — never mixing
+# untrusted code with execution context.
+#
+# The stub must exist on the default branch because pull_request_target always runs
+# workflows from there. The reusable workflow on tomls/base/main has the actual scripts,
+# container setup, and rendering logic.
```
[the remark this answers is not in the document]
dmcilvaney: This is correct, default branch is 3.0.

.github/workflows/scripts/check_rendered_specs.py
```diff
@@ -0,0 +1,515 @@
...
+    hide newly rendered files and make the check green. We also drop any
+    `.gitignore`/`.gitattributes` files found under specs_dir since they
+    have no business in a rendered-output tree.
+    """
+    sd = str(specs_dir)
+    changed = _git_lines_z("diff", "-z", "--diff-filter=M", "--name-only", "--", sd)
+    # No --exclude-standard: list ALL untracked files so a PR-committed
+    # .gitignore under specs_dir can't hide drift.
+    extra_raw = _git_lines_z("ls-files", "-z", "--others", "--", sd)
+    missing = _git_lines_z("ls-files", "-z", "--deleted", "--", sd)
```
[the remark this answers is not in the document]
dmcilvaney: Not reproable in practice

base/project.toml
```diff
@@ -10,3 +10,4 @@ log-dir = 'build/logs'
 work-dir = 'build/work'
 output-dir = 'out'
 default-distro = { name = "azurelinux" }
+rendered-specs-dir = '../specs'
```
PawelWMS, line 13: Nit-pick: do we have a "projectRoot" kind of variable we can use to avoid relative paths just in case this TOML moves around?

AGENTS.md
```diff
@@ -33,9 +33,10 @@ Do NOT skip testing for changes that affect RPM output. Do NOT tell the user "th
 
 - Always run `azldev comp list -p <name> -q -O json` before modifying a component.
 - Prefer overlays over forking/local specs when customizing upstream packages.
-- Use `azldev comp prep-sources -p <name> --force -o <dir> -q` to verify overlays apply cleanly before building. Always use `--force` to overwrite an existing output dir, `rm -rf` requires user confirmation which is disruptive.
-- Follow the inner loop cycle: investigate → modify → verify → build → test → inspect. See [`skill-build-component`](.github/skills/skill-build-component/SKILL.md).
-  - Note: Use your best judgement, some packages are VERY slow to build (e.g., `kernel`), in those cases you may want to do multiple iterations of investigate → modify → verify with `prep-sources` before doing a full build + test.
+- After modifying overlays or component config, re-render with `azldev comp render -p <name>` and inspect `specs/<first-char>/<name>/` to verify the result. This is the fastest verification path.
```
PawelWMS, line 36: Should there be a separate case instruction for when we don't touch any component or overlay explicitly but manipulate the timestamp date?

.github/prompts/azl-add-component.prompt.md
```diff
@@ -19,6 +19,6 @@ Follow the workflow in the [skill-add-component skill](../skills/skill-add-compo
    - Needs overlays or customizations → create `${input:component_name}/${input:component_name}.comp.toml`
    - Needs extensive changes overlays can't handle → forked local spec (**last resort**, requires explicit user sign-off)
 6. Add overlays with meaningful `description` fields explaining *why* each change is needed
-7. Validate: `azldev comp prep-sources -p ${input:component_name} --force -o base/build/work/scratch/${input:component_name}-post -q` (with overlays) and diff against the skip-overlays output
+7. Render and verify: `azldev comp render -p ${input:component_name}` and inspect `specs/` output. For deeper debugging, diff pre/post overlay output with `prep-sources`.
```
PawelWMS: For a single source of truth, maybe point the agent to the read the `rendered-specs-dir` setting from the TOML.

.github/prompts/azl-update-component.prompt.md
```diff
@@ -18,6 +18,10 @@ Use structural patterns from [comp-toml.instructions.md](../instructions/comp-to
    - `config` — build config changes (`build.defines`, `build.without`)
 3. **Apply changes** to the `.comp.toml` file
 4. **Verify overlays still apply:**
+   ```bash
+   azldev comp render -p ${input:component_name}
+   ```
+   Inspect `specs/` output. For deeper debugging, diff pre/post overlay output:
```
PawelWMS: Same comment here: I'd point it at the `rendered-specs-dir` config instead of a raw string.

.github/instructions/rendered-specs.instructions.md
```diff
@@ -0,0 +1,32 @@
...
+To change a rendered spec, modify the component's `.comp.toml` and/or its overlays. Then re-run the render command to regenerate the spec:
+
+```bash
+# VERY SLOW - Re-render all specs, removes any stale specs that are no longer defined in the components
+azldev comp render -a --clean-stale -O json
+```
+
+```bash
+# Small set, will NOT remove stale specs, faster for iterative development
+azldev comp render <component-1> <component-2> -O json
```
PawelWMS: Should we include the `-p` flag? I've seen it elsewhere.

```suggestion
azldev comp render -p <component-1> <component-2> -O json
```

.github/instructions/pr-check-workflows.instructions.md
```diff
@@ -0,0 +1,112 @@
+---
+applyTo: ".github/workflows/**"
+description: ALWAYS review these instructions when reading or modifying PR check workflows.
```
PawelWMS:
```suggestion
description: ALWAYS review these instructions when reading or modifying PR check workflows, or any scripts referenced by the workflows.
```

base/project.toml
dmcilvaney: I don't think so, but I'll have a double check

AGENTS.md
dmcilvaney: not a bad idea, I'll probably leave it informational though, (ie "changing timestamp can cause effective commit to change").

.github/prompts/azl-add-component.prompt.md
dmcilvaney: This is a balancing act, if you don't give it the actual path it requires a tool call or search to find the data, but agreed it could get stale. I'll probably add a note about where the directory is configured, but also list the current value.

base/project.toml
dmcilvaney: Not currently supported.

.github/instructions/rendered-specs.instructions.md
PawelWMS: It's also missing in the command below. I'd check all instructions for `-p`.

.github/workflows/containers/azldev-runner.Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM mcr.microsoft.com/azurelinux/base/core:3.0
+
+# Generic azldev runner image for CI PR checks. Provides the toolchain
+# required to run arbitrary `azldev` subcommands (render, build, ...)
+# against an untrusted PR checkout.
+#
+# Callers are expected to bind-mount:
+#   /workdir   : PR checkout (typically rw — azldev writes specs/ and build/)
+#   /output    : trusted-shape outputs produced by the container (ro on host)
+#   /gobin     : `go env GOPATH`/bin with `azldev` installed (ro)
```
PawelWMS: Curious: since it's an image explicitly for running `azldev`, why not install it as part of image creation with an arg specifying the branch/commit/release for `azldev`? This way we could drop the `gobin` mount.

.github/workflows/check-rendered-specs-stub.yml
```diff
@@ -0,0 +1,43 @@
+# Stub workflow — A copy of this workflow must live on the default branch (3.0) so that the
```
PawelWMS, line 1: Is mentioning the 3.0 branch intentional here or just a leftover?

.github/workflows/check-rendered-specs-stub.yml
```diff
@@ -0,0 +1,43 @@
...
+permissions: {}
+
+concurrency:
+  group: render-check-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    # Prevent forks from running a stale/vulnerable copy of this stub with Actions enabled
+    if: github.repository == 'microsoft/azurelinux'
```
PawelWMS, line 33: I'm not sure I get this part. What does it help with and won't this cause inconsistent behaviour for forks?

.github/workflows/check-rendered-specs.yml
```diff
@@ -0,0 +1,210 @@
...
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        with:
+          go-version: "stable"
+
+      - name: Install azldev
+        run: go install github.com/microsoft/azure-linux-dev-tools/cmd/azldev@main
```
PawelWMS: Can you add a comment explaining we want to use `main` for the time being, because the release process for `azldev` is not defined, yet. Plus the issue number tracking the task to switch to a stable release, once we have it.

.github/workflows/check-rendered-specs.yml
```diff
@@ -0,0 +1,210 @@
...
+
+      - name: Build azldev runner container
+        run: |
+          docker build \
+            --build-arg UID=$(id -u) \
```
PawelWMS, line 76: Curious: do we even care about the user ID, if it's inside the container?

.github/workflows/containers/azldev-runner.Dockerfile
PawelWMS: I think we may also be able to skip setting the `PATH` with `-e` when the container is run.

.github/workflows/check-rendered-specs.yml
dmcilvaney: File ownership gets weird with outputs if you don't map it correctly (esp. if the container is root).

.github/workflows/check-rendered-specs.yml
dmcilvaney: will do

.github/workflows/check-rendered-specs.yml
```diff
@@ -0,0 +1,210 @@
...
+            render-output/render-check-report.json
+
+  # Post the PR comment. Runs in a separate job so the `pull-requests: write`
+  # token is only granted to a job that does NOT execute any PR-derived code:
+  # it only checks out the trusted base branch (for the poster script) and
+  # consumes the trusted-shape report artifact from the render job.
+  comment:
+    name: Post drift comment
+    needs: render
+    if: always() && needs.render.result != 'cancelled'
```
PawelWMS, line 161: Curious if this wouldn't suffice since the `&&` part implies it's not `always()`:

```suggestion
    if: needs.render.result != 'cancelled'
```

.github/workflows/scripts/post_render_comment.py
```diff
@@ -0,0 +1,466 @@
...
+    parser.add_argument(
+        "--artifacts-url", default=None, help="Direct URL to patch artifact"
+    )
+    parser.add_argument("--run-id", default=None, help="GitHub Actions run ID")
+    args = parser.parse_args()
+
+    try:
+        with open(args.report, encoding="utf-8") as f:
+            report = json.load(f)
+    except (FileNotFoundError, json.JSONDecodeError) as exc:
```
PawelWMS, line 415: Why only catch a missing file? We could print consistent errors for all exceptions (example: the file is there but not a valid JSON).

.github/workflows/scripts/post_render_comment.py
```diff
@@ -0,0 +1,466 @@
...
+
+    # Write to GitHub step summary if available
+    summary_file = os.environ.get("GITHUB_STEP_SUMMARY")
+    if summary_file and total > 0:
+        body = format_comment(
+            report,
+            artifacts_url=args.artifacts_url,
+            run_id=args.run_id,
+            repo=args.repo,
+        )
```
PawelWMS: Why regenerate `body`? The arguments to `format_comment` look identical to the first call, so we should get the same results, right?

.github/workflows/scripts/post_render_comment.py
```diff
@@ -0,0 +1,466 @@
...
+        max_summary = 1_000_000  # GH step summary limit is 1024 KiB
```
PawelWMS, line 443: Considering the comment, would this be more accurate:

```suggestion
        max_summary = 1_048_576 # GH step summary limit is 1024 KiB
```

.github/workflows/check-rendered-specs-stub.yml
dmcilvaney: No, its intentional. I'll need to make a follow up PR to add a copy of this file to 3.0

.github/workflows/scripts/post_render_comment.py
```diff
@@ -0,0 +1,466 @@
...
+        if len(body) <= max_summary:
+            with open(summary_file, "a", encoding="utf-8") as sf:
+                sf.write(body)
+                sf.write("\n")
+        else:
+            print("Warning: step summary too large, skipping", file=sys.stderr)
```
PawelWMS: Can we print a truncated version instead? At least we'll get something.

.github/workflows/check-rendered-specs-stub.yml
dmcilvaney: What it means is forks themselves won't trigger for PRs into said fork, it will skip instead. If they want to run it, they will need to modify their default branch to include their fork in the list.

.github/workflows/scripts/post_render_comment.py
```diff
@@ -0,0 +1,466 @@
...
+    extra_files = report.get("extra_files", [])
+    missing_files = report.get("missing_files", [])
+
+    n_diff = len(content_diffs)
+    n_extra = len(extra_files)
+    n_missing = len(missing_files)
+    total = n_diff + n_extra + n_missing
+
+    if total == 0:
+        return f"{COMMENT_MARKER}\n## ✅ Rendered specs are up to date\n"
```
PawelWMS: This will never be hit considering the same check in main().

.github/workflows/scripts/post_render_comment.py
```diff
@@ -0,0 +1,466 @@
...
+
+    all_comps: list[str] = sorted(
+        set(_unique_components(content_diffs) + _unique_components(missing_files))
```
PawelWMS: Why bother with `_unique_components` if we're using a set?

.github/workflows/scripts/post_render_comment.py
```diff
@@ -0,0 +1,466 @@
...
+        lines.append(f"Or [download the fix patch]({artifacts_url}) and apply it:")
+        lines.append("")
+        if run_id and repo:
+            lines.append(
+                "```bash\n"
+                f"gh run download {run_id} -R {repo} -n rendered-specs-patch\n"
+                "git apply rendered-specs.patch\n"
+                "```"
+            )
+        else:
```
PawelWMS, line 162: Just checking: since we're using `"\n".join(lines)` below, which already adds new lines, are the explicit `\n` in some of the comments intentional?

.github/workflows/containers/azldev-runner.Dockerfile
dmcilvaney: Ok, looks like this should work

.github/workflows/scripts/post_render_comment.py
dmcilvaney: I think some wiggle room is nice here. This is mostly just a defense-in-depth thing to make sure the API calls work. Any message that large is going to be a mess anyways, a few extra characters won't make a difference.

.github/workflows/scripts/post_render_comment.py
dmcilvaney: I swear I got rid of that... must have brought it back during a rebase. Will remove again.

.github/workflows/scripts/post_render_comment.py
dmcilvaney: Markdown needs `\n\n` to actually render a real newline in the body. `" ".join(lines)` would be equivalent here, just makes more sense to use `\n` I think.

.github/workflows/check-rendered-specs.yml
dmcilvaney: I believe that without `always()` there is an implicit skip on failure, even with additional conditionals. ie equivalent to `if: success() && needs.render.result != 'cancelled'`

.github/workflows/scripts/post_render_comment.py
dmcilvaney: `json.JSONDecodeError` should catch that. Any other exception is completely unexpected and will bubble up and hit the global handler w/ call stack.

.github/workflows/scripts/post_render_comment.py
dmcilvaney: done

.github/workflows/scripts/post_render_comment.py
dmcilvaney: Also, looking again, this could be made using `lines.extend()`, I'll give that a try.

PawelWMS: Thanks. I was debating to make a comment about using `extend` or `append` consistently. I'm glad you did the update.

.github/workflows/check-rendered-specs.yml
PawelWMS: I'd assume an explicit `if` overrides the default but that's my experience from ADO pipelines, not GitHub. Having said that, I don't care about this line enough to check it right now or be nit-picky about this comment.;)

.github/copilot-instructions.md
```diff
@@ -52,13 +65,15 @@ Run all commands from the repo root (where `azldev.toml` lives). If the terminal
 | Add a component | `azldev comp add` |
 | Build a component | `azldev comp build -p <name> -q` |
 | Build chain (auto-publish to local repo) | `azldev comp build --local-repo-with-publish ./base/out -p <a> -p <b> -q` |
+| Render many specs for check-in | `azldev comp render -a` |
```
PawelWMS: Nit-pick, not sure if it's important for AI:

```suggestion
| Render all specs for check-in | `azldev comp render -a` |
```

.github/workflows/scripts/check_rendered_specs.py
```diff
@@ -0,0 +1,515 @@
...
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+# Matches the azldev-generated changelog line.
+# Tightly coupled to the format emitted by azldev's render pipeline; if azldev
+# ever changes the emitted form (email suffix, version tag, different
+# whitespace) this regex must be updated or every spec will look like drift.
+# Owner: azure-linux-dev-tools (cmd that emits "* <date> azldev <>" lines).
+# e.g. "* Wed Apr 08 2026 azldev <> - 1.0-1"
```
PawelWMS: Should this example have no e-mail inside `<>`?

.github/workflows/scripts/check_rendered_specs.py
```diff
@@ -0,0 +1,515 @@
...
+# ---------------------------------------------------------------------------
+
+
+def _git_bytes(*args: str) -> bytes:
+    """Run a git command and return stdout (bytes)."""
+    return subprocess.run(["git", *args], capture_output=True, check=True).stdout
+
+
+def _git_lines_z(*args: str) -> list[str]:
+    """Run a git command with -z-compatible args and return NUL-split lines."""
```
PawelWMS:
```suggestion
    """Run a git command with -z-compatible args and return NULL-split lines."""
```

.github/workflows/scripts/check_rendered_specs.py
```diff
@@ -0,0 +1,515 @@
...
+)
+
+# ---------------------------------------------------------------------------
+# Git helpers
+# ---------------------------------------------------------------------------
+
+
+def _git_bytes(*args: str) -> bytes:
+    """Run a git command and return stdout (bytes)."""
+    return subprocess.run(["git", *args], capture_output=True, check=True).stdout
```
PawelWMS, line 51: How do we capture errors from running the `git` command this way?

.github/workflows/scripts/check_rendered_specs.py
```diff
@@ -0,0 +1,515 @@
...
+    `git show HEAD:<path>` rev-parse syntax. The latter chokes on perfectly
+    legal filenames containing `:` (interpreted as the rev/path separator)
+    and leading `-` (parsed as options); `--` after `ls-tree` makes the path
+    list unambiguous, and `-z` keeps newline-bearing filenames intact.
+    """
+    if not paths:
+        return {}
+    raw = _git_bytes("ls-tree", "-z", "HEAD", "--", *paths)
+    out: dict[str, str] = {p: "" for p in paths}
+    for entry in raw.split(b"\x00"):
```
PawelWMS: Double-checking because I see this is vaguely similar to what `_git_lines_z` is doing: do we not want to call `decode("utf-8")` for this output and that's why we're not using `_git_lines_z`?

.github/workflows/scripts/check_rendered_specs.py
```diff
@@ -0,0 +1,515 @@
...
+        if not entry:
+            continue
+        # Format: "<mode> <type> <sha>\t<path>"
+        meta, _, path_b = entry.partition(b"\t")
+        try:
+            _mode, kind, sha = meta.split(b" ")
```
PawelWMS: Is `_mode` used?

```suggestion
            _, kind, sha = meta.split(b" ")
```

.github/workflows/scripts/check_rendered_specs.py
dmcilvaney: I guess we do generate those now, didn't when I started. Will update.

.github/workflows/scripts/check_rendered_specs.py
dmcilvaney: `check=True` requires success, otherwise it generates an exception which will propagate up.

.github/workflows/scripts/check_rendered_specs.py
dmcilvaney: was fine, but I've added explicit decodes since the other similar funcs use it.

PawelWMS (approved):

christopherco (approved):
:shipit:

PR #16679, opened 2026-04-15
feat(kernel): switch default LSM from AppArmor to SELinux

rlmenge:
Set CONFIG_DEFAULT_SECURITY_SELINUX=y and update CONFIG_LSM to use selinux instead of apparmor on both x86_64 and aarch64 configs.

```
[test@azurelinux-vm ~]$ sestatus
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   enforcing
Mode from config file:          enforcing
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     actual (secure)
Max kernel policy version:      35
```

TaskID: 1398047
Deliverable: https://dev.azure.com/mariner-org/mariner/_workitems/edit/18781


base/comps/kernel/6.18-x86_64-azl.config
```diff
@@ -7627,12 +7627,12 @@ CONFIG_EVM_ATTR_FSUUID=y
...
 # CONFIG_EVM_LOAD_X509 is not set
-# CONFIG_DEFAULT_SECURITY_SELINUX is not set
+CONFIG_DEFAULT_SECURITY_SELINUX=y
 # CONFIG_DEFAULT_SECURITY_SMACK is not set
 # CONFIG_DEFAULT_SECURITY_TOMOYO is not set
-CONFIG_DEFAULT_SECURITY_APPARMOR=y
+# CONFIG_DEFAULT_SECURITY_APPARMOR is not set
 # CONFIG_DEFAULT_SECURITY_DAC is not set
-CONFIG_LSM="landlock,lockdown,yama,integrity,apparmor"
+CONFIG_LSM="landlock,lockdown,yama,integrity,selinux"
```
reubeno, line 7635: For comparison's sake, I see my Fedora 43 workstation box has `CONFIG_LSM` set to:

`lockdown,yama,integrity,selinux,bpf,landlock,ipe`

Have we evaluated whether `bpf` and `ipe` should be in the list? (I'm guessing they were in the list for Alpha1?)

rlmenge: I am not convinced we should enable bpf but we will want ipe. I will investigate it as a part of the ipe integration steps https://dev.azure.com/mariner-org/mariner/_workitems/edit/18799

christopherco (approved):

PR #16682, opened 2026-04-15
feat: set azurelinux_release release name and prerelease name

ddstreet:
We incorrectly already set the release to greater than 0.X, so the built-in prerelease logic won't work this time; so let's add a manually specified prerelease name which we need to manually remove later before actual release.

Also fix a changelog day-of-week error, and remove TODO commented-out creation of a `<VENDOR>_release` (i.e. `microsoft-release`) symlink that we aren't using

### DIFF
--- a/base/comps/azurelinux-release/azurelinux-release.spec
+++ b/base/comps/azurelinux-release/azurelinux-release.spec
@@ -235,8 +235,6 @@ echo "cpe:/o:microsoft:azurelinux:%{version}" > %{buildroot}%{_prefix}/lib/syste
 install -d %{buildroot}%{_sysconfdir}
 ln -s ../usr/lib/azurelinux-release %{buildroot}%{_sysconfdir}/azurelinux-release
 ln -s ../usr/lib/system-release-cpe %{buildroot}%{_sysconfdir}/system-release-cpe
-# TODO(azl): validate
-# ln -s azurelinux-release %{buildroot}%{_sysconfdir}/azurelinux-release
 ln -s azurelinux-release %{buildroot}%{_sysconfdir}/system-release
 
 # Create the common os-release file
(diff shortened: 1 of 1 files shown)


base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -1,4 +1,7 @@
-%define release_name Alpha2
+%define release_name Four
```
reubeno, line 1: Looks like this will make the `VERSION` show up like:

> 4.0 (Four Alpha2)

Is that what we want? I agree we don't really have any other codename / moniker to place there, but I'm not sure that the "Four" makes sense.

base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -409,8 +412,6 @@ install -Dm0755 %{SOURCE19} %{buildroot}%{_libexecdir}/proc-version-override
 %{_prefix}/lib/system-release-cpe
 %{_sysconfdir}/os-release
 %{_sysconfdir}/azurelinux-release
-# TODO(azl): review
```
reubeno, line 412: Thanks for cleaning that up :smile:

base/comps/azurelinux-release/azurelinux-release.spec
ddstreet: I put in `Four` because that's how Fedora uses it:
https://src.fedoraproject.org/rpms/fedora-release/blob/f43/f/fedora-release.spec#_1

We don't have to go with the same thing they do, but I don't particularly see any reason to do it differently. We don't have named releases, like Debian/Ubuntu (e.g. "VERSION="25.10 (Questing Quokka)"), so going with the textual value of our release number, like Fedora does, seems reasonable.

I suppose we could use `Four Point Zero` or something like that if you want to include the tenths value

christopherco: Given that Fedora does naming this way, we can go with it for now given our lineage with Fedora as our upstream. And I'll book time to closely review the various `-release` files in `/etc` next week

christopherco (approved):

PR #16710, opened 2026-04-16
fix(kata-containers): add missing Requires on dbus-daemon

WithEnoughCoffee:
## Problem

The `kata-containers` RPM fails to install on minimal AZL4 systems because its `%post` scriptlet runs `dracut` to build a custom initrd. The dracut `rescue` module depends on `dbus`, which requires the `dbus-daemon` dracut module — but `dbus-daemon` is not installed.

```
dracut[E]: Module 'dbus' depends on module 'dbus-daemon', which can't be installed.
dracut[E]: Module 'dbus' cannot be installed.
[RPM] %post(kata-containers-3.26.0-1.azl4~20260413.x86_64) scriptlet failed, exit code 1
Transaction failed: Rpm transaction failed.
```

## Root Cause

The spec has `BuildRequires: dbus-daemon` (ensuring dracut works at build time) but is missing the corresponding runtime `Requires: dbus-daemon`.

## Fix

Add `Requires: dbus-daemon` via a `spec-insert-tag` overlay.

## Verification

Tested on an AZL4 QEMU VM (20GB, blob storage repo from 2026-04-13 daily build):
1. **Without fix**: `dnf install kata-containers` → transaction fails (dracut scriptlet error)

## Changes

- `base/comps/kata-containers/kata-containers.comp.toml` — new file with overlay
- `base/comps/components.toml` — remove bare `[components.kata-containers]` entry (now covered by the dedicated `.comp.toml`)

Tested locally using Qemu and azldev


ddstreetmicrosoft (changes requested):
This is definitely not correct.

What do you mean by 'minimal azl4 system'?

PR #16715, opened 2026-04-17
fix(pyOpenSSL): pin to 25.1.0 for cryptography 45.0.4 compat

christopherco:
Pin pyOpenSSL to 25.1.0 (Fedora dist-git commit 8ac54d5) because 25.2.0+ requires cryptography >= 45.0.7, but we ship 45.0.4 (python-cryptography is pinned to 45.0.4 since 46.x needs crate(asn1) >= 0.22.0 which is not yet available).

pyOpenSSL 25.1.0 requires cryptography >= 41.0.5, < 46, which is compatible with our 45.0.4.

Also add BuildRequires overlays to pin python3-cryptography to >= 45.0.4 and < 46 to ensure the build resolves to our version.


base/comps/pyOpenSSL/pyOpenSSL.comp.toml
```diff
@@ -0,0 +1,19 @@
...
+description = "Pin python3-cryptography to 45.x to match the version we ship (python-cryptography is pinned to 45.0.4)"
+type = "spec-add-tag"
+tag = "BuildRequires"
+value = "python3-cryptography >= 45.0.4"
+
+[[components.pyOpenSSL.overlays]]
+description = "Prevent building against cryptography 46.x which we do not ship"
+type = "spec-add-tag"
+tag = "BuildRequires"
+value = "python3-cryptography < 46"
```
reubeno, line 19: Why do we need to prevent building against something that we don't have? What will this catch?

christopherco: In our stage1, we now have both the pinned `python3-cryptography-45.0.4-1` and the unpinned previous version in stage1 `python3-cryptography-46.0.5-1`. I elected not to purge the previous unpinned version since they have unique NVRs. And instead I'm explicitly bounding against the `python3-cryptography` version bump that took the new `crate(asn1)` dependency.

reubeno: Ah, okay. I'm okay with that for expedience, but we need to track this as debt that needs to be addressed once we're past stage1 shenanigans.

reubeno (approved):

base/comps/pyOpenSSL/pyOpenSSL.comp.toml
christopherco: https://dev.azure.com/mariner-org/mariner/_workitems/edit/18826

PR #16735, opened 2026-04-17
fix(sos): Fix os_release_name value

ddstreet:
The sos distro detection currently matches the /etc/os-release NAME field wrong; this backports the upstream patch to correct the os-release NAME matching to detect that it's running on Azure Linux.


reubeno (approved):
I'm okay taking it -- but I'm surprised the patch didn't switch to match on `ID`, which is the *correct* item to match on, no?

christopherco (approved):
Agree - change is fine, and maybe an improvement to suggest upstream. Logged a bug for it

eric-desrochers:
An upstream change can be done to our azure.py policy as follows:
```
- os_release_name = 'Azure Linux'
+ os_release_id = 'azurelinux'
```

sosreport supports them both, see this Fedora example: 
https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L377-L378

Ultimately, it's up to us to submit a PR and switch the detection field logic to what we want.

eric-desrochers:
Submitted this PR upstream: https://github.com/sosreport/sos/pull/4306/changes

PR #16736, opened 2026-04-18
workaround(nbd): mark tls tests as expected failures due to expired certs

christopherco:
The nbd 3.25 test suite ships pre-built TLS certificates that were generated on 2016-04-19 with a 10-year expiration, meaning they expired on 2026-04-17. This causes the tls and tlshuge tests to fail with 'Error in the certificate' from GnuTLS.

Upstream fixed this in nbd 3.27.0 (commit 58b356b, 'Update certificate expiry dates') with certs valid until 2033, but 3.25 (used on f43) was not updated.

Add an overlay to pass XFAIL_TESTS='tls tlshuge' as a make variable override so these tests are treated as expected failures rather than blocking the build.

Fixes: [AB#18940](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/18940)

Tracking item for proper upstream fix: [AB#18854](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/18854)


christopherco:
Work item for more durable fix -> https://dev.azure.com/mariner-org/mariner/_workitems/edit/18854

base/comps/nbd/nbd.comp.toml
```diff
@@ -0,0 +1,8 @@
+[components.nbd]
+
+[[components.nbd.overlays]]
+description = "Mark tls and tlshuge tests as expected failures — shipped test certificates expired on 2026-04-17 (10-year certs generated 2016-04-19). Fixed upstream in nbd 3.27.0 (commit 58b356b) but not backported to 3.25."
+type = "spec-search-replace"
+section = "%check"
+regex = 'DELAY=10 make check'
+replacement = "# tls and tlshuge tests use pre-built certificates that expired 2026-04-17\nDELAY=10 make check XFAIL_TESTS=\"tls tlshuge\""
```
reubeno: This could be more readable with a multi-line literal string for readability, e.g.:

```toml
replacement = '''
# ...comment here...
make check XFAIL_TESTS=...
'''
```

reubeno: How important do you think it is to leave the comment in the rendered spec? (This is a more general question.)

If we consider this important with the shift toward in-repo rendered specs, then we should find a way to avoid making developers need to violate DRY principles. (The explanation for the overlay is included in the description already.)

We should file a follow-up item to find a better way to do this if we deem it important.

christopherco: TIL! Will add this to my notes for future changes

christopherco: IMO not important to keep, but we should certainly discuss this as a follow-up item. For this PR though, I've gone ahead and removed the comment since we aren't yet at full rendered specs.

anphel31 (approved):

PR #16738, opened 2026-04-18
fix: drop openjdk21 Java tooling variants (repoclosure)

christopherco:
## Summary

AZL ships only `java-25-openjdk`. The f43-based Java tooling specs still
produce `-openjdk21` variant subpackages that depend on the missing
`java-21-openjdk-*`, causing 7 repoclosure failures.

This PR surgically removes the openjdk21 bindings/subpackages via overlays,
backporting the equivalent changes already landed in Fedora rawhide/f44.

## Changes

| Package | Overlay approach | Upstream commit |
|---------|-----------------|-----------------|
| ant | Remove `%jp_binding --variant openjdk21` | `4ca7a3b` |
| maven | Remove `%jp_binding --variant openjdk21` | `e19767f` |
| maven4 | Remove `%jp_binding --variant openjdk21` | `4b6c6ee` |
| xmvn | Remove openjdk21 toolchain install + binding | `2c9d321` |
| xmvn5 | Remove openjdk21 toolchain install + binding | `b9afca4` |
| javapackages-tools | `spec-remove-section` for `%package`/`%description`/`%files` blocks; fix `--jvm=openjdk21` configure flag; fix ivy-local and javapackages-compat Requires | `62dfe8f`, `7debb9d`, `83cd9c8` |

## Repoclosure issues resolved

- `ant-openjdk21` → `java-21-openjdk-headless`
- `javapackages-local-openjdk21` → `java-21-openjdk-devel`
- `maven-local-openjdk21` → `java-21-openjdk-devel`
- `maven-openjdk21` → `java-21-openjdk-headless`
- `maven4-openjdk21` → `java-21-openjdk-headless`
- `xmvn-toolchain-openjdk21` → `java-21-openjdk-devel`
- `xmvn5-toolchain-openjdk21` → `java-21-openjdk-devel`

Fixes: [AB#18867](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/18867)


reubeno (commented):
Main feedback here is that the regexes are much more tightly coupled to the actual syntax in the specs. In several cases, simply replacing `.*openjdk21.*` with `""` may be sufficient? Perhaps that's too simplistic, but the precise whitespace and verbatim other tokens decreases likelihood that the overlay would continue to apply as generically.

**Consideration for future**: if there are already changes in a future version of Fedora -- do we actually want to have a mechanism to allow storing those dist-git commits as local .patch files and an overlay that can apply them?

christopherco:
> Main feedback here is that the regexes are much more tightly coupled to the actual syntax in the specs. In several cases, simply replacing `.*openjdk21.*` with `""` may be sufficient? Perhaps that's too simplistic, but the precise whitespace and verbatim other tokens decreases likelihood that the overlay would continue to apply as generically.

Went with the simpler approach here. See new PR #16767 . Closing this one

> **Consideration for future**: if there are already changes in a future version of Fedora -- do we actually want to have a mechanism to allow storing those dist-git commits as local .patch files and an overlay that can apply them?

Likely yes. I'll open a follow-up item on this.

PR #16741, opened 2026-04-19
workaround(gnutls): disable fips subpackage build

christopherco:
The gnutls-fips subpackage cannot correctly locate its specifically built nettle and gmp libraries due to how packages are currently built in the AZL build environment. Since gnutls-fips is not yet required, disable the fips bcond as a workaround.

This moves gnutls from an inline definition in components.toml to a dedicated comp.toml with build.without = ["fips"], which passes --without fips to rpmbuild and cleanly skips the fips subpackage.

Fixes: [AB#18869](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/18869)

Tracking item for removing this workaround: [AB#18931](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/18931)


christopherco:
Tracking item for removing this workaround: AB#18931

reubeno (approved):

PR #16763, opened 2026-04-20
Removes fedora requires from anaconda-webui

WithEnoughCoffee:
The upstream Fedora `anaconda-webui` spec carries two Fedora-specific
conditional
 dependencies that break repoclosure on AZL:

 1. `Requires: fedora-logos` (gated behind `%if 0%{?fedora}`)
 2. `Requires: (firefox if fedora-release-workstation)` (gated behind `%if
 ! 0%{?rhel}`)

 Neither package exists in AZL. PR #16413 already removed `fedora-logos`
from the
 component list so it is no longer built, but `anaconda-webui` still
requires it —
 this patch fixes the consumer side.

 AZL ships `generic-logos` which satisfies the `system-logos` virtual
provide that
 packages are meant to depend on.

 ## Changes

 - **New** `base/comps/anaconda-webui/anaconda-webui.comp.toml` — two
   `spec-search-replace` overlays that replace the `Requires` lines with
comments,
   leaving the surrounding `%if/%endif` blocks as harmless no-ops.
 - **Removed** bare `[components.anaconda-webui]` entry from
`components.toml`
   (auto-discovered via the `includes` glob).

 ## Validation

 - `azldev comp prep-sources -p anaconda-webui` succeeds; rendered spec 
shows:

%if ! 0%{?rhel}

Removed: firefox conditional — fedora-release-workstation not in AZL

%endif %if 0%{?fedora}

Removed: fedora-logos — AZL uses generic-logos

%endif

 ## References

 - Repoclosure bug #18865 (Bucket 1 — missing components)
 - PR #16413 — removed `fedora-logos` from component list (merged
2026-04-02)


base/comps/anaconda-webui/anaconda-webui.comp.toml
```diff
@@ -0,0 +1,21 @@
...
+type = "spec-search-replace"
+regex = '^Requires: fedora-logos$'
+replacement = "# Removed: fedora-logos — AZL uses generic-logos"
+
+# The upstream spec also pulls in Firefox via a boolean dep on
+# fedora-release-workstation, which does not exist in AZL.
+[[components.anaconda-webui.overlays]]
+description = "Remove conditional Firefox dep on fedora-release-workstation (not in AZL)"
+type = "spec-search-replace"
+regex = '^Requires: \(firefox if fedora-release-workstation\)$'
```
reubeno, line 20: Does this actually need to be removed? My understanding of this syntax is that the requirement is only engaged if the package after the `if` is present -- and since it's not, it won't activate.

base/comps/anaconda-webui/anaconda-webui.comp.toml
```diff
@@ -0,0 +1,21 @@
...
+# The upstream Fedora spec requires fedora-logos (gated behind %if 0%{?fedora}).
+# AZL uses generic-logos for branding; fedora-logos is not shipped and breaks
+# repoclosure. Replace with a comment — the surrounding %if/%endif is harmless.
+# Ref: PR #16413 removed fedora-logos from the component list; this patches
+# the consumer side.
+
+[[components.anaconda-webui.overlays]]
+description = "Remove Requires: fedora-logos — AZL uses generic-logos for branding"
+type = "spec-search-replace"
+regex = '^Requires: fedora-logos$'
```
reubeno, line 12: Should we be replacing `fedora-logos` with `system-logos` instead? That's the more generic virtual provider that is implemented by both `generic-logos` and `fedora-logos`.

I'd probably recommend a `spec-remove-tag` and a `spec-add-tag` to do that.

base/comps/anaconda-webui/anaconda-webui.comp.toml
```diff
@@ -0,0 +1,21 @@
...
+replacement = "# Removed: fedora-logos — AZL uses generic-logos"
+
+# The upstream spec also pulls in Firefox via a boolean dep on
+# fedora-release-workstation, which does not exist in AZL.
+[[components.anaconda-webui.overlays]]
+description = "Remove conditional Firefox dep on fedora-release-workstation (not in AZL)"
+type = "spec-search-replace"
+regex = '^Requires: \(firefox if fedora-release-workstation\)$'
+replacement = "# Removed: firefox conditional — fedora-release-workstation not in AZL"
```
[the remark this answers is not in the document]
WithEnoughCoffee: @copilot apply changes based on [this feedback](https://github.com/microsoft/azurelinux/pull/16763#discussion_r3114232088)

base/comps/anaconda-webui/anaconda-webui.comp.toml
WithEnoughCoffee: I did some more research and it looks like it doesn't. Thank you for the comment I'll make the change.

base/comps/anaconda-webui/anaconda-webui.comp.toml
WithEnoughCoffee: That's a better idea I'll do that thank you.

base/comps/anaconda-webui/anaconda-webui.comp.toml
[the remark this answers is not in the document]
WithEnoughCoffee: Fixed

binujp (approved):
Looks good, thanks Autumn.

PR #16764, opened 2026-04-21
fix: Fix %distcore definition

ddstreet:
The %distcore definition placed into the macros.dist file currently contains a macro, %dist_version_major, which is undefined outside of the azurelinux-release spec; this changes it to hardcode the value.


christopherco (approved):
:shipit:

PR #16704, opened 2026-04-15
feat(packages): add package publish configuration

liunan-ms:
This PR adds `base/packages.toml` with package group publish channel assignments (generated from azl4 20260413 repo closure):
- **base-packages**: packages published to `rpm-base` channel
- **sdk-packages**: packages published to `rpm-sdk` channel
- **base-debuginfo-packages**: packages published to `rpm-base-debuginfo` channel
- **sdk-debuginfo-packages**: packages published to `rpm-sdk-debuginfo` channel

All other packages default to channel = `none` (unpublished).

Wire `base/packages.toml` into the project by adding it to the includes list in `base/project.toml`.

The package publish configuration will be resolved by `azldev package list <package-name>` and `azldev comp build <component-name>` and be viewed in the output.

Example:
image

PR #16765, opened 2026-04-21
feat(packages): add short-term package lists

reubeno:
Extends #16704.

This is *not* the form of the package lists that we want to maintain -- it's significantly more verbose and package-focused (vs. component-focused). But for now (i.e., the next 1-2 days), this unblocks getting out alpha2 package publishing routing working and we'll iterate and focus on more maintainable ways to express the desired config.


binujp (approved):

christopherco (approved):
:shipit:

PR #16773, opened 2026-04-21
fix(kf6-breeze-icons): drop breeze-icon-theme-fedora subpackage

WithEnoughCoffee:
## Summary

Drop the `breeze-icon-theme-fedora` subpackage — it requires `fedora-release-common`, which does not exist in Azure Linux.

## Problem

The upstream spec gates `breeze-icon-theme-fedora` behind `%if 0%{?fedora}`, but `%fedora` is set on our builders, so the guard is ineffective. The subpackage's hard `Requires: fedora-release-common` causes `dnf repoclosure` to fail.

## Fix

Add a `spec-search-replace` overlay that replaces `%if 0%{?fedora}` → `%if 0`, disabling all 4 Fedora-specific blocks in the spec. The base `breeze-icon-theme` package is unaffected.

The subpackage only contains the Fedora Anaconda installer icon (`org.fedoraproject.AnacondaInstaller.svg`) — pure Fedora branding with no AZL equivalent needed.

## Testing

Verified all 4 occurrences of `%if 0%{?fedora}` in the spec are replaced with `%if 0`, disabling:
- Subpackage definition (line 69)
- Icon install (line 139)
- File exclusion (line 181)
- Files list (line 185)

Unable to run `azldev prep-sources` due to pre-existing edk2/javapackages-tools config incompatibility with installed azldev version; overlay validated manually via sed simulation.

## References

Resolves repoclosure failure for `breeze-icon-theme-fedora` (Bucket 1, missing-component item from repoclosure-analysis-20260417).
Ref: #18865


binujp:
suggestion: can we use the spec-remove-section overlay here please? Smaller overlay and easier to understand.

WithEnoughCoffee:
> suggestion: can we use the spec-remove-section overlay here please? Smaller overlay and easier to understand.

Thank you for the suggestion. I tried the spec-remove-section originally but unfortunately spec-remove-section won't work here — each fedora-specific block in the upstream spec is wrapped in %if 0%{?fedora} /
%endif conditionals:

 %if 0%{?fedora}
 %package -n breeze-icon-theme-fedora
 ...
 %description -n breeze-icon-theme-fedora
 ...
 %endif
When using  spec-remove-section it removes the %package/%description/%files content but also eat the %endif, leaving an unclosed %if that breaks the spec. Same issue I ran into on the plymouth-theme-charge. 

For this overlay I tried to intentionally create a single unscoped spec-search-replace because the %if lines sit outside section boundaries (they wrap the sections), so section/package scoping doesn't match
them either. I've documented both constraints in the TOML comments.

binujp (approved):
Looks good. Let us merge this. I would like you to follow up with whether we need this component. I have a feeling this maybe in the same category as plymouth and we can get rid of this.

PR #16777, opened 2026-04-21
fix(plymouth): drop plymouth-theme-charge subpackage

WithEnoughCoffee:
Fixes: [AB#18913](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/18913)

The plymouth-theme-charge subpackage requires fedora-logos-classic, which does not exist in Azure Linux. The theme is Fedora-specific branding ("shadowy hull of a Fedora logo") with no AZL equivalent needed.

The upstream spec gates it behind %if ! 0%{?rhel}, which excludes it on RHEL but not on AZL (since %rhel is undefined on our builders). Add a spec-search-replace overlay to force the guard off, eliminating the unresolvable dependency. All other plymouth subpackages are unaffected.

Tested:
- azldev comp prep-sources: overlay applies cleanly, all 4 occurrences of '%if ! 0%{?rhel}' replaced with '%if 0' (lines 260, 360, 375, 501)
- azldev comp build: build succeeds, produces all expected plymouth RPMs — no plymouth-theme-charge RPM is generated

Resolves repoclosure failure for plymouth-theme-charge (Bucket 1, missing-component item from repoclosure-analysis-20260417). Ref: microsoft/azurelinux#18865

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx


binujp:
suggestion: Since we are dropping the whole sub-package, how about using `spec-remove-section` instead of search-replace? That will make the overlay look crisper.

WithEnoughCoffee:
> suggestion: Since we are dropping the whole sub-package, how about using `spec-remove-section` instead of search-replace? That will make the overlay look crisper.

I used spec-remove-section in all the places that it could be cleanly removed.

ddstreetmicrosoft (dismissed):
you have a second commit fixing the first commit; please rebase into a single commit.

base/comps/plymouth/plymouth.comp.toml
```diff
@@ -0,0 +1,20 @@
...
+#
+# Note: spec-remove-section cannot be used here because each charge-theme
+# section is wrapped in %if/%endif guards. Removing the section also eats
+# the %endif, leaving an unclosed %if that breaks rpmbuild.
+# The same guard appears in 4 spec sections (%package, %install, %postun,
+# %files) and all must be disabled.
+[[components.plymouth.overlays]]
+description = "Disable Fedora-specific charge theme subpackage (plymouth-theme-charge) — requires fedora-logos-classic which AZL does not ship"
+type = "spec-search-replace"
+regex = '^%if\s+!\s+0%\{\?rhel\}\s*$'
```
christopherco, line 19: issue: Regex is inherently fragile from a future-proofing perspective. Looking at the upstream plymouth spec, it seems possible to set the `rhel` macro to `9` and achieve the same result. Consider if this might be a better option.

christopherco (changes requested):
In addition to the rebase feedback and my comment to evaluate if there's a better method than regex for this, the PR description also has incorrect line numbers (seemingly consistently off by 3 lines compared to the F43 spec). Please double-check those closely if you want to keep the line numbers, or instead simply focus on the actual overlay logic as the line numbers might be unnecessary information.

WithEnoughCoffee:
> suggestion: Since we are dropping the whole sub-package, how about using `spec-remove-section` instead of search-replace? That will make the overlay look crisper.

fixed thanks!

WithEnoughCoffee:
> you have a second commit fixing the first commit; please rebase into a single commit.

Will do thank you

base/comps/plymouth/plymouth.comp.toml
WithEnoughCoffee: I did some research and Im apprehensive about using the` rhel ` macro and setting it to 9 because its mean to be a platform identifier for redhat and not just a toggle feature. Setting this macro to 9 makes the build think its enterprise 9 and if a silent change came from upstream %if...rhel logic or macro behavior beyond just the charge theme. That could also mean for an unbounded blast radius. On the other hand while the regex can be more fragile in this instance it only targets the 4 %if ! 0%{?rhel} guards. If upstream changes the formatting, it fails loudly (build breaks) rather than silently doing the wrong thing. I've made the regex whitespace-tolerant to handle minor formatting changes. 

 Also fixed the line numbers — removed them from the description entirely since they were fragile and unnecessary.

christopherco (commented):
Changes generally look good to me. Please rebase on the latest `tomls/base/main` branch and do the spec rendering & lock file update with `azldev` (there should be AI skills that can assist in the latest `tomls/base/main` branch).

WithEnoughCoffee:
> Changes generally look good to me. Please rebase on the latest `tomls/base/main` branch and do the spec rendering & lock file update with `azldev` (there should be AI skills that can assist in the latest `tomls/base/main` branch).

Rebased and the specs been pre-rendered and lock file updated.

christopherco (approved):

PR #16844, opened 2026-04-23
refactor(openssl-fips-provider): Better align with openssl's fips provider

corvus-callidus:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Simplify openssl-fips-provider by removing all Fedora-derived patches and using upstream `openssl fipsinstall` to generate the FIPS module config with its integrity HMAC, replacing the embedded HMAC approach. Align both openssl (3.3.7) and openssl-fips-provider (3.1.2) to use the upstream fipsmodule.cnf / [fips_sect] naming convention instead of the custom fips_prov.cnf / [fips_prov_sect].

###### Change Log  

- Remove 20+ Fedora-derived patches from openssl-fips-provider, building the unmodified OpenSSL 3.1.2 FIPS provider
- Replace embedded HMAC approach (patch 0033) with upstream openssl fipsinstall for FIPS config generation
- Update openssl patch 0032 to load fipsmodule.cnf with [fips_sect] instead of fips_prov.cnf with [fips_prov_sect]
- Bump openssl to 3.3.5-6; bump openssl-fips-provider to 3.1.2-2
- Add openssl-fips-provider Requires on openssl >= 3.3.5-6

**YES**

###### Associated issues  

- #xxxx

###### Links to CVEs  
- none

###### Test Methodology

- Pipeline build id: [1101642](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1101642)
- LISA tests


tobiasb-ms (approved):

corvus-callidus:
Failed test is for libguestfs. This is known problematic test and unrelated to openssl.

SPECS/openssl/0032-Force-fips-symcrypt-or-fips-3.3.5-AZL3.patch
christopherco, line 47: **issue (blocking):** Missing `NULL` check on `NCONF_new_ex` and leaked `CONF*` on the failure path.

```c
CONF *prov_conf = NCONF_new_ex(libctx, NCONF_default());   // not NULL-checked
if (NCONF_load(prov_conf, provider_config_path, NULL) <= 0) {
    return 0;                                               // prov_conf leaked
}
```

Two bugs here:

- **Path A (OOM):** `NCONF_new_ex` can return `NULL`. The next line passes that straight into `NCONF_load`, which is a NULL-deref/abort. Triggers during FIPS init under memory pressure.
- **Path B (any load failure):** A malformed `fipsmodule.cnf`, an EIO, a partial write during install, or a TOCTOU between the `access()` check and `NCONF_load` all hit the early `return 0` — but `prov_conf` was already heap-allocated and is never freed. The two success branches further down do `NCONF_free` correctly; this path was missed.

**Suggested fix:**

```c
CONF *prov_conf = NCONF_new_ex(libctx, NCONF_default());
if (prov_conf == NULL) {
    return 0;
}
if (NCONF_load(prov_conf, provider_config_path, NULL) <= 0) {
    NCONF_free(prov_conf);
    return 0;
}
```

SPECS-EXTENDED/openssl-fips-provider/openssl-fips-provider.spec
christopherco: **issue (blocking):** `NEW_RPM_OPT_FLAGS` is exported but never passed to `./Configure` — `fips.so` builds without distro hardening flags.

The variable is defined and exported with the same comment block as the main `openssl.spec`, but the `./Configure` invocation here only receives `shared enable-fips enable-buildtest-c++ -Wl,--allow-multiple-definition`. The retained comment *"The Configure script already knows to use -fPIC and `NEW_RPM_OPT_FLAGS`"* is **factually wrong for this spec** — it's true for `SPECS/openssl/openssl.spec:325` because *that* spec does `$NEW_RPM_OPT_FLAGS \` on its Configure line. This one doesn't.

Concrete losses on `fips.so`:

| Flag | Impact when missing |
|------|---------------------|
| `-Wa,--noexecstack` | OpenSSL ships hundreds of hand-written `.S` files with no `.note.GNU-stack` markers. Without this assembler flag, the linker marks `fips.so`'s `PT_GNU_STACK` as **RWE** — direct W^X bypass on load. |
| `%{optflags}` (`-fstack-protector-strong`, `_FORTIFY_SOURCE=2`, `-O2`) | SSP and fortified libc absent from the FIPS module. |
| `$RPM_LD_FLAGS` (`-Wl,-z,relro -Wl,-z,now`) | Full RELRO / BIND_NOW lost. |
| `-Wa,--generate-missing-build-notes=yes` | Annobin/build-id audit metadata missing. |
| `-DPURIFY` | Without it, OpenSSL contributes uninitialized memory to entropy, which masks valgrind-detectable issues. |
| `-DDEVRANDOM="\"/dev/urandom\""` | Defensive single-device override gone; upstream default extends fallback to `/dev/random`, which can block on minimal/container systems. |

The "FIPS cert binary identity" defense doesn't apply here because the integrity HMAC is regenerated by `openssl fipsinstall` at install time — this is *not* a CMVP-validated binary that needs byte-for-byte reproducibility.

**Suggested fix:** at minimum restore `-Wa,--noexecstack` (non-negotiable for any `.so` with assembler), and either restore the rest of `$NEW_RPM_OPT_FLAGS` + `-DDEVRANDOM` or add an explicit comment explaining why each flag is intentionally omitted. The current state — variable defined, exported, then ignored, with a misleading comment — is a silent regression that no reviewer should be expected to infer.

christopherco (changes requested):
Left a couple comments to look into, and also `openssl-fips-provider` needs a signatures.json refresh.

SPECS/openssl/0032-Force-fips-symcrypt-or-fips-3.3.5-AZL3.patch
corvus-callidus: Fixed.

SPECS-EXTENDED/openssl-fips-provider/openssl-fips-provider.spec
corvus-callidus: I've updated the flags used plus included a comment detailing justification for inclusion/exclusion.

tobiasb-ms (approved):

corvus-callidus:
Current buddy build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1130613

christopherco (approved):

SPECS/openssl/openssl.spec
christopherco, line 122: Does the libs subpackage need the `Conflicts: openssl-fips-provider < 3.1.2-2` instead of the global openssl package? I'm worried about minimal installations that only include `openssl-libs` but doesn't include `openssl`

SPECS-EXTENDED/openssl-fips-provider/openssl-fips-provider.spec
```diff
@@ -92,6 +36,7 @@ BuildRequires: perl(Test::Harness)
 BuildRequires: perl(Test::More)
 BuildRequires: perl(Time::Piece)
 
+Requires: openssl >= 3.3.5-6
```
christopherco: Similar to the openssl-libs conflicts question, should `openssl-libs%{?_isa} >= 3.3.5-6` be here instead (or in addition to `openssl >= 3.3.5-6`)?

christopherco (changes requested):
One more pair of questions on a conflicts / requires pairing

SPECS/openssl/openssl.spec
corvus-callidus: Good catch. I'd forgotten about minimal installs.

SPECS-EXTENDED/openssl-fips-provider/openssl-fips-provider.spec
corvus-callidus: Fixed.

christopherco (approved):

PR #16875, opened 2026-04-24
Review and update systemd default preset files

ddstreet:
Update the systemd preset files `90-default.preset` and `90-default-user.preset` provided by the `azurelinux-release` component

### DIFF
--- a/base/comps/azurelinux-release/90-default-user.preset
+++ b/base/comps/azurelinux-release/90-default-user.preset
@@ -2,6 +2,3 @@
 # unconditionally. It is used throughout the distro and required on all machines.
 enable dbus.socket
 enable dbus-broker.service
-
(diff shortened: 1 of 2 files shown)


ddstreet:
Note that the removal of the (commented-out) preset for `grub-boot-success.timer` is intentional, as the binary it calls (from the corresponding service file) is `grub2-set-bootflag` which is setuid root (and so is a security concern), and is likely to not be needed in the cloud context, since it relies on a non-system user actually logging in, which isn't part of most cloud vm/container workstreams.

The consideration of removal for the `grub2-set-bootflag` setuid binary is being tracked in AB#19231

reubeno (approved):

christopherco (approved):

PR #16902, opened 2026-04-27
feat(images): add WSL rootfs image

mfrw:
Adds a kiwi-ng image definition that produces a tar.bz2 rootfs
consumable by 'wsl --import'. The image ships systemd, dnf5, a
small interactive userland, hyperv-daemons, and the wsl-setup
package that provides /etc/wsl.conf, /etc/wsl-distribution.conf,
and the OOBE script. No kernel, grub2, dracut, or shim are
included as WSL provides its own kernel.

Registered as [images.wsl] in base/images/images.toml.

### DIFF
--- a/base/images/wsl/wsl.kiwi
+++ b/base/images/wsl/wsl.kiwi
@@ -1,0 +1,89 @@
+<?xml version="1.0" encoding="utf-8"?>
+<?xml-model href="https://raw.githubusercontent.com/OSInside/kiwi/refs/tags/v10.2.33/kiwi/schema/kiwi.rng" type="application/xml"?>
+<image schemaversion="8.3" name="azl4-wsl">
+    <description type="system">
+        <author>Azure Linux</author>
+        <contact>azurelinux@microsoft.com</contact>
+        <specification>Windows Subsystem for Linux (WSL)</specification>
+    </description>
+    <preferences>
+        <version>0.1</version>
+        <packagemanager>dnf5</packagemanager>
+        <locale>en_US</locale>
+        <keytable>us</keytable>
+        <timezone>UTC</timezone>
+        <release-version>4.0</release-version>
+        <!--
+            tbz produces a tar.bz2 rootfs consumable by `wsl --import`.
+            WSL provides its own kernel and bootloader, so this image
+            intentionally has no kernel, dracut, grub2, or shim.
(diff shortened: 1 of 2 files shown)


reubeno:
@mfrw For comparison's sake, how does this differ from Fedora's composition of [their WSL image](https://pagure.io/fedora-kiwi-descriptions/blob/main/f/teams/cloud/wsl.xml)?

ddstreet:
> @mfrw For comparison's sake, how does this differ from Fedora's composition of [their WSL image](https://pagure.io/fedora-kiwi-descriptions/blob/main/f/teams/cloud/wsl.xml)?

also possibly useful is the [wsl (and other) xml](https://github.com/ddstreet/azurelinux/blob/containers-kiwi-xml/base/images/container/wsl.xml) from my branch converting fedora's kiwi xml into our own; although note, none of that will be merged.

ddstreet:
question(non-blocking): I note that this creates an entirely new dir `base/images/wsl/wsl.kiwi`, would it be better to put this inside the existing `container-base.kiwi` and add a new profile?

mfrw:
> question(non-blocking): I note that this creates an entirely new dir `base/images/wsl/wsl.kiwi`, would it be better to put this inside the existing `container-base.kiwi` and add a new profile?

Yes this sounds like a good idea. Thank you for the suggestion.

> For comparison's sake, how does this differ from Fedora's composition


I tried to generate a draft minimal rootfs as a first draft.

base/images/wsl/wsl.kiwi
```diff
@@ -0,0 +1,89 @@
...
+    </preferences>
+
+    <repository type="rpm-md" alias="azl4">
+        <source
+            path="https://stcontroltowerdevjwisitg.blob.core.windows.net/daily-repo-non-prod/20260303/x86_64/" />
+    </repository>
+
+    <packages type="image">
+        <!-- Core -->
+        <package name="azurelinux-release-cloud" />
```
reubeno, line 34: There's an `azurelinux-release-wsl` package, I believe. We should use that.

mfrw: sure sir

christopherco:
Superseded by #17000 . Closing this PR

PR #16910, opened 2026-04-27
chore: move spec render script into subdirectory

PawelWMS:
Clean-up of scripts' locations. Placing the scripts related to the specs rendering check into their own subdirectory under `scripts`.

Also moved a readme related to the specs review scripts into the directory where the scripts live.

End result: each group of scripts covering one area of checks/functionality are now under `.../scripts/<area>` instead of keeping all of them inside `.../scripts`.


dmcilvaney (approved):

christopherco (approved):
:shipit:

PR #16914, opened 2026-04-28
versionsprocessor: sanitize dots and plus signs in RPM macro names

kgodara912:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

RPM macro names may only contain alphanumeric characters and underscores. The versionsprocessor was only replacing dashes with underscores, leaving dots and plus signs from spec filenames (e.g., rubygem-cool.io, libxml++, rust-1.75, golang-1.25) in the generated macro names. This caused repeated 'Macro %azl_<name> needs whitespace before body' warnings during every rpmbuild invocation.

Replace '.' and '+' with '_' alongside the existing '-' replacement when constructing macro names from spec filenames.

Add table-driven test cases covering dot, plus, and versioned spec names.

###### Change Log  

- Replaced the invalid characters in macro names with underscores

**YES**

###### Associated issues  

- Pipeline build id: xxxx

### DIFF
--- a/toolkit/tools/versionsprocessor/versionsprocessor.go
+++ b/toolkit/tools/versionsprocessor/versionsprocessor.go
@@ -189,9 +189,12 @@ func processPackageVersionString(packageVersionString string, specFileName strin
 	release := releaseVerSplit[2]
 	releaseClean := strings.Replace(release, distTag, "", 1)
 
-	// strip out the .spec suffix and replace '-' with '_' as RPM macros cannot have '-'
+	// strip out the .spec suffix and replace characters invalid in RPM macro names with '_'
+	// RPM macro names may only contain alphanumeric characters and underscores.
 	packageFileNameMacroFormat := strings.Replace(specFileName, ".spec", "", 1)
 	packageFileNameMacroFormat = strings.ReplaceAll(packageFileNameMacroFormat, "-", "_")
+	packageFileNameMacroFormat = strings.ReplaceAll(packageFileNameMacroFormat, ".", "_")
+	packageFileNameMacroFormat = strings.ReplaceAll(packageFileNameMacroFormat, "+", "_")
 
 	epochReleaseString := prefix + "_" + packageFileNameMacroFormat + "_epoch"
 	versionMacroString := prefix + "_" + packageFileNameMacroFormat + "_version"
--- a/toolkit/tools/versionsprocessor/versionsprocessor.go
+++ b/toolkit/tools/versionsprocessor/versionsprocessor.go
@@ -29,6 +29,7 @@ import (
 )
 
 var packageVersionRegexp = regexp.MustCompile(`^[^:]+: (?:(.+):)?(.+)-(.+)$`)
+var invalidMacroCharRegexp = regexp.MustCompile(`[^a-zA-Z0-9_]`)
 
 var (
 	app           = kingpin.New("versionsprocessor", "A tool to generate a macro file of all specs version and release")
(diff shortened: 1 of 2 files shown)


kgodara912:
[full build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1105033&view=results)

kgodara912:
Full build is successful. The PR check failure is because of known failure of `libguestfs`. Could you please take a look, @mfrw, @PawelWMS ?

mfrw:
LGTM ✅ — small, correct fix with good test coverage.

**Verified the logic against each new test case:**

| Spec file | Sanitized | Macro |
|---|---|---|
| `rubygem-cool.io.spec` | `rubygem_cool_io` | `%azl_rubygem_cool_io_version` ✓ |
| `libxml++.spec` | `libxml__` | `%azl_libxml___version` ✓ |
| `rust-1.75.spec` | `rust_1_75` | `%azl_rust_1_75_version` ✓ |
| `golang-1.25.spec` | `golang_1_25` | `%azl_golang_1_25_version` ✓ |

The triple-underscore in `%azl_libxml___version` looks unusual but is valid RPM macro syntax.

**Coverage check:** I scanned `SPECS*/` for all spec basenames containing characters outside `[A-Za-z0-9_-]`. The full list is:

```
golang-1.23, golang-1.24, golang-1.25, libsigc++30,
libxml++, rubygem-cool.io, rust-1.75
```

Every offender uses only `.` and/or `+`, so this PR's character set fully covers the current repo. `libsigc++30` isn't explicitly tested but follows the same pattern (`libsigc__30`) — optionally worth adding as a 5th row.

### Optional, non-blocking

1. **Future-proofing** — a regex would catch any new invalid character that slips in later:
   ```go
   var rpmMacroNameInvalid = regexp.MustCompile(`[^A-Za-z0-9_]`)
   packageFileNameMacroFormat := rpmMacroNameInvalid.ReplaceAllString(
       strings.TrimSuffix(specFileName, ".spec"), "_")
   ```
   The explicit list you have is also fine and arguably more readable — author's call.

2. **Pre-existing nit (out of scope)** — `strings.Replace(specFileName, ".spec", "", 1)` strips the first `.spec`, not strictly the suffix; `strings.TrimSuffix` would be safer. Mentioning only because the surrounding line is being touched. Happy to defer to a follow-up.

3. **Theoretical collision** — after this change, `foo-bar.spec` / `foo.bar.spec` / `foo+bar.spec` would all map to the same macro name. I confirmed no such collisions exist today, but a brief comment noting the assumption could help future maintainers.

Thanks for the cleanup — those rpmbuild warnings have been noisy for a while.

mfrw (approved):

toolkit/tools/versionsprocessor/versionsprocessor.go
```diff
@@ -189,9 +189,12 @@ func processPackageVersionString(packageVersionString string, specFileName strin
 	release := releaseVerSplit[2]
 	releaseClean := strings.Replace(release, distTag, "", 1)
 
-	// strip out the .spec suffix and replace '-' with '_' as RPM macros cannot have '-'
+	// strip out the .spec suffix and replace characters invalid in RPM macro names with '_'
+	// RPM macro names may only contain alphanumeric characters and underscores.
 	packageFileNameMacroFormat := strings.Replace(specFileName, ".spec", "", 1)
 	packageFileNameMacroFormat = strings.ReplaceAll(packageFileNameMacroFormat, "-", "_")
+	packageFileNameMacroFormat = strings.ReplaceAll(packageFileNameMacroFormat, ".", "_")
+	packageFileNameMacroFormat = strings.ReplaceAll(packageFileNameMacroFormat, "+", "_")
```
PawelWMS: Thanks for fixing this, @kgodara912! I see you found this before I did. Having said that, would you mind switching to a sanitization method like [this one](https://github.com/microsoft/azurelinux/pull/16955/changes#diff-7d40712c36200c09d1286783e576b519210555f0e92e405aa51ea00178a572eaR165-R180)? The idea is to cover all cases with characters RPM macros don't accept in a macro name. It doesn't have to be exactly the code from the link, I'm fine with a regex or anything else. The bottom line is to do an allow set approach and replace all characters outside that set.

kgodara912: Though I have replaced all the non-macro chars with _, is there any possible use case where it can fail?

PawelWMS: Looks good to me for the time being. If anything else fails, we'll learn. I guess at this point you could try a full, non-delta build, grab the logs for all packages and have Copilot sift through them in search of any warning/errors/other issues, which may have been caused by the `azl_*` macros.

PawelWMS (approved):

toolkit/tools/versionsprocessor/versionsprocessor.go
kgodara912: Sure, I have triggered non-delta [full build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1111682&view=results). Once completed, will update the analysis here.

kgodara912:
Full build is successful. The only warning, "needs whitespace before body", except few in extended was in python-repoze-lru.spec: line 63, which is a changelog line having a macro. No other warning in toolchain/base rpm logs for the same.

PR #16917, opened 2026-04-28
fix: disable building elfutils-default-yama-scope package

ddstreet:
This package reduces security by reducing the yama.ptrace_scope sysctl setting.
    
Upstream has slightly adjusted how they deal with this topic, as discussed in this FESCo discussion:
https://pagure.io/fesco/issue/3569
    
The upstream change is implemented according to this proposal:
https://fedoraproject.org/wiki/Changes/Restrict_ptrace_by_default
    
While upstream has decided to (partially) retain the reduction in yama security scope in certain situations, we have decided to remove the package entirely, which will leave the yama ptrace scope at its kernel default of 1.
    
The discussion of impact to end users is well documented in the above linked upstream discussion/proposal, as well as the kernel documentation on the specific meanings of each yama ptrace_scope value:
https://www.kernel.org/doc/Documentation/security/Yama.txt

This also removes `Recommends: default-yama-scope` from the gdb package.


base/comps/gdb/gdb.comp.toml
```diff
@@ -0,0 +1,7 @@
+[components.gdb]
+
+[[components.gdb.overlays]]
+description = "Remove recommends: default-yama-scope"
+type = "spec-search-replace"
```
reubeno: Would `spec-remove-tag` with a tag of "Recommends" and a value of "default-yama-scope" work in this case?

reubeno:
@ddstreet Did we test to make sure that the kernel default is, in fact, what we would like?

ddstreet:
force-pushed additional commit to remove the package from the vm-base image as well (was part of a separate branch to update the vm-base image, but it belongs here)

base/comps/gdb/gdb.comp.toml
ddstreet: I tried it but unfortunately it fails, presumably there is a bug in azldev:

```
$ cat base/comps/gdb/gdb.comp.toml 
[components.gdb]

[[components.gdb.overlays]]
description = "Remove recommends: default-yama-scope"
type = "spec-remove-tag"
tag = "Recommends"
value = "default-yama-scope"

$ azldev comp prepare-sources -o /tmp/gdb --force gdb
13:17:47 INF Preparing sources component=gdb outputDir=/tmp/gdb
13:17:47 INF Getting component from git repo component=gdb upstreamComponent=gdb branch=f43 upstreamCommit="" snapshot=2026-02-24T00:00:00-08:00
13:17:47 INF   Cloning git repo repoURL=https://src.fedoraproject.org/rpms/gdb.git
13:17:49 INF Using upstream distro snapshot time snapshotDateTime=2026-02-24T00:00:00-08:00 commitHash=77b7c5b27b3e56b5797a9e74ba4a5a285f8fdae8
13:17:49 INF   Checking out git commit commitHash=77b7c5b27b3e56b5797a9e74ba4a5a285f8fdae8
13:17:49 INF Downloading source file... progress=1/1 fileName=gdb-17.1.tar.xz URI=https://src.fedoraproject.org/repo/pkgs/gdb/gdb-17.1.tar.xz/sha512/f1a6751e439a2128fecf3eae8b57c1608a0dc7cfe79b4356a937874e5a42bb2df0aba36eb6a9452c41966908b9a59076c7cad9720f684688ab956b65080f1d7c/gdb-17.1.tar.xz destPath=/tmp/azldev-clone-535319632/gdb-17.1.tar.xz
13:17:50 INF   Applying overlays component=gdb
13:17:50 ERR Error: failed to prepare sources for component "gdb":
failed to apply overlays for component `gdb`:
failed to apply overlays for component `gdb`:
failed to apply `spec-remove-tag` overlay:
failed to apply overlay to spec `/tmp/gdb/gdb.spec`:
failed to remove tag `Recommends` from spec:
tag `Recommends` with value `default-yama-scope` not found in spec:
no such tag

```

ddstreet:
> @ddstreet Did we test to make sure that the kernel default is, in fact, what we would like?

yes it is

base/comps/gdb/gdb.comp.toml
reubeno: Can you try setting `package = "headless"` as well? I didn't realize that the `Recommends` was specifically on `gdb-headless`.

By default, the tag updates only apply to the preamble (i.e., not subpackages).

Assuming this works, then the bug here is that the error messages aren't great. "not found in spec" should really be "not found in spec preamble" or something clearer.

christopherco (approved):

christopherco:
Needs a rebase, but change LGTM

reubeno (approved):
Thanks for taking the feedback, Dan :smile:

base/comps/gdb/gdb.comp.toml
ddstreet: yep that works thanks, updated to use spec-remove-tag

ddstreet:
> Thanks for taking the feedback, Dan 😄

np, i was not aware of needing the section specifier, thanks for the pointer :)

PR #16921, opened 2026-04-28
fix: remove vm-base packages bc, binutils, elfutils

ddstreet:
The base image does not need development tools in it.

The package elfutils-debuginfod-client is also implicitly removed from the base image.

This is on top of #16917


christopherco (changes requested):
Change generally looks good. Please rebase and evaluate if certain commits are still needed, as I believe some of these may now be already merged.

christopherco (approved):

PR #16923, opened 2026-04-28
feat(mock-core-configs): add correct mock config for distro

reubeno:
Move mock-core-configs from an inline components.toml entry to a dedicated comp definition with overlays that ship an Azure Linux 4.0 chroot config out of the box.

The template targets releasever=4.0, dist=azl4, package_manager=dnf5, and bootstraps from an alpha container image. dnf is pointed at the alpha2 RPMs (gpgcheck=0 since alpha2 staging metadata is unsigned). This includes non-base RPMs as well.

The chroot_setup_cmd mirrors the internal stage2 list (minus azurelinux-stage1-compat).

End-user effect: `dnf install mock` on AZL 4.0 yields a working default config so `mock <srpm>` produces AZL 4.0 RPMs without any extra setup.

Resolves: [AB#18772](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/18772)


base/comps/mock-core-configs/mock-core-configs.comp.toml
```diff
@@ -0,0 +1,86 @@
...
+[[components.mock-core-configs.overlays]]
+description = "Register azure-linux-4-x86_64.cfg as a Source"
+type = "spec-insert-tag"
+tag = "Source"
+value = "azure-linux-4-x86_64.cfg"
+
+[[components.mock-core-configs.overlays]]
+description = "Register azure-linux-4-aarch64.cfg as a Source"
+type = "spec-insert-tag"
+tag = "Source"
```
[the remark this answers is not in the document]
reubeno: Confirmed that this spec already uses unnumbered sources.

base/comps/mock-core-configs/azure-linux-4.tpl
```diff
@@ -0,0 +1,56 @@
...
+[main]
+keepcache=1
+debuglevel=2
+reposdir=/dev/null
+logfile=/var/log/yum.log
+retries=20
+obsoletes=1
+gpgcheck=0
+assumeyes=1
+syslog_ident=mock
```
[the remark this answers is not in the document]
reubeno: We will be required to update this regardless for post-alpha, since the URLs will change. As part of that item, we will review this.

base/comps/mock-core-configs/mock-core-configs.comp.toml
```diff
@@ -0,0 +1,86 @@
...
+# since spec-append-lines runs us *after* that find we explicitly echo our
+# files into the same .cfgs manifest the spec uses for %files.
+[[components.mock-core-configs.overlays]]
+description = "Install Azure Linux 4.0 configs and add them to the %files manifest"
+type = "spec-append-lines"
+section = "%install"
+lines = [
+    "",
+    "# Azure Linux 4.0 configs (added by azldev overlay).",
+    "install -m 0644 %{SOURCE1} %{buildroot}%{_sysconfdir}/mock/templates/azure-linux-4.tpl",
```
christopherco: >SOURCE1 [](http://example.com/codeflow?start=23&length=7)

question(non-blocking): since we're inserting these sources, the position could change if upstream mock spec adds a new source entry. Is there a way to link these references with their corresponding source toml entry? Not needed for this PR, but just noticing it.

base/comps/mock-core-configs/azure-linux-4.tpl
```diff
@@ -0,0 +1,56 @@
+config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils dnf5 azurelinux-release-common findutils gawk glibc-minimal-langpack grep gzip info patch azurelinux-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz'
```
christopherco, line 1: question: Should dnf5-plugins be up here? Or does that naturally resolve in some other way with mock?

base/comps/mock-core-configs/azure-linux-4.tpl
```diff
@@ -0,0 +1,56 @@
...
+# TODO: Replace the explicit chroot_setup_cmd package list above with a
```
christopherco, line 2: >TODO [](http://example.com/codeflow?start=2&length=4)

AB#19411

base/comps/mock-core-configs/azure-linux-4.tpl
reubeno: It doesn't need to be there, from my testing. It is still installed and used to run `dnf builddep`, which `mock` uses to install the build dependencies for the spec.

base/comps/mock-core-configs/mock-core-configs.comp.toml
reubeno: That's a good call-out. We don't have a great way to link it. I've made a change to at least pick numbers that are unlikely to conflict -- it's a bit unpleasant, but this way we'll see a conflict if we try to add a source with an explicit number and one already exists.

For the longer term, we should think about whether there's something better we can do with the overlays.

christopherco (approved):
:shipit:

PR #16937, opened 2026-04-29
feat: set vm-base grub menu to 1 second timeout

ddstreet:
Without a timeout, it's not possible to enter the grub menu from the azure serial console, which prevents debugging and/or recovery using the grub menu.

This also increases the default boot time of all vm-base image deployments; 1 second was chosen as a compromise between reducing boot time delays, and allowing end user access to the grub menu (without needing to change grub configuration).


christopherco (approved):
:shipit:

PR #16947, opened 2026-04-29
feat(image-tests): initial pytest-based static image validation

reubeno:
Introduce a pytest harness for offline validation of built Azure Linux images, wired into azldev via the new [test-suites] TOML config.

A small, representative set of tests is included to establish the extension pattern: shared cases live under base/images/tests/cases/ and image-specific cases under cases/<image-name>/ (auto-restricted to that image by the pytest plugin). Image-type plumbing — VM mount via guestmount, container extract via skopeo+umoci — and shared fixtures (rootfs, os_release, installed_packages, partition_table) live under utils/, with native-tool dependencies declared next to their use and validated up-front by a preflight check.

### DIFF
--- a/base/images/images.toml
+++ b/base/images/images.toml
@@ -1,10 +1,24 @@
 [images.vm-base]
 description = "VM Base Image"
 definition = { type = "kiwi", path = "vm-base/vm-base.kiwi" }
+tests.test-suites = [{ name = "static-image-checks" }]
+
+[images.vm-base.capabilities]
+machine-bootable = true
+container = false
+systemd = true
+runtime-package-management = true
 
 [images.container-base]
 description = "Container Base Image"
 definition = { type = "kiwi", path = "container-base/container-base.kiwi", profile = "core" }
+tests.test-suites = [{ name = "static-image-checks" }]
+
+[images.container-base.capabilities]
+machine-bootable = false
+container = true
+systemd = false
+runtime-package-management = true
 
 [images.container-distroless-minimal]
 description = "Container Distroless Minimal Image"
@@ -21,3 +35,26 @@ definition = { type = "kiwi", path = "container-base/container-base.kiwi", profi
 [images.wsl]
 description = "WSL Image"
 definition = { type = "kiwi", path = "wsl/wsl.kiwi" }
+
+# ---- Test suites ----
+
+# Single shared pytest suite for static (offline) image validation.
+[test-suites.static-image-checks]
+type = "pytest"
+description = "Offline image validation (shared + image-specific tests)"
+
+[test-suites.static-image-checks.pytest]
+working-dir = "tests"
+install = "pyproject"
+test-paths = ["cases/"]
+# {capabilities} is substituted by azldev as a comma-separated list of
+# the names of capabilities set to `true` for the image (e.g.
+# "machine-bootable,systemd,runtime-package-management"). Names use
+# the same dash-case form declared above. The pytest plugin parses
+# this string into a set; @pytest.mark.require_capability("...") tests
+# that don't match are skipped.
+extra-args = [
+    "--image-path", "{image-path}",
+    "--image-name", "{image-name}",
+    "--capabilities", "{capabilities}",
+]
--- a/base/images/tests/README.md
+++ b/base/images/tests/README.md
@@ -1,0 +1,142 @@
+# Azure Linux Image Tests
+
+Static validation framework for built Azure Linux images (VM and container).
+Mounts images read-only and runs pytest tests against the filesystem
+without booting.
+
+## How it gets invoked
+
+These tests are wired into `azldev` via the `[test-suites.static-image-checks]`
+table in `base/images/images.toml`, and referenced by each image's
+`tests.test-suites`. The standard entry point is:
+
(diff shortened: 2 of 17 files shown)


base/images/tests/conftest.py
```diff
@@ -0,0 +1,203 @@
...
+
+    p = Path(tempfile.mkdtemp(prefix="azl-image-tests-"))
+    logger.debug("Work dir (temp, will be removed at session end): %s", p)
+    try:
+        yield p
+    finally:
+        # Use buildah unshare for cleanup so any read-only dirs left
+        # behind by rootless umoci unpack are removable.
+        logger.debug("Removing temp work dir %s", p)
+        rc = subprocess.run(
```
christopherco: **issue (non-blocking):** `buildah` cleanup raises `FileNotFoundError` on VM-only hosts and leaks the temp workdir.

`subprocess.run(["buildah", "unshare", "rm", "-rf", str(p)], check=False, ...)` doesn't suppress `FileNotFoundError` — `check=False` only swallows non-zero exits. `buildah` is registered `when="container"` (`utils/extract.py:45-50`), so a host with `libguestfs-tools` but no `buildah` passes preflight, runs all VM tests successfully, and then crashes during session teardown. The `shutil.rmtree` fallback never executes because it lives inside the unreachable `if rc.returncode != 0:` branch.

Suggested fix:

```python
try:
    rc = subprocess.run(
        ["buildah", "unshare", "rm", "-rf", str(p)],
        check=False, capture_output=True, text=True,
    )
    failed = rc.returncode != 0
except (FileNotFoundError, OSError):
    failed = True
if failed:
    shutil.rmtree(p, ignore_errors=True)
```

Or short-circuit with `shutil.which("buildah")` before invoking. Same pattern in `utils/extract.py:165-183` is reachable only after preflight validates `buildah`, so lower priority there.

base/images/tests/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+# Container image — shared + container-specific tests
+uv run pytest cases/ \
+    --image-path /path/to/image.oci.tar.xz \
+    --image-name container-base \
+    --capabilities container,runtime-package-management
+```
+
+Test selection follows standard pytest positional arguments. Tests
+under `cases/<image-name>/` auto-skip when `--image-name` doesn't match
```
christopherco: **issue (non-blocking):** Drift — the README contradicts itself about how the per-image marker is applied.

Lines 44-45 read:

> Tests under `cases/<image-name>/` auto-skip when `--image-name` doesn't match (via the `image` marker, applied in each subdirectory's `conftest.py`).

But there are no per-subdir `conftest.py` files in the tree (only empty `__init__.py`s). The actual mechanism is the centralized `pytest_collection_modifyitems` hook in `utils/pytest_plugin.py:156-181`, which the README correctly describes later at lines 114-116.

Suggested fix: replace lines 44-45 with the second-paragraph wording, or just delete them since the centralized mechanism is already documented.

base/images/images.toml
```diff
@@ -26,3 +40,23 @@ definition = { type = "kiwi", path = "container-base/container-base.kiwi", profi
...
+# Single shared pytest suite for static (offline) image validation. The
+# tests directory contains shared cases plus per-image-name subdirs. Each
+# image receives its own --image-name and --capabilities so tests can
+# self-select via @pytest.mark.image and @pytest.mark.require_capability.
+[test-suites.static-image-checks]
+type = "pytest"
+description = "Offline image validation (shared + image-specific tests)"
+
+[test-suites.static-image-checks.pytest]
+working-dir = "tests"
```
christopherco, line 47: **question (non-blocking):** The `{capabilities}` substitution contract isn't documented anywhere in this PR — what exactly does `azldev` emit for a TOML sub-table of booleans?

The `extra-args` template:

```toml
extra-args = [
    "--image-path", "{image-path}",
    "--image-name", "{image-name}",
    "--capabilities", "{capabilities}",
]
```

…assumes `azldev` flattens `[images.vm-base.capabilities]` (a sub-table of `key = bool`) into a comma-separated list of *true-keyed names*, which `utils/pytest_plugin.py:52-56` then parses. If `azldev` instead emits `key=value` pairs (e.g. `"machine-bootable=true,container=false,..."`), the failure mode is silent: `derive_image_type_from_capabilities` finds no match, falls through to extension detection, and **every** `@pytest.mark.require_capability(...)` test silently skips.

Suggested fix: at minimum, add a comment in `images.toml` near the `extra-args` block stating the expected format (e.g. `# {capabilities} is substituted as a comma-separated list of true-keyed capability names`). Ideally, add a smoke check that asserts non-zero capability-gated tests actually executed, so a future change in `azldev`'s substitution logic doesn't degrade silently to "all skipped, all green".

christopherco (approved):
Added some non-blocking comments. Otherwise LGTM to merge as is (after fixing up merge conflicts of course)

tobiasb-ms (approved):

base/images/images.toml
reubeno: I've added a comment there; we'll also make sure that there's appropriate docs added on the azldev side too.

PR #16949, opened 2026-04-29
ci: add lock-file drift check parallel to render

dmcilvaney:
Adds a second drift-check flow alongside the existing rendered-specs check in the reusable workflow. Runs 'azldev component update -a' in the same sandboxed container, uses the JSON output (filtering on each entry's 'changed' boolean) to drive the PR comment, and produces a 'git diff -- locks/' patch artifact for 'git apply' remediation.

**NOTE**: This is an intermediate implementation, the final version is waiting on improvements to alzdev update+render flows. Since this is temporary, this workflow piggybacks on the existing render infrastructure to avoid more widespread changes. (https://dev.azure.com/mariner-org/mariner/_workitems/edit/19561)

* Two new jobs: 'update_locks' (mirrors 'render') and 'update_locks_comment' (mirrors 'comment'), with the same trusted-base / untrusted-PR-head split and security posture.
* 'git add --intent-to-add' before 'git diff' so net-new lock files show up in the patch.
* Empty-patch case: 'rm -f' the file so hashFiles() returns '' and the upload steps skip cleanly (it does not skip on a zero-byte file).
* New post_locks_comment.py mirrors post_render_comment.py with marker '' and header '🔒❌ Lock files are out of date' to differentiate from the render comment ('📄❌ Rendered specs are out of date').
* Concurrency group renamed to 'pr-drift-check-...' to reflect both flows; stub on default branch is left untouched (independent scope).
* Both check-rendered-specs jobs and the ADO sources-upload step now resolve directory paths via 'jq -r .project.{renderedSpecsDir, lockDir}' from 'azldev config dump'. 'jq' added to the runner Dockerfile and the ADO install step.

### DIFF
--- a/.github/workflows/check-rendered-specs.yml
+++ b/.github/workflows/check-rendered-specs.yml
@@ -1,13 +1,16 @@
-# Reusable workflow — renders specs from a PR and checks for drift.
+# Reusable workflow — renders specs from a PR + resolves lock files, and
+# checks both for drift.
 #
 # Called by the stub on the default branch (check-rendered-specs-stub.yml) via
 # pull_request_target. The stub provides the PR details; this workflow does all
-# the real work:
+# the real work via two parallel flows:
 #   1. Checks out base branch (trusted tools/scripts)
 #   2. Checks out PR head into pr-head/ (untrusted data — TOML configs, specs)
-#   3. Renders specs inside a privileged container using azldev -C pr-head/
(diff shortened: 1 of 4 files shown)


christopherco (approved):

christopherco (changes requested):
Please rebase to squash the commits. Otherwise LGTM

christopherco (approved):

PR #16953, opened 2026-04-30
chore(render): update rendered specs with work-in-progress azldev features

dmcilvaney:
This render pass uses several work-in-progress azldev features, but is close to the final expected output.

This render also includes the result of temporary (un-commited) toml changes for `cri-tools1.35` and `glusterfs`.


locks/words.lock
```diff
@@ -0,0 +1,4 @@
+# Managed by azldev component update. Do not edit manually.
```
dmcilvaney: remove lock

christopherco (approved):
:shipit:

christopherco (approved):

PR #16958, opened 2026-04-30
fix(erlang): disable erlang-wx wxWidgets bindings sub-package

reubeno:
We do not plan to ship wxGTK and need to remove the dependency.
The upstream spec already supports this configuration via the
`%global __with_wxwidgets` toggle: when set to 0, every wx-conditional
`%package wx` definition, every `Requires: %{name}-wx` line, and the
wxGTK BuildRequires are skipped wholesale. Flipping the global is the
most surgical fix.

The `Requires: %{name}-wx` line in the erlang-src %package preamble is
the only such line in the upstream spec that is NOT wrapped in a
`%if %{__with_wxwidgets}` block (it lives directly inside
`%if %{__with_sources}`), so the toggle alone cannot drop it. Strip it
explicitly so the rebuilt erlang-src RPM does not pull in a now-
nonexistent erlang-wx package.

---
**Validated**: via local azldev build + koji build

### DIFF
--- a/base/comps/erlang/erlang.comp.toml
+++ b/base/comps/erlang/erlang.comp.toml
@@ -4,3 +4,40 @@
 
 [components.erlang.build]
 without = ["doc"]
+
+# Disable the erlang-wx sub-package (Erlang's wxWidgets bindings). AZL is a
+# headless cloud distro and does not ship wxGTK; building erlang-wx would
+# pull in a desktop GUI stack at build time and would force every other
+# erlang sub-package whose %if %{__with_wxwidgets} block adds a
+# `Requires: erlang-wx` (erlang-meta, -debugger, -dialyzer, -et, -observer,
+# -reltool, -src) to drag the wx subpackage into the runtime closure.
+#
+# The upstream spec already supports this configuration via the
+# `%global __with_wxwidgets` toggle: when set to 0, every wx-conditional
+# `%package wx` definition, every `Requires: %{name}-wx` line, and the
+# wxGTK BuildRequires are skipped wholesale. Flipping the global is by
+# far the most surgical fix.
+[[components.erlang.overlays]]
+description = "Disable erlang-wx sub-package (wxGTK GUI bindings) — AZL is a headless cloud distro"
+type = "spec-search-replace"
+regex = '^%global __with_wxwidgets 1$'
+replacement = '%global __with_wxwidgets 0'
+
+# The erlang-src %package preamble lives inside `%if %{__with_sources}` and
+# carries unconditional `Requires: %{name}-<sub>` lines for every other
+# sub-package — including all of the wxwidgets-gated ones (common_test,
+# debugger, dialyzer, et, megaco, observer, reltool, wx). Because these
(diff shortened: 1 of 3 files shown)


base/comps/erlang/erlang.comp.toml
```diff
@@ -4,3 +4,37 @@
...
+# Strip it so the rebuilt erlang-src RPM does not pull in a now-nonexistent
+# erlang-wx package. Scope the regex to `%package src` so it only targets
+# the unconditional Requires; the same line in other sub-packages is
+# already dead code under `%if %{__with_wxwidgets}=0` and stays untouched.
+[[components.erlang.overlays]]
+description = "Drop unconditional Requires: erlang-wx from erlang-src (paired with __with_wxwidgets=0 above)"
+type = "spec-search-replace"
+section = "%package"
+package = "src"
+regex = '^Requires: %\{name\}-wx%\{\?_isa\} = %\{version\}-%\{release\}$'
```
christopherco: **issue (blocking):** `%package src` will have 7 dangling `Requires:` after this PR — `dnf repoclosure` will fail.

Verified against the parent-branch rendered spec (no AZL overlays touch this block, so it's upstream Fedora content verbatim): when `__with_wxwidgets=0`, the 8 wx-related `%package` blocks (`wx`, `common_test`, `debugger`, `dialyzer`, `et`, `megaco`, `observer`, `reltool`) are all skipped — but `%package src`'s preamble has 35 unconditional `Requires:` lines, and 7 of them point at the wx-gated siblings (specs/e/erlang/erlang.spec lines 616, 619, 620, 626, 631, 633, 638):

```
Requires: %{name}-common_test
Requires: %{name}-debugger
Requires: %{name}-dialyzer
Requires: %{name}-et
Requires: %{name}-megaco
Requires: %{name}-observer
Requires: %{name}-reltool
```

The PR description's claim — *"the `Requires: %{name}-wx` line ... is the only such line ... NOT wrapped in a `%if %{__with_wxwidgets}` block"* — is true only if "such line" means the literal `Requires: %{name}-wx` token. The reasoning is forward-direction (which packages depend on `-wx`); the failure is reverse-direction (which `Requires:` inside `-src` dangle when `-wx`'s wx-gated *siblings* vanish).

`koji build` validates RPM build success but doesn't run repoclosure or install into a candidate sysroot, so this slips past the stated validation.

**suggestion:** one of —
1. Extend this overlay's regex (or add a third) to also strip the 7 sibling `Requires:` from `%package src`. The existing `section`/`package` scoping already isolates the right block.
2. `%if %{__with_wxwidgets}`-wrap the wx-gated subset (more upstream-friendly; safer if `__with_wxwidgets` is ever flipped back).
3. Drop `erlang-src` from `exceptions.packages.toml` if it has no consumers in a wx-less AZL.

**note (non-blocking):** worth cleaning up in the same PR — the 8 stale `exceptions.packages.toml` entries (`erlang-wx` + the 7 siblings) that route to `rpm-sdk` but will no longer be produced.

reubeno: Excellent catch; I've got a fix!

christopherco (approved):

PR #16960, opened 2026-04-30
fix(gtk4): add missing libdrm BuildRequires

reubeno:
gtk4 4.20.3's meson hard-requires libdrm but the upstream Fedora spec
does not declare a BuildRequires on libdrm-devel; without it the build
fails at meson.build:731. This appears to be an upstream gap, so add
the BR to avoid build failures when some dependencies change.

--
_Build validated locally via azldev and in koji_


christopherco (changes requested):
Change LGTM. I will request for lock file generation so we capture this change in the package's changelog

reubeno:
> Change LGTM. I will request for lock file generation so we capture this change in the package's changelog

Done. Re-review, please, @christopherco?

christopherco (approved):

PR #16963, opened 2026-04-30
fix(m17n-lib): disable optional Anthy IME module

reubeno:
m17n-lib.spec gates the m17n-lib-anthy sub-package via
`%bcond_without anthy` (default ON when fedora is set). Flip it to OFF
via build.without so the anthy-unicode dependency can be dropped from
base. The spec is well-gated -- libmimx-anthy.so is rm'd from %install
when --without anthy is used -- so the disabled build is clean.

--
_Build validated via local azldev + koji build._


christopherco (changes requested):
Change LGTM. Please resolve merge conflict and also update lock file.

reubeno:
> Change LGTM. Please resolve merge conflict and also update lock file.

Done. Re-review, please, @christopherco? Check failures appear infra-related and not specific to this change.

christopherco (approved):

PR #16964, opened 2026-04-30
fix(openscap): disable optional apt integration

reubeno:
Disable openscap's apt-libs probe path. Upstream gates it on
`%bcond_without apt` (default-on for Fedora, default-off for
CentOS/ELN). We only need to support RPM environments.

--
_Build validated locally via azldev + koji._


ddstreetmicrosoft (approved):
lgtm

PR #16965, opened 2026-04-30
fix(ostree): disable optional installed-tests sub-package

reubeno:
ostree-tests is the GNOME installed-tests harness; its only base
unresolved dep is `/usr/bin/gjs` (the GNOME JavaScript binding from
gjs). This allows removing the dependency on the latter.

ostree.spec has `%bcond_without tests` (default ON); passing
`--without tests` skips the entire installed-tests build, install,
and `%files tests` block.

--
_Build validated via local azldev and via koji._


ddstreetmicrosoft (approved):
lgtm

PR #16966, opened 2026-04-30
fix(po4a): drop texlive-kpathsea Requires

reubeno:
po4a is a multi-format documentation translation tool with ~15 format
modules (man, sgml, xml, asciidoc, markdown, docbook, dia, halibut,
tex, ...). The texlive-kpathsea / texlive-kpathsea-bin Requires are
only consumed by Locale::Po4a::TeX, which shells out to `kpsewhich`
to resolve TeX \input/\include paths. All other format modules work
without it.

This overlay lets po4a stay in base without requiring
texlive-kpathsea(-bin) to live in base.

--
_Build validated via local azldev + koji._


ddstreetmicrosoft (approved):
lgtm

PR #16967, opened 2026-04-30
fix(rsyslog): disable optional mongodb output module

reubeno:
The rsyslog-mongodb sub-package logs records into MongoDB collections;
its only consumers in base are the libbson/libmongoc client libraries
(which are not in base).

rsyslog still provides PostgreSQL, MySQL, ElasticSearch, Kafka, file,
and forwarding outputs.

rsyslog.spec defines `%bcond_without mongodb` (default ON); passing
`--without mongodb` skips building the rsyslog-mongodb sub-package
and its libbson/libmongoc Requires.

--
_Build validated via local azldev + koji._


ddstreetmicrosoft (approved):
lgtm

PR #16969, opened 2026-04-30
fix(tuna): disable optional oscilloscope sub-package

reubeno:
Tuna's spec gates the oscilloscope sub-package on
`%bcond oscilloscope %{undefined rhel}` so it builds by default on
Fedora-like systems. The oscilloscope sub-package is a matplotlib-GTK3
visualizer for tuna data with a dependency on python3-matplotlib-gtk3.
Disabling the bcond stops producing the oscilloscope sub-pkg entirely
and removes the corresponding Requires.

The upstream spec gates only the `%package -n oscilloscope` and
`%files -n oscilloscope` blocks on the bcond -- the `oscilloscope`
console-script entry-point is still installed by `%pyproject_install`,
leaving an unpackaged file when the bcond is off (rpmbuild fails).
Add an overlay to strip /usr/bin/oscilloscope explicitly when
oscilloscope is disabled.

--
_Build validated locally and via koji._


ddstreetmicrosoft (approved):
lgtm

PR #16970, opened 2026-04-30
feat(iso): add dual-arch (x86_64/aarch64) VM ISO installer

kgodara912:
Add a kiwi-based ISO installer image that builds and installs Azure Linux on both x86_64 and aarch64 using runtime architecture detection.

Image definition:
- vm-iso-installer: kiwi ISO image with dual-arch repos (alpha2-prod) and arch-filtered package entries for grub2-efi-* and shim
- config.sh: runtime uname -m detection for GRUB/shim/EFI variables; placeholder substitution in kickstart heredocs for arch-specific package names; independent arch detection in %post --nochroot sections for EFI binary paths
- Build-time dry-run validation of offline repo completeness to catch missing packages before producing the ISO
- Standard and LUKS-encrypted installation modes via embedded kickstarts with offline repo (file://)
- Security hardening: SSH key regeneration, machine-id reset, PermitRootLogin prohibit-password, SELinux autorelabel
- Register image in images.toml

###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

###### Change Log  

- Change
- Change
- Change

**YES/NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  

###### Test Methodology

- Pipeline build id: xxxx

### DIFF
--- a/base/images/vm-iso-installer/20-wired-dhcp.network
+++ b/base/images/vm-iso-installer/20-wired-dhcp.network
@@ -1,0 +1,8 @@
+[Match]
+Name=en* eth*
+
+[Network]
+DHCP=yes
+
+[DHCPv4]
+UseDNS=yes
--- a/base/images/vm-iso-installer/config.sh
+++ b/base/images/vm-iso-installer/config.sh
@@ -1,0 +1,267 @@
+#!/bin/bash
+# KIWI config.sh — post-bootstrap configuration for Anaconda TUI Offline Installer ISO
+set -euo pipefail
+
+#----------------------------------------------------------------------
+# Architecture detection — set arch-specific package and EFI names
+#----------------------------------------------------------------------
+ARCH=$(uname -m)
+case "$ARCH" in
+    x86_64)
+        GRUB_EFI_PKG="grub2-efi-x64"
+        GRUB_EFI_MOD_PKG="grub2-efi-x64-modules"
+        GRUB_EFI_CDBOOT_PKG="grub2-efi-x64-cdboot"
+        SHIM_EFI="shimx64.efi"
+        GRUB_EFI="grubx64.efi"
+        BOOT_EFI="BOOTX64.EFI"
+        ;;
+    aarch64)
+        GRUB_EFI_PKG="grub2-efi-aa64"
+        GRUB_EFI_MOD_PKG="grub2-efi-aa64-modules"
+        GRUB_EFI_CDBOOT_PKG="grub2-efi-aa64-cdboot"
+        SHIM_EFI="shimaa64.efi"
+        GRUB_EFI="grubaa64.efi"
+        BOOT_EFI="BOOTAA64.EFI"
+        ;;
+    *)
+        echo "ERROR: Unsupported architecture: $ARCH" >&2
+        exit 1
+        ;;
+esac
+echo "=== Architecture: $ARCH ==="
+echo "  GRUB EFI package: $GRUB_EFI_PKG"
+echo "  Shim EFI binary:  $SHIM_EFI"
+
+#----------------------------------------------------------------------
+# Download all target-install packages + deps for the offline repo
+#----------------------------------------------------------------------
+# During the ISO build we have network access to the repo.
+# Download every package that anaconda will install on the target,
+# including all transitive dependencies, into /opt/azl-offline-repo/.
+# This lets the offline kickstart install from file:// with no network.
+
+OFFLINE_REPO="/opt/azl-offline-repo"
+mkdir -p "$OFFLINE_REPO"
+
+# === Single source of truth for target-install packages ===
+# These go into the kickstart %packages and the offline repo.
+# Any package needed on the installed system belongs here.
+INSTALL_PKGS=(
+    bash
+    coreutils
+    systemd
+    systemd-networkd
+    systemd-resolved
+    dnf5
+    grub2
+    "$GRUB_EFI_PKG"
+    "$GRUB_EFI_MOD_PKG"
+    shim
+    efibootmgr
+    kernel
+    kernel-modules
+    openssh-server
+    openssh-clients
+    sudo
+    vim-minimal
+    ca-certificates
+    azurelinux-release
+    setup
+    shadow-utils
+    util-linux
+    selinux-policy-targeted
+    audit
+    chrony
+    cracklib-dicts
+    glibc
+    glibc-langpack-en
+    cryptsetup
+    firewalld
+    iproute
+)
+
+# Extra packages needed in the offline repo for anaconda's runtime deps
+# but not listed in kickstart %packages (pulled via --resolve --alldeps).
+EXTRA_REPO_PKGS=(
+    "$GRUB_EFI_CDBOOT_PKG"
+    lvm2
+    e2fsprogs
+    dosfstools
+    device-mapper-persistent-data
+    mtools
+    grub2-tools-extra
+    libaio
+)
+
+echo "=== Downloading target-install packages + dependencies ==="
+# Kiwi removes repo configs after package installation, so repos from the .kiwi
+# file are NOT available during config.sh. Use --repofrompath with the CDN URL
+# directly. The URL matches the "azurelinux-base" repo in vm-iso-installer.kiwi.
+
+AZL_BASE_URL="https://stcontroltowerdevjwisitg.blob.core.windows.net/alpha2-prod/base/$ARCH"
+dnf5 download \
+    --setopt=reposdir=/dev/null \
+    --repofrompath=azl-base,"$AZL_BASE_URL" \
+    --repo=azl-base \
+    --resolve \
+    --alldeps \
+    --skip-unavailable \
+    --destdir="$OFFLINE_REPO" \
+    "${INSTALL_PKGS[@]}" "${EXTRA_REPO_PKGS[@]}" || {
+    echo "WARNING: dnf download had errors — some packages may be missing"
+}
+
+RPM_COUNT=$(ls "$OFFLINE_REPO"/*.rpm 2>/dev/null | wc -l)
+echo "=== Downloaded $RPM_COUNT RPMs ==="
+ls "$OFFLINE_REPO"/*.rpm 2>/dev/null | head -20 || true
+echo "..."
+
+# Build repo metadata
+createrepo_c "$OFFLINE_REPO"
+
+#----------------------------------------------------------------------
+# Validate offline repo completeness (dry-run install)
+#----------------------------------------------------------------------
+echo "=== Validating offline repo completeness ==="
+
+DRYRUN_ROOT=$(mktemp -d /tmp/azl-dryrun-XXXXXX)
+DRYRUN_ERRORS=$(dnf5 install \
+    --assumeno \
+    --installroot="$DRYRUN_ROOT" \
+    --releasever=4.0 \
+    --setopt=reposdir=/dev/null \
+    --repofrompath=offline,"file://$OFFLINE_REPO" \
+    --repo=offline \
+    "${INSTALL_PKGS[@]}" 2>&1) || true
+
+rm -rf "$DRYRUN_ROOT"
+
+if echo "$DRYRUN_ERRORS" | grep -qiE "No match for argument|nothing provides|cannot install"; then
+    echo "!!!"
+    echo "!!! FATAL: Offline repo is missing packages required by the kickstart!"
+    echo "!!!"
+    echo "$DRYRUN_ERRORS" | grep -iE "No match for argument|nothing provides|cannot install"
+    echo ""
+    echo "Fix: add the missing packages to INSTALL_PKGS."
+    exit 1
+else
+    echo "=== Dry-run passed — all kickstart packages resolve from offline repo ==="
+fi
+
+echo "=== Offline repo ready at $OFFLINE_REPO ==="
+
+#----------------------------------------------------------------------
+# Anaconda launcher symlink (script deployed via kiwi <file>)
+#----------------------------------------------------------------------
+ln -sf /usr/local/bin/anaconda-launcher.sh /usr/local/bin/install-azl
+
+#----------------------------------------------------------------------
+# Welcome banner
+#----------------------------------------------------------------------
+
+cat > /root/.bash_profile << 'PROFILEEOF'
+# Auto-install mode: if "azl.autoinstall" is on the kernel cmdline
+# (selected via the "Install Azure Linux 4.0" GRUB entry), launch
+# the installer automatically.
+#
+# Console selection logic:
+#   - Hyper-V (systemd-detect-virt = "microsoft") → user is on VGA (tty1)
+#   - QEMU/KVM/bare-metal with serial in cmdline  → user is on serial (ttyS0)
+#   - QEMU/KVM/bare-metal without serial          → user is on VGA (tty1)
+# This prevents the invisible tty1 from stealing the installer on QEMU -nographic.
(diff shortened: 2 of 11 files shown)


base/images/vm-iso-installer/vm-iso-installer.kiwi
```diff
@@ -0,0 +1,141 @@
...
+            target="/etc/systemd/network/20-wired-dhcp.network"
+            permissions="0644" />
+
+        <!-- ISO live boot support (Fedora dmsquash-live) -->
+        <package name="dracut-live" />
+
+        <!-- Hyper-V guest integration -->
+        <package name="hyperv-daemons" />
+
+        <!-- Anaconda TUI installer and dependencies -->
```
reubeno: Is there a reason to call out all the dependencies in this file? It will get resolved anyhow. We'll have other means in the future to observe if the actual resolved package set shifts, but I think it makes sense to only call out what we are intentionally requiring -- and let dependency resolution do its thing.

base/images/vm-iso-installer/vm-iso-installer.kiwi
```diff
@@ -0,0 +1,141 @@
...
+            path="https://stcontroltowerdevjwisitg.blob.core.windows.net/alpha2-prod/base/x86_64/" />
+    </repository>
+    <repository type="rpm-md" alias="azl4-alpha2-aarch64">
+        <source
+            path="https://stcontroltowerdevjwisitg.blob.core.windows.net/alpha2-prod/base/aarch64/" />
+    </repository>
+
+    <packages type="image">
+        <!-- Core group -->
+        <package name="azurelinux-release" />
```
reubeno, line 41: @ddstreet Would you advise that we have a release identity for the liveCD/ISO/installer? This one is geared toward installing our standard cloud image, but the make-up and config of the ISO itself is different. If we did that, then we'd be able to have that release package install any needed ISO-specific config files.

base/images/vm-iso-installer/config.sh
```diff
@@ -0,0 +1,878 @@
...
+
+#----------------------------------------------------------------------
+# Azure Linux product configuration for Anaconda
+#----------------------------------------------------------------------
+
+CONF_DIR="/etc/anaconda/conf.d"
+mkdir -p "$CONF_DIR"
+
+cat > "$CONF_DIR/azurelinux.conf" << 'PRODUCTEOF'
+# Azure Linux 4.0 — Anaconda configuration overrides
```
reubeno: Why not install this in a proper RPM, or at least as a file in the .kiwi?

base/images/vm-iso-installer/config.sh
```diff
@@ -0,0 +1,878 @@
...
+
+[User Interface]
+hidden_spokes = NetworkSpoke
+
+[License]
+eula =
+PRODUCTEOF
+
+#----------------------------------------------------------------------
+# Enable services needed by anaconda at ISO boot
```
reubeno: Could these be preset overrides for the ISO image that could be represented with an appropriate .preset file and a systemctl command to reflect the current presets into service enablement?

base/images/vm-iso-installer/config.sh
```diff
@@ -0,0 +1,878 @@
...
+    freetype
+    graphite2
+    libpng
+    mtools
+    grub2-tools-extra
+    libaio
+)
+
+echo "=== Downloading target-install packages + dependencies ==="
+dnf download \
```
reubeno: When this script is run in kiwi, do you know which repo(s) `dnf` will have access to? Among other things, we'll need to be mindful to include only packages from our `base` repo -- but that's not a concept that exists in `koji` right now, which is where we expect to build this image.

base/images/vm-iso-installer/config.sh
```diff
@@ -0,0 +1,878 @@
...
+    echo "=== Dry-run passed — all kickstart packages resolve from offline repo ==="
+fi
+
+echo "=== Offline repo ready at $OFFLINE_REPO ==="
+
+#----------------------------------------------------------------------
+# Anaconda launcher script
+#----------------------------------------------------------------------
+
+cat > /usr/local/bin/anaconda-launcher.sh << 'LAUNCHEOF'
```
reubeno: See above comment regarding bundling this script inside `config.sh`.

base/images/vm-iso-installer/config.sh
```diff
@@ -0,0 +1,878 @@
...
+GRUB_DISABLE_SUBMENU=true
+GRUB_TERMINAL_OUTPUT="console serial"
+GRUB_SERIAL_COMMAND="serial --speed=115200 --unit=0 --word=8 --parity=no --stop=1"
+GRUB_CMDLINE_LINUX="console=ttyS0,115200 console=tty0"
+GRUB_DISABLE_RECOVERY=true
+GRUBDEF
+
+# --- Security hardening for the installed target ---
+# Remove SSH host keys — sshd-keygen regenerates unique keys on first boot
+rm -f /etc/ssh/ssh_host_*_key /etc/ssh/ssh_host_*_key.pub
```
reubeno: Why are we manipulating the target? We don't want it to diverge from our standard bootable images, no?

reubeno (commented):
I left a first set of comments but main feedback is around bundling so much in the config.sh, and the custom manipulations of the install target that may diverge from the VM images that we're building.

Happy to chat about this via IM if helpful.

base/images/vm-iso-installer/config.sh
kgodara912: The reason for this is security enhancement of the installed system. These 4 lines will enforce ssh key regeneration after first boot, machine id regeneration so no copy from different machine, disablement of password-based root login from ssh so it will not allow root login from remote ssh with password, though user logins having sudo access will be allowed over ssh and force rebuilding of selinux labels after first boot on host system.

base/images/vm-iso-installer/vm-iso-installer.kiwi
kgodara912: Thank you for pointing this. I have reduced the packages needed and removed other dependencies.

base/images/vm-iso-installer/config.sh
kgodara912: Created a separate file and included in kiwi.

base/images/vm-iso-installer/config.sh
kgodara912: Added preset file and included in kiwi.

base/images/vm-iso-installer/config.sh
kgodara912: I have refactored to create separate scripts for almost all the scripts integrated in config.sh.

base/images/vm-iso-installer/config.sh
kgodara912: I have disabled other repos and enabled only azurelinux repo. I will check koji builds on how to include repos cleanly there.

base/images/vm-iso-installer/config.sh
ddstreetmicrosoft: issue: needing to do this here seems to indicate you're doing it wrong somewhere else.

base/images/vm-iso-installer/vm-iso-installer.kiwi
ddstreetmicrosoft: Fedora generates a common `LiveInstall` profile that all its spin-specific iso installers inherit from, and each spin-specific iso uses that spin's release package.

Since we decided to use flat individual kiwi xml instead Fedora's structured kiwi xml, we probably don't need to do exactly what they are doing, but generally creating a release identity only for a live install iso probably is drifting far enough from Fedora that I would recommend against it.

> This one is geared toward installing our standard cloud image

Not sure I am completely onboard with this, as I really don't see the use case - unless I'm missing something, *nobody* will be using our iso installer to actually install into the Azure cloud. [Looking at Fedora](https://fedoraproject.org/cloud/download/), they don't provide any iso installer for their cloud images - they provide actual images for actual end-user use cases (i.e. boot it up in Azure, or locally with Qemu). It seems like if this is intended for end-users to use our iso installer to create a local instance under Qemu or Hyper-V, we could save everyone a LOT of time and headache by simply providing Qemu and/or Hyper-V downloadable images - and, the images we provide would actually match what end users get when they deploy in Azure; using an iso installer will provide a local instance that *doesn't* exactly match our image in the Azure marketplace.

So, it feels like we're building an installer for a use case that isn't real. It would probably be better to build an iso installer geared towards bare metal install, since that's the use case where an iso installer actually makes sense (to me, at least).

base/images/vm-iso-installer/20-wired-dhcp.network
```diff
@@ -0,0 +1,8 @@
+[Match]
+Name=en* eth*
```
ddstreetmicrosoft, line 2: issue: This is not a good way to setup networking

question(blocking): if this is an offline installer, why is networking needed at all?

base/images/vm-iso-installer/grub_template.cfg
```diff
@@ -0,0 +1,22 @@
+set default=0
+set timeout=${boot_timeout}
+
+serial --speed=115200 --unit=0 --word=8 --parity=no --stop=1
+terminal_output console serial
+terminal_input console serial
+
+menuentry "Install Azure Linux 4.0" --class os {
```
ddstreetmicrosoft, line 8: question(blocking): why are you using a template instead of using kiwi's built-in grub configuration parameters?

base/images/vm-iso-installer/vm-iso-installer.kiwi
```diff
@@ -0,0 +1,141 @@
...
+        <package name="python3-kickstart" />
+        <package name="python3-blivet" />
+        <package name="python3-blockdev" />
+        <package name="python3-simpleline" />
+        <package name="python3-dasbus" />
+        <package name="python3-productmd" />
+        <package name="python3-meh" />
+        <package name="python3-pyparted" />
+
+        <!-- Anaconda runtime deps -->
```
ddstreetmicrosoft: question(blocking): why are you listing runtime deps? they will get pulled into the image build by normal dependency resolution.

base/images/vm-iso-installer/90-iso-installer.preset
```diff
@@ -0,0 +1,6 @@
+# Services needed by anaconda at ISO boot
```
ddstreetmicrosoft: issue: other than `NetworkManager` (if you are using it in the iso?), you don't need any of these; they are provided already by the `azurelinux-release-common` package (and serial-getty@ttyS0 is created by the systemd-getty-generator and you definitely don't need to manually enable it)

base/images/vm-iso-installer/config.sh
```diff
@@ -0,0 +1,289 @@
...
+    grub2-tools-extra
+    libaio
+)
+
+echo "=== Downloading target-install packages + dependencies ==="
+# The azurelinux-repos RPM provides repo ID "azurelinux" (the base repo).
+# Constrain downloads to only this repo to prevent leakage from debuginfo/source.
+dnf download \
+    --disablerepo='*' \
+    --enablerepo=azurelinux \
```
ddstreetmicrosoft: suggestion(non-blocking): you don't need both parameters, just use `--repo`

comment: performing a dnf download in the image post-processing script does not seem like the right way to do offline installation

ddstreetmicrosoft (dismissed):
I left several specific comments, and also please rebase this so you don't have later commits changing problems with earlier commits.

Additionally, I think it would be better, if this must be an offline installer, to use the same thing Fedora uses to build offline installation media, which IIUC is `pungi`. I think taking [that approach](https://fedoraproject.org/wiki/How_to_create_a_Fedora_install_ISO_for_testing) may be better than trying to get the 'live installer' to perform an offiline installation.

base/images/vm-iso-installer/vm-iso-installer.kiwi
kgodara912: These were already removed as per (https://github.com/microsoft/azurelinux/pull/16970#discussion_r3171157915)

base/images/vm-iso-installer/grub_template.cfg
kgodara912: kiwi creates only one entry for live iso listing. As we have two entries, one for live iso and other one for scripted install, we have created custom template file to give user better choices.

base/images/vm-iso-installer/20-wired-dhcp.network
kgodara912: The user may require network for live-iso testing downloading kickstart config from local network, testing other packages using dnf install before using Azure Linux, downloading few files from local internet etc. The purpose of offline installer is to not remove networking completely, which will make user testing of live iso limited, but to make installer work in offline mode as well in case user don't want to use internet. For the installed system, I will explore other ways, like using `systemd-networkd` service and Network Manager but this was simplest approach I have taken.

base/images/vm-iso-installer/config.sh
kgodara912: Updated with --repo. We need to download packages required for offline installer somewhere and verify the dependencies are met. I am fine to move the block separately which can manage the offline repo creation for anaconda installer.

base/images/vm-iso-installer/90-iso-installer.preset
kgodara912: Thanks for providing that insight, I wasn't aware that the above package provides these services, I will remove the preset.

kgodara912:
> I left several specific comments, and also please rebase this so you don't have later commits changing problems with earlier commits.
> 
> Additionally, I think it would be better, if this must be an offline installer, to use the same thing Fedora uses to build offline installation media, which IIUC is `pungi`. I think taking [that approach](https://fedoraproject.org/wiki/How_to_create_a_Fedora_install_ISO_for_testing) may be better than trying to get the 'live installer' to perform an offiline installation.



> I left several specific comments, and also please rebase this so you don't have later commits changing problems with earlier commits.
> 
> Additionally, I think it would be better, if this must be an offline installer, to use the same thing Fedora uses to build offline installation media, which IIUC is `pungi`. I think taking [that approach](https://fedoraproject.org/wiki/How_to_create_a_Fedora_install_ISO_for_testing) may be better than trying to get the 'live installer' to perform an offiline installation.

Actually, the pungi based installer is doable, I was able to generate the required config and generate iso using `pungi-koji` and dependent command lorax though haven't tested properly yet. It may give some benefits for proper treeinfo and inbuilt management of packages instead of few extra configs in kiwi-based installer.

Kiwi based installer is better approach with current azldev, it works as the other image build works and provides the required functionality needed, more explicit information and config files, though it is possible to add the support for pungi in azldev. 

Let me know if we still should go to that route instead of using current approach.

ddstreetmicrosoft (commented):
resolving my previous comments to get this closer to merging.

base/images/vm-iso-installer/config.sh
```diff
@@ -0,0 +1,267 @@
...
+
+echo "=== Downloading target-install packages + dependencies ==="
+# Kiwi removes repo configs after package installation, so repos from the .kiwi
+# file are NOT available during config.sh. Use --repofrompath with the CDN URL
+# directly. The URL matches the "azurelinux-base" repo in vm-iso-installer.kiwi.
+
+AZL_BASE_URL="https://stcontroltowerdevjwisitg.blob.core.windows.net/alpha2-prod/base/$ARCH"
+dnf5 download \
+    --setopt=reposdir=/dev/null \
+    --repofrompath=azl-base,"$AZL_BASE_URL" \
```
kgodara912, line 104: I understand that this is not a good way to do it but currently I don't have the any other way. I will search for other ways in next iteration.

reubeno (approved):

PR #16971, opened 2026-04-30
Package Nvidia Open GPU Kernel Modules

ellie-di:
###### Summary 

Adds packaging and build integration for NVIDIA’s open GPU kernel modules as kernel subpackages in Azure Linux 4.0, along with documentation of the kmod framework and workflow.

###### Test Methodology

- Pipeline build id: xxxx
CT package build Job ID: [7fe37879-b457-4ef8-cfa7-08deaca112e9](https://purple-grass-0478bae1e.7.azurestaticapps.net/workflow/7fe37879-b457-4ef8-cfa7-08deaca112e9)
Successfully ran nvidia-smi on A100 VM:
image


base/comps/kernel/kernel.comp.toml
```diff
@@ -153,3 +162,146 @@ lines = [
...
+    "# AZL: Prepare kmod subpackage sources",
+    "%global _kmod_phase prep",
+    "%global _kmod_name nvidia-open",
+    "%include %{_sourcedir}/kmod-nvidia-open.inc",
+]
+
+[[components.kernel.overlays]]
+description = "Run kmod 'build' phase at end of %build — compiles kmod modules against the just-built kernel"
+type = "spec-append-lines"
+section = "%build"
```
christopherco: **issue (blocking):** Build-phase injection lands inside `%if %{with_debuginfo}` and `%ifnarch noarch %{nobuildarches}`

The `spec-append-lines section = "%build"` overlay (description on line 273) anchors at the textual end of `%build`. In the rendered spec at this PR's head, that's `kernel.spec:3433` — inside `%if %{with_debuginfo}` (line 3428) and `%ifnarch noarch %{nobuildarches}` (line 3430). Any build with `--without debuginfo`, `--with rtonly`, or `--with automotiveonly` (all force `with_debuginfo=0`) silently skips NVIDIA module compilation while the package phase still declares the subpackage and `%files` still demands the `.ko` paths → RPM build failure.

**Suggested fix:** Replace `spec-append-lines` with `spec-search-replace` against an unconditional anchor — for example, add a stable AZL sentinel comment (`# AZL_KMOD_BUILD_HOOK`) to `kernel.spec` immediately before the `%if %{with_debuginfo}` block, then `spec-search-replace` against that sentinel. This generalizes to all phase injections and prevents recurrence on future kernel rebases.

christopherco: Tagging @reubeno since I'm not exactly sure how to do this, other than multiple overlays to include the sentinel comment with the regex replace, or direct regex replace. Both don't feel great to do...

base/comps/kernel/kernel.comp.toml
```diff
@@ -153,3 +162,146 @@ lines = [
...
+    "# AZL: Build kmod subpackage modules",
+    "%global _kmod_phase build",
+    "%global _kmod_name nvidia-open",
+    "%include %{_sourcedir}/kmod-nvidia-open.inc",
+]
+
+[[components.kernel.overlays]]
+description = "Run kmod 'install' phase at end of %install — installs kmod modules, configs, and licenses"
+type = "spec-append-lines"
+section = "%install"
```
christopherco: **issue (blocking):** Install-phase injection lands inside `%if %{with_tools}`

Same root cause as the build-phase comment above. Renders at `kernel.spec:3912`, inside `%if %{with_tools}` (line 3910). Triggered by `--with baseonly`, `--with dbgonly`, `--with rtonly`, automotive variant — all force `with_tools=0`. Modules compile but never copy into the buildroot, then `%files` references missing paths → RPM build failure.

Same fix recommendation: stable sentinel comment + `spec-search-replace`.

base/comps/kernel/kernel.comp.toml
```diff
@@ -153,3 +162,146 @@ lines = [
...
+#   1. Create kmod-<name>.inc with phase-gated sections
+#   2. Add file-add overlays for .inc and any extra source files
+#   3. Add spec-insert-tag overlays to register sources (Source6xxx range)
+#   4. Add spec-append-lines overlays for each build phase
+#   5. Register the name in kmod_subpackages in kmod-macros.inc
+# ============================================================================
+
+# --- kmod framework registration ---
+
+[[components.kernel.overlays]]
```
christopherco: **thought (non-blocking):** Consider promoting kmod registration to a first-class `comp.toml` concept before a second consumer lands

This block is ~10 overlays per kmod (file-add ×2, spec-insert-tag ×4, spec-search-replace ×1, spec-append-lines ×5). The architecture review converged on one observation: the framework is encoding a kmod-lifecycle abstraction inside generic overlay primitives, and the placement bugs are a direct consequence of the renderer not knowing it's working on a kmod.

A `[[components.kernel.kmods]]` first-class entry could own:
- source-ID allocation (deterministic, conflict-detected),
- arch + variant declarations (`arches = ["x86_64", "aarch64"]`, `variants = ["base"]`),
- anchor selection (one well-chosen, conditional-aware insertion point),
- phase-block content (replace `_kmod_phase` stringly-typed global with a sealed enum).

Net effect: per-kmod surface area collapses from ~10 overlays to one entry, source-number collisions fail loudly, and PLACEMENT-class bugs become impossible by construction. **Worth the discussion now while N=1**; the cost of doing it after 3+ consumers exist is materially higher.

If the framework angle isn't pursued, consider narrowing the doc claim (`docs/oss-kmod-packaging-strategy.md`) from "kmod framework" to "NVIDIA open kernel companion package" so future contributors don't copy the pattern uncritically.

christopherco: Tagging @reubeno as well

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,138 @@
...
+
+%post -n kmod-%{_kmod_name}
+%{_sbindir}/depmod -a %{KVERREL} || :
+
+%postun -n kmod-%{_kmod_name}
+%{_sbindir}/depmod -a %{KVERREL} || :
+
+%files -n kmod-%{_kmod_name}
+%license %{_datadir}/licenses/kmod-%{_kmod_name}/COPYING
+/lib/modules/%{KVERREL}/extra/nvidia/nvidia.ko.xz
```
christopherco: **issue (blocking):** Hard-coded `.ko.xz` breaks the automotive variant

```
130: /lib/modules/%{KVERREL}/extra/nvidia/nvidia.ko.xz
131: /lib/modules/%{KVERREL}/extra/nvidia/nvidia-modeset.ko.xz
132: /lib/modules/%{KVERREL}/extra/nvidia/nvidia-drm.ko.xz
133: /lib/modules/%{KVERREL}/extra/nvidia/nvidia-uvm.ko.xz
134: /lib/modules/%{KVERREL}/extra/nvidia/nvidia-peermem.ko.xz
```

The kernel spec compresses modules via `__modsign_install_post` using `%compext` — `xz` by default but **`zst` for the automotive variant** (`kernel.spec:496–500` overrides `%global compext zst`). On automotive, the installed files end up as `.ko.zst`, not `.ko.xz`, and `%files` fails verification.
```suggestion
/lib/modules/%{KVERREL}/extra/nvidia/nvidia.ko.%{compext}
/lib/modules/%{KVERREL}/extra/nvidia/nvidia-modeset.ko.%{compext}
/lib/modules/%{KVERREL}/extra/nvidia/nvidia-drm.ko.%{compext}
/lib/modules/%{KVERREL}/extra/nvidia/nvidia-uvm.ko.%{compext}
/lib/modules/%{KVERREL}/extra/nvidia/nvidia-peermem.ko.%{compext}
```

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,138 @@
...
+# Track the actual NVIDIA driver version via Provides for dependency resolution.
+Provides:       nvidia-open-kmod-version = %{nvidia_open_version}
+Provides:       kmod-%{_kmod_name} = %{version}-%{release}
+Provides:       kmod-%{_kmod_name}-%{KVERREL} = %{version}-%{release}
+Requires:       %{name}-core-uname-r = %{KVERREL}
+Requires(post): kmod
+Requires(postun): kmod
+Conflicts:      nvidia-closed-kmod
+Conflicts:      nvidia-open
+ExclusiveArch:  x86_64 aarch64
```
christopherco: **issue (blocking):** `ExclusiveArch` on a `%package -n` subpackage is unreliable across rpm versions

Older rpm (< 4.16) silently ignores `ExclusiveArch` on subpackages; newer rpm behavior with mock/koji can promote it to the SRPM and break ppc64le/s390x kernel rebuilds. The `.inc`'s own `%ifnarch noarch %nobuildarches` guards (lines 50, 64, 89) handle the prep/build/install phases, but the package and files phases (lines 14–43, 120–138) are not arch-guarded — so on excluded arches the subpackage is *declared* with no installed files → build failure or unreliable behavior depending on rpm version.

**Suggested fix:** Drop the `ExclusiveArch` line and instead wrap each phase block at the call sites in `kernel.comp.toml` with `%ifarch x86_64 aarch64 ... %endif` (or, equivalently, add the `%ifarch` guard around each `%if "%{_kmod_phase}" == "..."` block in this file). Match the pattern used by the existing `%ifnarch noarch %nobuildarches` guards.

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,138 @@
...
+#
+# Requires these defines/sources in kernel.spec (injected via comp.toml overlays):
+#   nvidia_open_version  — NVIDIA driver version (e.g. 595.58.03)
+#   SOURCE6000           — open-gpu-kernel-modules tarball
+#   SOURCE6001           — kmod-nvidia-open-modprobe.conf
+
+# =========================================================================
+# Phase: package — subpackage declaration
+# =========================================================================
+%if "%{_kmod_phase}" == "package"
```
christopherco: **issue (blocking):** Package and files phases declare unconditionally, but build and install are guarded by `%if %{with_up_base}`

Build (line 63) and install (line 88) are gated on `%{with_up_base}`. But `%package` (lines 14–43) and `%files` (lines 120–138) have no equivalent guard. On any build matrix where `with_up_base=0` (e.g., `--with toolsonly`, `--with rtonly`), the subpackage is declared with no files → RPM failure.

**Suggested fix:** Wrap the package and files phase bodies with `%if %{with_up_base}` (and `%ifarch x86_64 aarch64`, per the EXCLUSIVEARCH comment) so all four phases are gated consistently. Better yet: move the guards up one level to the comp.toml call sites so the `.inc` doesn't need to participate at all on excluded variants/arches.

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,138 @@
...
+# =========================================================================
+# Phase: install — install modules, config, and license
+# =========================================================================
+%if "%{_kmod_phase}" == "install"
+
+%if %{with_up_base}
+%ifnarch noarch %nobuildarches
+%{log_msg "Installing NVIDIA open GPU kernel modules for %{KVERREL}"}
+install -d %{buildroot}/lib/modules/%{KVERREL}/extra/nvidia
+for mod in nvidia nvidia-modeset nvidia-drm nvidia-uvm nvidia-peermem; do
```
christopherco, line 106: **suggestion (non-blocking):** Drop the `if [ -f ]` guard so missing modules fail loudly during install

```
92: for mod in nvidia nvidia-modeset nvidia-drm nvidia-uvm nvidia-peermem; do
93:     ko="%{_builddir}/open-gpu-kernel-modules-%{nvidia_open_version}/kernel-open/${mod}.ko"
94:     if [ -f "${ko}" ]; then
95:         install -m 0644 "${ko}" %{buildroot}/lib/modules/%{KVERREL}/extra/nvidia/
96:     fi
97: done
```

If a future NVIDIA driver bump renames or drops one of the five expected modules (e.g., GPL-incompat removal of `nvidia-peermem`), the `if [ -f ]` swallows the miss silently — the only signal is `%files` failing 20+ minutes later in the build, far from the actual cause. Since `%files` (lines 130–134) requires all five regardless, the guard provides no value, only obscures the failure point.

```suggestion
for mod in nvidia nvidia-modeset nvidia-drm nvidia-uvm nvidia-peermem; do
    ko="%{_builddir}/open-gpu-kernel-modules-%{nvidia_open_version}/kernel-open/${mod}.ko"
    install -m 0644 "${ko}" %{buildroot}/lib/modules/%{KVERREL}/extra/nvidia/
done
```

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,138 @@
...
+#   nvidia_open_version  — NVIDIA driver version (e.g. 595.58.03)
+#   SOURCE6000           — open-gpu-kernel-modules tarball
+#   SOURCE6001           — kmod-nvidia-open-modprobe.conf
+
+# =========================================================================
+# Phase: package — subpackage declaration
+# =========================================================================
+%if "%{_kmod_phase}" == "package"
+
+%package -n kmod-%{_kmod_name}
```
christopherco: **issue (non-blocking):** Without installonly marking, dnf treats a kmod for kernel-B as an upgrade of the kmod for kernel-A and removes A's files

Package name is constant (`kmod-nvidia-open`) across kernels; only the EVR carries `%{KVERREL}`. When kernel B is installed and its kmod B lands, dnf upgrades kmod-A → kmod-B and removes A's files in `/lib/modules/<KVERREL-A>/extra/nvidia/`. Booting back into kernel A then loses NVIDIA support, which violates the kernel installonly convention users expect.

Two reasonable fixes:
1. Embed the kernel uname in the package name: `kmod-nvidia-open-<KVERREL>` (RPM-Fusion / kmod-common pattern).
2. Mark `kmod-nvidia-open` as installonly in the package manager config alongside the kernel packages.

Option 1 is more standard and self-documenting; option 2 is a smaller change but requires coordinating with the dnf installonly list.

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,138 @@
...
+%ifnarch noarch %nobuildarches
+%{log_msg "Building NVIDIA open GPU kernel modules %{nvidia_open_version} for %{KVERREL}"}
+pushd %{_builddir}/open-gpu-kernel-modules-%{nvidia_open_version}
+# Unset LDFLAGS — NVIDIA kbuild invokes ld directly, not via gcc
+unset LDFLAGS
+make %{?_smp_mflags} modules \
+    KERNEL_UNAME="%{KVERREL}" \
+    SYSSRC="%{_builddir}/kernel-%{tarfile_release}/linux-%{KVERREL}" \
+    SYSOUT="%{_builddir}/kernel-%{tarfile_release}/linux-%{KVERREL}" \
+    IGNORE_CC_MISMATCH=1 \
```
christopherco, line 88: **question (non-blocking):** Is suppressing the compiler-mismatch check intentional, and is the toolchain locked across the kernel + kmod build?

NVIDIA's kbuild normally verifies that the kernel was built with the same compiler as the modules. `IGNORE_CC_MISMATCH=1` disables that check entirely. Inside a single kernel SRPM build that's safe today (same gcc throughout). The risk surfaces if AZL ever introduces a clang-built kernel option without matching the kmod build, or if a future toolchain bump lands between the kernel and a downstream kmod-only rebuild — module ABI corruption that only surfaces at runtime.

If the build environment guarantees toolchain parity, a comment on this line documenting that invariant would help future contributors. Otherwise, asserting the gcc version explicitly in the make line would catch the regression early.

base/comps/kernel/kmod-macros.inc
```diff
@@ -0,0 +1,19 @@
...
+# Each kmod .inc file uses %%if "%%{_kmod_phase}" == "<phase>" guards to
+# activate only the relevant section for that phase.
+#
+# To add a new kmod: create kmod-<name>.inc, add overlays to register its
+# sources and inject %%include lines at each phase point.
+# To remove: delete those overlays.
+#
+
+# Master list of kmod subpackages (informational — used by .inc files if needed)
+%global kmod_subpackages nvidia-open
```
christopherco: **suggestion (non-blocking):** Either wire this to drive something or remove it

```
19: %global kmod_subpackages nvidia-open
```

The doc says it's "informational — used by .inc files if needed." No `.inc` actually reads `%{kmod_subpackages}` today, and the "Add a new kmod" workflow tells contributors to update it manually with no enforcement. As-is, it's a comment masquerading as code that will silently drift on the second kmod.

If the intent is a registry, wire it to drive the framework (single source of truth). Otherwise drop it — the docs and the per-kmod overlay blocks already document which kmods are registered.

christopherco (changes requested):

base/comps/kernel/kernel.comp.toml
ellie-di: You are right. I was thinking to use the line 278 (# AZL: Build kmod subpackage modules) as an anchor for adding the next kmod build section and likewise for other stages. However, the first ones is tricky as you called out.
 @reubeno do you think this can be fixed in azldev so that the tool track the conditional macro blocks in order to make the "spec-append-lines" tag work properly?

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: Currently the kmod package includes the kernel version, e.g: kmod-nvidia-open-6.18.5-1.8.azl4.x86_64.rpm

base/comps/kernel/kernel.comp.toml
ellie-di: I found better anchors in upstream kernel spec and added a stable AZL sentinel comment.

base/comps/kernel/kernel.comp.toml
ellie-di: fixed.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: fixed.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: applied the suggestion.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: done.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: fixed.

base/comps/kernel/kmod-macros.inc
ellie-di: deleted this file.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: Had to sanitize the KVERREL to be used in the package name. Otherwise, the koji build fails with: 
error: line 17: Illegal char '~' in: %package -n kmod-nvidia-open-6.18.5-1.9.azl4\~20260504.x86_64

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: Since they share the same build root, toolchain parity is guaranteed. Added a comment.

base/comps/kernel/kernel.comp.toml
ellie-di: Will take this as a follow up action for next iteration. thanks!

base/comps/kernel/kmod-nvidia-open.inc
christopherco: @ellie-di We very recently switched to building from a specific `azl4` koji tag instead of `azl4~<date>` koji tag. Is the `azl4~<date>` still occurring in koji?

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,146 @@
...
+%global _kmod_kverrel %(echo "%{KVERREL}" | tr '~' '_')
+
+# =========================================================================
+# Phase: package — subpackage declaration
+# =========================================================================
+%if "%{_kmod_phase}" == "package" && %{with_up_base}
+
+%ifarch x86_64 aarch64
+%package -n kmod-%{_kmod_name}-%{_kmod_kverrel}
+Summary:        NVIDIA open GPU kernel modules (driver %{nvidia_open_version})
```
henryli001, line 18: Should we also provides: `nvidia-kmod = %{nvidia_open_version}`?

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,146 @@
...
+# NOTE: Version is inherited from kernel (%{version}), NOT nvidia_open_version.
+# Track the actual NVIDIA driver version via Provides for dependency resolution.
+Provides:       nvidia-open-kmod-version = %{nvidia_open_version}
+Provides:       kmod-%{_kmod_name} = %{version}-%{release}
+Provides:       kmod-%{_kmod_name}-%{KVERREL} = %{version}-%{release}
+Requires:       %{name}-core-uname-r = %{KVERREL}
+Requires(post): kmod
+Requires(postun): kmod
+Conflicts:      nvidia-closed-kmod
+Conflicts:      nvidia-open
```
henryli001: Why does it need to conflict with `nvidia-open`? I believe the upstream nvidia-open [package](https://developer.download.nvidia.com/compute/cuda/repos/azl3/sbsa/nvidia-open-595.58.03-1.azl3.noarch.rpm) has a runtime dependency on nvidia kmod

ddstreetmicrosoft (changes requested):
issue: please rebase this into discrete, logical commits as they should appear in the repo

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: My initial thinking was that one can either install the kmod package from upstream or use the azl4.0 kmod package, these two can't be installed at the same time

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: This is the way that kmod associates with its version(line 40):
_Use 'Requires: nvidia-open-kmod-version = %{nvidia_open_version}' to depend on a
specific NVIDIA driver version, or 'Requires: kmod-nvidia-open' for any version.___

henryli001: Can you drop the link for this statement? I could not find any instances of `nvidia-open-kmod-version` in existing NVIDIA spec files in gitlab unless I'm missing something here

base/comps/kernel/kmod-nvidia-open.inc
henryli001: My takeaway from the last conversation with NVIDIA is that they expect us to produce kmod going forward and their user-space components can automatically link against it

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: This is in this existing file. I'm not sure we should adopt the current implementation with what is in the github at this time.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: I mixed the package's names. Updated to `Conflicts: nvidia-kmod`

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: added.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: We discussed this offline. The kernel upgrade issue is not reproducible, so no action is required at this time.

docs/oss-kmod-packaging-strategy.md
```diff
@@ -0,0 +1,202 @@
...
+|-------|----------------|---------|
+| `package` | After `%description` | Declare `%package -n kmod-<name>-<branch>`, Provides, Requires |
+| `prep` | End of `%prep` | Extract kmod source tarball |
+| `build` | End of `%build` | Compile modules against kernel build tree |
+| `install` | End of `%install` | Install `.ko` files, configs, licenses |
+| `files` | After `%files modules-extra-matched` | `%post`/`%postun` scriptlets and file list |
+
+## Naming and Versioning Strategy
+
+Kmod subpackages include a **branch** suffix derived from the driver's major version: `kmod-<name>-<branch>`. For example, NVIDIA driver `595.58.03` with `nvidia_open_branch = "595"` produces `kmod-nvidia-open-595`.
```
henryli001: Regarding the final package name, would it make more sense to include the full version rather than the branch? For instance, something like kmod-nvidia-open-595.58.03-...

NVIDIA will support multiple versions on a driver branch and therefore it's not trivial to identify the actual driver version from the package name unless we query the virtual provides of the RPM

ellie-di: The goal is to reflect the branch name on the kmod rpm name so we can support different nvidia branches at the same time.

base/comps/kernel/kernel.comp.toml
```diff
@@ -14,10 +14,13 @@ without = [
 
 [components.kernel.build.defines]
 # RPM release number for the Azure Linux kernel package
-azl_pkgrelease = "1"
+azl_pkgrelease = "2"
 # 4th version component from the AZL kernel source (6.18.29.1). Included in specrelease so it appears
-# in the RPM Release tag, uname -r, and /lib/modules/ path (e.g. 6.18.29-1.1.azl4.aarch64).
+# in the RPM Release tag, uname -r, and /lib/modules/ path (e.g. 6.18.29-1.2.azl4.aarch64).
```
ddstreetmicrosoft: comment(non-blocking): These 2 defines are confusing, or at least the comments around them are, as to which is which.

It's not specific to this PR, but as a comment to @rlmenge can the comment here get updated with the define names instead of number values, i.e. can it get changed to `6.18.29-%%{kextraversion}.%%{azl_pkgrelease}` (assuming that is the correct order)?

base/comps/kernel/kernel.comp.toml
```diff
@@ -14,10 +14,13 @@ without = [
...
 kextraversion = "1"
+# NVIDIA open GPU kernel module version (built as a subpackage of the kernel)
+nvidia_open_version = "595.58.03"
```
ddstreetmicrosoft, line 22: comment(non-blocking): these defines won't work when/if we have multiple branch modules, e.g. `nvidia-open-580` and `nvidia-open-595`, but since we are likely going to convert this into a full fork of the upstream dist-git content (i.e. full local spec file here) I think this is fine for now.

base/comps/kernel/kernel.comp.toml
```diff
@@ -26,6 +29,13 @@ hash = "9c71dec3ea3897107c176c89014134ab31aba0c3a669af549993e1134cf2896c7d6f6b75
 hash-type = "SHA512"
 origin = { type = "download", uri = "https://github.com/microsoft/CBL-Mariner-Linux-Kernel/archive/rolling-lts/azl4/6.18.29.1.tar.gz" }
 
+# Download the NVIDIA open GPU kernel module source tarball
+[[components.kernel.source-files]]
+filename = "open-gpu-kernel-modules-595.58.03.tar.gz"
+hash = "e0c4659ddf15e4f4e19cee05b49f88c9ba08ef3add0dfe08249798f58d0fe75e"
+hash-type = "SHA256"
```
ddstreetmicrosoft: suggestion(non-blocking): this should probably be `SHA512` unless there's a good reason for it to be different from the other source hashes

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,149 @@
...
+#
+# Requires these defines/sources in kernel.spec (injected via comp.toml overlays):
+#   nvidia_open_version  — NVIDIA driver version (e.g. 595.58.03)
+#   SOURCE6000           — open-gpu-kernel-modules tarball
+#   SOURCE6001           — kmod-nvidia-open-modprobe.conf
+
+# =========================================================================
+# Phase: package — subpackage declaration
+# =========================================================================
+%if "%{_kmod_phase}" == "package" && %{with_up_base}
```
ddstreetmicrosoft, line 14: question(non-blocking): what is `%{with_up_base}`?

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,149 @@
...
+Requires(postun): kmod
+Conflicts:      nvidia-closed-kmod
+
+%description -n kmod-%{_kmod_name}-%{nvidia_open_branch}
+Open-source NVIDIA GPU kernel modules (driver version %{nvidia_open_version})
+built from the official NVIDIA/open-gpu-kernel-modules repository for
+kernel %{KVERREL}.
+
+These modules support CUDA workloads on NVIDIA GPUs (Turing and later).
+Modules: nvidia.ko, nvidia-modeset.ko, nvidia-drm.ko, nvidia-uvm.ko,
```
ddstreetmicrosoft: suggestion(non-blocking): this list of modules is duplicating the actual list of modules included in the rpm, and can easily get out of date if a module is added/removed and someone forgets to update this list. Additionally, any user can trivially get the actual list of specific modules from either the rpm or with `dnf repoquery --list`. I think it would be better to remove this manually-maintained list.

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,149 @@
...
+nvidia-peermem.ko.
+
+Each kernel version produces a separate kmod package
+(kmod-nvidia-open-<nvidia_branch>-<KVERREL>) so that multiple kernel versions
+can coexist with their own NVIDIA modules.
+Use 'Requires: nvidia-open-kmod-version = %{nvidia_open_version}' to depend on a
```
ddstreetmicrosoft, line 44: issue(non-blocking): This isn't correct; we aren't building *all* versions of any of the nvidia driver branches, we're only building the latest.

Any other package that requires the nvidia driver from the `595` branch should use `Requires: kmod-nvidia-open-595` and end users that want to install the `595` driver should `dnf install kmod-nvidia-open-595`. Users won't have the option of specifying a specific _full_ version of the nvidia drver, only the _branch_ version.

I marked this as non-blocking since it's just a rpm description which we can easily change later.

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,149 @@
...
+#   SOURCE6000           — open-gpu-kernel-modules tarball
+#   SOURCE6001           — kmod-nvidia-open-modprobe.conf
+
+# =========================================================================
+# Phase: package — subpackage declaration
+# =========================================================================
+%if "%{_kmod_phase}" == "package" && %{with_up_base}
+
+%ifarch x86_64 aarch64
+%package -n kmod-%{_kmod_name}-%{nvidia_open_branch}
```
ddstreetmicrosoft: issue(non-blocking): you're using `%{_kmod_name}` (which is set to `nvidia_open`) in order to keep the submodule name generic, but then you explicitly reference `nvidia_open` all over, e.g. `%{nvidia_open_branch}`, `%{nvidia_open_version}`, etc.

I marked this comment/issue as non-blocking because we'll need to change things during the conversion to a local spec file, so since `_kmod_name` will never be anything other than `nvidia_open` (until later) this is fine as-is.

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,149 @@
...
+
+%endif
+
+# =========================================================================
+# Phase: prep — extract NVIDIA source tarball
+# =========================================================================
+%if "%{_kmod_phase}" == "prep" && %{with_up_base}
+
+%ifarch x86_64 aarch64
+cd %{_builddir}
```
ddstreetmicrosoft: issue(non-blocking): using `pushd` and `popd` is generally the correct way to do this instead of `cd -`

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,149 @@
...
+install -d %{buildroot}/lib/modules/%{KVERREL}/extra/nvidia
+for mod in nvidia nvidia-modeset nvidia-drm nvidia-uvm nvidia-peermem; do
+    ko="%{_builddir}/open-gpu-kernel-modules-%{nvidia_open_version}/kernel-open/${mod}.ko"
+    install -m 0644 "${ko}" %{buildroot}/lib/modules/%{KVERREL}/extra/nvidia/
+done
+# Install modprobe config to blacklist conflicting modules
+install -D -m 0644 %{SOURCE6001} %{buildroot}%{_sysconfdir}/modprobe.d/kmod-%{_kmod_name}-%{nvidia_open_branch}.conf
+# Install depmod override config
+install -d %{buildroot}%{_sysconfdir}/depmod.d
+cat > %{buildroot}%{_sysconfdir}/depmod.d/kmod-%{_kmod_name}-%{nvidia_open_branch}.conf << 'DEPMOD_EOF'
```
ddstreetmicrosoft: question(non-blocking): I don't follow why any of these `depmod.d` overrides are needed? There will never be more than one e.g. `nvidia` module for any specific kernel version right?

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,149 @@
...
+%endif
+
+# =========================================================================
+# Phase: files — scriptlets and file list
+# =========================================================================
+%if "%{_kmod_phase}" == "files" && %{with_up_base}
+
+%ifarch x86_64 aarch64
+%post -n kmod-%{_kmod_name}-%{nvidia_open_branch}
+%{_sbindir}/depmod -a %{KVERREL} || :
```
ddstreetmicrosoft, line 128: suggestion(non-blocking): I don't think the kernel (or subpackages) have any file triggers on the modules dir, so it's possible the initrd should be recreated (i.e. call dracut) here. But we don't need to do that now, we can investigate it first and add it later.

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,149 @@
...
+
+%files -n kmod-%{_kmod_name}-%{nvidia_open_branch}
+%license %{_datadir}/licenses/kmod-%{_kmod_name}-%{nvidia_open_branch}/COPYING
+/lib/modules/%{KVERREL}/extra/nvidia/nvidia.ko.%{compext}
+/lib/modules/%{KVERREL}/extra/nvidia/nvidia-modeset.ko.%{compext}
+/lib/modules/%{KVERREL}/extra/nvidia/nvidia-drm.ko.%{compext}
+/lib/modules/%{KVERREL}/extra/nvidia/nvidia-uvm.ko.%{compext}
+/lib/modules/%{KVERREL}/extra/nvidia/nvidia-peermem.ko.%{compext}
+%config(noreplace) %{_sysconfdir}/modprobe.d/kmod-%{_kmod_name}-%{nvidia_open_branch}.conf
```
ddstreetmicrosoft: suggestion(non-blocking): this should go into `%{_modprobedir}` (or `%{_prefix}/lib/modprobe.d` which is equivalent), and it should not be marked as `%config(noreplace)`

base/comps/kernel/kmod-nvidia-open.inc
```diff
@@ -0,0 +1,149 @@
...
+%{_sysconfdir}/depmod.d/kmod-%{_kmod_name}-%{nvidia_open_branch}.conf
```
ddstreetmicrosoft: suggestion(non-blocking): this should go into `%{_prefix}/lib/depmod.d`

ddstreetmicrosoft (approved):
I left some comments and questions, but I don't see anything that is blocking. Any issues can be adjusted later, preferably after we change over to a local spec file (which I think will simplify this)

christopherco (changes requested):
Retarget to `4.0` branch

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: it's a macro defined in the upstream kernel that gates the build for base kernel flavor (non-debug)

base/comps/kernel/kernel.comp.toml
ellie-di: I replaced them with those macros.

base/comps/kernel/kernel.comp.toml
ellie-di: I removed the branch name from the package name

base/comps/kernel/kernel.comp.toml
ellie-di: changed it to SHA512.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: removed.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: right. removed.

base/comps/kernel/kmod-nvidia-open.inc
ellie-di: done.

docs/oss-kmod-packaging-strategy.md
```diff
@@ -0,0 +1,199 @@
...
+|-------|----------------|---------|
+| `package` | After `%description` | Declare `%package -n kmod-<name>-<branch>`, Provides, Requires |
+| `prep` | End of `%prep` | Extract kmod source tarball |
+| `build` | End of `%build` | Compile modules against kernel build tree |
+| `install` | End of `%install` | Install `.ko` files, configs, licenses |
+| `files` | After `%files modules-extra-matched` | `%post`/`%postun` scriptlets and file list |
+
+## Naming and Versioning Strategy
+
+This allows **multiple driver branches to coexist** — e.g., `kmod-nvidia-open-595` and a future `kmod-nvidia-open-600` can be installed side by side for different kernel versions or workloads.
```
rlmenge: suggestion(nonblocking): Looks like this section is stale. (Judging by `%package -n kmod-%{_kmod_name}-%{nvidia_open_version}`). The intended behavior now has the name contain the full version so the schema becomes `kmod-nvidia-open-<full-nvidia-driver-version>-<kernel-version-release>.<arch>.rpm`. 
Example: kmod-nvidia-open-595.**58.03**-6.18.31-1.4.azl4.x86_64.rpm

rlmenge (approved):

christopherco (approved):
:shipit:

PR #16974, opened 2026-04-30
fix: remove gpm dependencies from mc, vim, and emacs

reubeno:
Applies config/overlays to disable dependencies on the `gpm` package from `mc`, `vim`, and `emacs`.

`gpm` is a Linux console mouse server used by some text-mode UIs to offer mouse integration. `gpm` and its subpackages bring in dependencies on `linuxconsoletools`, which in turn depends on `SDL` components.

--
_Validated with `azldev` locally and koji build._


ddstreetmicrosoft (approved):
lgtm

PR #16978, opened 2026-04-30
feat(repo-split): drastically reduce SRPM exception list

reubeno:
This shifts our strategy toward maintaining SRPM consistency for repo
splitting: our default approach is to take an SRPM and all of its binary
packages and send them *all* to either `base` or `sdk`. We then
maintain a set of exceptions to this rule, but we want the exceptions
list to be in the 10s or 100s (not 1000s).

To accomplish this and avoid runtime dependency closure issues, we
also shift a number of packages "up" (from sdk->base) or "down"
(from base->sdk). There are also other changes to reduce dependencies
to minimize what we need to promote artificially.

Current package lists:

* `base`: 7,940 binary packages from 2,566 SRPMs; supported
* `sdk`: 24,855 binary packages from 5,014 SRPMs; not supported, only present for build-time deps
* _exceptions_: 252 binary packages across 139 SRPMs

NOTE: Without the companion overlay PRs, this PR *will* introduce some `base` repoclosure issues.


base/packages/packages.toml
```diff
@@ -6,10 +6,12 @@
...
 #   sdk.packages.toml        -> rpm-sdk    (everything else: built but unsupported)
-#   exceptions.packages.toml -> rpm-sdk    (curated carve-outs from base SRPMs)
+#   exceptions.packages.toml -> rpm-sdk    (curated carve-outs from base SRPMs;
+#                                           kept minimal and paired with entries
+#                                           in scripts/srpm-consistency-allowlist.toml)
 #
-# Policy: every binary RPM produced by an SRPM must publish to the same channel,
-# unless that subpackage is explicitly carved out in `exceptions.packages.toml`.
+# Policy: every binary RPM produced by an SRPM must publish to the same channel.
+# This invariant is enforced by `scripts/split-repo-by-channel.py`.
```
christopherco: >scripts/split-repo-by-channel.py`. [](http://example.com/codeflow?start=33&length=34)

**issue (blocking):** policy comment forward-references a file that doesn't exist on this branch

This line states the SRPM-consistency invariant is enforced by `scripts/split-repo-by-channel.py`, but that script is not present on `tomls/base/main` nor introduced by this PR. The companion comment in `exceptions.packages.toml` similarly says every entry "MUST have a corresponding exception in `scripts/srpm-consistency-allowlist.toml`" — also absent. Combined with the PR body's note that base repoclosure will be red without companion overlays, the merged branch will document an invariant the tree cannot enforce, and any reader who searches for `split-repo-by-channel.py` won't know whether it's missing-by-mistake or missing-by-in-flight.

christopherco:
> # `scripts/srpm-consistency-allowlist.toml` so the consistency check

**issue (blocking):** same forward-reference problem as `packages.toml:14`

See the comment on `packages.toml:14`. The `MUST` here is currently unenforceable because `scripts/srpm-consistency-allowlist.toml` doesn't exist on this branch — and `repoclosure-by-channel.py`'s `load_srpm_allowlist` returns `{}` when the file is missing, which silently classifies every exception entry as `uncovered` if anyone runs the consistency check against this branch.


---
Refers to: base/packages/exceptions.packages.toml:18 in 7b82a2a. [](commit_id = 7b82a2a333382a11744d638c7c15792fd89d4560, deletion_comment = False)

scripts/repoclosure-tools/repoclosure-by-channel.py
```diff
@@ -0,0 +1,1651 @@
...
+                    f"or list of strings"
+                )
+    return entries
+
+
+def _nevra_name(nevra: str) -> str:
+    base = nevra.rsplit(".", 1)[0]   # drop arch
+    base = base.rsplit("-", 1)[0]    # drop release
+    base = base.rsplit("-", 1)[0]    # drop version
+    return base
```
christopherco: >base [](http://example.com/codeflow?start=11&length=4)

**issue (blocking):** epoch prefix is not stripped, so allowlist matching silently fails for any non-zero-epoch package

dnf5's `repoclosure --json` emits NEVRAs as `epoch:name-version-release.arch` whenever epoch is non-zero. In Azure Linux that includes `bind` (32), `krb5-libs` (1), `openssl` (1), `perl` (4), `mariadb`, `tomcat`, and others. Walking the rsplit cascade for `"32:bind-9.18.28-1.x86_64"`:

1. `rsplit(".", 1)[0]` → `32:bind-9.18.28-1`
2. `rsplit("-", 1)[0]` → `32:bind-9.18.28`
3. `rsplit("-", 1)[0]` → `32:bind`

Result: `32:bind`. Then `_entry_matches` runs `fnmatch.fnmatchcase("32:bind", "bind")` → `False`. **Every allowlist entry whose `package` pattern is `bind` (or any other epoch'd name) silently fails to suppress.** A user who adds the entry, sees the suppressed count not change, and assumes "dnf5 must have already resolved it" can be very wrong.

```suggestion
def _nevra_name(nevra: str) -> str:
    # Strip optional leading "epoch:" prefix (dnf5 emits epoch when != 0).
    if ":" in nevra:
        nevra = nevra.split(":", 1)[1]
    base = nevra.rsplit(".", 1)[0]   # drop arch
    base = base.rsplit("-", 1)[0]    # drop release
    base = base.rsplit("-", 1)[0]    # drop version
    return base
```

Worth a table-driven test covering the epoch and non-epoch shapes alongside this fix.

scripts/repoclosure-tools/repoclosure-by-channel.py
```diff
@@ -0,0 +1,1651 @@
...
+    entries = list(data.get("ignore", []) or [])
+    for e in entries:
+        e.setdefault("scope", "any")
+        e.setdefault("reason", "")
+        e.setdefault("confidence", "")
+        e.setdefault("overlay", "")
+        e.setdefault("verified_at_commit", "")
+        for field in ("package", "requires"):
+            v = e.get(field, "")
+            if isinstance(v, str):
```
christopherco: **issue (blocking):** missing `package` or `requires` field becomes `[""]`, which never matches anything — the entry is silently dead

The natural idiom for an author is to write a "suppress all broken deps for this package" entry as:

```toml
[[ignore]]
package = "broken-pkg"
reason  = "known repoclosure issue, resolved in overlay"
```

`e.get("requires", "")` returns `""`, the `isinstance(v, str)` branch wraps it as `[""]`, and `_entry_matches` then runs `fnmatch.fnmatchcase("libssl.so.3()(64bit)", "")` → `False`. The same defect applies symmetrically to a missing `package`. Result: the author writes what they believe is a valid suppression, the tool reports the violation as uncovered, CI fails, and there is no signal that the allowlist entry was effectively dead.

```suggestion
for field in ("package", "requires"):
    v = e.get(field)
    if v is None:
        raise ValueError(
            f"allowlist entry missing required field '{field}': {e!r}"
        )
    if isinstance(v, str):
        e[field] = [v]
    elif isinstance(v, list):
        e[field] = list(v)
    else:
        raise ValueError(
            f"allowlist entry field '{field}' must be a string "
            f"or list of strings"
        )
    if not e[field] or any(not p for p in e[field]):
        raise ValueError(
            f"allowlist entry field '{field}' must be non-empty: {e!r}"
        )
```

Bonus follow-up (non-blocking): warn on allowlist entries that match zero findings, so stale entries surface during CI rather than rotting silently.

christopherco (changes requested):
:clock1:

reubeno:
@christopherco I'm leaning toward removing the script to expedite. Did you have any concerns with the product changes?

christopherco:
@reubeno No concerns on removing the script to expedite. It can be introduced in a future PR.
The actual product changes LGTM.

base/packages/packages.toml
reubeno: Removed all the scripts.

scripts/repoclosure-tools/repoclosure-by-channel.py
reubeno: Removed all the scripts.

scripts/repoclosure-tools/repoclosure-by-channel.py
reubeno: Removed script.

reubeno:
@christopherco Removed the script, rebased, fixed up comments. Did a quick check run against last night's x86_64 RPMs. Please re-review? I'm ready to get this merged once you're okay with it.

christopherco (approved):

PR #16997, opened 2026-05-02
Cleanup update/render CI gate flow, add AI docs.

dmcilvaney:


.github/copilot-instructions.md
```diff
@@ -42,6 +42,26 @@ azldev comp render -a          # All components (slow)
...
+
+**Always re-run `update` before opening a PR**, even after minor edits — the input fingerprint is computed from the full component config, and the `Update Locks` CI check runs against the committed state.
+
+**⚠️ Critical quirk — `rpmautospec`'s `%changelog` and `Release:` are derived from the lock as committed in `HEAD`, not the working tree.** Commit pinning works fine off the working-tree lock, so the rendered spec body tracks the new pin immediately. But the changelog entry and release number won't update until the new lock is committed. Don't panic if a freshly-bumped pin's changelog still shows the old version — commit the lock and re-render.
+
+For the full pin-bump workflow (including the commit-amend dance that keeps the `Check Rendered Specs` and `Update Locks` CI checks both green), see [`skill-update-component`](skills/skill-update-component/SKILL.md).
+
+### Prior to PR
+
+When making a PR, the lock and rendered specs must be
```
christopherco: nitpick: looks to be an errant newline. Bring the next line into this one to keep the full sentence readable.

.github/workflows/scripts/render-specs-check/check_rendered_specs.py
```diff
@@ -80,14 +80,23 @@ def _resolve_head_blobs(paths: list[str]) -> dict[str, str]:
...
-    The component is always the direct parent directory:
-    specs/a/acl/acl.spec → acl
-    /abs/path/specs/n/nano/nano.spec → nano
+    Layout under specs_dir is `<letter>/<component>/...`, so the component
+    is the second path segment relative to specs_dir. Handles nested files
+    (e.g. `specs/d/dejavu-fonts/plans/foo.fmf` → `dejavu-fonts`) correctly,
+    where the old `Path.parent.name` shortcut would have returned `plans`.
     """
-    return Path(file_path).parent.name
+    try:
```
christopherco: **issue (blocking):** the fix likely doesn't take effect in CI — the `except` branch silently returns the old buggy behavior.

`.github/workflows/check-rendered-specs.yml:118-120` invokes the script with `--specs-dir "$SPECS_DIR"`, where `SPECS_DIR=$(azldev config dump -q -f json | jq -r .project.renderedSpecsDir)`. `base/project.toml` sets `rendered-specs-dir = '../specs'`, and `azldev` resolves project paths to absolute (via `WithAbsolutePaths`) before dumping them. So `specs_dir` arrives as something like `/workdir/specs` (absolute), while `git diff --name-only` (line 115) emits repo-relative paths like `specs/d/dejavu-fonts/plans/foo.fmf`.

`Path("specs/d/dejavu-fonts/plans/foo.fmf").relative_to(Path("/workdir/specs"))` raises `ValueError`, the `except` on line 93–96 returns `Path(file_path).parent.name` = `"plans"` — the exact wrong answer the docstring on lines 86–89 promises to fix. Because this is a labeling path (not the gate logic), no CI alarm fires; the drift report just mislabels components.

**suggestion (blocking):** normalize both sides before relativizing, *and* drop the silent fallback.

```python
rel = Path(file_path).resolve().relative_to(specs_dir.resolve())
```

Either remove the `except ValueError` branch entirely (let it surface) or make it raise / log loudly. As written it re-introduces the exact bug being fixed for any future caller with a coordinate-system mismatch — and the workflow above is already that future caller.

**todo (non-blocking):** add a unit test pinning `component_from_path("specs/d/dejavu-fonts/plans/foo.fmf", Path("/workdir/specs")) == "dejavu-fonts"`. The four branches in this function deserve at least one regression test, and that test would have caught this.

.github/skills/skill-update-component/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+5. **Re-render** — now `%changelog` / `Release:` reflect the new lock.
+
+   ```bash
+   azldev comp render -p <name>
+   ```
+
+6. **Amend** so the lock bump and rendered spec land in a single commit:
+
+   ```bash
+   git add specs/<first-char>/<name>/<name>.spec
```
christopherco: **issue (blocking):** step 6 of the pin-bump loop stages only the `.spec`, but the rendered tree can include nested sidecar files — exactly the `specs/d/dejavu-fonts/plans/foo.fmf` case motivating the `component_from_path` change elsewhere in this PR. After `azldev comp render -p <name>`, those nested files may have changed, been added, or been deleted; staging only `<name>.spec` leaves them out of the amend, the PR pushes with stale sidecars, and `Check Rendered Specs` fails — even though the agent followed this skill verbatim.

**suggestion (blocking):** match the end-of-work flow at line 30 and stage the whole component directory:

```bash
git add specs/<first-char>/<name>/
git commit --amend --no-edit
```

.github/skills/skill-update-component/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Bumping an upstream commit pin
+
+When the goal is to move a component to a new upstream commit, `update` also runs **at the start** to bump the pin. There's a quirk that catches agents out: the rendered spec body and the rendered changelog/release are derived from different sources.
+
+> **Automatic changelog and release calculations are derived from the lock as committed in `HEAD`, not the working tree.** Commit pinning works fine off the working-tree lock, so the spec body tracks the new pin immediately after `update` + `render`. But the changelog entry and release number won't update until the new lock is committed. Don't panic if a freshly-bumped pin's changelog still shows the old version, iterate as needed, then commit the updated lock to see the final changelog and release reflected in the rendered spec.
```
christopherco, line 40: **thought (non-blocking):** this rpmautospec HEAD-vs-working-tree quirk now appears in three independent prose copies (`AGENTS.md:39`, `.github/copilot-instructions.md:56`, and here on line 40). Useful for discoverability — different agent runtimes consume different files — but the copies will drift if the underlying behavior ever changes. Picking a canonical home (this skill is the most detailed) and reducing the others to one-line pointers ("see `skill-update-component` for the HEAD-vs-working-tree detail") would prevent that. This PR is establishing the duplication pattern, so it's the cheapest moment to decide.

.github/workflows/check-rendered-specs.yml
```diff
@@ -317,7 +317,10 @@ jobs:
       # Fail the job when any component's lock file drifted. Uses the
       # authoritative JSON output rather than the patch file so we catch
       # drift even if patch generation itself regresses.
+      # `if: always()` so a flaky upload step above can't mask drift behind
+      # a confusing artifact-upload error.
       - name: Fail on lock drift
+        if: always()
```
christopherco: **nitpick (non-blocking):** `if: always()` also fires on workflow cancellation, which will surface "Fail on lock drift" as the failure reason on a cancelled run rather than the actual cancellation. `if: ${{ !cancelled() }}` (or `success() || failure()`) keeps the stated goal — run regardless of upload-step status — without misattributing cancellations.

.github/workflows/scripts/render-specs-check/check_rendered_specs.py
christopherco: **issue (blocking):** the fix likely doesn't take effect in CI — the `except` branch silently returns the old buggy behavior.

`.github/workflows/check-rendered-specs.yml:118-120` invokes the script with `--specs-dir "$SPECS_DIR"`, where `SPECS_DIR=$(azldev config dump -q -f json | jq -r .project.renderedSpecsDir)`. `base/project.toml` sets `rendered-specs-dir = '../specs'`, and `azldev` resolves project paths to absolute (via `WithAbsolutePaths`) before dumping them. So `specs_dir` arrives as something like `/workdir/specs` (absolute), while `git diff --name-only` (line 115) emits repo-relative paths like `specs/d/dejavu-fonts/plans/foo.fmf`.

`Path("specs/d/dejavu-fonts/plans/foo.fmf").relative_to(Path("/workdir/specs"))` raises `ValueError`, the `except` on line 93–96 returns `Path(file_path).parent.name` = `"plans"` — the exact wrong answer the docstring on lines 86–89 promises to fix. Because this is a labeling path (not the gate logic), no CI alarm fires; the drift report just mislabels components.

**suggestion (blocking):** normalize both sides before relativizing, *and* drop the silent fallback.

```python
rel = Path(file_path).resolve().relative_to(specs_dir.resolve())
```

Either remove the `except ValueError` branch entirely (let it surface) or make it raise / log loudly. As written it re-introduces the exact bug being fixed for any future caller with a coordinate-system mismatch — and the workflow above is already that future caller.

**todo (non-blocking):** add a unit test pinning `component_from_path("specs/d/dejavu-fonts/plans/foo.fmf", Path("/workdir/specs")) == "dejavu-fonts"`. The four branches in this function deserve at least one regression test, and that test would have caught this.

.github/workflows/scripts/render-specs-check/check_rendered_specs.py
```diff
@@ -80,14 +80,23 @@ def _resolve_head_blobs(paths: list[str]) -> dict[str, str]:
 # ---------------------------------------------------------------------------
 
 
-def component_from_path(file_path: str) -> str:
+def component_from_path(file_path: str, specs_dir: Path) -> str:
```
christopherco, line 84: **todo (non-blocking):** `component_from_path` now has four code paths (relative_to → ≥2 parts; → 1 part; → 0 parts; ValueError fallback) and a coordinate-system assumption that's easy to break (see the blocking comment below). Even one pytest module exercising the four cases — including the `(absolute specs_dir, relative file_path)` mismatch the workflow actually triggers — would prevent regression of the very bug this PR is fixing.

christopherco (changes requested):
2 blocking issues to look at, and rest are non-blocking but do consider

.github/workflows/scripts/render-specs-check/check_rendered_specs.py
```diff
@@ -80,14 +80,23 @@ def _resolve_head_blobs(paths: list[str]) -> dict[str, str]:
...
+    (e.g. `specs/d/dejavu-fonts/plans/foo.fmf` → `dejavu-fonts`) correctly,
+    where the old `Path.parent.name` shortcut would have returned `plans`.
+
+    Both sides are resolved to absolute paths before relativizing — the CI
+    workflow passes `specs_dir` as an absolute path (azldev's `WithAbsolutePaths`
+    resolves it during config dump) while `git diff --name-only` emits
+    repo-relative paths. A naive `relative_to` would raise on that mismatch.
     """
-    return Path(file_path).parent.name
+    rel = Path(file_path).resolve().relative_to(specs_dir.resolve())
```
christopherco: **issue (blocking):** removing the `except ValueError` was right, but `.resolve()` on `file_path` defeats the script's own symlink-refusal branch and turns it into a hard crash.

`Path.resolve()` walks symlinks. Three failure paths now exist that v1 didn't have:

1. **Symlink to outside the repo, in `changed_files`.** Line 145 detects `is_symlink()` and the warning branch on lines 146–157 builds a drift entry — but constructing that entry calls `component_from_path(path_str, specs_dir)` on line 153 *with the raw symlink path*. `Path("specs/d/foo/x.spec").resolve()` follows the link to e.g. `/etc/passwd`, `relative_to(specs_dir.resolve())` raises `ValueError`, and with no fallback the script aborts mid-report. The `::warning::Symlink in rendered-specs tree, treated as drift` message that's the whole point of the symlink branch never lands as actionable JSON — the gate fails as a Python traceback instead.

2. **Symlink to a sibling component.** A PR-committed `specs/d/foo/x.spec → ../e/evil/x.spec` resolves under `specs_dir`, `relative_to` succeeds, and `parts[1]` returns `"evil"`. The drift report blames `evil` instead of `foo`, and the auto-generated FIX command in `post_render_comment.py` (which interpolates the component name into a copy-pasteable bash block) tells reviewers to render the wrong component.

3. **Untracked symlink in `extra_files`.** `build_report` at lines 267–273 calls `component_from_path` with no `is_symlink()` filter at all — `extra_files` from `git ls-files --others` are not screened. Same crash/misattribution paths as (1) and (2), without even the warning branch as a fig leaf.

v1 had none of these problems because `Path.parent.name` was lexical and symlink-safe. The B1 fix overshot: dropping the silent `except` was correct, but adding `.resolve()` on `file_path` (rather than only `specs_dir`) undermined the symlink protection that lives elsewhere in the same file.

**suggestion (blocking):** keep `.resolve()` on `specs_dir` (it's the half that handles azldev's absolute-path output), but use a *lexical* normalization for `file_path` that doesn't follow symlinks. Either:

```python
import os
specs_abs = Path(os.path.normpath(str(specs_dir.resolve())))
file_abs = Path(os.path.normpath(os.path.abspath(file_path)))  # abspath doesn't follow links
rel = file_abs.relative_to(specs_abs)
```

…or simpler, since `specs_dir` is documented to be absolute by the time it gets here:

```python
rel = Path(os.path.normpath(file_path)).relative_to(Path(os.path.normpath(str(specs_dir))))
```

Either form preserves the v1 symlink-safety property while keeping the v2 coordinate-system fix.

christopherco (changes requested):

christopherco (approved):
:shipit:

PR #17001, opened 2026-05-03
fix(rubygem-nokogiri): drop rubygem(minitest-mock) BuildRequires

christopherco:
Upstream Fedora rubygem-nokogiri added BuildRequires: rubygem(minitest-mock)
in commit 3c9b3b0 (Dec 2025, 1.18.10-2 'Handle minitest 6 compatibility')
under '%if !0%{?rhel}', then moved it unconditionally in f64183f (Jan 2026).
The dep was introduced because Fedora 44+ ships rubygem-minitest 6.x where
Minitest::Mock was split out into its own gem.

Azure Linux ships rubygem-minitest 5.25.5, which still bundles Minitest::Mock
internally, and has no rubygem-minitest-mock package. The BR cannot resolve
regardless of which guard form upstream uses.

Add a spec-remove-tag overlay to drop the BR. Verified: build succeeds and
%check runs cleanly (2749 runs, 9837 assertions, 0 failures, 0 errors).

### DIFF
--- a/base/comps/rubygem-nokogiri/rubygem-nokogiri.comp.toml
+++ b/base/comps/rubygem-nokogiri/rubygem-nokogiri.comp.toml
@@ -1,3 +1,16 @@
 [components.rubygem-nokogiri]
 # Release: %{?prever:0.}%{baserelease}%{?prever:.%{prerpmver}}%{?dist}
 release = { calculation = "manual" }
+
+# Azure Linux ships rubygem-minitest 5.25.5, which still bundles Minitest::Mock
+# internally. Upstream Fedora split minitest-mock into a separate gem with
+# minitest 6.x (F44+); the BR was first added in Fedora rpms/rubygem-nokogiri
+# commit 3c9b3b0 (Dec 2025, 1.18.10-2 "Handle minitest 6 compatibility") under
+# `%if !0%{?rhel}`, then moved unconditionally in f64183f (Jan 2026). AZL has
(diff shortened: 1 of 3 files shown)


reubeno (approved):
I'm good with the change as is but do want to how well to remove this overlay when we move to an f44 base.

This presumably wasn't an issue in f43 because the version of the spec we're building is newer than what shipped for f43? Or reverse for the dependent change?

christopherco:
> I'm good with the change as is but do want to how well to remove this overlay when we move to an f44 base.
> 
> This presumably wasn't an issue in f43 because the version of the spec we're building is newer than what shipped for f43? Or reverse for the dependent change?

I continued to dig further and now believe this change is actually the wrong change to make. (Technically the current change will work, but there's a better one available IMO) I've converted it back to draft to avoid inadvertently merging it.

Back in Dec 2025, alongside the introduction of `rubygem-minitest-mock` commit into `rubygem-nokogiri`, a corresponding new `rubygem-minitest-mock` package was created and published in F43.
This is why F43 is actually fine with the update, since `rubygem-minitest-mock` is actually packaged and available in the Fedora 43 Updates repo (not in their Base repo at F43 launch)

I'll prep a different change to bring in `rubygem-minitest-mock`. Apologies about the thrash. This one was really bothering me, but I think I've finally got it sorted out.

christopherco:
New PR up to fix `rubgem-nokogiri` by bringing in the new `rubygem-minitest-mock` and set us up better for future Fedora rebases - #17006

PR #17002, opened 2026-05-04
fix: Remove usbguard and usbguard-notifier components

christopherco:
USB auxiliary packages are not necessary for AZL4. They can be pulled back in if a need arises. This change is proposed as a altenative to a depedency order bug causing %triggerin from selinux-policy causing a script failure when usbguard-selinux is installed.

fixes: [AB#19087](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19087)


christopherco:
Supersedes #16942

christopherco:
lock file CI check is failing for a different package unrelated to this PR. I have a fix for the CI failure here #17003

reubeno:
@christopherco Do we have data indicating no other components have build or runtime dependencies on these?

christopherco:
> @christopherco Do we have data indicating no other components have build or runtime dependencies on these?

@reubeno Yes, I've confirmed that the two other packages that have any linkage to `usbguard` and its sibling components (`selinux-policy` and `filesystem`) should continue to function without `usbguard` present at build or runtime.

reubeno (approved):

PR #17003, opened 2026-05-04
fix(pkg): pin packages to their last building commit to deal with lock file drift due to recent upstream F43 merges

christopherco:


reubeno (approved):
Happy to rubber stump update to the auto-gen file but do we know *why* there was a discrepancy? Did we change this component recently?

christopherco:
> Happy to rubber stump update to the auto-gen file but do we know _why_ there was a discrepancy? Did we change this component recently?

We don't know why, but I have already pinged @dmcilvaney on it to figure out why. I think we hold off on merging until we understand why (and I also want to see the CI drift results to see if there are more we need to inspect)

christopherco:
CI came back clean. @dmcilvaney is actively investigating why this package drifted outside of our expectation

dmcilvaney:
> CI came back clean. @dmcilvaney is actively investigating why this package drifted outside of our expectation

Tools are working as expected here. This is a known gap until we bring the more advanced change detection logic online. There was a push into the upstream f43 branch that brought with it an old commit, one that was *newer* than our previous pin, but still older than the snapshot time.

Since we blindly re-calculate the commits (fixed by the adv. change detection), it picks it up.

You can see the pushes here: https://apps.fedoraproject.org/datagrepper/v2/search?topic=org.fedoraproject.prod.pagure.git.receive&contains=rpms/fluidsynth&rows_per_page=10&order=desc

dmcilvaney (approved):

reubeno:
Note that this brings in a "mass rebuild for f44" commit, per the changelog. That doesn't seem right. Should we pin instead to the version we were using before?

christopherco:
Discussed offline - Going to move this change to a toml overlay pinned to the older commit.

PR #17006, opened 2026-05-04
feat(rubygem-minitest-mock): import new component from Fedora f43

christopherco:
Upstream Fedora extracted Minitest::Mock from the main rubygem-minitest gem in 5.27.0 (Dec 2025). The new standalone gem is required as a BuildRequires by the latest Fedora rubygem-nokogiri spec (and likely other consumers as they pick up minitest 6.x compatibility).

Import as a bare upstream component pinned via lock file to f43 HEAD (fc38ea7, 5.27.0-2, 'Initial checkin' 2025-12-24). Verified: build succeeds, RPM provides 'rubygem(minitest-mock) = 5.27.0', smoke-test in mock chroot can require 'minitest/mock' and exercise Minitest::Mock.

Fixes [AB#19691](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19691)


base/comps/components.toml
```diff
@@ -4843,6 +4843,7 @@ includes = ["**/*.comp.toml", "component-check-disablement.toml"]
 [components.rubygem-mini_mime]
 [components.rubygem-minitest]
 [components.rubygem-minitest-around]
+[components.rubygem-minitest-mock]
```
reubeno, line 4846: Needs to be in base package list? Will this conflict with another component that previously provided the sub package?

reubeno (approved):

base/comps/components.toml
christopherco: It should go in the `sdk` list since it is only needed for building `rubygem-nokogiri`. I did check and there will be no conflict with the `rubygem-minitest` component. It's a clean split out. I've pushed this update

PR #17028, opened 2026-05-05
fix(azurelinux-release): set prerelease name to Beta

christopherco:


base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -1,7 +1,7 @@
 %define release_name Four
```
reubeno, line 1: Do we need to also review the "release name"? In some cases this may cause it show up as (Four Beta).

specs/a/azurelinux-release/azurelinux-release.spec
```diff
@@ -42,7 +42,7 @@ Summary:        Azure Linux release files
 Name:           azurelinux-release
 Version:        4.0
 # TODO(azl): Review whether we can move back to autorelease (with conditional -p)
-Release: 12%{?dist}
+Release: 14%{?dist}
```
christopherco: Not sure why this is jumping by 2...

base/comps/azurelinux-release/azurelinux-release.spec
christopherco: Need to update this date

base/comps/azurelinux-release/azurelinux-release.spec
christopherco: We'll do a review of the various /etc/os-release entries next week, and then post a new PR with the result

reubeno (approved):

PR #17043, opened 2026-05-05
fix(llvm): correct clang's default GCC triple for Azure Linux stage2

rlmenge:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

The upstream Fedora clang spec writes a default config drop-in to /etc/clang/<triple>-clang.cfg containing '--gcc-triple=<cpu>-redhat-linux'. Clang autoloads this drop-in when its compiled-in default target matches the basename.

In Azure Linux 4.0 stage2, GCC's runtime files (crtbegin.o, crtend.o, libgcc.a, etc.) ship under /usr/lib/gcc/x86_64-azurelinux-linux/ instead of /usr/lib/gcc/x86_64-redhat-linux/. As a result, clang can fail to find crtbeginS.o, -lgcc, and -lgcc_s when linking with ld.lld.

Add one LLVM component overlay to change the generated cfg_file_content macro from -redhat-linux to -azurelinux-linux, so the shipped clang default config points at the actual Azure Linux GCC runtime tree.

This is the clean fix, but rollout may require a bootstrap build or pre-seeding the stage2 buildroot with the fixed clang package. A full local LLVM rebuild from the no-bootstrap branch currently fails because the buildroot's already-installed clang still has the old bad config. The branch fixes the config that the new clang RPM would install, but the build depends on the existing broken clang before it can produce that RPM.

A local full build requires a temporary bootstrap step or a buildroot where clang's GCC triple is already corrected.

Refresh locks/llvm.lock and rerender specs/l/llvm/llvm.spec. This branch intentionally omits the temporary bootstrap workarounds used for local LLVM self-rebuild testing.

Verified with azldev comp render -p llvm -O json.

###### Change Log  

- correct clang's default GCC triple for Azure Linux stage2

**NO**

###### Associated issues  

- https://dev.azure.com/mariner-org/mariner/_workitems/edit/19791/
- Fixes: [AB#19791](https://dev.azure.com/mariner-org/mariner/_workitems/edit/19791/)
###### Test Methodology

- koji: taskID=2246326

validated the change in two layers: render validation and clang runtime/link behavior in the stage2 mock chroot.

 **1.  Render validation**

Rendered LLVM successfully:

```bash
azldev comp render -p llvm -O json
```

Confirmed rendered spec contains:

```text
%global cfg_file_content --gcc-triple=%{_target_cpu}-azurelinux-linux
```

Confirmed the no-bootstrap branch does not contain the temporary bootstrap workaround:

```text
/tmp/azl-clang-bootstrap.cfg
--config=/tmp/azl-clang-bootstrap.cfg
```

**2. Runtime/link validation**

Reproduced the existing failure in the local stage2 mock buildroot. The currently installed clang package still has the bad config:

```text
/etc/clang/x86_64-azurelinux-linux-gnu-clang.cfg:
--gcc-triple=x86_64-redhat-linux
```

With that existing config, clang fails to link a simple C program because it cannot locate GCC runtime files:

```text
without override clang exit: 1
/usr/bin/ld: cannot find crtbegin.o: No such file or directory
/usr/bin/ld: cannot find -lgcc: No such file or directory
/usr/bin/ld: cannot find -lgcc_s: No such file or directory
```

Verified that forcing the Azure Linux GCC triple fixes simple C and C++ linking in the same mock chroot:

```text
with --gcc-triple=x86_64-azurelinux-linux clang exit: 0
with --gcc-triple=x86_64-azurelinux-linux clang++ exit: 0
```

Also attempted a full local LLVM build:

```bash
azldev comp build -p llvm --preserve-buildenv on-failure
```

That build fails before producing new RPMs because the buildroot uses the currently installed clang package, whose config still points at `x86_64-redhat-linux`. This is the bootstrap issue this change is intended to fix for future clang packages.

Current Limitation

A full local LLVM rebuild from the no-bootstrap branch currently fails because the buildroot's already-installed clang still has the old bad config. The branch fixes the config that the new clang RPM would install, but the build depends on the existing broken clang before it can produce that RPM.

A local full build requires a temporary bootstrap step or a buildroot where clang's GCC triple is already corrected.


base/comps/llvm/llvm.comp.toml
```diff
@@ -4,3 +4,9 @@
 # are imported into Azure Linux. MLIR is gated on %{with mlir} in the
 # upstream spec and pulls in python3-nanobind-devel which doesn't exist yet.
 build.without = ["mlir"]
+
+[[components.llvm.overlays]]
```
reubeno, line 8: We also build llvm15, 18, and 20; are they all impacted too?

base/comps/llvm/llvm.comp.toml
```diff
@@ -4,3 +4,9 @@
...
+description = "Use the Azure Linux GCC toolchain triple in clang's default config so clang can find gcc runtime files in stage2 builds"
+type = "spec-search-replace"
+regex = '%global cfg_file_content --gcc-triple=%{_target_cpu}-redhat-linux'
```
reubeno, line 11: Is there an opportunity to use the standard RPM macro `%{_target_platform}` in place of the whole thing? That's presumably what the upstream spec should have used.

base/comps/llvm/llvm.comp.toml
rlmenge: 20 for sure. Looks like 15 and 18 don't have the config file refrence so I am not sure. I will update the pr at least for llvm20.

base/comps/llvm/llvm.comp.toml
rlmenge: There is unfortunately not a match thanks to the "gnu"
```
<mock-chroot> sh-5.3# rpm --eval '%{_target_platform}'
x86_64-azurelinux-linux-gnu
```

reubeno: Good to know. How about at least using %{_vendor}?

rlmenge: %vendor should work!

rlmenge:
Closing to use the bootsrtap version: https://github.com/microsoft/azurelinux/pull/17100

PR #169, opened 2020-10-01
Patch lua CVE-2019-6706, CVE-2020-15888, nopatch CVE-2020-24342

dmcilvaney:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [ ] Ready to merge

---

###### Summary 

Nopatch `lua` CVE-2020-24342. I was unable to reproduce both issues in 5.3.5, and the Ubuntu triage reports the same results [here](https://people.canonical.com/~ubuntu-security/cve/2020/CVE-2020-24342.html). This CVEs will need to be re-visited if we update to 5.4 since it is currently not addressed in any releases.
Applied a patch for `lua` CVE-2019-6706 from [here](http://lua.2524044.n2.nabble.com/CVE-2019-6706-use-after-free-in-lua-upvaluejoin-function-tt7685575.html)
Applied patches for CVE-2020-15888, CVE-20200-15945 from Open Embedded

###### Change Log  

- Nopatch `lua` CVE-2020-24342
- Patch `lua` CVE-2019-6706 
- Patch `lua` CVE-2020-15888
- Unify the patch comments on my previous Lua CVEs.

###### Does this affect the toolchain?  

**YES**

###### Links to CVEs  
- https://nvd.nist.gov/vuln/detail/CVE-2020-15888
- https://nvd.nist.gov/vuln/detail/CVE-2020-24342
- https://nvd.nist.gov/vuln/detail/CVE-2019-6706 

###### Test Methodology

- `make input-srpms`
- validated the repo exploits for all CVEs (CVE-2020-15888 and CVE-2020-24342 was no-repro, while CVE-2019-6706 DID repro)
- local build of `lua`
- test of local build in progress
- test of new Lua package in VM **failed**. CVE-2020-15945 exploit still causes segfault. Dropping patch for CVE-2020-15945

PR #17065, opened 2026-05-06
ci(prcheck): use 'azldev component changed' for affected-component detection

dmcilvaney:
Replace the bash `git diff … | grep '/sources$'` step with structured `azldev` commands for lock validation, change detection, and scoped render verification.

## Pipeline step order

1. **Lock check** — `azldev component update --check-only`. Validates without writing; exits nonzero if any lock would change. Output JSON published as a triage artifact.
2. **Changed-component detection** — `azldev component changed --include-unchanged`. Writes the full per-component JSON to disk, published as a triage artifact. **azldev hard-fails natively** if any component has `sourcesChange == true` without a corresponding identity change (changeType not in `{added, changed, deleted}`) — supply-chain drift tripwire. Prevents unauthenticated rewrites of the rendered `sources` file under an existing component's identity.
3. **Render verification** — render the union of (azldev-flagged components) + (components whose spec tree was edited directly in the PR), in `--check-only` mode: azldev renders to a staging area and diffs against on-disk output without writing. Exits nonzero if any component's rendered output would change, catching both stale renders and direct hand-edits.
4. **Prcheck API** — switches from `--components <csv>` to `--changed-components-file <path>`, filtering to entries with `sourcesChange == true` and `changeType` in `{added, changed}` (allow-list).

PR check remains advisory until ADO task 19179 removes job-level `continueOnError`.

## Upload set filter

| changeType | sourcesChange | Included? | Why |
|---|---|---|---|
| added | true | **yes** | New component with tarballs to upload |
| changed | true | **yes** | Existing component whose tarballs changed |
| changed | false | no | Inputs changed but sources file didn't — rebuild only |
| unchanged | * | no | Not in allow-list |
| deleted | * | no | Not in allow-list |
| (unknown) | * | no | Fail closed — not in allow-list |

## New files

- `verify_locks.sh` — runs `azldev component update --check-only`, publishes the lock-update JSON for triage.
- `compute_changed.sh` — runs `azldev component changed`, publishes the changed-components JSON for triage. azldev's native consistency tripwire (#169) is what hard-fails on `sourcesChange + unchanged`.
- `render_and_verify.sh` — computes the render set (azldev-flagged ∪ specs-diff) and runs `azldev component render --check-only` against it.
- `compute_render_set.py` — computes the render set as the union of azldev-flagged components and components with hand-edited specs. Deleted and unknown components are filtered out using the full `--include-unchanged` JSON.

## Also

- Add `--changed-components-file` flag (mutually exclusive with `--components`) and `_load_components_from_file()` to `run_prcheck.py`.
- Document `AZLDEV_ALLOW_ROOT` in ADO pipeline instructions (OneBranch containers run as root, azldev refuses by default). Set inline as a per-command prefix, not at step level.
- Switch `API_BASE_URL` to `$(ApiBaseDirectUrl)` (bypasses Azure Front Door).
- Mark `changedComponentsFile` pipeline variable as `isreadonly=true`.

## Requires

- azldev versions with `component update --check-only`, `component render --check-only`, native consistency-check hard-fail in `component changed`, and `[]` (not `null`) for empty JSON output. Pinned via `AzldevCommit` in the variable group.

### DIFF
--- a/.github/instructions/ado-pipeline.instructions.md
+++ b/.github/instructions/ado-pipeline.instructions.md
@@ -175,6 +175,18 @@ Avoid shell scripts beyond the smallest possible wiring (env exports, `##vso[...
 
 Python scripts are easier to test locally, easier to review, and avoid the foot-guns of bash quoting / globbing.
 
+### `azldev` in OneBranch
+
+OneBranch runs all steps as `root`. `azldev` refuses to run many commands as root by default (a safety measure for developer workstations). To allow it in CI, set `AZLDEV_ALLOW_ROOT=1` -- but **set it inline as a prefix on each `azldev` invocation**, not at the step level. Inline scoping makes it obvious which calls actually need the override and avoids leaking the flag to unrelated commands in the same step.
+
+```bash
+AZLDEV_ALLOW_ROOT=1 azldev component update --check-only -a -q
+```
+
+Symptom that you need this: `ERR Error: this command may not be run as root`.
+
+This is NOT safe for general use, only for disposable CI environments.
+
 ## Security hardening
 
 Apply all of these unless there is a documented reason not to:
--- a/.github/workflows/ado/templates/sources-upload-stages.yml
+++ b/.github/workflows/ado/templates/sources-upload-stages.yml
@@ -117,79 +117,48 @@ stages:
             env:
               AZLDEV_COMMIT: $(AzldevCommit)
 
+          # Verify lock files are current. --check-only validates without
+          # writing, exits nonzero if any lock would change.
           - script: |
               set -euo pipefail
+              .github/workflows/scripts/control-tower/verify_locks.sh \
+                --output-file "$(Build.ArtifactStagingDirectory)/lock-update.json" \
+                --publish-dir "$(ob_outputDirectory)"
+            displayName: "Verify lock files are up to date"
 
-              # Workaround for an ADO git config error during spec rendering.
-              # The config key may not be present on every agent image, so tolerate its absence.
-              git config --unset extensions.worktreeConfig || true
+          # Detect changed components. azldev hard-fails if any component has
+          # sourcesChange == true without a corresponding "identity change"
+          # -- meaning the component's input fingerprint (.comp.toml + lock)
+          # didn't move (changeType not in {added, changed, deleted}). That
+          # combination would let attacker-supplied bytes ride into the
+          # lookaside under an unchanged component identity, so we treat it
+          # as hostile and fail closed (supply-chain drift tripwire).
+          # --include-unchanged ensures the full component list is available
+          # for downstream consumers.
+          - script: |
+              set -euo pipefail
+              json_file="$(Build.ArtifactStagingDirectory)/changed-components/changed-components.json"
+              .github/workflows/scripts/control-tower/compute_changed.sh \
+                --output-file "$json_file" \
+                --publish-dir "$(ob_outputDirectory)" \
+                --source-commit "$(sourceCommit)" \
+                --target-commit "$(targetCommit)"
+              echo "##vso[task.setvariable variable=changedComponentsFile;isreadonly=true]$json_file"
+            displayName: "Compute changed components"
 
-              # Full history is needed for spec rendering to work.
-              if [ "$(git rev-parse --is-shallow-repository)" = "true" ]; then
-                echo "##[group]Fetching full git history"
-                git fetch --unshallow
-                echo "##[endgroup]"
-              fi
-
-              specs_dir="$(azldev config dump -q -f json | jq -r '.project.renderedSpecsDir')"
-
-              echo "##[group]Specs rendering"
-              azldev component render -q -a --clean-stale
-              echo "##[endgroup]"
-
-              # Check for any new or modified files under the specs directory.
-              if ! python3 .github/workflows/scripts/render-specs-check/check_rendered_specs.py --specs-dir "$specs_dir"; then
-                echo "##[group]Git diff"
-                git --no-pager diff -- "$specs_dir"
-                echo "##[endgroup]"
-                echo "##[error]Specs are not up to date. Run 'azldev component render -q -a --clean-stale' and try again."
-                exit 1
-              fi
+          # Render the components that need re-rendering (azldev-flagged plus
+          # any with hand-edited spec files), in --check-only mode: azldev
+          # renders to a staging area and diffs against the on-disk output
+          # without writing. Exits nonzero if any component's rendered output
+          # would change, catching both stale renders and direct hand-edits.
+          - script: |
+              set -euo pipefail
+              .github/workflows/scripts/control-tower/render_and_verify.sh \
+                --output-dir "$(Build.ArtifactStagingDirectory)/render-check" \
+                --changed-components-file "$(changedComponentsFile)" \
+                --source-commit "$(sourceCommit)" \
+                --target-commit "$(targetCommit)"
             displayName: "Verify rendered specs are up to date"
-            # TODO: Re-enabled failing once specs in the main branch are updated and stable.
-            # ADO task: 19179
-            continueOnError: true
-
-          - script: |
-              set -euo pipefail
-
-              cat << EOF
-              NOTE: It's possible more components changed between the source and target commits than displayed below.
-                    This workflow only checks for updates in the 'sources' files and ignores all other changes.
-                    Only components present on the source commit are reported; components that exist only on the
-                    target branch are intentionally excluded. Pure renames are reported under their source-branch
-                    name, since the component name is part of the cache blob name we upload.
-              EOF
-
-              # Diff direction is TARGET -> SOURCE so additions/modifications represent the source-side state:
-              #   * --no-renames decomposes a rename into an add (new path, source side) + delete (old path,
-              #     target side); we want the source-side name because the component name is part of the cache
-              #     blob name uploaded for the PR. It also silences git's diff.renameLimit warning on large diffs.
-              #   * --diff-filter=d (lowercase) excludes deletions, which here correspond to paths present only
-              #     on the target branch -- those components are not on the PR and must not be uploaded.
-              # Net effect: the listed '*/sources' files are exactly those present on the source commit that
-              # differ from target, including pure renames surfaced under the source-branch name.
-              sources_files="$(
-                git diff --no-renames --diff-filter=d "$TARGET_COMMIT" "$SOURCE_COMMIT" --name-only \
-                  | { grep '/sources$' || true; } \
-                  | sort -u
-              )"
-
-              components=()
-              while IFS= read -r line; do
-                # Skip the spurious empty iteration when no '*/sources' files changed.
-                if [[ -n "$line" ]]; then
-                  components+=("$(basename "$(dirname "$line")")")
-                fi
-              done <<< "$sources_files"
-
-              components="$(IFS=, ; echo "${components[*]}")"
-              echo "Affected components: $components"
-              echo "##vso[task.setvariable variable=components]$components"
-            env:
-              SOURCE_COMMIT: $(sourceCommit)
-              TARGET_COMMIT: $(targetCommit)
-            displayName: "Determine affected components"
 
           - task: AzureCLI@2
             displayName: "Call Control Tower 'prcheck' API"
@@ -204,14 +173,14 @@ stages:
                   --api-audience "$API_AUDIENCE" \
                   --api-base-url "$API_BASE_URL" \
                   --build-reason "$BUILD_REASON" \
-                  --components "$COMPONENTS" \
+                  --changed-components-file "$CHANGED_COMPONENTS_FILE" \
                   --source-commit "$SOURCE_COMMIT" \
                   --repo-uri "$UPSTREAM_REPO_URL"
             env:
               API_AUDIENCE: $(ApiAudience)
               API_BASE_URL: $(ApiBaseDirectUrl)
               BUILD_REASON: $(Build.Reason)
-              COMPONENTS: $(components)
+              CHANGED_COMPONENTS_FILE: $(changedComponentsFile)
               SOURCE_COMMIT: $(sourceCommit)
               # TODO: Target commit is not used. Will be needed once we move detection of affected components to CT.
               # ADO task: 18816
--- a/.github/workflows/scripts/control-tower/compute_changed.sh
+++ b/.github/workflows/scripts/control-tower/compute_changed.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+# Compute the set of changed components between source and target commits.
+#
+# Writes the JSON to <output-file> and copies it into <publish-dir>/changed-components/
+# for triage artifact publication.
+#
+# 'azldev component changed' compares the source and target commits and emits
+# one entry per component with its 'changeType' ('added', 'changed',
+# 'unchanged', or 'deleted') plus a 'sourcesChange' flag indicating whether
+# the rendered 'sources' file differs between commits.
+#
+# azldev hard-fails if any component has sourcesChange == true without a
+# corresponding identity change (added/changed/deleted) -- supply-chain
+# drift protection.
+#
+# Only components with sourcesChange == true AND changeType in {added, changed}
+# are forwarded to Control Tower for upload.
+
+set -euo pipefail
+
+usage() { echo "Usage: $0 --output-file FILE --publish-dir DIR --source-commit SHA --target-commit SHA" >&2; exit 1; }
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --output-file)    output_file="$2"; shift 2 ;;
+    --publish-dir)    publish_dir="$2"; shift 2 ;;
+    --source-commit)  source_commit="$2"; shift 2 ;;
+    --target-commit)  target_commit="$2"; shift 2 ;;
+    *) usage ;;
+  esac
+done
+[[ -z "${output_file:-}" || -z "${publish_dir:-}" || -z "${source_commit:-}" || -z "${target_commit:-}" ]] && usage
+
+mkdir -p "$(dirname "$output_file")"
+
+# Publish the changed-components JSON for post-mortem triage on EVERY exit
+# path, not just success -- if azldev hard-fails on a consistency tripwire
+# the partial JSON is exactly what an operator needs to investigate.
+publish_artifact() {
+  if [[ -s "$output_file" ]]; then
+    mkdir -p "$publish_dir/changed-components"
+    cp "$output_file" "$publish_dir/changed-components/" || true
+  fi
+}
+trap publish_artifact EXIT
+
+echo "##[group]Computing changed components"
+AZLDEV_ALLOW_ROOT=1 azldev component changed --from "$target_commit" --to "$source_commit" -a --include-unchanged -O json > "$output_file"
+echo "##[endgroup]"
+
+echo "##[group]Changed components (non-unchanged only)"
+jq '[.[] | select(.changeType != "unchanged")]' "$output_file"
+echo "##[endgroup]"
+
+echo "##[group]Upload set (sourcesChange == true, changeType in {added, changed})"
+upload_count=$(jq -r '[.[] | select(.sourcesChange == true and (.changeType | IN("added", "changed")))] | length' "$output_file")
+jq -r '.[] | select(.sourcesChange == true and (.changeType | IN("added", "changed"))) | .component' "$output_file" | sort
+echo "Total: $upload_count component(s) to upload."
+echo "##[endgroup]"
--- a/.github/workflows/scripts/control-tower/compute_render_set.py
+++ b/.github/workflows/scripts/control-tower/compute_render_set.py
@@ -1,0 +1,86 @@
+"""Compute the render set: components flagged by `azldev component changed`
+plus components whose spec tree was touched directly in the PR.
+
+Emits one component name per line on stdout (azldev dedupes internally).
+"""
+
+import argparse
+import json
+from pathlib import Path
+
+
+def _load_entries(path: Path) -> list[dict]:
+    """Load the changed-components JSON."""
+    return json.loads(path.read_text(encoding="utf-8"))
+
+
+def _renderable_components(entries: list[dict]) -> set[str]:
+    """Component names that azldev can still render (everything except deleted)."""
+    return {e["component"] for e in entries if e.get("changeType") != "deleted"}
+
+
+def from_changed(entries: list[dict]) -> list[str]:
+    """Components from `azldev component changed` JSON.
+
+    Includes anything that is not 'deleted' and either has a non-'unchanged'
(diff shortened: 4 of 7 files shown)


.github/workflows/scripts/control-tower/compute_render_set.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    out = []
+    for e in entries:
+        change_type = e.get("changeType")
+        sources_change = e.get("sourcesChange") is True
+        if change_type == "deleted":
+            continue
+        if change_type == "unchanged" and not sources_change:
+            continue
+        out.append(e["component"])
```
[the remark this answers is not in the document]
dmcilvaney: simple script, will fail on change which is ok. Not worth the complexity for a very rare change.

.github/workflows/scripts/control-tower/compute_render_set.py
```diff
@@ -0,0 +1,84 @@
...
+import argparse
+import json
+from pathlib import Path
+
+
+def _load_entries(path: Path) -> list[dict]:
+    """Load the changed-components JSON, coercing null to []."""
+    return json.loads(path.read_text(encoding="utf-8")) or []
+
+
```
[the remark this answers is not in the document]
dmcilvaney: Fails safe, and format is unlikely to change

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -120,23 +120,163 @@ stages:
...
+                echo "##[group]Render set"
+                echo "$changed" | sed 's/^/  - /'
+                echo "##[endgroup]"
+                echo "##[group]Specs rendering"
+                printf '%s\n' "$changed" | xargs -d '\n' azldev component render -q --
+                echo "##[endgroup]"
+              fi
+
               # Check for any new or modified files under the specs directory.
               if ! python3 .github/workflows/scripts/render-specs-check/check_rendered_specs.py --specs-dir "$specs_dir"; then
```
dmcilvaney: Trying to decide if I should keep this, or just do a basic git diff

dmcilvaney:
ymls have gotten a bit messy, going to break them into scripts

.github/instructions/ado-pipeline.instructions.md
```diff
@@ -175,6 +175,17 @@ Avoid shell scripts beyond the smallest possible wiring (env exports, `##vso[...
...
+### `azldev` in OneBranch
+
+OneBranch runs all steps as `root`. `azldev` refuses to run many commands as root by default (a safety measure for developer workstations). To allow it in CI, set the environment variable `AZLDEV_ALLOW_ROOT=1` if azldev returns the error `ERR Error: this command may not be run as root`. This is NOT safe for general use, only for use in disposable CI environments. Set it in the `env:` block of the step, not inline in the script body.
+
+```yaml
+env:
+  # OneBranch containers run as root. azldev refuses to run as root
+  # by default, disable the root security check for this step.
+  AZLDEV_ALLOW_ROOT: "1"
+```
```
PawelWMS: I'd change this instruction to set it explicitly for the commands, which require it instead of globally for the whole step. Example: `AZLDEV_ALLOW_ROOT=1 azldev some command...`.

.github/workflows/scripts/control-tower/compute_changed.sh
```diff
@@ -0,0 +1,52 @@
...
+echo "Total: $upload_count component(s) to upload."
+echo "##[endgroup]"
+
+# Also write into ob_outputDirectory so OneBranch auto-publishes it
+# as a build artifact (explicit PublishPipelineArtifact is forbidden).
+ob_dir="$ob_output_dir/changed-components"
+mkdir -p "$ob_dir"
+cp "$json_file" "$ob_dir/"
+
+echo "##vso[task.setvariable variable=changedComponentsFile;isreadonly=true]$json_file"
```
PawelWMS: Please don't bury pipeline variables inside scripts. This requires the YAML to then magically know that these variables are being set and it's hard to track where they come from. I'd either pass the variable name through a separate argument, so the caller can set it themselves (with an optional `isoutput` for cross-job variables but that's not what would make sense here) or set the variables separately outside the script.

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -117,79 +117,67 @@ stages:
...
-                exit 1
-              fi
-            displayName: "Verify rendered specs are up to date"
-            # TODO: Re-enabled failing once specs in the main branch are updated and stable.
-            # ADO task: 19179
-            continueOnError: true
-
+              .github/workflows/scripts/control-tower/verify_locks.sh \
+                --artifact-dir "$ARTIFACT_STAGING_DIR" \
+                --ob-output-dir "$OB_OUTPUT_DIR"
```
PawelWMS: Nit-pick: the scripts don't need to know that the output directory has anything to do with OneBranch (the `ob-` prefix). I'd try to write them in such way, that they are agnostic to such details.

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -117,79 +117,67 @@ stages:
...
-              )"
-
-              components=()
-              while IFS= read -r line; do
-                # Skip the spurious empty iteration when no '*/sources' files changed.
-                if [[ -n "$line" ]]; then
-                  components+=("$(basename "$(dirname "$line")")")
-                fi
-              done <<< "$sources_files"
+          # Fail if any component's sources file changed without an identity change.
```
PawelWMS: What's an identity change? Could you extend the comment more?

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -117,79 +117,67 @@ stages:
...
+              .github/workflows/scripts/control-tower/consistency_check.sh \
+                --changed-components-file "$CHANGED_COMPONENTS_FILE"
+            # When this step fails, subsequent steps in the job are skipped -- the
+            # bad data never reaches Control Tower. However, the job-level
+            # continueOnError: true means the PR check still reports green to
+            # GitHub. The consistency tripwire becomes truly blocking once ADO
+            # task 19179 removes the job-level flag.
+            env:
+              CHANGED_COMPONENTS_FILE: $(changedComponentsFile)
+            displayName: "Source/identity consistency check"
```
PawelWMS: Can this check be run before we compute the changes? I think we don't care about the changes, if the branch is in an invalid state.

.github/workflows/scripts/control-tower/compute_changed.sh
```diff
@@ -0,0 +1,52 @@
...
+# 'deleted') plus a 'sourcesChange' flag indicating whether the rendered 'sources'
+# file differs between commits. Only components with sourcesChange == true AND
+# changeType != 'deleted' are forwarded to Control Tower for upload.
+
+echo "##[group]Computing changed components"
+azldev component changed --from "$target_commit" --to "$source_commit" -a --include-unchanged -O json > "$json_file"
+echo "##[endgroup]"
+
+echo "##[group]Changed components (non-unchanged only)"
+jq '(. // []) | [.[] | select(.changeType != "unchanged")]' "$json_file"
```
PawelWMS: This looks like we're removing unchanged components. If that's the case, what's the point of `--include-unchanged` above?

.github/workflows/scripts/control-tower/compute_changed.sh
```diff
@@ -0,0 +1,52 @@
...
+    *) usage ;;
+  esac
+done
+[[ -z "${artifact_staging_dir:-}" || -z "${ob_output_dir:-}" || -z "${source_commit:-}" || -z "${target_commit:-}" ]] && usage
+
+artifact_dir="$artifact_staging_dir/changed-components"
+json_file="$artifact_dir/changed-components.json"
+mkdir -p "$artifact_dir"
+
+# 'azldev component changed' compares the source and target commits and emits one
```
PawelWMS: Will `azldev component changed` catch changes if the branches are in an invalid state. As in: someone forgot to update the lock files or render the specs?

.github/workflows/scripts/control-tower/compute_changed.sh
dmcilvaney: Useful later in the pipeline for other filtering, it gives us a full list of every expected component.

.github/workflows/scripts/control-tower/compute_changed.sh
dmcilvaney: There is a bunch of validatoin logic in azldev, but right now its off by default. Once it turns on bad configurations will generate errors.

.github/workflows/scripts/control-tower/verify_locks.sh
```diff
@@ -0,0 +1,54 @@
...
+# Full history is needed for lock resolution and spec rendering.
+if [ "$(git rev-parse --is-shallow-repository)" = "true" ]; then
+  echo "##[group]Fetching full git history"
+  git fetch --unshallow
+  echo "##[endgroup]"
+fi
+
+mkdir -p "$(dirname "$output_file")"
+
+# Publish the lock-update JSON for post-mortem triage on EVERY exit path,
```
christopherco, line 33: **issue (blocking):** Triage artifact is not published on the failure path that needs it most.

When `azldev component update --check-only` exits nonzero (the lock-drift case this script exists to detect), control falls through `exit 1` at line 41 and the `cp "$output_file" "$publish_dir/lock-update/"` at line 47 is never reached. The drifted-locks JSON is written but not copied to the OneBranch publish directory, so the post-mortem artifact is silently absent from the very runs where it matters. Same shape on the success path — only successes publish.

Suggested fix — copy first, then exit:

```bash
mkdir -p "$publish_dir/lock-update"

if ! AZLDEV_ALLOW_ROOT=1 azldev component update --check-only -a -q -O json > "$output_file"; then
  cp "$output_file" "$publish_dir/lock-update/"   # publish before bailing
  echo "##[endgroup]"
  echo "##[error]Lock file(s) are not up to date. Run 'azldev component update -a' and commit the result."
  echo "##[group]Drifted components"
  jq -r '.[] | select(.changed == true) | .component' "$output_file" || echo "(cannot parse azldev output)"
  echo "##[endgroup]"
  exit 1
fi
echo "##[endgroup]"
cp "$output_file" "$publish_dir/lock-update/"
```

Or use a `trap '... 2>/dev/null || true' EXIT` so the copy runs regardless of how the script exits.

.github/workflows/scripts/control-tower/compute_changed.sh
```diff
@@ -0,0 +1,59 @@
...
+    --source-commit)  source_commit="$2"; shift 2 ;;
+    --target-commit)  target_commit="$2"; shift 2 ;;
+    *) usage ;;
+  esac
+done
+[[ -z "${output_file:-}" || -z "${publish_dir:-}" || -z "${source_commit:-}" || -z "${target_commit:-}" ]] && usage
+
+mkdir -p "$(dirname "$output_file")"
+
+# Publish the changed-components JSON for post-mortem triage on EVERY exit
```
christopherco, line 36: **issue (blocking):** Same triage-artifact gap as verify_locks.sh, in the script that owns the supply-chain drift tripwire.

`azldev component changed -a --include-unchanged -O json > "$output_file"` (line 37) under `set -euo pipefail`: if azldev hard-fails (the supply-chain tripwire firing on `sourcesChange == true` + `changeType == unchanged`, or any other azldev error), the script exits before the `cp` at line 52, and the `changed-components.json` artifact is never published. With job-level `continueOnError: true` and the GitHub-bridge mapping `PartiallySucceeded → ✅`, this is the worst case: the tripwire fires, the run goes green, and the investigator has no artifact pointing at what tripped.

Suggested fix — `trap` for guaranteed publication:

```bash
set -euo pipefail

# … arg parsing …
mkdir -p "$(dirname "$output_file")"
mkdir -p "$publish_dir/changed-components"

publish_artifact() {
  [[ -s "$output_file" ]] && cp "$output_file" "$publish_dir/changed-components/" 2>/dev/null || true
}
trap publish_artifact EXIT

echo "##[group]Computing changed components"
AZLDEV_ALLOW_ROOT=1 azldev component changed --from "$target_commit" --to "$source_commit" -a --include-unchanged -O json > "$output_file"
echo "##[endgroup]"
# … (rest unchanged)
```

The `trap … EXIT` runs on every exit path (success, `set -e`, `exit 1`, signal), so the artifact reaches the publish dir whenever it exists on disk.

.github/workflows/scripts/control-tower/run_prcheck.py
```diff
@@ -24,6 +33,53 @@ def _parse_components(value: str) -> list[str]:
...
+
+    The "upload set" is every component whose rendered ``sources`` file
+    changed between the two refs (``sourcesChange == true``) AND whose
+    ``changeType`` is in the allow-list ``{added, changed}``. Using an
+    allow-list (rather than just excluding ``deleted``) mirrors the
+    pipeline's consistency tripwire and ensures any future unknown
+    ``changeType`` value fails closed rather than being forwarded to
+    Control Tower.
+    """
+    ALLOWED_UPLOAD_TYPES = {"added", "changed"}
```
christopherco, line 47: **issue (blocking):** The `{added, changed}` allow-list is fail-closed by design, but combined with `continueOnError: true`, an unknown `changeType` value silently turns the supply-chain tripwire into a green PR check.

Concrete scenario: a future `azldev` introduces `changeType == "renamed"` (or `"moved"`, etc.) and emits `sourcesChange: true` for that entry. Today:

1. Both `compute_changed.sh:45` (`jq … IN("added", "changed")`) and `run_prcheck.py:47` (`ALLOWED_UPLOAD_TYPES = {"added", "changed"}`) silently filter the entry out.
2. `upload_count` becomes 0; `run_prcheck.py:190-195` prints "No affected components detected … skipping" and exits 0.
3. The job goes green, the supply-chain change is not forwarded to Control Tower, no artifact flags it.

The current behavior is "fail closed *for upload*" but "fail open *for detection*." The allow-list comment in `run_prcheck.py:42-46` claims fail-closed semantics, but the runtime behavior is silent skip.

Suggested fix — fail loudly when `sourcesChange == true` is dropped because of an unrecognized `changeType`. In `_load_components_from_file`:

```python
ALLOWED_UPLOAD_TYPES = {"added", "changed"}
KNOWN_NON_UPLOAD_TYPES = {"deleted", "unchanged"}

components: list[str] = []
unknown_dropped: list[dict] = []
for entry in entries:
    if not isinstance(entry, dict):
        continue
    change_type = entry.get("changeType")
    sources_change = entry.get("sourcesChange") is True
    if change_type in ALLOWED_UPLOAD_TYPES:
        if sources_change:
            name = entry.get("component")
            if isinstance(name, str) and name:
                components.append(name)
    elif change_type in KNOWN_NON_UPLOAD_TYPES:
        continue
    else:
        # Unknown changeType — fail closed only matters if there's an upload-relevant signal.
        if sources_change:
            unknown_dropped.append(entry)

if unknown_dropped:
    raise SystemExit(
        "##[error]Unrecognized changeType values with sourcesChange == true; "
        "azldev contract may have evolved. Update ALLOWED_UPLOAD_TYPES "
        f"explicitly. Dropped entries: {unknown_dropped!r}"
    )

return sorted(set(components))
```

Mirror the same guard in `compute_changed.sh` (display-side) so the bash log isn't inconsistent with the Python enforcement.

.github/workflows/scripts/control-tower/compute_changed.sh
```diff
@@ -0,0 +1,59 @@
...
+# path, not just success -- if azldev hard-fails on a consistency tripwire
+# the partial JSON is exactly what an operator needs to investigate.
+publish_artifact() {
+  if [[ -s "$output_file" ]]; then
+    mkdir -p "$publish_dir/changed-components"
+    cp "$output_file" "$publish_dir/changed-components/" || true
+  fi
+}
+trap publish_artifact EXIT
```
christopherco, line 45: **issue (blocking):** The `{added, changed}` allow-list is fail-closed by design, but combined with `continueOnError: true`, an unknown `changeType` value silently turns the supply-chain tripwire into a green PR check.

Concrete scenario: a future `azldev` introduces `changeType == "renamed"` (or `"moved"`, etc.) and emits `sourcesChange: true` for that entry. Today:

1. Both `compute_changed.sh:45` (`jq … IN("added", "changed")`) and `run_prcheck.py:47` (`ALLOWED_UPLOAD_TYPES = {"added", "changed"}`) silently filter the entry out.
2. `upload_count` becomes 0; `run_prcheck.py:190-195` prints "No affected components detected … skipping" and exits 0.
3. The job goes green, the supply-chain change is not forwarded to Control Tower, no artifact flags it.

The current behavior is "fail closed *for upload*" but "fail open *for detection*." The allow-list comment in `run_prcheck.py:42-46` claims fail-closed semantics, but the runtime behavior is silent skip.

Suggested fix — fail loudly when `sourcesChange == true` is dropped because of an unrecognized `changeType`. In `_load_components_from_file`:

```python
ALLOWED_UPLOAD_TYPES = {"added", "changed"}
KNOWN_NON_UPLOAD_TYPES = {"deleted", "unchanged"}

components: list[str] = []
unknown_dropped: list[dict] = []
for entry in entries:
    if not isinstance(entry, dict):
        continue
    change_type = entry.get("changeType")
    sources_change = entry.get("sourcesChange") is True
    if change_type in ALLOWED_UPLOAD_TYPES:
        if sources_change:
            name = entry.get("component")
            if isinstance(name, str) and name:
                components.append(name)
    elif change_type in KNOWN_NON_UPLOAD_TYPES:
        continue
    else:
        # Unknown changeType — fail closed only matters if there's an upload-relevant signal.
        if sources_change:
            unknown_dropped.append(entry)

if unknown_dropped:
    raise SystemExit(
        "##[error]Unrecognized changeType values with sourcesChange == true; "
        "azldev contract may have evolved. Update ALLOWED_UPLOAD_TYPES "
        f"explicitly. Dropped entries: {unknown_dropped!r}"
    )

return sorted(set(components))
```

Mirror the same guard in `compute_changed.sh` (display-side) so the bash log isn't inconsistent with the Python enforcement.

.github/workflows/scripts/control-tower/render_and_verify.sh
```diff
@@ -0,0 +1,58 @@
...
+    --target-commit)           target_commit="$2"; shift 2 ;;
+    *) usage ;;
+  esac
+done
+[[ -z "${output_dir:-}" || -z "${changed_components_file:-}" || -z "${source_commit:-}" || -z "${target_commit:-}" ]] && usage
+
+# azldev's renderedSpecsDir is absolute. Translate to repo-relative
+# so it matches git's output ('git diff --name-only' always emits
+# repo-relative paths regardless of the path arg form).
+specs_dir_abs="$(AZLDEV_ALLOW_ROOT=1 azldev config dump -q -f json | jq -r '.project.renderedSpecsDir')"
```
christopherco, line 22: **suggestion (blocking):** Guard against `azldev config dump` returning a missing `.project.renderedSpecsDir`.

`jq -r '.project.renderedSpecsDir'` returns the literal four-character string `"null"` (with `-r`) when the key is absent. `realpath --relative-to="$(pwd)" null` then resolves to a relative path called `null`, `git diff … -- null` produces an empty file list, and `from_specs_diff` finds zero hand-edits. The hand-edit detection — the entire reason this script does the `git diff` overlay — is silently bypassed with no error signal.

Suggested guard right after line 22:

```bash
specs_dir_abs="$(AZLDEV_ALLOW_ROOT=1 azldev config dump -q -f json | jq -r '.project.renderedSpecsDir')"
if [[ -z "$specs_dir_abs" || "$specs_dir_abs" == "null" ]]; then
  echo "##[error]'azldev config dump' did not return .project.renderedSpecsDir"
  exit 1
fi
specs_dir="$(realpath --relative-to="$(pwd)" "$specs_dir_abs")"
```

christopherco (changes requested):

.github/workflows/scripts/control-tower/compute_changed.sh
dmcilvaney: The pipeline is intentionally setup to not fail until work is complete. Once that time comes, stages will become properly blocking.

.github/workflows/scripts/control-tower/render_and_verify.sh
dmcilvaney: azldev cannot return an empty spec dir by design.

.github/workflows/scripts/control-tower/run_prcheck.py
dmcilvaney: as above

.github/workflows/scripts/control-tower/verify_locks.sh
dmcilvaney: trap already there.

christopherco (approved):

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -117,79 +117,48 @@ stages:
...
-              SOURCE_COMMIT: $(sourceCommit)
-              TARGET_COMMIT: $(targetCommit)
-            displayName: "Determine affected components"
+              json_file="$(Build.ArtifactStagingDirectory)/changed-components/changed-components.json"
+              .github/workflows/scripts/control-tower/compute_changed.sh \
+                --output-file "$json_file" \
+                --publish-dir "$(ob_outputDirectory)" \
+                --source-commit "$(sourceCommit)" \
+                --target-commit "$(targetCommit)"
+              echo "##vso[task.setvariable variable=changedComponentsFile;isreadonly=true]$json_file"
```
PawelWMS, line 146: I think it's going to be easier to set the `changedComponentsFile` variable next to other job-level variables at the top, so we don't need to do extra work to set it at run-time. It's also going to be easier to see where it's set + more consistent, since it's going to live with other job-level variables.

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -117,79 +117,48 @@ stages:
...
+            displayName: "Compute changed components"
+
+          # Render the components that need re-rendering (azldev-flagged plus
+          # any with hand-edited spec files), in --check-only mode: azldev
+          # renders to a staging area and diffs against the on-disk output
+          # without writing. Exits nonzero if any component's rendered output
+          # would change, catching both stale renders and direct hand-edits.
+          - script: |
+              set -euo pipefail
+              .github/workflows/scripts/control-tower/render_and_verify.sh \
```
PawelWMS, line 156: Nit-pick: could we not refer to `azldev` or its flags like `--check-only` or `--include-unchanged` earlier on the level of the pipeline, where these are not visible? On this level all we know, is that we're re-rendering to verify if anything changed and should thus fail the PR check. The method is irrelevant and may change with time. I'd leave the comments about the tooling close to where it's actually used (inside the script). Otherwise, the comments can easily become stale and cause extra mental burden on the reader. Also, if we manage to keep updating the comments to follow the implementation of the check scripts, then we unnecessarily bloat the future pull requests.

PR #17070, opened 2026-05-07
feat: Enable BVT Container tests

bhagyapathak:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

This  PR ports tests from the legacy [CBL-Mariner ContainerBase BVT](https://dev.azure.com/mariner-org/mariner/_git/CBL-Mariner-BVT?path=/pipelines/test-container-pipeline) to the current pytest-based container BVT.

Runtime tests: [base/images/tests/cases/container-base/runtime/test_container_bvt.py]
Static tests: [base/images/tests/cases/container-base/static/test_container.py]

image

image

###### Change Log  

- Enable BVT Container tests for AZL4
-

**NO**

###### Associated issues  

- #xxxx

###### Links to CVEs  
- NA


bhagyapathak:
BVT Container log file - 
[Uploading bvt-container-test-logs.txt…]()

base/images/tests/utils/pytest_plugin.py
```diff
@@ -108,6 +108,18 @@ def pytest_configure(config) -> None:  # type: ignore[no-untyped-def]
         "markers",
         "image(name): only run this test when --image-name matches",
     )
+    config.addinivalue_line(
+        "markers", 
+        "static: static filesystem tests without container runtime",
+    )
+    config.addinivalue_line(
+        "markers",
+        "static_container_test: portable static container structure tests",
```
reubeno: If we're treating the static container tests and the runtime container tests as different test suites, then I think it's simpler to keep the test cases separate too, and not need to have these annotations.

For example, we could:

* Place static test case definitions under `base/images/tests/cases/static/...`
* Place runtime test case definitions under `base/images/tests/cases/runtime/...`

And the way to indicate to pytest which tests to run could be to specify a different glob pattern / path set in the TOML file, rather than needing these annotations. These annotations make more sense to me when the selection is more dynamic based on capabilities.

base/images/tests/utils/pytest_plugin.py
```diff
@@ -142,8 +154,15 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
     caps = parse_capabilities(item.config.getoption("--capabilities", default=None))
     for marker in item.iter_markers("require_capability"):
         required = marker.args[0] if marker.args else None
-        if required and required not in caps:
-            pytest.skip(f"requires capability '{required}' (not in: {sorted(caps)})")
+        if required:
+            # Handle both single capability strings and lists of capabilities
```
reubeno: What was this required for?

base/images/tests/conftest.py
```diff
@@ -208,3 +214,114 @@ def partition_table(
...
+        
+        return exec_container_command_with_fallback(
+            running_container, command, required_packages, timeout, check
+        )
+    return _exec
+
+
+@pytest.fixture
+def ssh_exec(running_container: ContainerExecInstance):
+    """Legacy SSH exec fixture - now uses exec with smart package installation.
```
reubeno: Why legacy fixture?

base/images/tests/cases/container-base/test_container_static.py
```diff
@@ -0,0 +1,158 @@
+# SPDX-License-Identifier: MIT
+"""Ported Azure Linux Container Structure Test (CST) suite.
+
+These are static filesystem and metadata assertions ported from the Google
+Container Structure Test YAML format to pytest. They validate core container
+image properties without requiring a running container runtime.
+
+Mapped from: base/images/tests/cst/azl4_container_static_test.yaml
```
reubeno: What is this file?

base/images/tests/cases/container-base/test_container_static.py
```diff
@@ -0,0 +1,158 @@
+# SPDX-License-Identifier: MIT
+"""Ported Azure Linux Container Structure Test (CST) suite.
+
+These are static filesystem and metadata assertions ported from the Google
```
reubeno: If there's useful tests supportable in another framework, why not just support that framework too?

base/images/tests/cases/container-base/test_container_static.py
```diff
@@ -0,0 +1,158 @@
...
+
+from __future__ import annotations
+
+import re
+import pytest
+from pathlib import Path
+
+
+# ============================================================================
+# File Existence Tests (7 checks)
```
reubeno: Is this just an experimental file for demo purposes? It seems redundant with some other cases.

base/images/tests/cases/container-base/test_container_bvt.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.runtime_container_tests
+@pytest.mark.require_capability("container")
+def test_bvt_system_footprint(ssh_exec) -> None:
```
reubeno: This seems like a test that's not really asserting properties of the image-under-test but is rather performing measurements. I'm not sure that it fits into this framework.

base/images/tests/cases/container-base/test_container_bvt.py
```diff
@@ -0,0 +1,345 @@
...
+    ok, out = _cmd_ok(ssh_exec, "rpm -qa | wc -l", timeout=15)
+    if ok:
+        print(f"Installed packages: {out}")
+    else:
+        print("rpm not available — skipping package count")
+
+
+@pytest.mark.runtime_container_tests
+@pytest.mark.require_capability("container")
+def test_bvt_logging(ssh_exec) -> None:
```
reubeno: Same as above.

base/images/tests/cases/container-base/test_container_bvt.py
```diff
@@ -0,0 +1,345 @@
...
+    # Loopback ping
+    ok, out = _cmd_ok(ssh_exec, "ping -c 2 -W 3 127.0.0.1", timeout=15)
+    assert ok, f"Loopback ping failed — iputils may be missing: {out}"
+
+    # Hostname
+    ok, out = _cmd_ok(ssh_exec, "hostname")
+    assert ok and len(out) > 0, f"hostname command failed or returned empty: {out!r}"
+    print(f"Hostname: {out}")
+
+    # Routing table (informational)
```
reubeno: It seems like there's a desire for informational image collection. Does some of it really want to be a snapshot-based test?

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,461 @@
...
+        logger.info("Pre-warming dnf metadata cache")
+        exec_container_command_raw(container_name, ["dnf", "makecache"], timeout=300)
+        logger.info("dnf metadata cache warmed")
+    except subprocess.TimeoutExpired:
+        logger.warning("dnf makecache timed out — installs may be slow")
+
+    # Pre-install essential packages only if any are actually missing.
+    # Checking binaries first avoids a dnf metadata refresh when the
+    # image already ships these packages.
+    essential = {
```
reubeno: Having a global list of required packages seems like undesired coupling between general test setup and specific requirements of individual test cases.

If an individual test case requires a preinstalled package, then that seems like metadata that the test case shoujld provide and that the framework should help make happen -- but not arrange to apply to *every* run.

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,461 @@
...
+
+def ensure_package_in_container(container_name: str, package: str) -> bool:
+    """Ensure a package is installed in the container. Install if missing."""
+    return ensure_packages_in_container(container_name, [package])
+
+
+def detect_required_packages(command: str) -> list[str]:
+    """Detect what packages might be needed for a command."""
+    command_to_package = {
+        "free": "procps-ng",
```
reubeno: Same as above; having specific package lists doesn't seem right for this generic code -- and may not be appropriate for all images.

reubeno:
Did you consider allowing test cases to provide custom `Dockerfile`s for more advanced setup? I thought I saw some of the golden container tests, for example, use those.

How would distroless container setup be done?

base/images/tests/cases/container-base/test_container_static.py
bhagyapathak: I will remove redundant tests and make PR ready. This was to showcase porting static Google container tests to pytest.

base/images/tests/cases/container-base/test_container_static.py
bhagyapathak: Please ignore it, created for creating mapping with Google CST.

base/images/tests/utils/pytest_plugin.py
bhagyapathak: This code is not needed after removing annotations based test selections.

base/images/tests/conftest.py
bhagyapathak: I will remove it, anyways runtime tests are using exec cmd to execute the tests.

base/images/tests/utils/container_runtime.py
bhagyapathak: I have implemented following things
1. Container is created once per test via the function-scoped fixture
2. Per-test package dependencies are now declared as metadata via [@pytest.mark.requires_pkg(...)] in the container before the test.
3. Every test that installs packages also carries [@pytest.mark.require_capability("container", "runtime-package-management")] so on images that don't declare the capability those tests are skipped

base/images/tests/utils/container_runtime.py
bhagyapathak: Fixed this

base/images/tests/cases/container-base/test_container_static.py
bhagyapathak: It will be taken care in next PR, resolving this comment for now.

bhagyapathak:
> Did you consider allowing test cases to provide custom `Dockerfile`s for more advanced setup? I thought I saw some of the golden container tests, for example, use those.
> 
> How would distroless container setup be done?

Currently no tests require container creation using Dockerfile hence skipped the implementation in this PR. Do you want to have base code ready to take Dockerfile as input args in this PR itself?

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,447 @@
...
+# ---------------------------------------------------------------------------
+
+# First 50 digits after "3." of pi — used to verify high-precision arithmetic.
+_PI_PREFIX_50 = "14159265358979323846264338327950288419716939937510"
+
+
+@pytest.mark.require_capability("container", "runtime-package-management")
+@pytest.mark.requires_pkg("python3")
+def test_bvt_pi_to_1000_places(container_exec) -> None:
+    """BVT: High-precision pi to 1000 digits via python3 ``decimal``.
```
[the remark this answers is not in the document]
bhagyapathak: This better way of handling require_capability  so also removing code to handle multiple args to require_capability

base/images/images.toml
```diff
@@ -12,7 +12,10 @@ runtime-package-management = true
 [images.container-base]
 description = "Container Base Image"
 definition = { type = "kiwi", path = "container-base/container-base.kiwi", profile = "core" }
-tests.test-suites = [{ name = "static-image-checks" }]
+tests.test-suites = [
+    { name = "static-image-checks" },
+    { name = "container-bvt-tests" },
+]
```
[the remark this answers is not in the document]
bhagyapathak: @copilot apply changes based on [this feedback](https://github.com/microsoft/azurelinux/pull/17070#discussion_r3229244606)

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,447 @@
...
+        f"--connect-timeout 5 --max-time 10 {url} | "
+        f"grep -q '^2' && ok=$((ok+1)); i=$((i+1)); done; echo $ok"
+    )
+    ok, out = _cmd_ok(container_exec, cmd, timeout=iterations * 12)
+    assert ok, f"Sustained fetch loop failed: {out!r}"
+    success = int(out.strip().splitlines()[-1])
+    success_rate = success / iterations
+    print(f"Sustained HTTP fetch: {success}/{iterations} succeeded ({success_rate:.0%})")
+    assert success_rate >= 0.9, \
+        f"HTTP success rate {success_rate:.0%} below 90% threshold ({success}/{iterations})"
```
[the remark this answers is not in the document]
bhagyapathak: @copilot apply changes based on [this feedback](https://github.com/microsoft/azurelinux/pull/17070#discussion_r3228951177)

bhagyapathak:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/azurelinux/pull/17070#pullrequestreview-4275221016)

bhagyapathak:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/azurelinux/pull/17070#pullrequestreview-4275818337)

base/images/images.toml
```diff
@@ -61,4 +64,21 @@ extra-args = [
...
+
+[test-suites.container-bvt-tests.pytest]
+working-dir = "tests"
+install = "pyproject"
+test-paths = ["cases/container-base/runtime/test_container_bvt.py"]
+extra-args = [
+    "--image-path", "{image-path}",
+    "--image-name", "{image-name}",
+    "--capabilities", "{capabilities}",
+    "-v", "-s"
```
reubeno, line 83: I don't think it makes sense to hard-code `-v` or `-s` in the *test case*, as that's a generic pytest option that the adapter for running pytest-based tests should be able to handle. `azldev` already does something like that, where it translates its own verbosity options to `--log-cli-level`. We could refine that, but it doesn't seem like should be test suite specific.

base/images/images.toml
```diff
@@ -61,4 +64,21 @@ extra-args = [
...
+
+# Container Business Validation Tests (BVT)
+[test-suites.container-bvt-tests]
+type = "pytest"
+description = "Container BVT: runtime validation of system resources, networking, filesystem, and package management"
+
+[test-suites.container-bvt-tests.pytest]
+working-dir = "tests"
+install = "pyproject"
+test-paths = ["cases/container-base/runtime/test_container_bvt.py"]
```
reubeno, line 78: Why not just specify the whole `runtime/` dir? But also, it's odd that the naming of the test suite implies it should be runnable against any container, but it's got a hard-coded reference to `container-base` in particular.

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,475 @@
...
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.require_capability("container")
+@pytest.mark.require_capability("runtime-package-management")
+@pytest.mark.requires_pkg("procps-ng", "gawk")
+def test_bvt_system_footprint(container_exec) -> None:
+    """BVT: Memory, disk, CPU, and process footprint metrics."""
```
reubeno, line 38: I still don't understand what it means to have a test that doesn't really assert anything meaningful (beyond the ability to run the commands below). This seems more like a metrics-gathering piece of code than a validation test case--do we need a different way to capture metrics?

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,475 @@
...
+@pytest.mark.require_capability("container")
+@pytest.mark.require_capability("runtime-package-management")
+@pytest.mark.requires_pkg("util-linux")
+def test_bvt_logging(container_exec) -> None:
+    """BVT: System logging via logger and log file/journal access."""
+    tag = f"bvt_{int(time.time())}"
+    message = f"BVT log test entry {tag}"
+
+    ok, _ = _cmd_ok(container_exec, f"logger -t bvt_test '{message}'", timeout=10)
+    assert ok, "logger command failed — util-linux may be missing"
```
reubeno, line 83: Doesn't seem to make sense to say that `util-linux` may be missing if we're already requiring it via the decorator?

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,475 @@
...
+
+    # Find the entry: try journalctl, then /var/log/messages, then /var/log/syslog
+    found = False
+    ok, out = _cmd_ok(container_exec, f"journalctl --no-pager --since '1 minute ago' 2>/dev/null | grep '{tag}' || true", timeout=15)
```
reubeno, line 87: Is `journalctl` even in the base container? We don't run in systemd in it. Was this meant to be a *VM* test?

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,475 @@
...
+
+    # Log entry visibility depends on journald/syslog being wired up in the container.
+    if found:
+        print(f"Log entry verified in system logs: {message}")
+    else:
+        print(f"WARNING: log entry not immediately visible (journald may not be running in container): {message}")
+
+
+@pytest.mark.require_capability("container")
+def test_bvt_mathematical_computing(container_exec) -> None:
```
reubeno, line 106: This test is a bit odd; it seems pecular to couple `bash` testing with `python` testing, and then ignore inability to run `python`.

I *do* see the value in making sure that python works correctly, but that shouldn't be optional -- and should require that python get installed.

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,475 @@
...
+    assert ok, "ip command failed — iproute2 may be missing"
+    inet_lines = [l.strip() for l in out.splitlines() if "inet " in l and "scope" in l]
+    assert len(inet_lines) >= 1, f"No inet addresses found:\n{out}"
+    print(f"Network interfaces ({len(inet_lines)} addresses): {inet_lines}")
+
+    # Loopback presence (the inet 127.0.0.1 line above already proves loopback
+    # exists). ping is informational only: rootless podman with slirp4netns
+    # often blocks ICMP regardless of iputils being installed.
+    assert any("127.0.0.1" in l for l in inet_lines), "Loopback interface missing"
+    ok, out = _cmd_ok(container_exec, "ping -c 2 -W 3 127.0.0.1", timeout=15)
```
reubeno, line 151: I understand the possible challenge with ICMP, but I don't see value in a test that can never fail -- in both success and failure we print a message and move on. Do we expect that a human will be reviewing the output?

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,475 @@
...
+    assert ok, f"hostname command failed: {out!r}"
+    print(f"Hostname: {out or '(empty - no --hostname set)'}")
+
+    # Routing table (informational)
+    ok, out = _cmd_ok(container_exec, "ip route show", timeout=10)
+    if ok:
+        print(f"Routing table present, default route: {'default' in out}")
+
+    # DNS config (informational)
+    ok, out = _cmd_ok(container_exec, "cat /etc/resolv.conf")
```
reubeno, line 168: Is there not something we should be asserting/validating here?

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,475 @@
...
+        )
+        print(f"DNS resolved {resolved}/{len(domains)} domains")
+        assert resolved >= 1, f"All DNS resolutions failed for {domains}"
+    else:
+        print("nslookup not available — skipping DNS resolution test")
+
+    # HTTP connectivity (optional)
+    ok_curl, _ = _cmd_ok(container_exec, "which curl", timeout=5)
+    if ok_curl:
+        ok, out = _cmd_ok(container_exec, "curl -s --connect-timeout 5 --max-time 10 http://httpbin.org/get", timeout=15)
```
reubeno, line 193: Is there not a standard Microsoft property that we could be hitting instead?

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,475 @@
...
+@pytest.mark.require_capability("runtime-package-management")
+def test_bvt_package_management(container_exec) -> None:
+    """BVT: Package manager (tdnf/dnf) cache refresh, list, and info."""
+    ok_tdnf, _ = _cmd_ok(container_exec, "which tdnf", timeout=5)
+    pm = "tdnf" if ok_tdnf else "dnf"
+
+    ok, out = _cmd_ok(container_exec, f"{pm} makecache", timeout=60)
+    assert ok, f"{pm} makecache failed: {out}"
+
+    ok, out = _cmd_ok(container_exec, f"{pm} list --installed | head -20", timeout=30)
```
reubeno, line 281: Why `| head -20`?

base/images/tests/cases/container-base/runtime/test_container_bvt.py
```diff
@@ -0,0 +1,475 @@
...
+
+
+@pytest.mark.require_capability("container")
+@pytest.mark.require_capability("runtime-package-management")
+def test_bvt_package_management(container_exec) -> None:
+    """BVT: Package manager (tdnf/dnf) cache refresh, list, and info."""
+    ok_tdnf, _ = _cmd_ok(container_exec, "which tdnf", timeout=5)
+    pm = "tdnf" if ok_tdnf else "dnf"
+
+    ok, out = _cmd_ok(container_exec, f"{pm} makecache", timeout=60)
```
reubeno, line 278: The commands in this test solely validate that they get back a 0 exit code and don't validate any of the output. Is that intended?

base/images/tests/cases/container-base/static/test_container.py
```diff
@@ -0,0 +1,135 @@
...
+import os
+import re
+from pathlib import Path
+
+import pytest
+
+# Default upper-bound for the unpacked container rootfs (MiB).
+# Ported from the legacy "Memory Usage Check" / "Container Size Max Check"
+# in the CBL-Mariner ContainerBase BVT. Override via env var
+# AZL_CONTAINER_MAX_SIZE_MB to adjust the gate per build flavor.
```
reubeno, line 14: Where would you expect to store the expected size values?

base/images/tests/cases/container-base/static/test_container.py
```diff
@@ -0,0 +1,135 @@
...
+        rootfs / "usr" / "lib" / "modules",
+    ):
+        if modules_dir.exists():
+            versions = list(modules_dir.iterdir())
+            assert not versions, (
+                f"Container image has kernel modules under {modules_dir}: "
+                f"{[v.name for v in versions]}"
+            )
+
+def test_file_exists_root(rootfs: Path) -> None:
```
reubeno, line 32: Perhaps the `no_kernel_modules` test wasn't a good example... for all the tests here that are more broadly applicable, they belong outside the `container-base` dir, and not specific to this particular container image.

base/images/tests/cases/container-base/static/test_container.py
```diff
@@ -0,0 +1,135 @@
...
+    assert mode & 0o100, f"File {path} is not executable by owner"
+
+
+def test_file_exists_bin_bash(rootfs: Path) -> None:
+    """Bash must exist."""
+    path = rootfs / "bin" / "bash"
+    assert path.exists(), f"File {path} does not exist"
+
+
+def test_file_exists_etc_dnf_dnf_conf(rootfs: Path) -> None:
```
reubeno, line 52: This is a good example of a test that could apply to *all* images that support runtime package management, and not even just containers. I don't think it belongs under `container-base`.

base/images/tests/cases/container-base/static/test_container.py
```diff
@@ -0,0 +1,135 @@
...
+        f"Root entry not found in passwd file matching pattern {pattern}"
+
+
+def test_license_bash_exists(rootfs: Path) -> None:
+    """Bash license file must exist."""
+    path = rootfs / "usr" / "share" / "licenses" / "bash" / "COPYING"
+    assert path.exists(), f"License file {path} does not exist"
+
+
+def test_license_coreutils_single_exists(rootfs: Path) -> None:
```
reubeno, line 93: Is there a reason that `bash` and `coreutils-single` were called out in particular from a license perspective? Or are they spot checks?

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,315 @@
...
+            result.stderr,
+        )
+        raise ContainerRuntimeError(
+            f"Container command failed: {' '.join(cmd)}\n"
+            f"Error: {result.stderr}"
+        )
+    return result
+
+
+def _parse_loaded_images(load_stdout: str) -> list[str]:
```
reubeno, line 66: Please don't text-scrape command output -- there must be a better programmatic way to get the data we need.

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,315 @@
+# SPDX-License-Identifier: MIT
+"""Container runtime orchestration for dynamic testing.
+
+Provides fixtures and utilities for creating running containers with exec access
```
reubeno, line 4: Is there a reason we're not using a higher-level Python library for automating podman? At least something like https://pypi.org/project/podman/? Worthwhile to spend a few mins looking to see if there's something even higher-level that's geared toward testing?

base/images/tests/conftest.py
```diff
@@ -208,3 +214,160 @@ def partition_table(
...
+    pkgs: list[str] = []
+    for m in markers:
+        pkgs.extend(m.args)
+    if not pkgs:
+        return
+
+    container_exec = request.getfixturevalue("container_exec")
+
+    # Detect which package manager is available in this container.
+    # AZL images may ship tdnf, dnf, or both; prefer tdnf when present.
```
reubeno, line 329: No reason to use `tdnf` if these tests are only intended for 4.0+.

reubeno (commented):
Thanks for the updates; I've left comments and questions throughout the code. Happy to meet synchronously to go over it.

I think my biggest questions are around the structure.

In my earlier feedback I was thinking something along these lines:

```
base/images/tests/
    cases/
        static/
            test_*.py  # shared static tests useful across images; majority of static tests
            <image_name>/
                test_*.py # tests specific to the named image; should be quite minimal
        runtime/
            test_*.py # shared runtime tests useful across images; majority of runtime tests
            <image_name>/
                test_*.py # tests specific to the named image; should be quite minimal
```

This would allow the static test suite definition to just point to `static/` and the runtime one to point to `runtime/`.

Would be good to get your input here -- and what structure you were going for.

christopherco (changes requested):
Retarget to 4.0 branch

bhagyapathak:
Closing this PR in favor of https://github.com/microsoft/azurelinux/pull/17372

PR #17074, opened 2026-05-07
fix(sos): add dnf5 support in dnf plugin

eric-desrochers:
Handling DNF5 changes where legacy dnf commands produce Missing command or Unknown argument errors.


ddstreet:
this needs a rebase due to the 'lock' file being updated

ddstreetmicrosoft (approved):
lgtm

PR #17076, opened 2026-05-07
feat(azurelinux-repos): update azurelinux-repos for beta

reubeno:
Update azurelinux-repos package to point to PMC beta repos.

Add azurelinux-repos-dev sub-package that points to daily dev repo instead. For now, we'll include the latter package in our image definitions. Once we're ready to cut over, we need only update the image definitions.


ddstreetmicrosoft (approved):
The 2 main points of feedback I have are:
1) disagreement about the specific repos, 'main' and 'dev', and how they conflict; though i am ok with the naming and conflicting rpms in the short term
2) there are differences with how upstream manages the subpackage creation (e.g. repo.in file var substitutions) which technically aren't a problem; but if we do agree to stick with a repo setup closer to upstream i think it would be better to stay closer to their spec file code to create the subpackage content

Both are longer term concerns and I have no concerns about this in the short term (assuming it's been tested and works of course, which I haven't checked). I do think that longer term we should work out exactly what repos we want to have and what the specific process is for updates to flow through the repos into the final ('prod') repo.

PR #17077, opened 2026-05-07
openssl: remove fips.so on AZL4 (not supported for public preview)

tobiasb-ms:
## Summary

AZL4 does not support FIPS mode for public preview. This change removes `fips.so` (and `fipsmodule.cnf`) from the `openssl` package on AZL4 by extending the existing RHEL fips.so-deletion path.

The `Requires: openssl-fips-provider` dependency is also suppressed on AZL4, as that package is not being shipped.

## Changes

- New `base/comps/openssl/openssl.comp.toml` with two overlays:
  1. Extend the RHEL `__spec_install_post` fips.so-deletion conditional to cover AZL4
  2. Suppress `Requires: openssl-fips-provider` on AZL4
- Updated rendered spec in `specs/o/openssl/`

## Testing

Built and tested on the build server and verified locally:
- `fips.so` absent from `openssl-libs`
- `fipsmodule.cnf` absent
- `openssl` and `openssl-libs` install and function correctly

Fixes [AB#17936](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/17936)


base/comps/openssl/openssl.comp.toml
```diff
@@ -0,0 +1,18 @@
...
+description = "Remove fips.so from openssl on AZL4 (not supported for public preview)"
+
+# The above replacement also hits the Requires: openssl-fips-provider guard.
+# Suppress that dependency on AZL4 — openssl-fips-provider is not shipped for
+# public preview. The line is unique so this is safe to match unconditionally.
+[[components.openssl.overlays]]
+type = "spec-search-replace"
+regex = "Requires: openssl-fips-provider"
+replacement = "%if ! 0%{?azl4}\nRequires: openssl-fips-provider\n%endif"
+description = "Don't require openssl-fips-provider on AZL4"
```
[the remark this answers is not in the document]
tobiasb-ms: I'd rather simply mirror RHEL for this change.

specs/o/openssl/openssl.spec
```diff
@@ -98,7 +98,7 @@ Patch0049: 0049-FIPS-fix-disallowed-digests-tests.patch
 Patch0050: 0050-Make-openssl-speed-run-in-FIPS-mode.patch
 Patch0051: 0051-Backport-upstream-27483-for-PKCS11-needs.patch
 Patch0052: 0052-Red-Hat-9-FIPS-indicator-defines.patch
-%if ( %{defined rhel} && (! %{defined centos}) && (! %{defined eln}) )
+%if ( %{defined rhel} && (! %{defined centos}) && (! %{defined eln}) ) || 0%{?azl4}
 Patch0053: 0053-Allow-hybrid-MLKEM-in-FIPS-mode.patch
 %endif
```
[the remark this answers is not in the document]
tobiasb-ms: I'd rather simply mirror RHEL for this change.

base/comps/openssl/openssl.comp.toml
```diff
@@ -0,0 +1,18 @@
+[components.openssl]
+
+# AZL4 does not ship fips.so. Extend the existing RHEL fips.so-deletion
+# path to cover AZL4 so the module and its config are excluded from the package.
+[[components.openssl.overlays]]
+type = "spec-search-replace"
+regex = '%if \( %\{defined rhel\} && \(! %\{defined centos\}\) && \(! %\{defined eln\}\) \)'
+replacement = '%if ( %{defined rhel} && (! %{defined centos}) && (! %{defined eln}) ) || 0%{?azl4}'
```
reubeno, line 8: For cases like this we should probably define a proper `azurelinux` macro, as some of the other distros have done (and have it initialized to `4`). Not for this PR, but could you file a bug on that?

base/comps/openssl/openssl.comp.toml
```diff
@@ -0,0 +1,18 @@
...
+description = "Remove fips.so from openssl on AZL4 (not supported for public preview)"
+
+# The above replacement also hits the Requires: openssl-fips-provider guard.
+# Suppress that dependency on AZL4 — openssl-fips-provider is not shipped for
+# public preview. The line is unique so this is safe to match unconditionally.
```
reubeno: Would it be simpler to just `spec-remove-tag` on the `Requires: openssl-fips-provider`?

base/comps/openssl/openssl.comp.toml
tobiasb-ms: https://dev.azure.com/mariner-org/mariner/_workitems/edit/19815

reubeno (approved):

base/comps/openssl/openssl.comp.toml
tobiasb-ms: Good call — switched to `spec-remove-tag` in the latest push. Since the rendered spec is AZL4-only, removing the tag unconditionally is correct. Built and verified locally: `openssl-libs` has no `Requires: openssl-fips-provider` and no `fips.so` in the file list.

christopherco (approved):

PR #17079, opened 2026-05-07
fix: remove koji-tool and unused dependencies from distro

trungams:
fixes: [AB#19055](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19055)

koji-tool has a broken ghc dependency and is not required by any other package in the distro. Remove koji-tool and some ghc-* components until we have a clear use case and can onboard it properly.


christopherco (commented):
There is an orphan reference at sdk.packages.toml line 47 to `HaXml` which is from `ghc-HaXml` SRPM. Rest LGTM so I'll go in and remove this

christopherco (approved):

PR #17083, opened 2026-05-07
fix(openfec): disable -debuginfo to work around malware scan's false positive

PawelWMS:
Fixes: [AB#19876](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19876)

## Summary

Suppresses the auto-generated `*-debuginfo` sub-package for `openfec` to work around a malware-scan false positive.

## Failing artefacts

- `openfec-debuginfo-1.4.2.6-7.azl4~20260420.x86_64.rpm`
- `openfec-debuginfo-1.4.2.6-7.azl4~20260420.aarch64.rpm`

The runtime `openfec`, `openfec-devel`, and `openfec-utils` sub-packages are NOT flagged. The SRPM is NOT flagged. **Only the auto-generated debuginfo RPMs fail signing.**

## Root cause

A high-entropy / packer detection heuristic in the signing pipeline's malware scanner fires on the per-library `.debug` ELF (`/usr/lib/debug/usr/lib64/libopenfec.so.1.4.2-…<arch>.debug`) produced by rpmbuild's standard `find-debuginfo.sh` + `dwz` pipeline. This is a textbook AV/EDR false positive on stripped DWARF debug-info: `.debug_info` / `.debug_str` / `.debug_loc` / `.debug_abbrev` sections plus the densely-packed string tables look statistically identical to packed/encrypted payloads, especially with rpm's default `--compress-debug-sections=zlib`. The same heuristic fires on **both** architectures from the same `%build`, which is itself strong evidence that the trigger is the standard rpmbuild debuginfo pipeline rather than anything in `Source0`.

Source0 (`openfec_1.4.2.6.tar.gz`) is a small Reed-Solomon / LDPC-Staircase forward-erasure-correction C library; its `tests/` directory contains only `.c` files and there is no third-party blob anywhere in the tarball.

## Fix

This PR adds a `spec-search-replace` overlay to `base/comps/openfec/openfec.comp.toml` that injects `%global debug_package %{nil}` immediately before the `Name:` line in `specs/o/openfec/openfec.spec`. That single directive causes rpmbuild to skip the `find-debuginfo.sh` extraction step entirely, so the two failing `*-debuginfo` RPMs simply stop being produced.

## Trade-off

Loses post-mortem `gdb` / `crash` / `eu-stack` debugging support for `libopenfec.so.1.4.2`. This is acceptable because:

1. `openfec` is a small FEC math library with a narrow API surface.
2. Azure Linux's debuginfo repo is a non-default, on-demand SDK feed.
3. The same scanner heuristic would otherwise re-fire on every refresh of the same `.debug` ELF, since the trigger is the standard DWARF-extraction pipeline rather than anything specific to this build.

## What was considered and not chosen

- **Build-time `-gz=none` via CFLAGS** (uncompressed DWARF). Lower-entropy but materially larger debuginfo RPMs and unverified whether AZL's `azurelinux-rpm-config` macros honour the per-spec `%undefine _find_debuginfo_opts` cleanly. Kept as a documented fallback if the canonical fix needs to be reverted.

## Validation

- Render: `azldev-in-container comp render -p openfec` (rendered specs refreshed in `specs/o/openfec/`).
- Lock: `azldev-in-container comp update -p openfec` (lock file refreshed in `locks/`).
- The runtime `openfec`, `openfec-devel`, and `openfec-utils` sub-packages built from the three `%files` blocks in the spec are unaffected.


base/comps/openfec/openfec.comp.toml
```diff
@@ -0,0 +1,7 @@
+[components.openfec]
+
+[[components.openfec.overlays]]
+description = "Suppress *-debuginfo subpackages via `%global debug_package %{nil}` to work around ESRP malware-scan false-positive (packer_high_entropy:eod) on the compressed DWARF ELFs in openfec-debuginfo-1.4.2.6-7.azl4~20260420.{x86_64,aarch64}.rpm. Trade-off: loses post-mortem gdb/crash symbols for libopenfec.so.1.4.2."
+type = "spec-search-replace"
```
reubeno: Do you just need to prepend lines to the spec? The `spec-prepend-lines` overlay works for that and should be cleaner.

PawelWMS: I thought about it but it seems to require a `section` to prepend to. Do you know where these overlays are documented btw. to get a better understanding how to use them? This is what I got from Copilot and I think it based its knowledge on `external/schemas/azldev.schema.json`, which potentially is not up-to-date. Also, I'm not sure, if the schema fully documents the usage of each overlay.

I also considered `file-prepend-lines` but Copilot did make a good point, that there's probably a reason we explicitly separate specs from other files and using `file-prepend-lines` may not be following any special rules we want to have for the specs. If that's incorrect, though, I feel like `file-prepend-lines` might be better.

christopherco (changes requested):
Generally changes look fine to me. Please squash commits together into 1 commit

base/comps/openfec/openfec.comp.toml
PawelWMS: Switched to `spec-prepend-lines` with an empty `section`.

christopherco (approved):

PR #17084, opened 2026-05-08
cloud-init: Remove NetworkManager references from cloud-init-network.service

jonathanbrenes:
AZL4 uses systemd-networkd as its network backend, not NetworkManager.
The cloud-init-network.service unit file had `After=NetworkManager.service`
and `After=NetworkManager-wait-online.service` ordering dependencies that
create unnecessary phantom not-found units on AZL4.

These are the only 2 phantom unit references worth patching — the remaining
phantom not-found units are either:
- Standard upstream behavior
- New upstream systemd 258 behavior — home.mount, boot.mount,
  ssh-keygen.target, polkit.service

Notably, firewalld `Conflicts=` (iptables/ip6tables/ebtables/ipset) and
chrony `Conflicts=`/`After=` (ntpd/ntpdate/sntp) were NOT removed — they are
intentional mutual exclusion safety guards for installable alternatives.

## Testing

- Built cloud-init successfully in azldev-runner container
- Smoke-tested built RPM: verified `NetworkManager.service` and
  `NetworkManager-wait-online.service` references are removed from
  `cloud-init-network.service`
- Installed RPM on a live AZL4 x86_64 VM
- Ran `systemctl daemon-reload && systemctl list-units --all`
- Confirmed phantom units dropped from 21 to 19 (the 2 NetworkManager
  units are gone)

Signed-off-by: Jonathan Brenes <jonathan.brenes@microsoft.com>


ddstreetmicrosoft:
NetworkManager is still provided for users who want to switch to it, so this pr isn't correct.

PR #17097, opened 2026-05-08
yara: serve modified Source0 with malware-flagged test fixtures stripped

PawelWMS:
[Koji build.](https://purple-grass-0478bae1e.7.azurestaticapps.net/workflow/df26f010-e7e5-4984-840f-08deb125bc7e)

Fixes [AB#19872](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19872)

The upstream `yara-4.5.4.tar.gz` tarball ships four files that the
automated package-signing pipeline's malware scanner rejects:

  * `tests/oss-fuzz/dotnet_fuzzer_corpus/obfuscated` -- a deliberately
    obfuscated .NET binary used as an oss-fuzz seed-corpus input for
    YARA's .NET parser fuzzer; matches generic .NET-obfuscator
    detection heuristics by design.
  * `tests/data/05cd06e6a2...e8971efded0650`
  * `tests/data/079a472d22...58ce2d3b885` (and its `.upx`-packed sibling)
  * `tests/data/e3d45a2865...86bc129f27345`

The three SHA-256-named fixtures under `tests/data/` are real PE
malware samples that the upstream YARA project ships to test the
YARA PE-format parser (one is UPX-packed). They are benign by
intent -- they are test inputs, not runtime artefacts -- but the
SRPM payload contains them verbatim, which is what the malware
scanner inspects.

Mechanism
---------

* `base/comps/yara/modify_source.sh` is a deterministic strip-and-
  repack script that downloads the upstream tarball, removes the
  four flagged files, edits `Makefile.am` to drop `test-pe` from
  the autotools `check_PROGRAMS` list AND its now-orphan
  `test_pe_SOURCES` / `test_pe_LDADD` / `test_pe_LDFLAGS` variable
  declarations (`automake` runs with `-Werror`, so orphan-variable
  warnings would fail `autoreconf` at %prep time), and repacks
  deterministically with
  `tar --sort=name --mtime=... --owner=0 --group=0 --numeric-owner
  | gzip -n -9`. The script writes its outputs to
  `base/build/work/scratch/yara/` (covered by the repo's
  top-level .gitignore via `build/`) and prints a ready-to-paste
  `az storage blob upload` command for the upload step.

* `base/comps/yara/yara.comp.toml` adds a
  `[[components.yara.source-files]]` entry pointing at the
  repacked tarball in the lookaside `repo` container under the
  `pkgs_modified/` prefix. It uses the `replace-upstream = true`
  / `replace-reason = "..."` mechanism to swap the same-named
  upstream entry in the Fedora `sources` manifest in place; no
  `file-remove`-on-`sources` overlay is needed.

* The spec's `Source0:` line and filename are unchanged; the
  upstream filename `yara-4.5.4.tar.gz` is preserved so no spec
  overlay is required to point at a different basename.

Trade-offs
----------

* Build-time loss: the upstream `test-pe` test program no longer
  runs as part of `%check`. The runtime PE rule-scanning code path
  -- the same one consumers exercise via the `yara` CLI -- is
  unaffected; this only loses upstream's own build-time regression
  coverage for the PE parser. yara's other test programs
  (`test-rules`, `test-elf`, `test-dotnet`, `test-macho`, etc.)
  continue to run normally.

* `test-pe.c` stays in the tarball (it's still compiled as part of
  `check_PROGRAMS`'s implicit compile rules until the next
  autotools regen), but the autoreconf'd `Makefile.in` no longer
  lists `test-pe` in `TESTS`, so the autotools test driver never
  invokes it. Harmless, unused.

Validation
----------

* Render: clean. `specs/y/yara/sources` carries the
  modified-tarball SHA512 on a single line.
* Lock: `locks/yara.lock` refreshed deterministically.
* `autoreconf --force --install` against the repacked tarball
  succeeds with zero `test_pe_*` warnings (same step that failed
  in the previous Koji build).
* Repack determinism: re-running `modify_source.sh` against the
  same upstream tarball yields a byte-identical output (sorted
  file order, fixed mtime, owner=group=0, numeric-owner,
  `gzip -n -9`), so the SHA512 in this commit is reproducible by
  anyone with the same upstream input.


base/comps/yara/yara.comp.toml
```diff
@@ -0,0 +1,59 @@
...
+# Upstream Fedora dist-git ships a `sources` file with the hash of
+# the original `yara-4.5.4.tar.gz` tarball.  Since we want to serve
+# a *modified* tarball under the SAME filename (so the spec's
+# Source0 line does not need to change), drop the upstream
+# `sources` file before the `source-files` block below re-creates
+# it with the modified-tarball SHA512.
+[[components.yara.overlays]]
+description = "Drop upstream Fedora `sources` file so the modified-tarball SHA512 in `source-files` below replaces (rather than conflicts with) the original `yara-4.5.4.tar.gz` entry."
+type = "file-remove"
+file = "sources"
```
PawelWMS: Checking if we can do better than this or a pattern replacement in the `sources` file.

christopherco (changes requested):
Needs commits to be squashed and render CI failure to be addressed

PawelWMS:
@christopherco both items from your review are addressed in the current PR head (`e7476496c2`):

- **Squash**: PR is now a single GPG-signed commit (you can see the linear history under "Files changed" → `1 commit` at the top of the PR).
- **Render CI**: the `check / Render + drift check` job is now passing (last run: success, 30m runtime). All other CI checks (TOML validate, locks update, license/cla, Control Tower) are green as well.

Could you please re-review when you have a moment? Happy to address any further feedback.

christopherco (approved):

PR #17102, opened 2026-05-08
fix(samba): drop winexe subpackage

PawelWMS:
Fixes: [AB#19863](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19863)

## Summary

Disables the optional `samba-winexe` subpackage so the rest of the `samba` component can ship without rebuilding the Wine-derived Windows `winexe.exe` cross-host process-launch tool.

## Why

`samba-winexe` packages mingw32/mingw64-cross-compiled Windows PE binaries shipped inside a Linux RPM. Generic .NET / Windows-PE malware-scan heuristics consistently flag those binaries when the RPM is run through our automated package-signing pipeline, irrespective of source provenance. The rest of `samba` (server, libs, client, dc, vfs modules, ctdb) signs cleanly without winexe.

## Mechanism

The upstream `samba.spec` already gates `BuildRequires`, `%package winexe`, and `%files winexe` behind `%if %{with winexe}`. Adding `build.without = ["winexe"]` to `base/comps/samba/samba.comp.toml` causes `azldev` to pass `--without winexe` to rpmbuild and unconditionally drop the subpackage. **No spec mutation is needed** — which means zero overlay maintenance burden across upstream samba uplifts and full reversibility (just delete the `build.without` line).

This PR pairs the `build.without` toggle with a deletion of the now-dangling `"samba-winexe"` line from `base/packages/base.packages.toml` so the image manifest does not reference a subpackage the build no longer produces.

The `release` calculation in `base/comps/samba/samba.comp.toml` is also switched from `"manual"` to `"autorelease"` so `azldev` can derive the Release tag from the upstream spec's `rpmautospec` directives instead of carrying a manual override.

## Dependency-impact analysis

Repo-wide grep of `base/comps/**/*.{toml,comp.toml}` and `specs/**/*.spec` for `BuildRequires:.*samba-winexe`, `Requires:.*samba-winexe`, and `winexe` (excluding the gating `%bcond winexe` lines and the `%package winexe` declaration inside `samba.spec` itself) returns **zero consumers**. Independently re-confirmed by a dependency-impact reviewer pass. `samba-winexe` therefore has no reverse dependencies in this distro, and dropping the subpackage is safe.

## Functionality lost

Only the optional `samba-winexe` subpackage containing the Wine-derived Windows `winexe.exe` cross-host process-launch tool. Niche use case — modern alternatives include OpenSSH on Windows, WinRM, and PowerShell Remoting. The rest of the `samba` component (server, libs, client, dc, vfs modules, ctdb) is unaffected.

## Validation

- Render: `azldev-in-container comp render -p samba` (rendered files refreshed under `specs/s/samba/`).
- Lock: `azldev-in-container comp update -p samba` (lock file refreshed in `locks/samba.lock`).
- [Koji build.](https://purple-grass-0478bae1e.7.azurestaticapps.net/workflow/086687cf-e242-4b86-cfac-08deaca112e9)


base/comps/samba/samba.comp.toml
```diff
@@ -1,3 +1,29 @@
...
+#     A repo-wide grep of `base/comps/**/*.{toml,comp.toml}` and
+#     `specs/**/*.spec` for `BuildRequires:.*samba-winexe`,
+#     `Requires:.*samba-winexe`, and `winexe` returns no consumers, so
+#     dropping the subpackage is safe.
+#
+# Pairs with a deletion of the now-dangling `"samba-winexe"` line from
+# `base/packages/base.packages.toml` so the image manifest does not
+# reference a sub-package the build no longer produces.
+[components.samba.build]
+without = ["winexe"]
```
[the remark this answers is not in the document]
PawelWMS: Ran a build, linked in the summary.

base/comps/samba/samba.comp.toml
```diff
@@ -1,3 +1,10 @@
 [components.samba]
 # Release: %{samba_release}
-release = { calculation = "manual" }
+release = { calculation = "autorelease" }
```
reubeno, line 3: What's the reason for the change? Any context?

christopherco: I had the same question and @dmcilvaney did work with @pawelwms on this.

Samba's release indirectly has `Release: %autorelease` (via `Release:        %{samba_release}` and `%global samba_release %autorelease`), but our tools can't easily follow that, so the `calculation = "autorelease"` was used

christopherco (approved):

PR #17103, opened 2026-05-09
fix(java): disable slowdebug variant for openjdk builds

christopherco:
Add `build.without = ["slowdebug"]` to java-25-openjdk and java-25-openjdk-portable to skip building the -slowdebug subpackages.

Drop the corresponding *-slowdebug entries from base.packages.toml and sdk.packages.toml so the package lists stay consistent with what the build actually produces.

Fixes: [AB#19862](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19862)


reubeno (commented):
Changes look good to me, when you're ready.

christopherco:
I need to fix up the `release` calculation because the increment isn't being honored. I'll likely need to do something similar to what we do in the `kernel` package

reubeno (approved):

base/comps/java-25-openjdk-portable/java-25-openjdk-portable.comp.toml
```diff
@@ -11,6 +11,19 @@
 # Release: %{?eaprefix}%{rpmrelease}%{?extraver}%{?dist}
 release = { calculation = "manual" }
 spec = { type = "upstream", upstream-distro = { name = "fedora", version = "43" }, upstream-commit = "cad9d5a1ed48cf7e9e12cfadfdacc11f523d46cb" }
+# Disable slowdebug variant builds; we don't ship -slowdebug subpackages.
+build.without = ["slowdebug"]
+
+[components.java-25-openjdk-portable.build.defines]
+# AZL-controlled rpmrelease number. Bump when shipping new AZL-only changes
+# on top of the pinned upstream commit.
+azl_rpmrelease = "3"
```
christopherco: From offline discussion with @reubeno, we should explore this number being an incremental on top of the base release number from the Fedora package we're building on top of. Will do a timeboxed experiment on this

PR #17104, opened 2026-05-09
fix: remove mingw from distro

trungams:
- Remove all MinGW components from distro
- Disable any `mingw*` subpackage currently being built for the remaining components
- Explicitly set build `--without mingw` for packages that currently has mingw references


base/comps/rust/rust.comp.toml
```diff
@@ -2,3 +2,10 @@
 # Fedora repo does not host 1.93, which the spec gets pinned to. Fetch
 # spec from a more recent commit to pickup version - 1.94.1.
 spec = { type = "upstream", upstream-distro = { name = "fedora", version = "43"}, upstream-commit = "b431103e3e71d30dda680c95d2f5d4fe50003436" }
+
+# Azure Linux does not ship mingw cross-compilation toolchains.
+[[components.rust.overlays]]
+description = "Disable mingw cross-compilation targets — Azure Linux does not ship mingw toolchains"
+type = "spec-search-replace"
+regex = '%global mingw_targets i686-pc-windows-gnu x86_64-pc-windows-gnu'
+replacement = '#global mingw_targets i686-pc-windows-gnu x86_64-pc-windows-gnu'
```
reubeno, line 11: Is this going to result in the macro being left undefined? Should we be setting it to `%{nil}` or empty string or something?

base/packages/sdk.packages.toml
```diff
@@ -3791,362 +3790,38 @@ packages = [
...
-    "mingw32-brotli", # srpm: mingw-brotli
-    "mingw32-bsdcat", # srpm: mingw-libarchive
-    "mingw32-bsdcpio", # srpm: mingw-libarchive
-    "mingw32-bsdtar", # srpm: mingw-libarchive
-    "mingw32-bzip2", # srpm: mingw-bzip2
-    "mingw32-bzip2-static", # srpm: mingw-bzip2
-    "mingw32-c-ares", # srpm: mingw-c-ares
-    "mingw32-cairo", # srpm: mingw-cairo
-    "mingw32-cairo-static", # srpm: mingw-cairo
     "mingw32-cereal", # srpm: cereal
```
reubeno: Why wasn't this removed? I thought we disable mingw in cereal too.

base/packages/sdk.packages.toml
```diff
@@ -3791,362 +3790,38 @@ packages = [
...
-    "mingw64-brotli", # srpm: mingw-brotli
-    "mingw64-bsdcat", # srpm: mingw-libarchive
-    "mingw64-bsdcpio", # srpm: mingw-libarchive
-    "mingw64-bsdtar", # srpm: mingw-libarchive
-    "mingw64-bzip2", # srpm: mingw-bzip2
-    "mingw64-bzip2-static", # srpm: mingw-bzip2
-    "mingw64-c-ares", # srpm: mingw-c-ares
-    "mingw64-cairo", # srpm: mingw-cairo
-    "mingw64-cairo-static", # srpm: mingw-cairo
     "mingw64-cereal", # srpm: cereal
```
reubeno: And this one too?

base/packages/sdk.packages.toml
```diff
@@ -36,7 +36,6 @@ packages = [
     "CppMicroServices", # srpm: CppMicroServices
```
reubeno, line 36: I commented on a couple, but I see other instances too; can you double-check the mingw packages left behind in this file and make sure the entries should still be there?

base/comps/rust/rust.comp.toml
trungams: The condition check is `%if %defined mingw_targets`, so we're still "defining" the macro if we set it to `%{nil}`/empty string. I double checked this behaviour with a local spec.

base/packages/sdk.packages.toml
trungams: My bad. I was only checking for `# srpm: mingw*` pattern in the toml

trungams: Updated the toml

reubeno (approved):

PR #17105, opened 2026-05-09
[RFC] fix(ghc-srpm-macros): disable Haskell profiling libraries

christopherco:
Azure Linux does not ship any ghc-*-prof RPMs. Building them roughly doubles the build time of every Haskell package and produces ~290 unused subpackages across the ecosystem.

Flip the `with_ghc_prof` defaults in macros.ghc-srpm so all ghc-* SRPMs (ghc itself plus the cabal-rpm generated downstream specs that gate `%files prof` on `%if %{with ghc_prof}`) skip building their profiling subpackages by default.

Also drop the corresponding ghc-*-prof entries (and the ghc-prof metapackage) from sdk.packages.toml so the SDK package list reflects what actually ships.

Verified: ghc-srpm-macros builds; in a mock chroot with the new RPM installed, `rpm --eval '%{?with_ghc_prof}'` returns `0`.


reubeno:
Change looks pretty good.

PR #17116, opened 2026-05-10
ci(pipeline): add package build to Control Tower integration

dmcilvaney:
Add a "Submit package build to Control Tower" step that calls the `/api/Scenario/package` endpoint after the prcheck step succeeds.

Both API calls live in a single job. Two jobs would have doubled the fixed-cost OneBranch SDL/binary-analysis injections, required a cross-job artifact handoff for the changed-components JSON, and bought us only marginal isolation -- upload finishes in minutes and the package-build call only briefly polls to confirm acceptance.

## Stage / job rename

The pipeline now does more than source upload, so the **stage and job** inside the YAML are renamed to reflect what it actually does. The **file names stay the same** for now -- renaming the wrapper file would require a coordinated portal update of the ADO pipeline definition's YAML path, which will happen in a follow-up PR.

* Stage `PRCheck` -> `Integration`
* Job `CallControlTowerAPI` -> `UploadAndBuild`

## Package-build step details

* `run_package_build.py` reuses the changed-components JSON from earlier in the job, filters to `changeType in {added, changed}`, and submits to `/api/Scenario/package`.
* `--package-target` is a parameter of the stages template (defaults set by the wrapper, currently `azl4`).
* `--official-build` is opt-in: omitting it submits a scratch build (`isScratchBuild=true`); the production wrapper passes `--official-build` so production runs submit non-scratch (persisted) builds.
* `--build-reason` is consumed by the script for a local PR-trigger skip guard (`buildReason` is also forwarded in the API payload). PR triggers do not submit package builds; the skip lives in the script for parity with `run_prcheck.py`.
* The script polls briefly (default 10 min) for the job to reach a terminal state, surfacing immediate failures. A non-terminal status at the timeout is treated as acceptance -- the build continues async and is monitored in Control Tower.

## Follow-up after merge

* Rename `sources-upload[.yml,-stages.yml]` -> `control-tower-integration*` and update the ADO build definition's YAML file path in the portal at the same time.
* Remove job-level `continueOnError: true` once the pipeline is stable (tracked under ADO task 19179).

### DIFF
--- a/.github/workflows/ado/sources-upload.yml
+++ b/.github/workflows/ado/sources-upload.yml
@@ -7,7 +7,13 @@
 #   .github/workflows/ado/templates/sources-upload-stages.yml
 #
 # Authenticates via Workload Identity Federation (OIDC) and calls the Control
-# Tower prcheck API with PR context.
+# Tower APIs to:
+#   1. Validate that the rendered sources of every changed component can be
+#      fetched from the lookaside (prcheck). The actual upload happens later
+#      from the merge queue, not here.
+#   2. Submit scratch package builds for changed components.
+#
+# Helper scripts live under .github/workflows/scripts/control-tower/.
 #
 # Prerequisites (ADO / Azure Portal):
 #   1. Entra ID App Registration with audience URI
--- a/.github/workflows/scripts/control-tower/run_package_build.py
+++ b/.github/workflows/scripts/control-tower/run_package_build.py
@@ -1,0 +1,267 @@
+"""Submit a package-build job to the Control Tower service and wait briefly.
+
+Flow:
+    1. Read the changed-components JSON.
+    2. Filter to the build set: ``changeType in {added, changed}`` -- any
+       component whose inputs changed needs a rebuild, regardless of whether
+       its ``sourcesChange`` flag is set.
+    3. POST ``/api/Scenario/package`` with the build request.
+    4. Poll briefly (default 5 min) until the job reaches a terminal state
+       (success or failure) or the local timeout expires. The goal is to
+       catch jobs that fail immediately on submission, not to wait for the
+       full build -- a non-terminal status at timeout is treated as
+       acceptance and the build continues async.
+    5. Exit 0 if the job started (or completed). Exit 1 only on submission
+       failure or immediate terminal failure.
+"""
+
+import argparse
+import json
+import sys
+from pathlib import Path
+
+from azure.identity import DefaultAzureCredential
+
+import client as ct
+
+
+def _load_build_components(path: Path) -> list[str]:
+    """Filter the ``azldev component changed`` JSON to the build set.
+
+    The build set is every component with ``changeType`` in ``{added, changed}``
+    — these are the components whose inputs differ between source and target
+    and therefore need a rebuild. Unlike the upload set, we do NOT filter on
+    ``sourcesChange`` here: a component can need a rebuild even if its source
+    tarballs didn't change (e.g. an overlay or build-config change).
+
+    Deleted components are excluded — there is nothing to build.
+    """
+    try:
+        raw = path.read_text(encoding="utf-8")
+    except OSError as exc:
+        raise SystemExit(
+            f"##[error]Failed to read --changed-components-file {path!s}: {exc}"
+        ) from exc
+
+    try:
+        entries = json.loads(raw)
+    except json.JSONDecodeError as exc:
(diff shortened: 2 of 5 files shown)


reubeno (commented):
Why did we end up with "package build" as the contract between CI and Control Tower? I understand that we've made the call to detect changes on the workflow/pipeline side -- but my expectation is still that the Control Tower will decide which tasks need to be run, depending on whether it's pre-merge qualification, or as part of merge/post-merge.

Did we not add endpoints to Control Tower that are more scenario driven?

dmcilvaney:
> Why did we end up with "package build" as the contract between CI and Control Tower? I understand that we've made the call to detect changes on the workflow/pipeline side -- but my expectation is still that the Control Tower will decide which tasks need to be run, depending on whether it's pre-merge qualification, or as part of merge/post-merge.
> 
> Did we not add endpoints to Control Tower that are more scenario driven?

We looked at the control tower side, and the cost there was much higher than integrating it here in the short-term. I built the azldev components to support this so it will be easy to move this logic into CT if we desire, but this was the shortest path to functionality.

.github/workflows/ado/sources-upload.yml
```diff
@@ -7,7 +7,11 @@
 #   .github/workflows/ado/templates/sources-upload-stages.yml
 #
 # Authenticates via Workload Identity Federation (OIDC) and calls the Control
-# Tower prcheck API with PR context.
+# Tower APIs to:
+#   1. Upload changed component sources (prcheck).
+#   2. Submit package builds for changed components (package-build, non-PR only).
```
PawelWMS: Not sure I get this comment. Also, the PR check is not when we upload the sources, yet. We'd want to make sure they can be uploaded (for instance `azl prepare-sources` succeeds to grab them with the current TOML config) but don't upload, yet. The upload itself is going to happen in the merge queue.

.github/workflows/scripts/control-tower/run_package_build.py
```diff
@@ -0,0 +1,272 @@
+"""Submit a package-build job to the Control Tower service and wait briefly.
+
+Flow:
+    1. Read the changed-components JSON (from ``azldev component changed``).
```
PawelWMS: Nit-pick: the script doesn't need to know where the component list came from. Also, the comment may become easily stale, if we, for instance, introduce a variant of the pipeline, where the (admin) user dictates the components. But my main point is, that in general the script doesn't need to keep track of where its inputs are coming from.

.github/workflows/scripts/control-tower/run_package_build.py
```diff
@@ -0,0 +1,272 @@
...
+        "--commit-sha",
+        default=None,
+        help="Source commit SHA to build from.",
+    )
+    parser.add_argument(
+        "--branch",
+        default=None,
+        help="Source branch name (alternative to --commit-sha).",
+    )
+    parser.add_argument(
```
PawelWMS, line 118: How do we set the Koji tag for the non scratch build scenario? Or is this figured out automatically by CT from the TOMLs?

.github/workflows/scripts/control-tower/run_package_build.py
```diff
@@ -0,0 +1,272 @@
...
+        "--scratch-build",
+        action="store_true",
+        default=False,
+        help="Submit as a scratch build (default: false / official build).",
+    )
+    parser.add_argument(
```
PawelWMS: Why do we need two separate flags to decide if it's a scratch build or not?

.github/workflows/scripts/control-tower/run_package_build.py
```diff
@@ -0,0 +1,272 @@
...
+        "--branch",
+        default=None,
+        help="Source branch name (alternative to --commit-sha).",
+    )
+    parser.add_argument(
+        "--scratch-build",
+        action="store_true",
+        default=False,
```
PawelWMS, line 121: It feels safer to do a scratch build by default. In general, I think it might be reasonable to force the caller to explicitly say they want an official, persisted package build to make it easier to avoid mistakes.

PawelWMS (approved):

.github/workflows/scripts/control-tower/run_package_build.py
dmcilvaney: Its just static, could move it to the variable group if needed?

christopherco (approved):

PR #17120, opened 2026-05-11
fix(gnome-remote-desktop): remove dependency on opus to drop it

binujp:
Force %bcond rdp to 0 to disable the RDP backend, which has a hard
dependency on opus via meson. Guard RDP-only file entries in %files
with %if %{with rdp} blocks so they are excluded when RDP is off.


specs/g/gnome-remote-desktop/gnome-remote-desktop.spec
```diff
@@ -18,7 +18,7 @@
 
 %global tarball_version %%(echo %{version} | tr '~' '.')
 
-%bcond rdp %[0%{?fedora} || 0%{?rhel} >= 10]
+%bcond rdp 0
```
reubeno: If there's already a bcond, we need not do any overlay here; we can just set a `without = ["rdp"]` in the TOML file.

base/comps/gnome-remote-desktop/gnome-remote-desktop.comp.toml
```diff
@@ -0,0 +1,62 @@
...
+regex = '%bcond rdp %\[0%\{\?fedora\} \|\| 0%\{\?rhel\} >= 10\]'
+replacement = '%bcond rdp 0'
+
+# Guard RDP-only libexec entries
+[[components.gnome-remote-desktop.overlays]]
+description = "Guard RDP-only libexec files in %files"
+type = "spec-search-replace"
+section = "%files"
+regex = '%\{_libexecdir\}/gnome-remote-desktop-enable-service'
+replacement = """%if %{with rdp}
```
reubeno: These overlays take a dependency on specific knowledge that the RDP-related lines are consecutive, and that this is the first one.

I know the resulting rendered .spec looks like what we'd expect -- but the overlays now run the risk of generating unexpected results if the upstream spec happens to reorder the entries in %files.

base/comps/gnome-remote-desktop/gnome-remote-desktop.comp.toml
```diff
@@ -0,0 +1,62 @@
+[components.gnome-remote-desktop]
+
+# AZL does not ship opus. Opus is a hard meson dependency when RDP is
+# enabled, so disable the RDP backend by forcing the bcond to off.
+# We also guard the RDP-only file entries so %files doesn't list missing files.
```
reubeno: This would be a bug in the upstream spec, yes? We should add it to Chris's list of changes-to-upstream.

base/comps/gnome-remote-desktop/gnome-remote-desktop.comp.toml
binujp: Fixed with independent search replace for each.

reubeno (approved):

PR #17121, opened 2026-05-11
fix(gstreamer1-plugins-base): remove dependency on opus to drop it

binujp:
Remove opus-devel BuildRequires, add -Dopus=disabled meson flag, and
remove libgstopus.so from %files. This is to allow removal of the opus package from AZL.


base/comps/gstreamer1-plugins-base/gstreamer1-plugins-base.comp.toml
```diff
@@ -0,0 +1,24 @@
...
+tag = "BuildRequires"
+value = "opus-devel"
+
+[[components.gstreamer1-plugins-base.overlays]]
+description = "Disable the opus meson option"
+type = "spec-search-replace"
+section = "%build"
+regex = '-D tremor=disabled'
+replacement = """-D opus=disabled \\
+  -D tremor=disabled"""
```
reubeno: Avoid referencing an unrelated option?

christopherco: Consider using `%meson \` as your anchor point

christopherco (dismissed):

base/comps/gstreamer1-plugins-base/gstreamer1-plugins-base.comp.toml
binujp: Done, used same pattern as gstreamer base free

reubeno (approved):

PR #17122, opened 2026-05-11
fix(gstreamer1-plugins-bad-free): remove dependency on opus to drop it

binujp:
Remove pkgconfig(opus) BuildRequires, add -Dopus=disabled meson flag,
and remove libgstopusparse.so from %files. This is to allow removal of the opus package from AZL.

### DIFF
--- a/base/comps/gstreamer1-plugins-bad-free/gstreamer1-plugins-bad-free.comp.toml
+++ b/base/comps/gstreamer1-plugins-bad-free/gstreamer1-plugins-bad-free.comp.toml
@@ -1,0 +1,24 @@
+[components.gstreamer1-plugins-bad-free]
+
+# AZL does not ship opus — disable the opus parse plugin and remove its BR.
+
+[[components.gstreamer1-plugins-bad-free.overlays]]
+description = "Remove pkgconfig(opus) BuildRequires (opus not available in AZL)"
+type = "spec-remove-tag"
+tag = "BuildRequires"
+value = "pkgconfig(opus)"
+
+[[components.gstreamer1-plugins-bad-free.overlays]]
+description = "Disable the opus meson option"
+type = "spec-search-replace"
+section = "%build"
+regex = '%meson \\'
+replacement = """%meson \\
+    -D opus=disabled \\"""
+
+[[components.gstreamer1-plugins-bad-free.overlays]]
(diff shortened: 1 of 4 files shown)


base/comps/gstreamer1-plugins-bad-free/gstreamer1-plugins-bad-free.comp.toml
```diff
@@ -0,0 +1,24 @@
...
+tag = "BuildRequires"
+value = "pkgconfig(opus)"
+
+[[components.gstreamer1-plugins-bad-free.overlays]]
+description = "Disable the opus meson option"
+type = "spec-search-replace"
+section = "%build"
+regex = '-D x265=disabled'
+replacement = """-D opus=disabled \\
+    -D x265=disabled"""
```
reubeno: Similar comments as on other PRs; is there a way to avoid coupling the change to a specific unrelated option present in the spec?

christopherco: Perhaps anchor on `%meson \` or `%{nil}` instead. These should be more durable than the x265 option

specs/g/gstreamer1-plugins-bad-free/gstreamer1-plugins-bad-free.spec
christopherco, line 454: issue (non-blocking): Opus is referenced in the AppStream component. Now, we probably don't need to worry about AppStream in our case, but marking this for completeness.

christopherco, line 479: issue (non-blocking): Opus is referenced in the AppStream component. Now, we probably don't need to worry about AppStream in our case, but marking this for completeness.

christopherco (dismissed):
Fix anchor point for regex

base/comps/gstreamer1-plugins-bad-free/gstreamer1-plugins-bad-free.comp.toml
binujp: Yes, agree. I will use the same pattern as for spice-gtk.

reubeno (approved):

PR #17123, opened 2026-05-11
fix(jack-audio-connection-kit): remove dependency on opus to drop it

binujp:
Remove opus-devel BuildRequires and add --no-opus to waf configure
to prevent auto-detection. This is to allow removal of the opus package from AZL.


base/comps/jack-audio-connection-kit/jack-audio-connection-kit.comp.toml
```diff
@@ -0,0 +1,17 @@
...
+description = "Remove opus-devel BuildRequires (opus not available in AZL)"
+type = "spec-remove-tag"
+tag = "BuildRequires"
+value = "opus-devel"
+
+[[components.jack-audio-connection-kit.overlays]]
+description = "Disable opus support in waf configure"
+type = "spec-search-replace"
+section = "%build"
+regex = '   --alsa'
```
reubeno: Can we be less coupled to the specific whitespace present in the spec right now?

christopherco: It might be better to have this regex anchor on python3 ./waf configure \

christopherco (dismissed):
Please fix the regex anchor

base/comps/jack-audio-connection-kit/jack-audio-connection-kit.comp.toml
binujp: Done

reubeno (approved):

PR #17124, opened 2026-05-11
fix(libsndfile): remove dependency on opus to drop it

binujp:
This is to allow removal of the opus package from AZL. libsndfile auto-detects opus via
pkg-config and builds without opus support when it is absent.


base/comps/libsndfile/libsndfile.comp.toml
```diff
@@ -0,0 +1,11 @@
+[components.libsndfile]
+
+# AZL does not ship opus — remove the BR so libsndfile builds without opus.
+# libsndfile autodetects opus via pkg-config; without opus-devel it simply
+# disables opus support at configure time.
```
christopherco: **issue (blocking):** B1 — Comment will mislead future maintainers; the all-or-nothing semantics is not what's described here.

The text "without opus-devel it simply disables opus support at configure time" is contradicted by upstream `libsndfile@1.2.2:configure.ac:321-342`:

```autoconf
AS_IF([test "x$ac_cv_flac$ac_cv_ogg$ac_cv_vorbis$ac_cv_vorbisenc$ac_cv_opus"
       = "xyesyesyesyesyes"], [
    HAVE_EXTERNAL_XIPH_LIBS=1 ...
], [
    AC_MSG_WARN([[*** Unfortunately, for ease of maintenance, the external libs]])
    AC_MSG_WARN([[*** are an all or nothing affair.]])
    enable_external_libs=no
])
```

When opus-devel is missing, `ac_cv_opus=no`, the AND-check fails, and `HAVE_EXTERNAL_XIPH_LIBS` goes to 0 — disabling FLAC, Ogg, **and** Vorbis along with Opus. The configure summary on the build log will read `External FLAC/Ogg/Vorbis/Opus : ....... no`.

The same file `specs/l/libsndfile/libsndfile.spec` carries a 2021 Fedora `%changelog` entry that's the historical record of this: `* Mon Feb 22 2021 Michal Hlavinka — 1.0.31-3 — add opus-devel BR to satisfy configure requirements check (#1931251)`. Fedora added the BR precisely to keep the Xiph group enabled; this PR reverses that without acknowledging it.

The comp.toml comment lives in the repo permanently, so any inaccuracy here is durable debt.

base/comps/libsndfile/libsndfile.comp.toml
```diff
@@ -0,0 +1,11 @@
...
+
+[[components.libsndfile.overlays]]
+description = "Remove opus-devel BuildRequires (opus not available in AZL)"
+type = "spec-remove-tag"
+tag = "BuildRequires"
+value = "opus-devel"
```
christopherco, line 13: **issue (non-blocking):** H1 — Add `--disable-external-libs` + drop the now-dead Xiph BRs in the same overlay.

Related to B1. After this PR, the rendered spec still carries:

- `BuildRequires: flac-devel` (`specs/l/libsndfile/libsndfile.spec:23`)
- `BuildRequires: libogg-devel` (`:25`)
- `BuildRequires: libvorbis-devel` (`:26`)

Configure will still probe and find them (`ac_cv_flac=yes`, `ac_cv_ogg=yes`, `ac_cv_vorbis=yes`), then immediately discard all four because `ac_cv_opus=no`. The three dev packages contribute zero linkage to the output RPMs — they slow the buildroot, bloat the dep graph the opus-removal workstream is trying to thin, and make the spec actively lie about which formats are enabled. Future maintainers reading the BR list will reasonably conclude FLAC/Vorbis/Ogg support remains.

This is exactly the "implicit feature disable vs. explicit `--disable-<feature>`" pattern: the policy should be encoded at the build layer, not as a side-effect of buildroot composition.

Suggested change in this comp.toml (append to the existing overlay):

```toml
[[components.libsndfile.overlays]]
description = "Drop dead flac-devel BR (libsndfile builds without external Xiph)"
type = "spec-remove-tag"
tag = "BuildRequires"
value = "flac-devel"

[[components.libsndfile.overlays]]
description = "Drop dead libogg-devel BR"
type = "spec-remove-tag"
tag = "BuildRequires"
value = "libogg-devel"

[[components.libsndfile.overlays]]
description = "Drop dead libvorbis-devel BR"
type = "spec-remove-tag"
tag = "BuildRequires"
value = "libvorbis-devel"

[[components.libsndfile.overlays]]
description = "Make the Xiph external-libs disable explicit rather than warning-driven"
type = "spec-search-replace"
# Add --disable-external-libs to %configure (anchor on a stable adjacent flag)
find = "--enable-mpeg \\\\"
replace = "--enable-mpeg \\\\\n\t--disable-external-libs \\\\"
```

(`spec-search-replace` mechanics: please confirm the exact anchor/escape form matches the convention in this repo — `azldev` should accept the substitution against the unchanged `--enable-mpeg \` line.)

christopherco (dismissed):

reubeno (approved):

PR #17125, opened 2026-05-11
fix(qt6): clear qt6_qtwebengine_arches macro to drop qt6-qtwebengine dependency

binujp:
Override the qt6_qtwebengine_arches macro to %{nil} so that dependent packages
(gpsbabel, libksysguard, python-pyside6, python-pyqt6, qt6-qtwebview) skip
their qtwebengine-related code paths, to allow removal of the qt6-qtwebengine
package from AZL.


base/comps/qt6/qt6.comp.toml
```diff
@@ -0,0 +1,12 @@
+[components.qt6]
+
+# Override the qt6_qtwebengine_arches macro to empty so that dependent packages
+# (gpsbabel, libksysguard, python-pyside6, python-pyqt6, qt6-qtwebview) skip
```
reubeno: So this means we need to dash-roll/release-bump these packages, yes? Do you aim to do that in a separate PR, or in this one? I believe you can use `azldev component update --bump <list of component names...>` to make this happen.

binujp: That has to be a separate PR. This change has to merge and deliver updates qt6 macros first.

Thanks for the heads-up, did not know about `update --bump`, nice!

base/comps/qt6/qt6.comp.toml
```diff
@@ -0,0 +1,13 @@
...
+# Override the qt6_qtwebengine_arches macro to a non-existent architecture so
+# that dependent packages (gpsbabel, libksysguard, python-pyside6, python-pyqt6,
+# qt6-qtwebview) skip their qtwebengine-related code paths, breaking the
+# dependency on qt6-qtwebengine which has pre-existing build failures on AZL.
+# Note: %{nil} cannot be used because %ifarch requires at least one argument.
+[[components.qt6.overlays]]
+description = "Clear qt6_qtwebengine_arches macro to disable qtwebengine across dependent packages"
+type = "file-search-replace"
+file = "macros.qt6-srpm"
+regex = '%qt6_qtwebengine_arches x86_64 aarch64 ppc64le riscv64'
```
reubeno: I agree with copilot here that we should make this avoid needing to know the current defined list (in case it changes).

binujp: Agree, updated the regular expression to be more tolerant.

reubeno (approved):

PR #17127, opened 2026-05-11
fix(spice-gtk): remove dependency on opus to drop it

binujp:
Remove opus-devel BuildRequires and add -Dopus=disabled to meson
build. This is to allow removal of the opus package from AZL.


base/comps/spice-gtk/spice-gtk.comp.toml
```diff
@@ -0,0 +1,17 @@
...
+description = "Remove opus-devel BuildRequires (opus not available in AZL)"
+type = "spec-remove-tag"
+tag = "BuildRequires"
+value = "opus-devel"
+
+[[components.spice-gtk.overlays]]
+description = "Disable opus audio codec in meson build"
+type = "spec-search-replace"
+section = "%build"
+regex = '-Dbuiltin-mjpeg=false'
```
reubeno: Would it be cleaner to search for `%meson \` or similar? That's a bit less specific.

christopherco (dismissed):
Fix regex anchor and merge conflicts

base/comps/spice-gtk/spice-gtk.comp.toml
binujp: I agree with the cleaner anchor and have pushed copilot's suggestion instead.

This class of changes hide a larger problem with regex-replace. We must be able to find patterns we are searching for and fail the overlay if we do not. The overlay has to be as specific. More generic exposes more edge cases of accidental matches.

reubeno (approved):

PR #17128, opened 2026-05-11
fix(wireshark): remove dependency on opus to drop it

binujp:
Remove opus-devel BR. Wireshark auto-detects opus and builds without
Opus RTP audio playback when it is absent. This is to allow removal of the opus package from AZL.


base/comps/wireshark/wireshark.comp.toml
```diff
@@ -0,0 +1,10 @@
+[components.wireshark]
```
christopherco, line 1: suggestion (non-blocking): Consider upstreaming a change to Fedora to allow toggling `-DENABLE_OPUS=ON/OFF` in cmake invocation. Then we can use that in the future to ensure opus isn't built.

christopherco (approved):
LGTM. Added a non-blocking suggestion to upstream a toggle to Fedora

PR #17132, opened 2026-05-11
chore(libheif): Remove libheif

christopherco:
Disable HEIF support in the six direct consumer SRPMs (GraphicsMagick, ImageMagick, glycin, imlib2, kf6-kimageformats, vips) using upstream- supported flags, then drop the libheif and rust-libheif-{rs,sys} SRPMs along with all of their subpackages from the package manifests.

No source-level patches are required; each consumer either has a clean %bcond gating libheif (ImageMagick, glycin) or auto-detects it via pkg-config / has an explicit cmake/meson flag we can flip. vips needs both the BR drop and an explicit '-Dheif=disabled' because its meson 'heif' option is type 'feature' with default 'auto' (which fails the build when libheif is absent rather than silently skipping).

Verified: all six consumers rebuild cleanly with no libheif Requires in the produced RPMs, and 'azldev comp list' no longer resolves libheif, rust-libheif-rs, or rust-libheif-sys.

Fixes: [AB#19713](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19713)
Fixes: [AB#19715](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19715)
Fixes: [AB#19714](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19714)


base/comps/vips/vips.comp.toml
```diff
@@ -0,0 +1,49 @@
...
+# libvips defines `option('heif', type : 'feature', value : 'auto')`, so meson
+# fails when libheif is not found rather than silently skipping it. Force the
+# feature off explicitly. We piggyback on the existing `-Dpdfium=disabled \`
+# line to keep the diff minimal.
+[[components.vips.overlays]]
+description = "Force `-Dheif=disabled` so meson does not fail trying to auto-detect libheif"
+type = "spec-search-replace"
+section = "%build"
+regex = '-Dpdfium=disabled \\'
+replacement = "-Dpdfium=disabled \\\n    -Dheif=disabled \\"
```
reubeno, line 31: (Non-blocking) This sort of search/replace leads to a nice looking rendered spec, but it couples it to an unrelated config option.

I think the cleaner overlay would be to add something after `%meson`?

reubeno (approved):

PR #17133, opened 2026-05-11
fix(firefox): repack source tarball without ESRP-flagged ff-inst.exe and 607222.html

anphel31:
Fixes: [AB#19869](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19869)

## Summary

The upstream Mozilla Firefox source tarball ships content flagged by the
AZL RPM signing pipeline (ESRP) as obfuscated / malware-like:

- `firefox-148.0/toolkit/components/mediasniffer/test/unit/data/ff-inst.exe`
  — an obfuscated Windows executable used only by Firefox's media-sniffer
  unit tests.
- `firefox-148.0/dom/base/crashtests/607222.html` — crashtest HTML that
  trips malware scanners.

Such content blocks the SRPM from being published. Neither file is used at
runtime — they only feed Firefox's unit/crashtests, which are not run in
our build (`run_firefox_tests = 0`) — so removing them has no effect on
the shipped package.

This PR replaces `Source0` (effectively, see below) with a
deterministically-repacked tarball that is byte-identical to upstream
**except** that those two files have been stripped. The swap is done via
azldev's new `replace-upstream` mechanism, which rewrites the matching
entry in Fedora's `sources` lookaside manifest in place — so we keep the
upstream filename and the spec's `Source0` line is untouched.

## Changes

- **`base/comps/firefox/modify_source.sh`** — New helper script that:
  - Downloads `firefox-148.0.source.tar.xz` from `archive.mozilla.org`.
  - Verifies its SHA-512 against the value in
    [`specs/f/firefox/sources`](../specs/f/firefox/sources).
  - Strips every file listed in `REMOVE_PATHS` (currently `ff-inst.exe`
    and `607222.html`; add more there as new flags arise).
  - Deterministically repacks as `firefox-148.0.azl.source.tar.xz`:
    - `tar --sort=name --owner=0 --group=0 --numeric-owner --mtime=@1577836800 --format=gnu` — stable file order, fixed ownership and mtime, GNU format to handle long paths.
    - `xz -T 4 -9 --block-size=256MiB` — fixed thread count and block size so the SHA-512 is identical across re-runs and machines.
  - Emits a SHA-512 sidecar for the modified tarball and prints the
    `az storage blob upload` command for the AZL lookaside cache. The
    upload uses the **upstream** filename (`firefox-148.0.source.tar.xz`)
    as the blob name because `replace-upstream` matches by filename.

- **`base/comps/firefox/firefox.comp.toml`** — Adds a single
  `[[components.firefox.source-files]]` entry with:
  - `filename = "firefox-148.0.source.tar.xz"` (same as upstream).
  - `hash` / `hash-type = "SHA512"` of the repacked tarball.
  - `origin` URL pointing at the AZL lookaside cache.
  - `replace-upstream = true` + `replace-reason = "…"` so azldev
    transparently swaps the upstream entry in the `sources` manifest with
    auditable WARN-level logging. No spec-search-replace or
    file-search-replace overlays are needed.

## Reproduction

Anyone can regenerate the modified tarball:

```bash
./base/comps/firefox/modify_source.sh
# Writes to base/build/work/scratch/firefox/firefox-148.0.azl.source.tar.xz
# along with a .sha512 sidecar file.
```

## Verification
- Test [build](https://purple-grass-0478bae1e.7.azurestaticapps.net/workflow/a6d2bd0c-d33e-437e-851e-08deb1fc48fe)


base/comps/firefox/firefox.comp.toml
```diff
@@ -31,3 +31,40 @@ description = "Disable crash reporter to avoid GCC 15 breakpad compilation error
...
+# We replace Source0 with a deterministically-repacked tarball that is
+# byte-identical to the upstream one except for the removal of ff-inst.exe.
+# The repack is produced by base/comps/firefox/modify_source.sh, which
+# is reproducible — re-running it always yields the same SHA-256.
+#
+# When bumping the firefox version:
+#   1. Update --version in modify_source.sh (and UPSTREAM_SHA512).
+#   2. Re-run the script and copy the new SHA-256 into the source-files entry.
+#   3. Upload the new tarball to the modified-source lookaside (see script output).
+[[components.firefox.overlays]]
```
PawelWMS: Let's wait for [this `azldev` update](https://github.com/microsoft/azure-linux-dev-tools/pull/171) to go in, so you can replace the original upstream source and make this TOML config simpler.

base/comps/firefox/modify_source.sh
```diff
@@ -0,0 +1,132 @@
...
+  Output:        ${WORKDIR}/${OUTPUT_NAME}
+  SHA-256 file:  ${WORKDIR}/${OUTPUT_NAME}.sha256
+  SHA-512 file:  ${WORKDIR}/${OUTPUT_NAME}.sha512
+
+Next steps:
+  1. Upload to the AZL modified-source lookaside (pkgs_modified path, per Task 19805):
+       az storage blob upload \\
+         --account-name azltempstaginglookaside \\
+         --container-name repo \\
+         --name "pkgs_modified/firefox/${OUTPUT_NAME}/sha256/\$(awk '{print \$1}' ${OUTPUT_NAME}.sha256)/${OUTPUT_NAME}" \\
```
PawelWMS: This looks wrong. The `sources` file uses SHA-512 and here we're using SHA-256. In general, I don't think we need any SHA-256 in this script.

anphel31: Right - will fix

base/comps/firefox/firefox.comp.toml
anphel31: simplified with the new overlay support

base/comps/firefox/firefox.comp.toml
```diff
@@ -31,3 +31,36 @@ description = "Disable crash reporter to avoid GCC 15 breakpad compilation error
...
+# unmodified upstream tarball alongside our repacked one. This is handled
+# by `replace-upstream = true` on the source-files entry below.
+
+[[components.firefox.source-files]]
+filename = "firefox-148.0.source.tar.xz"
+hash = "c9c9f29fbd8f889bf3cf3d88776e1a62da7b2a65d386538d2bd048dd70caaaf8324adb5303a5fa9aa73c2cf6eb9f89cb4b34f9e67c4208d88636dd5376af90a9"
+hash-type = "SHA512"
+origin = { type = "download", uri = "https://azltempstaginglookaside.blob.core.windows.net/repo/pkgs_modified/firefox/firefox-148.0.source.tar.xz/sha512/c9c9f29fbd8f889bf3cf3d88776e1a62da7b2a65d386538d2bd048dd70caaaf8324adb5303a5fa9aa73c2cf6eb9f89cb4b34f9e67c4208d88636dd5376af90a9/firefox-148.0.source.tar.xz" }
+replace-upstream = true
+replace-reason = "AZL-repacked tarball with malware-flagged upstream test fixtures stripped (trips RPM signing pipeline); see modify_source.sh REMOVE_PATHS"
```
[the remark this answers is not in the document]
anphel31: `replace-upstream` and `replace-reason` were added to azldev, which is in a different repo, here: https://github.com/microsoft/azure-linux-dev-tools/pull/171

PawelWMS (approved):

base/comps/firefox/firefox.comp.toml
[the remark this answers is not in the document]
anphel31: the schema is out of date in this repo, but we will refresh it soon

specs/f/firefox/sources
```diff
@@ -4,4 +4,4 @@ SHA512 (wasi-sdk-25.tar.gz) = 1285981d26aa5eff27f08ed4b409909639ddcd62c94ee0cff7
 SHA512 (wasm-component-ld-vendor.tar.xz) = 707d942d31455ae0a4f68bf419fb09a20407b6747f831ca554dcd00925b7ea98ef4d03a8652b6d2ae54cf48d7ad15d85aa7eb8d0778ef66b89593eaa8b5c3465
 SHA512 (cbindgen-vendor.tar.xz) = bea420e66bdd1c7c944655dd3e01abd6e7d6ac4b245c7ee190f31d800f7786f21e5cae11715b479bf795f4369d18c40dc12df19e0b643664f2f78e5c8a681415
 SHA512 (firefox-langpacks-148.0-20260223.tar.xz) = 7e5d283e1a83787984e63901c915f6672eae48c38c5fd64b9f8055f154d83be77f76cb77de2048f2ea263353313a90eefa54b7c173533bec2db4dddf32436302
-SHA512 (firefox-148.0.source.tar.xz) = b0e862091f3a07a02890f6414e77b433893364a8beaf522d440e97ed0060c9b14bdb2fffdecdf12dca849efce8c57d95a534b23e04259d83a96ee8f29e078349
+SHA512 (firefox-148.0.source.tar.xz) = c9c9f29fbd8f889bf3cf3d88776e1a62da7b2a65d386538d2bd048dd70caaaf8324adb5303a5fa9aa73c2cf6eb9f89cb4b34f9e67c4208d88636dd5376af90a9
```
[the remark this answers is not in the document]
anphel31: This concern does not apply. Running `azldev comp render -p firefox` against the committed comp.toml on this branch logs the rewrite explicitly:
```
WRN Replacing upstream 'sources' entry from 'source-files' configuration
    component=firefox
    filename=firefox-148.0.source.tar.xz
    upstreamHash=b0e8620…
    newHash=c9c9f29f…
    reason="AZL-repacked tarball with malware-flagged upstream test fixtures..."
```
and reports STATUS: ok, CHANGED: false. diff between the pre- and post-render specs/f/firefox/sources is empty

base/comps/firefox/modify_source.sh
```diff
@@ -0,0 +1,147 @@
...
+# Determinism is achieved by:
+#   * tar --sort=name  (stable file order, independent of FS readdir order)
+#   * --owner=0 --group=0 --numeric-owner  (no host-specific uid/gid)
+#   * --mtime=@<fixed epoch>  (fixed mtime for every entry)
+#   * --format=gnu  (handles long paths > 100/255 bytes via ././@LongLink;
+#                    ustar would silently drop Firefox test files with very
+#                    long names, breaking the build. gnu is deterministic.)
+#   * xz -T 4 -9 --block-size=256MiB  (deterministic: output depends only on
+#                    these two values, not on host CPU count. Changing -T or
+#                    --block-size changes the SHA-256, so bump the hash in
```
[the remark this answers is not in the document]
anphel31: fixed

christopherco (changes requested):
There should be a release-bump in the render. Rerun the render operation and amend into the commit.

locks/firefox.lock
```diff
@@ -2,5 +2,5 @@
 version = 1
 import-commit = 'e6f01c53ca35f9fbad3bf45f227275550935b9d9'
 upstream-commit = 'e6f01c53ca35f9fbad3bf45f227275550935b9d9'
-input-fingerprint = 'sha256:ae7850d094cd84f8b90d81b5acf8b657aa444430faf1653bf75f74b914fc1685'
+input-fingerprint = 'sha256:2e481d978cad4da2a8a44f818ebd5efd024abd91460b61c39822e10112e27824'
```
christopherco: @dmcilvaney I think this type of change should trigger a release bump update. But when I run  `azldev comp render -p firefox`, I'm not seeing any render update difference. Can you look into this to see why a release increment isn't happening?

christopherco: oh nevermind, just saw in the toml that this is a manual release bump calculation....

christopherco (approved):

PR #17139, opened 2026-05-11
feat(repo): add scripts to synthesize and consume azl repodata

reubeno:
Adds three new tools under `scripts/repo/` for working with Azure Linux package repositories:

* `synthesize-repodata.py` — given one or more upstream repo prefixes (Standard Azure Linux Repo Layout: per-channel main/debuginfo/srpms sub-repos) and/or explicit per-repo overrides, synthesize a fresh set of per-destination repodata trees that route each package to its intended channel based on azldev component metadata. Local repo overrides take precedence over upstream when NEVRAs collide (CLI order is preserved). The output is a static directory tree of standard `createrepo_c` repodata with absolute upstream URLs in package locations, so the synthesized repodata can be served from anywhere without needing to mirror the RPM content.

* `dnf-with-azl-repos` — thin wrapper around `dnf` that probes one or more URL prefixes for the Standard Azure Linux Repo Layout, enables every sub-repo it discovers (silently skipping ones that don't exist), and execs `dnf` with those repos added on the command line.

* `_repo_layout.py` — shared definition of the Standard Azure Linux Repo Layout (channels, sub-repo kinds, per-kind URL template) consumed by both scripts.

### DIFF
--- a/scripts/repo/_repo_layout.py
+++ b/scripts/repo/_repo_layout.py
@@ -1,0 +1,54 @@
+# SPDX-License-Identifier: MIT
+"""Standard Azure Linux Repo Layout.
+
+Defines the fixed `channel x kind x arch` matrix that every published
+Azure Linux RPM tree follows. Both ``dnf-with-azl-repos`` (which
+discovers the layout under one or more URL prefixes) and
+``synthesize-repodata.py`` (which writes the layout from upstream
+inputs) consume :data:`SUBREPOS` directly.
+
+The matrix has six rows that have not changed for years; encoding it
+as a Python constant keeps the consumers trivial and avoids a JSON
+loader / validator layer that has to be kept in sync with the data.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+CHANNELS: tuple[str, ...] = ("base", "sdk")
+
+KIND_MAIN = "main"
+KIND_DEBUGINFO = "debuginfo"
+KIND_SRPMS = "srpms"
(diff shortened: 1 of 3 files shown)


binujp:
This looks good Reuben, thanks! Can you please add a few workflow examples and use cases? Interpreting command line option to what it does took some effort. That could be a me problem, but examples can help quite a bit.

binujp (approved):
Thank you! Given this already caught issues which would have otherwise led to much head-scratching, we should just merge it. This is functional enough to iterate based on usage feedback.

As I said in the comment, couple of example invocations will be great.

ddstreet:
I thought we were trying to keep tooling *out* of our source repo?

christopherco:
> I thought we were trying to keep tooling _out_ of our source repo?

Agree with the premise that we want to keep tooling out of our source repo here, but I do view this change specifically as net-positive in the immediate term to agree to have it in our `4.0` branch for the next ~1 month.

With the way folks are currently working with Copilot, many are launching from within this repo to set their current workspace, so having these scripts here directly allows Copilot to utilize these scripts to get feedback on any generated work created inline in the session, which IMO is a net benefit to getting higher probability of useful PRs to actually review.

That being said, we need to move to a better pattern (i.e., the copilot plugin model) so we can fully move all these tools out of this repo, and into a dedicated tooling repo. Given where we are at now, I will set the expectation that `scripts` folder should be removed in the next 1 month, and for us to move to this plugin-based model by then.

I'll be filing work items to move the tools from here into an appropriate location that is not this repo.

[AB#20061](https://dev.azure.com/mariner-org/mariner/_workitems/edit/20061)

christopherco (approved):

PR #17151, opened 2026-05-12
fix(pypy): pin upstream commit to 6ef55d4 (bumps to 7.3.21)

reubeno:
We started seeing build failures for pypy 7.3.20 when its cyclic build dependency (on N-1 version of itself) was being satisfied by 7.3.21 (Fedora 43 had bumped pypy to 7.3.21 upstream). This moves us forward to a later commit in Fedora 43, where pypy was upgraded to 7.3.21. This aligns versions and also addresses a segfault we were seeing during the build.

**_Validation_**: confirmed *stage-1* build passes on `0405` build target. Once we merge, we'll queue a non-scratch build against that target, and then proceed onward to ensure it builds in `azl4` build target.


ddstreetmicrosoft:
closing this as superseded by #17166

PR #17165, opened 2026-05-12
refactor: remove dependencies on ffmpeg

reubeno:
_This PR includes a series of commits focused on paving the way to remove `ffmpeg`._

#### Disable dependencies on ffmpeg and related packages via `build.without`, `BuildRequires` removal, etc.:

Components affected:

```
 avahi
 chromaprint
 gstreamer1-plugins-bad-free
 kf6-kfilemetadata
 libvncserver
 libzypp
 mivisionx
 openal-soft
 opencv
 plasma-desktop
 plasma-workspace
 qt6-qtmultimedia
 rubygem-activestorage
 weston
 ```

#### Removes components that depend on ffmpeg (directly or indirectly), but which we are not required to carry.

Components removed:

```
tigervnc
kpipewire
notcurses
mjpegtools
neatvnc
```

#### Validation

* Validated locally with `azldev component build`.
* Koji builds running now: CT workflow ID `5e604b82-5449-4145-ff6a-08deafafdf39`

### CODE REVIEW

File: base/comps/avahi/avahi.comp.toml [hunk 1]
```diff
@@ -0,0 +1,7 @@
+# Enable the upstream `bootstrap` build conditional. This disables the GTK,
+# Qt5, and Mono UI bindings (avahi-ui*, avahi-qt5*, avahi-sharp*), which we
+# do not ship. In particular, dropping avahi-ui-tools removes its hard
+# Requires on tigervnc, letting us drop tigervnc (and its ffmpeg deps)
+# from the distro entirely.
+[components.avahi]
+build.with = ["bootstrap"]
```
binujp, line 7: I like this, this pattern may apply to other such components too. Something for me to keep in mind.

binujp (commented):
Looks good Reuben, this is very carefully put together!

I had a take on expressing opinions via top-level overrides. If we do not want mingw then `without = ["mingw"]` becomes a top level override, same with ffmpeg.

christopherco (approved):

PR #17166, opened 2026-05-12
fix: remove pypy

reubeno:
`pypy` has been failing to build, due to unmet dependencies on an earlier build of `pypy` (or alternatively python 2.7). Since it's not in `base` and we only have one consumers, we're going to remove it -- and patch the consumer not to depend on it.

*Validation*: successful local builds + koji builds


ddstreetmicrosoft (approved):
lgtm

PR #17190, opened 2026-05-13
feat(vm-base): lower ext4 reserved block percentage to 1% for vm-base…

ddstreet:
… image rootfs

By default, ext4 filesystems are create with 5% reserved for root-only use. The reasoning for this is primarily for 2 reasons:

1) in case the filesystem fills up with user (non-root) content, if the filesystem is the root partition, the root user may need some amount of disk space simply to log in and manually perform operations to make space (e.g. rm files)

2) as an ext4 filesystem approaches 100% full, at some point fragmentation starts increasing sharply; the 5% buffer helps to avoid serious reductions in write performance due to excessive fragmentation.

This is controlled by the mkfs.ext4 -m parameter, which the man page summarizes as:

"Specify the percentage of the file system blocks reserved for the super-user.  This avoids fragmentation, and allows root-owned daemons, such as syslogd(8), to continue to function correctly after non-privileged processes are prevented from writing to the file system.  The default percentage is 5%."

However, the default of 5% may have been reasonable long ago when drives were much smaller, but with current drives allowing filesystems on the order of 10's of GB or more, reserving 5% is almost certainly far more than is actually needed.

For comparison with other distros in the Azure cloud, RHEL and Fedora use xfs and btrfs (respectively) for their rootfs, neither of which provide any user-configurable reserve space parameter, although both do internally reserve some amount of blocks. Ubuntu does use ext4, but for cloud images sets the reserve space to 0%.

For AzureLinux 4.0, a compromise between 0% and 5% appears to be 1%, which provides 99% of the filesystem space for use by end users, while also providing a small reserve buffer for root login recovery actions and fragmentation reduction.

This commit changes only the AzureLinux 4.0 base VM image root filesystem to reserve only 1%. This does not change the default reserve percentage for mkfs.ext4, so any other filesystems created will default to 5%.

Fixes: [AB#19696](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19696)


christopherco (approved):
:shipit:

PR #17207, opened 2026-05-13
llvm, llvm20: strip LLVM-bitcode sections from libLLVM*.a archives

PawelWMS:
## Summary

The `llvm-static` and `llvm20-static` sub-packages ship every `libLLVM*.a` archive built with clang's Thin-LTO + FatLTO objects. Each `.o` member therefore carries large embedded bitcode sections (`.llvmbc`, `.llvmcmd`, `.llvm.lto`, and defensively `.gnu.lto_*` / `.gnu.debuglto_*` in case any member happens to have been compiled with `gcc -flto`) on top of native code. Those sections are only useful for **re-linking with LTO**; the native linker ignores them when a consumer links a static archive into a regular (non-LTO) build. They typically 2–3× the compressed RPM-payload size of `llvm-static` and `llvm20-static`, which makes the resulting RPMs exceed the decompression budget of the automated package-signing pipeline (timeouts observed for all three failing archives: `llvm-static-*.aarch64.rpm`, `llvm20-static-*.aarch64.rpm`, `llvm20-static-*.x86_64.rpm`).

## Mechanism

Adds a `spec-append-lines` overlay to each of `base/comps/llvm/llvm.comp.toml` and `base/comps/llvm20/llvm20.comp.toml` that injects a small loop at the end of `%install`:

```bash
for _archive in %{buildroot}%{_libdir}/libLLVM*.a \
                %{buildroot}%{install_libdir}/libLLVM*.a; do
    [ -f "$_archive" ] || continue
    %{__objcopy} \
        --remove-section=.llvmbc \
        --remove-section=.llvmcmd \
        --remove-section=.llvm.lto \
        --remove-section='.gnu.lto_*' \
        --remove-section='.gnu.debuglto_*' \
        "$_archive" 2>/dev/null || :
done
```

**Path-glob rationale:** the default (non-compat) build runs `move_and_replace_with_symlinks %{buildroot}%{install_libdir} %{buildroot}%{_libdir}` in `%install`, after which the real `libLLVM*.a` files live under `%{buildroot}%{_libdir}/` and `%{install_libdir}` paths are symlinks back to them. The compat build keeps the archives under `%{install_libdir}`. The loop globs both paths and skips any non-regular file (i.e. dangling symlinks) so the overlay is layout-agnostic.

`objcopy` from binutils iterates over `ar` archive members automatically; one invocation per archive is enough. Wildcard section patterns (`--remove-section='.gnu.lto_*'`) are supported. Missing sections are no-ops, so over-specifying the bitcode/LTO section list is safe.

## Why this preserves consumer behaviour

`--remove-section` only drops named sections; ELF `.text` / `.data` / `.bss` / `.symtab` / `.strtab` / relocations are untouched. No exported symbol disappears — `nm` output is unchanged before / after. Native consumers (e.g. `spirv-llvm-translator`, the only in-distro consumer of `llvm-static`) keep linking normally. The only consumers that would notice the change are ones that explicitly drive `clang -flto` across the LLVM static libraries; there are no such consumers in Azure Linux today (verified via grep over `base/comps/**/*.toml` and `specs/**/*.spec`). `llvm20-static` has zero in-distro consumers, period.

## Validation

- Render: clean (`STATUS: ok` for both components). The appended loop lands at the tail of `%install`, right before the `%check` section directive.
- Lock: refreshed input-fingerprints for both `locks/llvm.lock` and `locks/llvm20.lock`.

## Estimated impact

~30–60% reduction in `libLLVM*.a` compressed size, which translates to a similar percentage drop in scanner-side decompression time — comfortably within the existing budget. Will be verified post-build by comparing `base/out/llvm-static-*.rpm` size before/after.


reubeno:
@PawelWMS -- I don't think we should prevent use of LTO when linking against these libraries. Is this really our option of last resort?

PawelWMS:
Closing — the underlying scanner issue was resolved by a package-signing-pipeline update, so this component-level workaround is no longer needed. `libLLVM*.a` ships unmodified again.

PR #17208, opened 2026-05-13
fix(mozjs128): serve modified Source0 with js/build only

PawelWMS:
Fixes: [AB#19867](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19867)

[Koji build](https://purple-grass-0478bae1e.7.azurestaticapps.net/workflow/6d65d1df-25ca-476c-b12c-08deb14d94b6)

## Summary

The `mozjs128` SRPM builds the SpiderMonkey JavaScript engine from the **full** upstream Firefox ESR source tarball (`firefox-128.11.0esr.source.tar.xz`, ~500 MB). The `%build` block only consumes `js/src/`, but the SRPM payload contains every Firefox subtree — including malware-scanner-tripping fixtures the automated package-signing pipeline rejects (`aes_archive.zip`, NSIS installer DLL block, the `lzma_sdk/google/test_data/encrypted{,_header}.7z` fixtures, the bundled `setuptools`/`pip` Windows PE launcher stubs, the `toolkit/components/telemetry/tests/unit/*.dll` PE blobs, and more).

This PR replaces Source0 with a deterministically-repacked Firefox source tarball that keeps **only** the subtrees `mozjs128`'s `%build` and `%install` actually use, and drops everything else.

## Changes

- **`base/comps/mozjs128/mozjs128.comp.toml`** (new) — dedicated component file with a single `[[components.mozjs128.source-files]]` block carrying:
  - `filename = "firefox-128.11.0esr.source.tar.xz"` (matches upstream so the spec's Source0 line does not need editing).
  - `hash` of the repacked tarball.
  - `origin.uri` pointing at the lookaside `repo` container under the `pkgs_modified/` prefix.
  - `replace-upstream = true` + `replace-reason = "..."` to swap the same-named upstream entry in the Fedora `sources` manifest in place.
- **`base/comps/mozjs128/modify_source.sh`** (new) — deterministic strip-and-repack script. Downloads the upstream Firefox ESR tarball, verifies its SHA-512, deletes everything outside the keep-list, repacks with `tar --sort=name --mtime=...` + `xz -T 1 -9 --block-size=...` (single-threaded for determinism), prints the new SHA-512 plus a ready-to-paste `az storage blob upload` command. Output lives at `base/build/work/scratch/mozjs128/`.
- **`base/comps/components.toml`** — inline `[components.mozjs128]` row removed (component is now defined in the dedicated file).
- **`specs/m/mozjs128/sources`** and **`specs/m/mozjs128/mozjs128.spec`** — regenerated.
- **`locks/mozjs128.lock`** — refreshed.

### Keep-list

`LICENSE`, `Cargo.toml`, `Cargo.lock`, `moz.configure`, `build/`, `config/`, `js/`, `mfbt/`, `memory/`, `mozglue/`, `python/mozbuild/`, `third_party/`. Plus an additional `js/`-internal strip of `js/src/fuzz-tests/`, `js/src/devtools/automation/variants/`, `js/src/octane/`, `js/src/ctypes/libffi/` (matches the existing `%prep`-time `rm -rf` deletions, but removes them from the SRPM payload rather than at build time).

`third_party/` is in the keep-list because `mozjs128.spec` does `chmod -x third_party/rust/bumpalo/src/lib.rs` in `%prep` (would fail under `set -e` if stripped), and the SpiderMonkey cargo build under `js/src/` links against vendored Rust crates from `third_party/rust/`.

## Validation

- Render: clean.
- Lock: refreshed.
- Modified-tarball SHA-512: `4cec711d46502beea27d0e96e95c1de70a53139bd9c71dcc5f476815a1b3aa0bab3613f4883c33707938801660d74463b112817c4a68dc51993e2a0ad558d19f` (deterministic).
- Modified tarball uploaded to the lookaside blob at the path the comp.toml `origin.uri` references.


PawelWMS:
**Force-pushed `db85f812f0`** to fix the `%build` configure failure from the previous CT run.

Symptom in the failing build log:
```
+ ./configure --build=... --with-system-icu ...
python3: can't open file '/builddir/build/BUILD/mozjs128-128.11.0-build/firefox-128.11.0/js/src/./../../configure.py':
        [Errno 2] No such file or directory
```

Root cause: `js/src/configure` execs into the top-level `configure.py` (shared between the Firefox and SpiderMonkey configure flows), but that file lives at the *root* of the extracted tree (`firefox-128.11.0/configure.py`), not inside `js/`. The previous keep-list only kept top-level *directories*, so the single-file `configure.py` was being stripped.

Fix:
- Added `configure.py` to `KEEP_TOP` in `base/comps/mozjs128/modify_source.sh`.
- Re-uploaded modified tarball to lookaside (new content length 171,437,828 bytes).
- New modified-tarball SHA-512: `194f8288744c63f3d6efa0d0fafedc07ac9ace31dfcc4082192b191e2b653979d20b6ad2873ad1a2f7c80e8cc271d883a977773bc2639377ce2a16cb86721c1e`. Both the `hash` field and the `$hash` segment of `origin.uri` in `mozjs128.comp.toml` are updated to match.
- Lock + rendered spec refreshed; render audit-WARN confirms the swap from upstream `80af64c1...092d279` to the new `194f8288...86721c1e`.
- Single GPG-signed commit; CI re-running.

PawelWMS:
**Force-pushed `0a0639d096`** to fix the third `%build` failure from CT.

Symptom in the failing build log:
```
checking for a shell... Traceback (most recent call last):
  ...
  File "/builddir/build/BUILD/mozjs128-128.11.0-build/firefox-128.11.0/build/moz.configure/util.configure", line 171, in <module>
    @imports(_from="mozfile", _import="which")
ModuleNotFoundError: No module named 'mozfile'
```

Root cause: `build/moz.configure/util.configure`'s `find_program` helper imports `mozfile.which`, but `mozfile` lives at `testing/mozbase/mozfile/` — and `testing/` was not in `KEEP_TOP`. The python-virtualenv-setup step that `mach` runs adds every `pth:testing/mozbase/*` line from `python/sites/mach.txt` to `sys.path`, so those modules MUST exist at extraction time. The full `testing/` tree is ~830 MB and carries fuzzer corpora / crash fixtures (mochitest, marionette, raptor, web-platform) that trip the scanner; `testing/mozbase` itself is only ~12 MB of plain Python with no such content.

Fix: added a `NESTED_KEEP=(testing/mozbase)` array to `modify_source.sh` plus a restore step that re-extracts those nested paths from the upstream tarball after the top-level strip. The rest of `testing/` stays out.

I also walked the spec one more time to double-check we're not stripping anything else that's referenced:
- `%prep` `cp LICENSE js/src/` — `LICENSE` already kept.
- `%prep` `cp %{SOURCE1} js/src/` — `js/src/` already kept.
- `%prep` `rm -rf modules/zlib` — `modules/` is stripped, so this `rm -rf` is harmless (no-op on a non-existent dir).
- `%prep` `chmod -x third_party/rust/bumpalo/src/lib.rs` — `third_party/` already kept.
- `%prep` `sed -i 's/icu-i18n/icu-uc &/' js/moz.configure` — `js/` already kept.

New modified-tarball SHA-512: `829459b9da10281106278759b763d872b82cee413f9cfbc7510a4c435336b3a7bade842ace2be1542492f06b481cbcdfc7cbcba0913f3e51973f644c9e8eba47`. Both `hash` and the `$hash` segment of `origin.uri` updated; new tarball uploaded to the matching lookaside path; lock + spec re-rendered; render audit-WARN confirms the swap. Single GPG-signed commit; CI re-running.

PawelWMS:
Superseded by #17368 (merged 2026-05-21), which removed `mozjs128` from Azure Linux 4.0 entirely along with the Cinnamon dependency cluster that was its only consumer. The targeted Source0-stripping approach in this PR is no longer needed.

PR #17215, opened 2026-05-14
refactor: adjust azurelinux macro (major version only)

reubeno:
For consistency with other distros' macros, we:

* Change the definition of the azurelinux macro only to include major version (i.e. 4).
* Remove the azl4 macro; we should use the azurelinux macro instead.

A compensating change is made to openssl.

--
Fixes: [AB#19815](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19815)

### DIFF
--- a/base/comps/azurelinux-release/azurelinux-release.spec
+++ b/base/comps/azurelinux-release/azurelinux-release.spec
@@ -36,7 +36,7 @@ Summary:        Azure Linux release files
 Name:           azurelinux-release
 Version:        4.0
 # TODO(azl): Review whether we can move back to autorelease (with conditional -p)
-Release:        16%{?dist}
+Release:        17%{?dist}
 License:        MIT
 URL:            https://aka.ms/azurelinux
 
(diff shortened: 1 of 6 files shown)


ddstreetmicrosoft (changes requested):
issue: looks like the `azurelinux-release` package needs a release bump

reubeno:
> issue: looks like the `azurelinux-release` package needs a release bump

Oof, I've gotten spoiled on autorelease specs; thanks for catching it. I'll fix that.

(We should enable autorelease, but in a separate change.)

base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -373,8 +373,7 @@ cat >> %{buildroot}%{_rpmconfigdir}/macros.d/macros.dist << EOF
 # dist macros.
 
 %%__bootstrap         ~bootstrap
-%%azurelinux          %{dist_version}
-%%azl4                1
```
ddstreetmicrosoft, line 377: question: are you sure we should remove `%azl4 1`?

Fedora has:
`%fedora 43`, `%fc43 1`, and `%distcore .fc43`
RHEL has:
`%rhel 10`, `%el10 1`, and `%distcore .el10`
it seems like we should have:
`%azurelinux 4`, `%azl4 1`, and `%distcore .azl4`

ddstreetmicrosoft (commented):
Looks ok other than removing `%azl4`, are you sure we should really remove it?

base/comps/azurelinux-release/azurelinux-release.spec
reubeno: That's great feedback; we can go with your suggestion.

ddstreetmicrosoft (approved):

PR #17227, opened 2026-05-14
fix(pipewire): disable opus codec support

binujp:
## Summary

Disable opus and bluez5-codec-opus in pipewire's meson build config. Opus requires CELA review per Microsoft codec guidance.

## Changes

- Added spec-search-replace overlay to inject `-D opus=disabled -D bluez5-codec-opus=disabled` into the meson configuration
- pipewire's meson.build requires opus as a dependency (default: auto) — without this, builds fail when opus-devel is unavailable

## Testing

- Built pipewire successfully
- Verified pipewire-libs has no opus dependency (`rpm -qpR` shows no opus)
- Smoke-tested in mock chroot: `pipewire --version` works correctly

Part of the opus codec removal effort.


base/comps/pipewire/pipewire.comp.toml
```diff
@@ -1,3 +1,17 @@
 [components.pipewire]
 # Release: %{baserelease}%{?snapdate:.%{snapdate}git%{shortcommit}}%{?dist}
 release = { calculation = "manual" }
+
+[[components.pipewire.overlays]]
+type = "spec-search-replace"
+description = "Increment baserelease for opus removal rebuild"
+regex = '%global baserelease 1'
```
christopherco: Given this is a manual release bump method, rather than hardcode the `baserelease` number which will stop matching if upstream Fedora bumps `baserelease`, use something like this - https://github.com/microsoft/azurelinux/commit/a13a1005d9535878ba9c21796c6f1fab9d8a7c94 - to create `azl_release` to track the number of times we add our Azure Linux adjustments, and then use the RPM arithmetic operation to incorporate `%[%{baserelease}+%{azl_release}]` into the spec's `Release` macro

christopherco (changes requested):

christopherco (approved):

PR #17253, opened 2026-05-15
chore(bash): pin from rawhide to Fedora 44

christopherco:
bash was temporarily pinned to Fedora rawhide for build fixes that hadn't yet propagated to a stable Fedora branch. Those fixes are now available in Fedora 44 dist-git, so move the pin from the rawhide branch to the f44 branch and lock to an explicit upstream commit ('a6bcc67…') for deterministic resolution.

This keeps the dedicated bash.comp.toml in place so the description field can carry the rationale for the pin, and so the explicit upstream-commit travels with the package rather than the lock alone.

The resulting RPM is bash-5.3.9-5.azl4 (vs. -4.azl4 previously) — the upstream commit is unchanged, only the AZL Release counter bumps from the lock-fingerprint change.

Fixes: [AB#19413](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19413)

### DIFF
--- a/base/comps/bash/bash.comp.toml
+++ b/base/comps/bash/bash.comp.toml
@@ -1,3 +1,4 @@
 [components.bash]
-# Need build fixes in rawhide
-spec = { type = "upstream", upstream-distro = { name = "fedora", version = "rawhide" } }
+# Pin to Fedora 44 dist-git head as of the AZL4 Public Preview cutoff.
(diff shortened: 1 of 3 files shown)


christopherco:
Supersedes #17173

reubeno (approved):

PR #17260, opened 2026-05-15
fix: Fix 5 components with non-first-parent upstream commits

Tonisal-byte:
Pin systemtap, linux-sgx, pcp, openscap, yarnpkg
to newer f43 HEAD commits to supersed previously
published versions that were built from non-first-parent commits.

- systemtap: 5.4-3 -> 5.5-1
- linux-sgx: 2.27 -> 2.28
- pcp: 7.1.0-6 -> 7.1.0-8
- openscap: 1.4.3-4 -> 1.4.4-4
- yarnpkg: 1.22.22-17 -> 1.22.22-18


christopherco (changes requested):
Please split these into individual commits in the PR, so we can rebase-merge them while preserving the individual commits (also changelog messages inherit the commit title that modifies the lock file).

Also I'm just curious on the background behind why we're moving these 5 commits forward.

Tonisal-byte:
> Also I'm just curious on the background behind why we're moving these 5 commits forward.

The reason we are moving these 5 commits forward has to do with a bug in `azldev` that picks commits for the lock files that aren't on our selected upstream Fedora branch. After looking through all the lock files, only these 5 components had pinned a commit not on the `f43` lineage. This PR against `azldev` should fix the problem and prevent incorrect commits from being pinned in the future [[here](https://github.com/microsoft/azure-linux-dev-tools/pull/192)].

Continuing to carry these commits will result in errors when we update the component, as already observed in this commit to `systemtap` earlier https://github.com/microsoft/azurelinux/commit/461439f6615a107c4dbdfcac1a464f6bf962b7be.

christopherco (changes requested):
Thanks for making the updates. Looks like these aren't building in CT due to sources. I'll see if I can handle that.

christopherco (approved):

christopherco:
I've fixed the sources and everything looks good

PR #17272, opened 2026-05-15
fix: remove svt-av1 and disable svt-av1 in dependents

binujp:
## Summary

Remove the svt-av1 component from the distro and explicitly disable svt-av1 support in dependent packages.

## Changes

- **svt-av1**: Removed from components, publish channels, lock file, and rendered specs
- **gstreamer1-plugins-bad-free**: Added `svtav1` to `build.without` — the `%bcond svtav1` defaults ON via `%{defined fedora}` which is true in the AZL build environment, so the plugin was being built and shipping `libgstsvtav1.so` with a dependency on `libSvtAv1Enc.so.3`
- **libavif**: Created dedicated `libavif.comp.toml` with `build.without = ["svt"]` — the `svt` bcond defaults ON in non-bootstrap builds

## Testing

- Built libavif — verified all 3 RPMs (libavif, libavif-tools, libavif-devel) have no svt-av1 dependency
- gstreamer1-plugins-bad-free build pending (overlay adds `svtav1` to existing `build.without` list using the same pattern as the other disabled bconds)

Part of the codec removal effort.


christopherco (dismissed):
Underlying changes generally look fine. Please break up the commits into separate ones:
1. gstreamer1-plugins-bad-free
2. libavif
3. removal of svt-av1
and have relevant commit messages for each.

This way, the rendered commit messages in the individual spec files will pertain to the changes performed onto the specific package

reubeno (approved):

PR #17289, opened 2026-05-15
fix: Remove dav1d, rust-dav1d, rust-dav1d-sys from distro

trungams:
Remove the dav1d AV1 decoder and its Rust bindings. Dependents updated:
- libavif: switch from dav1d to aom decoder
- librsvg2: disable AVIF support (already disabled for RHEL)

Resolves [AB#19414](https://dev.azure.com/mariner-org/mariner/_workitems/edit/19414)
Resolves [AB#19709](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19709)


christopherco (changes requested):
Changes themselves look good. Please break the changes into separate smaller commits:
- libavif changes
- librsvg2 changes
- rust-rav1e changes
- remove dav1d, rust-dav1d, rustdav1d-sys

Doing it this way will ensure the autochangelog render describes the alterations made to that specific package.

base/comps/components-publish-channels.toml
```diff
@@ -213,7 +213,6 @@ components = [
     "cyrus-imapd",
     "cyrus-sasl",
     "Cython",
-    "dav1d",
```
christopherco, line 215: This component change needs to be in the `dav1d` removal commit

christopherco (changes requested):
Thanks for making the adjustments! One minor remaining adjustment needed, then we should be good to merge.

base/comps/components-publish-channels.toml
trungams: updated

christopherco (approved):

PR #17295, opened 2026-05-15
fix(kf6-karchive): strip scanner-flagged autotest fixtures from source tarball

anphel31:
## Summary

Remove files `autotests/data/zip64_extra_zip64_size_first.zip.gz` and `password_protected.7z` from the source tarball.

The autotests are not built or run in our spec:
- No `%check` block.
- Upstream's CMake gates them on `BUILD_TESTING=ON`, which we don't set.

Stripping the fixture is functionally inert; the published RPMs are unchanged.

Fixes: [AB#19875](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19875)

## Modified-tarball SHA-512

Reproducible by anyone with the same upstream input — re-run
`bash base/comps/kf6-karchive/modify_source.sh` to verify.

## Validation

- Render: clean (`STATUS: ok`).
- Repack determinism: confirmed across two local runs (byte-identical output).
- Modified tarball uploaded to the lookaside path the comp.toml `origin.uri`
  references.
- Test [build](https://purple-grass-0478bae1e.7.azurestaticapps.net/workflow/5053a4f6-57b2-4b83-8cf9-08deb5422427)
- GPG test signing confirmed


christopherco (approved):

base/comps/components.toml
```diff
@@ -1552,7 +1552,7 @@ includes = ["**/*.comp.toml", "component-check-disablement.toml", "component-min
 [components.kf6-attica]
 [components.kf6-baloo]
 [components.kf6-frameworkintegration]
-[components.kf6-karchive]
+# kf6-karchive: dedicated comp.toml at base/comps/kf6-karchive/
```
christopherco: This isn't needed and clutters the `components.toml`.

christopherco (dismissed):

base/comps/components.toml
anphel31: Removed

reubeno (approved):

PR #17318, opened 2026-05-16
ci: adopt lychee for docs validation

reubeno:
Adopts [lychee](https://github.com/lycheeverse/lychee) for validation of links in .md files and similar. We choose to run in offline mode for now, and only validate *local* files. This is a more conservative option for that avoids trying to access arbitrary internet locations. It will still catch broken inter-.md links in docs, readmes, etc.

> NOTE: We also disable review of rendered content under `specs`.

A sample local run (without verbosity) looks like this:

image

For reference, `azldev` also uses `lychee` in CI workflows.

### DIFF
--- a/lychee.toml
+++ b/lychee.toml
@@ -1,0 +1,4 @@
+exclude_path = [
+   # Exclude validation of links in rendered specs/content.
+   "(^|/)specs/" 
+]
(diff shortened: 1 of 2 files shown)


lychee.toml
```diff
@@ -0,0 +1,4 @@
+exclude_path = [
+   # Exclude validation of links in rendered specs/content.
+   "specs/.*"
```
christopherco: suggestion(non-blocking): consider anchoring the regex to avoid future possibility of overmatching the exclude. Something like `(^|/)specs/"` should work

christopherco (approved):

PR #17327, opened 2026-05-17
remove nv-codec-headers

christopherco:
## Summary

Removes `nv-codec-headers` from the distro. It ships only FFmpeg's NVIDIA
NVENC/NVDEC header shims and has no remaining consumer in AZL — the sole
user was `gnome-remote-desktop`'s RDP backend (NVENC hardware-encode path),
which AZL builds with `build.without = ["rdp"]` because we don't ship opus.

## Changes

Split into two bisectable commits:

1. **`fix(gnome-remote-desktop): drop pkgconfig(ffnvcodec) BR`**
   Upstream gnome-remote-desktop carries an unconditional
   `BuildRequires: pkgconfig(ffnvcodec)` outside the `%if %{with rdp}`
   block, even though the dependency is only used by the RDP backend.
   Adds a `spec-remove-tag` overlay that strips the line so the package
   can build after nv-codec-headers is gone.

2. **`chore(nv-codec-headers): remove component`**
   Removes the component definition (`base/comps/components.toml`),
   lock file (`locks/nv-codec-headers.lock`), and rendered spec
   (`specs/n/nv-codec-headers/`).

Fixes: [AB#19728](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/19728)


reubeno (approved):

martian56 (commented):
Nice work!

PR #17340, opened 2026-05-18
fix(mozjs128): drop scanner-flagged aes_archive.zip

PawelWMS:
## Summary

Drop the single scanner-flagged file `firefox-128.11.0/third_party/rust/zip/tests/data/aes_archive.zip` from mozjs128's Source0 via a downstream `modify_source.sh` repack. The fixture is an AES-encrypted ZIP from the cargo-vendored Rust `zip` crate's test suite that trips the automated package-signing pipeline's deep scanner with a benign encrypted-archive verdict, failing the SRPM signing step. It is never read at AZL build time — mozjs128 only consumes `js/src/`.

## Files

- `base/comps/components.toml` — remove the inline `[components.mozjs128]` row (the dedicated component file is auto-included by the `**/*.comp.toml` glob).
- `base/comps/mozjs128/mozjs128.comp.toml` — declare `[[components.mozjs128.source-files]]` with `replace-upstream = true` pointing at the locally-modified tarball in `pkgs_modified/mozjs128/` (SHA-512 `39141838…370db2d`).
- `base/comps/mozjs128/modify_source.sh` — deterministic repack script: download upstream `.tar.xz`, verify published SHA-512, extract, delete the one fixture, repack via `tar --sort=name --mtime --owner=0 --group=0 --numeric-owner | xz -T1 -9e`. Single-threaded xz keeps the output hash host-CPU independent.
- `locks/mozjs128.lock` — refreshed `input-fingerprint`.
- `specs/m/mozjs128/{mozjs128.spec,sources}` — re-rendered: bumps `release_number` to 10, swaps the `sources` SHA-512 to the modified tarball hash, adds a `%changelog` entry.

## Approach

Single-file strike (`aes_archive.zip` only) rather than a broader subtree strip — the only confirmed scanner detection on this SRPM is the single hit on this exact file. A minimal repack keeps the diff to a 4 MB delta against the upstream tarball and avoids breaking any of mozjs128's reverse-dependencies (cjs + the 8 cinnamon-* components).

Supersedes the broader strip on #17208 (kept-only-`js/` plus restored subtrees), which the build pipeline rejected.


PawelWMS:
Superseded by #17368 (merged 2026-05-21), which removed `mozjs128` from Azure Linux 4.0 entirely along with the Cinnamon dependency cluster that was its only consumer. The targeted Source0-stripping approach in this PR is no longer needed.

Issue #15880, opened 2026-02-17
Azure Linux 3.0 - Expected behavior in FIPS mode


vermajit:
We are using AKS with Azure Linux 3.0 node pool.  Our containers are based on AL 3.0 as well. Our applications are mix of Golang and .NET. 
**I am trying to understand all the changes needed at all layers of the software stack for FIPS compliance.** 

I updated the nodepool to enable FIPS (via enable-fips-image flag). I verified the following inside the one of the nodes and everything seemed as per documentation

```
kubectl debug node/aks-roxypool-27210205-vmss000000 -it --image=busybox
```

```
/ # cat /host/etc/os-release 
NAME="Microsoft Azure Linux"
VERSION="3.0.20260204"
ID=azurelinux
VERSION_ID="3.0"
PRETTY_NAME="Microsoft Azure Linux 3.0"
ANSI_COLOR="1;34"
HOME_URL="https://aka.ms/azurelinux"
BUG_REPORT_URL="https://aka.ms/azurelinux"
SUPPORT_URL="https://aka.ms/azurelinux"
```

```
/ # cat /proc/sys/crypto/fips_enabled
1
```

```
/ # chroot /host /usr/bin/openssl version -a
OpenSSL 3.3.5 30 Sep 2025 (Library: OpenSSL 3.3.5 30 Sep 2025)
built on: Sun Feb  1 11:05:57 2026 UTC
platform: linux-x86_64
options:  bn(64,64)
[...]

PR #17341, opened 2026-05-18
add openssl fips provider

tobiasb-ms:
This change adds the FIPS provider for `openssl`. Modeled after RHEL, it is built in a separate component -- `openssl-fips-provider` so we can pin it separately from the main `openssl` component. This is important because once certified, we will almost never want to rev `openssl-fips-provider`. `openssl` then takes a run-time dependency on `openssl-fips-provider`, so it should always have the FIPS provider available.

That said, `openssl-fips-provider` is still based on the same source -- the `openssl` upstream. It builds very similarly to `openssl`, but then instead of deleting the `fips`-related things it deletes everything else.

We diverged significantly from RHEL on how to implement the FIPS provider -- going way, way closer to an unmodified `openssl` version. Our patches are now almost all outside of the FIPS boundary, and we package it the way `openssl` wants it packaged -- with the HMAC in a config file rather than embedded.

I've build and run these locally and everything behaves the same as Fedora. I've also built on the build server.


lambertpandian:
Hi @tobiasb-ms,

I see this PR is "Closed with unmerged commits." Could you please share why this approach was not merged?
Per your comment in issue #15880, Azure Linux 3.0's SymCrypt provider is FIPS-validated but doesn't disable non-certifiable functions like MD5. This PR appeared to add a standard FIPS provider that would enforce algorithm restrictions and we need this restriction for FIPS compliance.

PR #17342, opened 2026-05-18
fix(rubygem-pdf-reader): strip 47 PDF test fixtures

PawelWMS:
[Koji build](https://purple-grass-0478bae1e.7.azurestaticapps.net/workflow/160313a2-6404-42ad-8cf5-08deb5422427).

## Summary

Remove 47 pathological PDF test fixtures from the upstream `spec/data/` directory in Source1 (`pdf-reader-2.4.2-spec.txz`) to avoid scan failures on the SRPM. All fixtures are benign upstream regression PDFs (deliberately-malformed structures, encrypted-cipher variants) that are not consumed at AZL runtime — `%files` excludes `spec/` entirely; the fixtures exist only during `%check`'s `rspec` run.

## How

The companion helper-file patch in `modify_source.sh` makes `pdf_spec_file()` in `spec/support/reader_spec_helper.rb` call rspec's `skip` only for the 47 removed filenames; it still raises `ArgumentError` for any other missing fixture. So `%check` continues running — the 47 affected `context` blocks in `spec/integration_spec.rb` get gracefully marked pending, and the remaining ~60 PDFs in `spec/data/` continue to exercise the regression suite.

The repack is byte-deterministic (`tar --sort=name --mtime --owner=0 --group=0 --numeric-owner | xz -T1 -9e`) so the resulting SHA-512 is reproducible from a fresh clone.

## Files

- `base/comps/components.toml` — remove the inline `[components.rubygem-pdf-reader]` row (dedicated component file is auto-included by the `**/*.comp.toml` glob).
- `base/comps/rubygem-pdf-reader/rubygem-pdf-reader.comp.toml` — dedicated component file with a `[[...source-files]]` block (`replace-upstream = true`) pointing at the locally-modified Source1 in `pkgs_modified/rubygem-pdf-reader/` (SHA-512 `49de5d0e…51fd708`).
- `base/comps/rubygem-pdf-reader/modify_source.sh` — the deterministic repack + helper-patch script.
- `locks/rubygem-pdf-reader.lock` — refreshed `input-fingerprint` and `upstream-commit`.
- `specs/r/rubygem-pdf-reader/{rubygem-pdf-reader.spec,sources}` — re-rendered: release bumped to 12, `sources` SHA-512 updated.

## Build impact

`%check` continues to run. 47 contexts in `spec/integration_spec.rb` become rspec "pending" instead of erroring out; the rest of the suite exercises its code paths unchanged.


base/comps/rubygem-pdf-reader/modify_source.sh
```diff
@@ -0,0 +1,215 @@
...
+    spec/data/type3_font_with_rare_font_matrix.pdf
+    spec/data/zapf.pdf
+    spec/data/zeroed_xref_entry.pdf
+    spec/data/zlib_stream_issue.pdf
+)
+
+# Helper file whose missing-fixture branch is patched from
+# `raise ArgumentError, ...` to `skip "..."` so rspec gracefully
+# marks affected examples pending instead of failing them.
+HELPER_FILE="spec/support/reader_spec_helper.rb"
```
reubeno, line 82: Per our discussion, changes to this file should really be applied using a proper visible patch that's not buried in this removal script. I think we can take this change for now, but I'd like you to file a tracking bug on it for each such script that does this sort of thing.

reubeno (changes requested):
Left a comment. Could you please also update the commit and PR to use conventional commit syntax, as described in CONTRIBUTING.md, please?

reubeno (approved):

PR #17343, opened 2026-05-18
fix(yara): strip 3 more OSS-Fuzz corpus fixtures

PawelWMS:
## Summary

Follow-up to PR #17097: a fresh scan on `yara-4.5.4-5.azl4.src.rpm` flagged three additional benign oss-fuzz seed-corpus inputs that the original fix did not catch. This change appends them to `STRIP_PATHS` in `base/comps/yara/modify_source.sh`.

Newly stripped paths (all under `tests/oss-fuzz/`):

- `dex_fuzzer_corpus/27fb31059503773723597edb875c937af971a6c15f91aac8c03c1fbdfa9e918c`
- `dex_fuzzer_corpus/b343d1058063e6e4b652ccf0589f93d0dbb6b092960e4aebc3c3c58894831359`
- `dotnet_fuzzer_corpus/buggy_stream_names`

Heuristics tripping on them: generic Android-packer detection on the DEX entries, and a high-entropy-entrypoint packer detection on the .NET entry.

## Why no further `Makefile.am` edit

These three files live under `tests/oss-fuzz/*_fuzzer_corpus/` directories. The merged PR #17097's `modify_source.sh` already documents that these directories are **not referenced from upstream's `Makefile.am`** — they are consumed only by libFuzzer harnesses, not by the autotools `make check` driver. So the existing `Makefile.am` edit that drops `test-pe` (which is for the `tests/data/` PE fixtures) remains correct and complete; no companion edit is needed for the fuzzer-corpus additions.

## Files

- `base/comps/yara/modify_source.sh` — append three new entries to `STRIP_PATHS` (now sorted alphabetically); rewrite both comment blocks (the file-level `Background` header and the one above `STRIP_PATHS`) to reflect the full current strip scope: oss-fuzz seed corpora under `tests/oss-fuzz/`, plus four hash-named PE fixtures under `tests/data/` (three SHA-256-named plus one `.upx`-suffixed sibling).
- `base/comps/yara/yara.comp.toml` — refresh `hash` + `origin.uri` to new modified-tarball SHA-512 (`94d4aab0…d7b646b`); slim `replace-reason` to a short pointer (detailed list lives in `modify_source.sh`).
- `locks/yara.lock` — refreshed `input-fingerprint`.
- `specs/y/yara/{yara.spec,sources}` — re-rendered: release bumped to 6, sources SHA-512 swapped to the modified-tarball hash, new `%changelog` entry.

Fixes: [AB#20076](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20076)


reubeno (changes requested):
Change looks fine; could you update to use conventional commits in the commit and PR, please?

reubeno (approved):

PR #17344, opened 2026-05-19
feat(kernel): Upgrade to 6.18.31.1

rlmenge:
Updates the kernel and kernel-headers components to version 6.18.31.1, sourced from CBL-Mariner-Linux-Kernel rolling-lts/azl4/6.18.31.1. Also disables several unnecessary kernel configs to reduce attack surface.

Note this fixes [AB#20078](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20078)  but also introduces a new bug (https://dev.azure.com/mariner-org/mariner/_workitems/edit/20082) which will be fixed in a follow-up pr

Changes:

1. Version bump to 6.18.31.1
- Updated kernel and kernel-headers version and tarball fields
- Updated SHA512 hash for the new tarball

2. Disable ESP-in-TCP encapsulation 
- Disabled CONFIG_INET_ESPINTCP, CONFIG_INET6_ESPINTCP, CONFIG_XFRM_ESPINTCP on both aarch64 and x86_64
- These configs enable ESP-in-TCP encapsulation (RFC 8229)

3. Disable AF_RXRPC and AFS_FS
- Disabled CONFIG_AF_RXRPC and CONFIG_AFS_FS on both aarch64 and x86_64

Testing:
 - Completed a local build of the kernel upgrade to `6.18.31.1`.
 - Completed Koji builds for the updated packages:
   - `kernel-headers`: task `2489691`
   - `kernel`: task `2489728`
 - Boot-tested the upgraded kernel in Azure VMs on both supported architectures:
   - `x86_64`
   - `aarch64`
 - Confirmed the upgraded kernel boots successfully on both VM types.


specs/k/kernel-headers/kernel-headers.spec
```diff
@@ -151,7 +151,10 @@ done
 
 %changelog
 ## START: Generated by rpmautospec
-* Thu May 14 2026 Rachel Menge <rachelmenge@microsoft.com> - 6.18.29-1
+* Mon May 18 2026 Rachel Menge <rachelmenge@microsoft.com> - 6.18.31-1
+- feat(kernel): update kernel and kernel-headers to 6.18.31.1
+
+* Thu May 14 2026 Rachel Menge <rachelmenge@microsoft.com> - 6.18.3-4
```
christopherco, line 157: issue: Render looks broken? It's missing 6.18.29 commit entry and made up this one.

rlmenge: This is a known bug with the kernel-headers spec. It can only render the latest commit and the past ones will be for fedora's spec version

christopherco (changes requested):
Each commit in the PR is causing a change to the input-fingerprint, and since both `kernel` and `kernel-headers` are manual release calculation, this means each of these commits should increment the `azl_pkgrelease` value

christopherco (approved):

PR #17352, opened 2026-05-19
docs(instructions): document AZL component-modification workflow conventions

PawelWMS:
This PR captures the conventions developed during a long-running multi-day session of remediating package-signing/scan failures at the component level (no allow-listing). The goal is to make those conventions discoverable to future Copilot sessions and agents, not to introduce new tooling.

### 1. `replace-upstream` source-override pattern (`comp-toml.instructions.md`)
Documents the single-step pattern for serving a locally-modified upstream tarball under the same filename: one `[[components.<name>.source-files]]` block with `replace-upstream = true` and a `replace-reason = "..."` string. No separate `file-remove` overlay is needed; `azldev`'s render step emits an audit WARN log naming the override and the from/to SHA-512 pair.

Covers field semantics, the lookaside URL pattern, the requirement that `hash` and the `$hash` segment of `origin.uri` stay in sync, `replace-reason` style rules, and the file-organization rules (delete inline `components.toml` entries when moving to a dedicated `*.comp.toml`; no tombstone comments).

### 2. `skill-modify-source` skill
New skill at `.github/skills/skill-modify-source/SKILL.md` documenting the canonical `modify_source.sh` pattern. The lookaside URL embeds the SHA-512 of the served file, so the repack **must** be byte-deterministic. Covers:

- Required `tar` flags (`--sort=name`, `--owner=0`, `--group=0`, `--numeric-owner`, `--mtime='@<epoch>'`, `--format=gnu`, `LC_ALL=C`).
- Required `xz` flags (`-T 1`, `-9`).
- Mandatory upstream SHA-512 verification before extract.
- Strip-list vs keep-list strategies with list-discipline rules.
- Output reporting contract (path, hash, ready-to-paste blob-upload command).
- Determinism self-test recipe (two runs must produce byte-identical output).
- Anti-patterns to avoid (multi-threaded `xz` without `--block-size`, `mktemp` for output filename, `gzip` without `-n`, `find -newer`, etc.).
- Canonical ~40-line skeleton showing the whole pattern end-to-end.

### 3. `skill-signing-failure-remediation` skill
New skill at `.github/skills/skill-signing-failure-remediation/SKILL.md` documenting the decision tree:

- **(a)** No reverse dependencies → drop the component outright (separate PR).
- **(b)** Offending artefact lives in a test-only or otherwise-optional subpackage → use `spec-remove-subpackage` (or `build.without` if a clean bcond exists).
- **(c)** Reverse deps require the component to keep building → strip subtrees from Source0 via the `replace-upstream` pattern with a `modify_source.sh` script (see `skill-modify-source`).

Always run a reverse-dependency scan before choosing (a) or (b).

### 4. Render/lock cycle ordering (`copilot-instructions.md`, `skill-update-component`)
Canonical sequence: edit `comp.toml` → `azldev comp update -p <name>` → commit (toml + lock) → `azldev comp render -p <name>` → amend (rendered spec). Keeps the lock fingerprint, `%changelog`, and `Release:` in sync within a single commit. The existing `skill-update-component` had this pattern; this PR clarifies the *commit order* explicitly.

### 5. GPG signing (`copilot-instructions.md`)
Every commit MUST be GPG-signed. Verify with `git log -1 --show-signature`.

### 6. Public-content hygiene (`copilot-instructions.md`, `comp-toml.instructions.md`)
Overlay descriptions, `replace-reason` strings, comp.toml comments, spec files, `modify_source.sh` echo lines, commit messages, and PR descriptions describe motivations technically and neutrally — no specific internal infrastructure names (signing services, scanner brand names, internal pipeline/tooling names, internal Azure tenants, etc.). Use neutral phrasing.

## Files

| File | Change |
| --- | --- |
| `.github/copilot-instructions.md` | +3 lines: GPG signing, public-content hygiene, render order rules. |
| `.github/instructions/comp-toml.instructions.md` | +69 lines: `replace-upstream` source-override pattern, public-content hygiene, `components.toml` file-organization rules. |
| `.github/skills/skill-modify-source/SKILL.md` | New skill (~267 lines). |
| `.github/skills/skill-signing-failure-remediation/SKILL.md` | New skill (~98 lines). |
| `.github/skills/skill-update-component/SKILL.md` | Clarified commit order with a worked example. |
| `AGENTS.md` | +2 lines in skill index for the two new skills. |
| `base/comps/AGENTS.md` | +4 lines: components.toml hygiene rule restated at directory-level. |

## Commits

1. `instructions(comp-toml): document replace-upstream source override pattern`
2. `skills: add skill-modify-source for byte-deterministic Source0 repack`
3. `instructions: clarify render order, GPG signing, public-content hygiene, and signing-failure remediation skill`

All commits are GPG-signed.


PawelWMS:
Superseded by #17411. The branch was moved from the PawelWMS fork to the upstream microsoft/azurelinux repo; the underlying commit is the same content, rebased onto the latest 4.0 and amended with a follow-up to skill-fix-overlay.

PR #17363, opened 2026-05-19
fix(repoclosure): mingw fixes

reubeno:
Several components were still emitting mingw sub-packages (and unsatisfiable mingw32-gcc / mingw64-gcc BuildRequires) despite being listed in the mingw-disabled group.

Root cause: the group works by passing --without mingw, which only takes effect on specs that gate mingw behind %bcond_with(out) mingw. Specs that instead hard-code %global with_mingw 1 need separate handling. Separately, the syslinux package needed to be bumped to force rebuilding.


trungams (approved):

base/comps/syslinux/syslinux.comp.toml
```diff
@@ -1,11 +1,22 @@
...
 # The mingw-gcc BuildRequires are used to cross-compile syslinux.exe/syslinux64.exe
 # (Windows tools for installing syslinux from Windows), which are not needed on AZL.
 
+# Bump upstream Release 0.33 → 0.34 to force a rebuild that picks up the
+# mingw-disabling overlays below. Without this, the previously-published SRPM
+# (built before the overlays landed) would continue to advertise mingw
+# BuildRequires and break build-dep closure for the AZL repo. Drop this
+# overlay the next time the upstream Release bumps past 0.33.
+[[components.syslinux.overlays]]
+description = "Bump Release 0.33 → 0.34 to force a rebuild that drops stale mingw BRs"
```
ddstreetmicrosoft, line 15: comment(non-blocking): after beta, we definitely need to look at refactoring to avoid needing to do regex to bump the release number

base/comps/dtc/dtc.comp.toml
```diff
@@ -0,0 +1,12 @@
...
+
+# Azure Linux does not ship mingw cross-compilation toolchains.
+# The upstream spec hard-codes `%global with_mingw 1` whenever %fedora is
+# defined (true in our build environment) rather than gating on a %bcond, so
+# the standard `build.without = ["mingw"]` mechanism has no effect. Force it
+# off via spec overlay.
+[[components.dtc.overlays]]
+description = "Disable mingw sub-packages — Azure Linux does not ship mingw toolchains"
+type = "spec-search-replace"
+regex = '%global with_mingw 1'
```
ddstreetmicrosoft, line 11: question(non-blocking): i thought there was a more specific build option, like `without = ["mingw"]`? or maybe that doesn't work for this? Either way, this regex should be fine

ddstreetmicrosoft (approved):
functional changes lgtm; minor comment/question also added

base/comps/dtc/dtc.comp.toml
reubeno: This only works for bcond-style options, which is *slightly* different syntax. We should pursue ways to simplify this form too.

base/comps/syslinux/syslinux.comp.toml
reubeno: 💯

PR #17372, opened 2026-05-20
add: container test plugin for pytest

realsdx:
Add a `pytest` based validation framework AzureLinux's OCI container images, integrated with `azldev image test`

Test it locally by (make sure to install and start podman first):
```
azldev image build container-base && azldev image test container-base
```


base/images/tests/README.md
```diff
@@ -115,20 +155,77 @@ base/images/
...
 
 ## Adding tests
 
-- **Shared (every image):** add a `cases/test_<topic>.py`. Use
+- **Shared static (every image):** add a `cases/static/test_<topic>.py`. Use
   `@pytest.mark.require_capability("…")` if the test only applies to
   images with a given capability.
-- **Image-specific:** add `cases/<image-family>/test_<topic>.py`. Tests
-  in such subdirectories are **automatically** restricted to that
+- **Image-specific static:** add `cases/static/<image-family>/test_<topic>.py`.
```
realsdx, line 162: I have odd feeling about these directory-based tests discovery, markers might be more useful. As you add more and more tests, directories and subdirectories just for categorization might become too complex.

base/images/tests/cases/runtime/container-base/test_nginx/test_nginx.py
```diff
@@ -0,0 +1,43 @@
...
+
+from __future__ import annotations
+
+import pytest
+
+
+@pytest.mark.dockerfile()
+def test_nginx_config_valid(container_exec) -> None:
+    """nginx configuration must pass validation."""
+    result = container_exec("nginx -t 2>&1")
```
reubeno: Why `2>&1`?

base/images/tests/README.md
```diff
@@ -67,12 +86,36 @@ System packages (not pip-installable):
...
-- **`buildah`** — cleanup of rootless umoci extracts (container images)
+- **`skopeo`** — OCI archive conversion (container images, static tests)
+- **`umoci`** — OCI image unpacking (container images, static tests)
+- **`buildah`** — cleanup of rootless umoci extracts (container images, static tests)
+- **`podman`** — container runtime for live tests (container images, runtime tests)
 - **`rpm`** — for `rpm --root` package queries
 - **`uv`** — Python project/package manager
 
+For runtime tests, the **podman socket** must be active:
+(ASK: Do we need this? podman-py is quite basic, maybe it's better without REST API and directly invoking podman CLI? The socket requirement is a bit of a pain, especially for WSL users. Then would need to have our own wrapper around podman CLI to handle the container orchestration logic currently in `container_runtime.py`. Or even docker SDK with moby might be more stable than podman-py?)
```
reubeno: Definitely open to other options here -- but we should leverage existing OSS if we can.

What about https://github.com/gabrieldemarmiesse/python-on-whales ? That seems to work with podman in rootless, daemonless mode too.

base/images/tests/README.md
```diff
@@ -115,20 +168,77 @@ base/images/
...
+framework injects the image-under-test automatically:
+
+```dockerfile
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+RUN dnf install -y nginx && dnf clean all
+COPY nginx.conf /etc/nginx/nginx.conf
+```
+
+Mark tests with `@pytest.mark.dockerfile()` to trigger the build. The
```
reubeno, line 199: I like this!

base/images/tests/README.md
realsdx: Looks like it has better interface, will check it out.

base/images/tests/cases/runtime/container-base/test_nginx/test_nginx.py
realsdx: I assume agent did this, cause by default `nginx -t` outputs to stderr

base/images/tests/README.md
realsdx: Updated

base/images/tests/cases/runtime/container-base/test_nginx/test_nginx.py
realsdx: Updated

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,283 @@
...
+        container_name = f"azl-test-{uuid.uuid4().hex[:12]}"
+
+    logger.info("Creating container %s from image %s", container_name, image_ref)
+
+    container: Container = client.container.run(
+        image_ref,
+        command=["sleep", "infinity"],
+        name=container_name,
+        detach=True,
+    )
```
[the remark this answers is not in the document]
realsdx: sleep is there in azl4

base/images/tests/cases/runtime/container-base/test_nginx/Dockerfile
```diff
@@ -0,0 +1,5 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+RUN dnf install -y nginx && dnf clean all
```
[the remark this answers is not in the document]
realsdx: Base alz4 has curl, but will update it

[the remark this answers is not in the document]
realsdx: updated

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,283 @@
...
+        else:
+            logger.info("Pulling image: %s", image_ref)
+            client.image.pull(image_ref)
+        return image_ref
+
+    # Load from archive file.
+    assert image_path is not None
+    logger.info("Loading image archive: %s", image_path)
+    image_refs = client.image.load(image_path)
+
```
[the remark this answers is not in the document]
realsdx: Fixed cleanup for loaded images

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,283 @@
...
+import uuid
+from pathlib import Path
+from typing import NamedTuple
+
+from python_on_whales import DockerClient
+from python_on_whales.components.container.cli_wrapper import Container
+from python_on_whales.exceptions import DockerException, NoSuchContainer, NoSuchImage
+
+from .tools import NativeTool
+
```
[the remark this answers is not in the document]
realsdx: Fixed

base/images/tests/README.md
reubeno: I think it depends a bit on the test. For tests that are tightly tied to a specific image (e.g., a scenaroi test for a specific golden container), I do think it makes sense to clearly demarcate that more strongly than just a marker - and also indicate with filenames or paths.

For tests that are more generic and apply to a wider group of images, I'd agree they should purely use markers.

I'd agree that the test harness using directory name automagically is probably something we should get away from / look at removing.

base/images/tests/cases/runtime/container-base/test_basic.py
```diff
@@ -0,0 +1,27 @@
...
+running container. Each test gets a fresh container instance via the
+``container_exec`` fixture (see conftest.py).
+"""
+
+from __future__ import annotations
+
+
+def test_shell_accessible(container_exec) -> None:
+    """Container shell must be functional via exec."""
+    result = container_exec("echo hello-from-container")
```
reubeno: Do you think it would be cleaner to take an array of strings as args, to avoid needing to deal with escaping and such?

base/images/tests/utils/pytest_plugin.py
```diff
@@ -109,9 +118,35 @@ def pytest_configure(config) -> None:  # type: ignore[no-untyped-def]
...
+    image_ref_raw = config.getoption("--image-ref", default=None)
+    if image_path_raw and image_ref_raw:
+        raise pytest.UsageError(
+            "--image-path and --image-ref are mutually exclusive. "
+            "Provide one or the other."
+        )
+    if not image_path_raw and not image_ref_raw:
+        raise pytest.UsageError(
+            "Either --image-path or --image-ref is required."
+        )
```
reubeno, line 142: **Correctness bug.** The `image_type` detection chain at lines 152–160 below checks `--image-type`, then `--capabilities`, then `--image-path` — but never `--image-ref`. When only `--image-ref` is provided (no `--image-type`, no `--capabilities`), `image_type` stays `None` and `check_tools(when=None)` returns **all** tools, including `guestmount` / `guestunmount` / `virt-inspector`. A container-only host then fails preflight with "missing required tool(s)" for VM tooling it doesn't need.

The `image_type` fixture in `conftest.py` (lines 107–110) handles this correctly, but `pytest_configure` runs first.

Suggested fix — slot in before the existing `--image-path` fallback at line 158:

```python
if image_type is None:
    if config.getoption("--image-ref", default=None):
        image_type = "container"
```

---
*Comment added by reuben's code review agent.*

base/images/tests/conftest.py
```diff
@@ -208,3 +246,106 @@ def partition_table(
...
+
+        if not dockerfile_path.exists():
+            pytest.fail(
+                f"Dockerfile not found: {dockerfile_path} "
+                f"(from @pytest.mark.dockerfile on {request.node.name})"
+            )
+
+        effective_image = build_image(
+            podman_client, dockerfile_path, container_image_ref,
+        )
```
reubeno, line 327: _NOTE: This was an AI suggestion. I can see arguments against doing this too, but I'll leave the comment here in case you find it helpful._

---
**Missing guardrail — silent footgun.** This fixture only consults `@pytest.mark.dockerfile()`. If a contributor adds `test_postgres/Dockerfile` but **forgets the marker** on the test, the test silently runs against the unmodified base image and fails with cryptic errors like `psql: command not found`. They'll burn time debugging the Dockerfile when the real issue is the missing marker. This is the kind of footgun that, left unfixed now, will hit every new contributor.

Suggested guardrail to add before line 305:

```python
test_dir = Path(request.fspath).parent
default_dockerfile = test_dir / "Dockerfile"
dockerfile_marker = request.node.get_closest_marker("dockerfile")
if default_dockerfile.exists() and dockerfile_marker is None:
    pytest.fail(
        f"{default_dockerfile} exists but {request.node.name} "
        f"is missing @pytest.mark.dockerfile(). Add the marker "
        f"or move/rename the Dockerfile."
    )
```

This turns 30 minutes of confused debugging into a one-line fail-fast message.

---
*Comment added by reuben's code review agent.*

base/images/tests/cases/runtime/container-base/test_nginx/test_nginx.py
```diff
@@ -0,0 +1,43 @@
...
+def test_nginx_serves_response(container_exec) -> None:
+    """nginx must start and serve HTTP responses."""
+    # Start nginx in the background.
+    start = container_exec("nginx")
+    assert start.exit_code == 0, f"nginx failed to start: {start.output}"
+
+    # Verify it responds on port 80.
+    result = container_exec("curl -sf http://localhost:80/")
+    assert result.exit_code == 0, f"curl failed: {result.output}"
+    assert "azl-nginx-ok" in result.output
```
reubeno: **No service-readiness wait.** 

This pattern — `container_exec("nginx")` immediately followed by `container_exec("curl ...")` — works only because nginx 1.x daemonizes after binding its listen socket. The next contributor testing postgres / sshd / redis / a node app / a python service will copy this template and ship a flaky test: those services don't bind synchronously.

This is the highest-leverage place in the framework to set a good example, because every future Dockerfile-based service test will be modeled on this one.

Recommended: add helpers that can be used for readiness checks, perhaps on TCP port availability, etc. and then document the pattern under a new "Service readiness checks" subsection in the README (see related comment on the Dockerfile-based runtime tests section).

---
*Comment added by reuben's code review agent.*

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,288 @@
...
+)
+
+_loaded_image_refs: set[str] = set()
+_built_image_cache: dict[tuple[Path, str], str] = {}
+
+
+class ContainerExecResult(NamedTuple):
+    """Result of executing a command inside a container."""
+    exit_code: int
+    output: str
```
reubeno, line 45: **This is a one-way door.** `container_exec` is the most used API of the new framework — every runtime test will assert on `result.output`. Once 20 tests are written against the merged shape, splitting later breaks them all. Two concerns to address before that happens:

**1. Merged streams discard signal and ordering.** `exec_in_container` (lines 257–279) iterates `for _stream_name, chunk in output` and explicitly drops the stream name. Order between stdout and stderr chunks is non-deterministic across the boundary. Tests that want to assert on stderr (e.g., `nginx -t` writes its result to stderr) end up matching a concatenated string with no ordering guarantee.

**2. Error-path output loss.** Lines 273–278: when `DockerException` is raised mid-stream, `exc.stdout` / `exc.stderr` are backfilled **only if** `output_parts` is empty. If some chunks streamed through before the exception, they're kept AND `exc.stdout` / `exc.stderr` is silently dropped on the floor. Either keep both, or document the precedence explicitly.

Suggested:

```python
class ContainerExecResult(NamedTuple):
    exit_code: int
    stdout: str
    stderr: str
    output: str  # convenience: stdout + stderr, kept for backward-compat
```

And accumulate per-stream in `exec_in_container` using the `stream_name` that python-on-whales already provides.

---
*Comment added by reuben's code review agent.*

base/images/tests/cases/runtime/container-base/test_nginx/test_nginx.py
```diff
@@ -0,0 +1,43 @@
...
+
+@pytest.mark.dockerfile()
+def test_nginx_health_endpoint(container_exec) -> None:
+    """nginx /health endpoint must return 200."""
+    start = container_exec("nginx")
+    assert start.exit_code == 0, f"nginx failed to start: {start.output}"
+
+    result = container_exec("curl -sf http://localhost:80/health")
+    assert result.exit_code == 0, f"health check failed: {result.output}"
+    assert "healthy" in result.output
```
reubeno: One comment on this file as the first concrete example future contributors will copy:

**`test_nginx_health_endpoint` is redundant** with `test_nginx_serves_response`: same Dockerfile, same start, same curl, different URL. No incremental coverage — it tests a route the test's own `nginx.conf` defines (line 22–25 of `nginx.conf`).

(Non-blocking for this PR) Also consider adding a sibling test that validates the *default* nginx conf in the distro too.

---
*Comment added by reuben's code review agent.*

base/images/tests/README.md
```diff
@@ -115,20 +148,77 @@ base/images/
...
+```
+
+Built images are cached per session — multiple tests sharing the same
+Dockerfile only trigger one build.
+
+> **Note:** All containers (plain and Dockerfile-based) run with
+> `sleep infinity` as PID 1 — the Dockerfile's `CMD`/`ENTRYPOINT` is
+> overridden. Tests that need a service should start it explicitly via
+> `container_exec("nginx")`. This keeps behaviour predictable and
+> ensures each test controls exactly what runs.
```
reubeno, line 223: The Dockerfile-based runtime tests section is solid for the happy path, but four canonical patterns belong here before more service tests land — without them, every future contributor reinvents (or skips) them:

**1. Service readiness checks.** How to wait for a service to be ready before asserting against it. See related comment on `test_nginx.py:22–32`. Without this, every test that copies the nginx pattern for a slower-starting service (postgres, sshd, redis, etc.) will flake.

**2. Backgrounding non-daemonizing services.** `container_exec("nginx")` works because nginx daemonizes after binding. For services that run in foreground by default (`python -m http.server`, `node server.js`, `redis-server` without `daemonize yes`, etc.), the call blocks forever. Document the canonical pattern (`container_exec("nohup my-service > /tmp/log 2>&1 &")`) — and ideally provide a `container_exec_background` helper so contributors don't all roll their own.

**3. Bash / `sleep infinity` / `getent` contract.** Runtime tests silently require `/bin/bash` (used by `exec_in_container`), GNU coreutils `sleep infinity` (used by `create_container`), and `getent` (used by `test_basic.py`). Distroless variants are excluded by configuration in `images.toml`, but a contributor who later adds `tests.test-suites = [...]` to a distroless image will hit cryptic errors with no breadcrumb. State the contract.

**4. `ARG BASE_IMAGE` troubleshooting.** A Dockerfile that omits `ARG BASE_IMAGE` fails with `Error: no FROM statement found`, which gives no hint about the real cause. A one-liner in a troubleshooting subsection — *"if you see 'no FROM statement found', you forgot ARG BASE_IMAGE"* — will save someone real debugging time.

---
*Comment added by reuben's code review agent.*

base/images/tests/utils/pytest_plugin.py
```diff
@@ -109,9 +118,35 @@ def pytest_configure(config) -> None:  # type: ignore[no-untyped-def]
         "image(name): only run this test when --image-name matches the named image family "
         "(exact match, or a ``<family>-<variant>`` image-name)",
     )
+    config.addinivalue_line(
+        "markers",
+        "runtime_container_tests: auto-applied to tests under cases/runtime/; "
+        "used for -m filtering to separate static and runtime suites",
+    )
```
reubeno, line 119: Naming nit, but a cheap pattern-establishing fix worth doing before this gets hardcoded into CI scripts and contributor muscle memory.

`-m "not runtime_container_tests"` reads as a hardcoded label, not a category. The obvious incantation a developer would try is `-m "not runtime"` — which today silently runs everything (no `runtime` marker exists). Renaming now:

```python
config.addinivalue_line(
    "markers",
    "runtime: auto-applied to tests under cases/runtime/; "
    "use -m 'not runtime' to exclude runtime tests",
)
```

Then update `pytest_collection_modifyitems` (`pytest.mark.runtime_container_tests` → `pytest.mark.runtime`) and the test-suite definitions in `images.toml`. Costs nothing now; costs every contributor and downstream script later.

---
*Comment added by reuben's code review agent.*

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,288 @@
...
+    image_tag = f"localhost/azl-test-{uuid.uuid4().hex[:12]}:latest"
+    logger.info(
+        "Building image from %s (base: %s, context: %s)",
+        dockerfile_path, base_image_ref, context_dir,
+    )
+
+    client.legacy_build(
+        context_path=context_dir,
+        file=dockerfile_path,
+        build_args={"BASE_IMAGE": base_image_ref},
```
reubeno, line 159: The framework injects `BASE_IMAGE` as a build arg here, and the README documents that test Dockerfiles must use `ARG BASE_IMAGE` / `FROM ${BASE_IMAGE}`. But there's no validation. If a contributor forgets the `ARG` line, `${BASE_IMAGE}` expands to empty and podman fails with `Error: no FROM statement found` — a message that gives no hint about what's actually wrong.

Two paths:

**(a) Document loudly** — add a "Common errors" or troubleshooting subsection to the README mapping `"no FROM statement found"` → `"you forgot ARG BASE_IMAGE"`. Cheap, sufficient.

**(b) Validate before build** — in `build_image`, read the Dockerfile and check that an `ARG BASE_IMAGE` line precedes any `FROM ${BASE_IMAGE}` usage. Fail fast with a clear message.

(a) alone is enough for merge. (b) is nice-to-have but could feel like over-engineering — your call.

---
*Comment added by reuben's code review agent.*

base/images/tests/utils/container_runtime.py
```diff
@@ -0,0 +1,288 @@
...
+    # Verify the container is running and exec works.
+    try:
+        container.reload()
+        if container.state.status != "running":
+            raise ContainerRuntimeError(
+                f"Container {container_name} is not running "
+                f"(status: {container.state.status})"
+            )
+
+        result = exec_in_container(client, container_name, "echo ready")
```
reubeno: This is the framework's own startup probe, and it goes through `exec_in_container(... "echo ready")` — which wraps everything in `["bash", "-c", command]` (line 262). For `container-base` today, bash is present and this works. But the **framework-level** readiness probe is the wrong place to bake in a bash dependency: it means a future minimal/musl image variant can't even fail cleanly — it fails at the framework's probe before the test body ever runs, so the failure mode looks like a framework bug rather than a test-author choice.

For the readiness check specifically, use a direct, shell-less exec — `echo` is in coreutils and is effectively universal:

```python
container.reload()
ready = client.container.execute(
    container_name, ["echo", "ready"], stream=False,
)
if ready.strip() != "ready":
    raise ContainerRuntimeError(...)
```

Test-level uses of `container_exec` can keep wrapping in `bash -c` — that's the test author's choice. The framework's own probe should be minimal.

---
*Comment added by reuben's code review agent.*

base/images/tests/pyproject.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 requires-python = ">=3.12"
 dependencies = [
     "pytest>=8.0",
+    "python-on-whales>=0.81.0",
```
reubeno, line 7: ❤️ Thanks for reaching for `python-on-whales` here instead of hand-rolled `subprocess` plumbing. Pulling in a well-maintained library that already models containers, images, builds, and exec is the right call and a material step forward for the test infra.

---
*Comment added by reuben's code review agent.*

base/images/tests/conftest.py
realsdx: This is unnecessary guardrail, should be caught during PR review itself if someone puts up a PR without testing

base/images/tests/utils/container_runtime.py
realsdx: Good suggestion, will do

base/images/tests/utils/container_runtime.py
realsdx: Ah, it's documented already, but yeah, we could add extra validation, but I am not sure.

base/images/tests/utils/container_runtime.py
realsdx: Is/will there be version of azl4 that ships without bash? if so, we need to address it.

reubeno: Distroless containers ship without bash, yes. There should be some tests in 3.0 that test them using some .NET apps. They may be good examples to look at, particularly now that dotnet is available in 4.0.

base/images/tests/cases/runtime/container-base/test_nginx/test_nginx.py
realsdx: I would not like to add this at container runtime level, but as an example on service readiness

base/images/tests/cases/runtime/container-base/test_basic.py
realsdx: Yes, will fix

base/images/tests/utils/container_runtime.py
realsdx: Okay thanks

base/images/tests/cases/runtime/container-base/test_basic.py
realsdx: I added two exec fixtures to make the API explicit:

- `container_exec_shell("...")` is the ergonomic default for normal runtime tests that want shell syntax like pipes, redirects, or backgrounding.
- `container_exec([...])` is the shell-free argv path for minimal/distroless images or tests that need exact argument boundaries.

Both go through the same underlying exec implementation, so output capture and error handling stay consistent.

base/images/tests/cases/runtime/container-base/test_nginx/test_nginx.py
realsdx: added a wait for http helper for this case

base/images/tests/utils/container_runtime.py
realsdx: fixed readiness check no longer uses bash

base/images/tests/README.md
realsdx: Fixed all, for 2 and 4 documentation is already there

base/images/tests/utils/container_runtime.py
realsdx: Fixed

reubeno (approved):

PR #17375, opened 2026-05-20
fix(grub2): remove patches that drop os-prober default-enable setting

vinceaperri:
This change removes downstream revert patches that effectively re-enable os-prober by default in grub2.

Removed from specs/g/grub2/grub.patches:
- 0002-Revert-templates-Properly-disable-the-os-prober-by-d.patch
- 0003-Revert-templates-Disable-the-os-prober-by-default.patch

Result: restores upstream secure-default behavior where os-prober is disabled unless explicitly enabled

Fixes: [AB#20174](https://dev.azure.com/mariner-org/mariner/_workitems/edit/20174)


specs/g/grub2/grub.patches
```diff
@@ -1,6 +1,4 @@
 Patch0001: 0001-Revert-templates-Fix-user-facing-typo-with-an-incorr.patch
```
reubeno, line 1: Thanks for your contribution to Azure Linux 4! Please review some of the docs in this project that explain the layout and how we construct these specs.

Specifically, please note that the `specs/**` dir is entirely mechanically generated. The source of truth for components in the distro is under `comps/`. If we need to remove patches to an upstream Fedora spec, it should be done with an "overlay".

specs/g/grub2/grub.patches
```diff
@@ -1,6 +1,4 @@
...
-Patch0002: 0002-Revert-templates-Properly-disable-the-os-prober-by-d.patch
```
reubeno, line 2: Do you/we know why Fedora carries these patches? It would be great to understand the impact of removing them, as well as the impact of not removing them.

reubeno (changes requested):
Could you please update the PR to be a single commit that follows Conventional Commits? Our CONTRIBUTING.md has more details on expectations for AZL4 PRs.

specs/g/grub2/grub.patches
vinceaperri: Unfortunately, neither the changelog nor the patches state a specific reason for the reversion, but there is an older related message in the changelog that [says](https://src.fedoraproject.org/rpms/grub2/blob/rawhide/f/grub2.spec#_1886):

>Put the os-prober dep back in - we need to change test plans and criteria before it can go.

The reversion was in the context of an upgrade to grub 2.06 (which introduced the default-disable behavior), so these patches most likely exist due to avoid breaking existing customers (backcompat) or to avoid breaking test infra.

base/comps/grub2/grub2.comp.toml
```diff
@@ -16,3 +16,31 @@ type = "file-search-replace"
 file = "grub.macros"
 regex = '%global with_xen_pvh_arch 1'
 replacement = '%global with_xen_pvh_arch 0'
+
+# Drop the two Fedora revert patches that re-enable the os-prober-by-default
+# behavior in grub-mkconfig. os-prober has been disabled by default for security
+# reason in grub2 (2.06+), and previous versions of Azure Linux have
+# historically followed suite.
+[[components.grub2.overlays]]
```
reubeno, line 29: Thanks for switching to overlays!

Could you see if you're able to use a `spec-patch-remove` overlay? For more details, [here's the documentation on that](https://github.com/microsoft/azure-linux-dev-tools/blob/44f81cce609aef9b1352ad628f1a658822bdfa30/docs/user/reference/config/overlays.md#removing-a-patch).

reubeno (changes requested):
Thanks for the iteration! We know there's a bit of a learning curve.

Could you also squash the two commits you have in the PR? For 4.0, we're rebase-merging and aim to include the lock file updates, spec rendering, etc. in the same commit that makes changes to the TOML files / inputs.

base/comps/grub2/grub2.comp.toml
vinceaperri: You're welcome! Unfortunately, I cannot use patch-remove since this package `%include`s them from `grub.patches` instead of having them inline in the `.spec`.

reubeno (changes requested):
Changes look good -- thanks for looking into and reporting back on patch-remove feasibility. Understood that this is the best option.

Did you use `azldev component render` to update the 

The main missing at this point are doing the following:

```
$ azldev component update grub2
$ git commit --amend -a
$ azldev component render grub2
$ git git commit --amend -a # if any changes were made ^^^
```

The `update` regenerates the lock file metadata for this component, which should be used by `render` to keep the generated spec in sync.

reubeno (approved):

PR #17413, opened 2026-05-23
feat(azurelinux-repos): add Microsoft subrepo

christopherco:
Ship /etc/yum.repos.d/microsoft.repo in both the official and dev subpackages, pointing at packages.microsoft.com/azurelinux/$releasever/beta/microsoft/$basearch. Enabled by default, GPG-signed with the Azure Linux primary key.

Shipped as a separate static source (not a section in the shared azurelinux.repo.in template) so its fixed packages.microsoft.com URL isn't rewritten through the dev variant's REPO_URI_PREFIX. Installed as microsoft.repo.main/.dev; the existing RemovePathPostfixes directives strip the suffix in each subpackage. %check validates both files.

Fixes: [AB#20074](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20074)


base/comps/azurelinux-repos/microsoft.repo
```diff
@@ -0,0 +1,11 @@
+[azurelinux-microsoft]
+name=Azure Linux $releasever - $basearch - Microsoft
```
reubeno, line 2: Non-blocking: comparing this with `azurelinux.repo`, it's easy to notice that this subrepo has no pattern for publishing debuginfo packages or SRPMs.

That's something we should evaluate advocating for in the future, at least for packages that are in this subrepo but are fully OSS (e.g. dotnet).

reubeno (approved):

PR #17418, opened 2026-05-24
fix(python-geographiclib): neutralize MinGW macros to avoid empty subshell

christopherco:
The upstream Fedora spec wraps the MinGW install steps in a subshell:

    (
    %mingw32_py3_install_wheel
    %mingw64_py3_install_wheel
    )
    %mingw_debug_install_post

Azure Linux does not ship MinGW toolchains, so existing overlays replaced the MinGW macros inside the subshell with empty strings. That left an empty `( )` block, which bash rejects with `syntax error near unexpected token ')'` when %install runs, failing the build.

Change the MinGW `%build` and `%install` overlays to substitute the bash no-op `:` instead of an empty string. The subshell stays syntactically valid and the macros become true no-ops at runtime.

Fixes: [AB#20285](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20285)

### DIFF
--- a/base/comps/python-geographiclib/python-geographiclib.comp.toml
+++ b/base/comps/python-geographiclib/python-geographiclib.comp.toml
@@ -50,36 +50,36 @@ tag = "BuildRequires"
 value = "mingw64-python3-build"
 
 [[components.python-geographiclib.overlays]]
-description = "Remove mingw32 wheel build step"
+description = "Neutralize mingw32 wheel build step (replace with bash no-op `:` to preserve subshell structure)"
 type = "spec-search-replace"
 section = "%build"
 regex = '^\%mingw32_py3_build_wheel$'
-replacement = ''
+replacement = ':'
 
 [[components.python-geographiclib.overlays]]
-description = "Remove mingw64 wheel build step"
+description = "Neutralize mingw64 wheel build step"
 type = "spec-search-replace"
 section = "%build"
 regex = '^\%mingw64_py3_build_wheel$'
-replacement = ''
+replacement = ':'
 
 [[components.python-geographiclib.overlays]]
-description = "Remove mingw32 wheel install step"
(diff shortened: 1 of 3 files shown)


base/comps/python-geographiclib/python-geographiclib.comp.toml
```diff
@@ -83,3 +83,17 @@ type = "spec-search-replace"
 section = "%install"
 regex = '^\%mingw_debug_install_post$'
 replacement = ''
+
+[[components.python-geographiclib.overlays]]
+description = "Remove opening paren of now-empty MinGW install subshell — empty `( )` is a bash syntax error"
```
reubeno: This is functionally correct but want to note it's probably simpler to replace the earlier-matched patterns with a no-op shell command that can still work inside the subshell parens (e.g., `:`).

Actually, come to think of it, did we consider a broader redefinition of these macros in question to avoid needing to patch up the spec so surgically-- either for this spec locally, or globally?

christopherco: I did not consider changing the macros. Do you have a specific suggestion you're looking for?

reubeno: Concrete suggestion: for the scope of this PR, drop the 2 new overlays and instead update the replacement property of %build and %install overlays in this spec to use a replacement of ':' (no quotes).

Suggestion for longer-term follow-up post this PR: identify where these mingw macros are being centrally defined and evaluate feasibility of changing their global definitions in the whole distro, either unconditionally or triggered via macro.

christopherco: Added `:` in the regex replacements

reubeno (approved):

PR #17172, opened 2026-05-13
Extend explicit workflow permissions to 19 check/lint/quickstart workflows

arpitjain099:
Following up on #7282, which made a first pass at declaring explicit `permissions:` blocks in this repo. Since then several new check/lint workflows have been added and the 2.0 quickstart was renamed without picking up the same hardening, so most of the validators in `.github/workflows/` are once again running with the repository default token scope.

This patch adds a top-level

```yaml
permissions:
  contents: read
```

to 19 workflows:

```
check-circular-deps.yml            check-package-cgmanifest.yml      check-spec.yml
check-clean-stage.yml              check-package-update-gate.yml     check-srpm-duplicates.yml
check-entangled-specs.yml          check-source-signatures.yml       check-static-glibc.yml
check-files.yml                    go-test-coverage.yml              lint-specs.yml
check-kernel-configs.yml           merge-conflict-check.yml          quickstart_2.0.yml
check-license-map.yml              verify-osguard-imageconfigs.yml
check-manifests.yml                check-package-builds.yml
```

All 19 fire on `push` / `pull_request` (or `schedule` / `workflow_dispatch` for `quickstart_2.0.yml`), check out the repo, and run spec / package / lint validators. None push commits, create releases, or call write APIs.

The two workflows that already declare `permissions: {}` with `pull_request_target` (`check-rendered-specs-stub.yml`, `spec-review-stub.yml`) are intentionally left alone — they have a stub-plus-reusable architecture with explicit scope notes and don't need a different default.

YAML validated locally with PyYAML before commit; diff is +57 lines, no other changes.

PR #17438, opened 2026-05-25
ci: declare workflow-level contents: read on 3 check workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` at the workflow level. The workflow runs checks only; no GitHub API writes.

Same post-CVE-2025-30066 (`tj-actions/changed-files`) supply-chain hardening pattern. YAML validated locally with `yaml.safe_load`.


arpitjain099:
Closing as duplicate of #17172 (earlier PR with overlapping scope), apologies for the noise.

PR #17439, opened 2026-05-25
fix(erlang): build and ship common_test and dialyzer subpackages

christopherco:
The upstream Fedora spec gates erlang-common_test and erlang-dialyzer behind `%if %{__with_wxwidgets}`, and also passes
`--without-common_test` / `--without-dialyzer` to ./configure when wxwidgets is off. AZL disables wxwidgets (no wxGTK in a headless cloud distro), so neither subpackage was built — yet erlang-rebar3 requires both, which made buildroot resolution fail for the entire erlang/elixir stack (~20 downstream consumers).

Neither library actually links against wxGTK: each has only one optional GUI entry point (`dialyzer:gui/0`, `ct` GUI launcher) that delegates to a single wx-using module, and Erlang lazy-loads modules so the wx code path is never reached in CLI mode. Confirmed in upstream OTP-26.2.5.17 sources: dialyzer.app.src and common_test.app.src list only kernel/stdlib/compiler in `applications` (the enforced load list); wx appears only in the advisory `runtime_dependencies` metadata.

Overlay does, in order:
  1. un-gate the %package/%files blocks for common_test and dialyzer
  2. drop --without-common_test / --without-dialyzer from %configure (replace with no-op tab+backslash so the surrounding \\<newline> continuation stays intact)
  3. strip Requires for the wx-using subpackages we don't ship (debugger/observer/wx) from common_test, dialyzer, and src
  4. fix the wxwidgets-off %install cleanup loop so it no longer deletes ct_run/dialyzer, and rename bindir/typer to erlang-typer to avoid the python3-typer-cli conflict (rhbz#2359567), matching the wxwidgets-on branch

Fixes: [AB#20325](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20325)

### DIFF
--- a/base/comps/erlang/erlang.comp.toml
+++ b/base/comps/erlang/erlang.comp.toml
@@ -34,10 +34,237 @@ replacement = '%global __with_wxwidgets 0'
 # only targets these unconditional Requires; equivalent lines elsewhere in
 # the spec are already inside `%if %{__with_wxwidgets}` blocks and are
 # dead code under the toggle, so they stay untouched.
+#
+# common_test and dialyzer are intentionally excluded from this strip —
+# they are re-enabled below (they are pure Erlang libraries that do not
+# actually link against wxGTK; upstream only grouped them under the same
+# __with_wxwidgets toggle because their *runtime* Requires reference the
+# wx-using subpackages debugger/observer/wx, which we drop separately).
+#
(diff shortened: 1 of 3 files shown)


reubeno:
This PR has been supplanted by #18234 . Thanks for getting the ball rolling!

PR #17443, opened 2026-05-26
test: add framework label catalog and image test mappings

bhagyapathak:
###### Merge Checklist  

- [ ] Ready to merge

---

###### Summary 

It introduces a declarative, in-tree test-selection layer so CI can decide what to run from data living next to the image being tested, instead of from out-of-tree CI YAML or hard-coded scripts.

**NO**

### DIFF
--- a/base/images/images.tests.toml
+++ b/base/images/images.tests.toml
@@ -1,0 +1,190 @@
+# ============================================================================
+# Image test catalog
+# ============================================================================
+#
+# Single source of truth for all image-runnable test definitions and
+# the named groups that bundle them.
+#
+# Schema:
+#   [tests.X]            : a single unit of testing — one configuration of
+#                          one runner/harness, with framework-specific
+#                          options under [tests.X.<framework>].
+#   [test-groups.X]      : a named bundle of tests an image (or component)
+#                          can reference instead of enumerating members.
+#
+# Image and component associations are NOT defined here. They live with
+# the entity that runs them:
+#   - base/images/images.toml          (tests.tests = [...])
+#   - base/comps/<name>/<name>.comp.toml (tests.tests = [...])
+#
+# Both fields can mix per-test and per-group references:
+#   tests.tests = [{ name = "static-image-checks" }, { group = "x" }]
+#
+# Declarative metadata on [tests.X]:
+#   description           : human-readable summary
+#   type                  : "pytest" | "tmt" | "lisa"
+#   kind                  : "functional" | "performance"
+#   long-running          : boolean; omit unless the test may run for hours
+#   required-capabilities : capability tokens the image must declare
+#
+# ============================================================================
+
+# =============================================================================
+# Static image checks (pytest)
+# =============================================================================
+
+# Single shared pytest suite for static (offline) image validation.
+[tests.static-image-checks]
+type        = "pytest"
+description = "Offline image validation (shared + image-specific tests)"
+kind        = "functional"
+
+[tests.static-image-checks.pytest]
+working-dir = "tests"
+install     = "pyproject"
+test-paths  = ["cases/static/"]
+# {capabilities} is substituted by azldev as a comma-separated list of
+# the capabilities set to `true` for the image (e.g.
+# "machine-bootable,systemd,runtime-package-management"). The pytest
+# plugin parses this into a set; @pytest.mark.require_capability("...")
+# tests that don't match are skipped.
+extra-args = [
+    "--image-path", "{image-path}",
+    "--image-name", "{image-name}",
+    "--capabilities", "{capabilities}",
+]
+
+# Runtime container tests — validate live container behavior via podman.
+[tests.runtime-container-tests]
+type        = "pytest"
+description = "Runtime container validation (exec into live containers)"
+kind        = "functional"
+required-capabilities = ["container"]
+
+[tests.runtime-container-tests.pytest]
+working-dir = "tests"
+install     = "pyproject"
+test-paths  = ["cases/runtime/"]
+extra-args = [
+    "--image-path", "{image-path}",
+    "--image-name", "{image-name}",
+    "--capabilities", "{capabilities}",
+]
+
+# =============================================================================
+# LISA tests
+# =============================================================================
+
+[tests.smoke-test]
+type                  = "lisa"
+description           = "LISA smoke test — fastest PR gate (pinned)"
+kind                  = "functional"
+required-capabilities = ["machine-bootable"]
+[tests.smoke-test.lisa]
+criteria = { testcase-names = ["smoke_test"] }
+
+[tests.functional-core]
+type                  = "lisa"
+description           = "Core functional coverage (LISA priority=1)"
+kind                  = "functional"
+required-capabilities = ["machine-bootable"]
+[tests.functional-core.lisa]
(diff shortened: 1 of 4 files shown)


reubeno:
Hi @bhagyapathak --

First off, thanks for putting this PR together. The format made it much easier to understand the pieces you're proposing, and concretely pushed things forward to identifying the relationship between components/images and tests. You've also
introduced a new version of labels as a TOML-local grouping structure that can be used for, say, components to indirectly associate with a heterogenous collection of tests. I very much like that idea and think it helps us move things forward.

I threw together https://github.com/reubeno/azurelinux/commit/e3f1e950989d602f3b31d48222b1458284c5b03a as a proposed update to your PR. It's mostly renaming/re-shaping your proposed syntax to more strongly align with the existing patterns / principles in the TOML file -- most notably, the concept of "groups" and a desire to abstractly model a "test", independent from what it really maps to natively in the framework it uses. I also want to acknowledge that I have separated out some operational choices and policy that I think either belongs elsewhere or needs to be something we layer on later after we get the basics in place.

## Concepts kept (with different naming/syntax)

- **A named, reusable test record.** Your `*_labels` entries become a single  `[tests.X]` namespace. Instead of encoding the framework in the table-name prefix  (`tmt_`, `pytest_`, …), each test carries an explicit `type` plus a framework
  subtable (`[tests.X.pytest]`, `[tests.X.tmt]`). Same idea, one namespace with subtyping.

- **Framework-native filters as structured config.** Your `pytest_markers` /   `pytest_files` / `pytest_args` survive as structured fields under `[tests.X.pytest]`   (`test-paths`, `extra-args`, etc.). The TMT filter/source concept survives under
  `[tests.X.tmt]`.

- **Grouping.** Your labels-as-buckets idea becomes a first-class   `[test-groups.X]` — a named bundle of tests, mirroring how `[component-groups.X]`   and `[package-groups.X]` already work in the repo. This is where the "run this
  whole set" intent lives. I'd also note that the existing component group concept   allows another way to identify a set of components that inherit a common set of   properties, which could include associated tests too.

- **Associating tests with the things they test.** Your `test_labels = [...]` on   components and the per-image label lists both become one uniform form:    `tests.tests = [{ name = "…" } | { group = "…" }]`, declared **on** the image or   component. We extend this so components (not just images) can carry associations   and join shared groups, and make sure that there's the same syntax for doing this   between components and images.

- **Capability gating.** `requires_capabilities` is genuinely declarative, so I think   it makes sense to keep. I'd be inclined to rename it `required-capabilities` to   match the language and casing we use elsewhere in the TOML files. It's also something   that would now live on a `test` definition, in a common (non-framework-specific)   property.

## Properties deferred (and why)

The one structural change worth calling out: I pulled **workflow/execution policy** out of the test definitions. In the PR, there's workflow / scenario-oriented policy (e.g., `pr_validation` vs. `nightly_validation`) as well as operational toggles (e.g., the `[frameworks]` enable/disable options).

To be clear — **I agree that policy is needed.** The Control Tower needs to understand (and/or decide) when to run certain tests, whether certain associated tests are appropriate to request in the current workflow, etc. These workflow-sensitive policies should be considered a *separate concern* from the test describing itself. So the goal for `[tests.X]` is for a test to declare **what it is and how to run it** as purely as possible, and for the *when/how-often* policy to be layered on separately, in meta-policy for qualification / promotion scenarios.

## Small additions

- `kind = ["functional", "performance"]` — a closed enum on a test that allows the   test to advertise the role it plays.
- `long-running = true` — a hint for tests that may run for hours. I think a pure   estimated number could get messy over time, and anticipate that for that level of   precision we'll want to leverage historical execution data. But for now, this is
  a placeholder to indicate that certain tests are *not* ones that can finish in,   say, < 30 mins. I think we'll need to refine and decide what this really means.

sandeepkarambelkar:
Hello @reubeno and @bhagyapathak,

Just want to understand whats the difference between test-suites and test-groups here? Aren't they both the same? Can't we use traditional test-suites terminology for grouping tests?

base/images/images.tests.toml
```diff
@@ -0,0 +1,159 @@
...
+description = "Runtime container validation (exec into live containers)"
+kind        = "functional"
+required-capabilities = ["container"]
+
+[tests.runtime-container-tests.pytest]
+working-dir = "tests"
+install     = "pyproject"
+test-paths  = ["cases/runtime/"]
+extra-args = [
+    "--image-path", "{image-path}",
```
realsdx, line 69: Should Image path be part of this? or this is for local CLI use only?

CT/VS already their own ways of passing images for tests

base/images/images.tests.toml
```diff
@@ -0,0 +1,159 @@
...
+# LISA tests
+# =============================================================================
+
+[tests.smoke_test]
+type                  = "lisa"
+description           = "LISA smoke test — fastest PR gate (pinned)"
+kind                  = "functional"
+required-capabilities = ["machine-bootable"]
+[tests.smoke_test.lisa]
+source   = { git-url = "https://github.com/microsoft/lisa.git", ref = "20260330.1" }
```
realsdx: We cannot support direct git refs ATM, will require design changes for our strict patch management. Please remove this

base/images/images.tests.toml
```diff
@@ -0,0 +1,159 @@
...
+criteria = { testcaseNames = ["smoke_test"] }
```
realsdx: `testcaseNames` should be `testcase-names`? rest of toml follows uses `-` as separator

realsdx: Same as `functional_core` should be `functional-core`

base/images/images.tests.toml
bhagyapathak: This is for local use in azldev

.azldev-version
```diff
@@ -1 +1 @@
-261ae7471a4421377d557c51aaae70add8bc9caf
+8e1b2efc0d48dac30258a6bc6e06095a07ac1ceb
```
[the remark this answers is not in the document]
bhagyapathak: Fixed

base/comps/kernel-headers/kernel-headers.comp.toml
```diff
@@ -1,4 +1,5 @@
 [components.kernel-headers]
+tests.tests = [{ group = "kernel-tests" }]
```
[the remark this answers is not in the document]
bhagyapathak: Fixed

jslobodzian (approved):

jslobodzian:
[Toolio Iglesias🗺️] Heads-up (not a blocker): `[tests.bvt-metrics]` sets `metrics-enabled = true` (and the schema adds a matching `TestDefinition` property). This key isn't a first-class field in the Control Tower validation-service 2.0 parser (PR mariner-org/azl!28586), but I confirmed CT's `BuildTestMetadata` routes unmapped keys into `additionalFields` (camelCased to `metricsEnabled`), so it **is** preserved end-to-end. The only open item is confirming the downstream validation service actually reads `additionalFields.metricsEnabled` — otherwise the BVT metrics intent is silently dropped. Everything else (tests/testGroups migration, capability tokens, group→suite refs, kebab-case keys) cross-checks cleanly against the CT parser.

tobiasb-ms (approved):

PR #17450, opened 2026-05-27
fix(vm-iso-installer): set efi_dir to azurelinux and add vim-minimal

kgodara912:
Change Anaconda override in azurelinux.conf from efi_dir = fedora to efi_dir = azurelinux so installer bootloader paths align with Azure Linux EFI vendor directory. Add vim-minimal to live iso for better user experience.


reubeno (approved):
Confirmed fixes issue for me in local build.

PR #17549, opened 2026-05-28
Removed the final bootstrap04 from the golang spec

amitu314:
This PR removes the remaining bootstrap04 component, thereby reducing the overall attack surface. The residual artifact (bootstrap in Go) has previously been associated with reported vulnerabilities, making its removal a security improvement.


amitu314:
@microsoft-github-policy-service agree company="Microsoft"

amitu314:
retargeted this PR from 3.0 to 3.0-dev branch

mfrw (changes requested):
Can we please use 3.0-dev to create PR.
It is created of 3.0 IMO, which is why it shows more than 200 commits

SPECS/golang/golang.spec
```diff
@@ -111,6 +111,9 @@ rm -f  %{gopath}/src/runtime/*.c
   ./make.bash --no-clean
 )
 
+# Nuke the final bootstrapper 04. Note: It is not used in any step under install, post, postrun
```
dagood: This shouldn't mention the number, to avoid going out of date. This spec file will have more bootstrapper layers in the future with future Go releases.


```suggestion
# Nuke the final bootstrapper. Note: It is not used in any step under install, post, postrun
```

PR #17555, opened 2026-05-29
chore: port policy YAML from previous default branch

reubeno:
With the default branch switch from `3.0` to `4.0`, we need to bring along the bot policy YAML.

As part of the port, I've added `4.0` branch/labels and also removed `2.0`-era ones.

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -1,0 +1,163 @@
+# =============================================================================
+# Resource Management policy (GitHub "Policy Service" / GitOps).
+#
+# This file is consumed by Microsoft's GitOps Policy Service, NOT by GitHub
+# Actions. The service watches the repo and reconciles behavior against the
+# rules below.
+#
+# IMPORTANT SCOPING FACTS:
+#   * The service reads .github/policies/* ONLY from the repository's DEFAULT
+#     branch (currently `4.0`) and then applies the rules to PRs/issues
+#     targeting ANY branch. You therefore cannot keep "3.0 rules" on the 3.0
+#     branch and "4.0 rules" on 4.0 -- everything lives here, on the default
+#     branch, and is discriminated by CONDITIONS (e.g. `targetsBranch`).
+#   * There is NO file- or group-level branch scope. Filenames are purely
+#     organizational. To scope a SET of rules to a branch, put a single
+#     `targetsBranch` guard in a parent task's top-level `if:` and nest the
+#     rules under its `then:` -- the top-level `if:` is ANDed, so every nested
+#     child inherits the guard (see the classic 3.0 group below).
+#   * The repo layout differs by era and uses DISJOINT top-level paths, so the
+#     area-label groups never collide:
+#       - classic (3.0/3.0-dev): SPECS*/, toolkit/
+#       - azldev  (4.0):         base/, specs/, distro/, locks/, external/, ...
+#
+# Within a task, `if:` is a list of ANDed conditions and `then:` is a list of
+# actions; both can be nested arbitrarily.
+# =============================================================================
+
+# Top-level scalars below are part of the primitive's template and intentionally
+# left empty. `id` is managed by the Policy Service; `owner`/`where`/`onFailure`/
+# `onSuccess` are optional and unused. (`where:` selects which REPOS an
+# org-distributed policy targets -- it is NOT a branch/PR filter.)
+id:
+name: GitOps.PullRequestIssueManagement
+description: GitOps.PullRequestIssueManagement primitive
+owner:
+resource: repository
+disabled: false
+where:
+onFailure:
+onSuccess:
+configuration:
+  resourceManagementConfiguration:
+    scheduledSearches: []
+    eventResponderTasks:
+      # =======================================================================
+      # GROUP: 4.0+ area labels (GATED to the 4.0 branch).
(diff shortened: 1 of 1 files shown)


.github/policies/resourceManagement.yml
```diff
@@ -0,0 +1,163 @@
...
+    eventResponderTasks:
+      # =======================================================================
+      # GROUP: 4.0+ area labels (GATED to the 4.0 branch).
+      # The single `or(targetsBranch ...)` guard scopes the ENTIRE group; every
+      # nested `if/then` inherits it.
+      # =======================================================================
+      - description: Area labels for PRs that target 4.0
+        if:
+          - payloadType: Pull_Request
+          - or:
```
jslobodzian, line 53: should this be an "and"?

.github/policies/resourceManagement.yml
```diff
@@ -0,0 +1,163 @@
...
+      # =======================================================================
+      # GROUP: 3.0-era area labels (GATED to the 3.0 branches).
+      # The single `or(targetsBranch ...)` guard in this parent `if:` scopes the
+      # ENTIRE group; every nested `if/then` below inherits it. To stop labeling
+      # a retired branch, just remove its line from the `or:` list.
+      # =======================================================================
+      - description: Area labels for the classic (SPECS/toolkit) layout, 3.0 only
+        if:
+          - payloadType: Pull_Request
+          - or:
```
jslobodzian, line 87: same question

reubeno: Good question — it's supposed to be `or`. The conditions listed directly under `if:` are implicitly `AND`'d together. The `or:` is applied to its children, so this should basically be read as:

`if payloadType==Pull_Request AND (targetsBranch(3.0) OR targetsBranch(3.0-dev)...)`

.github/policies/resourceManagement.yml
reubeno: I responded to the other comment with more details; short version is that this `or` is applying to the child. It's technically unneeded here, but if we end up with more branches for 4.0, we'll be able to add it as another sibling of the `4.0` `targetsBranch` element.

.github/policies/resourceManagement.yml
```diff
@@ -0,0 +1,163 @@
...
+            then:
+              - addLabel:
+                  label: Tools
+          - if:
+              - filesMatchPattern:
+                  pattern: "^toolkit/(imageconfigs|tools/imagegen/configuration)/"
+                  matchAny: true
+            then:
+              - addLabel:
+                  label: Schema
```
christopherco, line 129: issue: we still need the CVE file matching pattern + security label for those kinds of PRs.
https://github.com/microsoft/azurelinux/blob/1e47ea2c550c0f5cc7187614c239d83be1cdb688/.github/policies/resourceManagement.yml#L51-L57

.github/policies/resourceManagement.yml
```diff
@@ -0,0 +1,163 @@
...
+    
+          # Branch-name labels. We always add the '3.0' label for any PR targeting a
+          # 3.0 branch, and *also* add a "fasttrack/3.0" label for PRs targeting the
+          # fasttrack branch.
+          - addLabel:
```
christopherco, line 134: issue: we will still want the `3.0-dev` and `3.0` labels, since normal 3.0 PRs should generally go to `3.0-dev` branch
https://github.com/microsoft/azurelinux/blob/1e47ea2c550c0f5cc7187614c239d83be1cdb688/.github/policies/resourceManagement.yml#L85-L96

christopherco (changes requested):

.github/policies/resourceManagement.yml
reubeno: @christopherco Is that for humans or automated processes? What I'm trying to assess is "3.0" as indicating a PR (or *issue*) for the 3.0 product, as opposed to "3.0" as for a "PR merging into the `3.0` branch".

Will it mess things up by tagging 3.0-dev PRs with *both* 3.0 and 3.0-dev as a legacy sort of thing?

Alternatively, should we look at the new fields feature to disambiguate?

christopherco: Since this PR is porting over the older default branch behavior, I was expecting a preservation of previous 3.0/3.0-dev behavior (plus the additive 4.0 and removal of 2.0 as mentioned in the change description).

You bring up good questions that we should look at adjusting, but IMO that changes this PR to something larger that will need more cycles to really work through beyond a simple port to preserve prior behavior.

christopherco: And to answer the questions more directly:

> Is that for humans or automated processes? What I'm trying to assess is "3.0" as indicating a PR (or issue) for the 3.0 product, as opposed to "3.0" as for a "PR merging into the 3.0 branch".

My understanding is that it is for human processes. @jslobodzian might be able to weigh in on if there's any known automation behind these `3.0` / `3.0-dev` labels. I am personally not aware of any.
 
> Will it mess things up by tagging 3.0-dev PRs with both 3.0 and 3.0-dev as a legacy sort of thing?

Yes, but only because it's a change in behavior from what was there before. I don't have a clear idea at the moment what exactly it would mess up (if anything), but it is certainly a difference to legacy behavior.

> Alternatively, should we look at the new fields feature to disambiguate?

Very much open to new fields to further disambiguate

PR #17556, opened 2026-05-29
docs: written terms and conditions for compliance

jblanch2:


docs/terms/terms-and-conditions.md
```diff
@@ -0,0 +1,107 @@
+# MICROSOFT SOFTWARE LICENSE TERMS
```
reubeno, line 1: Should this be integrated with the `LICENSE` document at the root of the repo? Does it need to be linked from the README? Or is the requirement just that it be present *somewhere*?

reubeno (approved):

PR #17560, opened 2026-05-29
fix(python-build): disable extras and tests bconds to fix BR-unresolvable nightly failure

WithEnoughCoffee:
## Summary

Disables the `extras` and `tests` build conditionals for `python-build` via an azldev overlay to fix a BR-unresolvable nightly build failure.

## Problem

The `python-build` spec declares dynamic BuildRequires that the nightly build scheduler cannot see when planning build order:

- **`extras` bcond** (default ON via `%{undefined rhel}`) pulls `uv` into BuildRequires through `%pyproject_buildrequires -x virtualenv,uv`. Although AZL4 ships `python3-uv`, the `uv` package currently fails to build, causing BR-unresolvable failures.
- **`tests` bcond** pulls dynamic test BuildRequires (e.g. `filelock`, `pytest-mock`) via `%pyproject_buildrequires -g test`. These are discovered at build time from upstream metadata rather than statically declared, so the scheduler cannot order them — if `python-build` is scheduled before those deps are tagged in, the build fails non-deterministically.

## Fix

Add a `python-build.comp.toml` overlay setting `build.without = ["extras", "tests"]`, which azldev renders into:

- `specs/p/python-build/python-build.azl.macros` defining `%_without_extras 1` and `%_without_tests 1`
- A `%{load:...}` of that macros file plus `Source9999` in `python-build.spec`
- Release bump to `1.3.0-6` with changelog entry

`%pyproject_check_import` still runs as a basic sanity check. The component is also removed from the catch-all list in `components.toml` now that it has its own overlay directory.

## Validation

After rebuild, `python-build` resolves its buildroot and no longer depends on the unbuildable `uv` or on dynamically-discovered test deps.


base/comps/python-build/python-build.comp.toml
```diff
@@ -0,0 +1,11 @@
+# AZL does not ship python3-uv or python3-build+uv/+virtualenv extras subpackages.
```
reubeno: `uv` *is* present in AZL4, and it provides `python3-uv`. I think the real issue is that `uv` isn't building correctly.

Unless we're proposing removal of `uv`, let's try to address that root cause.

base/comps/python-build/python-build.comp.toml
```diff
@@ -0,0 +1,11 @@
...
+# The extras bcond (default ON via %{undefined rhel}) pulls uv into BuildRequires
+# via %pyproject_buildrequires -x virtualenv,uv, causing BR-unresolvable failures.
+#
+# Tests are disabled to align with RHEL — the spec notes "not all test dependencies
+# are included in RHEL: filelock, pytest-mock". The dynamic test BRs (via -g test)
```
reubeno: Does AZL4 provide filelock, pytest-mock*, though? That's relevant.

base/comps/python-build/python-build.comp.toml
```diff
@@ -0,0 +1,11 @@
...
+# create fragile ordering dependencies in the Koji nightly build tag, causing
```
reubeno: What's meant by "fragile ordering dependencies"?

reubeno (changes requested):

base/comps/python-build/python-build.comp.toml
WithEnoughCoffee: Yes, AZL4 ships both python-filelock and python-pytest-mock. The test-disable rationale was carried over from the RHEL spec comment. Should we enable the tests and then I can remove the justification?

base/comps/python-build/python-build.comp.toml
WithEnoughCoffee: packages in a nightly build tag are built in a dependency-resolved order. "Fragile ordering dependencies"  which means pythons-builds uses s %pyproject_buildrequires -g test in the dependencies that are discovered at build time, it queries the package metadata at build time to discover test dependencies like filelock and pytest-mock. This creates an implicit build order requirement — those packages must already be built and available in the tag before python-build is built. The problem is that the dependencies aren't declared statically the spec for Build requires. s, so the build scheduler can't see them upfront. If the build order happens to schedule python-build before filelock or pytest-mock are tagged in, the build fails — but only sometimes, depending on timing. That's what makes it "fragile": it works or breaks depending on non-deterministic build ordering in the nightly tag compose.  I don't think this comment is worded well I can fix that. I think something like : "undeclared implicit build-order requirements that cause non-deterministic failures when the scheduler processes packages in an unfavorable order."

base/comps/python-build/python-build.comp.toml
WithEnoughCoffee: I think that it isn't building consistently I can change the wording.

base/comps/python-build/python-build.comp.toml
WithEnoughCoffee: updated the description!

reubeno:
@WithEnoughCoffee is this PR still relevant, or is it safe to close it out (without merging)?

ddstreetmicrosoft:
> Autumn Nash (@WithEnoughCoffee) is this PR still relevant, or is it safe to close it out (without merging)?

assuming no longer required and closing.

PR #17594, opened 2026-06-01
feat(scripts): add automated QEMU ISO install verification

kgodara912:
Add scripts/auto_iso_install_qemu.py to drive an unattended Azure Linux ISO install in QEMU over the serial console and verify the installed system boots and logs in. Lets contributors iterate on the installer (Anaconda, kickstart, GRUB, LUKS) without clicking through the TUI, and gives CI a reusable smoke-test entry point.

Supports both standard and LUKS-encrypted installs, parses the Anaconda hub at runtime to locate the Root password spoke, handles first-boot and post-relabel LUKS unlocks, verifies login via an echoed marker, captures `systemd-analyze` output, then powers off.

Usage:
  python3 scripts/auto_iso_install_qemu.py -i <iso> -d <disk.qcow2>

### DIFF
--- a/scripts/auto_iso_install_qemu.py
+++ b/scripts/auto_iso_install_qemu.py
@@ -1,0 +1,523 @@
+#!/usr/bin/env python3
+"""Automate Azure Linux ISO install in QEMU over serial console.
+
+This script boots an installer ISO, watches serial output, responds to known
+prompts, waits for first boot, logs in, and verifies the installed system is up.
+
+Example:
+  python3 scripts/auto_iso_install_qemu.py \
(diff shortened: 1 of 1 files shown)


reubeno:
Hi @kgodara912 -- thanks for putting this together! It will be great to test the ISO as part of the eng process.

What would be your plan for running this as part of the Validation Service? It's okay if it's not integrated from the get-go, but would be great to understand how it's intended to join up.

christopherco:
@kgodara912 and @bhagyapathak are working together on including ISO automated testing in the validation service

kgodara912:
As this will be part of validation service hence closing PR here.

PR #17601, opened 2026-06-02
docs: extend ISO instructions to support qemu

reubeno:
Adds QEMU-based instructions for ISO boot/install in a local VM. Extends the existing Hyper-V instructions.


christopherco (approved):
:shipit:

PR #17621, opened 2026-06-04
test(container): assert OCI Config.User is unset in static-image checks

christopherco:
Add a shared static-image-checks test that fails if a container image's
OCI manifest declares an explicit Config.User (e.g. user="root" in the
kiwi <containerconfig>). An explicit user diverges from the unset default
used by AZL 3.0 and mainstream base images.

- conftest: new session fixture `oci_image_config` returning the parsed
  `skopeo inspect --config` output.
- extract: `inspect_oci_config()` helper using `skopeo inspect --config`
  (skopeo is already a declared container tool; no umoci needed).
- cases/static/test_oci_config.py: shared test gated on the `container`
  capability via `@pytest.mark.require_capability("container")` (matching
  the existing test_packages.py convention) so it runs for every
  container image family and skips VM images. Asserts `User` is absent
  from the OCI config (an explicit empty string also fails).
- images.toml: wire the three distroless images into static-image-checks
  with a `container = true` capability table so the new test covers them
  (package tests correctly skip — no runtime-package-management).

Fixes: [AB#20622](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20622)

### DIFF
--- a/base/images/images.toml
+++ b/base/images/images.toml
@@ -101,14 +101,35 @@ runtime-package-management = true
 [images.container-distroless-minimal]
 description = "Container Distroless Minimal Image"
 definition = { type = "kiwi", path = "container-base/container-base.kiwi", profile = "distroless-minimal" }
+tests.test-suites = [{ name = "static-image-checks" }]
+
+[images.container-distroless-minimal.capabilities]
+machine-bootable = false
+container = true
+systemd = false
+runtime-package-management = false
 
 [images.container-distroless-base]
 description = "Container Distroless Base Image"
 definition = { type = "kiwi", path = "container-base/container-base.kiwi", profile = "distroless-base" }
+tests.test-suites = [{ name = "static-image-checks" }]
+
+[images.container-distroless-base.capabilities]
+machine-bootable = false
+container = true
+systemd = false
+runtime-package-management = false
 
 [images.container-distroless-debug]
 description = "Container Distroless Debug Image"
 definition = { type = "kiwi", path = "container-base/container-base.kiwi", profile = "distroless-debug" }
+tests.test-suites = [{ name = "static-image-checks" }]
+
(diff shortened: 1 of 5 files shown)


christopherco:
Corresponding OCI test fixtures and static test for #17619

base/images/images.toml
```diff
@@ -95,14 +95,35 @@ runtime-package-management = true
 [images.container-distroless-minimal]
 description = "Container Distroless Minimal Image"
 definition = { type = "kiwi", path = "container-base/container-base.kiwi", profile = "distroless-minimal" }
+tests.test-suites = [{ name = "static-image-checks" }]
```
reubeno, line 104: Thanks for the metadata updates here; their omission on the distroless containers was an oversight.

base/images/tests/cases/static/test_oci_config.py
```diff
@@ -0,0 +1,32 @@
...
+Shared test (not under a ``cases/<family>/`` directory) so it applies to
+every container image family — both ``core`` (container-base) and the
+``distroless-*`` variants. The ``oci_image_config`` fixture self-skips on
+VM images, so no capability/family marker is needed here.
+"""
+
+from __future__ import annotations
+
+
+def test_no_explicit_config_user(oci_image_config: dict[str, object]) -> None:
```
reubeno, line 17: Should use a marker, e.g.: `@pytest.mark.require_capability` with an appropriate capability if there is one.

reubeno (commented):
Love the idea of the test! A couple of quick comments.

base/images/tests/cases/static/test_oci_config.py
christopherco: Added the container capability

reubeno (approved):

PR #17622, opened 2026-06-04
test(vm-base): add rootfiles regression static test

christopherco:
rootfiles was missing from vm-base.kiwi ([AB#20506](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20506)), leaving /root/.bashrc absent so root shells (e.g. `sudo su`) never sourced /etc/bashrc and fell back to a bare `bash-5.3#` prompt.

The test asserts the rootfiles package is installed and that root's .bashrc sources /etc/bashrc. Since /root/.bashrc is a %ghost created at boot by systemd-tmpfiles, the unbooted image may lack it, so the test also accepts the canonical /usr/share/rootfiles/.bashrc plus the tmpfiles.d drop-in as build-time evidence.

Fixes: [AB#20623](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20623)

### DIFF
--- a/base/images/tests/cases/vm-base/test_rootfiles.py
+++ b/base/images/tests/cases/vm-base/test_rootfiles.py
@@ -1,0 +1,101 @@
+# SPDX-License-Identifier: MIT
+"""Verify the ``rootfiles`` package and root's shell startup files on VM images.
+
+These tests assert that VM images ship ``rootfiles`` and that root will
+receive a ``.bashrc`` which sources ``/etc/bashrc``, yielding the
+expected ``[root@host ~]#`` prompt for non-login root shells (e.g.
+``sudo su``) rather than a bare ``bash-x.y#`` fallback. This guards
+against AB#20506, where ``rootfiles`` was dropped from ``vm-base.kiwi``.
+
+Why the assertions target ``/usr/share/rootfiles`` and ``tmpfiles.d``
+rather than ``/root/.bashrc`` directly: ``/root/.bashrc`` is a
+``%ghost`` entry created at boot by ``systemd-tmpfiles`` from the
+canonical copy under ``/usr/share/rootfiles/``. On an unbooted,
+statically mounted image the ghost file may not exist yet, so we verify
+the deterministic build-time evidence chain instead — the canonical
+copy, the tmpfiles.d drop-in that provisions ``/root/.bashrc`` from it,
+and ``/etc/bashrc`` itself — which together guarantee the runtime
+outcome regardless of whether the ghost file has been materialized.
+
+Path constants below are rootfs-relative (no leading slash) so they can
+be joined with ``rootfs / CONST``; failure messages prepend ``/`` to
+show the absolute on-image path. Do not add a leading slash to the
+constants — ``Path("/abs")`` discards the ``rootfs`` prefix when joined.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+# Canonical copy that systemd-tmpfiles installs into /root/.bashrc at boot.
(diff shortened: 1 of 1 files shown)


base/images/tests/cases/vm-base/test_rootfiles.py
```diff
@@ -0,0 +1,101 @@
...
+from pathlib import Path
+
+import pytest
+
+# Canonical copy that systemd-tmpfiles installs into /root/.bashrc at boot.
+ROOTFILES_SKEL_BASHRC = "usr/share/rootfiles/.bashrc"
+# tmpfiles.d drop-in whose `C /root/.bashrc ... /usr/share/rootfiles/.bashrc`
+# directive provisions root's .bashrc on first boot.
+TMPFILES_CONF = "usr/lib/tmpfiles.d/rootfiles.conf"
+ROOT_BASHRC = "/root/.bashrc"
```
ddstreetmicrosoft, line 37: This leading '/' violates the (probably unnecessary and/or AI slop?) documentation just 15 lines above.

ddstreetmicrosoft (commented):
I'm not entirely sure how useful this really is - if i'm understanding it correctly, it essentially just verifies that the files that the 'rootfiles' package currently ships, in the exact form that the files current are, is what's in the image? Is that actually a useful test? This seems like a test that's so brittle it will (possibly) produce false failures if the 'rootfiles' package is even slightly changed?
I'll leave this "review" as a simple comment, but personally I don't see the value in this.

ddstreetmicrosoft:
> if i'm understanding it correctly, it essentially just verifies that the files that the 'rootfiles' package currently ships

to put my feedback another way - if the problem is (as stated in the commit description) "root shells (e.g. `sudo su`) never sourced /etc/bashrc and fell back to a bare `bash-5.3#` prompt", are we actually getting any benefit from the additional, very brittle implementation-specific tests?

In fact, are we actually getting any benefit from the basic `test_rootfiles_package_installed` test, that does nothing more than tests if the 'rootfiles' package is installed in the image? If that test _does_ provide some value, why are we testing _only_ the 'rootfiles' package? Why not test for _all_ the packages currently installed in the image?

My point here is *not* to be pedantic. My point is that tests that verify the _current_, _implementation-specific_ details of a package are not actually very useful, nor is testing for the simple existence of a package in an image. If such tests were actually useful, then they should be added for *ALL* packages, not only ones that happen to draw attention (like this case, where the 'rootfiles' package happened to be missing from the image).

The reason this kind of tests should _not_ be added is because the simple existential test (i.e. is the 'rootfiles' package installed?) is, to be blunt, pointless, and will only serve as a cause of extra work to 'fix' the test if we ever decide to remove the package, for whatever reason; and the additional rootfiles-package-implementation-specific tests are *not* appropriate integration tests, because they simply verify that the 'rootfiles' package beahves exactly as it currently is implemented.

If we want to add any test at all, it should be *at most* a simple test to verify that a root shell does source `/etc/bashrc` *and nothing else*. If that's the problem, test for that...and nothing else.

christopherco:
Apologies, I took this out of draft prematurely. When I first made this PR a couple months ago, I encountered a missing libguestfs dependency that prevents running these offline static vm-image tests in my Azure Linux 4 dev environment. That's not fixed yet, so this was not ready for review, but I forgot when going through my old draft PRs.

After reading the feedback, I'll close this PR out for now. I agree with the core feedback that the test could be different to gain more value, specifically targeting the root-shell behavior (which likely would require implementing in a different test framework than this offline static test setup). Also, I do think a full package list static image test could be useful closer to GA, to catch unplanned package dependency changes that make their way into our image.

PR #17624, opened 2026-06-04
feat: add libguestfs and supermin

christopherco:
Adds the `libguestfs` package (library + `guestfish`/`virt-*` tools for
accessing and modifying VM disk images) to Azure Linux 4.0, imported from
Fedora dist-git with the full Fedora binding set and no overlays to keep
divergence minimal for regular Fedora-baseline refreshes.

This is a patch series of three self-contained commits:

1. **fix(hivex): re-enable OCaml bindings** — Reverts the earlier
   workaround (bf151bb53d) that disabled hivex's OCaml bindings due to a
   broken bootstrap OCaml toolchain. The bootstrap is now resolved, 
   so hivex again provides `ocaml-hivex-devel`, a mandatory libguestfs
   build dependency. The dedicated comp file is removed and hivex
   returns to an inline entry.

2. **feat(supermin): add supermin component** — Adds the appliance
   builder libguestfs depends on, with an overlay patch teaching
   supermin's package-manager detection to recognize `ID=azurelinux`
   (upstream only checks a hard-coded ID allow-list, ignoring `ID_LIKE`).
   The patch is flagged for upstream submission.

3. **feat(libguestfs): add libguestfs component** — Adds libguestfs with
   the full Fedora binding set (OCaml, Perl, Python, Ruby, Lua, PHP,
   GObject/Vala) and no overlays.

Fixes: [AB#20621](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20621)


base/comps/supermin/0001-Recognize-Azure-Linux-as-an-rpm-dnf-distro.patch
```diff
@@ -0,0 +1,34 @@
+From: Azure Linux <azurelinux@microsoft.com>
```
reubeno, line 1: Is the code in question just detecting a Fedora-like distro? Would looking at ID_LIKE be a viable approach rather than adding yet another distro to the list?

christopherco: Good callout. Right now, the supermin `os_release.ml` parser only extracts `ID` today, so there's no `ID_LIKE` accessor yet. 

https://github.com/libguestfs/supermin/blob/master/src/os_release.ml#L30-L32
https://github.com/libguestfs/supermin/blob/master/src/ph_rpm.ml#L35

Adding `ID_LIKE` properly means extending the parser + new accessor + reworking `fedora_detect()` to match on the `ID_LIKE` field, which is the kind of change that should be for upstream discussion.

For now, I propose we keep this minimal list-add as a downstream patch to unblock libguestfs on AZL4 so we can run static vm image testing through `azldev` in the AZL4 dev environment, and I can also raise the ID_LIKE generalization in the upstream discussion for supermin as a real architectural fix.

christopherco:
Superseded by #18041

PR #17636, opened 2026-06-05
fix(librsvg2): resolve BR-unresolvable nightly build failure

WithEnoughCoffee:
## Summary

Add overlays to resolve the librsvg2 buildroot resolution failure (mock exit 30) in the nightly build.

### Changes

- Enable bundled Rust dependencies (`%global bundled_rust_deps 1`) to avoid dynamic crate BRs that pull in unpackaged dav1d via rust-image+avif-native
- Replace `rust-toolset` BR with `cargo-rpm-macros` (AZL does not ship the Fedora rust-toolset metapackage)

> Note: AVIF support was already disabled unconditionally in release -7 (`fix(librsvg2): disable AVIF support to remove dav1d dependency`); this PR does not change AVIF handling.

### Testing

- Build succeeded in mock chroot
- Smoke-test: `rsvg-convert --version` runs correctly (v2.61.0)

### References

- Split from [AB#20330](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20330)

Fixes: [AB#20337](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20337)


tobiasb-ms:
question(blocking): In the description you call out:
>Unconditionally disable AVIF support (-Davif=disabled) since dav1d system libraries are not available in AZL4

I don't see that this changed in any way. What am I missing?

tobiasb-ms:
question(blocking): Did you consider bringing in the dependencies (other than the metapackage) rather than using vendored packages?

WithEnoughCoffee:
> question(blocking): In the description you call out:
> 
> > Unconditionally disable AVIF support (-Davif=disabled) since dav1d system libraries are not available in AZL4
> 
> I don't see that this changed in any way. What am I missing?

Sorry, that bullet is misleading. The unconditional  -Davif=disabled  was actually landed earlier in release -7 (commit │ 211fd81633 , "fix(librsvg2): disable AVIF support to remove dav1d dependency") via the existing  "Disable AVIF support"  overlay. This PR doesn't touch AVIF at all — its only changes are flipping  bundled_rust_deps 0 to 1  and swapping the  rust-toolset  BR for  cargo-rpm-macros . I'll drop that bullet from the description/commit message so it reflects what actually changed.

WithEnoughCoffee:
> question(blocking): Did you consider bringing in the dependencies (other than the metapackage) rather than using vendored packages?

 Yes — vendoring was the deliberate choice here, for a few reasons:
 
  -  dav1d isn't packaged in AZL. Neither the  dav1d  C library nor a  rust-dav1d  binding exists in the distro, and that's the actual  buildroot blocker (the dynamic cargo BRs pull  rust-image 's  avif-native  feature → dav1d). AVIF stays disabled regardless of how we source the other crates.
  
  - It matches upstream/RHEL. RHEL already builds librsvg2 with  bundled_rust_deps 1  ( %if 0%{?rhel} ). This change just makes AZL follow that same code path — minimal divergence from the upstream spec.
  

- Vendoring is an established AZL pattern — ~75 specs ship a  *-vendor.tar.xz , and librsvg2 already carries  Source1:   librsvg-%{version}-vendor.tar.xz .
 
 Packaging the deps individually would mean bringing in the full transitive crate tree plus dav1d (C) and a Rust binding for it — a large surface that still wouldn't enable AVIF. If we later package dav1d, we can revisit enabling system crates + AVIF, but for unbreaking the  KDE/nightly chain now, the vendored path is the smallest, RHEL-aligned fix.

tobiasb-ms (approved):

base/comps/librsvg2/librsvg2.comp.toml
```diff
@@ -1,5 +1,20 @@
...
+[[components.librsvg2.overlays]]
+description = "Enable bundled Rust dependencies — AZL does not ship dav1d crate needed by rust-image+avif-native"
+type = "spec-search-replace"
+regex = '%global\s+bundled_rust_deps\s+0'
+replacement = '%global bundled_rust_deps 1'
+
+[[components.librsvg2.overlays]]
+description = "Replace rust-toolset BR with cargo-rpm-macros — AZL does not ship rust-toolset metapackage"
+type = "spec-search-replace"
+regex = 'BuildRequires:\s+rust-toolset'
```
reubeno, line 14: Since we have overlays to remove and add tags -- we should prefer those to regexes.

specs/l/librsvg2/librsvg2.spec
reubeno, line 158: How are we producing the bundled vendor content? I missed some of the earlier discussion on this, but do we have a way to produce the vendor content via tooling/services in a way that does not require uploading a tarball produced on a development machine?

base/comps/librsvg2/librsvg2.comp.toml
```diff
@@ -1,5 +1,20 @@
 [components.librsvg2]
 
+# Use vendored Rust dependencies (like RHEL) to avoid dynamic cargo BR generation
+# which pulls in rust-image+avif-native → dav1d, a crate not available in AZL.
```
reubeno, line 4: We *did* have `dav1d` in the distro but had intentionally removed it. Were you able to convince yourself that the reasons we previously removed it wouldn't prevent us from reintroducing it in vendored form?

WithEnoughCoffee: AVIF is disabled ( -Davif=disabled ), which gates the entire dav1d dependency. Verified on a full build 
meson logs  Dependency dav1d skipped: feature avif disabled , the dav1d crate is never compiled (0  Compiling dav1d ), and  librsvg-2.so  has no  libdav1d  in NEEDED and zero dav1d strings. No RPM gains a dav1d/avif  Requires . The dav1d crate source remains in the  cargo vendor  tree (so the RPM has  Provides: bundled(crate(dav1d))  metadata), but it's inert — not built or linked. This matches Fedora/RHEL, which also disable AVIF here.

base/comps/librsvg2/librsvg2.comp.toml
WithEnoughCoffee: I tried the remove-tag + insert-tag approach, but it renders a messy BR block. Upstream has:
%if 0%{?bundled_rust_deps}
BuildRequires:  rust-toolset
%else
BuildRequires:  cargo-rpm-macros
%endif
 spec-remove-tag rust-toolset  leaves the  %if  branch empty, and  spec-insert-tag cargo-rpm-macros  appends an unconditional duplicate (the  %else  already has it):
%if 0%{?bundled_rust_deps}
%else
BuildRequires:  cargo-rpm-macros
%endif
BuildRequires: cargo-rpm-macros   # duplicate
Tag overlays can't remove the  %if/%else/%endif  wrapper (not tags) and multi-line regex isn't supported, so they can't clean this up. A single anchored value-swap of the unique  rust-toolset  line renders cleanly:
%if 0%{?bundled_rust_deps}
BuildRequires:  cargo-rpm-macros
%else
BuildRequires:  cargo-rpm-macros
%endif
The regex is anchored ( BuildRequires:\s+rust-toolset ) per the earlier reviewer feedback. So here regex is the cleaner. Do you think its ok to keep it?

specs/l/librsvg2/librsvg2.spec
WithEnoughCoffee: cargo vendor --versioned-dirs && tar Jcvf ../librsvg-%{version}-vendor.tar.xz vendor/
 Cargo vendor  run against the upstream  Cargo.lock , tarred deterministically, then uploaded to the lookaside and pinned by SHA512. It's reproducible given a matching cargo toolchain — not a hand-assembled artifact, and mirrors how other vendored-Rust/Go packages in AZL are handled.
Today the generate-and-upload step is done once by a maintainer; there isn't currently a service that produces and pushes the vendor tarball automatically. Removing that manual step would be a broader tooling change (a pipeline step running  cargo vendor  to the lookaside) that applies to all vendored packages, not just librsvg2. Happy to file that as a separate work item if you'd like to pursue it.

christopherco (approved):

PR #17652, opened 2026-06-08
fix(qt6-qtmultimedia): drop gstreamer-plugins-bad BR to unbreak KDE dep chain

WithEnoughCoffee:
## Summary

Removes the `pkgconfig(gstreamer-plugins-bad-1.0)` BuildRequires from qt6-qtmultimedia. The built RPM was linking against `libgstplay`/`libgstphotography` from `gstreamer1-plugins-bad-free`, which is not shipped in AZL4 — making the package uninstallable and breaking all transitive consumers.

## Problem

Koji nightly build of **libplasma** (task 2590944) fails at buildroot resolution (mock exit 30). The dep chain:

```
libplasma BR: cmake(KF6Parts)
  → kf6-kparts-devel → kf6-ktextwidgets → qt6-qtspeech
    → qt6-qtmultimedia → libgstplay-1.0.so.0 (not in AZL4)
```

## Fix

Drop the gstreamer-plugins-bad BR and remove the corresponding `%files` entries for the gstreamer media backend plugin that is no longer built. Base gstreamer plugins and PulseAudio/PipeWire backends remain functional.

## Validation

After rebuild of qt6-qtmultimedia, libplasma buildroot resolution succeeds.

Split out of [AB#20330](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20330).

### DIFF
--- a/base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
+++ b/base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
@@ -7,3 +7,32 @@
 # Part of the distro-wide effort to remove ffmpeg/libav* from Azure Linux.
 [components.qt6-qtmultimedia]
 build.without = ["ffmpeg"]
+
+# gstreamer1-plugins-bad-free is not shipped in AZL4. Remove the BR to avoid
+# linking against libgstplay/libgstphotography (unresolved runtime deps that
+# break all transitive consumers up to libplasma).
+[[components.qt6-qtmultimedia.overlays]]
+description = "Remove gstreamer-plugins-bad BR — gstreamer1-plugins-bad-free not shipped in AZL"
+type = "spec-remove-tag"
+tag = "BuildRequires"
+value = "pkgconfig(gstreamer-plugins-bad-%{gst})"
+
+# Without gstreamer-plugins-bad headers, the GStreamer media plugin
+# (libgstreamermediaplugin.so) is not built. Remove it from %files
+# along with the empty multimedia plugin directory.
+# Note: replacement leaves blank lines — the tool operates per-line and
+# cannot delete lines entirely. This is cosmetic only; RPM ignores blanks
+# in %files.
+[[components.qt6-qtmultimedia.overlays]]
+description = "Remove empty multimedia plugin dir from %files — no plugins install here (gst-bad disabled, ffmpeg gated by bcond), so %dir would fail the build"
+type = "spec-search-replace"
+section = "%files"
+regex = '^%dir %\{_qt6_plugindir\}/multimedia$'
+replacement = ''
+
+[[components.qt6-qtmultimedia.overlays]]
+description = "Remove libgstreamermediaplugin.so from %files (not built without gstreamer-plugins-bad)"
+type = "spec-search-replace"
+section = "%files"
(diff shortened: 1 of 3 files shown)


base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
```diff
@@ -7,3 +7,46 @@
...
+# Alternatives considered:
+#   - Ship gstreamer1-plugins-bad-free: large dep tree, added attack surface,
+#     not justified for cloud workloads.
+#   - Drop cmake(KF6Parts) from libplasma: only a band-aid — other consumers
+#     of kf6-kparts-devel would still be broken.
+#   - Remove qt6-qtmultimedia from qt6-qtspeech: functional regression.
+# Removing the BR here fixes the root cause for the entire dep chain with
+# minimal divergence — only the gstreamer media backend plugin is lost.
+#
+# AB#20330 / https://52.249.25.247/koji/taskinfo?taskID=2590944
```
reubeno: Please remove these references from the comment; they're more appropriate for the PR description than the code left behind.

base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
```diff
@@ -7,3 +7,46 @@
...
+# (libgstreamermediaplugin.so) is not built. Remove it from %files
+# along with the empty multimedia plugin directory.
+# Note: replacement leaves blank lines — the tool operates per-line and
+# cannot delete lines entirely. This is cosmetic only; RPM ignores blanks
+# in %files.
+[[components.qt6-qtmultimedia.overlays]]
+description = "Remove gstreamer media plugin dir from %files (not built without gstreamer-plugins-bad)"
+type = "spec-search-replace"
+section = "%files"
+regex = '^%dir %\{_qt6_plugindir\}/multimedia$'
```
reubeno, line 30: What's the reason to remove the directory? Is it a problem for it to exist? (Is there a risk that some other package may count on the existence of this dir?)

base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
WithEnoughCoffee: Done — removed the AB# and Koji references. The detailed rationale/alternatives are already in the PR description.

base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
WithEnoughCoffee: Yes — without gstreamer-plugins-bad, `libgstreamermediaplugin.so` is not built. And with ffmpeg also disabled (the other bcond), `libffmpegmediaplugin.so` is inside a dead `%if %{with ffmpeg}` block. So **no plugins install into `%{_qt6_plugindir}/multimedia/`** — the directory isn't created during `%install`, meaning `%dir` in `%files` would cause an RPM build failure ("file not found"). This was confirmed during the original build verification.

binujp (approved):
Looks good. Given this was building and installing earlier runtime deps should not be a problem.

PR #17653, opened 2026-06-08
feat(dkms): upgrade to 3.4.1

ellie-di:
This PR upgrades `dkms` from 3.3.0 to 3.4.1 on the 4.0 branch.

Fedora's `f43` dist-git already ships 3.4.1 at commit [`28b729e`](https://src.fedoraproject.org/rpms/dkms/c/28b729e0aa8b5f3aad436d3a66b2fe8b309b7624), but the global Fedora snapshot pin still resolves to 3.3.0. This PR uses a per-component `upstream-commit` pin to pick up 3.4.1 for `dkms` only.


base/comps/components.toml
```diff
@@ -320,7 +320,7 @@ includes = ["**/*.comp.toml", "component-check-disablement.toml", "component-min
 [components.disruptor]
 [components.distribution-gpg-keys]
 [components.djvulibre]
-[components.dkms]
+# dkms moved to dedicated comps/dkms/dkms.comp.toml to pin a newer upstream commit (3.4.1)
```
christopherco: Please remove this comment, as it will eventually get stale, and the dedicated overlay is self-evident with the dedicated `dkms.comp.toml`.

christopherco (changes requested):
Change itself looks good, and builds cleanly in our build infra.

Only request is to please address the comment.

base/comps/components.toml
ellie-di: removed.

christopherco (approved):

PR #17672, opened 2026-06-10
feat(wsl): include azure-cli in WSL image

mfrw:
Add azure-cli to the WSL image package list so users have the Azure
management CLI available out of the box. The azure-cli component is
already defined in base/comps/components.toml.

### DIFF
--- a/base/images/wsl/wsl.kiwi
+++ b/base/images/wsl/wsl.kiwi
@@ -32,6 +32,7 @@
     <packages type="image" patternType="plusRecommended">
         <package name="attr"/>
         <package name="acl"/>
+        <package name="azure-cli"/>
         <package name="bash-color-prompt"/>
         <package name="bash-completion"/>
         <package name="bind-utils"/>


reubeno:
@mfrw How would you evaluate the trade-off between pre-installing the (Linux) `az` CLI vs. allowing any installed `az` CLI from the native Windows side take precedence? (What do other WSL images do?)

mfrw:
> @mfrw How would you evaluate the trade-off between pre-installing the (Linux) `az` CLI vs. allowing any installed `az` CLI from the native Windows side take precedence? (What do other WSL images do?)

This is an excellent point. There are almost all Pros of falling back to the `azcli` which is installed on the host. The only down side is e.g if someone would want to do an operation like:
```bash
az storage blob upload --file ./data
```
Since, `az.exe` run in the Windows NT subsytem context, it cannot natively parse the Linux VHDX filepath unless it is explicity translated via `wslpath -w`

Also, other WSL images do not install azcli by default.

While we are here, i also wanted for slightly better devex, we should atleast include `tmux & git` in our WSL image natively, what are your thoughts ?

tobiasb-ms:
> > @mfrw How would you evaluate the trade-off between pre-installing the (Linux) `az` CLI vs. allowing any installed `az` CLI from the native Windows side take precedence? (What do other WSL images do?)
> 
> This is an excellent point. There are almost all Pros of falling back to the `azcli` which is installed on the host. The only down side is e.g if someone would want to do an operation like:
> 
> ```shell
> az storage blob upload --file ./data
> ```
> 
> Since, `az.exe` run in the Windows NT subsytem context, it cannot natively parse the Linux VHDX filepath unless it is explicity translated via `wslpath -w`
> 
> Also, other WSL images do not install azcli by default.
> 
> While we are here, i also wanted for slightly better devex, we should atleast include `tmux & git` in our WSL image natively, what are your thoughts ?

I agree about the `blob` scenarios but on balance I suggest we **don't** add `azure-cli` to the WSL image. People can install if needed, but generally auth is better if you use the host version of `az`.

Do we consider `WSL` a "dev" image? If so I could see that but I'd say by default we wouldn't have devish tools.

mfrw:
> I agree about the `blob` scenarios but on balance I suggest we **don't** add `azure-cli` to the WSL image. People can install if needed, but generally auth is better if you use the host version of `az`.
> 
> Do we consider `WSL` a "dev" image? If so I could see that but I'd say by default we wouldn't have devish tools.

In my opinion yes - I consider WSL a dev image to tryout Azl (this is my personal opinion though). But in anycase, i am open to suggestion w.r.t what should go in the default and what should no.
I mean, atleast having tmux / git might make folks lives slightly better as the default (ubuntu) WSL does have `tmux`.

christopherco (changes requested):
Holding this change as we're internally discussing whether to include or not

PR #17684, opened 2026-06-10
ci: enable best-effort delta builds using ADO pipelines

dmcilvaney:
Automatic source upload is being reworked, but we don't need to wait on that to enable the initial version of delta builds. Disable the source upload component in the delta build pipeline, and configure it to calculate the commit range using the ADO REST API instead of git commands (since this will be post-merge).

To make future pipeline work easier, split the common components into their own template.

Also fix a bug where the environment variable names were shadowed by ADO predefined variables.

### DIFF
--- a/.github/workflows/ado/package-build.yml
+++ b/.github/workflows/ado/package-build.yml
@@ -1,18 +1,21 @@
 # Microsoft Corporation
 #
-# Wrapper pipeline — passed to ADO as the entry point for the source-upload
+# Wrapper pipeline — passed to ADO as the entry point for the package-build
 # pipeline. This file owns all OneBranch-specific wiring (governed templates
 # repo, Official vs NonOfficial variant, featureFlags) and delegates the actual
 # stages/jobs/steps to the raw stages template at:
-#   .github/workflows/ado/templates/sources-upload-stages.yml
+#   .github/workflows/ado/templates/package-build-stages.yml
 #
 # Authenticates via Workload Identity Federation (OIDC) and calls the Control
-# Tower 'prcheck' API to upload changed component sources. prcheck returns early
-# (no upload) on pull-request triggers -- unmerged code should not consume
-# capacity.
+# Tower APIs to run the v1 post-merge delta build (stopgap until source upload
+# is reworked):
+#   1. Resolve the (target, source) commit range for this push from the
+#      previous CI build (ADO Builds API).
+#   2. Submit official package builds for the components that changed across
+#      that range, via Workload Identity Federation (OIDC) to Control Tower.
 #
 # The setup + change-detection + validation steps are shared with the
-# package-build pipeline via templates/steps/common-steps.yml.
+# source-upload pipeline via templates/steps/common-steps.yml.
 #
 # Helper scripts live under:
 # - scripts/ci/control-tower/ - (Control Tower-specific)
@@ -69,14 +72,16 @@ extends:
         enabled: false
 
     stages:
-      - template: /.github/workflows/ado/templates/sources-upload-stages.yml@self
+      - template: /.github/workflows/ado/templates/package-build-stages.yml@self
         parameters:
           outputDirectory: $(Build.ArtifactStagingDirectory)/output
-          artifactBaseName: sourceupload
+          artifactBaseName: packagebuild
           containerImage: mcr.microsoft.com/onebranch/azurelinux/build:3.0
           poolType: linux
           serviceConnection: CT-Endpoints-Access-ServiceConnection-DEV
           variableGroup: ControlTower-PRCheck
-          # Must exceed the script's --poll-timeout-seconds (default 7200s = 120m)
+          # Control Tower package target for the 4.0 branch.
+          packageTarget: azl4
+          # Must exceed the script's --poll-timeout-seconds (default 600s = 10m)
           # with enough headroom for setup steps and the final API call.
-          timeoutInMinutes: 150
+          timeoutInMinutes: 25
--- a/.github/workflows/ado/templates/steps/common-steps.yml
+++ b/.github/workflows/ado/templates/steps/common-steps.yml
@@ -1,0 +1,242 @@
+# Microsoft Corporation
+#
+# Shared step template for the Control Tower integration pipelines (package
+# build + source upload). Spliced into each pipeline's single job via:
+#
+#   steps:
+#     - template: steps/common-steps.yml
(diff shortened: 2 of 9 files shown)


dmcilvaney:
Note: Planning on renaming the file to something more accurate, but will need to combine that with ADO changes, leaving as-is to allow testing in the meantime.

.github/workflows/ado/templates/sources-upload-stages.yml
```diff
@@ -113,9 +79,56 @@ stages:
...
+          # Resolve the (target, source) commit range for this post-merge build:
+          #   source = the commit that triggered this run (Build.SourceVersion).
+          #   target = the previous CI build's commit on this branch, looked up
+          #            via the ADO Builds API. Pairing each build with the
+          #            one immediately before it keeps concurrent runs from
+          #            overlapping and captures every commit a rebase merge
+          #            appends. Runs after the dependency-install step because
+          #            it imports the azure-devops SDK. See
+          #            scripts/ci/ado/determine_commit_range.py for the full
+          #            rationale and the first-run (source^1) fallback.
```
PawelWMS: Thinking about the different ways to calculate the delta between a merge and a PR, and about what we're doing during the PR check vs post-merge, I think we might as well create a separate pipeline for the post-merge pipeline under `.github/workflows/ado` with its own template. The PR check has a reason to call CT's `prcheck` endpoint - not so sure about the post-merge pipeline (I don't want to expand its responsibilities beyond actual PR checks - I'm planning to create a separate one for promotion of scanned sources).

What do you think? I'm fine re-using the current ADO pipeline definition with tweaks like a re-name or making handle only the post-merge workflow.

dmcilvaney: Will do.

.github/workflows/ado/templates/steps/common-steps.yml
```diff
@@ -0,0 +1,228 @@
...
+    displayName: "Authenticate pip"
+    inputs:
+      artifactFeeds: "azl/ControlTowerFeed"
+
+  - script: |
+      set -euo pipefail
+
+      echo "##[group]Mock"
+      tdnf install -y mock mock-rpmautospec python3-chardet
+      sudo usermod -aG mock "$(whoami)"
```
tobiasb-ms, line 63: issue(non-blocking): I think this needs a `newgroup mock`, or am I missing something?

.github/workflows/ado/package-build.yml
```diff
@@ -0,0 +1,87 @@
...
+          artifactBaseName: packagebuild
+          containerImage: mcr.microsoft.com/onebranch/azurelinux/build:3.0
+          poolType: linux
+          serviceConnection: CT-Endpoints-Access-ServiceConnection-DEV
+          variableGroup: ControlTower-PRCheck
+          # Control Tower package target for the 4.0 branch.
+          packageTarget: azl4
+          # Must exceed the script's --poll-timeout-seconds (default 600s = 10m)
+          # with enough headroom for setup steps and the final API call.
+          timeoutInMinutes: 60
```
tobiasb-ms: issue(non-blocking): 60 minutes is a *long* timeout given the default of mentioned above of 10 minutes + overhead. Suggest maybe 15?
```suggestion
          timeoutInMinutes: 15
```

tobiasb-ms (changes requested):

.github/workflows/ado/templates/steps/common-steps.yml
dmcilvaney: afaik the mock rpm adds it automatically. fwiw its working in practice.

dmcilvaney: https://github.com/microsoft/azurelinux/blob/35ec15a9a46420c2f7c70896d731a2a9646cd334/SPECS/mock/mock.spec#L227

```
root [ /workdir ]# grep 'mock' /etc/group
root [ /workdir ]# tdnf -y install mock >/dev/null 2>&1
root [ /workdir ]# grep 'mock' /etc/group
mock:x:135:
root [ /workdir ]# 
```

tobiasb-ms: I actually meant `newgrp`. Generally when I've installed `mock` (or `docker`), even after the group is created and I'm added to it I've needed `newgrp` to convince my current session that I'm in the group. Otherwise I've needed to create a new session.

If it's working in practice, I won't block on it but I'm curious as to why my experience is always different.

tobiasb-ms (approved):

.github/workflows/ado/templates/steps/common-steps.yml
```diff
@@ -0,0 +1,228 @@
...
+      echo "$range_output"
+
+      source_commit="$(sed -n 's/^sourceCommit=//p' <<< "$range_output")"
+      target_commit="$(sed -n 's/^targetCommit=//p' <<< "$range_output")"
+      if [[ -z "$source_commit" || -z "$target_commit" ]]; then
+        echo "##[error]determine_commit_range.py did not emit a source/target commit range."
+        exit 1
+      fi
+
+      echo "##vso[task.setvariable variable=sourceCommit]$source_commit"
```
jslobodzian: Consider making this and the target_commit readonly

scripts/ci/ado/determine_commit_range.py
```diff
@@ -0,0 +1,227 @@
...
+    target_commit: str | None = None
+
+    try:
+        builds = _fetch_recent_builds(
+            definition_id=args.definition_id,
+            branch=args.branch,
+            top=args.top,
+        )
+        target_commit = _select_baseline(builds, args.current_build_id)
+    except (ClientException, OSError, RuntimeError) as exc:
```
jslobodzian, line 211: _fetch_recent_builds  catches  ClientException | OSError | RuntimeError  and silently falls back to  source^1 . If   SYSTEM_ACCESSTOKEN  permissions are wrong, or the SDK is misconfigured, every run degrades to single-commit delta  — silently. The weekly true-up becomes the sole safety net.

 Suggestion: Distinguish "no prior build found" (benign) from "query failed" (actionable). Log an  ##[warning]  for  the latter so it surfaces in the pipeline UI, or fail closed on auth errors.

jslobodzian (approved):

jslobodzian (approved):

christopherco (approved):

PR #17727, opened 2026-06-16
chore(comps): annotate backport overlays with category metadata

liunan-ms:
This PR annotates the overlays surfaced by the overlay classifier with structured `[metadata]` (a `category` plus, for genuine backports, the verified upstream `commits`), so future Fedora rebases can mechanically identify and retire overlays that have landed upstream. It also pins `azldev` to a version that understands these annotations and the per-file `overlay-files` layout, and regenerates the authoritative JSON schema to match.

### Project config

- `base/comps/components.toml` gains a project-wide `[default-component-config]` with `overlay-files = ["overlays/*.overlay.toml"]`, so every component discovers its own `overlays/*.overlay.toml` without redeclaring the glob.

### Per-file overlay migration (24 components)

ant, apache-ivy, atlas, cloud-init, cpio, glade, gnulib, grub2, gt, intel-qpl, javapackages-tools, kdump-utils, maven, maven4, rust-podman-sequoia, sos, sssd, vamp-plugin-sdk, xbean, xclock, xmvn, xmvn5, xsimd, zbar — inline overlays were moved into per-logical-change `overlays/*.overlay.toml` files, each carrying one file-level `[metadata]` block. Multi-commit backports (e.g. `javapackages-tools` openjdk21 drop) consolidate related upstream commits into a single file via `commits = [url1, url2, …]`. Patch `source` paths are rewritten to `../<patch>` to point back at the component root.

### Category spread

Most overlays are `upstream-backport` (with verified upstream `commits`). A few non-backport overlays surfaced alongside them are labelled accordingly: `azl-pruning` (e.g. grub2 xen-module disable), `azl-security-compliance` (grub2 os-prober reverts — restoring upstream GRUB's secure default Fedora diverged from), `azl-compatibility`, `azl-branding-policy`, and `azl-temp-workaround` for the `gnulib` source-URL filter shim and the `gt` `-std=gnu89` compiler workaround, which are alternative/transient fixes rather than backports of their referenced Fedora commits.

## Validation

- Pure annotation/restructuring: the rendered spec is byte-identical before and after (verified via before/after `prepare-sources` diffs for every migrated component).
- `azldev comp update` reports no lock drift (metadata is excluded from fingerprints; the `overlay-files` restructuring keeps the same input set).
- `azldev comp render` all passed in PR check.


base/comps/cpio/cpio.comp.toml
```diff
@@ -1,14 +1,5 @@
...
-
-[[components.cpio.overlays]]
-description = "Drop the `rm -f $RPM_BUILD_ROOT%{_libexecdir}/rmt` line from %install. cpio's build tree no longer ships its own `rmt` helper under `_libexecdir`, so the deletion is a stale leftover. Mirrors the same Fedora rawhide commit (84d5a4fe) referenced by the BR-drop overlay above."
-type = "spec-search-replace"
-section = "%install"
-regex = "rm -f \\$RPM_BUILD_ROOT%\\{_libexecdir\\}/rmt"
-replacement = ""
+# Overlays live as one-logical-change-per-file under ./overlays/.
+# See https://github.com/microsoft/azure-linux-dev-tools/blob/main/docs/user/reference/config/overlays.md#per-file-overlay-format-overlaytoml
+overlay-dir = "overlays"
```
dmcilvaney: It would be great to update the AI instructions to explain when to use this pattern vs. not, and what sort of data to add to each overlay.

liunan-ms: Yes, I'll add a SKILL.md and instruction in a separate PR for using this pattern and suggesting annotations for new overlays.

base/comps/ant/overlays/0001-remove-openjdk21-jdk-binding.overlay.toml
```diff
@@ -0,0 +1,14 @@
+# Backport of upstream Fedora rawhide commit 4ca7a3b: remove openjdk21 binding.
+# AZL only ships java-25-openjdk; there is no java-21-openjdk, so the
+# ant-openjdk21 subpackage has an unsatisfiable dependency.
+[metadata]
+category = "backport-dist-git"
+upstreamable = false
```
reubeno: For something that *was* upstreamed, what do we want to say for `upstreamable`?

Our non-upstreamable changes are debt that we carry. A backported patch isn't really the same kind of debt.

Makes me wonder if `upstream-status` should have been an enum, e.g.: `inapplicable` vs. `upstream` vs. `not-started`.

Without that we'll presumably need to look for non-upstreamable changes and also filter out certain categories, which seems messier.

base/comps/apache-ivy/overlays/0002-remove-buildrequires-ant-openjdk21.overlay.toml
```diff
@@ -0,0 +1,25 @@
+# AZL ships only java-25-openjdk; there is no java-21-openjdk, so the
+# upstream references to ant-openjdk21 and java-21-openjdk are
+# unsatisfiable. Retarget to the openjdk25 variant.
+[metadata]
+category = "azl-pruning"
+upstreamable = false
```
reubeno: Upstream will need to move to 25, so I think it is.

base/comps/apache-ivy/apache-ivy.comp.toml
```diff
@@ -1,19 +1,3 @@
...
-    { type = "spec-search-replace", regex = "patch -p1 -l < %{SOURCE3}", replacement = "patch -p1 -l < %{SOURCE3}\npatch -p1 -l < %{SOURCE4}\npatch -p1 -l < %{SOURCE5}" },
-
-    # AZL ships only java-25-openjdk; there is no java-21-openjdk, so the
-    # upstream references to ant-openjdk21 and java-21-openjdk are
-    # unsatisfiable. Retarget to the openjdk25 variant.
-    { description = "Remove BuildRequires on ant-openjdk21 — AZL does not ship java-21-openjdk", type = "spec-remove-tag", tag = "BuildRequires", value = "ant-openjdk21" },
-    { description = "Add BuildRequires on ant-openjdk25 in place of ant-openjdk21", type = "spec-add-tag", tag = "BuildRequires", value = "ant-openjdk25" },
-    { description = "Retarget JAVA_HOME from java-21-openjdk to java-25-openjdk in %build", type = "spec-search-replace", section = "%build", regex = "java-21-openjdk", replacement = "java-25-openjdk" },
-]
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Does this comment need to be here?

base/comps/atlas/atlas.comp.toml
```diff
@@ -1,13 +1,3 @@
...
-# The spec references the old name in %prep when building a pruned LAPACK archive.
-# Fixed upstream in atlas-3.10.3-33 (rawhide):
-#   https://src.fedoraproject.org/rpms/atlas/c/f253f3a2d6c1 (rhbz#2433875)
-# This overlay can be removed once we import from a Fedora version that includes the fix.
-[[components.atlas.overlays]]
-description = "Fix liblapack_pic.a reference — renamed to liblapack.a in lapack-static 3.12.0-10"
-type = "spec-search-replace"
-regex = 'liblapack_pic\.a'
-replacement = "liblapack.a"
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Ditto.

base/comps/cloud-init/cloud-init.comp.toml
```diff
@@ -1,158 +1,3 @@
...
-section = "%install"
-regex = '%py3_install -- --init-system=systemd'
-replacement = '%py3_install -- --init-system=systemd --distro=azurelinux'
-
-[[components.cloud-init.overlays]]
-description = "Set variant to azurelinux"
-type = "spec-search-replace"
-regex = 'python3 tools/render-template --variant %{\?rhel:rhel}%{!\?rhel:fedora}'
-replacement = "python3 tools/render-template --variant azurelinux"
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Please remove comment.

base/comps/cpio/cpio.comp.toml
```diff
@@ -1,14 +1,3 @@
...
-regex = ',\s*rmt\b'
-replacement = ''
-
-[[components.cpio.overlays]]
-description = "Drop the `rm -f $RPM_BUILD_ROOT%{_libexecdir}/rmt` line from %install. cpio's build tree no longer ships its own `rmt` helper under `_libexecdir`, so the deletion is a stale leftover. Mirrors the same Fedora rawhide commit (84d5a4fe) referenced by the BR-drop overlay above."
-type = "spec-search-replace"
-section = "%install"
-regex = "rm -f \\$RPM_BUILD_ROOT%\\{_libexecdir\\}/rmt"
-replacement = ""
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Please remove comment.

base/comps/glade/glade.comp.toml
```diff
@@ -1,9 +1,3 @@
 [components.glade]
-
-# Backport from Fedora rawhide: https://src.fedoraproject.org/rpms/glade/c/478c15c
-# Required by meson's gnome.generate_gir; missing from f43 spec.
-[[components.glade.overlays]]
-description = "Add missing BuildRequires for gobject-introspection-devel needed by meson gnome.generate_gir"
-type = "spec-add-tag"
-tag = "BuildRequires"
-value = "gobject-introspection-devel"
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Please remove comment.

base/comps/gnulib/gnulib.comp.toml
```diff
@@ -1,22 +1,6 @@
...
-type = "spec-update-tag"
-tag = "Source1"
-value = "https://erislabs.net/gitweb/?p=gnulib.git;a=blob_plain;hb=HEAD;f=debian/manpages/check-module.1#/check-module.1"
-
-[[components.gnulib.overlays]]
-description = "Temporary: append #/filename hint to Source2 so rendered spec filter can match gnulib-tool.1. Remove when snapshot includes f43 commit 41cae067."
-type = "spec-update-tag"
-tag = "Source2"
-value = "https://erislabs.net/gitweb/?p=gnulib.git;a=blob_plain;hb=HEAD;f=debian/manpages/gnulib-tool.1#/gnulib-tool.1"
+# All gnulib overlays live one-logical-change-per-file under ./overlays/, discovered
```
reubeno: Please remove comment.

base/comps/grub2/grub2.comp.toml
```diff
@@ -1,66 +1,3 @@
...
-# the three appended os-prober reverts. Matching the full filename is deliberate:
-# the regex and replacement must stay in lockstep, and if Fedora ever renames the
-# 0382 patch this overlay fails loudly at render time rather than silently
-# rewriting the line back to the stale name.
-regex = 'Patch0382: 0382-Set-correctly-the-memory-attributes-for-the-kernel-P\.patch'
-replacement = '''Patch0382: 0382-Set-correctly-the-memory-attributes-for-the-kernel-P.patch
-Patch0383: 0383-Revert-Revert-templates-Disable-the-os-prober-by-default.patch
-Patch0384: 0384-Revert-Revert-templates-Properly-disable-the-os-prober-by-d.patch
-Patch0385: 0385-Revert-Revert-templates-Fix-user-facing-typo-with-an.patch'''
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Please remove comment.

base/comps/gt/gt.comp.toml
```diff
@@ -1,17 +1,3 @@
...
-# Fixed via a patch that addresses the issue in F44 with
-# https://src.fedoraproject.org/rpms/gt/c/a8aec4724c07e306085f4b6e1a68e2ebde540359
-[[components.gt.overlays]]
-description = "Add -std=gnu89 to CFLAGS to fix K&R C function declarations failing with modern GCC"
-type = "spec-prepend-lines"
-section = "%build"
-lines = [
-    'export CFLAGS="$CFLAGS -std=gnu89"',
-]
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Please remove comment.

base/comps/gt/gt.comp.toml
```diff
@@ -1,17 +1,3 @@
 [components.gt]
```
reubeno, line 1: Side note: we should consider if we want to find a way to avoid the need for a 1-line file like this, but still include appropriate overlays. I'd imagine we could have the default overlay glob pattern include replacement (e.g., `{component}/overlays/*.overlay.toml`).

base/comps/intel-qpl/intel-qpl.comp.toml
```diff
@@ -1,22 +1,3 @@
...
-# "FATAL: ThreadSanitizer: unexpected memory mapping ..." aborts (e.g.
-# google/sanitizers#1716) when libqpl is loaded inside an unprivileged
-# container. Upstream made the same change in
-# https://src.fedoraproject.org/rpms/intel-qpl/c/7a433cf7338f7922f90b8819bcfd3b9fb9d4cb35
-[[components.intel-qpl.overlays]]
-description = "Disable ThreadSanitizer in release build (perf overhead + breaks under Docker seccomp)."
-type = "spec-search-replace"
-regex = '-DSANITIZE_THREADS=ON'
-replacement = "-DSANITIZE_THREADS=OFF"
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Please remove comment.

base/comps/javapackages-tools/javapackages-tools.comp.toml
```diff
@@ -1,79 +1,3 @@
...
-replacement = "Requires:       javapackages-local-openjdk25 = %{version}-%{release}"
-
-[[components.javapackages-tools.overlays]]
-description = "Fix javapackages-compat Requires to use javapackages-tools (upstream 83cd9c8)"
-type = "spec-search-replace"
-section = "%package"
-package = "javapackages-compat"
-regex = 'Requires:       javapackages-local = %\{version\}-%\{release\}'
-replacement = "Requires:       javapackages-tools = %{version}-%{release}"
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Please remove comment.

base/comps/kdump-utils/overlays/0001-only-first-match-find.overlay.toml
```diff
@@ -0,0 +1,11 @@
+# Upstream (currently unmerged) PR: https://github.com/rhkdump/kdump-utils/pull/153
+# The commit below is the PR's head commit (not yet merged upstream).
+[metadata]
+category = "backport-dist-git"
+upstreamable = true
+commits = ["https://github.com/rhkdump/kdump-utils/commit/c999bac950b817d229549d57f592980f4dc1655b"]
```
reubeno: It occurs to me that we have the same issue with this as we previously discussed with bugs -- with a list of commits, tools need to guess whether the commit is a Fedora commit or a commit further upstream.

base/comps/kdump-utils/kdump-utils.comp.toml
```diff
@@ -1,7 +1,3 @@
 [components.kdump-utils]
-
-# Upstream (currently unmerged) PR: https://github.com/rhkdump/kdump-utils/pull/153
-[[components.kdump-utils.overlays]]
-description = "Use only first match in _find_kernel_path_by_release()"
-type = "patch-add"
-source = "kdumpctl-Use-only-first-match-in-_find_kernel_path_b.patch"
+# Overlays live as one-logical-change-per-file under ./overlays/, discovered via
```
reubeno: Please remove comment. (I'll stop leaving these, but we should do the same for all other components.)

base/comps/sos/overlays/0001-fix-os-release-name.overlay.toml
```diff
@@ -0,0 +1,12 @@
+# Microsoft-authored sos Policy fix so os_release_name matches the os-release
+# NAME field on Azure Linux. Upstreamed as sosreport/sos commit 48c34be5.
+[metadata]
+category = "backport-dist-git"
```
reubeno: Is it appropriate to call this a backport-dist-git? We backported from the original upstream project, not dist-git.

base/comps/sos/overlays/0002-add-os-release-id.overlay.toml
```diff
@@ -0,0 +1,12 @@
+# Microsoft-authored sos Policy fix so os_release_id matches the os-release
+# ID field on Azure Linux. Upstreamed as sosreport/sos commit fcc6a5f4.
+[metadata]
+category = "backport-dist-git"
+upstreamable = true
+commits = ["https://github.com/sosreport/sos/commit/fcc6a5f498bc1e7edf7c2f2acf0f506ced0f21dc"]
```
reubeno: Same comment as above.

base/comps/sos/overlays/0003-add-missing-cloud-init.overlay.toml
```diff
@@ -0,0 +1,13 @@
+# Backport of upstream sos PR sosreport/sos#4318 (merged as 3aff70c2): add the
+# missing cloud-init systemd service units for the cloud-init >= 24.3
+# single-process optimization.
+[metadata]
+category = "backport-dist-git"
+upstreamable = true
+commits = ["https://github.com/sosreport/sos/commit/3aff70c2f55567c6aa911e4434e0ab59eb497044"]
```
reubeno: Ditto.

base/comps/sos/overlays/0004-add-dnf5-support-dnf.overlay.toml
```diff
@@ -0,0 +1,12 @@
+# Backport of upstream sos PR sosreport/sos#4311 (merged as b8ba38e8, closes
+# issue #4308): initial dnf5 support in the dnf plugin.
+[metadata]
+category = "backport-dist-git"
```
reubeno: Ditto.

base/comps/sssd/overlays/0002-fix-platform-m4-sourcing.overlay.toml
```diff
@@ -0,0 +1,11 @@
+# Backport of upstream SSSD PR sssd/sssd#8397 (merged as 308af8f2): fix
+# platform.m4 sourcing /etc/os-release overwriting the VERSION shell variable.
+[metadata]
+category = "backport-dist-git"
+upstreamable = false
```
reubeno: Why do we sometimes omit this and sometimes explicitly mark it false? Which is preferred?

base/comps/sssd/overlays/0002-fix-platform-m4-sourcing.overlay.toml
```diff
@@ -0,0 +1,11 @@
+# Backport of upstream SSSD PR sssd/sssd#8397 (merged as 308af8f2): fix
+# platform.m4 sourcing /etc/os-release overwriting the VERSION shell variable.
+[metadata]
+category = "backport-dist-git"
```
reubeno: Also not a backport from dist-git.

base/comps/vamp-plugin-sdk/overlays/0001-re-number-existing-patch.overlay.toml
```diff
@@ -0,0 +1,26 @@
...
+# rdfgen) depend on the archive files directly, not on sdkstatic. With
+# parallel make (-jN), the linker can read an archive while ranlib is
+# still rewriting its symbol table.
+#
+# Fix by moving ranlib into the individual archive target recipes so the
+# archive file is not considered up-to-date until ranlib has finished.
+# Mirrors the fix in Fedora rawhide vamp-plugin-sdk-2.10-4; the Makefile race
+# fix is upstream c4dm/vamp-plugin-sdk commit da2d66f.
+[metadata]
+category = "backport-dist-git"
```
reubeno: Ditto.

base/comps/xsimd/overlays/0001-apply-patch-fix-arm64.overlay.toml
```diff
@@ -0,0 +1,12 @@
+# Backport fix for ARM64 SVE static assertion failure from upstream commit 0fd437f.
+# Changes std::is_same to std::is_base_of to handle derived arch types like i8mm<neon64>.
+[metadata]
+category = "backport-dist-git"
```
reubeno: Not from dist-git

base/comps/components.toml
```diff
@@ -1,5 +1,15 @@
...
 
+# Project-wide default: every component automatically discovers per-file overlay
+# documents under its own `overlays/` directory. The glob is resolved relative
+# to each concrete component's config file, so components without an `overlays/`
+# directory simply match nothing (non-matching globs are ignored). This lets
+# components keep their overlays as one-logical-change-per-file `*.overlay.toml`
+# documents without repeating `overlay-files` in every comp.toml.
+# See https://github.com/microsoft/azure-linux-dev-tools/blob/main/docs/user/reference/config/overlays.md#per-file-overlay-format
+[default-component-config]
+overlay-files = ["overlays/*.overlay.toml"]
```
reubeno, line 11: The other PR has this in the distro config. Where do we want it?

reubeno (dismissed):

base/comps/ant/overlays/0001-remove-openjdk21-jdk-binding.overlay.toml
liunan-ms: I agree that the semantics of `upstreamable` on a backport are kind of awkward. Here is how we can read the current field:
`upstreamable` = false:
  - backport fixes that're already upstreamed
  - azl-only changes that should not be upstreamed

`upstreamable` = true:
- the change should be upstreamed

Expanding `upstreamable` into an enum will help differentiate the 2 cases while `upstreamable` = false, but in practice should we more care about the change should be upstreamed by looking for `upstreamable` = true because these are actional work? For backport fixes that are already upstreamed, we should remove the overlay once pinning to a new upstream version and azl-only changes will stay untouched.

base/comps/components.toml
liunan-ms: I think components.toml is the best place to define overlay-files in this PR. The overlay-files defined in distro config from the other PRs are temporarily for PR check, will be moved before PR merge.

base/comps/sos/overlays/0001-fix-os-release-name.overlay.toml
liunan-ms: This category name seems misleading. The intent of the category is "this is a backport of an existing fix, not a novel AZL derivation". Based on the feedback and our discussion, I think we decided to collapse both "backport from the original upstream project" and "backport from Fedora dist-git" into a single category, since in practice the distinction between "where we pulled the patch from" and "where the fix originated" is already captured precisely by the commits field (I'll switch the commits to the same structure as bugs, leaving the potential to track the database it comes from).

The `-dist-git` suffix is misleading. I could rename the category to just `backport` so the name matches the broader meaning. Happy to do that rename and commits field change (maybe switch upstreamable to enum) together as an azldev fix if you agree.

base/comps/sssd/overlays/0002-fix-platform-m4-sourcing.overlay.toml
liunan-ms: I explicitly mark the overlays that already have upstream fixes as `upstreamable = false`, omit it for the unsure ones, leaving them for future revisit. For example, fedora upstream has a patch to fix the same issue which is fixed by the overlays (not using the patch) in `gt`, so I'm not sure if this is upstreamable or even if this overlay should be a backport or not.

.azldev-version
```diff
@@ -1 +1 @@
-0256227f5434d9e00d7c8501b16848efa400a72b
+feeec457521dabc3948525c9c1df5e5c097f8348
```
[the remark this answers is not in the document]
liunan-ms: updated the PR description

base/comps/ant/overlays/0001-remove-openjdk21-jdk-binding.overlay.toml
liunan-ms: Addressed in commit 64bcb0fd — the `upstreamable` boolean has been replaced by an `upstream-status` enum as you suggested. Backported patches that already landed Fedora upstream now use `upstream-status = "upstreamed"`, cleanly distinguishing the following ones and AZL-specific changes (`inapplicable`):

- **upstreamable**: the patch we're carrying is itself upstream-shaped. Sending the exact same diff (or something very close) to Fedora / the upstream project would be accepted. This status is used for the overlays that is available in OSS project upstream (either released or in an open PR) but not in Fedora upstream yet and the local patches that we created and should be pushed to both OSS project and Fedora upstream. To differentiate the two cases in the existing overlays, the first one always has upstream commits in its metadata, the later one doesn't have any upstream commit yet. For new overlays, it's highly recommended to push the local fixes to OSS upstream first and grab the commit into the overlay metadata.
- **needs-upstream-hook**: the change is AZL-specific and would not be accepted upstream as-is (e.g. hard-coding `Vendor: Microsoft`, disabling a feature only we don't want, swapping a default path). But upstream could add a `bcond / %if / config` knob that would let us express the same customization without patching the spec. 

The schema has been regenerated to match.

base/comps/apache-ivy/overlays/0002-remove-buildrequires-ant-openjdk21.overlay.toml
liunan-ms: now uses `upstream-status = "upstreamable"`

base/comps/apache-ivy/apache-ivy.comp.toml
liunan-ms: The comment is removed in all comp.toml.

base/comps/gt/gt.comp.toml
liunan-ms: Good idea, and there's a WIP for it — PR [#269](https://github.com/microsoft/azure-linux-dev-tools/pull/269) adds support for pattern-discovered components. It's not wired into this PR yet though, because two issues still need to be worked out:

- `config dump` can't see pattern-discovered components. `config dump` serializes `env.Config()` before pattern discovery walks the filesystem and materializes the concrete `ComponentConfig` instances, so they don't show up in the dumped config.
- Component group members still need an explicitly declared `[components.<name>]` block to pass validation, so the 1-line file can't be fully eliminated yet.

I'll keep the minimal 1-line files as-is for now and follow up with a better design that addresses both once PR [#269](https://github.com/microsoft/azure-linux-dev-tools/pull/269) lands.

base/comps/kdump-utils/overlays/0001-only-first-match-find.overlay.toml
liunan-ms: Switched `commits` from bare URL strings to { url = "..." } URLRef objects as same as `bugs` field.

base/comps/sos/overlays/0001-fix-os-release-name.overlay.toml
liunan-ms: The category was renamed from `backport-dist-git` to `upstream-backport`, which is more accurate and covers backports from both the upstream project and dist-git. The commits list for this overlay now includes both the sosreport/sos upstream commit and the Fedora dist-git commit that carried it downstream.

base/comps/sssd/overlays/0002-fix-platform-m4-sourcing.overlay.toml
liunan-ms: Now `upstream-status` is a required field, which will push the overlay contributor to pick a definite status from `upstreamed`, `upstreamable`, `needs-upstream-hook`, `inapplicable`, or `unknown` for the change.

base/comps/sssd/overlays/0002-fix-platform-m4-sourcing.overlay.toml
liunan-ms: The category was renamed from `backport-dist-git` to `upstream-backport`, which is more accurate and covers backports from both the upstream project and dist-git.

base/comps/vamp-plugin-sdk/overlays/0001-re-number-existing-patch.overlay.toml
liunan-ms: same as above

base/comps/xsimd/overlays/0001-apply-patch-fix-arm64.overlay.toml
liunan-ms: same as above

base/comps/apache-ivy/overlays/0001-pack200-commons-compress.overlay.toml
```diff
@@ -0,0 +1,43 @@
+# Use Apache Commons Compress for pack200 support: stage two downstream patches
+# (Source4/Source5), wire them into %prep, and dos2unix the touched files.
+# Backport of upstream Apache Ivy commit d772841 (IVY-1652).
+[metadata]
+category = "upstream-backport"
+upstream-status = "upstreamable"
```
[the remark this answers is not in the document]
liunan-ms: `upstreamed` means the fix is already in Fedora upstream, `upstreamable` is the proper status as the fix is in Apache-ivy project but not in Fedora yet.

base/comps/xsimd/overlays/0001-apply-patch-fix-arm64.overlay.toml
```diff
@@ -0,0 +1,12 @@
+# Backport fix for ARM64 SVE static assertion failure from upstream commit 0fd437f.
+# Changes std::is_same to std::is_base_of to handle derived arch types like i8mm<neon64>.
+[metadata]
+category = "upstream-backport"
+upstream-status = "upstreamable"
```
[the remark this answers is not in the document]
liunan-ms: `upstreamed` means the fix is already in Fedora upstream, `upstreamable` is the proper status as the fix is in xsimd project but not in Fedora yet.

.azldev-version
```diff
@@ -1 +1 @@
-0256227f5434d9e00d7c8501b16848efa400a72b
+830e6fed79750f8e24ae17cd630248f50c90c8a9
```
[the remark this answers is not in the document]
liunan-ms: updated the PR description

tobiasb-ms (approved):
I confirmed via before/after runs of `azldev component list --all-components` that the only changes are the addition of `metadata` and a few new `description` fields.

Seems good.

tobiasb-ms (approved):

PR #17744, opened 2026-06-17
Addressing ptest failures for multiple packages

jykanase:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Fix failing core ptest across packages

## Summary

Bumps `Release` and adds a changelog entry on every affected spec. Per-spec rationale:

- **debugedit** — replace the broken `%make_build_check` macro with an explicit `make %{?_smp_mflags} check` in `%check` so the test target actually runs.
- **dtc** — add `fix-check-err-tuple.patch` to handle SWIG returning a tuple instead of an integer from `fdt_get_mem_rsv` in `pylibfdt` (`make check` itself is unchanged).
- **fakeroot** — split the `obj-sysv` and `obj-tcp` test runs and retry the `tcp` variant once to ride out intermittent `t.tar` failures in the chroot; also move `COPYING` from `%doc` to `%license` so the toolkit license-checker no longer flags it.
- **fillup** — drop `%{?_smp_mflags}` from `make test` in `%check`; the test target is not parallel-safe.
- **future** — add `future-fix-imp-removal.patch` (fall back to `importlib` when the `imp` module is unavailable, Python 3.12+); expand the pytest `-k` exclusion list and add 10 `--ignore` entries for test modules that still depend on `imp` or stdlib modules removed in 3.12.
- **keyutils** — tolerate `make -k test`; the `keyctl/requesting/valid` case invokes the kernel `request_key` callout which spawns `/sbin/request-key` in the host pid/mount namespace and cannot see the chroot-installed binary. The other ~140 tests pass.
- **libtool** — run `make check` with `TESTSUITEFLAGS='1-65 67-168'` to skip exactly test 66 (`link-order`, broken with `binutils >= 2.39` — debbugs #58105) and test 169 (`cmdline_wrap`, which cascades from any earlier failure including 66). Any other regression still hard-fails the build.
- **m2crypto** — replace deprecated `setup.py test` with `pytest`, copy system `six` over the bundled `M2Crypto/six.py` so the test run surfaces the real underlying error in the log, and tolerate (`|| :`) because M2Crypto 0.38.0 has two upstream incompatibilities with Python 3.12 that cannot be fixed at the spec level: (1) the bundled `six.moves` lazy importer is broken by 3.12 importlib changes, and (2) `SSL/ssl_dispatcher.py` imports `asyncore`, which was removed by PEP 594. These are fixed in M2Crypto >= 0.40; the day this package is rebased, `%check` becomes meaningful with no further work. Add `python3-pytest` and `python3-six` BRs.
- **man-db** — `make check XFAIL_TESTS="man1/lexgrog.1"`; the `lexgrog.1` page trips groff's line-adjust check due to the missing `CW` (constant-width) font in the build chroot. The other 11 man-page tests still hard-fail on regression.
- **nbdkit** — add the missing `tests/test-blocksize-extents-overflow.sh` (completes the CVE-2025-47712 backport — without it `make check` aborted with *No rule to make target*); skip 10 chroot-incompatible tests (5 TLS — Azure Linux gnutls 3.8.x rejects the test PKI; 3 OCaml — 1.35.3's OCaml plugin is incompatible with ocaml 5.x multicore; 2 file-plugin dir-mode — upstream `free(): invalid pointer` under `MALLOC_CHECK_=1`) and XFAIL `test-ocaml-errorcodes`.
- **perl-List-MoreUtils** — drop the `cpan -i List::MoreUtils::XS` bootstrap from `%check` (XS package is not available in the build env) and run only the pure-perl tests via `make test TEST_FILES="t/pureperl/*.t"`.
- **python-astunparse** — replace deprecated `setup.py test` with `pytest` and deselect `tests/test_unparse.py::UnparseTestCase::test_files`; astunparse 1.6.3 has no `_Match` handler for the Python 3.10+ match-statement AST. All other tests still hard-fail. Add `python3-pytest` BR.
- **python-cherrypy** — enumerate 11 specific `--deselect`s in the `tox` invocation (multipart-form half-close timing, timing-sensitive bus/conn/gc tests, range-header parsing, SIGHUP-daemonized reload). All other test failures still hard-fail.
- **python-coverage** — the previous `pip3 install -r requirements/dev.pip; tox` chain silently ran zero tests (the rpm-installed pip has no `RECORD` file, so the install aborted before tox was present). Replace it with a direct pytest invocation: build the encoded-source zip via `igor.py zip_mods`, copy the built `CTracer` `.so` into the source tree so `COVERAGE_CORE=ctrace` resolves correctly, ignore 4 test modules (`test_concurrency.py` and `test_oddball.py` import `flaky` which lives in SPECS-EXTENDED and can't be a core BR; `test_process.py` and `test_venv.py` need subprocesses / virtualenv) and deselect 10 environment-dependent tests. Add `python3-execnet`, `python3-hypothesis`, `python3-pytest-xdist`, `python3-sortedcontainers` BRs.
- **python-distlib** — extend the existing pytest `-k` exclusion to also skip `test_package_data`; depends on package-data layout that is not reliable in the build env.
- **python-fields** — pin `pytest>=5.4.0,<9` in the `%check` bootstrap; pytest 9.x rejects duplicate parametrize IDs (`test_tuple_pickle` uses `loads` twice).
- **python-gevent** — replace `%tox || :` with proper scoped exclusions via an updated `skip-irrelevant-tests.patch`: (1) add 2 atexit-sensitive stdlib tests to gevent's native `disabled_tests` list, (2) decorate one timing-flaky core test with `@unittest.skip`, (3) neutralize `TestEtcHosts` (synthetic resolver comparison that recurses in restricted DNS environments), (4) bump per-file test timeout 100s → 300s, (5) skip the monkey-patched stdlib `test_socket.py` entirely (upstream itself flags it as flaky under monkey-patching). `%tox` now runs clean — no `|| :`.
- **python-google-pasta** — replace deprecated `setup.py test` with `pytest`; exclude `PrefixSuffixGoldenTest` (51 failures from Python 3.12 AST changes) and the `fstring` / `inline_*` tests that error from the same AST changes. Add `python3-pip` BR.
- **python-logutils** — patch tests for Python 3.12 compatibility in `%prep` (replace `assertEquals` with `assertEqual` in `test_dictconfig.py`; add explicit `l.setLevel(logging.WARNING)` in `test_adapter.py`, `test_testing.py`, `test_queue.py`).
- **python-markdown** — pass lowercase `%{pkgname}` to `%autosetup -n` so the source dir matches `markdown-3.8.2`; rename two test methods (`test_raw_missing_close_bracket`, `test_unclosed_comment_*`) to skip them — both fail due to Python 3.12 `html.parser` behaviour changes.
- **python-pbr** — `tox 3.x` + `virtualenv 21.x` can no longer editable-install pbr from source under modern setuptools. Replace tox with a direct `stestr` invocation and exclude 5 tests (`test_freeze_command`, `test_console_script_{develop,install}`, `test_pep_517_support`, `test_requirement_parsing`) that recursively pip-install pbr-from-source inside a PEP 517 isolated env where pbr's `setup.py` cannot find itself. Fixed upstream in pbr >= 6.1. Add `python3-virtualenv` and `python3-wheel` BRs.
- **python-pexpect** — patch tests to use `assertRaisesRegex` (`assertRaisesRegexp` was removed in Python 3.12); `--ignore` `test_pxssh.py` (needs an SSH server) and `test_async.py` (uses `asyncio.coroutine`, removed in 3.12).
- **python-pycodestyle** — pin `packaging==23.2` in the test bootstrap to avoid the uninstall conflict with the system RPM-managed `packaging`.
- **python-pytest-benchmark** — `pip3 install -e .` so the pytest plugin entry point is discovered; suppress `ast.Str` `DeprecationWarning`; extend the `-k` exclusion list with 10 more tests broken by Python 3.12 `CodeType` / pytest output-format differences.
- **python-remoto** — rename `def setup(self)` → `def setup_method(self)` in `test_process.py` / `test_util.py`; newer pytest no longer auto-invokes `setup` on plain (non-`unittest`) classes.
- **python-responses** — pin `packaging==23.2` in `%check` to avoid the uninstall conflict with the system RPM-managed `packaging`.
- **python-routes** — pin `packaging==23.2` and drop the `tox==3.4.0` pin (tox 4.x is required for the `py312` environment).
- **python-setuptools_scm** — pin `packaging==23.2` in `%check` to avoid the uninstall conflict with the system RPM-managed `packaging`.
- **python-sphinx** — add `python3-filelock` BR (two test modules failed pytest collection without it); drop the unreliable `pip install --upgrade` bootstrap (rpm-installed pip has no `RECORD` file); extend `--ignore` with three more modules that depend on `html5lib` via docutils (`test_build_texinfo`, `test_domain_std`, `test_smartquotes` — `html5lib` cannot be a core BR because its transitive `webencodings` dep lives in SPECS-EXTENDED); deselect 7 version-specific tests (`test_enum_class`, `test_autodoc_type_aliases`, `test_autodoc_default_options`, `test_viewcode`, `test_viewcode_linenos`, two `test_intl` translation cases) that are fixed in later Sphinx releases. Any other failure still hard-fails.
- **python-sqlalchemy** — sed-patch `lib/sqlalchemy/util/_concurrency_py3k.py` to create a new event loop when none is running (Python 3.12 deprecated `get_event_loop()`); pass `--ignore=test/ext/mypy --ignore=test/typing/test_mypy.py` (version-specific mypy output mismatches); drop the redundant `doc/copyright.html` HTML rendering of `LICENSE` and add `%license LICENSE` to the `-doc` subpackage so the toolkit license-checker no longer flags it.
- **python-toolz** — replace `nosetests` with `pytest` (`nose` is incompatible with Python 3.12 — no `imp` module); pin `packaging==23.2`; exclude `test_inspect_wrapped_property` (Python 3.12 `inspect` behaviour change).
- **python-tox-current-env** — install `packaging==23.2` (avoid uninstall conflict) and tolerate `tox -e py%{python3_version_nodots}` because 177 of 185 integration tests invoke `tox` internally, which itself fails inside the build chroot.
- **python-tqdm** — `--ignore=tests/tests_pandas.py`; the file uses `pandas.core.common.is_builtin_func` which was removed in newer pandas releases.
- **python-twisted** — the full upstream `tox -e nocov-posix-alldeps` suite (>10000 tests) repeatedly hangs in CI past 5h on conch/names/dns/process/multicast modules with no captured culprit. Replace it with a curated `trial` smoke-test subset on 7 pure-logic modules (`test_compat`, `test_error`, `test_cooperator`, `test_paths`, `test_text`, `test_strerror`, `test_randbytes` — ~200 unit tests, no network, no spawned children, no privileged sockets, no extra PyPI deps). Drop the BRs and setup steps (`pip`, `sudo`, `net-tools`, `tzdata`, `git`, multicast route, test user) that only existed to bootstrap tox; add `python3-attrs`, `python3-constantly`, `python3-hyperlink`, `python3-typing-extensions` BRs that `trial` needs to import `twisted`. Once we upgrade the package, we can revert this.
- **python-urllib3** — pin `pytest>=7.4.0,<9` in the `%check` bootstrap; pytest 9.x rejects non-Collection iterables (`itertools.chain`) in `parametrize` `argvalues` used by the urllib3 test suite.

## Testing

Each spec was rebuilt locally with:

```
sudo make build-packages REBUILD_TOOLS=y \
     PACKAGE_BUILD_LIST=<pkg> PACKAGE_REBUILD_LIST=<pkg> \
     SRPM_PACK_LIST=<pkg> ALLOW_TOOLCHAIN_REBUILDS=y RUN_CHECK=y
```

and the `ptest` log was inspected to confirm only the intended skips / XFAILs trigger and that no real regression is masked.

**YES**


Kanishk-Bansal (approved):
LGTM

Kanishk-Bansal:
[Build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1141192&view=results)

kgodara912:
[Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1168540&view=results)

SPECS/m2crypto/m2crypto.spec
```diff
@@ -48,16 +50,50 @@ messenger for Zope.
...
+# Upstream M2Crypto >= 0.40 drops the asyncore-based dispatcher. Until
+# this package can be rebased on 0.40+, the tests cannot pass under
+# Python 3.12 and we tolerate failure with `|| :`. We still run pytest
+# (with the six.py shim so reviewers see the real underlying error) and
+# still install `parameterized`, so the day this package is updated the
+# %check section starts being meaningful with no further work.
 pip3 install parameterized
-#Testing: MiscSSLClientTestCase failing with SSLError not raised
-%python3 setup.py test
+cp -f $(%python3 -c "import six; print(six.__file__)") \
```
mfrw, line 73: Just to double verify - this is only happening in the check stage and does not ship anything :)

mfrw (approved):
LGTM
Thanks for this

kgodara912 (approved):
Multiple python packages and python were initially borrowed from 2.0 branch. As the python version kept upgrading for 3.0, few of tests went out of sync for those packages. Similarly, few other packages also became out of sync with base 3.0 packages which caused few failures for other packages. This PR attempts to either exclude those tests keeping future scope of improvements for those packages by upgrading. This PR will also ensure that we do not ignore positive test failure results for other tests. Buddy build is successful and all the package tests are passing or few excluded to allow other tests to pass and report the failures properly. We will raise a separate PR which can handle the upgrades to the packages in the long term. LGTM.

PR #17749, opened 2026-06-17
Addressing ptest failures for multiple packages

kgodara912:
###### Merge Checklist  

- [x] Ready to merge

---

###### Summary 

Fix failing core ptest across packages

## Summary

Bumps `Release` and adds a changelog entry on every affected spec. Per-spec rationale:

- **atftp** — drop `%check` entirely (set to `:`). Upstream `test/test.sh` is an interactive end-to-end network test (sudo+tmpfs, multiple SSH hosts for multicast) and is not runnable in a non-interactive build chroot. Upstream's own `redhat/atftp.spec.in` ships no `%check` for the same reason.
- **debugedit** — replace the broken `%make_build_check` macro with an explicit `make %{?_smp_mflags} check` in `%check` so the test target actually runs.
- **dtc** — add `fix-check-err-tuple.patch` to handle SWIG returning a tuple instead of an integer from `fdt_get_mem_rsv` in `pylibfdt` (`make check` itself is unchanged).
- **fakeroot** — split the `obj-sysv` and `obj-tcp` test runs and retry the `tcp` variant once to ride out intermittent `t.tar` failures in the chroot; also move `COPYING` from `%doc` to `%license` so the toolkit license-checker no longer flags it.
- **fillup** — drop `%{?_smp_mflags}` from `make test` in `%check`; the test target is not parallel-safe.
- **future** — add `future-fix-imp-removal.patch` (fall back to `importlib` when the `imp` module is unavailable, Python 3.12+); expand the pytest `-k` exclusion list and add 10 `--ignore` entries for test modules that still depend on `imp` or stdlib modules removed in 3.12.
- **keyutils** — tolerate `make -k test`; the `keyctl/requesting/valid` case invokes the kernel `request_key` callout which spawns `/sbin/request-key` in the host pid/mount namespace and cannot see the chroot-installed binary. The other ~140 tests pass.
- **libtool** — run `make check` with `TESTSUITEFLAGS='1-65 67-168'` to skip exactly test 66 (`link-order`, broken with `binutils >= 2.39` — debbugs #58105) and test 169 (`cmdline_wrap`, which cascades from any earlier failure including 66). Any other regression still hard-fails the build.
- **m2crypto** — replace deprecated `setup.py test` with `pytest`, copy system `six` over the bundled `M2Crypto/six.py` so the test run surfaces the real underlying error in the log, and tolerate (`|| :`) because M2Crypto 0.38.0 has two upstream incompatibilities with Python 3.12 that cannot be fixed at the spec level: (1) the bundled `six.moves` lazy importer is broken by 3.12 importlib changes, and (2) `SSL/ssl_dispatcher.py` imports `asyncore`, which was removed by PEP 594. These are fixed in M2Crypto >= 0.40; the day this package is rebased, `%check` becomes meaningful with no further work. Add `python3-pytest` and `python3-six` BRs.
- **man-db** — `make check XFAIL_TESTS="man1/lexgrog.1"`; the `lexgrog.1` page trips groff's line-adjust check due to the missing `CW` (constant-width) font in the build chroot. The other 11 man-page tests still hard-fail on regression.
- **nbdkit** — add the missing `tests/test-blocksize-extents-overflow.sh` (completes the CVE-2025-47712 backport — without it `make check` aborted with *No rule to make target*); skip 10 chroot-incompatible tests (5 TLS — Azure Linux gnutls 3.8.x rejects the test PKI; 3 OCaml — 1.35.3's OCaml plugin is incompatible with ocaml 5.x multicore; 2 file-plugin dir-mode — upstream `free(): invalid pointer` under `MALLOC_CHECK_=1`) and XFAIL `test-ocaml-errorcodes`.
- **perl-List-MoreUtils** — drop the `cpan -i List::MoreUtils::XS` bootstrap from `%check` (XS package is not available in the build env) and run only the pure-perl tests via `make test TEST_FILES="t/pureperl/*.t"`.
- **python-astunparse** — replace deprecated `setup.py test` with `pytest` and deselect `tests/test_unparse.py::UnparseTestCase::test_files`; astunparse 1.6.3 has no `_Match` handler for the Python 3.10+ match-statement AST. All other tests still hard-fail. Add `python3-pytest` BR.
- **python-cherrypy** — enumerate 11 specific `--deselect`s in the `tox` invocation (multipart-form half-close timing, timing-sensitive bus/conn/gc tests, range-header parsing, SIGHUP-daemonized reload). All other test failures still hard-fail.
- **python-coverage** — the previous `pip3 install -r requirements/dev.pip; tox` chain silently ran zero tests (the rpm-installed pip has no `RECORD` file, so the install aborted before tox was present). Replace it with a direct pytest invocation: build the encoded-source zip via `igor.py zip_mods`, copy the built `CTracer` `.so` into the source tree so `COVERAGE_CORE=ctrace` resolves correctly, ignore 4 test modules (`test_concurrency.py` and `test_oddball.py` import `flaky` which lives in SPECS-EXTENDED and can't be a core BR; `test_process.py` and `test_venv.py` need subprocesses / virtualenv) and deselect 10 environment-dependent tests. Add `python3-execnet`, `python3-hypothesis`, `python3-pytest-xdist`, `python3-sortedcontainers` BRs.
- **python-distlib** — extend the existing pytest `-k` exclusion to also skip `test_package_data`; depends on package-data layout that is not reliable in the build env.
- **python-fields** — pin `pytest>=5.4.0,<9` in the `%check` bootstrap; pytest 9.x rejects duplicate parametrize IDs (`test_tuple_pickle` uses `loads` twice).
- **python-gevent** — replace `%tox || :` with proper scoped exclusions via an updated `skip-irrelevant-tests.patch`: (1) add 2 atexit-sensitive stdlib tests to gevent's native `disabled_tests` list, (2) decorate one timing-flaky core test with `@unittest.skip`, (3) neutralize `TestEtcHosts` (synthetic resolver comparison that recurses in restricted DNS environments), (4) bump per-file test timeout 100s → 300s, (5) skip the monkey-patched stdlib `test_socket.py` entirely (upstream itself flags it as flaky under monkey-patching). `%tox` now runs clean — no `|| :`.
- **python-google-pasta** — replace deprecated `setup.py test` with `pytest`; exclude `PrefixSuffixGoldenTest` (51 failures from Python 3.12 AST changes) and the `fstring` / `inline_*` tests that error from the same AST changes. Add `python3-pip` BR.
- **python-logutils** — patch tests for Python 3.12 compatibility in `%prep` (replace `assertEquals` with `assertEqual` in `test_dictconfig.py`; add explicit `l.setLevel(logging.WARNING)` in `test_adapter.py`, `test_testing.py`, `test_queue.py`).
- **python-markdown** — pass lowercase `%{pkgname}` to `%autosetup -n` so the source dir matches `markdown-3.8.2`; rename two test methods (`test_raw_missing_close_bracket`, `test_unclosed_comment_*`) to skip them — both fail due to Python 3.12 `html.parser` behaviour changes.
- **python-pbr** — `tox 3.x` + `virtualenv 21.x` can no longer editable-install pbr from source under modern setuptools. Replace tox with a direct `stestr` invocation and exclude 5 tests (`test_freeze_command`, `test_console_script_{develop,install}`, `test_pep_517_support`, `test_requirement_parsing`) that recursively pip-install pbr-from-source inside a PEP 517 isolated env where pbr's `setup.py` cannot find itself. Fixed upstream in pbr >= 6.1. Add `python3-virtualenv` and `python3-wheel` BRs.
- **python-pexpect** — patch tests to use `assertRaisesRegex` (`assertRaisesRegexp` was removed in Python 3.12); `--ignore` `test_pxssh.py` (needs an SSH server) and `test_async.py` (uses `asyncio.coroutine`, removed in 3.12).
- **python-pycodestyle** — pin `packaging==23.2` in the test bootstrap to avoid the uninstall conflict with the system RPM-managed `packaging`.
- **python-pytest-benchmark** — `pip3 install -e .` so the pytest plugin entry point is discovered; suppress `ast.Str` `DeprecationWarning`; extend the `-k` exclusion list with 10 more tests broken by Python 3.12 `CodeType` / pytest output-format differences.
- **python-remoto** — rename `def setup(self)` → `def setup_method(self)` in `test_process.py` / `test_util.py`; newer pytest no longer auto-invokes `setup` on plain (non-`unittest`) classes.
- **python-responses** — pin `packaging==23.2` in `%check` to avoid the uninstall conflict with the system RPM-managed `packaging`.
- **python-routes** — pin `packaging==23.2` and drop the `tox==3.4.0` pin (tox 4.x is required for the `py312` environment).
- **python-setuptools_scm** — pin `packaging==23.2` in `%check` to avoid the uninstall conflict with the system RPM-managed `packaging`.
- **python-sphinx** — add `python3-filelock` BR (two test modules failed pytest collection without it); drop the unreliable `pip install --upgrade` bootstrap (rpm-installed pip has no `RECORD` file); extend `--ignore` with three more modules that depend on `html5lib` via docutils (`test_build_texinfo`, `test_domain_std`, `test_smartquotes` — `html5lib` cannot be a core BR because its transitive `webencodings` dep lives in SPECS-EXTENDED); deselect 7 version-specific tests (`test_enum_class`, `test_autodoc_type_aliases`, `test_autodoc_default_options`, `test_viewcode`, `test_viewcode_linenos`, two `test_intl` translation cases) that are fixed in later Sphinx releases. Any other failure still hard-fails.
- **python-sqlalchemy** — sed-patch `lib/sqlalchemy/util/_concurrency_py3k.py` to create a new event loop when none is running (Python 3.12 deprecated `get_event_loop()`); pass `--ignore=test/ext/mypy --ignore=test/typing/test_mypy.py` (version-specific mypy output mismatches); drop the redundant `doc/copyright.html` HTML rendering of `LICENSE` and add `%license LICENSE` to the `-doc` subpackage so the toolkit license-checker no longer flags it.
- **python-toolz** — replace `nosetests` with `pytest` (`nose` is incompatible with Python 3.12 — no `imp` module); pin `packaging==23.2`; exclude `test_inspect_wrapped_property` (Python 3.12 `inspect` behaviour change).
- **python-tox-current-env** — install `packaging==23.2` (avoid uninstall conflict) and tolerate `tox -e py%{python3_version_nodots}` because 177 of 185 integration tests invoke `tox` internally, which itself fails inside the build chroot.
- **python-tqdm** — `--ignore=tests/tests_pandas.py`; the file uses `pandas.core.common.is_builtin_func` which was removed in newer pandas releases.
- **python-twisted** — the full upstream `tox -e nocov-posix-alldeps` suite (>10000 tests) repeatedly hangs in CI past 5h on conch/names/dns/process/multicast modules with no captured culprit. Replace it with a curated `trial` smoke-test subset on 7 pure-logic modules (`test_compat`, `test_error`, `test_cooperator`, `test_paths`, `test_text`, `test_strerror`, `test_randbytes` — ~200 unit tests, no network, no spawned children, no privileged sockets, no extra PyPI deps). Drop the BRs and setup steps (`pip`, `sudo`, `net-tools`, `tzdata`, `git`, multicast route, test user) that only existed to bootstrap tox; add `python3-attrs`, `python3-constantly`, `python3-hyperlink`, `python3-typing-extensions` BRs that `trial` needs to import `twisted`. Once we upgrade the package, we can revert this.
- **python-urllib3** — pin `pytest>=7.4.0,<9` in the `%check` bootstrap; pytest 9.x rejects non-Collection iterables (`itertools.chain`) in `parametrize` `argvalues` used by the urllib3 test suite.

## Testing

Each spec was rebuilt locally with:

```
sudo make build-packages REBUILD_TOOLS=y \
     PACKAGE_BUILD_LIST=<pkg> PACKAGE_REBUILD_LIST=<pkg> \
     SRPM_PACK_LIST=<pkg> ALLOW_TOOLCHAIN_REBUILDS=y RUN_CHECK=y
```

and the `ptest` log was inspected to confirm only the intended skips / XFAILs trigger and that no real regression is masked.

**YES**


kgodara912:
[Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1141180&view=results)

kgodara912:
[Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1142236&view=results)
For twisted: [Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1142280&view=results)

SPECS/libtool/libtool.spec
```diff
@@ -81,6 +96,12 @@ make %{?_smp_mflags} check
 %{_libdir}/libltdl.so.7.3.2
 
 %changelog
+* Wed Jun 17 2026 Kshitiz Godara <kgodara@microsoft.com> - 2.4.7-2
+- Skip exactly tests 66 (link-order, broken with binutils >= 2.39) and
+  169 (cmdline_wrap, cascades from 66) via TESTSUITEFLAGS instead of
+  swallowing all `make check` failures. Any other regression now
+  fails the build.
```
mfrw, line 103: This is a really good approach IMHO.

mfrw (approved):
Hi sir, there are a few suggestion of duplicate use of python dependcies - rest the PR looks good.
Thank you for this - this is a massive effort :)
[sorry took a little more time to review - given the size of the change]

Kanishk-Bansal (commented):
## Per-package review of `%check` changes


**Core / non-python**
- **atftp** — LGTM: dropping `%check` to `:` matches upstream `redhat/atftp.spec.in`; the interactive network test genuinely can't run in a chroot.
- **debugedit** — LGTM: swapping the broken `%make_build_check` macro for an explicit `make %{?_smp_mflags} check` is the correct fix.
- **dtc** — LGTM: `fix-check-err-tuple.patch` correctly handles the SWIG tuple return and is defensive (`isinstance` guard keeps the int path working).
- **fakeroot** — LGTM: splitting `obj-sysv`/`obj-tcp` keeps sysv strict while only retrying the flaky tcp variant; `%doc`→`%license` for COPYING is correct. (Source0 moved to snapshot.debian.org + matching cgmanifest update — good.)
- **fillup** — LGTM: dropping `%{?_smp_mflags}` is the right call for a non-parallel-safe test target.
- **keyutils** — ⚠️ `make -k test || :` swallows *all* failures, not just `keyctl/requesting/valid`. Consider XFAIL/grep-on-summary so a real regression still fails the build.
- **libtool** — LGTM (nice): `TESTSUITEFLAGS='1-65 67-168'` skips exactly 66 + 169 and lets any other regression hard-fail — much better than swallowing.
- **man-db** — LGTM: `XFAIL_TESTS="man1/lexgrog.1"` is surgical and keeps the other 11 tests strict.
- **nbdkit** — LGTM: adding the missing `test-blocksize-extents-overflow.sh` completes the CVE backport; skips are itemized and `test-ocaml-errorcodes` correctly XFAIL'd (binary, not skip_test).
- **perl-List-MoreUtils** — LGTM with note: limiting to `t/pureperl/*.t` is reasonable since the XS pkg isn't in the build env, but it does reduce coverage to pure-perl only.

**Python — surgical deselects/ignores (good)**
- **python-astunparse** — LGTM: `pytest` + single `--deselect` of the `_Match`/3.10 case; rest stays strict.
- **python-cherrypy** — LGTM: 11 enumerated `--deselect`s, all others hard-fail.
- **python-coverage** — LGTM (thorough): replaces the silently-no-op tox chain with a real pytest run; the `flaky`-in-SPECS-EXTENDED rationale and CTracer `.so` copy are well explained.
- **python-distlib** — LGTM: one extra `-k` exclusion (`test_package_data`).
- **python-fields** — LGTM: `pytest<9` pin for the duplicate-parametrize-ID issue.
- **python-filelock** — LGTM: test-only `caplog.clear()` patch (gated by `with_check`) fixes the aarch64 async-leak flake without touching library code.
- **python-gevent** — LGTM (nice): replaces `%tox || :` with properly scoped patch exclusions — strictly better than the old blanket tolerate.
- **python-google-pasta** — LGTM: `pytest` + `-k` exclusion of the 3.12 AST-change cases.
- **python-logutils** — LGTM: `%prep` seds for `assertEquals`→`assertEqual` + explicit `setLevel` are minimal 3.12-compat fixes.
- **python-markdown** — LGTM: lowercase `%{pkgname}` for `%autosetup -n` is a real correctness fix; the 2 sed-renamed tests are 3.12 `html.parser` cases.
- **python-pbr** — LGTM: direct `stestr` with 5 documented excludes is a sound replacement for the broken tox+virtualenv editable install.
- **python-pexpect** — LGTM: `assertRaisesRegex` sed + `pytest-timeout` + single retry; ignores need SSH / removed `asyncio.coroutine`.
- **python-pytest-benchmark** — LGTM with note: `pip install -e .` to register the plugin entry point is correct; the `-k` exclusion list grew large (~15 tests) — worth tracking for a future rebase.
- **python-pytest-cov** — LGTM: `packaging==23.2` pin + the dist/html skip patch are well justified (coverage 7.x / xdist 3.x interaction).
- **python-pytest-xdist** — LGTM (nice): backports the upstream loadscope sort-by-gateway-id fix — a real de-flake, not a skip.
- **python-remoto** — LGTM: `setup`→`setup_method` rename for newer pytest.
- **python-sphinx** — LGTM: `python3-filelock` BR + dropping the no-op pip bootstrap + itemized ignores/deselects; html5lib/SPECS-EXTENDED rationale is sound.
- **python-sqlalchemy** — LGTM with note: sed-patching `get_event_loop()` works but is brittle to upstream formatting; mypy ignores + `%license` doc fix are fine.
- **python-toolz** — LGTM: `nose`→`pytest` (nose is dead on 3.12), one inspect-change exclusion.
- **python-tqdm** — LGTM: `timeout 600` wrapper + `pytest-timeout` + retry addresses the `test_imap` pool-leak hang; pandas ignore is justified.

**Python — packaging==23.2 pin (uninstall-no-RECORD conflict)**
- **python-pycodestyle / python-responses / python-routes / python-setuptools_scm** — LGTM: identical, well-understood `packaging==23.2` pin; `routes` also correctly drops the stale `tox==3.4.0` pin so py312 works.

**Python — blanket tolerate (`|| :`) — please double-check**
- **m2crypto** — Acceptable interim: failure is genuinely unfixable at spec level (asyncore/PEP 594 + vendored six on 3.12) and is clearly documented; the six-shim to surface the real error is a nice touch. Recommend a tracking issue to rebase to ≥0.40 and re-arm `%check`.
- **python-tox-current-env** — ⚠️ `tox ... || :` makes `%check` effectively non-gating (177/185 invoke tox internally). Documented, but it now catches zero regressions — a tracking issue would help.

**Python — large coverage reduction (justified but flagging)**
- **python-twisted** — ⚠️ biggest change: swaps the full >10k-test suite for a ~200-test `trial` smoke subset to escape the >5h CI hang. Reasonable as a temporary unblock and the BR cleanup is correct, but coverage drops a lot — please keep the "revert on upgrade" note tracked.
- **future** — LGTM with note: the `imp`→`importlib` fallback patch is good; the 10 `--ignore` + expanded `-k` list is large but each target is a removed-stdlib/3.12 module.
- **python-urllib3** — LGTM: `pytest<9` pin (also de-dupes the double `pytest>=7.4.0` line).

**Manifests** — `debugedit 5.0-3` and `libtool 2.4.7-2` (incl. `libltdl`) bumped consistently across `pkggen_core_*` and `toolchain_*` for both arches; cgmanifest fakeroot URL updated. ✅

SPECS/atftp/atftp.spec
```diff
@@ -85,8 +86,14 @@ ATFTPD_BIND_ADDRESSES=
...
-sed -i 's/^start_server$/chown -R nobody $DIRECTORY\nstart_server/g' test/test.sh || true
-make %{?_smp_mflags} check
+# Upstream's test/test.sh is an end-to-end network test that expects an
+# interactive shell, a writable tftpboot dir, the ability to mount a tmpfs
+# ramdisk via sudo, and (optionally) several remote hosts for the multicast
+# and mtftp scenarios. It is not designed to run in a non-interactive build
+# chroot, and the upstream-shipped redhat/atftp.spec.in deliberately ships
+# no %%check stage at all. We follow that precedent here. The build itself
+# still exercises the toolchain via `make` in %%build.
+:
```
Kanishk-Bansal, line 95: LGTM: dropping %check to : matches upstream redhat/atftp.spec.in; the interactive network test genuinely can't run in a chroot

SPECS/debugedit/debugedit.spec
```diff
@@ -22,7 +22,7 @@ Source0:      https://sourceware.org/ftp/%{name}/%{version}/%{name}-%{version}.t
 %make_install
 
 %check
-%make_build_check
+make %{?_smp_mflags} check
```
Kanishk-Bansal, line 25: LGTM: swapping the broken %make_build_check macro for an explicit make %{?_smp_mflags} check is the correct fix

SPECS/dtc/dtc.spec
```diff
@@ -1,13 +1,14 @@
...
 Version:        1.7.0
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        BSD OR GPLv2+
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 Group:          Development/Tools
 URL:            https://devicetree.org/
 Source0:        https://kernel.org/pub/software/utils/%{name}/%{name}-%{version}.tar.gz
+Patch0:         fix-check-err-tuple.patch
```
Kanishk-Bansal, line 11: LGTM: fix-check-err-tuple.patch correctly handles the SWIG tuple return and is defensive (isinstance guard keeps the int path working).

SPECS/fakeroot/fakeroot.spec
```diff
@@ -119,9 +119,10 @@ done
 %endif
 
 %check
-for type in sysv tcp; do
-  make -C obj-$type check VERBOSE=1
-done
+# sysv variant must pass all tests
+make -C obj-sysv check VERBOSE=1
+# tcp variant has intermittent t.tar failures in chroot environments
+make -C obj-tcp check VERBOSE=1 || make -C obj-tcp check VERBOSE=1
```
Kanishk-Bansal, line 125: LGTM: splitting obj-sysv/obj-tcp keeps sysv strict while only retrying the flaky tcp variant; %doc→%license for COPYING is correct. (Source0 moved to snapshot.debian.org + matching cgmanifest update — good.)

SPECS/fillup/fillup.spec
```diff
@@ -67,14 +67,18 @@ install -d %{buildroot}/%{_mandir}/man8
 install -m 644 SGML/fillup.8.gz %{buildroot}/%{_mandir}/man8
 
 %check
-make %{?_smp_mflags} test    OPTISPLUS="%{optflags}"
+make test    OPTISPLUS="%{optflags}"
```
Kanishk-Bansal, line 70: LGTM: dropping %{?_smp_mflags} is the right call for a non-parallel-safe test target

SPECS/debugedit/debugedit.spec
Kanishk-Bansal: dropping %{?_smp_mflags} is the right call for a non-parallel-safe test target

SPECS/keyutils/keyutils.spec
```diff
@@ -44,7 +44,12 @@ find %{buildroot} -name '*.a'  -delete
 %check
 # Installing keyutils binaries to be available for the tests to use.
 %make_install DESTDIR=/
-%make_build -k test
+# `keyctl/requesting/valid` invokes the kernel's request_key callout,
+# which spawns /sbin/request-key in the host pid/mount namespace and
+# cannot see binaries installed inside the build chroot. The other ~140
+# tests pass; failures are already printed inline by the test harness
+# (no separate test-suite.log is produced).
+%make_build -k test || :
```
Kanishk-Bansal: make -k test || : swallows all failures, not just keyctl/requesting/valid. Consider XFAIL/grep-on-summary so a real regression still fails the build.

SPECS/libtool/libtool.spec
```diff
@@ -42,7 +42,22 @@ find %{buildroot}%{_libdir} -name '*.la' -delete
...
+#     regression in libtool 2.4.7 vs binutils >= 2.39, not fixed in
+#     2.4.7. See https://debbugs.gnu.org/cgi/bugreport.cgi?bug=58105.
+#   * 169 "Run tests with low max_cmd_len" (cmdline_wrap.at:48) -- a
+#     recursive test that re-runs the suite with max_cmd_len=24. It
+#     cascades whenever any earlier test has failed (the test itself
+#     warns: "If we already have failures, then reruns will fail too!")
+#     and so always fails when 66 fails.
+# Use autotest's native range syntax to skip exactly these two; any
+# other regression still fails the build.
+make %{?_smp_mflags} check TESTSUITEFLAGS='1-65 67-168'
```
Kanishk-Bansal, line 60: LGTM (nice): TESTSUITEFLAGS='1-65 67-168' skips exactly 66 + 169 and lets any other regression hard-fail — much better than swallowing

SPECS/man-db/man-db.spec
```diff
@@ -52,7 +52,13 @@ rm %{buildroot}%{_datadir}/man/man1/zsoelim.1
...
 getent passwd man >/dev/null || useradd -c "man" -d /var/cache/man -g man \
         -s /bin/false -M -r man
-make %{?_smp_mflags} check
+# The man/check-man test runs `man -E UTF-8 -l <page>` against each
+# installed man page. lexgrog.1 trips groff's "cannot adjust line"
+# warning in the build chroot due to the missing CW (constant-width)
+# font; the other 11 pages format cleanly. Mark just lexgrog.1 as
+# expected-fail via Automake's XFAIL_TESTS so any new regression in
+# the other tests still fails the build.
+make %{?_smp_mflags} check XFAIL_TESTS="man1/lexgrog.1"
```
Kanishk-Bansal, line 61: LGTM: XFAIL_TESTS="man1/lexgrog.1" is surgical and keeps the other 11 tests strict

SPECS/nbdkit/nbdkit.spec
```diff
@@ -809,11 +809,36 @@ skip_test tests/test-cache-max-size.sh
...
+# check_PROGRAMS binary so we must mark it XFAIL instead (skip_test would
+# be overwritten when make check rebuilds the binary).
+skip_test \
+    tests/test-cc-ocaml.sh \
+    tests/test-shebang-cc-ocaml.sh
+
+# Skip file plugin dir-mode tests: upstream bug triggers
+# "free(): invalid pointer" when glibc malloc_check is enabled
+# (the spec exports MALLOC_CHECK_=1 below).
+skip_test tests/test-file-dir.sh tests/test-file-dirfd.sh
```
Kanishk-Bansal, line 835: LGTM: adding the missing test-blocksize-extents-overflow.sh completes the CVE backport; skips are itemized and test-ocaml-errorcodes correctly XFAIL'd (binary, not skip_test).

SPECS/perl-List-MoreUtils/perl-List-MoreUtils.spec
```diff
@@ -45,12 +45,9 @@ find %{buildroot} -name 'perllocal.pod' -delete
...
 %check
-# Install required module List::MoreUtils::XS for maketest
-export PERL_MM_USE_DEFAULT=1
-echo "yes" | cpan -a
-cpan local::lib
-cpan -i List::MoreUtils::XS
-make test
+# Skip XS tests - perl-List-MoreUtils-XS package is not available
+# Only run pure-perl tests
+make test TEST_FILES="t/pureperl/*.t"
```
Kanishk-Bansal, line 50: LGTM with note: limiting to t/pureperl/*.t is reasonable since the XS pkg isn't in the build env, but it does reduce coverage to pure-perl only

SPECS/python-astunparse/python-astunparse.spec
```diff
@@ -45,7 +46,19 @@ distribution; under Tools/parser in Python 3.
...
+# "AttributeError: 'Unparser' object has no attribute '_Match'"
+# when it tries to round-trip a stdlib file containing match. The
+# corresponding DumpTestCase.test_files (in test_dump.py) still
+# passes, so deselect only the broken UnparseTestCase instance.
+# Use pytest directly; setup.py test is deprecated since
+# setuptools 58 and emits warnings on every run.
+# py3_install writes to the buildroot, not the system site-packages,
+# so point PYTHONPATH at the in-tree lib/ directory for the test run.
+PYTHONPATH=lib %python3 -m pytest tests/ \
+  --deselect tests/test_unparse.py::UnparseTestCase::test_files
```
Kanishk-Bansal, line 61: LGTM: pytest + single --deselect of the _Match/3.10 case; rest stays strict

SPECS/python-cherrypy/python-cherrypy.spec
```diff
@@ -61,7 +61,33 @@ python3 setup.py install --root=%{buildroot}
...
+  --deselect cherrypy/test/test_config_server.py::ServerConfigTests::testMaxRequestSize \
+  --deselect cherrypy/test/test_conn.py::LimitedRequestQueueTests::test_queue_full \
+  --deselect cherrypy/test/test_core.py::CoreRequestHandlingTest::testRanges \
+  --deselect cherrypy/test/test_encoding.py::EncodingTests::test_multipart_decoding_bigger_maxrambytes \
+  --deselect cherrypy/test/test_encoding.py::EncodingTests::test_gc \
+  --deselect cherrypy/test/test_http.py::HTTPTests::test_post_filename_with_special_characters \
+  --deselect cherrypy/test/test_http.py::HTTPTests::test_post_multipart \
+  --deselect cherrypy/test/test_mime.py::SafeMultipartHandlingTest::test_Flash_Upload \
+  --deselect cherrypy/test/test_states.py::SignalHandlingTests::test_SIGHUP_daemonized \
+  --deselect cherrypy/test/test_tutorials.py::TutorialTest::test09Files
```
Kanishk-Bansal, line 90: LGTM: 11 enumerated --deselects, all others hard-fail.

SPECS/python-coverage/python-coverage.spec
```diff
@@ -43,8 +47,59 @@ executable, and which have been executed.
...
+  --deselect tests/test_debug.py::ShortStackTest::test_short_stack \
+  --deselect tests/test_debug.py::ShortStackTest::test_short_stack_skip \
+  --deselect tests/test_plugins.py::PluginTest::test_local_files_are_importable \
+  --deselect tests/test_report.py::SummaryTest::test_omit_files_here \
+  --deselect tests/test_report.py::SummaryTest::test_report_skip_covered_no_branches \
+  --deselect tests/test_report.py::SummaryTest::test_report_wildcard \
+  --deselect tests/test_report.py::SummaryTest::test_report_with_chdir \
+  --deselect tests/test_report.py::SummaryTest::test_run_omit_vs_report_omit \
+  --deselect tests/test_setup.py::SetupPyTest::test_metadata \
+  --deselect tests/test_testing.py::CoverageTestTest::test_sub_python_is_this_python
```
Kanishk-Bansal, line 102: LGTM (thorough): replaces the silently-no-op tox chain with a real pytest run; the flaky-in-SPECS-EXTENDED rationale and CTracer .so copy are well explained.

SPECS/python-distlib/python-distlib.spec
```diff
@@ -57,13 +57,17 @@ pip3 install iniconfig
 # on the ordering of the input, hence disabling it.
 # https://github.com/pypa/distlib/issues/161
 # test_is_writable depends on network access
-%pytest -k "not test_sequencer_basic and not test_is_writable"
+%pytest -k "not test_sequencer_basic and not test_is_writable and not test_package_data"
```
Kanishk-Bansal, line 60: LGTM: one extra -k exclusion (test_package_data).

SPECS/python-fields/python-fields.spec
```diff
@@ -71,11 +71,12 @@ rm -rf docs/_build/html/.buildinfo docs/_build/html/.doctrees
 
 %check
 # Perf tests require unmaintained 'characteristic' module
+# Pin pytest<9: pytest 9.x rejects duplicate parametrize IDs (test_tuple_pickle uses 'loads' twice)
 pip3 install atomicwrites>=1.3.0 \
     attrs>=19.1.0 \
     more-itertools>=7.0.0 \
     pluggy>=0.11.0 \
-    pytest>=5.4.0 \
+    'pytest>=5.4.0,<9' \
```
Kanishk-Bansal, line 79: LGTM: pytest<9 pin for the duplicate-parametrize-ID issue.

SPECS/python-filelock/python-filelock.spec
```diff
@@ -2,13 +2,16 @@
...
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        Unlicense
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 URL:            https://github.com/toxdev/filelock
 Source0:        https://files.pythonhosted.org/packages/source/f/%{srcname}/%{srcname}-%{version}.tar.gz#/%{name}-%{version}.tar.gz
 Patch0:         remove-python-3.14-classifier.patch
+%if 0%{?with_check}
+Patch1:         filelock-tests-isolate-caplog.patch
```
Kanishk-Bansal, line 13: LGTM: test-only caplog.clear() patch (gated by with_check) fixes the aarch64 async-leak flake without touching library code

SPECS/python-gevent/python-gevent.spec
```diff
@@ -64,14 +64,32 @@ Features include:
 
 %check
 # freeze packaging since we already have it available
-pip3 install packaging==23.2 tox tox-current-env 
+pip3 install packaging==23.2 tox tox-current-env
```
Kanishk-Bansal, line 67: LGTM (nice): replaces %tox || : with properly scoped patch exclusions — strictly better than the old blanket tolerate

SPECS/python-google-pasta/python-google-pasta.spec
```diff
@@ -33,7 +34,10 @@ Enable python source code refactoring through AST modifications.
 %py3_install
 
 %check
-python3 setup.py test
+pip3 install pytest
+# Exclude PrefixSuffixGoldenTest (51 failures from Python 3.12 AST changes)
+# and fstring/inline tests that error due to same AST changes
+python3 -m pytest -v -k "not (PrefixSuffixGoldenTest or fstring or test_inline_conditional_fails or test_inline_function_fails or test_inline_non_assign_fails or test_inline_non_constant_fails)"
```
Kanishk-Bansal, line 40: LGTM: pytest + -k exclusion of the 3.12 AST-change cases.

SPECS/python-logutils/python-logutils.spec
```diff
@@ -38,6 +38,10 @@ Requires:       python3-libs
 
 %prep
 %setup -q -n %{pkgname}-%{version}
+# Fix Python 3.12 compat: assertEquals removed, and tests need explicit log level
```
Kanishk-Bansal, line 41: LGTM: %prep seds for assertEquals→assertEqual + explicit setLevel are minimal 3.12-compat fixes

SPECS/python-logutils/python-logutils.spec
```diff
@@ -47,6 +51,7 @@ python3 setup.py install --root=%{buildroot}
 
 %if %{with check}
 %check
+# Patch fixes Python 3.12 compat: assertEquals -> assertEqual, explicit logger level
```
Kanishk-Bansal, line 54:
```suggestion
# sed fixes Python 3.12 compat: assertEquals -> assertEqual, explicit logger level
```

SPECS/python-markdown/python-markdown.spec
```diff
@@ -37,6 +37,9 @@ there are a few known issues.
 
 %prep
 %autosetup -p1 -n %{pkgname}-%{version}
+# Skip 2 tests that fail due to Python 3.12 html.parser behavior changes
+sed -i 's/def test_raw_missing_close_bracket/def _skip_test_raw_missing_close_bracket/' tests/test_syntax/blocks/test_html_blocks.py
+sed -i 's/def test_unclosed_comment_/def _skip_test_unclosed_comment_/' tests/test_syntax/blocks/test_html_blocks.py
```
Kanishk-Bansal, line 42: LGTM

SPECS/python-pbr/python-pbr.spec
```diff
@@ -42,8 +44,22 @@ export SKIP_PIP_INSTALL=1
...
+# build env containing only setuptools+wheel. pbr's own setup.py starts with
+# 'from pbr import util', which fails inside that isolated env
+# (ModuleNotFoundError: No module named 'pbr'). PYTHONPATH cannot reach in
+# either, because pip strips it from the build subprocess. tox 'usedevelop'
+# papered over this historically by injecting an editable pbr install before
+# the tests spawned the inner venv, but that workflow is broken with current
+# setuptools. The self-bootstrap is fixed upstream in pbr 6.1+.
+pip3 install stestr testscenarios testresources six
+python3 -m stestr run --suppress-attachments \
+    --exclude-regex '(test_freeze_command|test_console_script_develop|test_console_script_install|test_pep_517_support|test_requirement_parsing)'
```
Kanishk-Bansal, line 62: LGTM: direct stestr with 5 documented excludes is a sound replacement for the broken tox+virtualenv editable install.

SPECS/python-pexpect/python-pexpect.spec
```diff
@@ -59,6 +59,8 @@ pty module.
 
 %prep
 %autosetup -p1 -n %{modname}-%{version}
+# Fix Python 3.12: assertRaisesRegexp removed, use assertRaisesRegex
+sed -i 's/assertRaisesRegexp/assertRaisesRegex/g' tests/test_expect.py tests/test_misc.py tests/test_popen_spawn.py
```
Kanishk-Bansal, line 63: LGTM: assertRaisesRegex sed + pytest-timeout + single retry; ignores need SSH / removed asyncio.coroutine.

SPECS/python-pytest-benchmark/python-pytest-benchmark.spec
```diff
@@ -65,9 +65,10 @@ pip3 install atomicwrites>=1.3.0 \
     aspectlib \
     pygal \
     pytest-xdist
-PATH=%{buildroot}%{_bindir}:${PATH} \
-PYTHONPATH=%{buildroot}%{python3_sitelib} \
-    python%{python3_version} -m pytest -v tests -k 'not (test_commit_info_error or test_basic or test_skip or test_disable or test_only_benchmarks)'
+# Install the package itself so pytest can discover the plugin entry point
+pip3 install -e .
```
Kanishk-Bansal, line 69: LGTM

SPECS/python-pytest-cov/python-pytest-cov.spec
```diff
@@ -2,13 +2,14 @@
...
 Version:        4.1.0
-Release:        1%{?dist}
+Release:        2%{?dist}
 License:        MIT
 Vendor:         Microsoft Corporation
 Distribution:   Azure Linux
 URL:            https://pypi.python.org/pypi/pytest-cov
 Source0:        https://github.com/pytest-dev/%{srcname}/archive/v%{version}/%{srcname}-%{version}.tar.gz#/%{name}-%{version}.tar.gz
 Patch0:         0001-skip-tests-that-are-expected-to-fail.patch
+Patch1:         0002-skip-flaky-dist-and-html-tests.patch
```
Kanishk-Bansal, line 12: LGTM

SPECS/python-pytest-xdist/python-pytest-xdist.spec
```diff
@@ -36,6 +36,11 @@ BuildRequires:  python3-setuptools_scm
 BuildRequires:  python3-wheel
 
 Patch0:         pytest-xdist-3.5.0-fix-pytest9-parser.patch
+# Backport of upstream PR pytest-dev/pytest-xdist#1318: sort loadscope
+# nodes by gateway id before assigning work units, so worker bootstrap
+# order does not change the assignment. Fixes flaky
+# test_workqueue_ordered_by_size (upstream issues #985, #1068, #1248).
+Patch1:         pytest-xdist-3.5.0-fix-loadscope-deterministic-985.patch
```
Kanishk-Bansal, line 43: LGTM

SPECS/python-remoto/python-remoto.spec
```diff
@@ -36,6 +36,8 @@ remoto as well to inspect Ceph clusters.
 
 %prep
 %autosetup -n %{pkgname}-%{version}
+# Fix pytest compatibility: 'setup' not auto-called on plain classes in newer pytest
+sed -i 's/def setup(self)/def setup_method(self)/' remoto/tests/test_process.py remoto/tests/test_util.py
```
Kanishk-Bansal, line 40: LGTM: setup→setup_method rename for newer pytest

SPECS/python-sphinx/python-sphinx.spec
```diff
@@ -191,22 +192,42 @@ mkdir %{buildroot}%{python3_sitelib}/sphinxcontrib
...
+  --ignore tests/test_build_texinfo.py \
+  --ignore tests/test_domain_std.py \
+  --ignore tests/test_smartquotes.py \
+  --deselect tests/test_ext_autodoc.py::test_enum_class \
+  --deselect tests/test_ext_autodoc_configs.py::test_autodoc_type_aliases \
+  --deselect tests/test_ext_autodoc_configs.py::test_autodoc_default_options \
+  --deselect tests/test_ext_viewcode.py::test_viewcode_linenos \
+  --deselect tests/test_ext_viewcode.py::test_viewcode \
+  --deselect tests/test_intl.py::test_additional_targets_should_not_be_translated \
+  --deselect tests/test_intl.py::test_additional_targets_should_be_translated
```
Kanishk-Bansal, line 230: LGTM: python3-filelock BR + dropping the no-op pip bootstrap + itemized ignores/deselects; html5lib/SPECS-EXTENDED rationale is sound

SPECS/python-sqlalchemy/python-sqlalchemy.spec
```diff
@@ -77,6 +77,9 @@ This package includes the python 3 version of the module.
 
 %prep
 %setup -q -n %{srcname}-%{srcversion}
+# Fix Python 3.12: get_event_loop() raises DeprecationWarning when no current loop exists
+# Replace with creating a new event loop when none is running
+sed -i 's/return asyncio.get_event_loop_policy().get_event_loop()/loop = asyncio.new_event_loop(); asyncio.set_event_loop(loop); return loop/' lib/sqlalchemy/util/_concurrency_py3k.py
```
Kanishk-Bansal, line 82: LGTM + %license doc fix are fine

SPECS/python-toolz/python-toolz.spec
```diff
@@ -61,15 +61,22 @@ BuildRequires:  python%{python3_pkgversion}-pip
 %py3_install
 
 %check
-pip3 install nose
-nosetests
+pip3 install packaging==23.2 pytest
+# nose is incompatible with Python 3.12 (no 'imp' module), using pytest instead
+# Exclude test_inspect_wrapped_property (Python 3.12 inspect behavior change)
+python3 -m pytest toolz/tests/ -k "not test_inspect_wrapped_property"
```
Kanishk-Bansal, line 67: LGTM: nose→pytest (nose is dead on 3.12), one inspect-change exclusion

SPECS/python-tox-current-env/python-tox-current-env.spec
```diff
@@ -48,15 +48,22 @@ The tox-current-env plugin allows to run tests in current Python environment.
 
 
 %check
-pip3 install tox
-tox -e py%{python3_version_nodots}
+# pin packaging==23.2 to avoid uninstall conflict with system RPM
+pip3 install packaging==23.2 tox
+# 177/185 integration tests fail: they invoke tox internally which fails in chroot
+tox -e py%{python3_version_nodots} || :
```
Kanishk-Bansal, line 54: needs review tox ... || : makes %check effectively non-gating

SPECS/python-twisted/python-twisted.spec
```diff
@@ -71,13 +71,23 @@ ln -s ckeygen %{buildroot}/%{_bindir}/ckeygen3
...
 
 %check
-route add -net 224.0.0.0 netmask 240.0.0.0 dev lo
-chmod g+w . -R
-useradd test -G root -m
-sudo -u test pip3 install --upgrade pip
-sudo -u test pip3 install 'tox>=3.27.1,<4.0.0' PyHamcrest cython-test-exception-raiser
-chmod g+w . -R
-LANG=en_US.UTF-8 sudo -u test /home/test/.local/bin/tox -e nocov-posix-alldeps
+# The full upstream tox suite ('nocov-posix-alldeps') runs >10000 tests and
```
Kanishk-Bansal, line 74: might need a re review as we are swaping the full >10k-test suite for a ~200-test trial smoke subset

SPECS/python-urllib3/python-urllib3.spec
```diff
@@ -49,10 +49,10 @@ rm -rf test/contrib/
 %pyproject_install
 
 %check
-pip3 install --upgrade pip
+# Pin pytest<9: pytest 9.x rejects non-Collection iterables (chain) in parametrize argvalues
```
Kanishk-Bansal, line 52: LGTM

SPECS/m2crypto/m2crypto.spec
```diff
@@ -48,16 +50,50 @@ messenger for Zope.
...
+# (with the six.py shim so reviewers see the real underlying error) and
+# still install `parameterized`, so the day this package is updated the
+# %check section starts being meaningful with no further work.
 pip3 install parameterized
-#Testing: MiscSSLClientTestCase failing with SSLError not raised
-%python3 setup.py test
+cp -f $(%python3 -c "import six; print(six.__file__)") \
+  %{buildroot}%{python3_sitearch}/M2Crypto/six.py
+PYTHONPATH=%{buildroot}%{python3_sitearch} \
+  %python3 -m pytest tests/ -k "not test_tls1_nok" || :
```
Kanishk-Bansal, line 76: have you checked 0.48.0, does that has the fix?

SPECS/keyutils/keyutils.spec
```diff
@@ -73,6 +80,14 @@ find %{buildroot} -name '*.a'  -delete
 %{_mandir}/man3/*
 
 %changelog
+* Wed Jun 17 2026 Kshitiz Godara <kgodara@microsoft.com> - 1.6.3-2
+- Skip the chroot-unfriendly `keyctl/requesting/` tests in %%check
```
Kanishk-Bansal, line 84: LGTM now, thanks for making the changes

Kanishk-Bansal (approved):
all ptest fixes looks good to me, I have reviewed each package and have put the comments above
for `python-tox-current-env` mostly ptest fails hence added ||
for `python-twisted` we are taking a basic sanity check tests.

Rest all tests are proper fixes and skips.
Thank you so much @kgodara912 for doing this. 🥇

SPECS/python-tox-current-env/python-tox-current-env.spec
kgodara912: This I will address in a new PR.

SPECS/python-twisted/python-twisted.spec
kgodara912: This is kept as a minimal test set till we merge the upgrade PR. I will update the upgrade PR with these removal.

SPECS/m2crypto/m2crypto.spec
kgodara912: I haven't checked from upgrade perspective due to test only changes in this PR.

kgodara912:
[Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1149933&view=results) with all the packages. The GitHub PR check pipelines are running which will also build and test packages.

kgodara912:
Closing in favor of: https://github.com/microsoft/azurelinux/pull/17744/

PR #17755, opened 2026-06-17
Fix failing core ptest in systemd and libguestfs

kgodara912:
## Summary

Fix failing `%check` (ptest) stages for **`systemd`** and **`libguestfs`** in the Azure Linux build chroot. No upstream version changes — release-bump + targeted in-spec test fixes only.

## Changes

| Package | Fix |
|---|---|
| `systemd` | Skip tests in `%check` that require capabilities not available in the build chroot |
| `libguestfs` | Switch `make quickcheck` to the direct backend with `force_tcg`; extend supermin appliance with `sqlite*` |

### `systemd`

Constructs an explicit test list and excludes tests that fail in a chroot but pass in a real environment:

- **`test-fd-util`, `test-mount-util`, `test-mountpoint-util`, `test-path-util`, `test-time-util`, `test-calendarspec`, `test-date`** — need mount-namespace privileges (`MS_SLAVE` remount fails with `EINVAL` in the chroot).
- **`test-udev`** — invokes `systemd-detect-virt`, which is not on `PATH` yet at this stage.
- **`test-rpm-macros`** — environmental mismatch with the build chroot.

### `libguestfs`

- Force the **direct** libguestfs backend with `force_tcg` in `%check` so `make quickcheck` no longer requires a running `libvirtd` or `/var/run/libvirt/libvirt-sock`, and does not need `/dev/kvm`.
- Add `sqlite`, `sqlite-libs`, `sqlite-devel` to the supermin appliance via `--with-extra-packages` so `guestfsd` (which depends on `libsqlite3` through `librpm`) can start inside the appliance.
- Make the QEMU log dump on failure non-fatal when the log directory is missing.

## Verification

- [x] Both `Release:` bumps to the packages
- [x] No upstream version changes
- [x] No toolchain manifest update needed (neither `systemd` nor `libguestfs` appears in any toolchain / pkggen_core manifest; `systemd-bootstrap-250.3-19` in the manifests is a separate `SPECS/systemd-bootstrap/systemd-bootstrap.spec`, not affected)

## Related

#17749 ("Fix failing core ptest across packages")


kgodara912:
[Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1141176&view=results)

kgodara912:
[Buddy build](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1141693&view=results) for systemd after the recent changes in systemd.

Kanishk-Bansal (approved):
LGTM, 

in libguestfs forcing the direct backend with force_tcg is a nice minimal approach, since we don't have libvirtd / /var/run/libvirt/libvirt-sock (or /dev/kvm) in the build chroot.

in systemd the pakcage conflict refactor is a good add on which is unrelated to ptest fix, we are basically skipping the test although these passes in non chroot env, but due to env constraints we are skipping it here.

SPECS/libguestfs/libguestfs.spec
```diff
@@ -850,8 +850,9 @@ extra=--with-supermin-packager-config=$(pwd)/yum.conf
 %if %{without appliances}
   --enable-appliance=no \
 %endif
-  --disable-erlang
-  $extra
+  --disable-erlang \
+  $extra \
+  --with-extra-packages="sqlite sqlite-libs sqlite-devel"
```
Kanishk-Bansal, line 855:
```suggestion
  --with-extra-packages="sqlite*"
```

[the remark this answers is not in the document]
kgodara912: will check and update accordingly.

kgodara912: This doesn't work as it is not part of shell expansion.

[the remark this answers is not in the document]
kgodara912: The purpose of this line is different, it provides packages inside the supermin guestfs. The above alone is insufficient.

SPECS/systemd/systemd.spec
```diff
@@ -918,7 +922,16 @@ cp %{buildroot}/usr/lib/systemd/boot/efi/systemd-bootaa64.efi %{buildroot}/boot/
 
 %check
 %if %{with tests}
-meson test -C %{_vpath_builddir} -t 6 --print-errorlogs
+# Skip tests that require capabilities not available in the build chroot:
```
mfrw, line 929: 🥇

mfrw (approved):

kgodara912:
Buddy build for this is successful. The PR check failures are also not there for systemd and libguestfs. All the changes are related to tests. The PR ensures that any future genuine failures for these packages, we check properly instead of assuming that it was always failing. @PawelWMS, could you please check and if looks good, could you please approve it?

SPECS/libguestfs/libguestfs.spec
```diff
@@ -850,8 +850,9 @@ extra=--with-supermin-packager-config=$(pwd)/yum.conf
 %if %{without appliances}
   --enable-appliance=no \
 %endif
-  --disable-erlang
-  $extra
+  --disable-erlang \
+  $extra \
```
jslobodzian, line 854: Does this need to be set for non-check builds too?

kgodara912: Yes @jslobodzian, actually it does require these extra libraries to make the package work properly. Without these the supermin never worked in our images, this was the orginal PR which has all the details, https://github.com/microsoft/azurelinux/pull/13288

LeoMar4 (approved):

PR #17762, opened 2026-06-18
rocm-examples: drop FetchContent fallback for google-benchmark

himaja-kesari:
Under network-isolation Koji builds (azl4-bootstrap-network-isolation), rocm-examples 6.4.2 fails in %build because
Tutorials/reduction/benchmark/CMakeLists.txt does
find_package(benchmark CONFIG QUIET) and, if benchmark::benchmark is not created, falls through to FetchContent_MakeAvailable() that git-clones https://github.com/google/benchmark.git at v1.6.1. With network blocked in the mock chroot, this fails:

  Could not resolve host: github.com

google-benchmark-devel is already a BuildRequires for non-SUSE builds, so the network fallback is not needed. Add a small overlay patch that replaces the find_package(QUIET)/FetchContent fallback with a single find_package(benchmark CONFIG REQUIRED), so the build either picks up the system google-benchmark or fails fast with a clear error instead of trying to reach the internet.

Also moves the rocm-examples component definition out of the inline components.toml entry into its own rocm-examples.comp.toml so the overlay configuration has a home.

### DIFF
--- a/base/comps/components.toml
+++ b/base/comps/components.toml
@@ -4524,7 +4524,7 @@ includes = ["**/*.comp.toml", "component-check-disablement.toml", "component-min
 [components.rocm-bandwidth-test]
 [components.rocm-cmake]
 [components.rocm-core]
-[components.rocm-examples]
+# rocm-examples is defined in base/comps/rocm-examples/rocm-examples.comp.toml
 [components.rocm-omp]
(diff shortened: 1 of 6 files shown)


reubeno:
Thanks for your work on this! We were able to adapt this and have unblocked this package from building.

PR #17775, opened 2026-06-19
fix(check-rendered-specs): add source-origin preflight guard

liunan-ms:
This PR adds a preflight source-origin allowlist check before `azldev component update/render` in the rendered-spec PR path. The new guard validates the resolved Fedora dist-git and lookaside URIs from `azldev config dump` and fails fast if they differ from the trusted Fedora endpoints, preventing fork PRs from redirecting source fetches to attacker-controlled infrastructure.

Fixes: [AB#22017](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22017)

### DIFF
--- a/scripts/ci/render-specs-check/source_origin_guard.py
+++ b/scripts/ci/render-specs-check/source_origin_guard.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+"""Guard source-origin configuration for rendered-spec checks."""
+
+from __future__ import annotations
+
+import json
+import os
+import subprocess
+import sys
+from typing import Any
+
+# Intentionally duplicated from trusted defaults in distro/fedora.distro.toml.
+# Keep this allowlist in sync when Fedora source origins are intentionally updated.
+ALLOWED_FEDORA_DIST_GIT_BASE_URI = "https://src.fedoraproject.org/rpms/$pkg.git"
+ALLOWED_FEDORA_LOOKASIDE_BASE_URI = "https://src.fedoraproject.org/repo/pkgs/$pkg/$filename/$hashtype/$hash/$filename"
+ALLOWED_FEDORA_REPO_BASE_URIS = {
+    "https://na.edge.kernel.org/fedora/releases/$releasever/Everything/source/tree",
+    "https://na.edge.kernel.org/fedora/updates/$releasever/Everything/source/tree",
+    "https://na.edge.kernel.org/fedora/development/$releasever/Everything/source/tree",
+}
+
+
+def _run(
+    cmd: list[str], *, cwd: str | None = None, env: dict[str, str] | None = None
+) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(
+        cmd,
+        cwd=cwd,
+        env=env,
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+
+
+def _load_config() -> dict[str, Any]:
+    env = os.environ.copy()
+    env["AZLDEV_ALLOW_ROOT"] = "1"
+    result = _run(["azldev", "config", "dump", "-q", "-f", "json"], env=env)
+    if result.returncode != 0:
+        stderr = result.stderr.strip()
+        if stderr:
+            print(f"::error::azldev config dump failed: {stderr}", file=sys.stderr)
+        else:
+            print("::error::azldev config dump failed", file=sys.stderr)
+        raise SystemExit(result.returncode)
+    return json.loads(result.stdout)
+
+
+def _validate_fedora_sources(config: dict[str, Any]) -> list[str]:
+    errors: list[str] = []
+    fedora = config.get("distros", {}).get("fedora")
+    if not isinstance(fedora, dict):
+        return ["Resolved config is missing distros.fedora"]
+
+    dist_git = fedora.get("distGitBaseUri")
+    lookaside = fedora.get("lookasideBaseUri")
+    repos = fedora.get("repos")
+
+    if dist_git != ALLOWED_FEDORA_DIST_GIT_BASE_URI:
+        errors.append(
+            f"distros.fedora.distGitBaseUri resolved to {dist_git!r}, expected {ALLOWED_FEDORA_DIST_GIT_BASE_URI!r}"
+        )
+
+    if lookaside != ALLOWED_FEDORA_LOOKASIDE_BASE_URI:
+        errors.append(
+            f"distros.fedora.lookasideBaseUri resolved to {lookaside!r}, expected {ALLOWED_FEDORA_LOOKASIDE_BASE_URI!r}"
+        )
+
+    if not isinstance(repos, list):
+        errors.append("distros.fedora.repos is missing or not a list")
+        return errors
+
+    actual_repos: set[str] = set()
+    for index, item in enumerate(repos):
+        if not isinstance(item, dict):
+            errors.append(f"distros.fedora.repos entry {index} is not a dictionary: {item!r}")
+            continue
+        base_uri = item.get("baseUri")
+        if not isinstance(base_uri, str):
+            errors.append(f"distros.fedora.repos entry {index} has non-string baseUri: {base_uri!r}")
+            continue
+        actual_repos.add(base_uri)
+    missing_repos = sorted(ALLOWED_FEDORA_REPO_BASE_URIS - actual_repos)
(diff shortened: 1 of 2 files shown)


scripts/ci/render-specs-check/source_origin_guard.py
```diff
@@ -0,0 +1,112 @@
...
+    for item in repos:
+        if not isinstance(item, dict):
+            continue
+        base_uri = item.get("baseUri")
+        if isinstance(base_uri, str):
+            actual_repos.add(base_uri)
+    missing_repos = sorted(ALLOWED_FEDORA_REPO_BASE_URIS - actual_repos)
+    extra_repos = sorted(actual_repos - ALLOWED_FEDORA_REPO_BASE_URIS)
+
+    if missing_repos:
```
tobiasb-ms, line 87: question (non-blocking): What's the actual harm in missing an approved repo? Is it just so we can clean it up if a repo is no longer being used?

scripts/ci/render-specs-check/source_origin_guard.py
```diff
@@ -0,0 +1,112 @@
...
+        errors.append("Missing approved Fedora repo base URIs:")
+        errors.extend(f"  - {uri}" for uri in missing_repos)
```
tobiasb-ms: nitpick(non-blocking): For most uris that you print, you use `!r`, which seems to add single quotes. Do that here for consistency.
```suggestion
        errors.extend(f"  - {uri!r}" for uri in missing_repos)
```

scripts/ci/render-specs-check/source_origin_guard.py
```diff
@@ -0,0 +1,112 @@
...
+    if extra_repos:
+        errors.append("Unexpected Fedora repo base URIs:")
+        errors.extend(f"  - {uri}" for uri in extra_repos)
```
tobiasb-ms: nitpick(non-blocking): For most uris that you print, you use `!r`, which seems to add single quotes. Do that here for consistency.
```suggestion
        errors.extend(f"  - {uri!r}" for uri in extra_repos)
```

scripts/ci/render-specs-check/source_origin_guard.py
```diff
@@ -0,0 +1,112 @@
...
+            f"distros.fedora.lookasideBaseUri resolved to {lookaside!r}, expected {ALLOWED_FEDORA_LOOKASIDE_BASE_URI!r}"
+        )
+
+    if not isinstance(repos, list):
+        errors.append("distros.fedora.repos is missing or not a list")
+        return errors
+
+    actual_repos: set[str] = set()
+    for item in repos:
+        if not isinstance(item, dict):
```
tobiasb-ms, line 76: question(blocking): Is this actually a valid case where we should silently ignore it?

scripts/ci/render-specs-check/source_origin_guard.py
```diff
@@ -0,0 +1,112 @@
...
+            continue
+        base_uri = item.get("baseUri")
+        if isinstance(base_uri, str):
```
tobiasb-ms: question(blocking): Is it actually valid for `base_uri` to not be a string, or should we report an error on that?

tobiasb-ms (changes requested):

scripts/ci/render-specs-check/source_origin_guard.py
liunan-ms: Missing an approved repo means the resolved config is incomplete/broken, which should be caught early rather than silently failing later during fetch. Although repo removal is possible and expected over time, the key for the guard is: if a URI is intentionally removed, update the approved-origin policy in the same PR so the trust model stays explicit and auditable.

scripts/ci/render-specs-check/source_origin_guard.py
liunan-ms: Thanks for catching this! Updated to use {uri!r} for consistency with the rest of the script output formatting.

scripts/ci/render-specs-check/source_origin_guard.py
liunan-ms: Updated to use {uri!r} for consistency with the rest of the script output formatting.

scripts/ci/render-specs-check/source_origin_guard.py
liunan-ms: Agreed this should not be ignored. I changed the loop to emit an explicit validation error when a repos entry is not an object, instead of silently skipping it.

scripts/ci/render-specs-check/source_origin_guard.py
liunan-ms: Yes, agreed. I updated validation to fail when baseUri is missing or not a string, with an explicit error message that includes the entry index/value.

scripts/ci/render-specs-check/source_origin_guard.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+
+    if not isinstance(repos, list):
+        errors.append("distros.fedora.repos is missing or not a list")
+        return errors
+
+    actual_repos: set[str] = set()
+    for index, item in enumerate(repos):
+        if not isinstance(item, dict):
+            errors.append(f"distros.fedora.repos entry {index} is not an object: {item!r}")
```
tobiasb-ms: nitpick(non-blocking): It would be more clear to say this isn't a dictionary rather than the generic object:
```suggestion
            errors.append(f"distros.fedora.repos entry {index} is not a dictionary: {item!r}")
```

tobiasb-ms (approved):
issue(non-blocking): I'd prefer if you rebased so this is one commit -- the second commit is logically part of the initial implementation.

scripts/ci/render-specs-check/source_origin_guard.py
```diff
@@ -0,0 +1,115 @@
...
+import sys
+from typing import Any
+
+ALLOWED_FEDORA_DIST_GIT_BASE_URI = "https://src.fedoraproject.org/rpms/$pkg.git"
+ALLOWED_FEDORA_LOOKASIDE_BASE_URI = "https://src.fedoraproject.org/repo/pkgs/$pkg/$filename/$hashtype/$hash/$filename"
+ALLOWED_FEDORA_REPO_BASE_URIS = {
+    "https://na.edge.kernel.org/fedora/releases/$releasever/Everything/source/tree",
+    "https://na.edge.kernel.org/fedora/updates/$releasever/Everything/source/tree",
+    "https://na.edge.kernel.org/fedora/development/$releasever/Everything/source/tree",
+}
```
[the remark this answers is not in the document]
tobiasb-ms: (non-blocking): This seems like a good comment -- I suggest doing something with it.

scripts/ci/render-specs-check/source_origin_guard.py
liunan-ms: Updated the error text to ‘not a dictionary’ for clarity

scripts/ci/render-specs-check/source_origin_guard.py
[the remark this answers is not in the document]
liunan-ms: Added a comment linking these constants to distro/fedora.distro.toml and explaining the intentional duplication.

scripts/ci/render-specs-check/source_origin_guard.py
```diff
@@ -0,0 +1,115 @@
...
+
+def _validate_fedora_sources(config: dict[str, Any]) -> list[str]:
+    errors: list[str] = []
+    fedora = config.get("distros", {}).get("fedora")
+    if not isinstance(fedora, dict):
+        return ["Resolved config is missing distros.fedora"]
+
+    dist_git = fedora.get("distGitBaseUri")
+    lookaside = fedora.get("lookasideBaseUri")
+    repos = fedora.get("repos")
```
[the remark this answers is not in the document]
liunan-ms: Current guard is Fedora-key scoped, this PR closes the exploitable fork path. I agree broader all distros + component distro-name allowlist validation is useful defense-in-depth, but that should be a follow-up hardening scope beyond this fix.

christopherco (changes requested):
Code changes themselves look good. Requesting to update your commit message in your commit to add a short explanation of the changes. We're doing `rebase-merge` commits of PRs now, so the commit itself is what is added to the final git tree, including its commit title and message. We need our future `git log` to have useful commit message information.

christopherco (approved):

christopherco (approved):

PR #17782, opened 2026-06-23
fix(qt6-qtmultimedia): remove leftover GStreamer media plugin devel f…

WithEnoughCoffee:
PR #17652 dropped the gstreamer-plugins-bad BuildRequires (so the GStreamer media plugin is no longer built) and cleaned up the runtime %files, but the matching %files devel entries for QtGstreamerMediaPluginImpl were missed. This broke qt6-qtmultimedia-devel packaging. This PR removes those leftover devel %files entries to fix the build.

Note: an earlier revision of this PR also overlaid away a duplicate %dir for QtQuick3D/SpatialAudio ("File listed twice") and an unescaped %{version} in a comment ("Macro expanded in comment"). A full build confirmed both are only non-fatal RPM *warnings*, not errors, so those overlays were removed to minimize divergence from the Fedora spec.

### DIFF
--- a/base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
+++ b/base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
@@ -36,3 +36,18 @@ type = "spec-search-replace"
 section = "%files"
 regex = '^%\{_qt6_plugindir\}/multimedia/libgstreamermediaplugin\.so$'
 replacement = ''
+
+# Dropping the gstreamer-plugins-bad BR also stops the GStreamer media plugin
+# devel artifacts (QtGstreamerMediaPluginImpl headers/.a/.prl/cmake) from being
+# built. The runtime %files entries were removed above, but the matching
+# %files devel entries were missed, so qt6-qtmultimedia-devel packaging fails
+# with "File/Directory not found". A single line-based regex drops every devel
+# %files entry that references QtGstreamerMediaPluginImpl (headers, .a, .prl,
+# the cmake %dir, and the cmake glob).
+[[components.qt6-qtmultimedia.overlays]]
+description = "Remove all QtGstreamerMediaPluginImpl devel %files entries (not built without gstreamer-plugins-bad)"
+type = "spec-search-replace"
+section = "%files"
+package = "devel"
+regex = '^.*GstreamerMediaPluginImpl.*$'
+replacement = ''
--- a/locks/qt6-qtmultimedia.lock
+++ b/locks/qt6-qtmultimedia.lock
@@ -2,5 +2,5 @@
 version = 1
 import-commit = '66b4dd09782f23900e04dc2a93470c6ef8a02537'
 upstream-commit = '66b4dd09782f23900e04dc2a93470c6ef8a02537'
-input-fingerprint = 'sha256:796992d13039bed1d8a28f20772baec8dcbb76e0f433968e82c6767aabb56d2e'
+input-fingerprint = 'sha256:7912965e3b6ac8f0339d2f8e6ceff4805e727df0bcf72dd305691ae15153f20a'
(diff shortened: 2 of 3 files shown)


dmcilvaney (approved):

base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
```diff
@@ -36,3 +36,67 @@ type = "spec-search-replace"
...
+
+[[components.qt6-qtmultimedia.overlays]]
+description = "Remove Qt6GstreamerMediaPluginImplPrivate cmake files from %files devel (not built without gstreamer-plugins-bad)"
+type = "spec-search-replace"
+section = "%files"
+package = "devel"
+regex = '^%\{_qt6_libdir\}/cmake/Qt6GstreamerMediaPluginImplPrivate/\*\.cmake$'
+replacement = ''
+
+# The QtQuick3D/SpatialAudio directory is listed twice: once as a bare %dir and
```
tobiasb-ms: issue(blocking): Usually warnings like this are non-fatal. (Same with the `macro expanded in comment` warning below). If they aren't actually fatal I'd prefer to not drift any further from the Fedora spec than we need to. So unless the warnings really are fatal, please remove these overlays.

tobiasb-ms (changes requested):

tobiasb-ms (changes requested):
issue(blocking): Changes look good but please rebase this down to one commit -- they don't really make sense as separate commits.

base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
```diff
@@ -36,3 +36,48 @@ type = "spec-search-replace"
...
+regex = '^%dir %\{_qt6_libdir\}/cmake/Qt6GstreamerMediaPluginImplPrivate$'
+replacement = ''
+
+[[components.qt6-qtmultimedia.overlays]]
+description = "Remove Qt6GstreamerMediaPluginImplPrivate cmake files from %files devel (not built without gstreamer-plugins-bad)"
+type = "spec-search-replace"
+section = "%files"
+package = "devel"
+regex = '^%\{_qt6_libdir\}/cmake/Qt6GstreamerMediaPluginImplPrivate/\*\.cmake$'
+replacement = ''
```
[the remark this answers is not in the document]
WithEnoughCoffee: Good catch — the description was stale. I've updated it.

WithEnoughCoffee:
> issue(blocking): Changes look good but please rebase this down to one commit -- they don't really make sense as separate commits.

Got that taken care of.

tobiasb-ms (approved):

base/comps/qt6-qtmultimedia/qt6-qtmultimedia.comp.toml
```diff
@@ -36,3 +36,48 @@ type = "spec-search-replace"
...
+# devel artifacts (QtGstreamerMediaPluginImpl headers/.a/.prl/cmake) from being
+# built. The runtime %files entries were removed above, but the matching
+# %files devel entries were missed, so qt6-qtmultimedia-devel packaging fails
+# with "File/Directory not found". Remove them here too.
+[[components.qt6-qtmultimedia.overlays]]
+description = "Remove QtGstreamerMediaPluginImpl headers from %files devel (not built without gstreamer-plugins-bad)"
+type = "spec-search-replace"
+section = "%files"
+package = "devel"
+regex = '^%\{_qt6_headerdir\}/QtGstreamerMediaPluginImpl/$'
```
reubeno: Could these 4 regexes be unified? Can we just remove all %files entries that mention GstreamerMediaPluginImpl?

WithEnoughCoffee: That's a really good idea. I'll work on that.

reubeno (approved):

PR #17794, opened 2026-06-24
feat(wsl): add POSIX text/scripting baseline tools

mfrw:
The WSL image list enumerates user-facing basics explicitly (tar, which,
less, curl, file) but omitted several classic POSIX text-processing and
scripting tools that scripts and interactive users universally assume.

Add them so they are guaranteed present rather than relying on transitive
dependencies of a minimal install:

  - gawk      (awk)
  - diffutils (diff, cmp)
  - findutils (find, xargs)
  - patch
  - xz        (.tar.xz is one of the dominant tarball format)
  - grep
  - sed
  - gzip      (gzip, gunzip, zcat)

### DIFF
--- a/base/images/wsl/wsl.kiwi
+++ b/base/images/wsl/wsl.kiwi
@@ -39,9 +39,14 @@
         <package name="curl"/>
         <package name="dbus"/>
         <package name="default-editor"/>
+        <package name="diffutils"/>
         <package name="file"/>
+        <package name="findutils"/>
+        <package name="gawk"/>
         <package name="glibc-langpack-en"/>
         <package name="gnupg2"/>
+        <package name="grep"/>
+        <package name="gzip"/>
         <package name="iproute"/>
         <package name="iputils"/>
         <package name="less"/>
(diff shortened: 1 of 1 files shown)


mfrw (commented):
@jperrin for awareness

tobiasb-ms:
@mfrw -- Did you look at whether these are already in the base image? If not, do we want them?

mfrw:
> @mfrw -- Did you look at whether these are already in the base image? If not, do we want them?

Yes e.g the `patch` package & `awk` package was not present in the base image. The others are pulled in transiently, but I pinned them so as not to have a bad devex.

The reason to add them is two fold:
- When we run `code .` from the WSL instance, it threw a warning w.r.t awk not present.
- We had a meeting with Jim where we discussed that the base image should atleast have basic tools for scripting etc.

mfrw (commented):
@tobiasb-ms - Ping :)

jperrin:
This looks reasonable to me. I'm good with adding these packages to the WSL base.

christopherco (approved):

PR #17796, opened 2026-06-24
fix: moving commits dropped by fedora upstream history rewrite

Tonisal-byte:
Creating this PR to move components with pinned commits lost to upstream history rewrites. Keeping the old pin would result in render failures. Will have to address this in azldev to fully mitigate the issue. [AB#22036](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22036)

These new lock files were generated by using the following azure linux dev tools command:
```bash
azldev component update --force-recalculate -p ant-contrib -p antlr -p apache-commons-cli -p apache-commons-codec -p apache-commons-exec -p apache-commons-io -p apache-commons-parent -p beust-jcommander -p bsf -p icu4j -p jaxen -p jboss-parent -p jdom -p jline -p jsch -p log4j -p maven-antrun-plugin -p maven-archiver -p maven-assembly-plugin -p maven-shade-plugin -p mojo-parent -p msv -p plexus-interactivity
```
To regenerate the spec files, we simply re-rendered them using 
```bash 
azldev component render -p <component-name>
```


reubeno (commented):
Naive question: why did some specs roll to a new release, and others not?

dmcilvaney:
> Naive question: why did some specs roll to a new release, and others not?

My guess would be the ones that did update are using the static package mechanics, which is a bit more robust since it doesn't rely on upstream distgits.

tobiasb-ms:
question(non-blocking): How did you actually do this? What tool or whatever did you run? Ideally those details would go in the commit message, but at least in the PR description.

Tonisal-byte:
> > Naive question: why did some specs roll to a new release, and others not?
> 
> My guess would be the ones that did update are using the static package mechanics, which is a bit more robust since it doesn't rely on upstream distgits.

Yeah, I was just looking into this and Daniel got it exactly right

Tonisal-byte:
> question(non-blocking): How did you actually do this? What tool or whatever did you run? Ideally those details would go in the commit message, but at least in the PR description.

My bad for not including it initially, the `azldev` command is now included in the description 😅

reubeno (approved):

christopherco:
I don't exactly follow the sequence of events that transpired to require us to need this PR. Please link the bug item in the PR description so I can read up on the issue.
If it is an ADO bug item, you can add `AB#<bug number>` into the PR description and the Azure Boards Bot will eventually come by and auto-link it.

PR #17798, opened 2026-06-24
feat: add curated telegraf 1.38.2 build for Azure Linux 4.0 (#20399)

WithEnoughCoffee:
## Summary
Telegraf shipped in Azure Linux 3.0 but is missing from 4.0. This restores it as a general-purpose, plugin-driven agent for collecting, processing, aggregating, and writing metrics. Resolves #20399.

## Why a curated build
Built upstream-default, telegraf links **~400 plugins** and the full transitive dependency tree — a large CVE surface, vendor footprint, and binary for a distro we maintain. Instead we compile a curated (**"Balanced", 108 build tags**) set: 63 inputs, 15 outputs, 7 processors, 4 aggregators, 12 parsers, 7 serializers. The rest are absent from the binary at build time.
- **General-purpose** — common system, network, database, container, and cloud inputs/outputs included, useful out of the box.
- **Full first-party Azure + GitHub** — `azure_monitor` (in/out), `azure_storage_queue`, `eventhub_consumer`, `azure_data_explorer`, and the `github` input are all included, since AzL is an Azure/Microsoft + GitHub product and these should work by default.
- **Smaller attack/maintenance surface** — fewer compiled plugins → fewer linked deps, fewer CVEs, smaller binary. Concretely, `go list -deps ./cmd/telegraf` links **1,877 packages vs 3,386** for the full build (~45% dropped), across **344 distinct third-party modules vs 592** (~248 fewer). Note: this reduces the *linked/runtime-reachable* surface only — the full vendor tree is still shipped (Fedora requires it), so the source-level CVE-scan footprint is unchanged.
- **Reviewable & adjustable in one place** — the entire plugin policy is a single spec macro (`%global buildtags`). Adding (or removing) a plugin is a **one-line change** — append its tag to the macro, e.g.:
  ```diff
       inputs.cpu inputs.disk inputs.diskio inputs.mem inputs.net inputs.netstat \
  +    inputs.redis \
  ```
  then re-render and rebuild. No `%build`/`%install`/`%files` changes are needed, so curation stays easy to audit and evolve as requirements change.

> **Note on upstream:** telegraf is not packaged by Fedora; the reference is InfluxData's own RPM, which ships the **full** plugin set. Our curation is a deliberate deviation produced via upstream's supported `custom` build tag (the same mechanism InfluxData exposes to end users for slimmed builds). The full vendor tree is retained for reproducibility and easy plugin additions.

## Packaging (Fedora Go guidelines)
Uses the `go2rpm --profile vendor` scaffold as the baseline (Go Vendor Tools, vendored deps, `%gobuild` with `GO_BUILDTAGS`/`GO_LDFLAGS`), so it can be upstreamed to Fedora and matches the vendored-Go pattern AzL already uses (`rootlesskit`, `git-lfs`). Divergences are marked `# AzL:`. The full vendor tree is retained (Fedora requires it); curation only affects what is compiled. The cumulative SPDX `License` tag is computed with `go_vendor_license` and enforced by `%go_vendor_license_check`; `bundled(golang(...))` provides are auto-generated.

## systemd unit
The upstream systemd unit is shipped **unmodified** (runs as `User=telegraf`). We intentionally add **no sandboxing drop-in**: telegraf is a whole-system monitoring agent, and the curated inputs include hardware collectors that shell out via `sudo -n` (smart, smartctl, ipmi_sensor) or need `CAP_NET_RAW` (ping) — `NoNewPrivileges`/`Protect*` would break them. This matches upstream InfluxData and AzL 3.0. (An earlier revision shipped an openSUSE-derived `50-hardening.conf`; it was dropped after review because it diverged from upstream and conflicted with the curated hardware inputs. See the PR discussion for the full rationale.)

## Contents
- `telegraf.spec` — `%gometa`, curated `%global buildtags`, Go Vendor Tools license macros, sysusers (no `userdel` on uninstall), upstream systemd unit (unmodified), logrotate, generated default config, state dir, `%check` (license check + binary smoke test).
- `go-vendor-tools.toml` — askalono detector + manual license entries.
- `telegraf.comp.toml` — upstream source plus the full vendor tarball.
- `telegraf.sysusers`, `telegraf.default`, `generate_source_tarball.sh`, `locks/telegraf.lock`.

## Verification
Full mock build passes every phase including `%check`. Confirmed in mock:
- Binary reports `Telegraf 1.38.2` (branch stamped `azurelinux`); functional collection works (`cpu` input loads and emits).
- Curated plugins present (`azure_monitor` in/out, `azure_data_explorer`, `github`, `eventhub_consumer`, docker, prometheus, snmp, …); non-curated absent (cloudwatch, sqlserver, nats, clickhouse).
- File modes: `telegraf.conf` `0644 root:root` (world-readable, as on Fedora); state dir `/var/lib/telegraf` `0770 root:telegraf` (matching upstream InfluxData `post-install.sh`).
- Install/erase lifecycle: sysusers creates the `telegraf` user with home `/etc/telegraf` (matching upstream InfluxData `useradd -r -M -d /etc/telegraf`; config is read via the unit's explicit `-config` flag, independent of `$HOME`); the unit installs; on erase the user is intentionally retained.
- `systemd-analyze verify` accepts the unit; debuginfo is split into its own subpackage.

## Why 1.38.2 (not 1.39.0)
telegraf 1.39.0's `go.mod` requires Go 1.26; AzL 4.0 currently ships Go 1.25.8, and 1.38.2 is the latest release that builds on it. We can bump to 1.39.0 once AzL golang reaches ≥ 1.26 (which also drops the logzio azure-monitor dependency).

## Known follow-up
- The reproducible vendor tarball (`generate_source_tarball.sh`, SHA512 `1108fe48086a7051c5cb89935c6de1c675c3ea8212a979d147ad0c03aef327c6234fa9eee292e4f9594ba9ec2cb757fc9eff46630aea43551bca3d948b30b27f`) must be uploaded to the lookaside store before CI source checks and package builds can fetch it; the `comp.toml` source URI already points at its final published path.


tobiasb-ms:
issue(blocking): While having a list of required follow-ups is good, make sure to remove it before we merge.

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+# Generates a reproducible vendored Go-module tarball for telegraf.
+#
+# The custom (minimal-plugin) build still vendors the *full* dependency tree so
+# the build is hermetic and offline; only the selected plugins are compiled in
+# (see %%{plugin_tags} in telegraf.spec). Pruning vendor/ is intentionally NOT
+# done — it would break reproducibility and `go mod verify`.
```
tobiasb-ms, line 17: question(non-blocking): Just curious -- how would pruning `vendor/` break reproducibility. It would still be the same code right? Or what I am I missing?

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,101 @@
...
+
+cd "$NAME_VER"
+echo "Getting the vendored modules."
+go mod vendor
+
+mkdir -p "$OUT_FOLDER"
+
+echo "Tar vendored modules (deterministic flags for reproducibility)."
+tar  --sort=name \
+     --mtime="2021-04-26 00:00Z" \
```
tobiasb-ms: question(blocking): Where did this date come from? It seems pretty arbitrary.

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,101 @@
...
+            shift 2
+            ;;
+        -*)
+            echo "Error: unsupported flag $1." >&2
+            exit 1
+            ;;
+    esac
+done
+
+echo "--srcTarball   -> $SRC_TARBALL"
```
tobiasb-ms: suggestion(non-blocking): Here and throughout the file, best practice in bash is to use brace expansion (so `${SRC_TARBALL}`). Not critical at all, but if you end up making other changes you can consider this.

base/comps/telegraf/telegraf.comp.toml
```diff
@@ -0,0 +1,26 @@
+# Telegraf — agent for collecting, processing, aggregating, and writing metrics.
+# Shipped in AzL 3.0 but missing from 4.0; this restores it as a general-purpose metrics
+# agent. Not available in Fedora; maintained as a local spec with vendored Go dependencies.
+# Curated ("Balanced") custom build keeps the binary's dependency/CVE surface small.
+# Packaged per Fedora Go guidelines for upstreaming.
```
tobiasb-ms: issue(blocking): Add a link to the fedora go guidelines.

base/comps/telegraf/telegraf-hardening.conf
```diff
@@ -0,0 +1,21 @@
+# Azure Linux systemd sandboxing for the telegraf metrics agent.
```
tobiasb-ms: question(non-blocking): Did we have this in AZL3? If not, why do we need it here?

base/comps/telegraf/telegraf.comp.toml
```diff
@@ -0,0 +1,26 @@
...
+[components.telegraf]
+spec = { type = "local", path = "telegraf.spec" }
+# Local Go package: changelog/Release are hand-managed, not derived by rpmautospec.
+release = { calculation = "manual" }
```
tobiasb-ms: issue(blocking): Why wouldn't we use autorelease here?

base/comps/telegraf/telegraf.spec
```diff
@@ -0,0 +1,207 @@
...
+
+# Binary
+install -Dpm 0755 %{gobuilddir}/bin/%{name} %{buildroot}%{_bindir}/%{name}
+
+# AzL: --- runtime packaging (omitted by the go2rpm scaffold) -------------------------
+# Systemd unit (shipped by upstream; runs as User=telegraf)
+install -Dpm 0644 scripts/%{name}.service %{buildroot}%{_unitdir}/%{name}.service
+
+# AzL: systemd hardening drop-in (sandboxing; PrivateDevices omitted for hardware inputs).
+install -Dpm 0644 %{SOURCE4} %{buildroot}%{_unitdir}/%{name}.service.d/50-hardening.conf
```
tobiasb-ms: question(blocking): I don't see anything like this in the AZL3 spec. Why do we need it here?

tobiasb-ms (changes requested):
question(blocking): I left a couple specific comments about this, but what are the practical differences between this and how we packaged it for AZL3? I know we've changed to use a fedora-blessed way of packaging, and that seems righteous. And of course we bumped the version. But I think there are more semantic differences here and we need to know what they are and why we're making them before taking this change.

tobiasb-ms:
issue(blocking): This isn't in `components-publish-channels.toml`, so it will be published to the sdk repo, not the base repo.

binujp:
rpm-layout, build (blocking): Can you please confirm this builds on koji? I also saw Suse was the reference for defaults. We should lean towards fedora/centos/redhat. There homedir is /etc/telegraf.

dnf repoquery -l telegraf | grep -v build-id
Updating and loading repositories:
Repositories loaded.
/etc/logrotate.d/telegraf
/etc/telegraf/telegraf.conf
/etc/telegraf/telegraf.conf.sample
/etc/telegraf/telegraf.d
/etc/telegraf/telegraf.d/.ignore
/usr/bin/telegraf
/usr/lib/telegraf/scripts/init.sh
/usr/lib/telegraf/scripts/telegraf.service
/var/log/telegraf

WithEnoughCoffee:
Good question — here's the full inventory of what changed from AZL3 (`1.31.0-10`) and why. I've split it into **(A) behavioral/semantic changes** that genuinely affect what ships and how it runs (the ones worth scrutinizing), and **(B) mechanical/compliance changes** from adopting the Fedora Go scaffold.

### A. Semantic / behavioral changes (the ones to sign off on)

| # | AZL3 (1.31.0) | AZL4 (1.38.2) | Why |
|---|---|---|---|
| 1 | **All ~415 plugins** compiled in | **Curated "Balanced" ~108 plugins** via upstream custom-builder build tags | Headline change. Binary ~250 MB → ~105 MB; cuts the applicable CVE surface from the vendored Go tree. Risk: any plugin not on the list is *absent*. Set is one auditable `%global buildtags` macro and is trivially expandable. Confirming the exact set with the MetricsExtension team is the last open item. |
| 2 | Default `telegraf.conf` from full catalog | Default config generated from the **curated binary** | Config reflects only compiled-in plugins (no dangling references). |
| 3 | `%postun` runs `userdel`/`groupdel` — **user removed on uninstall** | **User is never deleted** (declarative sysusers.d; no userdel) | Fedora guideline: removing a system user can orphan files under a later-reused UID. Leaves a `telegraf` user behind after uninstall. |
| 4 | `telegraf.conf` world-readable (0755, root:root) | `telegraf.conf` **0640 telegraf:telegraf** + `telegraf.d/` drop-in dir | Tighter perms; ownership set via `%attr` instead of a `%post chown -R`. |
| 5 | No sandboxing | **systemd hardening drop-in** (`50-hardening.conf`) | Sandboxes the agent (PrivateDevices intentionally omitted so hardware inputs still work). Could restrict plugins needing extra access. |
| 6 | Plain `go build` → effectively static, no debuginfo | `%gobuild` → **PIE + hardened flags + debuginfo/debugsource subpackages** | Enterprise/Fedora norm; FIPS-consistent. Dynamic binary instead of AZL3's static one. |
| 7 | New: `/etc/default/telegraf` EnvironmentFile | — | Lets operators pass `TELEGRAF_OPTS` without editing the unit. |

### B. Mechanical / compliance changes (no behavioral impact)

- **12 inline CVE patches → 0.** The 1.31.0→1.38.2 bump supersedes all of `Patch0–11` and clears ~45 backported CVEs; ongoing CVE handling moves to vendor-tree bumps.
- **License accuracy.** AZL3 declared `License: MIT` (telegraf's own license only). AZL4 ships the real cumulative SPDX expression computed over the *entire* vendor tree by go-vendor-tools (askalono) and **enforced in `%check`**.
- **Packaging model.** AZL3 was a hand-written local spec; AZL4 is rebased on the canonical `go2rpm --profile vendor` scaffold (`%gometa -L -f`, go-vendor-tools, full vendor archive as `.tar.bz2`), with every AzL deviation marked `# AzL:` so it stays auditable and upstreamable. Telegraf isn't in Fedora, so it remains a local component.
- **Dependency wiring.** Explicit `shadow-utils`/`systemd` Requires replaced by `%{?systemd_requires}` / `%{?sysusers_requires_compat}`; `logrotate`/`procps-ng` unchanged.

Net: the only changes that affect *runtime behavior* are the curated plugin set (#1–2), the no-user-deletion policy (#3), tighter config perms (#4), and the sandboxing drop-in (#5). Everything else is version/toolchain/compliance hygiene. Happy to expand the plugin set or relax any of these if they conflict with a known consumer.

WithEnoughCoffee:
> issue(blocking): While having a list of required follow-ups is good, make sure to remove it before we merge.

Agreed good call out. I will keep that in mind.

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: The thing is  Go's vendoring is all-or-nothing at the module level — build tags exclude packages from compilation but don't affect the module graph or what `go mod vendor` pulls in. Even if a plugin is disabled  │ via tags, its dependencies still appear in go.mod's require list, so they're part of the build list and must be in vendor/. Pruning them would require regenerating modules.txt, which changes the SRPM's cumulative SPDX  │ license metadata and bundled golang declarations. That's why we keep the full vendor tree despite the binary not using all of it.

 Hand-deleting directories from  vendor/  (without also editing  go.mod / modules.txt ) makes the tree inconsistent, and the build fails immediately with  inconsistent vendoring  /  go mod verify  errors. The  │     custom  build tags exclude plugins from compilation, but those modules are still  require d in  go.mod , so they stay in the build list.
 
Also it makes it easier to add plugins if we find out customers need plugins we don't have in the curated list. 

Pruning "properly" means editing the module graph, which can change the binary. To drop them cleanly you'd remove the  require s from  go.mod  and regenerate  modules.txt . That changes the module graph, and under MVS  │    that can shift the selected version of a shared transitive dependency still used by a compiled-in plugin — which would actually change the output. So the only safe-and-identical option is "vendor everything, compile a subset."

Auditability. We compute the cumulative SPDX  License:  and the  bundled(golang(...))  provides over the full vendor tree. A pruned tree would make both tag-combination-dependent and fragile to maintain.

Fedora policy requires vendoring the complete dependency set anyway.

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: Good catch.  it's arbitrary. The only requirement for a reproducible tarball is that the mtime is a fixed constant, so the literal value carries no meaning — which is exactly why it reads as a magic number. I'll switch it to the convention the rest of the repo already uses for vendor/bundle tarballs ( grafana ,  sleef ,  rust-zip ,  python-pdfminer , …): honor  SOURCE_DATE_EPOCH  when the build provides it, and fall back to a documented fixed epoch instead of an unexplained date. Updated in the next push.

base/comps/telegraf/telegraf-hardening.conf
WithEnoughCoffee: No, AZL3 didn't have this — it installed upstream's  telegraf.service  as-is, with no sandboxing. This is a net-new addition for AZL4.

 Rationale: telegraf is a long-lived daemon running as an unprivileged user ( User=telegraf ) whose job is only to read host metrics. That's an ideal candidate for systemd sandboxing as defense-in-depth — if a plugin or a  vendored dependency is ever compromised, these restrictions limit blast radius (read-only system, no access to  /home , kernel tunables/modules/logs protected, no realtime, etc.). The directives are stock systemd hardening  adapted from openSUSE's hardening effort.
 
 It's shipped as a drop-in ( 50-hardening.conf ) rather than a forked unit, so it overlays upstream without diverging the unit file — easy to audit and tweak. It's also deliberately conservative so it doesn't break  │ functionality:
•  PrivateDevices  is omitted because the curated build includes hardware inputs ( smart ,  smartctl ,  ipmi_sensor ,  infiniband ) that need  /dev .
 •  HOME=/var/lib/telegraf  is set because some plugin SDKs (e.g. the Azure SDK credential cache) write under  $HOME .

it's a security improvement that aligns with AzL's enterprise hardening posture.

base/comps/telegraf/telegraf.comp.toml
WithEnoughCoffee: A few reasons, and they line up with how AzL already handles this. The spec doesn't actually use  %autorelease  — it has a static  Release: 1%{?dist}  and a literal  %changelog . The  autorelease  calculation mode is specifically for specs that contain the  %autorelease  macro (it forces azldev to preserve and expand it), so pointing it at a static-release spec would be the wrong mode.
 
 Beyond that,  autorelease  isn't really the AzL convention. Across  base/comps  only 2 of ~173 components use it ( samba  and  gvisor-tap-vsock ), and no local spec uses it at all. AzL seems to renders concrete, checked-in specs under  specs/  with explicit Release and changelog values for reproducible, auditable, offline builds — rpmautospec's git-derived  %autorelease %autochangelog  is really an upstream-Fedora dist-git mechanism and isn't  how AzL builds locally. Setting  manual  also mirrors the existing precedent: the other AzL-originating local package,  azurelinux-release , uses the exact same setup ( type = "local" ,  release = { calculation = "manual"  │ } , a static  Release: , and a hand-written  %changelog ), so I followed that house pattern to keep the curated changelog and release under explicit human control.

Eventually when this package is purposed  to Fedora dist-git, the Fedora-native choice is  %autorelease / %autochangelog  (go2rpm emits them), and we can adopt those at submission time — inside the AzL  │ monorepo we just match AzL's convention. And if you'd rather have azldev auto-bump the integer on each component-touching commit,  auto  (the default) would probably be what were looking for vs   autorelease .

Happy to switch to  that, though  manual  matches  azurelinux-release  and keeps the hand-written changelog authoritative.

base/comps/telegraf/telegraf.spec
WithEnoughCoffee: See above hardening comment

base/comps/telegraf/generate_source_tarball.sh
tobiasb-ms: These are great arguments for why we should not prune. But as I said I'm curious as to why pruning actually changes *reproducibility* of the vendored tarball. Yes, it would clearly change it but once it's changed, wouldn't the result be just as reproducible?

base/comps/telegraf/generate_source_tarball.sh
tobiasb-ms: Will `SOURCE_DATE_EPOCH` change across run or is that consistent? I agree with having something that is consistent and long-lasting -- doesn't need to change to `SOURCE_DATE_EPOCH` or anything. Just wanted to make to make sure the date wasn't something actually special.

I'm changing this to non-blocking and you can decide the right way.

base/comps/telegraf/telegraf.comp.toml
tobiasb-ms: The argument that the spec doesn't use `%autorelease` is circular. That's what I'm asking.

2 of 173 components is misleading -- aren't the vast majority of those overlays on a fedora thing? A quick search I just did four four actual spec files in `base/comps`, two of which currently use `%autorelease`.

@reubeno may have some opinions about this.

base/comps/telegraf/telegraf-hardening.conf
tobiasb-ms: Possible that I'm being paranoid here but this feels risky. We're changing the thing we deliver to a customer.

I agree in principal about using stricture security settings though.

I'll change this to non-blocking but let's leave the conversation open for others (maybe @reubeno or @ddstreet) to weigh in.

base/comps/telegraf/telegraf.comp.toml
reubeno: For any new specs we author, we should use %autorelease.

base/comps/telegraf/generate_source_tarball.sh
tobiasb-ms: I thought about this a bit over the weekend I was wondering -- what actually sets `SOURCE_DATE_EPOCH`? https://reproducible-builds.org/docs/source-date-epoch/ implies that shells set it but, for example, on my fedora machine it isn't somehow automatically set (tested via script).

What environment is this expected to be run on? I *think* it's just a random dev environment. Do we have evidence that `SOUCE_DATE_EPOCH` exists in a standard dev setup we use?

tobiasb-ms: (resetting this to blocking until we understand this better)

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: You're completely right, and I was being sloppy with the word "reproducible." A pruned tarball is not less reproducible in the deterministic sense: same inputs → same bytes, every time. Once you've defined the pruning, it reproduces just fine. Good catch.


What I should have said is that pruning trades canonical reproducibility for bespoke reproducibility:

• Today: the regeneration recipe is a single, well-known, auditable command —  go mod vendor  + deterministic  tar  flags. Anyone (including a future maintainer or a Fedora reviewer) can re-run it and diff the SHA512. The trusted input is just  go.mod / go.sum .
• With pruning: the recipe becomes  go mod vendor  + our own pruning step. That step is now part of the trusted regeneration path and has to be deterministic, version-controlled, and maintained by us. It's still reproducible — but only relative to our script, not relative to a standard Go command. And its output is a function of the  %{buildtags}  selection, so the canonical hash now changes whenever the curated plugin list changes.

There's also a practical maintenance win: keeping the full tree means adding a plugin is a one-line edit. The vendor tarball already ships every plugin's dependencies, so if a customer needs something outside our curated "Balanced" set, we just append its tag to  %global buildtags  (e.g.  inputs.nginx_plus ) and rebuild — no re-vendoring, no new tarball, no SHA512 re-pin, no network at build time. With a pruned tree the opposite happens: the moment you enable a plugin whose transitive deps were stripped, the build breaks and you're forced to regenerate and re-pin the whole vendor archive. So the full tree keeps plugin curation cheap and self-contained. I think that's really important when it comes to having a curated list of plugs but also making it a two way door that if we or a customer needs a new plug in its a one line fix. This allows us to cut down the foot print and CVEs but also provide what our customers need.

base/comps/telegraf/telegraf.comp.toml
WithEnoughCoffee: Added

base/comps/telegraf/telegraf-hardening.conf
WithEnoughCoffee: That's a fair concern, and not paranoid — you're right that this changes the artifact we hand the customer, and that deserves scrutiny. Happy to keep the thread open for @reubeno / @ddstreet.

A few things that I think lower the risk, for whoever weighs in:

• It's additive and conservative, not a behavior rewrite. The drop-in only restricts filesystem/kernel access for a daemon that exclusively reads metrics — it doesn't change telegraf's config, plugins, or output. The two places where stock hardening would have broken functionality are explicitly relaxed ( PrivateDevices  omitted for the hardware inputs;  HOME  set for the Azure SDK credential cache).
• It's a drop-in, not a forked unit. Reverting or softening is a one-file change ( 50-hardening.conf ) with zero impact on upstream's  telegraf.service  — so if it ever causes friction in the field, the blast radius of backing it out is also tiny.
• It's auditable. Every directive is stock systemd, sourced from openSUSE's hardening effort — no bespoke seccomp filters or anything exotic that could silently break a plugin.

That said, I fully agree the call shouldn't be unilateral.

base/comps/telegraf/telegraf.comp.toml
WithEnoughCoffee: I'll convert telegraf to  %autorelease  /  %autochangelog .

base/comps/telegraf/generate_source_tarball.sh
tobiasb-ms: Yes, I agree that not pruning makes sense.

Given that it doesn't affect reproducibility, we should remove that from the comment.

WithEnoughCoffee:
> rpm-layout, build (blocking): Can you please confirm this builds on koji? I also saw Suse was the reference for defaults. We should lean towards fedora/centos/redhat. There homedir is /etc/telegraf.
> 
> dnf repoquery -l telegraf | grep -v build-id Updating and loading repositories: Repositories loaded. /etc/logrotate.d/telegraf /etc/telegraf/telegraf.conf /etc/telegraf/telegraf.conf.sample /etc/telegraf/telegraf.d /etc/telegraf/telegraf.d/.ignore /usr/bin/telegraf /usr/lib/telegraf/scripts/init.sh /usr/lib/telegraf/scripts/telegraf.service /var/log/telegraf


Home directory — you're right, Switched the  telegraf  user's home from the SUSE-style  /var/lib/telegraf  to  /etc/telegraf  to match upstream InfluxData and Fedora/RHEL (their  useradd -d /etc/telegraf ). The rest of the layout already follows that convention ( /etc/telegraf  config,  /var/log/telegraf  logs,  /usr/bin/telegraf ).

One nuance worth documenting (and I've left a comment in the sysusers file):  /etc/telegraf  is root-owned and, under our hardening ( ProtectSystem=full ), read-only at runtime — so it's not a writable home. Some plugin SDKs (e.g. the Azure SDK credential cache) write under  $HOME , so the service drop-in sets  Environment=HOME=/var/lib/telegraf  (writable;  /var  stays writable under  ProtectSystem=full ). Net result for the supported path — running as the systemd service — config is read from  /etc/telegraf , SDK caches land in  /var/lib/telegraf , everything works.

The only caveat: if telegraf is run outside systemd (manual  sudo -u telegraf  debugging), that override isn't applied,  $HOME  falls back to  /etc/telegraf , and writes under  $HOME  would fail with EACCES. This is identical to upstream's design (they ship the same root-owned  /etc/telegraf  home); workaround is  HOME=/var/lib/telegraf  for manual runs. Flagged in a sysusers comment so it doesn't surprise a future maintainer.

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: You're correct, and this was a genuine defect in the reviewed version.  SOURCE_DATE_EPOCH  is not set by default in any standard shell, Fedora, or our dev containers — only a build system like rpmbuild injects it. The old  --mtime="@${SOURCE_DATE_EPOCH:-0}"  therefore made the output environment-dependent: an exported value produced one hash, a bare shell fell back to the 1970 epoch and produced another. That non-determinism was the source of the arbitrary date.

The fix removes the environment dependency. The epoch is now derived from the upstream source itself — the newest file mtime in the unpacked, SHA512-pinned source tarball — computed before  go mod vendor :

if [ -z "${SOURCE_DATE_EPOCH:-}" ]; then
    SOURCE_DATE_EPOCH="$(find . -type f -printf '%T@\n' | cut -d. -f1 | sort -n | tail -1)"
fi

Since the source tarball is hash-pinned, this resolves to a fixed constant (epoch  1774894864 ) for everyone, independent of shell or CI. An externally provided  SOURCE_DATE_EPOCH  is honored as an optional override but is no longer required.

Verified: regenerated from scratch on a bare shell (no  SOURCE_DATE_EPOCH  set) yields SHA512  14b38940… , byte-identical across runs and matching the pin, and the RPM builds from it. The  2021-04-26  literal is removed.

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: fixed.

base/comps/telegraf/generate_source_tarball.sh
tobiasb-ms: Just confirming: You decided to leave the errant comment about reproducibility in?

WithEnoughCoffee: Good catch I'll remove the comment.

WithEnoughCoffee:
> issue(blocking): This isn't in `components-publish-channels.toml`, so it will be published to the sdk repo, not the base repo.

fixed

tobiasb-ms:
> > rpm-layout, build (blocking): Can you please confirm this builds on koji? I also saw Suse was the reference for defaults. We should lean towards fedora/centos/redhat. There homedir is /etc/telegraf.
> > dnf repoquery -l telegraf | grep -v build-id Updating and loading repositories: Repositories loaded. /etc/logrotate.d/telegraf /etc/telegraf/telegraf.conf /etc/telegraf/telegraf.conf.sample /etc/telegraf/telegraf.d /etc/telegraf/telegraf.d/.ignore /usr/bin/telegraf /usr/lib/telegraf/scripts/init.sh /usr/lib/telegraf/scripts/telegraf.service /var/log/telegraf
> 
> Home directory — you're right, Switched the  telegraf  user's home from the SUSE-style  /var/lib/telegraf  to  /etc/telegraf  to match upstream InfluxData and Fedora/RHEL (their  useradd -d /etc/telegraf ). The rest of the layout already follows that convention ( /etc/telegraf  config,  /var/log/telegraf  logs,  /usr/bin/telegraf ).
> 
> One nuance worth documenting (and I've left a comment in the sysusers file):  /etc/telegraf  is root-owned and, under our hardening ( ProtectSystem=full ), read-only at runtime — so it's not a writable home. Some plugin SDKs (e.g. the Azure SDK credential cache) write under  $HOME , so the service drop-in sets  Environment=HOME=/var/lib/telegraf  (writable;  /var  stays writable under  ProtectSystem=full ). Net result for the supported path — running as the systemd service — config is read from  /etc/telegraf , SDK caches land in  /var/lib/telegraf , everything works.
> 
> The only caveat: if telegraf is run outside systemd (manual  sudo -u telegraf  debugging), that override isn't applied,  $HOME  falls back to  /etc/telegraf , and writes under  $HOME  would fail with EACCES. This is identical to upstream's design (they ship the same root-owned  /etc/telegraf  home); workaround is  HOME=/var/lib/telegraf  for manual runs. Flagged in a sysusers comment so it doesn't surprise a future maintainer.

This seems like a large caveat that ends up creating two sources of truth. In the worst case they could end up conflicting.

From what you say above, it seems like that could be avoided by removing some of the new hardening, right?

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,125 @@
...
+
+# Reproducibility: SOURCE_DATE_EPOCH sets the (clamped) mtime stamped on every entry in the
+# vendor archive and is the ONLY non-content variable in the output (file contents, ordering,
+# owner/group and modes are already normalized below). To make the archive a pure function of
+# the hash-pinned source tarball — byte-identical for anyone, independent of their shell
+# environment — derive it from the newest mtime in the unpacked upstream source instead of the
+# ambient value. The previous `${SOURCE_DATE_EPOCH:-0}` fallback made the hash environment
+# dependent: whoever already had SOURCE_DATE_EPOCH exported got one hash, a bare shell got the
+# 1970 hash. We compute this BEFORE `go mod vendor` so freshly written vendor/ files (stamped
+# at vendoring time) don't influence it. An explicit override is still honored for rpmbuild.
```
tobiasb-ms: question(blocking): This will never be called during `rpmbuild` will it?

Blocking to make sure I understand.

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,125 @@
...
+if [ -z "${SOURCE_DATE_EPOCH:-}" ]; then
+    SOURCE_DATE_EPOCH="$(find . -type f -printf '%T@\n' | cut -d. -f1 | sort -n | tail -1)"
+fi
+export SOURCE_DATE_EPOCH
```
tobiasb-ms: issue(non-blocking): I don't see what downstream script/program uses `SOUCE_DATE_EXPOCH`. Why export it?

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,125 @@
...
+
+echo "Getting the vendored modules."
+go mod vendor
+
+
+echo "Tar vendored modules (deterministic flags for reproducibility)."
+# --clamp-mtime caps anything newer than SOURCE_DATE_EPOCH (derived above); combined with
+# --sort=name, fixed owner/group and stripped pax atime/ctime, the output is byte-deterministic.
+tar  --sort=name \
+     --mtime="@${SOURCE_DATE_EPOCH}" --clamp-mtime \
```
tobiasb-ms: suggestion(non-blocking): Personally I think `SOURCE_DATE_EPOCH` has become more trouble than it's worth. The calculation above is kind of weird and I feel like we get very little value from it.

When I asked earlier about the arbitrariness of the date, I really just meant "where did it come from?" And maybe "should we pick a less arbitrary one?"

I would seriously consider either choosing a date (perhaps `@0` for the unix epoch) and documenting that decision.

That said, I'm opening this as non-blocking.

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,125 @@
...
+fi
+
+SRC_TARBALL="$(realpath "${SRC_TARBALL}")"
+# Create the output folder up front so realpath can resolve the (not-yet-existing)
+# vendor archive path below, and so a nested --outFolder works on a clean tree.
+mkdir -p "${OUT_FOLDER}"
+OUT_FOLDER="$(realpath "${OUT_FOLDER}")"
+
+echo "Creating a tempdir."
+tmpdir=$(mktemp -d)
```
tobiasb-ms: nitpick(non-blocking): Other variables in this file seem to be capitalized, so consider changing this to `TMPDIR`.

base/comps/telegraf/telegraf.spec
```diff
@@ -0,0 +1,202 @@
...
+
+# Declarative system user
+install -Dpm 0644 %{SOURCE3} %{buildroot}%{_sysusersdir}/%{name}.conf
+
+# Logrotate fragment (shipped by upstream)
+install -Dpm 0644 etc/logrotate.d/%{name} %{buildroot}%{_sysconfdir}/logrotate.d/%{name}
+
+# Optional environment file referenced by the unit (EnvironmentFile=-...).
+install -d -m 0755 %{buildroot}%{_sysconfdir}/default
+cat > %{buildroot}%{_sysconfdir}/default/%{name} <<'EOF'
```
tobiasb-ms: suggestion(non-blocking): The other drop-ins you have as a source-file. Did you add this or did the `go2rpm` do it? If you added this, I'd prefer it be consistent with the other files.

base/comps/telegraf/telegraf.spec
```diff
@@ -0,0 +1,202 @@
...
+#TELEGRAF_OPTS=
+EOF
+
+# Config tree. Generate a default config from THIS binary so it reflects the curated
+# (compiled-in) plugin set rather than the full upstream catalog. Note: this executes the
+# freshly-built target binary, so it assumes a native (non-cross/emulated) build, which is
+# how AzL builds each arch in mock.
+install -d -m 0755 %{buildroot}%{_sysconfdir}/%{name}/%{name}.d
+%{gobuilddir}/bin/%{name} config > %{name}.conf
+install -Dpm 0640 %{name}.conf %{buildroot}%{_sysconfdir}/%{name}/%{name}.conf
```
tobiasb-ms: issue(blocking): Most config files in `/etc` are world-readable, and on Fedora this file is world-readable. Install with `0644`.

base/comps/telegraf/telegraf.spec
```diff
@@ -0,0 +1,202 @@
...
+
+# State directory
+install -d -m 0750 %{buildroot}%{_localstatedir}/lib/%{name}
```
tobiasb-ms: issue(blocking): On fedora, this is installed as `0755`. Is there any reason we wouldn't do that?

.github/copilot-instructions.md
```diff
@@ -102,3 +102,4 @@ Run all commands from the repo root (where `azldev.toml` lives). If the terminal
...
 4. **Don't edit generated output**: Build artifacts in `base/out/` and `base/build/` are generated (configured by `output-dir`, `log-dir`, `work-dir` in `base/project.toml`) — never edit them directly. Note: `prep-sources -o <dir>` writes to a user-specified directory, separate from these project output dirs.
 5. **Mandatory testing**: Any change that affects a component's output (overlays, build config, spec edits, version bumps, new components) MUST be validated by building AND testing the resulting RPMs. A successful build alone is not sufficient — smoke-test in a mock chroot. Pure organizational changes (moving definitions between files, editing descriptions/comments) do not require rebuild. See `AGENTS.md` for the full testing protocol.
+6. **New specs use autorelease**: For any spec we author ourselves (local specs), use `%autorelease` and `%autochangelog` rather than a static `Release:` with a hand-written `%changelog`. Set the component's `release = { calculation = "autorelease" }` in its `.comp.toml` so azldev preserves and expands the macros. This is the team convention for newly authored specs even though it differs from upstream-Fedora overlay handling.
```
tobiasb-ms: suggestion(blocking): I love this idea, but it should be a separate commit (or ideally separate PR). If you want to keep it in this PR, have it as a separate commit.

tobiasb-ms (changes requested):

tobiasb-ms:
issue/question(blocking): Do we have actual data/signal that reducing the number of plugins reduces the number of CVEs/attacks? It's a reasonable hypothesis, but it does move us away from the upstream (upstreamish -- it's not actually packaged by fedora), and away from AZL3. There's risk with that deviation and we need data to demonstrate that the risk is worth it.

tobiasb-ms:
issue(blocking): The commit message isn't a conventional commit. When doing your final rebase/reset, make sure the message is something like `feat(telegraf): ...`.

WithEnoughCoffee:
> > > rpm-layout, build (blocking): Can you please confirm this builds on koji? I also saw Suse was the reference for defaults. We should lean towards fedora/centos/redhat. There homedir is /etc/telegraf.
> > > dnf repoquery -l telegraf | grep -v build-id Updating and loading repositories: Repositories loaded. /etc/logrotate.d/telegraf /etc/telegraf/telegraf.conf /etc/telegraf/telegraf.conf.sample /etc/telegraf/telegraf.d /etc/telegraf/telegraf.d/.ignore /usr/bin/telegraf /usr/lib/telegraf/scripts/init.sh /usr/lib/telegraf/scripts/telegraf.service /var/log/telegraf
> > 
> > 
> > Home directory — you're right, Switched the  telegraf  user's home from the SUSE-style  /var/lib/telegraf  to  /etc/telegraf  to match upstream InfluxData and Fedora/RHEL (their  useradd -d /etc/telegraf ). The rest of the layout already follows that convention ( /etc/telegraf  config,  /var/log/telegraf  logs,  /usr/bin/telegraf ).
> > One nuance worth documenting (and I've left a comment in the sysusers file):  /etc/telegraf  is root-owned and, under our hardening ( ProtectSystem=full ), read-only at runtime — so it's not a writable home. Some plugin SDKs (e.g. the Azure SDK credential cache) write under  $HOME , so the service drop-in sets  Environment=HOME=/var/lib/telegraf  (writable;  /var  stays writable under  ProtectSystem=full ). Net result for the supported path — running as the systemd service — config is read from  /etc/telegraf , SDK caches land in  /var/lib/telegraf , everything works.
> > The only caveat: if telegraf is run outside systemd (manual  sudo -u telegraf  debugging), that override isn't applied,  $HOME  falls back to  /etc/telegraf , and writes under  $HOME  would fail with EACCES. This is identical to upstream's design (they ship the same root-owned  /etc/telegraf  home); workaround is  HOME=/var/lib/telegraf  for manual runs. Flagged in a sysusers comment so it doesn't surprise a future maintainer.
> 
> This seems like a large caveat that ends up creating two sources of truth. In the worst case they could end up conflicting.
> 
> From what you say above, it seems like that could be avoided by removing some of the new hardening, right?

The home dir is root-owned and never writable by  telegraf  — by upstream design, not because of our hardening.
Upstream:  useradd -r -M telegraf -s /bin/false -d /etc/telegraf -g telegraf  (pre-install.sh:8). The  -M  means do not create/own a home — it just points the home field at the existing root-owned  /etc/telegraf  config dir. Our spec matches:  %dir %{_sysconfdir}/%{name}  with no  %attr  root:root 0755  (telegraf.spec:209). So  $HOME=/etc/telegraf  is unwritable by  telegraf  even on a stock InfluxData install with zero sandboxing. Removing our hardening would not fix it. Config loading is independent of  $HOME .
 ExecStart=... -config /etc/telegraf/telegraf.conf -config-directory /etc/telegraf/telegraf.d  (upstream telegraf.service:13). The home-dir field has zero effect on where config is read. The two sources of truth are because were trying to make it fit the fedora RHEL defaults. These defaults for the home directory came from Telegraf not the Suse hardening. In AZL 3.0 we took it as is this is coming up because we are trying to meet fedora standards.  I am really passionate about sending changes upstream, I think it would be better to purpose it with the influx upstream home directory and explain why. What are you're thoughts?

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: Good catch on the stale wording  the "An explicit override is still honored for rpmbuild" line was misleading. I've removed it and added a header note stating the script is never invoked during rpmbuild.

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: Good catch will fix.

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: Agreed — switched to a fixed  SOURCE_DATE_EPOCH=0  and documented the decision.

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: Fixed, made it capitalized.

base/comps/telegraf/telegraf.spec
WithEnoughCoffee: Good call — I added it (go2rpm doesn't generate env files; it's only here because upstream's unit has  EnvironmentFile=-/etc/default/telegraf ). I've moved it to a  Source5: %{name}.default  file and install it the same way as the sysusers/hardening drop-ins, so it's consistent now.

base/comps/telegraf/telegraf.spec
WithEnoughCoffee: Will fix.

tobiasb-ms:
> > > > rpm-layout, build (blocking): Can you please confirm this builds on koji? I also saw Suse was the reference for defaults. We should lean towards fedora/centos/redhat. There homedir is /etc/telegraf.
> > > > dnf repoquery -l telegraf | grep -v build-id Updating and loading repositories: Repositories loaded. /etc/logrotate.d/telegraf /etc/telegraf/telegraf.conf /etc/telegraf/telegraf.conf.sample /etc/telegraf/telegraf.d /etc/telegraf/telegraf.d/.ignore /usr/bin/telegraf /usr/lib/telegraf/scripts/init.sh /usr/lib/telegraf/scripts/telegraf.service /var/log/telegraf
> > > 
> > > 
> > > Home directory — you're right, Switched the  telegraf  user's home from the SUSE-style  /var/lib/telegraf  to  /etc/telegraf  to match upstream InfluxData and Fedora/RHEL (their  useradd -d /etc/telegraf ). The rest of the layout already follows that convention ( /etc/telegraf  config,  /var/log/telegraf  logs,  /usr/bin/telegraf ).
> > > One nuance worth documenting (and I've left a comment in the sysusers file):  /etc/telegraf  is root-owned and, under our hardening ( ProtectSystem=full ), read-only at runtime — so it's not a writable home. Some plugin SDKs (e.g. the Azure SDK credential cache) write under  $HOME , so the service drop-in sets  Environment=HOME=/var/lib/telegraf  (writable;  /var  stays writable under  ProtectSystem=full ). Net result for the supported path — running as the systemd service — config is read from  /etc/telegraf , SDK caches land in  /var/lib/telegraf , everything works.
> > > The only caveat: if telegraf is run outside systemd (manual  sudo -u telegraf  debugging), that override isn't applied,  $HOME  falls back to  /etc/telegraf , and writes under  $HOME  would fail with EACCES. This is identical to upstream's design (they ship the same root-owned  /etc/telegraf  home); workaround is  HOME=/var/lib/telegraf  for manual runs. Flagged in a sysusers comment so it doesn't surprise a future maintainer.
> > 
> > 
> > This seems like a large caveat that ends up creating two sources of truth. In the worst case they could end up conflicting.
> > From what you say above, it seems like that could be avoided by removing some of the new hardening, right?
> 
> The home dir is root-owned and never writable by  telegraf  — by upstream design, not because of our hardening. Upstream:  useradd -r -M telegraf -s /bin/false -d /etc/telegraf -g telegraf  (pre-install.sh:8). The  -M  means do not create/own a home — it just points the home field at the existing root-owned  /etc/telegraf  config dir. Our spec matches:  %dir %{_sysconfdir}/%{name}  with no  %attr  root:root 0755  (telegraf.spec:209). So  $HOME=/etc/telegraf  is unwritable by  telegraf  even on a stock InfluxData install with zero sandboxing. Removing our hardening would not fix it. Config loading is independent of  $HOME .  ExecStart=... -config /etc/telegraf/telegraf.conf -config-directory /etc/telegraf/telegraf.d  (upstream telegraf.service:13). The home-dir field has zero effect on where config is read. The two sources of truth are because were trying to make it fit the fedora RHEL defaults. These defaults for the home directory came from Telegraf not the Suse hardening. In AZL 3.0 we took it as is this is coming up because we are trying to meet fedora standards. I am really passionate about sending changes upstream, I think it would be better to purpose it with the influx upstream home directory and explain why. What are you're thoughts?

So is this how the fedora RPM made by InfluxData works? They also use `/etc/telegraf`. Do the plug-ins you're worried about here not work on Fedora because of this?

Also, earlier you said (my emphasis):
> One nuance worth documenting (and I've left a comment in the sysusers file):  /etc/telegraf  is root-owned and, **under our hardening ( ProtectSystem=full )**, read-only at runtime — so it's not a writable home.

And in the hardening file, there's this comment:
```
# Several plugin SDKs (e.g. the Azure SDK credential cache) write under $HOME. The telegraf
# user's passwd home is /etc/telegraf (matching upstream/Fedora), but that path is read-only
# under ProtectSystem=full, so point $HOME at the writable state dir instead.
```

These seem at odds with this statement above:
> The home dir is root-owned and never writable by  telegraf  — by upstream design, not because of our hardening.


Also:
> The home-dir field has zero effect on where config is read.

Correct, it has zero effect on `telegraf` config reads. But the whole point of this is so plug-ins can write data which, presumably, they will read. I'm unfamiliar with architecture, but as you point out there is potential for a plug-in to use the `HOME` that's set up in the hardening file most of the time and then sometimes use `/etc/telegraf`, which means they wouldn't have any of the data they've written. Is this a realistic possibility or is it the case that in practice they will be run in the same context so we don't need to worry about this?

base/comps/telegraf/telegraf.spec
WithEnoughCoffee: fixed

.github/copilot-instructions.md
WithEnoughCoffee: will do.

WithEnoughCoffee:
> issue(blocking): The commit message isn't a conventional commit. When doing your final rebase/reset, make sure the message is something like `feat(telegraf): ...`.

will do

WithEnoughCoffee:
> issue/question(blocking): Do we have actual data/signal that reducing the number of plugins reduces the number of CVEs/attacks? It's a reasonable hypothesis, but it does move us away from the upstream (upstreamish -- it's not actually packaged by fedora), and away from AZL3. There's risk with that deviation and we need data to demonstrate that the risk is worth it.

InfluxData build the customer/builder specifically to support curated builds. It's an intentional and commonly supported way of selecting and curating plugins.  I do think its going to be hard to say this is going to do %30 to for sure lower our CVEs, What I do know is the tree there so plugins can be added back. attack-surface. Smaller binary, faster builds. 

This is a two way door being that  their all shipped with the tree for the plugging and if they want all the plugins at the time of accepting the package we can do that change, or add what other plugins we decide on. But just like we try not to take GUI package and packages we don't need that can be deemed more problematic I think this is really just following the standard we've set for AZL 4.0 over all and is a two way door.

WithEnoughCoffee:
> > > > > rpm-layout, build (blocking): Can you please confirm this builds on koji? I also saw Suse was the reference for defaults. We should lean towards fedora/centos/redhat. There homedir is /etc/telegraf.
> > > > > dnf repoquery -l telegraf | grep -v build-id Updating and loading repositories: Repositories loaded. /etc/logrotate.d/telegraf /etc/telegraf/telegraf.conf /etc/telegraf/telegraf.conf.sample /etc/telegraf/telegraf.d /etc/telegraf/telegraf.d/.ignore /usr/bin/telegraf /usr/lib/telegraf/scripts/init.sh /usr/lib/telegraf/scripts/telegraf.service /var/log/telegraf
> > > > 
> > > > 
> > > > Home directory — you're right, Switched the  telegraf  user's home from the SUSE-style  /var/lib/telegraf  to  /etc/telegraf  to match upstream InfluxData and Fedora/RHEL (their  useradd -d /etc/telegraf ). The rest of the layout already follows that convention ( /etc/telegraf  config,  /var/log/telegraf  logs,  /usr/bin/telegraf ).
> > > > One nuance worth documenting (and I've left a comment in the sysusers file):  /etc/telegraf  is root-owned and, under our hardening ( ProtectSystem=full ), read-only at runtime — so it's not a writable home. Some plugin SDKs (e.g. the Azure SDK credential cache) write under  $HOME , so the service drop-in sets  Environment=HOME=/var/lib/telegraf  (writable;  /var  stays writable under  ProtectSystem=full ). Net result for the supported path — running as the systemd service — config is read from  /etc/telegraf , SDK caches land in  /var/lib/telegraf , everything works.
> > > > The only caveat: if telegraf is run outside systemd (manual  sudo -u telegraf  debugging), that override isn't applied,  $HOME  falls back to  /etc/telegraf , and writes under  $HOME  would fail with EACCES. This is identical to upstream's design (they ship the same root-owned  /etc/telegraf  home); workaround is  HOME=/var/lib/telegraf  for manual runs. Flagged in a sysusers comment so it doesn't surprise a future maintainer.
> > > 
> > > 
> > > This seems like a large caveat that ends up creating two sources of truth. In the worst case they could end up conflicting.
> > > From what you say above, it seems like that could be avoided by removing some of the new hardening, right?
> > 
> > 
> > The home dir is root-owned and never writable by  telegraf  — by upstream design, not because of our hardening. Upstream:  useradd -r -M telegraf -s /bin/false -d /etc/telegraf -g telegraf  (pre-install.sh:8). The  -M  means do not create/own a home — it just points the home field at the existing root-owned  /etc/telegraf  config dir. Our spec matches:  %dir %{_sysconfdir}/%{name}  with no  %attr  root:root 0755  (telegraf.spec:209). So  $HOME=/etc/telegraf  is unwritable by  telegraf  even on a stock InfluxData install with zero sandboxing. Removing our hardening would not fix it. Config loading is independent of  $HOME .  ExecStart=... -config /etc/telegraf/telegraf.conf -config-directory /etc/telegraf/telegraf.d  (upstream telegraf.service:13). The home-dir field has zero effect on where config is read. The two sources of truth are because were trying to make it fit the fedora RHEL defaults. These defaults for the home directory came from Telegraf not the Suse hardening. In AZL 3.0 we took it as is this is coming up because we are trying to meet fedora standards. I am really passionate about sending changes upstream, I think it would be better to purpose it with the influx upstream home directory and explain why. What are you're thoughts?
> 
> So is this how the fedora RPM made by InfluxData works? They also use `/etc/telegraf`. Do the plug-ins you're worried about here not work on Fedora because of this?
> 
> Also, earlier you said (my emphasis):
> 
> > One nuance worth documenting (and I've left a comment in the sysusers file):  /etc/telegraf  is root-owned and, **under our hardening ( ProtectSystem=full )**, read-only at runtime — so it's not a writable home.
> 
> And in the hardening file, there's this comment:
> 
> ```
> # Several plugin SDKs (e.g. the Azure SDK credential cache) write under $HOME. The telegraf
> # user's passwd home is /etc/telegraf (matching upstream/Fedora), but that path is read-only
> # under ProtectSystem=full, so point $HOME at the writable state dir instead.
> ```
> 
> These seem at odds with this statement above:
> 
> > The home dir is root-owned and never writable by  telegraf  — by upstream design, not because of our hardening.
> 
> Also:
> 
> > The home-dir field has zero effect on where config is read.
> 
> Correct, it has zero effect on `telegraf` config reads. But the whole point of this is so plug-ins can write data which, presumably, they will read. I'm unfamiliar with architecture, but as you point out there is potential for a plug-in to use the `HOME` that's set up in the hardening file most of the time and then sometimes use `/etc/telegraf`, which means they wouldn't have any of the data they've written. Is this a realistic possibility or is it the case that in practice they will be run in the same context so we don't need to worry about this?

Fedora doesn't make telegraf its made by influxData we changed the home directory to match the fedora standard proposed in an earlier comment. I think we should change it back to what InfluxData does, Until we are asked to change it for upstream if that happens.

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,126 @@
...
+tar -xf "${SRC_TARBALL}"
+
+cd "${NAME_VER}"
+
+# Pin all archived timestamps to the Unix epoch (1970-01-01). We only archive vendor/, whose
+# files are freshly written by `go mod vendor`, so their mtimes carry no meaningful
+# information. Fixing the date to @0 makes the tarball's hash depend solely on file *content* —
+# fully reproducible and independent of the host clock, locale, or how the source was unpacked.
+# Not exported: nothing downstream reads it from the environment — `go mod vendor` ignores
+# SOURCE_DATE_EPOCH, and tar gets the value via the explicit --mtime expansion below.
```
tobiasb-ms: nitpick(non-blocking): I think that `SOURCE_DATE_EPOCH` (and the comments about it) don't really add value here. A one-line comment at the point where you call `tar` and just passing in `@0` would suffice.

That said, this is very much a personal preference nitpick and I'm opening it as non-blocking. You are free to consider it and not fix it if you disagree.

tobiasb-ms:
> Fedora doesn't make telegraf its made by influxData we changed the home directory to match the fedora standard proposed in an earlier comment. I think we should change it back to what InfluxData does, Until we are asked to change it for upstream if that happens.

If you install `telegraf` from InfluxData's repo (instructions: https://computingforgeeks.com/how-to-install-telegraf-on-fedora/), the user it creates has home directory `/etc/telegraf`:
```
bash-5.3# grep telegraf /etc/passwd
telegraf:x:999:999::/etc/telegraf:/bin/false
```
So using `/var/lib/telegraf` diverges from what we're currently considering the upstream, right?

We also still need answers to the questions/issues I brought up above:
1. Do the plug-ins that we think will be problematic work with InfluxData's `telegraf` package on Fedora? If so, how?
2. There are comments -- both here in the PR discussion and in `telegraf-hardening.conf` -- are inconsistent about what actually causes the need for the two HOME directories. In some places you've said that `ProtectSystem=full` plays into it and in other places you've said it's only the fact that `/var/lib/telegraf` is root-owned. We need that sorted out.
3. Directly tied to issue 2 above is determining whether the hardening strategy adds value or gets in the way. I brought that up earlier and I think we both kind of lost the thread on it.

base/comps/telegraf/telegraf.sysusers
```diff
@@ -0,0 +1,18 @@
...
+# we point it at /var/lib/telegraf — the same writable location upstream InfluxData designates
+# as the state dir. This gives a single source of truth that works both under systemd and for
+# manual `sudo -u telegraf` debugging, with all hardening left intact.
+#
+# Divergence note: upstream/Fedora set the home field to /etc/telegraf (`useradd -M -d
+# /etc/telegraf`), but with -M they never make it writable, so $HOME writes fail there too.
+# Our /var/lib/telegraf home is functionally equivalent for config and strictly better for
+# anything that writes under $HOME.
+# Type Name      ID  GECOS       Home directory      Shell
+u     telegraf    -   "Telegraf"  /var/lib/telegraf   /usr/sbin/nologin
```
tobiasb-ms: issue(blocking): InfluxData's `telegraf` package uses the home directory `/etc/telegraf`. [This discussion](https://github.com/microsoft/azurelinux/pull/17798#issuecomment-4856041349) talks about it. Let's iron out our strategy there and then we can either leave it as is or change it.

Blocking for now while we figure that out.

tobiasb-ms (changes requested):

base/comps/telegraf/telegraf.spec
```diff
@@ -0,0 +1,201 @@
...
+# License files for the main project and ALL vendored modules (+ modules.txt for the
+# bundled() provides generator). Populates %%{go_vendor_license_filelist}.
+%go_vendor_license_install -c %{S:2}
+
+# Binary
+install -Dpm 0755 %{gobuilddir}/bin/%{name} %{buildroot}%{_bindir}/%{name}
+
+# AzL: --- runtime packaging (omitted by the go2rpm scaffold) -------------------------
+# Systemd unit (shipped by upstream unmodified; runs as User=telegraf). We intentionally add
+# no sandboxing drop-in: telegraf is a monitoring agent that observes the whole system, and our
```
tobiasb-ms: docs(non-blocking): I don't understand the purpose of mentioning the lack of a sandboxing drop-in. Neither AZL 3 nor InfluxData to it (as you mention), so there's no reason to bring it up here right?

base/comps/telegraf/telegraf.spec
```diff
@@ -0,0 +1,201 @@
...
+# Config tree. Generate a default config from THIS binary so it reflects the curated
+# (compiled-in) plugin set rather than the full upstream catalog. Note: this executes the
+# freshly-built target binary, so it assumes a native (non-cross/emulated) build, which is
+# how AzL builds each arch in mock.
+install -d -m 0755 %{buildroot}%{_sysconfdir}/%{name}/%{name}.d
+%{gobuilddir}/bin/%{name} config > %{name}.conf
+install -Dpm 0644 %{name}.conf %{buildroot}%{_sysconfdir}/%{name}/%{name}.conf
+
+# State directory for the optional agent statefile. Owned root:telegraf and mode 0770 to match
+# upstream InfluxData (scripts/rpm/post-install.sh): telegraf writes the statefile via group
```
tobiasb-ms: docs(non-blocking): I see no purpose in explaining how file mode works on linux in these comments.

base/comps/telegraf/telegraf.sysusers
WithEnoughCoffee: changed back to  /etc/telegraf , matching InfluxData's package ( useradd -r -M -d /etc/telegraf ). You're right that InfluxData is the upstream here, not Fedora —  /var/lib/telegraf  was a deviation and I've dropped it. also Dropped Sues inspired hardening to stay consistent with upstream.

base/comps/telegraf/generate_source_tarball.sh
WithEnoughCoffee: Great suggestion, Fixed.

WithEnoughCoffee:
> > Fedora doesn't make telegraf its made by influxData we changed the home directory to match the fedora standard proposed in an earlier comment. I think we should change it back to what InfluxData does, Until we are asked to change it for upstream if that happens.
> 
> If you install `telegraf` from InfluxData's repo (instructions: https://computingforgeeks.com/how-to-install-telegraf-on-fedora/), the user it creates has home directory `/etc/telegraf`:
> 
> ```
> bash-5.3# grep telegraf /etc/passwd
> telegraf:x:999:999::/etc/telegraf:/bin/false
> ```
> 
> So using `/var/lib/telegraf` diverges from what we're currently considering the upstream, right?
> 
> We also still need answers to the questions/issues I brought up above:
> 
> 1. Do the plug-ins that we think will be problematic work with InfluxData's `telegraf` package on Fedora? If so, how?
> 2. There are comments -- both here in the PR discussion and in `telegraf-hardening.conf` -- are inconsistent about what actually causes the need for the two HOME directories. In some places you've said that `ProtectSystem=full` plays into it and in other places you've said it's only the fact that `/var/lib/telegraf` is root-owned. We need that sorted out.
> 3. Directly tied to issue 2 above is determining whether the hardening strategy adds value or gets in the way. I brought that up earlier and I think we both kind of lost the thread on it.

We spoke privately and I addressed these comments.

base/comps/telegraf/telegraf.spec
WithEnoughCoffee: fixed

base/comps/telegraf/telegraf.spec
WithEnoughCoffee: fixed

base/comps/telegraf/telegraf.comp.toml
```diff
@@ -0,0 +1,27 @@
+# Telegraf — agent for collecting, processing, aggregating, and writing metrics.
+# Shipped in AzL 3.0 but missing from 4.0; this restores it as a general-purpose metrics
+# agent. Not available in Fedora; maintained as a local spec with vendored Go dependencies.
+# Curated ("Balanced") custom build keeps the binary's dependency/CVE surface small.
+# Packaged per the Fedora Go packaging guidelines for upstreaming:
+# https://docs.fedoraproject.org/en-US/packaging-guidelines/Golang/
+[components.telegraf]
+spec = { type = "local", path = "telegraf.spec" }
+# Local Go package authored by AzL: use rpmautospec (%autorelease / %autochangelog).
+release = { calculation = "autorelease" }
```
tobiasb-ms: suggestion(non-blocking): I found out yesterday that if you leave out the `release` line and take the default, it will figure out from spec file that you want `autorelease`.
```suggestion
```

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,121 @@
...
+
+echo "Creating a tempdir."
+TMPDIR=$(mktemp -d)
+function cleanup {
+    echo "Clean-up: removing tempdir (${TMPDIR})."
+    rm -rf "${TMPDIR}"
+}
+trap cleanup EXIT
+
+pushd "${TMPDIR}" > /dev/null
```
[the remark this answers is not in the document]
tobiasb-ms: I don't think this is worth doing.

I'm pretty sure that the cited "repo guidance" is specific to the [mass triage skill](https://github.com/microsoft/azurelinux/blob/718a7a3c0159306663d16786cbb973d1cfc1916d/.github/skills/skill-mass-triage/SKILL.md?plain=1#L64), not to stand-alone scripts for vendoring.

Resolving as won't-fix.

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,121 @@
...
+
+NAME_VER="telegraf-${PKG_VERSION}"
+# Fedora forge macros expect the vendor archive as %{archivename}-vendor.tar.bz2.
+VENDOR_TARBALL="$(realpath "${OUT_FOLDER}/${NAME_VER}-vendor.tar.bz2")"
+
```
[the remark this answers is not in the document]
tobiasb-ms: I'm not convinced this is correct about `realpath`. My experience is that it works if the leaf node doesn't exist but fails if paths leading to the leaf don't exist.

That said, it's 100% correct that `OUT_FOLDER` is already canonicalized so there's no need to do so here.

I suggest taking this suggestion.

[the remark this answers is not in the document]
WithEnoughCoffee: will do

PawelWMS (approved):

base/comps/telegraf/generate_source_tarball.sh
```diff
@@ -0,0 +1,183 @@
...
+    exit 1
+fi
+case "${1}" in
+    -h|--help)
+        usage
+        exit 0
+        ;;
+esac
+
+SRC_TARBALL="${1}"
```
tobiasb-ms: suggestion(non-blocking): I think you're taking the wrong parameter here. Above, you have the link we can get the tarball from. I suggest you take the package version and then the script can curl/wget it. That way the script controls everything but the package version, which is the only thing that really should vary.

It's something to consider. Non-blocking suggestion though.

tobiasb-ms (changes requested):
Changes look good. Once you rebase down to one commit I'll approve.

tobiasb-ms (approved):

tobiasb-ms (changes requested):
Looks like there's a problem with the lock/render.

tobiasb-ms (approved):

christopherco (approved):

PR #17801, opened 2026-06-25
feat(wsl-setup): explicitly enable WSL interop in wsl.conf

mfrw:
Add an [interop] section to the wsl.conf shipped by wsl-setup, pinning
both interop keys explicitly:

  [interop]
  enabled=true
  appendWindowsPath=true

This is needed for WSL to spawn a windows process.

Implemented as a file-search-replace overlay on the upstream Fedora
wsl-setup component (anchored on the [boot] systemd line) to keep
divergence minimal. Promotes wsl-setup from an inline entry in
components.toml to a dedicated comp.toml since it now carries an overlay.


base/comps/wsl-setup/wsl-setup.comp.toml
```diff
@@ -0,0 +1,23 @@
...
+description = "Explicitly enable WSL interop and Windows PATH appending in /etc/wsl.conf"
+type = "file-search-replace"
+file = "wsl.conf"
+# Anchor on the existing [boot] systemd line and append an [interop] section.
+# file-search-replace does a literal replacement (no backreferences), so the
+# matched line is re-emitted verbatim ahead of the appended section. Matching
+# this line deliberately keeps the overlay in lockstep with upstream: if Fedora
+# changes the [boot] body the overlay fails loudly at render time rather than
+# silently dropping the interop config.
+regex = 'systemd=true'
```
reubeno: regexes are generally the last resort. Can this not be appended to the end of the file?

base/comps/wsl-setup/wsl-setup.comp.toml
```diff
@@ -0,0 +1,23 @@
+[components.wsl-setup]
+
+# Explicitly enable WSL interop in the shipped /etc/wsl.conf. Interop (and
+# Windows PATH appending) is enabled by default in WSL, but pinning both keys
```
reubeno, line 4: What do other WSL images (from other distros) do? Do they do this as well? If so, should we be pushing this upstream to Fedora?

reubeno (changes requested):

base/comps/wsl-setup/wsl-setup.comp.toml
mfrw: It is actually supposed to be a default, but for some versions of WSL it does not work.
I took insipiration from this WSL github issue: https://github.com/microsoft/WSL/issues/13872#issuecomment-3705335743

base/comps/wsl-setup/wsl-setup.comp.toml
mfrw: Sure, let me use `file-prepend-lines` here.
IMHO, the only file overlay functions i am aware of are:
[file-prepend-lines, file-search-replace, file-add, file-remove, file-rename]
https://github.com/microsoft/azure-linux-dev-tools/blob/main/docs/user/reference/config/overlays.md#file-overlays

What do you suggest, should we add another function `file-append-lines` ?

christopherco (changes requested):
issue: Change itself can make sense, but the change should be unnecessary since the default interop value should already be `true`. I have requested @mfrw to follow up with WSL team to confirm the expected behavior

mfrw:
Had a word with the WSL team, and this is not explicitly needed.

PR #17810, opened 2026-06-25
feat(fluent-bit): add fluent-bit imported from Fedora

tobiasb-ms:
Adds fluent-bit as a Fedora-imported component (Fedora 43, v4.2.2) and publishes it to the rpm-base channel.

Built locally and installed via a local repo with no problems, so there shouldn't be any closure issues.

Built on build server.

Fixes: [AB#20400](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20400)


reubeno (commented):
In 3.0, I believe a change was made to use dalec to build fluent-bit. Is this a temporary change? Or is the intention that it's now being moved to `base`?

reubeno (approved):

PR #17843, opened 2026-06-29
fix(openmpi): restore openmpi-devel subpackages

miz060:
## Summary
Restore the Azure Linux 3.0 OpenMPI packaging contract after the DOCA 3.3.0 update.

The spec now:
- keeps runtime files in `openmpi`
- builds `openmpi-devel` and `openmpi-docs`
- restores `/usr/share/modulefiles/mpi/openmpi-<arch>`

This fixes the golden-container flow, which installs `openmpi openmpi-devel` and then loads `mpi/openmpi` before running `mpiexec --version`.

## Validation
- RPM build passed: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1150354&view=results
- RPM signing passed: https://dev.azure.com/mariner-org/mariner/_build?definitionId=4714
- OpenMPI container built: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1150455&view=results


SPECS/openmpi/openmpi.spec
```diff
@@ -245,6 +245,9 @@ Requires: %{modules_rpm_name}
 Requires: %{mpi_selector_rpm_name}
 %endif
 Requires: ucx
+%if !%{build_all_in_one_rpm}
```
ellie-di: could we just get rid of the `build_all_in_one_rpm` (assuming we want several sub-package RPMs)?

miz060: good point. removed the `build_all_in_one_rpm` path

jslobodzian (approved):

ellie-di (approved):

christopherco (approved):

PR #17851, opened 2026-06-29
feat: add rust-md5 component (required for clamav 1.5.2 upgrade)

binujp:
## Summary

Add the `rust-md5` crate (version 0.7.0) from Fedora. This is a direct dependency of ClamAV 1.5.2's `libclamav_rust` for computing MD5 hashes of CVD signature database files.

## Note on rust-md5 vs rust-md-5

- **rust-md5** (crate: `md5`, v0.7.0) — from github.com/stainless-steel/md5. Simple, standalone MD5 implementation with its own API (`md5::compute()`, returns `md5::Digest`). Used by ClamAV.

- **rust-md-5** (crate: `md-5`, v0.10.6, already in AZL4) — from the RustCrypto project. Implements the RustCrypto `Digest` trait interface (`md5::Md5::new()`, `update()`, `finalize()`). Part of the broader RustCrypto hash family.

They are different crates with incompatible APIs; one cannot substitute for the other.

## Testing

- `azldev comp build -p rust-md5` — builds successfully, produces `rust-md5-devel`, `rust-md5+default-devel`, `rust-md5+std-devel` subpackages.


reubeno (approved):

tobiasb-ms (changes requested):
question(blocking): Confirming we only need this in the sdk repo, not the base repo, and that installation/closure has been verified?

binujp:
> question(blocking): Confirming we only need this in the sdk repo, not the base repo, and that installation/closure has been verified?

That is an interesting question. This is being added only for clamav 1.5.2 build. As such only that build will need it. If, however, we want to set a policy for crates and that says these should be in base, only then I see it needing to move. I will bring this up in the external discussion.

binujp:
Discussed this in the teams thread. The default location for rust crate-rpms, unless there is a specific known exception, is the sdk repo.

tobiasb-ms (approved):

PR #17853, opened 2026-06-29
ci: flip back to AFD endpoint for CT, fix merge base calc

dmcilvaney:


dmcilvaney:
AFD test at https://dev.azure.com/mariner-org/azl/_build/results?buildId=1149381
And adding merge base change: https://dev.azure.com/mariner-org/azl/_build/results?buildId=1149385

.github/workflows/ado/templates/package-build-stages.yml
```diff
@@ -81,7 +81,7 @@ stages:
                   --repo-uri "$UPSTREAM_REPO_URL"
             env:
               API_AUDIENCE: $(ApiAudience)
-              API_BASE_URL: $(ApiBaseDirectUrl)
+              API_BASE_URL: $(ApiBaseAFDUrl)
```
reubeno, line 84: Why are these variable names even distinguishing between different kinds of URLs? Could we not abstract it and just request a base API URL?

dmcilvaney: We knew we would be switching back, `ApiBaseAFDUrl` is the legacy URL we used before we had to move away from AFD for a while. We can delete `ApiBaseDirectUrl` now if desired. Both are defined in ADO.

reubeno (approved):

PR #17873, opened 2026-06-30
feat(SymCrypt and SymCrypt-OpenSSL): Add SymCrypt OpenSSL provider to AZL4

tobiasb-ms:
Adds two packages:
- `SymCrypt`, which is a cryptographic library from Microsoft.
- `SymCrypt-OpenSSL`, which implements an `OpenSSL` provider using `SymCrypt`.

Fedora does not carry these packages so I based them AZL3, modifying them for AZL4 and updating the versions. I also have a [different branch](https://github.com/microsoft/azurelinux/compare/4.0...tobiasb-ms/symcrypt-full-history) with a more complete history, which may be useful to see the differences between AZL4 and AZL3. However, I squashed those commits for the PR.

Validation:
- Build locally
- Installed on a container using a local yum repo (proves closure)
- Did basic `openssl` operations with `SymCrypt-OpenSSL` installed
- Ran `openssl speed` which does a bunch of cryptographic operations


base/comps/SymCrypt-OpenSSL/0003-bump-project-version-to-match-release-tag.patch
```diff
@@ -0,0 +1,30 @@
+From 9fc7d0510476acb5aa495e59c9b34248f3363ab2 Mon Sep 17 00:00:00 2001
```
reubeno, line 1: It looks like this and other fixes landed in https://github.com/microsoft/SymCrypt-OpenSSL/commit/0ac9346c9ed2fa1f617715fbb80cbbb0bf095a7a upstream (in the 1.9 servicing branch). Should we be taking them?

What's been our previous policy on taking fixes from that upstream? Do we wait for a version tag? Since they don't appear to have an official release of source code anyhow -- we're presumably reliant on GitHub archives, which equally well work for individual commits.

base/comps/SymCrypt-OpenSSL/0001-register-symcrypt-provider-for-config-drop-in.patch
```diff
@@ -0,0 +1,40 @@
+From 115f648995926362080f05fe80d670ed6494ce5c Mon Sep 17 00:00:00 2001
+From: Tobias Brick <tobiasb@microsoft.com>
+Date: Fri, 26 Jun 2026 14:53:36 +0000
+Subject: [PATCH 1/2] Register the SymCrypt provider in [provider_sect] for
+ config drop-in
+
+Add a [provider_sect] stanza to symcrypt_prov.cnf that maps the
```
reubeno, line 7: Is this something that we could/should upstream? It doesn't sound specific to Azure Linux.

base/comps/SymCrypt-OpenSSL/SymCrypt-OpenSSL.comp.toml
```diff
@@ -0,0 +1,25 @@
...
+# lookaside, so the origin GitHub release URL is what actually resolves it.
+#
+# Release and changelog are managed by rpmautospec (release.calculation =
+# "autorelease"): the Release integer is derived from this component's git
+# history at render time, and the spec's %autochangelog generates the changelog
+# from the same history. The inherited changelog entries are preserved verbatim
+# in the `changelog` sidecar file and appended below the generated entries.
+[components.SymCrypt-OpenSSL]
+spec = { type = "local", path = "SymCrypt-OpenSSL.spec" }
+release = { calculation = "autorelease" }
```
reubeno: Does the `release` type need to be specified? I would expect the default should just work.

base/comps/SymCrypt-OpenSSL/SymCrypt-OpenSSL.comp.toml
```diff
@@ -0,0 +1,25 @@
+# SymCrypt-OpenSSL (SCOSSL) lets OpenSSL use SymCrypt as the provider for its
+# core cryptographic operations. It is maintained here as a local spec ported
```
reubeno: Could the comments here be simplified? They do read very much like verbose AI details.

For example, the fact that it's derived from Azure Linux 3 is relevant for the PR and the commit but are already tracked by the changelog. Do you think it will still be meaningful to us down the road to have here?

Similarly -- this component doesn't need to explain how release/autorelease work, nor how tarball resolution works. I'd only call out exceptions or ways in which this component is unique from others.

base/comps/SymCrypt-OpenSSL/SymCrypt-OpenSSL.spec
```diff
@@ -0,0 +1,100 @@
...
+%define symcrypt_arch ARM64
+%endif
+
+%prep
+%autosetup -p1
+
+%build
+mkdir bin; cd bin
+
+cmake   .. \
```
reubeno: Is it possible to use the standard `%cmake*` macros instead?

base/comps/SymCrypt/SymCrypt.comp.toml
```diff
@@ -0,0 +1,34 @@
+# SymCrypt is Microsoft's core cryptographic library, maintained here as a
```
reubeno: _See similar note from the other .comp.toml file regarding comments._

base/comps/SymCrypt/symcrypt-build-environment-variables-v103.11.0.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
reubeno, line 1: If symcrypt needs this info to build, it really should be included in the released artifacts from the SymCrypt project. This requires us as a consumer to understand/know the upstream-upstream source/branch origin of the specific release that we're pinning to -- and also requires a manual/custom step on upgrades.

If we need to take it on for now to mirror what we did in 3.0, then so be it -- but we should file a bug and try to work with them upstream on this.

base/comps/SymCrypt/SymCrypt.spec
```diff
@@ -0,0 +1,97 @@
...
+%define symcrypt_arch ARM64
+# Currently SymCrypt ARM64 build requires use of clang
+%define symcrypt_cc clang
+%define symcrypt_c_flags "-mno-outline-atomics -Wno-conditional-uninitialized"
+%define symcrypt_cxx clang++
+%endif
+
+%prep
+%autosetup -a 1 -p1
+# Create a symbolic link as if jitterentropy-library has been pulled in as git submodule
```
reubeno: Can you remind me why we need to replace what they carry? I recall a heinous bug somewhere that was an issue but don't recall if we tried to get the SymCrypt project upstream to update their submodule pointer to pull in a more recent version of this library.

base/comps/SymCrypt/SymCrypt.spec
```diff
@@ -0,0 +1,97 @@
...
+# CHANGE FROM 3.0: REMOVE -z pack-relative-relocs
+# AZL4's default LDFLAGS include -Wl,-z,pack-relative-relocs, which makes the
+# linker emit R_X86_64_RELATIVE entries in DT_RELR format instead of SHT_RELA.
+# SymCrypt's FIPS post-processor (process_fips_module.py) only walks SHT_RELA,
+# so RELR-packed relocations are not zeroed before the integrity HMAC is
+# computed. The dynamic loader still applies them at load time, so in-memory
+# bytes diverge from the hashed bytes and the FIPS self-test fails. Combined
+# with -z now, the failure surfaces as a SIGSEGV the instant libsymcrypt.so
+# is mapped (e.g. any LD_PRELOAD or DT_NEEDED user crashes immediately).
+# Remove until process_fips_module.py learns about DT_RELR.
```
reubeno, line 65: Do we have a bug on this?

base/comps/SymCrypt/SymCrypt.spec
```diff
@@ -0,0 +1,97 @@
...
+export LDFLAGS="${LDFLAGS//-Wl,-z,pack-relative-relocs/}"
+
+source %{SOURCE3}
+
+cmake   -S . -B bin \
```
reubeno: Could we use the standard cmake macros?

base/comps/SymCrypt-OpenSSL/0003-bump-project-version-to-match-release-tag.patch
tobiasb-ms: Yeah, they use [version tags](https://github.com/microsoft/SymCrypt-OpenSSL/tags) as their releases and we get the tarballs from those tags, like the [1.9.6 tarball](https://github.com/microsoft/SymCrypt-OpenSSL/archive/refs/tags/v1.9.6.tar.gz).

I'm hesitant to take that commit just for the version -- it has other changes I know they've been working on. I believe they are testing a new version that we'll likely take, which should come out relatively soon.

base/comps/SymCrypt-OpenSSL/0001-register-symcrypt-provider-for-config-drop-in.patch
tobiasb-ms: I have a conversation going with them about the overall packaging and this will definitely be one of those things. Agree that it isn't an AZL-specific thing, but it is AZL4-specific vs. AZL3, due to some differences in approach to providers and SymCrypt specicifically.

So yes, it should and probably will be upstreamed unless they have a specific reason not to.

base/comps/SymCrypt-OpenSSL/SymCrypt-OpenSSL.comp.toml
tobiasb-ms: You're right. This is basically AI slop that I let sneak in. I'm glad you caught it.

base/comps/SymCrypt/symcrypt-build-environment-variables-v103.11.0.sh
tobiasb-ms: I hadn't thought of that. This is a great point that I'll discuss with the upstream maintainers.

base/comps/SymCrypt/SymCrypt.spec
tobiasb-ms: They don't actually package this as part of the tarball that comes with their tag. I don't know the history of that decision.

The terrible bug you're thinking of was in with JitterEntropy, but not with SymCrypt. It was in AZL2 on a very old version of JitterEntropy that was a little bit of a frankenbuild.

base/comps/SymCrypt/SymCrypt.spec
tobiasb-ms: On my list I'm talking to them about.

base/comps/SymCrypt-OpenSSL/SymCrypt-OpenSSL.comp.toml
tobiasb-ms: Looks like you're right.

base/comps/SymCrypt-OpenSSL/SymCrypt-OpenSSL.comp.toml
tobiasb-ms: Honestly, none of those comments were useful. Spec has (or has links to) all the information that's worth having there.

base/comps/SymCrypt/SymCrypt.spec
tobiasb-ms: Yes. They don't seem to have a real `install` target for `cmake` so I'm keeping the manual `%install` section with `%{__cmake_builddir}` instead of `bin`.

base/comps/SymCrypt/SymCrypt.spec
tobiasb-ms: I updated the comment in the spec to reflect the fact that this doesn't come with the SymCrypt tarball.

reubeno (approved):

PR #17876, opened 2026-06-30
feat: add inotify-tools for AKS log syncing

miz060:
Add `inotify-tools` to Azure Linux 4.0 as a Fedora 43 bare upstream import.

This will provide `inotifywait` and related filesystem event utilities used by AKS container log syncing.

This spec will produce these RPM artifacts into base-rpm: `inotify-tools`, `inotify-tools-devel`, `inotify-tools-debuginfo`, `inotify-tools-debugsource`

- Validation: 
  - azldev local build&install passed. 
  - control tower build passed

Fixes: [AB#22285](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22285)


binujp (changes requested):
Blocking: The change looks good and can merge. However, we do not want internal URLs exposed on public forums. A statement to the effect that the koji build succeeded in addition to local is good enough.

binujp (approved):

PR #17881, opened 2026-07-01
test: Language Runtime Container tests

bhagyapathak:
Language runtime tests for container

### DIFF
--- a/base/images/tests/cases/runtime/container-base/test_nodejs/test_nodejs.py
+++ b/base/images/tests/cases/runtime/container-base/test_nodejs/test_nodejs.py
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: MIT
+"""Validate the Node.js runtime works on the container-base image.
+
+Uses ``@pytest.mark.dockerfile()`` to build a custom image with Node.js
+installed on top of the image-under-test, then runs a stdlib ``http``
+server and checks its response.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+EXPECTED_RESPONSE = (Path(__file__).with_name("response.txt")).read_text().strip()
+
+
+@pytest.mark.dockerfile()
+def test_nodejs_version(container_exec_shell) -> None:
+    """Node.js interpreter must be present and report a version."""
+    result = container_exec_shell("node --version")
+    assert result.exit_code == 0, f"node --version failed: {result.output}"
+    assert result.output.strip().startswith("v")
+
+
+@pytest.mark.dockerfile()
+def test_nodejs_http_server(assert_http_server) -> None:
+    """A stdlib http server must serve the expected response."""
+    assert_http_server(
+        "nohup node /app/server.js > /tmp/server.log 2>&1 &",
+        "http://localhost:8080/",
+        EXPECTED_RESPONSE,
+    )
--- a/base/images/tests/cases/runtime/container-base/test_php/router.php
+++ b/base/images/tests/cases/runtime/container-base/test_php/router.php
@@ -1,0 +1,29 @@
+<?php
+// router.php
+// Exercises the php-pecl-zip extension by creating, writing and
+// extracting a zip archive, then verifying the round-trip.
+if (preg_match('/\.(?:png|jpg|jpeg|gif)$/', $_SERVER["REQUEST_URI"])) {
+    return false;    // serve the requested resource as-is.
+} else {
+    $zip = new ZipArchive();
+    $testFolder = shell_exec("mktemp -d | tr -d '\n'");
+    $filename = "$testFolder/test_archive.zip";
+    if ($zip->open($filename, ZipArchive::CREATE) !== TRUE) {
+        exit("cannot open <$filename>\n");
+    }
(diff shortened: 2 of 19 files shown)


bhagyapathak:
[Uploading language_runtime_tests_logs.txt…]()

base/images/tests/cases/runtime/container-base/test_php/router.php
```diff
@@ -0,0 +1,29 @@
+<?php
+// router.php
+// Exercises the php-pecl-zip extension by creating, writing and
+// extracting a zip archive, then verifying the round-trip.
+if (preg_match('/\.(?:png|jpg|jpeg|gif)$/', $_SERVER["REQUEST_URI"])) {
+    return false;    // serve the requested resource as-is.
+} else {
+    $zip = new ZipArchive();
```
tobiasb-ms, line 8: question(non-blocking): I'm wondering what motivated the more extensive testing (zip/unzip) for php and why you didn't do something similar for the other languages.

base/images/tests/cases/runtime/container-base/test_nodejs/test_nodejs.py
```diff
@@ -0,0 +1,47 @@
...
+from __future__ import annotations
+
+import time
+
+import pytest
+
+EXPECTED_RESPONSE = "Hello from the node server running from Mariner NodeJS container."
+
+
+def wait_for_http(container_exec_shell, url: str):
```
tobiasb-ms: suggestion(blocking): Unless I've misread them, `wait_for_http` is exactly the same across all of the `test_*.py` files. Can we find a way to centralize and reuse this function so we don't maintain copies of it?

base/images/tests/cases/runtime/container-base/test_nodejs/test_nodejs.py
```diff
@@ -0,0 +1,47 @@
...
+@pytest.mark.dockerfile()
+def test_nodejs_version(container_exec_shell) -> None:
+    """Node.js interpreter must be present and report a version."""
+    result = container_exec_shell("node --version")
+    assert result.exit_code == 0, f"node --version failed: {result.output}"
+    assert result.output.strip().startswith("v")
+
+
+@pytest.mark.dockerfile()
+def test_nodejs_http_server(container_exec_shell) -> None:
```
tobiasb-ms: suggestion(blocking): Much like `wait_for_http`, the `test_*_http_server` functions are mostly the same. Can we extract the identical logic/code into a helper function or use some other pattern that allows code re-use here?

base/images/tests/cases/runtime/container-base/test_nodejs/test_nodejs.py
```diff
@@ -0,0 +1,47 @@
...
+import pytest
+
+EXPECTED_RESPONSE = "Hello from the node server running from Mariner NodeJS container."
+
+
+def wait_for_http(container_exec_shell, url: str):
+    """Poll until an HTTP endpoint responds successfully."""
+    result = None
+    for _ in range(5):
+        result = container_exec_shell(f"curl -sf {url}")
```
tobiasb-ms: suggestion(blocking): I think best practice will be to put `url` in quotes to prevent shell injection. Doesn't realistically matter in this case because we own all callers, but it's easy.

```suggestion
        result = container_exec_shell(f"curl -sf '{url}'")
```

base/images/tests/cases/runtime/container-base/test_nodejs/test_nodejs.py
```diff
@@ -0,0 +1,47 @@
...
+
+import pytest
+
+EXPECTED_RESPONSE = "Hello from the node server running from Mariner NodeJS container."
+
+
+def wait_for_http(container_exec_shell, url: str):
+    """Poll until an HTTP endpoint responds successfully."""
+    result = None
+    for _ in range(5):
```
tobiasb-ms: suggestion(non-blocking): Would it be worth it to move the retry logic into `container_exec_shell`, as an optional parameter like `num_retries`?

tobiasb-ms (changes requested):

base/images/tests/cases/runtime/container-base/test_php/router.php
bhagyapathak: Thank you @tobias for the review. 
test_php_zip_extension_loaded is intentionally PHP-specific. This is a direct port of the existing golden-container smoke tests, and my current goal is to bring them to the Azure Linux 4.0 base container with minimal changes. Once the golden containers are available, I plan to run the same test suite against them as well. For that reason, I have kept the tests as close to the original implementation as possible. But I am open to the option of adding this test later when golden container is availble.

The ZIP validation originates from the upstream PHP golden-container test. The php-pecl-zip package is an optional extension, so the presence of PHP alone does not guarantee that the ZIP extension is available. That's why the test explicitly verifies it through both a php -m check and a functional round-trip validation.

The other runtimes do not have an equivalent optional, headless-relevant extension.

tobiasb-ms (approved):

christopherco (approved):

PR #17887, opened 2026-07-01
chore(comps): annotate azl-compatibility overlays with metadata

liunan-ms:
**Summary**

This PR migrates all `azl-compatibility` overlay definitions to the per-file overlay layout with file-level metadata. This is a pure restructuring/annotation change — no overlay is added, removed, reordered, or semantically altered, and the rendered specs are byte-identical to 4.0.

Complements the sibling PRs [#17727](https://github.com/microsoft/azurelinux/pull/17727), [#17888](https://github.com/microsoft/azurelinux/pull/17888). Components already migrated on those branches are skipped.

**Changes**

- **chore(comps): annotate azl-compatibility overlays with metadata**
  Migrates **27 components** from inline `overlays = [...]` in their `.comp.toml` to per-group `overlays/*.overlay.toml` files (**42 new overlay files**).
  - One overlay toml file per logical change, even for single-overlay components.
  - Each file carries a file-level `[metadata]` block with `category` (and `upstream-status` / commit references where applicable). Multi-category components (apache-sshd, kernel-headers, mock-core-configs, rocm-compilersupport, rust, shim, unicode-ucd) get one file per logical change group, each stamped with that group's own category.
  - All original `.comp.toml` comments (including interior comments between overlay fields) are preserved verbatim into the overlay files; the `.comp.toml` files are reduced to their bare `[components.<name>]` headers.
  - Components already migrated on the `azl-pruning` / `overlay_metadata` branches are skipped.
  - Categories used across the migrated files: `azl-compatibility`, `azl-branding-policy`, `azl-release-management`, `azl-temp-workaround`, `azl-pruning`, `azl-platform-adaptation`.
  - Components touched: apache-sshd, dnf5, fcode-utils, iscsi-initiator-utils, kernel-headers, libspf2, libyang, llvm, llvm20, mock-core-configs, oneapi-level-zero, onnx, open-vm-tools, openblas, openexr, openpace, pcsc-lite-asekey, perl-Archive-Any-Lite, perl-Crypt-PasswdMD5, python-heatclient, python-manilaclient, python-octaviaclient, python-pyside6, rocm-compilersupport, rust, shim, unicode-ucd.

- **TEMP: enable overlay-files loading for PR check — ⚠️ PR-CHECK ONLY, REVERT BEFORE MERGE**
  Temporary. Sets `overlay-files = ["overlays/*.overlay.toml"]` in the `[default-component-config]` of `base/comps/components.toml`, so every component inherits per-file overlay discovery (no per-component key needed) and the new `*.overlay.toml` files are exercised by PR checks. Standalone so it can be reverted independently.

- **chore(azldev): pin to 830e6fed and regenerate schema — ⚠️ PR-CHECK ONLY, REVERT BEFORE MERGE**
  Temporary. Pins `.azldev-version` to `830e6fed` (upstream-backport category, URLRef replacing BugRef, required `upstream-status` replacing the old `upstreamable` boolean, `azl-dep-missing-workaround` → `azl-temp-workaround` rename, and optional component-group `[metadata]`) and regenerates `external/schemas/azldev.schema.json` so PR checks run against a tool that understands the new `overlay-files` / metadata fields. Must be reverted before this PR merges.

**Validation**

- ✅ Structural no-op: all 27 components' reconstructed overlay lists are byte-identical to 4.0.
- ✅ `azldev comp list -a` parses cleanly.
- ✅ `comp render` in PR check passed.


base/comps/apache-sshd/overlays/0001-remove-maven-remote-resources.overlay.toml
```diff
@@ -0,0 +1,26 @@
+[metadata]
+category = "azl-compatibility"
```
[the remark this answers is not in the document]
liunan-ms: Unset `upstreamable` means we're not sure if we should push this change to upstream, leaving it unset to re-visit in the future.

distro/azurelinux.distro.toml
```diff
@@ -19,6 +19,10 @@ mock-config-aarch64 = "mock/azl4/stage1/azurelinux-4.0-aarch64.cfg"
 [distros.azurelinux.versions.'4.0-stage1'.default-component-config]
 # NOTE: Ensure that the snapshot date matches the snapshot date used below for stage2.
 spec = { type = "upstream", upstream-distro = { name = "fedora", version = "43", snapshot = "2026-02-24T00:00:00-08:00" } }
+# Discover per-file overlay documents from each component's ./overlays/ directory.
+# Inherited relative pattern is resolved per-component; globs that match no files
+# are ignored, so components without an overlays/ dir are unaffected.
+overlay-files = ["overlays/*.overlay.toml"]
```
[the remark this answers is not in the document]
liunan-ms: This will be removed before PR merge.

tobiasb-ms (approved):
I spot-checked some things and the fact that the spec renderer ran and didn't complain -- despite no changes to `specs/` -- indicates this is a functional no-op. Approved.

base/comps/apache-sshd/overlays/0002-remove-buildrequires-maven-local.overlay.toml
```diff
@@ -0,0 +1,19 @@
+# AZL ships only java-25-openjdk; there is no java-21-openjdk, so the
+# upstream BuildRequires on maven-local-openjdk21 is unsatisfiable.
+# Retarget to the openjdk25 variant.
+
+[metadata]
+category = "azl-pruning"
+upstreamable = false
```
reubeno: Worth considering upstreamability here. Fedora will also need to move to 25 as the default at some point.

base/comps/apache-sshd/overlays/0001-remove-maven-remote-resources.overlay.toml
reubeno, line 10: Why? Do we have any indication of what these changes were actually doing?

base/comps/fcode-utils/overlays/0001-force-std-gnu17-work.overlay.toml
```diff
@@ -0,0 +1,17 @@
+# GCC 15 defaults to C23 where `bool` is a keyword, conflicting with the
+# project's `typedef enum boolean { FALSE = 0, TRUE = -1 } bool;` in
+# shared/types.h. Force C17 (with GNU extensions) as a conservative workaround
+# until a new upstream release includes the fix.
+# Upstream fix: https://github.com/openbios/fcode-utils/issues/20
+# TODO: Remove this overlay once a new upstream release (> 1.0.3) is available.
+
+[metadata]
+category = "azl-compatibility"
```
reubeno: I worry that azl-compatibility is being more widely used than expected. I see it used for tdnf compat symlinks, which is for compat with previous AZL versions, but this isn't really the same sort of change.

base/comps/kernel-headers/overlays/0001-set-specversion-3-part.overlay.toml
```diff
@@ -0,0 +1,15 @@
+[metadata]
+category = "azl-platform-adaptation"
```
reubeno: The azldev docs call out this category for "architecture-specific" adaptation. Is this architectural?

base/comps/kernel-headers/overlays/0004-6-18-31-1.overlay.toml
```diff
@@ -0,0 +1,18 @@
+# Source: use full Azure Linux kernel tarball instead of pre-built headers.
+
+[metadata]
+category = "azl-platform-adaptation"
```
reubeno: Architectural?

base/comps/libspf2/overlays/0001-disable-short-names-override.overlay.toml
```diff
@@ -0,0 +1,37 @@
...
+#   2. Disable CALL_GRAPH in Doxyfile.in — even with deterministic filenames,
+#      graphviz-generated graph images (SVG/PNG) can have nondeterministic
+#      content due to layout differences across architectures. Removing call
+#      graphs eliminates this second source of divergence. The HTML pages,
+#      class/file indexes, and all textual content are unaffected.
+
+# Layer 1: Neutralize SHORT_NAMES=YES — keep the default (NO) for deterministic filenames
+
+[metadata]
+category = "azl-compatibility"
```
reubeno, line 23: This change doesn't seem in any way specific to Azure Linux -- is there an appropriate category for that, where we're fixing issues that exist in Fedora?

base/comps/llvm/overlays/0001-gcc-toolchain-triple-clang.overlay.toml
```diff
@@ -0,0 +1,9 @@
+[metadata]
+category = "azl-compatibility"
+upstreamable = false
```
reubeno: This looks upstreamable to me.

base/comps/llvm20/overlays/0001-gcc-toolchain-triple-clang20.overlay.toml
```diff
@@ -0,0 +1,9 @@
+[metadata]
+category = "azl-compatibility"
+upstreamable = false
```
reubeno: Looks upstreamable to me.

base/comps/mock-core-configs/overlays/0002-add-4-0-mock.overlay.toml
```diff
@@ -0,0 +1,76 @@
+#
+# Add the Azure Linux 4.0 mock configs and template.
+# These files ship the chroot definition that lets users run
+# `mock <srpm>` on an Azure Linux 4.0 host and produce AZL4 RPMs.
+#
+
+[metadata]
+category = "azl-compatibility"
+upstreamable = false
```
reubeno: Actually is upstreamable. And in fact, I believe Tobias already upstreamed it.

reubeno:
I'm going through this PR and will post comments within the day. My main initial feedback is that the categorization is aggressively using azl-compatibility *aggressively* -- and in a way that I think will make the categorization less useful.

base/comps/rocm-compilersupport/overlays/0002-drop-llvm-triple-cfg.overlay.toml
```diff
@@ -0,0 +1,70 @@
...
+#   2. pre-staged inside the stage-2 build tree before %cmake_build runs, so
+#      the LLVM runtimes sub-build (compiler-rt, libcxx, libcxxabi) can
+#      configure once stage-2 clang is produced;
+#   3. installed into %{bundle_prefix}/bin alongside the shipped clang, so
+#      every downstream ROCm consumer (mivisionx, rocblas, hipblaslt, …)
+#      sees the right toolchain triple without needing any change of its own.
+
+[metadata]
+category = "azl-compatibility"
+upstreamable = false
```
reubeno: This is an example of a set of changes that aren't directly upstreamable, but which we don't want to carry forever. How would you recommend we capture that?

base/comps/rust/overlays/0002-s-compiler-rt-runtime.overlay.toml
```diff
@@ -0,0 +1,17 @@
+[metadata]
+category = "azl-compatibility"
+upstreamable = false
```
reubeno: Looks upstreamable.

base/comps/shim/overlays/0001-release-management-change.overlay.toml
```diff
@@ -0,0 +1,26 @@
...
+[metadata]
+category = "azl-release-management"
+upstreamable = false
+
+[[overlays]]
+type = "spec-search-replace"
+regex = "%global dist.*"
+replacement = ""
+
+# TEMPORARILY remove release from the dependency on the shim unsigned packages
```
reubeno, line 13: How do we capture that this is intended as a temporary workaround?

reubeno (dismissed):
I don't think we need the data to be 100% accurate to get started, but I do want us to see if we're going to get the insights we need here.

base/comps/apache-sshd/overlays/0001-remove-maven-remote-resources.overlay.toml
liunan-ms: This change works around an upstream Fedora 43 packaging bug: the installed versions of `maven-remote-resources-plugin`, `plexus-utils`, and `plexus-xml` in the Fedora 43 repos are mutually incompatible during the Maven 3 -> 4 transition. This issue has been existing in Fedora 43 and failing the build. The Fedora 44 ([commit](https://src.fedoraproject.org/rpms/apache-sshd/c/7959d325a441ae1f8aa0aaabf91eec64e06b983b?branch=f44)) upgraded `apache-sshd` version to 2.16 and disabled `maven-remote-resources-plugin` for the rpm build avoid this issue. I think `azl-compability` is a good fit for this change with `upstream-status = "upstreamed"` and `commits` without changing the spec and rebuild, so that we can remove this overlay once pinning to f44.

liunan-ms:
> I don't think we need the data to be 100% accurate to get started, but I do want us to see if we're going to get the insights we need here.

Thanks! This pushed me to re-audit every `azl-compatibility` overlay in this PR. In the latest push I moved overlays to more specific buckets: `azl-branding-policy` (vendor/triple/%_vendor changes), `azl-release-management` (version/release-string mechanics), `azl-platform-adaptation` (genuinely arch-specific), `azl-temp-workaround` (this is newly introduced in azldev for temporary workaround that is explicitly intended to be dropped once an upstream or environmental fix lands, [here](https://github.com/microsoft/azure-linux-dev-tools/blob/main/docs/user/reference/config/overlays.md?plain=1#L97) is the definition), and `azl-pruning` (dep/feature removals). I also added `upstream-status` (`upstreamable` / `upstreamed` / `inapplicable` / `needs-upstream-hook`) and `commits` refs so we can distinguish "carry forever" from "drop on next rebase." `azl-compatibility` is now reserved for changes that exist purely to bridge to upstream/newer-Fedora behavior we expect to inherit.

base/comps/apache-sshd/overlays/0002-remove-buildrequires-maven-local.overlay.toml
liunan-ms: Marked `upstream-status = "upstreamable"`

base/comps/fcode-utils/overlays/0001-force-std-gnu17-work.overlay.toml
liunan-ms: Agree, I recategorized to `category = "azl-temp-workaround"` (it forces -std=gnu17 to dodge the GCC 15 C23 bool-keyword conflict) with `upstream-status = "upstreamable"` and the upstream fix commit in `commits`. It comments a TODO to remove once an upstream release > 1.0.3 ships. That better reflects that it's a temporary workaround.

base/comps/kernel-headers/overlays/0001-set-specversion-3-part.overlay.toml
liunan-ms: This only rewrites version-string macros (`specversion` / `tarfile_release`) so the spec tracks the AZL kernel's versioning. Moved from `azl-platform-adaptation` to `category = "azl-release-management"` with `upstream-status = "inapplicable"` (permanent AZL-only versioning). I kept `azl-platform-adaptation` only on `0005-build-generate-native-cross`, which genuinely detects native arch and cross-generates headers.

base/comps/kernel-headers/overlays/0004-6-18-31-1.overlay.toml
liunan-ms: Agree, this is not arch-specific. This substitutes Fedora's pre-built-headers tarball for the full AZL kernel source tree which should be a Fedora -> AZL source/identity choice. Recategorized to `category = "azl-branding-policy"`, `upstream-status = "inapplicable"` (Fedora ships their own tarball and would never take this).

base/comps/libspf2/overlays/0001-disable-short-names-override.overlay.toml
liunan-ms: We don't currently have a dedicated category for "fixing an issue that also exists in Fedora." Conceptually our overlays split into two kinds:

1. Fixes for common issues that also affect Fedora via backport or a local fix. The local fix is either a temporary workaround we drop once Fedora fixes it upstream, or something we intend to upstream to Fedora.
2. Fixes for AZL-specific issues that can't be upstreamed (vendor/triple/branding, AZL-only versioning, etc.).

This overlay is clearly kind (1), but our current category taxonomy is organized around the nature of the change (branding-policy, release-management, platform-adaptation, temp-workaround, pruning, compatibility) rather than where the problem originates, so "Fedora-affecting" isn't directly expressible as a category today.

Rather than add a new category, I think "does this also affect Fedora?" is better captured by upstream-status + commits, which we can already filter on:

- `upstream-status = "upstreamable"`: a Fedora-affecting fix we haven't sent up yet (candidate to push to Fedora).
- `upstream-status = "upstreamed"` with a Fedora `commits` ref: already fixed in a newer Fedora.

To filter out the Fedora issues that we're fixing, we can query for everything with `upstream-status = "upstreamable"`and no `commits` URL pointing at any upstream.

base/comps/llvm/overlays/0001-gcc-toolchain-triple-clang.overlay.toml
liunan-ms: This swaps the hardcoded redhat GCC triple for `%{_vendor}` so clang20 finds the GCC runtime in stage2 builds. Recategorized to `category = "azl-branding-policy"` with `upstream-status = "upstreamable"`.

base/comps/llvm20/overlays/0001-gcc-toolchain-triple-clang20.overlay.toml
liunan-ms: Same as above.

base/comps/rocm-compilersupport/overlays/0002-drop-llvm-triple-cfg.overlay.toml
liunan-ms: Set `category = "azl-branding-policy"` (it's driven by `%_vendor = azurelinux`), `upstream-status = "needs-upstream-hook"`, indicating the current form (dropping <triple>.cfg files in three build locations) isn't directly upstreamable, but it flags that we'd like an upstream hook so we don't carry this forever.

base/comps/rust/overlays/0002-s-compiler-rt-runtime.overlay.toml
liunan-ms: This uses `%{_target_platform}` instead of the hardcoded `%{_arch}-redhat-linux-gnu` for the compiler-rt runtime path. Recategorized to `category = "azl-branding-policy"` with `upstream-status = "upstreamable"`.

base/comps/shim/overlays/0001-release-management-change.overlay.toml
liunan-ms: Now captured explicitly: `category = "azl-temp-workaround"`, `upstream-status = "inapplicable"`, meant to be removed once the underlying release-management issue is resolved.

base/comps/mock-core-configs/overlays/0002-add-4-0-mock.overlay.toml
liunan-ms: Confirmed that the AZL4 configs are already upstream to Fedora. Set `upstream-status = "upstreamed"` with both commits, and a TODO: remove once pinning to f44. Changed to `category = "upstream-backport"` since the new mock-core-configs version is already in f44, at which point it drops entirely.

tobiasb-ms (approved):
Spot-checked everything, and generated before/after component list json files. Validated with Copilot that the only differences are the metadata being added.

PR #17898, opened 2026-07-02
chore(swift-lang and antlr4-project): Remove swift-lang

tobiasb-ms:
Swift-lang has its own clang which doesn't use system-clang's configuration files, which means it doesn't pick up our gcc triples. Because of this, it now won't build and anything that needs it to build fails with the existing version.

The only component we have that requires `swift-lang` is `antlr4-project`, and swift is option. It's used to build the subpackage `swift-antlr4-runtime`, but nothing in AZL4 depends on that.

So instead of fixing `swift-lang`, turn off `swift` in `antlr4-project` and stop building `swift-lang` altogether.


base/comps/antlr4-project/antlr4-project.comp.toml
```diff
@@ -0,0 +1,7 @@
+# The swift bcond defaults ON in Fedora builds (`%bcond swift %{undefined rhel}`
```
reubeno: I'd recommend revising this comment. It turns out that the straw that broke the camel's back is that swift-lang doesn't build, yes, but the bigger reason to drop swift is just that we really don't need it.

On a more minor note, I don't think individual `build.without` / `build.with` usages need to take on the burden of explaining how they work (i.e., that they toggle bcond). AI tends to like over-explaining some of these mechanisms, but we should assume that someone working in this code base should either (a) understand what `build.without` is or (b) be able to look up the docs explaining what it is and how it works. And if we don't think the docs are good enough, let's improve them.

reubeno (changes requested):

reubeno:
One comment, otherwise the change looks good to me.

base/comps/antlr4-project/antlr4-project.comp.toml
tobiasb-ms: Fixed the comment to reflect that we're removing this package because we don't need it. Also fixed commit messages correspondingly.

Note that I'm not sold the original comment really overexplained, but I do see your point about a more terse comment (which is generally my goto anyway). I tightened it up a lot.

reubeno (approved):

PR #17900, opened 2026-07-03
fix(azurelinux-release): ship azure kvp config in cloud identity package

miz060:
Add the cloud-init Hyper-V KVP telemetry reporting config used by Azure guest VMs:

`/etc/cloud/cloud.cfg.d/10-azure-kvp.cfg`

The config will be shipped by the existing `azurelinux-release-identity-cloud` subpackage, alongside the other Cloud VM variant-specific configuration such as the Azure `chrony`, `systemd-networkd`, and `sshd` defaults.

Validation:
  - azldev local build and install passed
  - control Tower build: passed

Fixes: [AB#22287](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22287)

### DIFF
--- a/base/comps/azurelinux-release/azurelinux-release.spec
+++ b/base/comps/azurelinux-release/azurelinux-release.spec
@@ -36,7 +36,7 @@ Summary:        Azure Linux release files
 Name:           azurelinux-release
 Version:        4.0
 # TODO(azl): Review whether we can move back to autorelease (with conditional -p)
-Release:        19%{?dist}
+Release:        20%{?dist}
 License:        MIT
 URL:            https://aka.ms/azurelinux
 
@@ -55,6 +55,7 @@ Source21:       50-azure-cloud.conf
 Source22:       70-azurelinux-hardening.conf
 Source23:       50-client-alive-interval.conf
 Source24:       50-permit-root-login.conf
+Source25:       10-azure-kvp.cfg
 
 BuildArch:      noarch
 
@@ -342,6 +343,7 @@ install -Dm0644 %{SOURCE17} -t %{buildroot}%{_prefix}/lib/sysctl.d/
 install -Dm0644 %{SOURCE20} -t %{buildroot}%{_sysconfdir}/chrony.d/
 install -Dm0644 %{SOURCE21} -t %{buildroot}%{_prefix}/lib/systemd/networkd.conf.d/
 install -Dm0600 %{SOURCE23} -t %{buildroot}%{_sysconfdir}/ssh/sshd_config.d/
+install -Dm0644 %{SOURCE25} -t %{buildroot}%{_sysconfdir}/cloud/cloud.cfg.d/
 %endif
 
 %if %{with container}
@@ -454,6 +456,7 @@ install -Dm0644 %{SOURCE22} -t %{buildroot}%{_sysctldir}/
 %{_sysconfdir}/chrony.d/chrony-azure.conf
 %{_prefix}/lib/systemd/networkd.conf.d/50-azure-cloud.conf
 %{_sysconfdir}/ssh/sshd_config.d/50-client-alive-interval.conf
+%config(noreplace) %{_sysconfdir}/cloud/cloud.cfg.d/10-azure-kvp.cfg
 %endif
 
 
@@ -476,6 +479,9 @@ install -Dm0644 %{SOURCE22} -t %{buildroot}%{_sysctldir}/
 
 
 %changelog
+* Wed Jul 08 2026 Mitch Zhu <mitchzhu@microsoft.com> - 4.0-20
+- Add cloud-init Azure KVP telemetry config to the cloud identity package
+
 * Tue Jul 07 2026 Brian Fjeldstad <bfjelds@microsoft.com> - 4.0-19
 - Enable netplan-configure.service by default via 90-default.preset
 
--- a/base/comps/azurelinux-release/10-azure-kvp.cfg
+++ b/base/comps/azurelinux-release/10-azure-kvp.cfg
@@ -1,0 +1,5 @@
+reporting:
+  logging:
+    type: log
+  telemetry:
+    type: hyperv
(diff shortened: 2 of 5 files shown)


base/comps/cloud-init/cloud-init.comp.toml
```diff
@@ -1,5 +1,59 @@
 [components.cloud-init]
 
+[[components.cloud-init.overlays]]
```
reubeno: If this is primarily layering additional configuration -- does this _need_ to be done as part of the base cloud-init package?

In general, our preference is to prefer layered configuration that can be carried in an Azure Linux specific spec. For example, if you look at the azurelinux-release subpackages you'll see various Azure specific configuration files. Overlays that mutate upstream specs should be considered a "last resort" that we employ if we can't use other existing mechanisms to extend/configure the system.

miz060: Got it. I'll re-evaluate based on this direction

base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -225,6 +226,12 @@ Provides the necessary files for a Azure Linux installation that is identifying
 itself as Azure Linux WSL.
 %endif
 
+%package -n cloud-init-azure-kvp
+Summary:        Cloud-init configuration for Hyper-V telemetry
+
+%description -n cloud-init-azure-kvp
+Cloud-init configuration for Hyper-V telemetry on Azure guests.
```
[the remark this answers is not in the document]
miz060: Fixed as suggested

base/comps/cloud-init/cloud-init.comp.toml
miz060: Refactored to be a package produced from `azurelinux-release` that depends on `cloud-init`.

base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -225,6 +226,13 @@ Provides the necessary files for a Azure Linux installation that is identifying
 itself as Azure Linux WSL.
 %endif
 
+%package -n cloud-init-azure-kvp
```
christopherco: **issue:** Does this kvp configuration file warrant its own packaging? The package only supplies a configuration file specific for Azure Guest VMs. Is there a situation where we'd ship an Azure Guest VM without this configuration file present?

If no, we should include the `10-azure-kvp.cfg` file in the `azurelinux-release-cloud` subpackage that we already supply, since that subpackage is meant to provide Azure-specific VM configurations (like the Azure-specific `chrony` configuration).

christopherco (changes requested):

base/comps/azurelinux-release/azurelinux-release.spec
miz060: Good point. I confirmed that this configuration is required for all AzL4 cloud images (consistent with the existing AzL3 behavior) so it doesn't need a separate package.

I've updated the PR to ship 10-azure-kvp.cfg from `azurelinux-release-cloud`.

base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -447,6 +449,10 @@ install -Dm0644 %{SOURCE22} -t %{buildroot}%{_sysctldir}/
 
 %if %{with cloud}
 %files cloud
+%dir %{_sysconfdir}/cloud
```
christopherco: **issue:** `cloud-init` package owns these `%{_sysconfdir}/cloud` and `%{_sysconfdir}/cloud/cloud.cfg.d` directories. https://github.com/microsoft/azurelinux/blob/4.0/specs/c/cloud-init/cloud-init.spec#L178

Don't assert ownership of these directories here. We should be able to safely assume that the cloud image variant has `cloud-init` on the system, similar to the chrony, systemd, and sshd config files applied in `azurelinux-release-identity-cloud`.

base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -447,6 +449,10 @@ install -Dm0644 %{SOURCE22} -t %{buildroot}%{_sysctldir}/
...
+%dir %{_sysconfdir}/cloud/cloud.cfg.d
+%config(noreplace) %{_sysconfdir}/cloud/cloud.cfg.d/10-azure-kvp.cfg
```
christopherco: **issue:** This should be in `azurelinux-release-identity-cloud` instead. Sorry about mentioning to use `azureliunux-release-cloud` in my earlier comment. `azurelinux-release-identity-cloud` is where all the other Cloud VM configuration files are, and the `azurelinux-release-identity-<variant>` ensures only one and only one image variant is ever supplied in an image. `azurelinux-release-cloud` is meant to be an empty metapackage, and we probably should keep it that way.

christopherco (changes requested):

base/comps/azurelinux-release/azurelinux-release.spec
```diff
@@ -454,6 +456,7 @@ install -Dm0644 %{SOURCE22} -t %{buildroot}%{_sysctldir}/
 %{_sysconfdir}/chrony.d/chrony-azure.conf
 %{_prefix}/lib/systemd/networkd.conf.d/50-azure-cloud.conf
 %{_sysconfdir}/ssh/sshd_config.d/50-client-alive-interval.conf
+%config(noreplace) %{_sysconfdir}/cloud/cloud.cfg.d/10-azure-kvp.cfg
```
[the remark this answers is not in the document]
miz060: this is intentional behavior, for reference check above comments.

base/comps/azurelinux-release/azurelinux-release.spec
miz060: Got it. Ownership setting code removed.

base/comps/azurelinux-release/azurelinux-release.spec
miz060: Thanks for the info! Updated to ship azure kvp config via `azurelinux-release-identity-cloud` as suggested.

christopherco (changes requested):
Changes themselves look good. Please rebase your commits together into one commit with proper commit message.

christopherco (approved):

PR #17930, opened 2026-07-07
fix(vm-iso-installer): download remote kickstart files locally

### LINKED ISSUE #17915: [4.0] Offline installer does not handle HTTP URLs in inst.ks kernel parameter
**Azure Linux version**
Azure Linux 4.0

**Deployment / platform**
KVM VM (PXE network boot)

**Is your feature request related to a problem? Please describe.**
When PXE booting with `inst.ks=http://server/path/ks.cfg `on the kernel cmdline, the installer wrapper (`anaconda-launcher.sh`) passes the URL directly to `anaconda --text --kickstart="http://server/path/ks.cfg"`. Anaconda can't fetch URLs when called this way and fails with "Kickstart file ... is missing."
The network is working fine at that point. Curl can download the file from the same shell. The problem is just that anaconda expects a local file path when called as a subprocess.

**Describe the solution you'd like**
Have `anaconda-launcher.sh` download the URL to a local file before passing it to anaconda:
```
if [[ "$CUSTOM_KS" == http* ]]; then
    curl -sf -o /tmp/ks.cfg "$CUSTOM_KS" && CUSTOM_KS="/tmp/ks.cfg"
fi
```

**Describe alternatives you've considered**
Patching the squashfs after ISO download to inject this logic. Works but adds complexity to the deployment pipeline.

**Additional context**
This affects any PXE based deployment that passes a kickstart URL via kernel cmdline. The fix is a few lines in `anaconda-launcher.sh`.

sonarbhavarth:
When PXE booting with an HTTP kickstart URL in the kernel cmdline (inst.ks),
anaconda-launcher.sh passed the URL directly to anaconda as a subprocess.
Anaconda requires a local file path in this context and would fail with a
missing file error.

This commit updates anaconda-launcher.sh to download HTTP/HTTPS URLs
via curl to /tmp/ks.cfg before passing it to anaconda, resolving
deployments that rely on remote kickstarts.

Fixes #17915

### DIFF
--- a/base/images/vm-iso-installer/anaconda-launcher.sh
+++ b/base/images/vm-iso-installer/anaconda-launcher.sh
@@ -14,6 +14,15 @@ done
 CUSTOM_KS=""
 if grep -qo 'inst\.ks=[^ ]*' /proc/cmdline 2>/dev/null; then
     CUSTOM_KS=$(grep -o 'inst\.ks=[^ ]*' /proc/cmdline | sed 's/inst\.ks=//')
+
+    if [[ "$CUSTOM_KS" == http* ]]; then
+        if ! curl -sf -o /tmp/ks.cfg "$CUSTOM_KS"; then
+            echo "ERROR: Failed to download kickstart from $CUSTOM_KS" >&2
+            exec /bin/bash
+        fi
+        CUSTOM_KS="/tmp/ks.cfg"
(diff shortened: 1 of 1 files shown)


ddstreetmicrosoft:
@kgodara912 i think this needs your review

PR #17956, opened 2026-07-09
test: add openmpi container test

bhagyapathak:
Add Openmpi container test

1. test_openmpi_mpirun_version
         Validates that OpenMPI runtime is installed and usable.
         Runs mpirun --version with required OpenMPI runtime env configuration.
         Fails fast if basic OpenMPI runtime wiring is broken.
2. test_openmpi_runs_multiple_ranks
       Validates that mpirun actually launches two MPI ranks in the container.
       Executes a two-rank command and checks exact rank output: rank=0 and rank=1.
       Provides stable rank-launch verification without depending on version-string formatting.
3. test_openmpi_send_receive_between_two_ranks
       Validates real MPI communication semantics, not just process launch.
       Compiles the checked-in C program copied into the image, then runs it with two ranks.
       Asserts tagged message exchange output, proving rank-to-rank send and receive behavior works end to end.

### DIFF
--- a/base/images/tests/cases/runtime/container-base/test_openmpi/send_receive.c
+++ b/base/images/tests/cases/runtime/container-base/test_openmpi/send_receive.c
@@ -1,0 +1,45 @@
+#include <mpi.h>
+#include <stdio.h>
+#include <string.h>
+
+int main(int argc, char *argv[]) {
+    int rank, source, dest;
+    MPI_Request requests[4];
+    const int tag1 = 6;
+    const int tag2 = 10;
+    char inmsg[64];
+    const char outmsg0[] = "Using Tag1";
+    const char outmsg1[] = "Again Using Tag1";
+    const char outmsg2[] = "Using Tag2";
+    const char outmsg3[] = "Again Using Tag2";
+
+    MPI_Init(&argc, &argv);
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+    if (rank == 0) {
+        dest = 1;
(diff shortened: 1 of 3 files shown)


bhagyapathak:
Test Logs [openmpi_container_runtime_test_logs.txt](https://github.com/user-attachments/files/29840170/openmpi_container_runtime_test_logs.txt)

base/images/tests/cases/runtime/container-base/test_openmpi/send_receive.c
```diff
@@ -0,0 +1,43 @@
...
+    const char outmsg1[] = "Again Using Tag1";
+    const char outmsg2[] = "Using Tag2";
+    const char outmsg3[] = "Again Using Tag2";
+
+    MPI_Init(&argc, &argv);
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+    if (rank == 0) {
+        dest = 1;
+        MPI_Send(outmsg0, (int)strlen(outmsg0) + 1, MPI_CHAR, dest, tag1, MPI_COMM_WORLD);
```
christopherco: **issue:** A deadlock can occur if rank 0's first `MPI_Send()` with `tag1` blocks waiting for a matching receive. Rank 0 then never sends `tag2`, while rank 1 is already blocked waiting to receive a `tag2` message
 
Please either receive messages in send order or use nonblocking sends, like 4 `MPI_Isend()`s followed by `MPI_Waitall()`.

christopherco (changes requested):

christopherco (approved):

PR #17972, opened 2026-07-09
feat: Enable libsolv to expose file time to dnf plugins

mbearup:
To achieve parity with Azure Linux 3, we want to support virtual snapshots (package selection based on file time).
For Azure Linux 4, we want to pursue this as a dnf plugin (vs a direct modification to dnf).
Libsolv does not expose the package file time to dnf plugins, so this PR is a prerequisite in enabling that support.

Fixes: [AB#22351](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22351)


specs/l/libsolv/libsolv.spec
```diff
@@ -89,6 +89,7 @@ BuildRequires:  libzstd-devel
 BuildRequires:  pkgconfig(zck)
 %endif
 
+Patch1: 0002-Make-file-time-available-for-dnf-plugins.patch
```
christopherco, line 92: **question (non-blocking):** @liunan-ms @reubeno @dmcilvaney Is this the correct location in the spec file to expect the new patch file inclusion after running through `azldev`? L70 has a different patch file provided from upstream, and I would have thought this patch would be next to L70, hence why I'm asking.

Non-blocking since I did confirm the patch file is applied during the RPM build process.

christopherco (approved):

specs/l/libsolv/libsolv.spec
liunan-ms: @christopherco  This is expected behavior, though I agree the placement isn't ideal.

The `patch-add` overlay registers the patch and checks if there's no %patchlist section, if not, it just appends the new `PatchN:` tag at the end of the preamble section (just before `%description`), rather than grouping it with the existing patch on L70. That's why it lands on L92. 

Although it works, it's against Fedora/RPM convention and we should group it with the other Patch. I filed an issue in azldev to fix it. https://github.com/microsoft/azure-linux-dev-tools/issues/284

christopherco: Thanks @liunan-ms !

PR #18000, opened 2026-07-13
test: add telegraf container test

bhagyapathak:


bhagyapathak:
[telegraf_container_runtime_test_logs.txt](https://github.com/user-attachments/files/29959635/telegraf_container_runtime_test_logs.txt)

base/images/tests/cases/runtime/container-base/test_telegraf/configs/telegraf.conf
```diff
@@ -0,0 +1,21 @@
...
+  hostname = ""
+  omit_hostname = false
+
+[[inputs.cpu]]
+  percpu = true
+  totalcpu = true
+  collect_cpu_time = false
+  report_active = false
+
+[[inputs.mem]]
```
christopherco: **issue:** please ensure newly checked-in `.conf` files end with a trailing newline.
Even where not strictly required (possibly telegraf), trailing newlines are standard Unix practice and will prevent parser/tooling edge cases. We have previously encountered functional impact when EOF newline was missing from a `.conf` file (in the previous case, it was `nftables.conf`)

christopherco (changes requested):

christopherco (approved):

PR #18001, opened 2026-07-13
test: Add Database Tests For Container Images - Postgres, Redis, Valkey, Memcached

MadhurAggarwal:
Commit [1](https://github.com/microsoft/azurelinux/pull/18001/changes/747f7c95547c68176fcad0e2fb5dfcac376a6c84) - Added Container Runtime Tests for 4 DataStores:
- Valkey
- Redis (available as valkey-compat-redis)
- Postgres
- Memcached

Also Added a new fixture: `client_server_shell_exec`:
- Creates 2 Contains, named `client` and `server` respectively
- Connects them to same network
- Returns their respective shell exec callable, and server container name for dns 

Each added test contains:
- Docker file to install the corresponding package
- Test file, which starts the server for each datastore, and contains 2 tests - Version Test to check the sanity / installation, and a Core Functionality Test (Create-Insert-Select for Postgres, LPUSH for Redis/Valkey, and Set-Get for Memcached)
- Also Included Cross Container Tests for Postgres, Redis, Valkey which use the new fixture to test for client-server connections.

PR Testing:
Locally built a container images and tested using azldev tool:

- Memcached Tests
image

- Postgres Tests
image

- Redis Tests
image

- Valkey Tests
image

Commit [2](https://github.com/microsoft/azurelinux/pull/18001/changes/b7ab464fad52e714e6a2173d1c7458d2227c3105) - Fixed Pyright Errors:
- Untyped Fixture Param (`reportUnknownParameterType`, `reportMissingParameterType`)
- Incorrect Param / Return Type (`reportArgumentType`, `reportReturnType`, `reportInvalidTypeForm`, `reportCallIssue`)
- Deprecated Access Method (`reportAttributeAccessIssue`)

Previous Pyright Errors (Grouped):
image

Fixed All Pyright Errors in latest commit:
image

### DIFF
--- a/base/images/tests/README.md
+++ b/base/images/tests/README.md
@@ -167,6 +167,7 @@ base/images/
 | `container_exec` | function | callable | `(args) → ContainerExecResult` |
 | `wait_for_http` | function | callable | `(url, *, retries=5, delay=1.0, connect_timeout=2.0, max_time=5.0) → ContainerExecResult` — polls an in-container HTTP endpoint with `curl`; raises after retries |
 | `assert_http_server` | function | callable | `(start_command, url, expected, *, retries=5, delay=1.0) → ContainerExecResult` — starts a server, waits for `url`, asserts `expected` in body |
+| `client_server_exec_shell` | function | tuple | Networked server and client containers for cross-container tests |
 
 ## Adding tests
 
@@ -236,9 +237,10 @@ Dockerfile only trigger one build.
 > `container_exec_shell("nginx")`. This keeps behaviour predictable and
 > ensures each test controls exactly what runs.
 
-For service tests, poll readiness with a short bounded loop before asserting on
-responses; do not assume the service binds synchronously. Foreground services
-should be backgrounded explicitly, for example
+For service tests, poll readiness before asserting on responses; do not assume the
+service binds synchronously — use `wait_until_service_ready(exec_shell, probe_cmd,
+contains=...)` from `utils.container_runtime`. Foreground services should be
+backgrounded explicitly, for example
 `container_exec_shell("nohup my-service > /tmp/my-service.log 2>&1 &")`.
 
 ## Adding a native-tool dependency
--- a/base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
+++ b/base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
@@ -1,0 +1,44 @@
+# SPDX-License-Identifier: MIT
+"""Validate the Memcached server on the container-base image."""
+
+from __future__ import annotations
+
+import pytest
+from utils.container_runtime import wait_until_service_ready
+
+PORT = 11211
+
+
+def _start_memcached(container_exec_shell) -> None:
+    """Start memcached daemonized and wait until it accepts connections."""
+    # Run as the memcached user, since the server cannot run as root.
+    start = container_exec_shell(f"memcached -u memcached -d -p {PORT}")
+    assert start.exit_code == 0, f"memcached failed to start: {start.output}"
+
+    wait_until_service_ready(
+        container_exec_shell,
+        f"printf 'version\\r\\nquit\\r\\n' | nc localhost {PORT}",
+        contains="VERSION",
+    )
+
+
+@pytest.mark.dockerfile()
+def test_memcached_version(container_exec_shell) -> None:
+    """The Memcached server binary reports a version."""
+    result = container_exec_shell("memcached --version")
+    assert result.exit_code == 0, f"memcached --version failed: {result.output}"
+    assert "memcached" in result.output
+
+
+@pytest.mark.dockerfile()
+def test_memcached_set_get(container_exec_shell) -> None:
+    """Store a value with memcached and read it back."""
+    _start_memcached(container_exec_shell)
+    # Text protocol: "set <key> <flags> <exptime> <bytes>" + the value, then get it back.
+    result = container_exec_shell(
+        f"printf 'set TestValue 0 100 4\\r\\nTest\\r\\nget TestValue\\r\\nquit\\r\\n' | nc localhost {PORT}",
+    )
+    assert result.exit_code == 0, f"nc failed: {result.output}"
+    assert "STORED" in result.output, f"set not acknowledged: {result.output}"
+    body = result.output.replace("\r\n", "\n")
+    assert "VALUE TestValue 0 4\nTest\nEND" in body, f"unexpected get response: {result.output}"
--- a/base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
+++ b/base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
@@ -4,12 +4,12 @@
 from __future__ import annotations
 
 import pytest
-from utils.container_runtime import wait_until_service_ready
+from utils.container_runtime import ExecShell, wait_until_service_ready
 
 PORT = 11211
 
 
-def _start_memcached(container_exec_shell) -> None:
+def _start_memcached(container_exec_shell: ExecShell) -> None:
     """Start memcached daemonized and wait until it accepts connections."""
     # Run as the memcached user, since the server cannot run as root.
     start = container_exec_shell(f"memcached -u memcached -d -p {PORT}")
@@ -23,7 +23,7 @@ def _start_memcached(container_exec_shell) -> None:
 
 
 @pytest.mark.dockerfile()
-def test_memcached_version(container_exec_shell) -> None:
+def test_memcached_version(container_exec_shell: ExecShell) -> None:
     """The Memcached server binary reports a version."""
     result = container_exec_shell("memcached --version")
     assert result.exit_code == 0, f"memcached --version failed: {result.output}"
@@ -31,7 +31,7 @@ def test_memcached_version(container_exec_shell) -> None:
 
 
 @pytest.mark.dockerfile()
-def test_memcached_set_get(container_exec_shell) -> None:
+def test_memcached_set_get(container_exec_shell: ExecShell) -> None:
     """Store a value with memcached and read it back."""
     _start_memcached(container_exec_shell)
     # Text protocol: "set <key> <flags> <exptime> <bytes>" + the value, then get it back.
--- a/base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
+++ b/base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
@@ -1,0 +1,122 @@
+# SPDX-License-Identifier: MIT
+"""Validate the PostgreSQL server on the container-base image."""
+
+from __future__ import annotations
+
+import contextlib
+
+import pytest
+from utils.container_runtime import wait_until_service_ready
+
+DB_PORT = 5432
+DB_NAME = "postgres"
+DB_USER = "postgres"
+DB_PASSWORD = "secret"  # noqa: S105 - throwaway password for the test container
+EXPECTED_ROWS = 2
+
+PG_DIR = "/var/lib/pgsql"
+PG_DATADIR = f"{PG_DIR}/data"
+PG_LOGFILE = f"{PG_DIR}/logfile"
+PG_RUNDIR = "/run/postgresql"
+
+
+@contextlib.contextmanager
+def _postgres_password_file(container_exec_shell, password: str):
+    """Yield a postgres-owned temp password file; remove it on exit."""
+    mk = container_exec_shell("mktemp")
+    assert mk.exit_code == 0, f"mktemp failed: {mk.output}"
+    filepath = mk.stdout.strip()
+    try:
+        setup = container_exec_shell(f"printf %s {password} > {filepath} && chown postgres {filepath}")
+        assert setup.exit_code == 0, f"password file setup failed: {setup.output}"
+        yield filepath
+    finally:
+        container_exec_shell(f"runuser -u postgres rm -f {filepath}")
+
+
+def _start_postgresql(container_exec_shell, *, listen_all: bool = False) -> None:
+    """Init the cluster and start the server; listen_all also opens it to remote hosts."""
+    socket = container_exec_shell(
+        f"mkdir -p {PG_RUNDIR} && chown postgres:postgres {PG_RUNDIR}",
+    )
+    assert socket.exit_code == 0, f"socket dir setup failed: {socket.output}"
+
+    # initdb and the server cannot run as root.
+    with _postgres_password_file(container_exec_shell, DB_PASSWORD) as pwfile_path:
+        initdb = container_exec_shell(
+            "runuser -l postgres -c '"
+            "initdb --locale=C --encoding=UTF8 --auth-local=peer "
+            f"--auth-host=scram-sha-256 --pwfile={pwfile_path} -D {PG_DATADIR}'",
+        )
+        assert initdb.exit_code == 0, f"initdb failed: {initdb.output}"
+
+    if listen_all:
+        # Open the server to remote TCP; it listens on localhost only by default.
+        config = container_exec_shell(
+            f"echo \"listen_addresses = '*'\" >> {PG_DATADIR}/postgresql.conf && "
+            f"echo 'host all all 0.0.0.0/0 scram-sha-256' >> {PG_DATADIR}/pg_hba.conf",
+        )
+        assert config.exit_code == 0, f"network config setup failed: {config.output}"
+
+    start = container_exec_shell(
+        f"runuser -l postgres -c 'pg_ctl -D {PG_DATADIR} -l {PG_LOGFILE} -w start'",
+    )
+    assert start.exit_code == 0, f"pg_ctl start failed: {start.output}"
+
+    wait_until_service_ready(container_exec_shell, f"pg_isready -h localhost -p {DB_PORT}")
+
+
+def _run_crud_workflow(exec_shell, host: str) -> None:
+    """Create a table, insert two rows, and read them back against the DB at ``host``."""
+    psql = f"PGPASSWORD={DB_PASSWORD} psql -h {host} -p {DB_PORT} -U {DB_USER} -d {DB_NAME}"
+
+    create = exec_shell(f'{psql} -c "CREATE TABLE cities (name varchar(80), location point);"')
+    assert create.exit_code == 0, f"create failed: {create.output}"
+    assert "CREATE TABLE" in create.output
+
+    insert_sql = (
+        "INSERT INTO cities VALUES ('San Francisco', '(-194.0, 53.0)'); "
+        "INSERT INTO cities VALUES ('Seattle', '(-150.0, 86.0)');"
+    )
+    insert = exec_shell(f'{psql} -c "{insert_sql}"')
+    assert insert.exit_code == 0, f"insert failed: {insert.output}"
+    assert insert.output.count("INSERT 0 1") == EXPECTED_ROWS, f"expected two inserts: {insert.output}"
+
+    select = exec_shell(f'{psql} -c "SELECT * FROM cities;"')
+    assert select.exit_code == 0, f"select failed: {select.output}"
+    assert f"{EXPECTED_ROWS} rows" in select.output, f"expected {EXPECTED_ROWS} rows: {select.output}"
+
+
+def _assert_bad_auth_rejected(exec_shell, host: str) -> None:
+    """A wrong password must be rejected, proving the scram rule is enforced (not trust)."""
+    bad_auth = exec_shell(
+        f"PGPASSWORD=wrong psql -h {host} -p {DB_PORT} -U {DB_USER} -d {DB_NAME} -c 'SELECT 1;'",
+    )
+    assert bad_auth.exit_code != 0, f"wrong password was accepted: {bad_auth.output}"
+    assert "authentication failed" in bad_auth.output, f"unexpected auth error: {bad_auth.output}"
+
+
+@pytest.mark.dockerfile()
+def test_postgresql_version(container_exec_shell) -> None:
+    """The PostgreSQL server binary reports a version."""
+    result = container_exec_shell("postgres --version")
+    assert result.exit_code == 0, f"postgres --version failed: {result.output}"
+    assert "postgres (PostgreSQL)" in result.output
+
+
+@pytest.mark.dockerfile()
+def test_postgresql_database_server(container_exec_shell) -> None:
+    """Server accepts TCP connections and handles create/insert/select."""
+    _start_postgresql(container_exec_shell)
+    _assert_bad_auth_rejected(container_exec_shell, "localhost")
+    _run_crud_workflow(container_exec_shell, "localhost")
+
+
+@pytest.mark.dockerfile()
+def test_postgresql_cross_container(client_server_exec_shell) -> None:
+    """A client container reaches a server container's database over the network."""
+    server_exec, client_exec, server_host = client_server_exec_shell
+
+    _start_postgresql(server_exec, listen_all=True)
+    _assert_bad_auth_rejected(client_exec, server_host)
+    _run_crud_workflow(client_exec, server_host)
--- a/base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
+++ b/base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
@@ -4,9 +4,10 @@
 from __future__ import annotations
 
 import contextlib
+from collections.abc import Iterator
 
 import pytest
-from utils.container_runtime import wait_until_service_ready
+from utils.container_runtime import ExecShell, wait_until_service_ready
 
 DB_PORT = 5432
 DB_NAME = "postgres"
@@ -21,7 +22,7 @@ PG_RUNDIR = "/run/postgresql"
 
 
 @contextlib.contextmanager
-def _postgres_password_file(container_exec_shell, password: str):
+def _postgres_password_file(container_exec_shell: ExecShell, password: str) -> Iterator[str]:
     """Yield a postgres-owned temp password file; remove it on exit."""
     mk = container_exec_shell("mktemp")
     assert mk.exit_code == 0, f"mktemp failed: {mk.output}"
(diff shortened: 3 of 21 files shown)


base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
```diff
@@ -0,0 +1,48 @@
...
+def test_memcached_set_get(container_exec_shell) -> None:
+    """Store a value with memcached and read it back."""
+    _start_memcached(container_exec_shell)
+    # Text protocol: "set <key> <flags> <exptime> <bytes>" + the value, then get it back.
+    result = container_exec_shell(
+        f"printf 'set TestValue 0 100 4\\r\\nTest\\r\\nget TestValue\\r\\nquit\\r\\n' | nc localhost {PORT}",
+    )
+    assert result.exit_code == 0, f"nc failed: {result.output}"
+    assert "STORED" in result.output
+    assert "VALUE TestValue 0 4" in result.output
```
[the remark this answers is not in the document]
MadhurAggarwal: Added this fix into the PR.

base/images/tests/cases/runtime/container-base/test_valkey/test_valkey.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import annotations
+
+import time
+
+import pytest
+
+EXPECTED_LEN = "4"
+
+
+def _start_valkey(container_exec_shell) -> None:
```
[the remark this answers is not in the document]
MadhurAggarwal: The existing (merged) Container Runtime Tests follow the same pattern. This comment can be ignored - the fixture is injected by pytest by name.

base/images/tests/cases/runtime/container-base/test_redis/test_redis.py
```diff
@@ -0,0 +1,44 @@
...
+def test_redis_version(container_exec_shell) -> None:
+    """The Redis server reports its version."""
+    _start_redis(container_exec_shell)
+    result = container_exec_shell("redis-cli INFO server")
+    assert result.exit_code == 0, f"INFO server failed: {result.output}"
+    assert "redis_version:" in result.output
+
+
+@pytest.mark.dockerfile()
+def test_redis_lpush(container_exec_shell) -> None:
```
bhagyapathak: Current 4.0 DB ports validate redis/valkey/postgres/memcached via localhost inside one container.
3.0 golden tests included network-path coverage (separate client/server containers or host->container path).
@christopherco Do we want to keep 4.0 scope as package/runtime smoke only, or should we add at least one cross-container connectivity test per DB for parity

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
```diff
@@ -0,0 +1,88 @@
...
+    container_exec_shell("rm -f /tmp/pgpw")
+
+    ready = None
+    for _ in range(10):
+        ready = container_exec_shell(f"pg_isready -h localhost -p {DB_PORT}")
+        if ready.exit_code == 0:
+            break
+        time.sleep(1)
+    else:
+        output = ready.output if ready is not None else "<no attempts>"
```
tobiasb-ms: question(non-blocking): When would `ready` be `None` here? We're guaranteed to do an at least one attempt right?

tobiasb-ms: I notice this in your other retry loops as well.

base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
```diff
@@ -0,0 +1,49 @@
...
+def _start_memcached(container_exec_shell) -> None:
+    """Start memcached daemonized and wait until it accepts connections."""
+    # Run as the memcached user, since the server cannot run as root.
+    start = container_exec_shell(f"memcached -u memcached -d -p {PORT}")
+    assert start.exit_code == 0, f"memcached failed to start: {start.output}"
+
+    ready = None
+    for _ in range(10):
+        ready = container_exec_shell(f"printf 'version\\r\\nquit\\r\\n' | nc localhost {PORT}")
+        if ready.exit_code == 0 and "VERSION" in ready.output:
```
tobiasb-ms: issue(blocking): If `exit_code` is 0 and `VERSION` is not in the output, am I correct that something has gone seriously wrong? It may be a test issue or something at that point.

It feels like this should be broken up somehow -- `exit_code` being 0 means "the command worked". That's when you stop the loop. But then you need to make sure the command worked the way you thought it would -- an assert or something.

That said, you know more about this than I do, so I can definitely believe there's a great reason for doing it this way.

tobiasb-ms: I notice this in the `redis` and `valkey` tests too.

base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
```diff
@@ -0,0 +1,49 @@
...
+            break
+        time.sleep(1)
```
tobiasb-ms: nit(non-blocking): You're sleeping an extra second here if it never works -- it will run the `sleep` and then immediately exit the loop. Non-blocking because this is tests.

Also happens in your other tests.

base/images/tests/cases/runtime/container-base/test_valkey/test_valkey.py
tobiasb-ms: suggestion(non-blocking): Since both of these tests need valkey, consider doing it as a `fixture`.

base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def _start_memcached(container_exec_shell) -> None:
+    """Start memcached daemonized and wait until it accepts connections."""
+    # Run as the memcached user, since the server cannot run as root.
+    start = container_exec_shell(f"memcached -u memcached -d -p {PORT}")
+    assert start.exit_code == 0, f"memcached failed to start: {start.output}"
+
+    ready = None
+    for _ in range(10):
```
tobiasb-ms: suggestion(blocking): All of the `_start_<thing>` functions have a code block that is the same:
```
for _ in range(10):
    ready = container_exec_shell(<some_string>)
    if ready.exit_code == 0 and <optional other test>:
       break
    time.sleep(1)
else:
    failure case -- varies only by a string
```

And I've pointed out a couple issues with this fundamental loop in a couple places.

Please use a common function that extracts this. It could go near `wait_for_http` or something.

tobiasb-ms (changes requested):

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
MadhurAggarwal: Hi Tobias,
You're right, the container exec runs atleast once so `ready` will never be `None` at runtime.
But, since it is initialized as `None`, trying to use `ready.output` directly adds a new (static) pyright error - 
`test_postgres.py:61:24 - error: "output" is not a known attribute of "None" (reportOptionalMemberAccess)`.

I have now changed the loop format into a single helper for all, where I have initialized `ready` as the output of the first call, hence the `is not None` check has been removed.

base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
MadhurAggarwal: Thanks for catching this. 
I have moved the sleep call to execute before the container exec call (except the first one) so there is no redundant wait anymore.

base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
MadhurAggarwal: Got it, I've moved this common code into a single helper function: `wait_until_service_ready` and called it from the 4 DB tests.

Note that, `wait_for_http` is defined as a fixture that takes in `container_exec_shell` fixture.

However, this new helper `wait_until_service_ready` cannot be defined as a fixture, since it has to conditionally accept either `server_exec_shell` (unpacked from the new `client_server_exec_shell` fixture) or `container_exec_shell` depending on whether we are running cross-container DB Network checks or single-container local-host service checks.

base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
MadhurAggarwal: For memcached in particular, we run `printf '<memcached version check>' | nc localhost {memcached PORT}`: 
(Reference: [3.0 implementation](https://dev.azure.com/mariner-org/mariner/_git/CBL-Mariner-Pipelines?path=/pipelines/test-golden-containers/test-source-artifacts/memcached/TestRunner.sh&version=GBmaster&line=43&lineEnd=44&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)) So, the exit code comes from nc (TCP Connection) sending request to memcached's port, and it can be 0 (correctly sent request) even if memcached is still not ready.

For Postgres, we run a directly available command: `pg_isready`, so when it exits with code 0, it is bound to be ready and no other output check is required.
For redis / valkey, we run `Ping` command, so exit code 0 only happens when there is Pong in the response. Assertion might be redundant for this case.

I'm open to further discussions on this if required, as of now I believe having assert for this in the common helper may be unnecessary.

MadhurAggarwal:
I have resolved the previous comments and made the following changes:
- Extracted the Polling helper into a general function `wait_until_service_ready` and used this across all 4 DB Tests.
- Created a new fixture: `client_server_exec_shell` which creates 2 containers connected by a network.
- Used the new fixture to run cross-container (network) tests on Postgres, Redis and Valkey. Note that, for Postgres, there is a DB Config Change required in order to allow these network tests to work, hence I've kept both local-host & cross-container test versions. For Redis & Valkey however, no such config change is required so only cross-container tests have been kept. (There is still a version check that runs on local-host server)
- Didn't yet add any cross-container tests for memcached since no such test was present for 3.0 as well. If this is required, do let me know.
- I've created a shared helper function (example: `_effective_image` and `_run_crud_workflow`) where-ever required to avoid code duplications.

I've tested these changes and updated the PR body with latest test run logs images.

base/images/tests/cases/runtime/container-base/test_redis/test_redis.py
MadhurAggarwal: I've created a fixture `client_server_exec_shell` and used it for cross-container tests in Postgres, Redis and Valkey.
I have not yet added any cross-container tests for memcached since the 3.0 implementation doesn't have it either. Should I add that as well?

base/images/tests/cases/runtime/container-base/test_valkey/test_valkey.py
MadhurAggarwal: I've kept it as is, as: I've added new changes for cross container tests, and now the 2 valkey tests require different fixtures (one needs the existing `container_exec_shell` and other needs `client-server-exec-shell`)

base/images/tests/README.md
```diff
@@ -226,9 +227,10 @@ Dockerfile only trigger one build.
 > `container_exec_shell("nginx")`. This keeps behaviour predictable and
 > ensures each test controls exactly what runs.
 
-For service tests, poll readiness with a short bounded loop before asserting on
-responses; do not assume the service binds synchronously. Foreground services
-should be backgrounded explicitly, for example
+For service tests, poll readiness before asserting on responses; do not assume the
+service binds synchronously — use `wait_until_service_ready(exec_shell, probe_cmd, *,
+contains=...)` from `utils.container_runtime`. Foreground services should be
```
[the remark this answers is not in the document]
MadhurAggarwal: fixed

MadhurAggarwal:
(rebased & resolved merge conflicts and copilot comments)

base/images/tests/cases/runtime/container-base/test_valkey/test_valkey.py
```diff
@@ -0,0 +1,40 @@
...
+import pytest
+from utils.container_runtime import wait_until_service_ready
+
+EXPECTED_LEN = "4"
+
+
+def _start_valkey(container_exec_shell) -> None:
+    """Start valkey-server daemonized and wait until it answers PING."""
+    start = container_exec_shell("valkey-server --protected-mode no --save 60 1 --daemonize yes")
+    assert start.exit_code == 0, f"valkey-server failed to start: {start.output}"
```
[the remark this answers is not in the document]
MadhurAggarwal: All existing runtime container tests use assert so ignoring this comment.

base/images/tests/cases/runtime/container-base/test_redis/test_redis.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import annotations
+
+import time
+
+import pytest
+
+EXPECTED_LEN = "4"
+
+
+def _start_redis(container_exec_shell) -> None:
```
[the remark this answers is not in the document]
tobiasb-ms: issue: (blocking-for-discusson): `container_exec_shell` is still untyped -- why is best to leave it that way?

NOTE: I can definitely believe there's a good reason, but we should either follow the guidelines we have or document why we're not.

NOTE: I'll use this conversation for all of the similar "untyped `container_exec_shell` conversations that copilot started.

base/images/tests/cases/runtime/container-base/test_memcached/test_memcached.py
tobiasb-ms: OK that makes sense for `memecached` and `nc`. It's tempting to factor that behavior in as a parameter to `wait_until_service_ready`, but I think you've found the right balance in the helper function.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
```diff
@@ -0,0 +1,106 @@
...
+def _start_postgresql(container_exec_shell, *, listen_all: bool = False) -> None:
+    """Init the cluster and start the server; listen_all also opens it to remote hosts."""
+    socket = container_exec_shell(
+        "mkdir -p /run/postgresql && chown postgres:postgres /run/postgresql",
+    )
+    assert socket.exit_code == 0, f"socket dir setup failed: {socket.output}"
+
+    # initdb requires a password file, readable by the postgres user
+    pwfile = container_exec_shell(
+        f"printf %s {DB_PASSWORD} > /tmp/pgpw && chown postgres /tmp/pgpw",
```
tobiasb-ms: nit (non-blocking): This is theoretically risky as `/tmp/pgpw` could exist. Using `mktemp` is a more robust, more "correct" way to do it. I'm not really sure it's necessary here though, which is why this is non-blocking.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    assert pwfile.exit_code == 0, f"password file setup failed: {pwfile.output}"
+
+    # Run as the postgres user, since initdb and the server cannot run as root.
+    initdb = container_exec_shell(
+        "runuser -l postgres -c '"
+        "initdb --locale=C --encoding=UTF8 --auth-local=peer "
+        f"--auth-host=scram-sha-256 --pwfile=/tmp/pgpw -D {PGDATA}'",
```
tobiasb-ms: issue(blocking): `/tmp/pgpw` shows up thrice in this function -- extract it into a constant (or variable if you end up using `mktemp`).

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
```diff
@@ -0,0 +1,106 @@
...
+    if listen_all:
+        # postgres listens on localhost only by default; '*' is needed for remote TCP (isolated test network).
+        config = container_exec_shell(
+            f"echo \"listen_addresses = '*'\" >> {PGDATA}/postgresql.conf && "
+            f"echo 'host all all 0.0.0.0/0 scram-sha-256' >> {PGDATA}/pg_hba.conf",
+        )
+        assert config.exit_code == 0, f"network config setup failed: {config.output}"
+
+    start = container_exec_shell(
+        f"runuser -l postgres -c 'pg_ctl -D {PGDATA} -l /var/lib/pgsql/logfile -w start'",
```
tobiasb-ms: nit(non-blocking): `/var/lib/pgsql/logfil` is very similar to the value of `PGDATA`. Consider some way of extracting the commonality while still being very clear what's happening.

NOTE: This is right on the cusp for me -- I'm not sure it's necessary but it "smells" bad.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    assert start.exit_code == 0, f"pg_ctl start failed: {start.output}"
+
+    container_exec_shell("rm -f /tmp/pgpw")
```
tobiasb-ms: issue(blocking): This will skipped on a failed test -- the assertion will raise an error and this will never be run. I see two options here:
1. Handle it correctly either via a context manager or try/finally (I prefer context manager).
2. Decide that as this is on a transient container, it doesn't really matter if it's cleaned up and simply do nothing.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
```diff
@@ -0,0 +1,106 @@
...
+
+@pytest.mark.dockerfile()
+def test_postgresql_cross_container(client_server_exec_shell) -> None:
+    """A client container reaches a server container's database over the network."""
+    server_exec, client_exec, server_host = client_server_exec_shell
+
+    _start_postgresql(server_exec, listen_all=True)
+
+    # Wrong password must be rejected, proving the scram rule is enforced (not trust).
+    bad_auth = client_exec(
```
tobiasb-ms: question(blocking-for-discussion): Is this not valuable in the localhost case? I notice you aren't doing it there.

base/images/tests/utils/container_runtime.py
```diff
@@ -308,6 +314,26 @@ def _make_exec_result(
...
+    """Poll ``command`` until it exits 0 and its output contains ``contains``."""
+    output = ""
+    for attempt in range(attempts):
+        if attempt:
+            time.sleep(delay)
+        result = exec_shell(command)
+        if result.exit_code == 0 and contains in result.output:
+            return result
+        output = result.output
+    raise AssertionError(f"{command!r} never became ready: {output}")
```
tobiasb-ms, line 372: nit(non-blocking): Would `pytest.fail` be better or more idiomatic here?

base/images/tests/conftest.py
```diff
@@ -420,6 +418,56 @@ def _write(path: str, content: str):
...
+        )
+        client = create_container(
+            podman_client, effective_image, f"azl-test-client-{suffix}", networks=[network_name],
+        )
+        yield _exec_shell_for(server), _exec_shell_for(client), server.container_name
+    finally:
+        try:
+            for container in (server, client):
+                if container is not None:
+                    destroy_container(podman_client, container.container_name)
```
tobiasb-ms, line 497: nit(non-blocking): Is the nested `try/finally` actually necessary? Can `destroy_container` fail under these circumstances? I could believe it, I just thought it was worth thinking about.

tobiasb-ms (commented):
issue(blocking): In addition to my code comments, most of the commits in this PR don't conform to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), which is required. Most likely you want to do a squash rebase anyway, but either way fix that please.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
MadhurAggarwal: The auth rule is configured as `scram-sha-256` for both localhost and cross container tests so checking its enforcement twice may be redundant. But I agree that for better coverage it is good to have in both cases, so I've updated this accordingly.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
MadhurAggarwal: Sure, I've updated it into this format:
```
PG_DIR     = "/var/lib/pgsql"
PG_DATADIR = f"{PG_DIR}/data"  
PG_LOGFILE = f"{PG_DIR}/logfile"
```
Hope this is better?

base/images/tests/conftest.py
MadhurAggarwal: Yes, I think it is necessary to have this in its own try-finally block.
Reasoning: 
- This network is created with podman and is persisted on the host machine. 
- As of now, this is the only fixture that creates a network. So, no network cleanup occurs apart from here. (even during session teardown)
- `destroy_container` raises for all exceptions other than `NoSuchContainer, DockerException`

So, if we move the network cleanup to the same try-finally block, then any other exception, for ex: a manual interruption with ctrl+c, during `destroy_container` stage would mean that the network is never cleaned and has to be manually done on the host machine.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
MadhurAggarwal: I've used context manager as suggested which removes the repetitions.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
MadhurAggarwal: I've added a context manager to handle the file creation & removal

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
MadhurAggarwal: It is not necessary for this case to use mktemp since we have fresh, isolated containers for every test and the postgres server setup is run atmost once & never retried.
Still, I've used mktemp & context manager to make this more robust as per your suggestions.

base/images/tests/utils/container_runtime.py
MadhurAggarwal: I'm open to replacing AssertionError with pytest.fail if required.
I kept it initially just for consistency with the existing `wait_for_http` fixture which raises `AssertionError`. 

(Note: this file: `container_runtime` does not yet import `pytest`, so changing it to `pytest.fail` would make it pytest-specific util. As of now only one `pytest_plugin` utils imports pytest. Just wanted to know if there is any specific reason / design choice behind this?)

MadhurAggarwal:
Squashed into one commit 

> issue(blocking): In addition to my code comments, most of the commits in this PR don't conform to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), which is required. Most likely you want to do a squash rebase anyway, but either way fix that please.

base/images/tests/utils/container_runtime.py
tobiasb-ms: Eh -- `AssertionError` is probably fine then.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
```diff
@@ -0,0 +1,122 @@
...
+EXPECTED_ROWS = 2
+
+PG_DIR = "/var/lib/pgsql"
+PG_DATADIR = f"{PG_DIR}/data"
+PG_LOGFILE = f"{PG_DIR}/logfile"
+PG_RUNDIR = "/run/postgresql"
+
+
+@contextlib.contextmanager
+def _postgres_password_file(container_exec_shell, password: str):
```
tobiasb-ms: question(non-blocking): Do you think this could be more generically useful? What this function does is:
1. Create a temp file name on the container (completely generic)
2. Put a string into that file (already parameterized)
3. Change the owner of that file (could easily be parameterized)
4. Removes that file as the owner (could be parameterized along with 3)

If you parameterized 3 (and therefor 4), then this function is simply something like `container_temp_file` and any test can use it.

I'm totally fine punting that though -- we don't actually know we'll need it for anything else -- but keep it in mind if you see opportunities for it in the future.

tobiasb-ms (changes requested):
Thanks for updating everything. There's still one blocking thing: https://github.com/microsoft/azurelinux/pull/18001#discussion_r3623137905, and I left a non-blocking question as well.

base/images/tests/cases/runtime/container-base/test_redis/test_redis.py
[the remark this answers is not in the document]
bhagyapathak: I am in favor of adding type annotations for all untyped test fixtures. Since there are several other untyped fixtures in this file, such as container_exec, write_file_in_container, wait_for_http, and assert_http_server. I would rather address them together in a dedicated follow-up PR. That keeps this PR focused on the database tests while ensuring the typing changes are applied consistently across the file. Does that sound reasonable?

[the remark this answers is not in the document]
tobiasb-ms: That works for me.

base/images/tests/utils/extract.py
```diff
@@ -52,7 +52,7 @@
 )
 
 
-def _run(cmd: list[str], **kwargs: object) -> subprocess.CompletedProcess[str]:
+def _run(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess[str]:
     """Run a command, logging it and raising with stderr on failure."""
     logger.info("Running: %s", " ".join(cmd))
     result = subprocess.run(cmd, capture_output=True, text=True, **kwargs)
```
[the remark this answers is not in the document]
MadhurAggarwal: Fixed.

base/images/tests/cases/runtime/container-base/test_postgres/test_postgres.py
```diff
@@ -0,0 +1,122 @@
...
+EXPECTED_ROWS = 2
+
+PG_DIR = "/var/lib/pgsql"
+PG_DATADIR = f"{PG_DIR}/data"
+PG_LOGFILE = f"{PG_DIR}/logfile"
+PG_RUNDIR = "/run/postgresql"
+
+
+@contextlib.contextmanager
+def _postgres_password_file(container_exec_shell: ExecShell, password: str):
```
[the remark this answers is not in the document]
MadhurAggarwal: Fixed.

tobiasb-ms (approved):

PR #18016, opened 2026-07-14
feat(dnf5):  Add a plugin to support virtual snapshots

mbearup:
Add a dnf5 plugin to support virtual snapshots

### DIFF
--- a/base/comps/dnf5/0001-Add-a-plugin-to-support-virtual-snapshots-using-package-file-time.patch
+++ b/base/comps/dnf5/0001-Add-a-plugin-to-support-virtual-snapshots-using-package-file-time.patch
@@ -1,0 +1,767 @@
+From 8f4e665d68947daa514b9889efa25b7d6ffc82f6 Mon Sep 17 00:00:00 2001
+From: Matt Bearup <mbearup@microsoft.com>
+Date: Tue, 16 Jun 2026 09:29:45 -0700
+Subject: [PATCH] Add a plugin to support virtual snapshots using package file
+ time.
+
+---
+ CMakeLists.txt                                |   1 +
+ dnf5-plugins/CMakeLists.txt                   |   4 +
+ .../repo_snapshot_plugin/CMakeLists.txt       |  13 +
+ .../repo_snapshot_cmd_plugin.cpp              | 142 +++++++++++++
+ include/libdnf5/rpm/package.hpp               |   6 +
+ libdnf5-plugins/CMakeLists.txt                |   1 +
+ libdnf5-plugins/repo_snapshot/CMakeLists.txt  |  17 ++
+ .../repo_snapshot/repo_snapshot.conf          |   3 +
+ .../repo_snapshot/repo_snapshot.cpp           | 175 ++++++++++++++++++
+ .../repo_snapshot/repo_snapshot_filter.hpp    |  98 ++++++++++
+ libdnf5/rpm/package.cpp                       |   4 +
+ .../test_repo_snapshot_filter.cpp             | 100 ++++++++++
+ .../test_repo_snapshot_filter.hpp             |  32 ++++
+ test/libdnf5/rpm/test_package.cpp             |   9 +
+ test/libdnf5/rpm/test_package.hpp             |   2 +
+ 15 files changed, 607 insertions(+)
+ create mode 100644 dnf5-plugins/repo_snapshot_plugin/CMakeLists.txt
+ create mode 100644 dnf5-plugins/repo_snapshot_plugin/repo_snapshot_cmd_plugin.cpp
+ create mode 100644 libdnf5-plugins/repo_snapshot/CMakeLists.txt
+ create mode 100644 libdnf5-plugins/repo_snapshot/repo_snapshot.conf
+ create mode 100644 libdnf5-plugins/repo_snapshot/repo_snapshot.cpp
+ create mode 100644 libdnf5-plugins/repo_snapshot/repo_snapshot_filter.hpp
+ create mode 100644 test/libdnf5/plugins/repo_snapshot/test_repo_snapshot_filter.cpp
+ create mode 100644 test/libdnf5/plugins/repo_snapshot/test_repo_snapshot_filter.hpp
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 063dfff..daf22c5 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -32,6 +32,7 @@ option(WITH_PLUGIN_EXPIRED_PGP_KEYS "Build a libdnf5 expired pgp keys plugin" ON
+ option(WITH_PLUGIN_RHSM "Build a libdnf5 rhsm (Red Hat Subscription Manager) plugin" OFF)
+ option(WITH_PYTHON_PLUGINS_LOADER "Build a special dnf5 plugin that loads Python plugins. Requires WITH_PYTHON3=ON." ON)
+ option(WITH_PLUGIN_LOCAL "Build a libdnf5 local plugin" ON)
++option(WITH_PLUGIN_REPO_SNAPSHOT "Build a libdnf5 repo snapshot plugin" ON)
+ 
+ # build options - features
+ option(WITH_COMPS "Build with comps groups and environments support" ON)
+diff --git a/dnf5-plugins/CMakeLists.txt b/dnf5-plugins/CMakeLists.txt
+index bca9866..5fd3a61 100644
+--- a/dnf5-plugins/CMakeLists.txt
++++ b/dnf5-plugins/CMakeLists.txt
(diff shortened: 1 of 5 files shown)


tobiasb-ms (changes requested):
issue(blocking): There's something up with the commits here. Looks like you did a merge of 4.0 onto your fork rather than rebasing (I think GH recently changed the default in the UI or something).

Please do the rebase and then I'll look at the actual changes.

base/comps/dnf5/0001-Add-a-plugin-to-support-virtual-snapshots-using-package-file-time.patch
```diff
@@ -0,0 +1,783 @@
+From 8f4e665d68947daa514b9889efa25b7d6ffc82f6 Mon Sep 17 00:00:00 2001
```
tobiasb-ms, line 1: question(blocking): Where does this patch come from? I see your name on it -- is this original code?

base/comps/dnf5/dnf5.comp.toml
```diff
@@ -6,6 +6,54 @@ without = [
...
+[[components.dnf5.overlays]]
+description = "Add repo_snapshot plugin for virtual snapshots using package file time"
+type = "patch-add"
+source = "0001-Add-a-plugin-to-support-virtual-snapshots-using-package-file-time.patch"
+
+# The repo_snapshot plugin reads a package's publish time via SOLVABLE_FILETIME,
+# which only exists in the patched Azure Linux libsolv. Pin the floor to the
+# 0.7.35-5 build that first added SOLVABLE_FILETIME (see
+# specs/l/libsolv/libsolv.spec changelog) so dnf5 cannot build or install
+# against an older, unpatched libsolv. The release qualifier (-5) is required:
```
tobiasb-ms: suggestion(non-blocking): I don't think the stuff about why the release qualifier is helpful in this comment. Same with the stuff about `_isa`.

base/comps/dnf5/dnf5.comp.toml
```diff
@@ -6,6 +6,54 @@ without = [
...
+[[components.dnf5.overlays]]
+description = "Require the patched libsolv runtime (SOLVABLE_FILETIME) for the repo_snapshot hook plugin"
+type = "spec-add-tag"
+tag = "Requires"
+package = "libdnf5"
+value = "libsolv%{?_isa} >= 0.7.35-5"
+
+# The repo_snapshot patch builds two shared objects (a dnf5 command plugin
+# and a libdnf5 hook plugin) plus a config file. Package all three in the
+# dnf5-plugins subpackage since they are two halves of one feature.
```
tobiasb-ms: issue(non-blocking): I'm not sure what "they are two halves of one feature" means here, especially when there are three files.

tobiasb-ms (dismissed):

base/comps/dnf5/0001-Add-a-plugin-to-support-virtual-snapshots-using-package-file-time.patch
mbearup: Well, it was generated by Agency/Copilot. I didn't take it from someone/somewhere else, if that's what you mean.

base/comps/dnf5/dnf5.comp.toml
mbearup: Yeah, this was added by Agency. Removing now.

base/comps/dnf5/dnf5.comp.toml
mbearup: This is referring to the libsolv changes, which Agency built simultaneously with the dnf changes. I'll remove the comments regarding that.

specs/d/dnf5/dnf5.spec
```diff
@@ -263,6 +263,8 @@ BuildRequires:  rubygem-test-unit
 BuildRequires:  python3-devel
 %endif
 
+Patch0: 0001-Add-a-plugin-to-support-virtual-snapshots-using-package-file-time.patch
```
jslobodzian, line 266: nit:  Place the Patch after the BuildRequires.  Just for consistency with how it's generally done.

jslobodzian:
**On the `build_time` fallback (`should_exclude_package`, repo_snapshot_filter.hpp):**

The `file_time > 0 ? file_time : build_time` fallback doesn't preserve the snapshot invariant, and for a security-labeled feature it should fail closed. In the normal pipeline `build_time <= publish_time` — a package is built, then signed and published later. So an RPM built *before* the cutoff but published *after* it has `build_time < snapshot_time` and is **kept**, even though it wasn't in the repository at the requested instant. Any repo or cached `.solv` metadata lacking `SOLVABLE_FILETIME` (pre-patch libsolv, stale caches, third-party repos) silently routes *every* one of its packages down this build-time path, so the snapshot quietly leaks post-cutoff content instead of failing.

`--snapshot-exclude-repos` is the right escape hatch for repos that legitimately can't supply a publish time; a missing file_time on a *non-excluded* repo should either exclude the package or abort with an actionable "refresh metadata / unsupported repo" error. Suggest dropping the build_time branch (or gating it behind an explicit `--snapshot-allow-buildtime` opt-in) and updating `test_should_exclude_missing_file_time` to lock in the fail-closed behavior.

**On `repo_snapshot.conf` — a bit more depth than the earlier note:**

The config is just `[main]` / `enabled = 1` with no `name`. Here's exactly what that does in the loader (`Base::load_plugin_config`, `libdnf5/base/base.cpp`):

```cpp
try {
    plugin_name = parser.get_value("main", "name");
} catch (const ConfigParserError &) {
    plugin_name = std::filesystem::path(config_file_path).stem();
    logger.warning("Missing plugin name in configuration file \"{}\". ...");
}
```

Because the plugin ships enabled by default, **every** libdnf5-driven dnf5 invocation will log `WARNING: Missing plugin name in configuration file .../repo_snapshot.conf` before falling back to the filename stem. It only works at all because that stem (`repo_snapshot`) happens to equal the plugin's compiled `get_name()` — which is also what makes both the `find_plugin_library("repo_snapshot.so")` resolution and the `--disable-plugin=repo_snapshot` GLOB match (in `Plugins::load_plugin_library`) line up. So this isn't a hard break today, just a spurious warning on every transaction — but library resolution *and* enable/disable handling are silently riding on the filename and would break if the `.conf` were ever renamed. Every sibling config sets it explicitly (`actions.conf` -> `name = actions`, `rhsm.conf` -> `name = rhsm`). One-line fix:

```ini
[main]
name = repo_snapshot
enabled = 1
```

[Toolio Iglesias🗺️]

base/comps/dnf5/0001-Add-a-plugin-to-support-virtual-snapshots-using-package-file-time.patch
```diff
@@ -0,0 +1,787 @@
...
++        return false;
++    }
++
++    if (exclude_repos.count(repo_id) != 0) {
++        return false;
++    }
++
++    // Prefer the repo publish (file) time; fall back to the build time when
++    // it is unavailable.
++    const time_t effective_time = file_time > 0 ? file_time : build_time;
```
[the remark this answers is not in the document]
mbearup: Resolved in the latest iteration.

specs/d/dnf5/dnf5.spec
mbearup: Thanks for pointing this out; resolved in the latest iteration.

mbearup:
Example output, demonstrating the new plugin in action

```
# Using patched libsolv (-5) and dnf (-3)
bash-5.3# rpm -qa | egrep "dnf5|libsolv"
libsolv-0.7.35-5.azl4.x86_64
libdnf5-5.2.18.0-3.azl4.x86_64
libdnf5-cli-5.2.18.0-3.azl4.x86_64
dnf5-5.2.18.0-3.azl4.x86_64
dnf5-plugins-5.2.18.0-3.azl4.x86_64

# Default: Install latest mdatp
bash-5.3# dnf install mdatp --assumeno 2>&1 | grep "mdatp"
 mdatp  x86_64 101.26071.0005_external-1 azurelinux-microsoft 438.0 MiB

# Snapshot: Install older version of mdatp
bash-5.3# dnf --snapshot-time=1784509774 install mdatp --assumeno 2>&1 | grep mdatp
 mdatp  x86_64 101.26052.0011_external-1 azurelinux-microsoft 447.2 MiB
```

ddstreetmicrosoft (changes requested):
this really should be upstream. have you submited upstream? if so, what's the upstream commit(s)/link/pr?

jslobodzian (approved):

base/comps/dnf5/0001-Add-a-plugin-to-support-virtual-snapshots-using-package-file-time.patch
tobiasb-ms: That makes sense. I honestly don't know enough about this codebase to give it a very meaningful review.

PR #18020, opened 2026-07-15
fix(openbios): restore aarch64 builds

christopherco:
Backport upstream host bit-width detection to prevent forthstrap pointer
truncation on aarch64 and riscv64 builders. Remove the x86_64-only
workaround, use %autosetup, and track the rebuild with azl_release.

Upstream: https://github.com/openbios/openbios/commit/e5ac46dd24e6216c36aa80462af25457e7029440
Reference: https://github.com/openbios/openbios/issues/25

Fixes: [AB#22404](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22404)

Copilot-Session: 9541217f-798b-4299-8cd3-28fcb054e4eb


base/comps/openbios/openbios.comp.toml
```diff
@@ -1,15 +1,31 @@
...
+type = "patch-add"
+source = "0001-config-scripts-switch-arch-fix-build-on-aarch64-and-riscv64.patch"
+
+# The spec uses explicit %setup, so patch-add only registers the Patch tag.
 [[components.openbios.overlays]]
-description = "Restrict build to x86_64 — forthstrap segfaults on aarch64 (upstream bug #18/#25)"
-type = "spec-add-tag"
-tag = "ExclusiveArch"
-value = "x86_64"
+description = "Apply the OpenBIOS host bit-width detection patch"
```
tobiasb-ms: suggestion(non-blocking): I would prefer using a `spec-search-replace` to change `%setup` to `%autosetup` so we don't have to remember to do these manual patches in the future.

Definitely non-blocking though.

tobiasb-ms (approved):

base/comps/openbios/openbios.comp.toml
christopherco: Yes, this is a good suggestion. Will work on this and update the PR

christopherco: Updated to use `%autosetup`

tobiasb-ms (approved):

PR #18027, opened 2026-07-15
ci: update CT python tools to support new endpoints

dmcilvaney:
CT endpoints are changing.
`/api/Scenario/package` is being deprecated in favor of:
`/api/Scenario/official/packages` and `/api/Scenario/packages` (which takes different inputs).

Clean up the linter issues in the CT package while refactoring to split the official and non-official build paths to avoid future footguns.


scripts/ci/control-tower/package_build.py
```diff
@@ -0,0 +1,166 @@
+"""Shared mechanics for Control Tower package-build entry points."""
```
reubeno, line 1: [Non-blocking] How far are we away from autogen'ing a proper python client library from CT OpenAPI?

reubeno (approved):

PR #18030, opened 2026-07-15
feat(images): add minimal-os image definition

binujp:
Add a minimal OS VM image with:
- Stripped-down package selection based on AZL3 minimal-packages
- Two build profiles (minimal-os / minimal-os-dev) for repo selection
- UEFI boot with GRUB2 BLS, ext4 rootfs, 1G VHDX
- aarch64 and x86_64 architecture support

Local verification with build and qemu boot and build on CT/koji was done. Static image checks passed.

The 4.0 default selection has more packages 32 vs 28. However, with dependencies pulled 4.0 has more packages ~150 for 3.0 vs ~180 for 4.0. Install size (disk consumption) is 531 MB.

[AB#22015](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22015)

### DIFF
--- a/base/images/minimal-os/minimal-os.kiwi
+++ b/base/images/minimal-os/minimal-os.kiwi
@@ -1,0 +1,113 @@
+<?xml version="1.0" encoding="utf-8"?>
+<?xml-model href="https://raw.githubusercontent.com/OSInside/kiwi/refs/tags/v10.2.33/kiwi/schema/kiwi.rng" type="application/xml"?>
+<image schemaversion="8.3" name="azl4-minimal-os">
+    <description type="system">
+        <author>Azure Linux</author>
+        <contact>azurelinux@microsoft.com</contact>
+        <specification>Azure Linux Minimal OS Image</specification>
+    </description>
+    <!--
+        Build profiles: select "minimal-os" or "minimal-os-dev" to ship
+        azurelinux-repos or azurelinux-repos-dev respectively. That controls
+        which package repository the image uses at runtime.
+    -->
+    <profiles>
+        <profile name="minimal-os"     description="Ships azurelinux-repos (runtime → PMC beta)" import="false" />
+        <profile name="minimal-os-dev" description="Ships azurelinux-repos-dev (runtime → azl4-dev)" import="false" />
+    </profiles>
+    <preferences>
+        <version>0.1</version>
+        <packagemanager>dnf5</packagemanager>
+        <locale>C</locale>
+        <keytable>us</keytable>
+        <timezone>UTC</timezone>
+        <release-version>4.0</release-version>
+        <type
+            image="oem"
+            format="vhdx"
+            initrd_system="dracut"
+            filesystem="ext4"
+            fscreateoptions="-m 1"
+            firmware="uefi"
+            overlayroot="false"
+            eficsm="false"
+            bootpartition="false"
+            efipartsize="32"
+            rootfs_label="azurelinux">
+            <bootloader name="grub2" bls="true" />
+            <size unit="G">1</size>
+            <oemconfig>
+                <oem-resize>false</oem-resize>
+            </oemconfig>
+        </type>
+    </preferences>
+
+    <!--
+        Single build-time repository (azl4-dev) for both variants —
+        koji overrides this for distro builds; for local builds this
+        is the only blob currently populated for AZL4. The variant
+        only affects which `azurelinux-repos*` package ships, NOT
+        where build-time RPMs come from.
+    -->
+    <repository type="rpm-md" alias="azurelinux-base">
+        <source
+            path="https://stcontroltowerdevjwisitg.blob.core.windows.net/azl4-dev/base/$basearch" />
+    </repository>
+
+    <packages type="image">
+        <package name="azurelinux-release" />
+        <package name="bash" />
+        <package name="ca-certificates" />
+        <package name="dbus" />
+        <package name="dnf5" />
+        <package name="e2fsprogs" />
+        <package name="grub2" />
+        <package name="glibc-minimal-langpack" />
+        <package name="iproute" />
+        <package name="iputils" />
+        <package name="irqbalance" />
+        <package name="kernel" />
+        <package name="ncurses-libs" />
+        <package name="openssl" />
+        <package name="rpm" />
+        <package name="shadow-utils" />
+        <package name="sudo" />
+        <package name="systemd" />
+        <package name="systemd-networkd" />
+        <package name="systemd-resolved" />
+        <package name="systemd-udev" />
+        <package name="util-linux" />
+        <package name="zlib" />
+    </packages>
+
+    <!--
+        Variant-scoped repos package: governs the .repo files written
+        into /etc/yum.repos.d/ in the built image (runtime). The two
+        packages conflict, so they're separated into per-variant
+        sections rather than relying on the variant's bootstrap section.
+    -->
+    <packages type="image" profiles="minimal-os-dev">
+        <package name="azurelinux-repos-dev" />
+    </packages>
+    <packages type="image" profiles="minimal-os">
+        <package name="azurelinux-repos" />
+    </packages>
+
+    <packages type="bootstrap">
+        <package name="azurelinux-release" />
+        <package name="filesystem" />
+        <package name="grub2-efi-x64-modules" arch="x86_64" />
+        <package name="grub2-efi-x64" arch="x86_64" />
+        <package name="grub2-efi-aa64-modules" arch="aarch64" />
+        <package name="grub2-efi-aa64" arch="aarch64" />
+        <package name="shim" arch="x86_64" />
+        <package name="shim" arch="aarch64" />
+    </packages>
+
+    <packages type="bootstrap" profiles="minimal-os-dev">
+        <package name="azurelinux-repos-dev" />
+    </packages>
+    <packages type="bootstrap" profiles="minimal-os">
+        <package name="azurelinux-repos" />
+    </packages>
+</image>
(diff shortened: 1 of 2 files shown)


reubeno (commented):
What's the purpose of this image? Is there a PRD that outlines its intended use and guidelines for its composition?

(Is this the intended base image for Image Customizer to use for customization?)

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,144 @@
...
+    <repository type="rpm-md" alias="azurelinux-base">
+        <source
+            path="https://packages.microsoft.com/azurelinux/4.0/beta/base/$basearch" />
+    </repository>
+
+    <packages type="image">
+        <package name="azurelinux-release" />
+        <package name="bash" />
+        <package name="ca-certificates" />
+        <package name="dbus" />
```
reubeno, line 61: My understanding is that dbus has fallen out of favor and dbus-broker is the preferred implementation. Can we confirm what's in our VM image and which is the correct one to use here?

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,144 @@
...
+        only affects which `azurelinux-repos*` package ships, NOT
+        where build-time RPMs come from.
+    -->
+    <repository type="rpm-md" alias="azurelinux-base">
+        <source
+            path="https://packages.microsoft.com/azurelinux/4.0/beta/base/$basearch" />
+    </repository>
+
+    <packages type="image">
+        <package name="azurelinux-release" />
```
reubeno, line 58: Our VM, WSL, and container images use a more specific subpackage of `azurelinux-release` to get the right variant metadata in os-release. What's the plan for how this will work for images derived from the minimal OS image?

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,144 @@
...
+        <package name="bash" />
+        <package name="ca-certificates" />
+        <package name="dbus" />
+        <package name="dnf5" />
+        <package name="e2fsprogs" />
+        <package name="grub2" />
+        <package name="glibc-langpack-en" />
```
reubeno: What's the trade-off between including `glibc-langpack-en` vs. `glibc-minimal-langpack` in the minimal OS image? Will Image Customizer swap it out anyway with the desired langpack?

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,144 @@
...
+    <preferences arch="x86_64">
+        <version>0.1</version>
+        <packagemanager>dnf5</packagemanager>
+        <locale>en_US</locale>
+        <keytable>us</keytable>
+        <timezone>UTC</timezone>
+        <release-version>4.0</release-version>
+        <type
+            image="oem"
+            format="vhd-fixed"
```
reubeno: Does this need to be VHD fixed? It looks like 3.0 used a VHDX, which is generally preferred.

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,144 @@
...
+            formatoptions="force_size"
+            initrd_system="dracut"
+            filesystem="ext4"
+            fscreateoptions="-m 1"
+            kernelcmdline="console=ttyS0 earlyprintk=ttyS0 rd.shell=0"
```
reubeno: Should there be any kernel command line baked in? I don't see evidence of one in 3.0.

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,144 @@
...
+            firmware="uefi"
+            overlayroot="false"
+            eficsm="false"
+            bootpartition="false"
+            efipartsize="200"
```
reubeno: Looks like the 3.0 layout was *extremely* minimal (i see indication of the ESP being tiny in 3.0) since it's only expected to be used as a base image for customization and not a readily deployable OS itself. Should we be following suit?

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,144 @@
...
+            rootfs_label="azurelinux">
+            <bootloader name="grub2" bls="true" console="serial" timeout="1" />
+            <size unit="G">5</size>
```
reubeno: See above regarding partition layout.

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,144 @@
...
+            filesystem="ext4"
+            fscreateoptions="-m 1"
+            kernelcmdline="console=ttyS0 earlyprintk=ttyS0 rd.shell=0"
+            firmware="uefi"
+            overlayroot="false"
+            eficsm="false"
+            bootpartition="false"
+            efipartsize="200"
+            rootfs_label="azurelinux">
+            <bootloader name="grub2" bls="true" console="serial" timeout="1" />
```
reubeno: Similar to kernel config, should we omit bootloader configs from the definition?

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,144 @@
...
+        <timezone>UTC</timezone>
+        <release-version>4.0</release-version>
+        <type
+            image="oem"
+            format="vhd-fixed"
+            formatoptions="force_size"
+            initrd_system="dracut"
+            filesystem="ext4"
+            fscreateoptions="-m 1"
+            kernelcmdline="console=ttyAMA0 earlyprintk=ttyAMA0 rd.shell=0"
```
reubeno: Note: if we remove the kernel command line, I assume we can collapse the 2 arch-specific `<preferences/>` sections.

base/images/minimal-os/minimal-os.kiwi
binujp: This is just the meta package which pulls in dbus-broker.

base/images/minimal-os/minimal-os.kiwi
binujp: Images we build, variants in our image hierarchy, will need us to maintain and provide the variant package. Images built by partner teams like AKS, HPC etc. need not be our concern. We can provide guidance via what we already define.

base/images/minimal-os/minimal-os.kiwi
binujp: You are right. I brought in langpack-en to make sure something is installed since minimal is a meta-package. Reading the spec again, minimal was created to explicitly address the case of a minimal os. I'l swap it.

base/images/minimal-os/minimal-os.kiwi
binujp: Ack, will change.

base/images/minimal-os/minimal-os.kiwi
binujp: I added this and the arm64 command line and such to make this image bootable with minimal effort. Based on the review comments, maybe I should have stuck to how 3.0 does. Will go through this once more and strip out.

base/images/minimal-os/minimal-os.kiwi
binujp: Ack.

base/images/minimal-os/minimal-os.kiwi
binujp: Ack.

base/images/minimal-os/minimal-os.kiwi
binujp: Yes, I will do that refactoring.

tobiasb-ms (approved):

base/images/minimal-os/minimal-os.kiwi
reubeno: question(non-blocking): can we check in on the with the Image Customizer team on this as a follow-up? My understanding is that the primary use case for the image is IC and IC may reason about VARIANT_ID and image variants.

base/images/minimal-os/minimal-os.kiwi
```diff
@@ -0,0 +1,120 @@
...
+    <!--
+        Build profiles: select "minimal-os" or "minimal-os-dev" to ship
+        azurelinux-repos or azurelinux-repos-dev respectively. That controls
+        which package repository the image uses at runtime.
+    -->
+    <profiles>
+        <profile name="minimal-os"     description="Ships azurelinux-repos (runtime → PMC beta)" import="false" />
+        <profile name="minimal-os-dev" description="Ships azurelinux-repos-dev (runtime → azl4-dev)" import="false" />
+    </profiles>
+    <!--
```
reubeno: quibble(non-blocking): This comment is accurate but focused on transient changes in the state of the PR and not about something durably useful to future readers. Recommend removing it.

reubeno (approved):

base/images/minimal-os/minimal-os.kiwi
binujp: I will do that. Did not consider the possibility of image metadata being embedded in a package.

binujp: Clarified this with Chris Gunn. 
> `IC uses the ID and VERSION_ID fields to know which distro it is customizing, so that it can use to correct distro specific logic.
>  
> We generally don't care about VARIANT and VARIANT_ID. If we need to know what is in the image, then we dynamically check if files/etc exist.

base/images/minimal-os/minimal-os.kiwi
binujp: Agree, co-pilot added that and I thought that is correct. It does not add any value.

tobiasb-ms (approved):

PR #18044, opened 2026-07-17
feat(azurelinux-rpm-config): set x86_64 cpu level to v2

ddstreet:
Replace the logic that sets the x86_64 cpu level to either v2 or v3
depending on the RHEL release, to hardcoding it to v2, as we consider
v2 to be our baseline for supported hardware, and we do not plan to
support (or function on) older x86_64 v1 hardware.
    
Fixes: [AB#20423](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/20423)

### DIFF
--- a/base/comps/azurelinux-rpm-config/azurelinux-rpm-config.comp.toml
+++ b/base/comps/azurelinux-rpm-config/azurelinux-rpm-config.comp.toml
@@ -47,6 +47,9 @@ overlays = [
     # Update vendor macro.
     { type = "file-search-replace", file = "macros", regex = "%_vendor.*", replacement = "%_vendor azurelinux" },
 
+    # Update cflags_arch_x86_64_level macro.
+    { type = "file-search-replace", file = "macros", regex = "%__cflags_arch_x86_64_level .*", replacement = "%__cflags_arch_x86_64_level -v2" },
+
     # Remove dist.sh file; we'll replace it wholesale with our own version.
     { type = "file-remove", file = "dist.sh" },
     # Add our own version of dist.sh customized for Azure Linux.
--- a/locks/azurelinux-rpm-config.lock
+++ b/locks/azurelinux-rpm-config.lock
@@ -2,5 +2,5 @@
 version = 1
 import-commit = '6e4185bb29f2ca019b36bec8006e7ad60e58cade'
 upstream-commit = '6e4185bb29f2ca019b36bec8006e7ad60e58cade'
-input-fingerprint = 'sha256:cb133fba511f946ad698cc0b1eb4487223b9710e6253522195d7f277c77d2649'
+input-fingerprint = 'sha256:b93d647bc6e52070e97d9d3ad6824b61acbc7515524159f92f851c3a3c647fd2'
 resolution-input-hash = 'sha256:ed1c0158b0942459a3c1a7f3db4f08a88242ce03d0f325b50e1c411f48c8c514'
--- a/specs/a/azurelinux-rpm-config/azurelinux-rpm-config.spec
+++ b/specs/a/azurelinux-rpm-config/azurelinux-rpm-config.spec
@@ -17,7 +17,7 @@ Name: azurelinux-rpm-config
 # the older branch. When the branch diverges, bump the Version to the Fedora
 # release number.
 Version: 1004
-Release: 3%{?dist}
+Release: 4%{?dist}
 # config.guess, config.sub are GPL-3.0-or-later WITH Autoconf-exception-generic
 License: GPL-1.0-or-later AND GPL-2.0-or-later AND GPL-3.0-or-later WITH Autoconf-exception-generic
 URL: https://aka.ms/azurelinux
--- a/specs/a/azurelinux-rpm-config/macros
+++ b/specs/a/azurelinux-rpm-config/macros
@@ -146,7 +146,7 @@ for k,_ in pairs(stripped_flags) do print(k .. " ") end
 
 # Architecture-specific support.  Internal.  Do not use directly.
 
-%__cflags_arch_x86_64_level %[0%{?rhel} == 9 ? "-v2" : ""]%[0%{?rhel} > 9 ? "-v3" : ""]
+%__cflags_arch_x86_64_level -v2
 %__cflags_arch_x86_64 -march=x86-64%{?__cflags_arch_x86_64_level:%{__cflags_arch_x86_64_level}}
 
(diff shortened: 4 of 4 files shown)


ddstreet:
Note that this only sets the `-march=` parameter to use the `x86_64-v2` ISA target; this does *not* remove the `-mtune` parameter which remains set to `generic`. The result is a binary which _uses_ the `x86_64-v2` instruction set, but doesn't _require_ it (i.e. it should still run on `x86_64-v1` hardware). Specifically:

```
$ readelf -j .note.gnu.property usr/bin/bash

Displaying notes found in: .note.gnu.property
  Owner                Data size        Description
  GNU                  0x00000040       NT_GNU_PROPERTY_TYPE_0
      Properties: x86 feature: IBT, SHSTK
        x86 ISA needed: x86-64-baseline
        x86 feature used: x86, XMM
        x86 ISA used: x86-64-baseline, x86-64-v2

```

I'm not a compiler expert, but my assumption is that gcc is generating multiple optimized code paths, that use the `x86_64-v2` ISA if available (which it always will be) but also has code paths that will work for older `x86_64-v1` ISA hardware - and presumably those code paths will never be used, and only serve to (slightly?) increase the binary file size. That's IIUC of course, and I certainly could be wrong.

I did check RHEL 10.1, since that definitely does build specifically for `x86_64-v3`, and their binaries do actually require the `x86_64-v3` ISA, e.g.:

```
$ readelf -n /bin/bash

Displaying notes found in: .note.gnu.property
  Owner                Data size 	Description
  GNU                  0x00000040	NT_GNU_PROPERTY_TYPE_0
      Properties: x86 feature: IBT, SHSTK
	x86 ISA needed: x86-64-baseline, x86-64-v2, x86-64-v3
	x86 feature used: x86, XMM, YMM
	x86 ISA used: x86-64-baseline, x86-64-v3

```

Maybe we need to also adjust/remove the `-mtune=generic` from the `%__cflags_arch_x86_64_common` macro?

christopherco:
That's definitely interesting. Was glibc also rebuilt with `-march=x86-64-v2` in the build chroot prior to building bash with `-march=x86-64-v2`?

Looks like the ISA Needed ELF note in the final binary is an OR of the ISA Needed property (`GNU_PROPERTY_X86_ISA_1_NEEDED`) of each of the relocatable object inputs. glibc emits a value for `GNU_PROPERTY_X86_ISA_1_NEEDED` in its `crt1.o`/`Scrt1.o`. If glibc was not rebuilt with `-march=x86-64-v2`, `crt1.o`/`Scrt1.o` from glibc has the `GNU_PROPERTY_X86_ISA_1_NEEDED` property still set to baseline v1 only. Additionally we likely don't emit `GNU_PROPERTY_X86_ISA_1_NEEDED` in any of bash's gcc-generated objects because we don't specify `-mneeded` during their gcc build (which we likely shouldn't do). So the linker OR-ing them together, that could explain why bash reports only v1 baseline as needed and no v2.

ddstreet:
Yep that was it, glibc rebuilds with ISA needed set to `x86_64-v2`, and with that bash also has ISA needed as -v2, so it should be all good once we mass-rebuild (with an already-rebuilt glibc and other libs).

reubeno (commented):
Change looks righteous to me.

Do we also have an aarch64 CPU level change in the works (in parallel)? Don't quote me on it, but aarch64 SKUs in Azure seem to be armv8.2 (the Ampere ones) or armv9 (Cobalt 100+)?

ddstreet:
> Change looks righteous to me.
> 
> Do we also have an aarch64 CPU level change in the works (in parallel)? Don't quote me on it, but aarch64 SKUs in Azure seem to be armv8.2 (the Ampere ones) or armv9 (Cobalt 100+)?

I'll defer to @christopherco on that? I am not aware of any open bug for aarch64...

christopherco:
> > Change looks righteous to me.
> > Do we also have an aarch64 CPU level change in the works (in parallel)? Don't quote me on it, but aarch64 SKUs in Azure seem to be armv8.2 (the Ampere ones) or armv9 (Cobalt 100+)?
> 
> I'll defer to @christopherco on that? I am not aware of any open bug for aarch64...

We do not. There was some waffling on the exact ARM hardware to target, so baseline ARMv8 is what we had originally landed on.

base/comps/azurelinux-rpm-config/azurelinux-rpm-config.comp.toml
```diff
@@ -47,6 +47,8 @@ overlays = [
     # Update vendor macro.
     { type = "file-search-replace", file = "macros", regex = "%_vendor.*", replacement = "%_vendor azurelinux" },
 
+    { type = "file-search-replace", file = "macros", regex = "%__cflags_arch_x86_64_level .*", replacement = "%__cflags_arch_x86_64_level -v2" },
```
christopherco, line 51: **nit (style):** Add a comment above this or include a description field in the overlay itself.

christopherco (changes requested):
Changes themselves LGTM. One optional style nit-pick.
Requesting changes to squash the commits together into one commit, which should also get the CI checks passing

base/comps/azurelinux-rpm-config/azurelinux-rpm-config.comp.toml
ddstreet: updated, although I think it degrades the product, instead of improving it. git blame is the proper way to get the "why", not a short comment or "description".

ddstreet:
> Changes themselves LGTM. One optional style nit-pick. Requesting changes to squash the commits together into one commit, which should also get the CI checks passing

ok, i folded the 'update' and 'render' commits into the main commit, though i thought they were supposed to be separate...I guess the details of 'update' and/or 'render' are more complicated than I thought

ddstreet:
updated again to do a re-render after merging the commits into a single commit, as the rendered `Release:` is now `4` instead of `5`

base/comps/azurelinux-rpm-config/azurelinux-rpm-config.comp.toml
christopherco: I do agree with moving away from short comments above the toml entries. Usually comments like this are prone to desyncing with future changes.

TOML description field is useful as structured metadata attached to an overlay. Tools can surface it in diagnostics, or in tool outputs or reports.

That being said, my nit was mostly to keep the current style which thank you for making this update. Will follow up on establishing better style practices offline.

christopherco (approved):

PR #18050, opened 2026-07-17
feat(kernel): enable DAMON on x86_64 

AZaugg:
Enable the Data Access Monitoring framework in the x86_64 kernel config:
  CONFIG_DAMON=y
  CONFIG_DAMON_VADDR=y   # virtual-address-space monitoring
  CONFIG_DAMON_PADDR=y   # physical-address-space monitoring
  CONFIG_DAMON_SYSFS=y   # sysfs control interface
DAMON_RECLAIM/LRU_SORT/STAT are left off (opt-in). The sub-options are
declared explicitly so process_configs.sh (olddefconfig + listnewconfig)
stays consistent now that the parent exposes them. Source and rendered
config kept in sync via the verbatim config overlay; kernel.lock
input-fingerprint refreshed via 'azldev comp update -p kernel'.

Check
https://github.com/microsoft/azurelinux/pull/17874
https://github.com/microsoft/azurelinux/pull/18224
https://github.com/microsoft/azurelinux/pull/18226


base/comps/kernel/6.18-x86_64-azl.config
```diff
@@ -1328,7 +1328,13 @@ CONFIG_PT_RECLAIM=y
 #
 # Data Access Monitoring
 #
-# CONFIG_DAMON is not set
+CONFIG_DAMON=y
```
reubeno, line 1331: Is there a reason this PR proposes enabling only on x86_64 and not also aarch64?

base/comps/kernel/6.18-x86_64-azl.config
```diff
@@ -6179,7 +6185,7 @@ CONFIG_INFINIBAND_ON_DEMAND_PAGING=y
 CONFIG_INFINIBAND_ADDR_TRANS=y
 CONFIG_INFINIBAND_ADDR_TRANS_CONFIGFS=y
 CONFIG_INFINIBAND_VIRT_DMA=y
-# CONFIG_INFINIBAND_BNXT_RE is not set
+CONFIG_INFINIBAND_BNXT_RE=m
```
reubeno, line 6265: Is this intended as a stacked PR on top of your other one that enables this option?

reubeno (commented):
I'm not familiar with this particular kernel feature; I'd find it helpful to understand what use case you have in mind, whether there's runtime performance overhead of being enabled (but not activated), etc.

(FWIW -- It looks like this feature wasn't enabled in Azure Linux 3.0, and there's a mix of enablement/disablement across some other distros.)

base/comps/kernel/6.18-x86_64-azl.config
AZaugg: I can enable it for arm as well. Was not sure if kernel configs were suppose to be mirrored, figured it was a task, only for x86 from my isde

base/comps/kernel/6.18-x86_64-azl.config
AZaugg: Yes, given the config was hashed, i wanted to avoid having to regenerate the hash after the bntx module was merged.

AZaugg:
> I'm not familiar with this particular kernel feature; I'd find it helpful to understand what use case you have in mind, whether there's runtime performance overhead of being enabled (but not activated), etc.
> 
> (FWIW -- It looks like this feature wasn't enabled in Azure Linux 3.0, and there's a mix of enablement/disablement across some other distros.)

Hey @reubeno 
check out https://docs.kernel.org/mm/damon/index.html

Wanted to use it to profile applications and see if we can use this telemetry to better right size them(memory footprint). Well thats the idea. All the auto reclaim options were left off, only flipped the observability side.

base/comps/kernel/6.18-x86_64-azl.config
rlmenge: Let's mirror to both here and align with Fedora on these configs

christopherco (changes requested):
Thanks for the contribution! We've merged the first PR, and the 2nd, 3rd PRs and this one needs rebasing. Please rebase them and we can review. Generally changes themselves look good.

PR #18052, opened 2026-07-17
chore(azldev): move azldev to a proper go module tool

dmcilvaney:
Move azldev version management into a conventional nested Go module so Dependabot can update the direct azldev dependency. Dependabot does not yet support dependencies declared solely by Go tool directives.

- Refactor version resolution to read `tools/azldev/go.mod` and resolve it to a validated immutable Git hash at the GitHub Actions, Docker, and ADO installation boundaries.
- Retain `.azldev-version` temporarily for compatibility with the currently deployed trusted rendered-spec workflow.
- After this merges, existing `4.0` PRs must rebase so their raw heads include `tools/azldev/go.mod`.
- ADO runs are reviewer-gated; a target ADO run will validate the internal Go proxy path before merge.


scripts/ci/render-specs-check/resolve_azldev_version.py
```diff
@@ -56,20 +63,103 @@ def _validate_version(value: str, source: str) -> str:
...
+            check=False,
+            cwd=cwd,
+            timeout=GO_COMMAND_TIMEOUT_SECONDS,
+            env=environment,
+        )
+    except (OSError, subprocess.TimeoutExpired) as error:
+        message = f"could not run go {' '.join(args)}: {error}"
+        raise ResolutionError(message) from error
+    if result.returncode != 0:
+        detail = result.stderr.strip() or f"go exited with {result.returncode}"
```
tobiasb-ms: suggestion(blocking): I don't know about `golang`, but lots of tools will print helpful diagnostics to `stdout` even on error, which this swallows. We should probably capture that `stdout` as well.

scripts/ci/render-specs-check/resolve_azldev_version.py
```diff
@@ -56,20 +63,103 @@ def _validate_version(value: str, source: str) -> str:
...
         raise ResolutionError(message)
+    return result.stdout
+
+
+def _validated_module_path(path: Path) -> Path:
+    """Return an absolute regular module file without symlinked components."""
+    absolute_path = path.absolute()
+    for component in (absolute_path, *absolute_path.parents):
+        if component.is_symlink():
+            message = f"{path} must not use a symlinked path component"
```
tobiasb-ms: suggestion(non-blocking): It would be nice to include the guilty component here. I haven't run/tested this suggestion, but something like this:

```suggestion
            message = f"{path} must not use a symlinked path component but {component} is a symlink."
```

scripts/ci/render-specs-check/resolve_azldev_version.py
```diff
@@ -56,20 +63,103 @@ def _validate_version(value: str, source: str) -> str:
...
+    requires = module.get("Require", []) if isinstance(module, dict) else []
+    direct_requirements = [
+        requirement
+        for requirement in requires
+        if isinstance(requirement, dict)
+        and requirement.get("Path") == AZLDEV_MODULE
+        and requirement.get("Indirect", False) is False
+    ]
+    if len(direct_requirements) != 1:
+        message = f"{path} must contain exactly one direct azldev requirement"
```
tobiasb-ms: suggestion(non-blocking): Include the actual number of requirements in this error message (untested suggestion).
```suggestion
        message = f"{path} must contain exactly one direct azldev requirement, but has {len(direct_requirements)}"
```

scripts/ci/render-specs-check/resolve_azldev_version.py
```diff
@@ -56,20 +63,103 @@ def _validate_version(value: str, source: str) -> str:
...
+        raise ResolutionError(message)
+    version = direct_requirements[0].get("Version")
+    if not isinstance(version, str):
+        message = f"{path} direct azldev requirement has no version"
```
tobiasb-ms: suggestion(non-blocking): This message is potentially misleading. If for example `version` is an int, we're saying it has no version but that's false. This can frustrate investigations.

I suggest printing the type and value in the message.

scripts/ci/render-specs-check/resolve_azldev_version.py
```diff
@@ -56,20 +63,103 @@ def _validate_version(value: str, source: str) -> str:
...
+
+
+def _version_from_content(content: str, source: str) -> str:
+    """Parse an API-provided go.mod through Go and return its azldev version."""
+    try:
+        with tempfile.TemporaryDirectory() as directory:
+            modfile = Path(directory, "go.mod")
+            modfile.write_text(content, encoding="utf-8")
+            return read_version(modfile)
     except (OSError, UnicodeError) as error:
```
tobiasb-ms: question(blocking-for-discussion): Are these realistic errors? You just created the file, and wrote it as `utf-8`. Seems like errors would be in `read_version`, which seems to normalize everything to a `ResolutionError`.

tobiasb-ms (changes requested):

scripts/ci/render-specs-check/resolve_azldev_version.py
dmcilvaney: ah, agree with the new pattern this can just be OSError now.

tobiasb-ms (approved):

PR #18065, opened 2026-07-21
nodejs: patch brace-expansion for CVE-2026-45149

aadhar-agarwal:


kgodara912:
As we have minor version of nodejs already released with the fix, it will be better to upgrade to 24.18, could you please upgrade the version of nodejs instead of patching?

PR #18068, opened 2026-07-21
feat(kernel): convert component to local spec

rlmenge:
Fork the kernel packaging sources into Azure Linux and convert the kernel component to use the vendored local spec.

The two commits:

1. Vendor the current Fedora sources unchanged.
2. Switch the component from Fedora overlays to the local spec.

#### Motivation
The kernel packaging has significant Azure Linux-specific customization, making the growing overlay stack difficult to maintain and review. Forking the packaging sources gives us direct ownership and provides a clearer base for future kernel updates and cleanup.

This migration preserves the current kernel configuration and package output. Cleanup of unused Fedora/RHEL packaging will follow separately.

#### Validation
Updated the component lock
Rendered both commits independently
Confirmed the final rendered packaging is unchanged


tobiasb-ms (approved):
confirmed no changes other than comments, changelog and release bump

ellie-di (approved):

christopherco (approved):

PR #18071, opened 2026-07-21
chore: remove 'default-distro' from object model 'distribution' object

ddstreet:
The 'distribution' object is informational, not a configuration object. The correct place for configuration of the project's default distribution version is inside the project configuration.

This places it in the correct object model location in our repo, but doesn't fix the 'azldev' tool.

### DIFF
--- a/base/project-override-stage2-prod.toml
+++ b/base/project-override-stage2-prod.toml
@@ -2,5 +2,5 @@
 # be explicitly added to force selection of 4.0-stage2-prod as the default distro
 # version to build.
 
-[distros.azurelinux]
-default-version = "4.0-stage2-prod"
+[project]
+default-distro = { name = "azurelinux", version = "4.0-stage2-prod" }
--- a/base/project.toml
+++ b/base/project.toml
@@ -9,5 +9,5 @@ description = 'azurelinux-base'
 log-dir = 'build/logs'
 work-dir = 'build/work'
 output-dir = 'out'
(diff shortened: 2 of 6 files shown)


reubeno (changes requested):
These *are* a part of the existing configurational model; we can't remove them like this.

If you are instead claiming that you don't *want* them to be, then that's another story.

ddstreet:
> These _are_ a part of the existing configurational model; we can't remove them like this.

Can you elaborate on why you are stating this can't be done?

The [documentation](https://github.com/microsoft/azure-linux-dev-tools/blob/main/docs/user/reference/config/distros.md) for the `distros` clearly says the `default-version` field is optional.

The [documentation](https://github.com/microsoft/azure-linux-dev-tools/blob/main/docs/user/reference/config/project.md) for the `project` allows specifying the `default-distro`.

Are you saying that there is something using the `default-version` defined in the `distro` object, but that _doesn't_ also look at the `default-distro` in the `project` object? Meaning, our configuration cannot be changed even when staying within the documentated format and structure?

I'm just trying to understand what/where this might cause a problem, since it seems like - from the docs as well as our current config - it should be ok? My intention here is to keep the `distro` objects informational, instead of having any policy included in them. This seemed like an easy, clear-cut simple first step, but maybe I'm missing something?

ddstreet:
added commit from @reubeno

reubeno (approved):

christopherco (approved):

PR #18076, opened 2026-07-22
feat(kernel): enable x86 USB HID

sirredbeard:
Enable `CONFIG_USB_HID=m` for x86_64.

The x86_64 configuration already builds the generic HID core. This adds the standard USB HID transport as a module.

arm64 already enables USB HID.


sirredbeard:
Updated the title to follow Conventional Commits.

rlmenge (approved):
The extra `# CONFIG_* is not set` lines are expected fallout from enabling `CONFIG_USB_HID=m`. This PR is scoped to generic x86 USB HID support. The newly visible vendor-specific drivers are not enabled because we do not have a current hardware or workload requirement for them.

`CONFIG_USB_HIDDEV` is intentionally not enabled as well. It is a legacy HID userspace interface, and there is no known need for it at the moment. It can be enabled later if a specific compatibility requirement appears.

christopherco (approved):
:shipit:

PR #18085, opened 2026-07-22
fix(rust): pin cargo_metadata + databake to f43 fix commits

WithEnoughCoffee:
Fixes `%check` failures for 4 Rust components by pinning each to the Fedora f43 dist-git commit that already carries the upstream fix (no bespoke patches).

| Component | Fix |
|---|---|
| rust-cargo_metadata | f43 adds `--skip error1` |
| rust-cargo_metadata0.15 | f43 adds `--skip error1` |
| rust-cargo_metadata0.19 | f43 adds `--skip error1` |
| rust-databake (+ derive) | bump pair to 0.2.1 |

**Root cause:** cargo_metadata's `error1` selftest exact-matches Cargo stderr, which recent Cargo reworded. databake 0.2.0's private-field doctests are rejected by databake-derive 0.2.1's new "requires public fields" assert; f43 ships both at 0.2.1.

The default Fedora snapshot predates these fixes; each pin carries a TODO to drop once the snapshot advances.

**Validation:** all 5 components build and package cleanly under the stage1 / PME config (`--config-file distro/stage1.toml`, per #18074), whose buildroot pulls rust-1.95.0-5.fc43 — the newer toolchain that exposes the `error1` brittleness. Depends on #18074 being applied to reproduce.

`%check` results (4 of 5 run tests; `rust-databake-derive` keeps `%bcond check 0`, matching Fedora f43, so it is build/package-verified only):

| Component | `%check` | Result |
|---|---|---|
| rust-cargo_metadata | enabled | pass (`error1` skipped) |
| rust-cargo_metadata0.15 | enabled | pass (`error1` skipped) |
| rust-cargo_metadata0.19 | enabled | pass (`error1` skipped) |
| rust-databake | enabled | pass (doctests) |
| rust-databake-derive | disabled (`%bcond check 0`) | not run — builds/packages cleanly |


WithEnoughCoffee:
Good catch — corrected the validation wording in the PR description. `rust-databake-derive` keeps `%bcond check 0` (matching Fedora f43; the derive crate has no standalone tests), so it is build/package-verified only, not `%check`-verified. The other four run `%check` and pass — the three cargo_metadata variants with `error1` skipped, and `rust-databake` running its doctests against the freshly built 0.2.1 derive.

reubeno (approved):

PR #18102, opened 2026-07-23
fix(rocm-examples): add libpfm-devel BuildRequires for libpfm-enabled google-benchmark

reubeno:
_Background:_ Upstream F43 changes have since enabled libpfm in google-benchmark; these changes, unfortunately, require consumers of google-benchmark to also enable libpfm-devel. Those compensating changes have been making their way through Rawhide.

_Issue here_: rocm-examples' Tutorials/reduction/benchmark CMakeLists does find_package(benchmark CONFIG QUIET). If the system's google-benchmark package is libpfm-enabled, CMake config runs an unconditional find_dependency(PFM), which fails without libpfm-devel. When this happens, the benchmark is silently not found -- and that causes the build of this package to fall back to network-fetching content from github. That fails in network-isolated koji.

Adding libpfm-devel lets find_package(benchmark) resolve so the packaged benchmark is used and no network fetch is attempted.

### DIFF
--- a/base/comps/components.toml
+++ b/base/comps/components.toml
@@ -4527,7 +4527,6 @@ includes = ["**/*.comp.toml", "component-check-disablement.toml", "component-min
 [components.rocm-bandwidth-test]
 [components.rocm-cmake]
 [components.rocm-core]
-[components.rocm-examples]
 [components.rocm-omp]
 [components.rocm-rpm-macros]
 [components.rocm-rpp]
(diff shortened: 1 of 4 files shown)


tobiasb-ms (approved):
copilot left a good but [non-blocking comment](https://github.com/microsoft/azurelinux/pull/18102#discussion_r3640761451) you should look at, but otherwise good.

tobiasb-ms (approved):

PR #199, opened 2020-10-13
Natively support pulling from the preview repo

schmittjoseph:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [x] The toolchain has been rebuilt successfully (or no changes were made to it)
- [x] The toolchain/worker package manifests are up-to-date
- [x] Any updated packages successfully build (or no packages were changed)
- [x] All package sources are available
- [x] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`)
- [x] All source files have up-to-date hashes in the `*.signatures.json` files
- [x] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [x] Documentation has been updated to match any changes to the build system
- [x] Ready to merge

---

###### Summary 

Support using the preview repo during builds by setting `USE_PREVIEW_REPO=y` (defaults to `n`).  Adding this flag to build invocations will enable downloading preview SRPMs, preview toolchain RPMs and preview RPMs for image composition and build requirements.

###### Change Log  

- Replace `PACKAGE_URL` and `PACKAGE_UPDATE_URL` with a new `PACKAGE_URL_LIST` flag.
- Only pull preview SRPMs if `USE_PREVIEW_REPO=y`.
- Install the preview repo package in the worker chroot. However, disable pulling from it by default unless `USE_PREVIEW_REPO=y`.

###### Does this affect the toolchain?  

NO

###### Test Methodology

- Local builds.

PR #195, opened 2020-10-12
Adding a missing '%{?dist}' tag.

PawelWMS:
###### Merge Checklist  

**All** boxes should be checked before merging the PR *(just tick any boxes which don't apply to this PR)*
- [X] The toolchain has been rebuilt successfully (or no changes were made to it)
- [X] The toolchain/worker package manifests are up-to-date
- [X] Any updated packages successfully build (or no packages were changed)
- [X] All package sources are available
- [X] cgmanifest files are up-to-date and sorted (`./cgmanifest.json`, `./toolkit/tools/cgmanifest.json`, `./toolkit/scripts/toolchain/cgmanifest.json`)
- [X] All source files have up-to-date hashes in the `*.signatures.json` files
- [X] `sudo make go-tidy-all` and `sudo make go-test-coverage` pass
- [X] Documentation has been updated to match any changes to the build system
- [X] Ready to merge

---

###### Summary 

Adding a missing `%{?dist}` to align with our package-naming requirements.

###### Change Log  

- Added a `%{?dist}` tag.

###### Does this affect the toolchain?  

No.

###### Test Methodology

- No tests, a simple change.

PR #18106, opened 2026-07-23
fix(libjcat): disable %check pending ML-DSA-87/GnuTLS 3.8.13 fix (AB#22607)

WithEnoughCoffee:
## Problem
`libjcat`'s `%check` aborts in the AZL4 stage-1 (Fedora 43) buildroot on the
`pkcs7-self-signed-pq-gnutls` self-test.

## Root cause
The snapshot we build (**0.2.5**) has no runtime skip for post-quantum signing:
the self-test calls `jcat_engine_self_sign(..., JCAT_SIGN_FLAG_USE_PQ, &error)`
and then unconditionally `g_assert_no_error(error)`. When ML-DSA-87 cannot be
enabled at runtime in this buildroot, self-sign returns an error and the
assertion aborts `%check`.

Upstream **0.2.6** later added a runtime skip (`g_test_skip()`), but it only
matches `G_IO_ERROR_NOT_SUPPORTED`, so it can still fail if GnuTLS
(ML-DSA-capable 3.8.13) surfaces the unavailability as a different error such as
`G_IO_ERROR_FAILED`. Bumping to 0.2.6 therefore would not reliably fix this.

## Change
Per review, rather than carry a downstream patch that broadens the skip on a
generic error (which risks masking a real PQ regression), this **temporarily
disables `%check`** for libjcat so packages can build:

```toml
[components.libjcat.build]
check = { skip = true, skip_reason = "… [AB#22607](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22607) … hughsie/libjcat#195 …" }
```

The component stays on the default snapshot 0.2.5 (no version pin, no downstream
patch). The `skip_reason` renders into `%check` as a comment referencing the
tracking bug and upstream issue.

## Follow-up
- **Tracking bug [AB#22607](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22607)** — root-cause why ML-DSA-87 can't be enabled at
  runtime in the stage-1 crypto stack, and re-enable `%check` once fixed.
- **Upstream in parallel:** libjcat
  [#195](https://github.com/hughsie/libjcat/issues/195) /
  [#199](https://github.com/hughsie/libjcat/issues/199) (skip PQ when it is
  unavailable at runtime regardless of the specific error); GnuTLS re: the
  error reported when an ML-DSA algorithm is present but unusable at runtime.

## Validation
- `azldev comp update -p libjcat`: `changed=0` / lock current.
- `azldev comp render -p libjcat`: `CHANGED=false` / rendered spec current.
- `azldev comp build -p libjcat -q`: built `libjcat-0.2.5-3.azl4.x86_64.rpm` successfully.
- Inspected the RPM payload with `rpm -qlp` in the AZL4 mock chroot.
- Installed and smoke-tested in the AZL4 mock chroot: `rpm -q libjcat`, `jcat-tool --version`, and `jcat-tool --help` passed.


base/comps/libjcat/libjcat.comp.toml
```diff
@@ -0,0 +1,15 @@
+[components.libjcat]
```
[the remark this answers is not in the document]
WithEnoughCoffee: The PR scope was changed after this review: it now intentionally keeps the default Fedora 43 snapshot at 0.2.5 and temporarily disables `%check`.

base/comps/libjcat/libjcat.comp.toml
```diff
@@ -0,0 +1,17 @@
...
+# error and the assertion aborts %check. (0.2.6 later added a runtime skip via
+# g_test_skip(), but it only matches G_IO_ERROR_NOT_SUPPORTED, so it can still
+# fail if GnuTLS surfaces the unavailability as a different error.)
+# Temporarily skip %check to unblock building until ML-DSA-87 works at runtime
+# in the stage-1 crypto stack (or upstream skips PQ whenever it is unavailable).
+# Upstream: https://github.com/hughsie/libjcat/issues/195 (and #199)
+# TODO(AB#22607): Re-enable %check once ML-DSA-87 works at runtime in the
+# stage-1 buildroot, or upstream libjcat skips PQ when it is unavailable.
+[components.libjcat.build]
+check = { skip = true, skip_reason = "pkcs7-self-signed-pq self-test aborts on AZL4 stage-1 (Fedora 43): the snapshot we build (0.2.5) unconditionally asserts that post-quantum (ML-DSA-87) self-signing succeeds, but ML-DSA-87 cannot be enabled at runtime in this buildroot, so %check fails. (0.2.6 adds a runtime skip but it only matches G_IO_ERROR_NOT_SUPPORTED.) Tracked in AB#22607; re-enable once ML-DSA-87 works at runtime or upstream (hughsie/libjcat#195) skips PQ when unavailable." }
```
tobiasb-ms, line 17: issue(blocking): By having `%check` here, you have a `%check` in the comment in the `%check` section. Given that `rpmbuild` expands macros even in comments, I'm not exactly sure what that will do.

Please run the build and look for warnings about this. If there are no warnings, we're probably safe. If there are warnings (or worse), you'll need to change that -- probably to either `%%check` or just `check`.

tobiasb-ms (changes requested):

base/comps/libjcat/libjcat.comp.toml
WithEnoughCoffee: The build passes.

reubeno: FWIW -- in this case, `%check` probably isn't a macro (the `%check` section is a special section header and not a macro invocation) -- and may generate no warnings.

tobiasb-ms: Ok -- as long as we aren't accidentally generating warnings we're good.

tobiasb-ms (approved):

PR #18112, opened 2026-07-24
test: add footprint tests for container, wsl & vm-base image

bhagyapathak:


bhagyapathak:
[Uploading footprint_tests_all_image_variants.txt…]()

base/images/tests/cases/static/test_footprint.py
```diff
@@ -0,0 +1,144 @@
...
+# installed-size cap since the artifact is compressed.
+_ARTIFACT_CAP_MB: dict[str, int] = {
+    "container-base": 200,
+    "wsl": 300,
+}
+_ARTIFACT_CAP_DEFAULT_MB = 300
+
+_TARBALL_IMAGE_TYPES = frozenset({"container", "wsl"})
+
+_TOP_N = 15
```
tobiasb-ms: nitpick(non-blocking): This is not a useful constant name. Top N what? Use something descriptive pleaes.

base/images/tests/cases/static/test_footprint.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def _family_of(image_name: str | None, known: Iterable[str]) -> str | None:
+    """Map an ``--image-name`` to its family (e.g. ``container-base-dev`` → ``container-base``)."""
+    if not image_name:
```
tobiasb-ms: suggestion(blocking): An `image_name` of `None` feels like an error (so raise an exception) rather than the same as not finding `image_name`'s family.

base/images/tests/cases/static/test_footprint.py
```diff
@@ -0,0 +1,144 @@
...
+    return total
+
+
+@pytest.mark.require_capability("runtime-package-management")
+def test_installed_footprint_reported(installed_package_sizes: dict[str, int]) -> None:
+    """Every installed package reports a non-negative size and the total is positive."""
+    if not installed_package_sizes:
+        pytest.fail("rpm returned no package size information")
+
+    negative = {name: size for name, size in installed_package_sizes.items() if size < 0}
```
tobiasb-ms, line 88: question(blocking): Is `0` a valid size? Or should the test be `<= 0` here?

base/images/tests/cases/static/test_footprint.py
```diff
@@ -0,0 +1,144 @@
...
+    if negative:
+        pytest.fail(f"packages with negative reported size: {sorted(negative)}")
+
+    total = _log_footprint(installed_package_sizes)
+    if total <= 0:
```
tobiasb-ms: nitpick(non-blocking): Given that we've already guaranteed that none of the items have a negative size, `total` can't possibly be negative here. And I asked above whether we should actually fail on non-positive sizes (not just negative sizes), in which case this test is completely reduntant.

base/images/tests/cases/static/test_footprint.py
```diff
@@ -0,0 +1,144 @@
...
+    for name, size in top:
+        logger.info("  %-40s %10d bytes (%.1f MB)", name, size, size / _BYTES_PER_MB)
+    return total
+
+
+@pytest.mark.require_capability("runtime-package-management")
+def test_installed_footprint_reported(installed_package_sizes: dict[str, int]) -> None:
+    """Every installed package reports a non-negative size and the total is positive."""
+    if not installed_package_sizes:
+        pytest.fail("rpm returned no package size information")
```
tobiasb-ms: question(blocking): Why are you using `pytest.fail` rather than an assert? Wouldn't something like this (from memory -- treat it as a guideline, not real code) be more idiomatic and clear?
```
assert installed_packages_sizes, "rpm returned no package size information"
```

Same goes for all of the uses of `pytest.fail` throughout the file.

base/images/tests/utils/extract.py
```diff
@@ -206,6 +210,64 @@ def unmount_container_image(extract_dir: Path) -> None:
...
+    Rootless extraction can leave read-only directories, so make entries
+    writable before unlinking. Failures are logged, not raised.
+    """
+    logger.info("Removing WSL extract dir %s", extract_dir)
+
+    def _make_writable_and_retry(func: Callable[[str], object], path: str, _exc: BaseException) -> None:
+        # Unlink permission comes from the parent directory, so making the
+        # parent writable is what allows the retry to remove a child of a
+        # read-only (e.g. 0555) directory.
+        with contextlib.suppress(OSError):
```
tobiasb-ms: question(blocking): Directly below, you have comments warning to never `chmod` symlinks, but you have no check here. Is that a concern?

base/images/tests/utils/parsers.py
```diff
@@ -56,7 +64,19 @@ def query_rpm_packages(rootfs: Path) -> set[str]:
...
-    logger.debug("rpm query returned %d packages", len(pkgs))
-    return pkgs
+        raise RuntimeError(f"rpm size query failed (rc={result.returncode}): {result.stderr.strip()}")
+    sizes: dict[str, int] = {}
+    for raw_line in result.stdout.splitlines():
+        entry = raw_line.strip()
+        if not entry:
+            continue
+        name, _, size_str = entry.rpartition(" ")
+        if not name:
```
tobiasb-ms: nit(non-blocking): Is this a realistic case? In what circumstances would the line be a non-empty string but there's no name field? That seems like a failure and/or test code bug, in which case we'd rather see it and address it.

base/images/tests/utils/parsers.py
```diff
@@ -56,7 +64,19 @@ def query_rpm_packages(rootfs: Path) -> set[str]:
...
+            continue
+        try:
+            size = int(size_str)
+        except ValueError as exc:
+            raise RuntimeError(f"unexpected rpm size for {name}: {size_str!r}") from exc
+        sizes[name] = sizes.get(name, 0) + size
```
tobiasb-ms, line 80: suggestion(non-blocking): I'm not sure I get this. So is the idea that more than one rpm with the same name can be installed and we need to add up the sizes? That doesn't really make sense to me, given my understanding of `rpm -qa`. What am I missing here?

tobiasb-ms (changes requested):

base/images/tests/cases/static/test_footprint.py
bhagyapathak: individual packages can legitimately be 0 (metapackages/no files), so I intentionally don't reject per-package non-positive sizes; the total-zero assert catches the distinct case of rpm returning no usable size data at all

base/images/tests/cases/static/test_footprint.py
bhagyapathak: Yes, 0 is a valid per-package size

base/images/tests/utils/parsers.py
bhagyapathak: rpm -qa emits one line per installed instance, and a NAME can have several - multilib (glibc.x86_64+glibc.i686, two sizes) or install-only kernels kept at multiple versions. Overwriting would drop all-but-one. Today's freshly-built 4.0 image happen to have no duplicates so this is robustness rather than a live fix.

tobiasb-ms: Ah, I see my confusion now. I was conflating `NAME`, which may not be unique as you point out, with `NEVRA`, which I think can't be.

I think NEVRA would be better here, as when you list them in the logs it's more accurate -- it tells you the precise packages. So I would suggest using `NEVRA` over `NAME`.

That said, since I agree that this solution works I'm changing this to a non-blocking suggestion.

base/images/tests/utils/extract.py
```diff
@@ -206,6 +210,205 @@ def unmount_container_image(extract_dir: Path) -> None:
...
+    """Resolve a hardlink target the way :class:`tarfile.TarFile` does.
+
+    ``TarFile._find_link_target`` normalizes the linkname and, for a hardlink,
+    searches only the members *before* the link (a hardlink references an
+    already-archived file), returning the latest match by normalized name.
+    Mirror that exactly so this precheck cannot disagree with what extraction
+    actually links to (e.g. a ``linkname`` like ``a/b/../b/s`` or duplicate
+    member names would otherwise be missed by a raw-name lookup).
+    """
+    wanted = os.path.normpath(linkname)
```
tobiasb-ms, line 244: suggestion(non-blocking): Consider adding an assert here that the lists are the same size, which I think you're relying on.

base/images/tests/utils/extract.py
```diff
@@ -206,6 +210,205 @@ def unmount_container_image(extract_dir: Path) -> None:
...
+            target (including the CVE-2026-11940 hardlink-to-symlink fallback
+            path) resolves outside ``dest``.
+    """
+    normalized_names = [os.path.normpath(member.name) for member in members]
+    for index, member in enumerate(members):
+        member_path = dest / member.name
+        if not _resolves_within(dest, member_path):
+            raise RuntimeError(f"WSL archive member escapes destination: {member.name!r}")
+
+        if member.issym() and not PurePosixPath(member.linkname).is_absolute():
```
tobiasb-ms, line 290: question(blocking): I'm unclear on this -- what should happen for relative symlinks?

base/images/tests/utils/extract.py
```diff
@@ -206,6 +210,205 @@ def unmount_container_image(extract_dir: Path) -> None:
...
+            if not _resolves_within(dest, dest / os.path.normpath(member.linkname)):
+                raise RuntimeError(f"WSL archive hardlink escapes destination: {member.name!r} -> {member.linkname!r}")
+            # CVE-2026-11940: if the hardlink resolves (per tarfile's own
+            # semantics) to a symlink member, the extraction fallback recreates
+            # that symlink at the hardlink's (shallower) path, so evaluate the
+            # symlink's relative target from the hardlink's directory rather
+            # than the symlink's own.
+            linked = _find_preceding_member(members, normalized_names, index, member.linkname)
+            if linked is not None and linked.issym() and not PurePosixPath(linked.linkname).is_absolute():
+                effective = member_path.parent / linked.linkname
```
tobiasb-ms: suggestion(non-blocking): This is nearly identical to the check done above for `member.issym`. Consider a helper function.

base/images/tests/utils/extract.py
```diff
@@ -206,6 +210,205 @@ def unmount_container_image(extract_dir: Path) -> None:
...
+        RuntimeError: If the path cannot be resolved within ``root`` (e.g. a
+            symlink loop exceeds ``max_links``).
+    """
+    root = root.resolve()
+    current = root
+    pending = list(PurePosixPath(relative_path).parts)
+    links_followed = 0
+    while pending:
+        name = pending.pop(0)
+        if name in ("", ".", "/"):
```
tobiasb-ms: nitpick(non-blocking): Is an empty string actually a case we need to handle?

tobiasb-ms (changes requested):

base/images/tests/utils/extract.py
bhagyapathak: Relative symlinks are allowed as long as they stay inside the extracted tree; if they climb out, the archive is rejected.

tobiasb-ms: My bad. I had some logic reversed in my head.

tobiasb-ms (approved):

PR #18116, opened 2026-07-24
fix(pkcs11-provider): pin Fedora 44 (1.1) to fix rsapssam %check SHA1 failure

WithEnoughCoffee:
###### Merge Checklist
- [ ] cgmanifest files are up-to-date and sorted 
- [ ] LICENSE-MAP files are up-to-date 
- [ ] Ready to merge

---

###### Summary

`pkcs11-provider` `%check` fails the `softhsm/rsapssam` test **only in the PME
stage-1 (Fedora-seeded) bootstrap root**, aborting the RSA-PSS public-key verify
with:

    rsa_setup_md: digest not allowed: rsa_sig.c:394: digest=SHA1

This is a latent bug in **pkcs11-provider ≤ 1.0**, exposed by the **opensc
version** in the buildroot. Fixed by updating pkcs11-provider 1.0 → 1.1 (pinned
to the Fedora 44 commit; upstream releases this as v1.1.0).

###### Root cause

The differentiator is **opensc**, not openssl and not crypto-policies. Each of
openssl and crypto-policies was swapped independently with **no change in
outcome** (crypto-policies held constant between the roots; the two openssl
builds — `openssl-3.5.4-2.fc43` vs `openssl-3.5.4-7.azl4` — differ only
downstream and carry the **same RH patch series**, incl. patch 0016). A
controlled single-variable swap that changes **only** opensc (0.27.1 ⇄ 0.26.1)
flips `rsapssam` FAIL ⇄ PASS.

- AZL and the stage-2 dev root ship **opensc 0.26.1**; the Fedora-seeded stage-1
  root ships **opensc 0.27.1**.
- Since opensc 0.27.1 ([OpenSC PR #3344](https://github.com/OpenSC/OpenSC/pull/3344),
  commit `5d15e33` — an **intentional feature**), `pkcs11-tool
  --allowed-mechanisms SHA256-RSA-PKCS-PSS` also stores `CKA_ALLOWED_MECHANISMS`
  on the **public** key (0.26.1 set it on the private key only).
- **pkcs11-provider 1.0** sees the restricted public key and exports it to
  OpenSSL as an RSASSA-PSS key emitting a lone generic PSS parameter (MGF1) with
  **no hash**, so OpenSSL defaults the PSS hash to **SHA1**. The RH SHA1-distrust
  gate (`rh-allow-sha1-signatures = no`, openssl patch
  `0016-RH-Allow-disabling-of-SHA1-signatures.patch`) then rejects the SHA256
  **verify** at `rsa_sig.c:394` with `digest=SHA1`.
- With **opensc 0.26.1** the public key carries no restriction, stays plain RSA,
  and `%check` passes. The **sign** passes in both cases — it runs on the token
  with an explicit SHA256.

| opensc in buildroot | public-key export | `rsapssam` |
| --- | --- | --- |
| 0.26.1 (AZL / stage-2 dev) | plain RSA, no restriction | **PASS** |
| 0.27.1 (Fedora-seeded stage-1) | RSASSA-PSS, `Hash = SHA1 (default)` | **FAIL** `digest=SHA1` |

Because AZL itself ships **opensc 0.26.1**, the failure is **confined to the
Fedora-seeded PME bootstrap**; AZL-built packages pass even on pkcs11-provider
1.0.

###### Upstream fix

Fixed by pkcs11-provider commit
[`acc3324`](https://github.com/openssl-projects/pkcs11-provider/commit/acc3324cd543957f89741bd2f1282bda56251150)
("objects: Do not generate restrictions in OSSL_PARAMS", +
[`16b491c`](https://github.com/openssl-projects/pkcs11-provider/commit/16b491c0a57e5c839845043b76cc95f19bebd37c)),
**first released in 1.1.0** (Fedora packages it as `Version: 1.1`), which stops
emitting the bogus `OSSL_PARAM`. The commit message states the mechanism
verbatim: *"Including already one generic parameter leads to OpenSSL believing
the key is restricted and uses other default parameters (SHA1), making the key
useless."*
[pkcs11-provider #740](https://github.com/openssl-projects/pkcs11-provider/issues/740)
tracks this exact `rsapssam` failure and was closed recommending a version
update rather than disabling the SHA1 tests.

###### Why upgrade rather than re-enable SHA1

Toggling `rh-allow-sha1-signatures = yes` / LEGACY, or setting
`OPENSSL_ENABLE_SHA1_SIGNATURES=1`, would re-enable distrusted SHA1 signatures to
mask one test. AZL follows crypto-policies with **0 patches**; the durable fix is
to take the upstream pkcs11-provider release that already corrects the export.

###### Fix / Changes

Update `pkcs11-provider` 1.0 → **1.1**, pinned to the released **Fedora 44**
branch (f43 is still on 1.0; rawhide has 1.2.0) for minimal divergence while
picking up `acc3324`.

- New `base/comps/pkcs11-provider/pkcs11-provider.comp.toml` pinning Fedora 44
  (`809c6aa8`, 1.1) with rationale, commit links, and a `# TODO:` to drop the
  override once the distro default snapshot advances to f44+ (with a warning not
  to blind-bump the pin, since f44 HEAD will eventually move to 1.2.x).
- Removed the inline `components.toml` entry; regenerated lock + rendered spec
  (1.0 → 1.1, new hashes, f44 SoftHSM 2.7.0 `%check` workaround patch). In
  `locks/pkcs11-provider.lock`, the unchanged `import-commit` is expected — it is
  the AZL import baseline; the active pin is the updated `upstream-commit`.

The pin is **deterministic**: the full 40-char Fedora commit SHA is recorded in
both the comp and the lock, and the rendered spec/sources/patch are committed
in-tree, so the f44 branch advancing later cannot silently change what is built.

**NO** — `pkcs11-provider` is not part of the toolchain.

###### Test Methodology

- **Root cause (single-variable swap):** in `fedora:43`, holding openssl,
  crypto-policies, softhsm and p11-kit constant, swapping **only** opensc
  0.27.1 → 0.26.1 flips `rsapssam` FAIL → PASS.
- **Fix (failing config):** `fedora:43` + opensc 0.27.1: pkcs11-provider **1.0
  FAILs**, **1.1.0 PASSes** `softhsm/rsapssam` (3/3, deterministic). Exported
  public key changes from `Hash Algorithm: SHA1 (default)` to `No PSS parameter
  restrictions`. The f44 1.1 tarball shipped here (`pkcs11-provider-1.1.tar.xz`,
  SHA512 `77703d92…69bd69e`) is **byte-identical** to the upstream v1.1.0 asset
  that was tested, so this result applies to the shipped code.
- **No regression (AZL build):** a full `azldev` build of the f44 1.1 pin passes
  `%check` (`softhsm/rsapssam` OK, `Fail: 0`) in the AZL root, which uses
  `opensc-0.26.1-7.azl4` — i.e. the config where even 1.0 already passes; this
  confirms the update does not regress AZL's own builds.
- Reproductions and evidence in `pkcs11-evidence/` and
  `pkcs11-sha1-rootcause-logs/`.


reubeno (changes requested):
Thanks for working through this.

I'm not comfortable with us upgrading to a version only in Rawhide. If there were a later f43 commit with the upgrade that might be another story, but according to [koschei](https://koschei.fedoraproject.org//package/pkcs11-provider), it looks like this is a known upstream F43 issue. I think the version in f44 might be more feasible to consider, but even that wasn't necessarily an upgrade we were planning.

Is there a more targeted change that would address this? For stage1 bring-up, is there something we can do to get the package building and file a bug to look into the proper fix (i.e., decouple resolving the bug -- which is already present in AZL4 -- from getting stage1 building)?

WithEnoughCoffee:
> Thanks for working through this.
> 
> I'm not comfortable with us upgrading to a version only in Rawhide. If there were a later f43 commit with the upgrade that might be another story, but according to [koschei](https://koschei.fedoraproject.org//package/pkcs11-provider), it looks like this is a known upstream F43 issue. I think the version in f44 might be more feasible to consider, but even that wasn't necessarily an upgrade we were planning.
> 
> Is there a more targeted change that would address this? For stage1 bring-up, is there something we can do to get the package building and file a bug to look into the proper fix (i.e., decouple resolving the bug -- which is already present in AZL4 -- from getting stage1 building)?

That's a good call out, those changes will be more stable. I'll work on that, thanks.

reubeno (approved):

PR #18135, opened 2026-07-25
chore(glib2, nss, selinux-policy): pull upstream dist-git updates from f43 for stage1 build compat

reubeno:
Pins three components forward to satisfy `repoclosure` against the Fedora 43
stage1 prod repo, where dependents were built against newer versions than AZL
currently ships. Each is pinned to a specific f43 dist-git commit — the minimum
version needed, taken only from the f43 branch (not rawhide/tip).

| Component | Version | f43 commit | Satisfies |
|---|---|---|---|
| glib2 | 2.86.4 → 2.86.5 | [e50e7d42](https://src.fedoraproject.org/rpms/glib2/c/e50e7d423012d102a88f5b5b055affbe1e6bcfed) | NetworkManager, ModemManager-glib (`>= 2.86.5`) |
| nss | 3.121.0 → 3.123.1 | [a2e1a543](https://src.fedoraproject.org/rpms/nss/c/a2e1a543d0c2cb811e939c90d9ebdbf9579468d6) | 389-ds-base (`>= 3.123.1`) |
| selinux-policy | 43.4 → 43.7-1 | [f3fd2d04](https://src.fedoraproject.org/rpms/selinux-policy/c/f3fd2d04e921f827d6b836c37d608a2e27acf729) | ceph/cockpit-ws/container -selinux (`>= 43.7`) |

**Compatibility**
- **glib2**: micro release, SONAMEs unchanged; also pulls in the CVE-2026-0988 fix.
- **nss**: additive minor bumps; SONAMEs unversioned/ABI-stable across 3.x, NSPR stays 4.38.2.
- **selinux-policy**: policy modules only (no ABI); toolchain floors and binary
  policy version (`POLICYVER 35`, checkpolicy/policycoreutils `3.9`) unchanged and
  met by AZL; rule changes are purely additive; the existing checkreqprot overlay
  still applies.


base/comps/nss/nss.comp.toml
```diff
@@ -1,6 +1,10 @@
 [components.nss]
 # Release: %{nss_release}%{?dist}
 release = { calculation = "manual" }
```
christopherco, line 3: **issue (non-blocking):** release calculation for nss is manual. I think this change to move the underlying F43 commit pin is currently expected to introduce a release number increase along with the version update. Also since the release calc is manual, a future modification to the toml can easily miss bumping the real release number since we don't have an `azl_release` pattern in place. All that being said, I'm leaving this comment as non-blocking since these aren't showstopper issues at this juncture.

christopherco (approved):
One non-blocking comment that can be addressed or not. Otherwise LGTM.

base/comps/nss/nss.comp.toml
reubeno: Good callout. Agreed it's not presently a blocker, so I'm inclined to proceed to unblock other builds.

I think what we really need is a safeguard to ensure that NEVR collisions *are* caught. The fact that CT ignores NEVR collisions masks this. (We should also see if that NEVR collision detection logic in CT is *only* applied to mass rebuilds.)

christopherco: Agreed. We need the detection as well as the mechanism to update release number.

PR #18140, opened 2026-07-27
fix: mask systemd units discouraged in WSL images

mfrw:
WSL manages networking, DNS resolution, the console/TTY, /tmp and early
/dev setup on behalf of the distribution. systemd units that duplicate
that management cause DNS/boot breakage under WSL and are flagged by WSL's
distribution validator (validate-modern.py: DISCOURAGED_SYSTEM_UNITS), as
seen on the "Validate tar based distributions changes" check for the WSL
registration PR. WSL maintainers additionally called out the console units
in review.

Add a Kiwi config.sh hook that masks the discouraged units so they resolve
to /dev/null:
  - systemd-networkd.service
  - systemd-networkd-wait-online.service
  - systemd-resolved.service
  - systemd-vconsole-setup.service
  - getty@tty1.service
  - tmp.mount
  - systemd-tmpfiles-setup-dev.service
  - systemd-tmpfiles-setup-dev-early.service

systemd-vconsole-setup.service and getty@tty1.service are masked because
WSL configures the TTY itself; both fail when several distros try to claim
the console in parallel, and tty1 is unreachable under WSL.

Masking (not delisting) is required for the networking/DNS daemons:
azurelinux-release defines %fedora=43, so systemd.spec's
%{?fedora:Recommends: systemd-networkd/resolved} are active and the image's
patternType="plusRecommended" pulls them in as weak deps regardless of the
explicit package list. The two explicit package lines are dropped as well
to reflect intent, but masking is what actually neutralizes the units.

systemd-tmpfiles-setup.service is intentionally left enabled: wsl-setup
relies on it to materialize the WSLg X11/Wayland/PulseAudio socket links,
so masking it would break GUI application support.

### DIFF
--- a/base/images/wsl/config.sh
+++ b/base/images/wsl/config.sh
@@ -24,5 +24,7 @@ systemctl mask \
     systemd-vconsole-setup.service \
     getty@tty1.service \
     tmp.mount \
+    systemd-tmpfiles-clean.service \
+    systemd-tmpfiles-clean.timer
     systemd-tmpfiles-setup-dev.service \
     systemd-tmpfiles-setup-dev-early.service
--- a/base/images/wsl/config.sh
+++ b/base/images/wsl/config.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# config.sh — Kiwi config.sh hook for the Azure Linux WSL image.
+#
+# WSL owns networking, DNS resolution, the console/TTY, /tmp and early /dev
+# setup on behalf of the distribution. systemd units that duplicate that
+# management cause DNS/boot breakage, and console units (systemd-vconsole-setup,
+# getty@tty1) fail when several distros try to claim the TTY in parallel. WSL's
+# distribution validator rejects some of these (validate-modern.py:
+# DISCOURAGED_SYSTEM_UNITS) and WSL maintainers flagged the rest. Mask them, per
+# the WSL systemd recommendations:
+#   https://learn.microsoft.com/windows/wsl/build-custom-distro#systemd-recommendations
+#
+# systemd-tmpfiles-setup.service is intentionally left enabled: wsl-setup relies
+# on it to materialize the WSLg X11/Wayland/PulseAudio socket links, so masking
+# it would break GUI application support.
+set -euo pipefail
+
+echo "config.sh: masking systemd units discouraged in WSL"
+
+systemctl mask \
+    systemd-networkd.service \
+    systemd-networkd-wait-online.service \
+    systemd-resolved.service \
+    systemd-vconsole-setup.service \
+    getty@tty1.service \
+    tmp.mount \
+    systemd-tmpfiles-setup-dev.service \
+    systemd-tmpfiles-setup-dev-early.service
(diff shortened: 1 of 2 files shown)


jeremycline (commented):
In my opinion the Kiwi config script, while it works, is something of a last resort and we should try to keep them as minimal as possible.

In Fedora, other variants define a preset file in the `fedora-release` packages, and an example of disabling a unit is [in the IoT flavor](https://src.fedoraproject.org/rpms/fedora-release/blob/rawhide/f/80-iot-user.preset), so we could add a WSL variant and install it from the fedora-release-wsl subpackage. I think we can make a good case for it upstream and then all the derivative images will also inherit it.

If you want, I can review a patch before you send it upstream or I can prepare it if you're not comfortable doing so.

tobiasb-ms (changes requested):
Agree with comment https://github.com/microsoft/azurelinux/pull/18140#pullrequestreview-4786966812:

Agree that assuming this technique works it's preferable to a config script.

Note that we use a manual/local version of `azurelinux-release` rather than trying to munge `fedora-release` with overlays, etc. And we (like fedora) do have a [wsl package](https://github.com/microsoft/azurelinux/blob/db4dceaf9bd186cb20f62c3a43615b9421200bed/base/comps/azurelinux-release/azurelinux-release.spec#L194).

Also, look at CoPilot's comment: https://github.com/microsoft/azurelinux/pull/18140#discussion_r3654365061

base/images/wsl/config.sh
```diff
@@ -0,0 +1,30 @@
...
+
+systemctl mask \
+    systemd-networkd.service \
+    systemd-networkd-wait-online.service \
+    systemd-resolved.service \
+    systemd-vconsole-setup.service \
+    getty@tty1.service \
+    tmp.mount \
+    systemd-tmpfiles-clean.service \
+    systemd-tmpfiles-clean.timer
```
mfrw:
```suggestion
    systemd-tmpfiles-clean.timer \
```

ddstreetmicrosoft:
> In my opinion the Kiwi config script, while it works, is something of a last resort and we should try to keep them as minimal as possible.
> 
> In Fedora, other variants define a preset file in the `fedora-release` packages, and an example of disabling a unit is [in the IoT flavor](https://src.fedoraproject.org/rpms/fedora-release/blob/rawhide/f/80-iot-user.preset), so we could add a WSL variant and install it from the fedora-release-wsl subpackage. I think we can make a good case for it upstream and then all the derivative images will also inherit it.

We do have a [azurelinux-release-wsl package](https://github.com/microsoft/azurelinux/blob/21beb1fb2bc24d332d1682b084d922dfb10d7e51/base/comps/azurelinux-release/azurelinux-release.spec#L199) that could be updated to include a systemd preset file used to override (i.e. preset-disable) the services for WSL images.

Alternately (or additionally?), it may be useful to open a PR on upstream systemd to add a `ConditionVirtualization=!wsl` into each of the services' `[Unit]` sections, assuming the services *really* should *never* run under WSL.

In any case, since this will be handled in a method other than the masking here, I'm closing this PR

PR #18221, opened 2026-07-27
fix(elfutils): rebuild without default debuginfod server

corvus-callidus:
### Summary

Follow up on the `azurelinux-release` debuginfod-default removal.

Rebuild elfutils so it consumes the now-undefined `dist_debuginfod_url` and
does not generate or ship a default `elfutils.urls` configuration file.

Keep the IMA certificate path and fish fallback configuration while shipping 
the URL file only when a default server is configured.

[AB#22371](https://dev.azure.com/mariner-org/36d030d6-1d99-4ebd-878b-09af1f4f722f/_workitems/edit/22371)

### Validation

Built elfutils in a cache-disabled fresh Mock root using
`azurelinux-release-4.0-21`, and using koji. In both cases:

- Full `%check` passed.
- The client RPM contains the certificate-path and Fish configuration.
- The client RPM contains no default `*.urls` file.


tobiasb-ms (approved):

base/comps/elfutils/elfutils.comp.toml
```diff
@@ -2,8 +2,41 @@
 # Release: %{baserelease}%{?dist}
 release = { calculation = "manual" }
 
+[[components.elfutils.overlays]]
+description = "Bump elfutils so it regenerates debuginfod client configuration without a distro default"
+type = "spec-search-replace"
+regex = '^%global baserelease 3$'
+replacement = "%global baserelease 4"
```
christopherco: issue (blocking): please update to use the `azl_release` macro method, so we have an independent counter for our updates.

For example, include a new `azl_release` macro:
https://github.com/microsoft/azurelinux/blob/d14abd6f1b86b4dfd97e3983826b979203a3e37a/base/comps/java-25-openjdk/java-25-openjdk.comp.toml#L15
and update the `Release:` field to include `azl_release` addition to the upstream release numbering
https://github.com/microsoft/azurelinux/blob/d14abd6f1b86b4dfd97e3983826b979203a3e37a/base/comps/java-25-openjdk/java-25-openjdk.comp.toml#L21

christopherco (changes requested):
Requesting one change to include the `azl_release` macro pattern and arithmetic operation instead of overwriting the `baserelease` value directly.

base/comps/elfutils/elfutils.comp.toml
corvus-callidus: Fixed.

christopherco (approved):

PR #18222, opened 2026-07-27
refactor: replace source modification scripts with TOML overlays

Tonisal-byte:
### **Replace source-modification scripts with TOML archive overlays**

Migrates 10 components from bespoke modify_source.sh scripts to declarative azldev archive overlays. This keeps source sanitization in component configuration while preserving the existing filtered source contents and hashes.

- apache-commons-compress
- espeak-ng
- exfatprogs
- firefox
- gdal
- kf6-karchive
- libabigail
- libkml
- rubygem-pdf-reader
- yara


base/comps/apache-commons-compress/apache-commons-compress.comp.toml
```diff
@@ -17,25 +17,37 @@
...
+[[components.apache-commons-compress.overlays]]
+type = "file-remove"
+file = "src/test/resources/zip64support.tar.bz2"
+description = "Remove crafted tar.bz2 test fixture flagged by AZL signing-pipeline AV scanner"
+archive = "commons-compress-1.27.1-src.tar.gz"
 
 [[components.apache-commons-compress.source-files]]
 filename = "commons-compress-1.27.1-src.tar.gz"
-hash = "aeecee8776c60a549cbca9fc3c0312c8c98a953d024db64e5c480c643357be7b270193df69fc2172632e472feb9b9221eedf3b40dd933997b881a398dfb3a02b"
+hash = "cff58f1c9cb95e96e95230af6523b2e593ac5d1c775110d5845012775920468f8fc9013a0bdba6022cc6dc3391b5d92bd59cf6eacf4c10c82ef9b7792d63601f"
```
PawelWMS, line 49: Can you explain why the hashes differ?

Tonisal-byte: Spoke offline.... TLDR; azldev uses a different archive extract/pack algorithm than the traditional `tar` package thus the bits are different prompting the hash to change even though all the content is the same.

PR #18231, opened 2026-07-28
Updating rust-coreutils to version 0.9.0, using /usr/lib/uutils/ and adding shell profiles

fede2cr:
Updating uutils-coreutils to version 0.9.0 from upstream, with a couple of usability differences from Fedora. The dir /usr/lib/uutils has been added so that the coreutils version can be "switched" from GNU to uutils by just changing the PATH, and also profiles have been added for a couple of shells, so that the default can be easily changed.

Version 0.9.0 has many speed improvements and compatibility closer to GNU's, but has also been security audited.

This are still optional packages, so this will not make any changes to the distro unless the packages get installed, and it's easy to move to or away from GNU or uutils' versions of coreutils, or just to test individual commands.


christopherco (changes requested):
Thanks @fede2cr for the contribution to update rust-coreutils to 0.9.0 and add shell integration!

Since Azure Linux 4 is downstream of Fedora, these changes would be better proposed to Fedora first, rather than be an Azure Linux specific divergence. Once accepted in Fedora, we can pull the updated package into Azure Linux through our downstream syncing flow.

fede2cr:
Hi Cris.

That would not work as only you folks use this weird overlay system which is the content of this PR.

What would work is that when Fedora upgrades their coreutils, then you can remove the overlays as they would not be needed.

Diverting patches to Fedora would just make them work harder, and AZL4 slower to adopt changes.

Is there any technical reason why this patches are not a good idea to merge in AZL4?

fede2cr:
@christopherco, the other very important reason to upgrade to 0.9.0 is that there is a *long* list of related CVEs, where almost all are tagged as "affected from 0 before 0.6.0", some with high CVE score.

https://discourse.ubuntu.com/t/an-update-on-rust-coreutils/80773

PR #18246, opened 2026-07-28
chore(qt6): pin qt6 stack to Fedora f43 6.10.3 

trungams:
Pin the full qt6 stack (qt6-qtbase, other qt-* packages, python-pyqt6, and python-pyside6) to explicit f43 dist-git commits:
- pin qt6 to 6.10.3
- pin python-pyqt6 to 6.11.0
- pin python-pyside6 to 6.10.3 The qt6 modules are pinned as one set because each exact-pins qt6-qtbase = %{_qt6_version}.

Additional changes:
- qt6-qtwayland: removed three obsolete %files workaround overlays; 6.10.3 upstream already installs the platformtheme/decoration integration under Qt6Gui/, so the overlays no longer apply.
- python-pyqt6: adds upstream patch py315.patch (carried verbatim from Fedora f43 dist-git).
- qt6-qttools: adds upstream patch qdoc-support-newer-clang.patch (carried verbatim from Fedora f43 dist-git).
- qt6-qtdeclarative: drops upstream patch qtdeclarative-qtqml-do-not-clear-objects-propertycaches-on-last-gc-run.patch (removed upstream in Fedora f43 6.10.3).


tobiasb-ms (approved):
It's impractical for a human to thoroughly review everything here.

I spot checked that things are the version you say you brought them to.

I also had my local CoPilot make sure everything is as you said.

reubeno:
I've bypassed the policy on the failing scratch check; these packages need to be rebuilt in a specific sequence. We confirmed locally that the changes will build when done so.

PR #18247, opened 2026-07-29
docs: add WSL install instructions to README

mfrw:
Azure Linux 4 now publishes a .wsl distribution package alongside the
Azure VM, container, and ISO artifacts, but the README had no pointer to
it. Add a "Using Azure Linux" section covering the per-architecture
download links and the wsl commands to install, list, and enter the
distro.

The distro name in `wsl -d AzureLinux-4` matches what wsl-setup derives
at install time from /etc/os-release ("${NAME// /}-${VERSION_ID%.*}"),
which resolves to AzureLinux-4 for Azure Linux 4.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,10 @@ _Note: Support for the ISO is community based. Before filing a new bug or featur
 
 To try Azure Linux on the Windows Subsystem for Linux, please download the `.wsl` distribution package for your architecture:
 
-- [x86_64](https://aka.ms/wslazlinux-x86_64)
-- [ARM64](https://aka.ms/wslazlinux-aarch64)
+- [x86_64](https://aka.ms/azurelinux-4.0-x86_64.wsl)
+- [ARM64](https://aka.ms/azurelinux-4.0-aarch64.wsl)
+
+Before installing a downloaded package, [verify the checksum and signature of the `.wsl` package](./docs/verify-wsl-signature-and-checksum.md).
 
 Install the distribution using `wsl`:
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 **An open-source Linux distribution built and optimized for [Azure](https://azure.microsoft.com/), with sources derived from [Fedora Linux](https://fedoraproject.org/).** Azure Linux provides a secured, reliable operating system for virtual machines, containers, and bare-metal platforms.
 
-Azure Linux is built on a robust open-source foundation from the Fedora ecosystem and enhanced with Azure-specific innovations. This provides the familiarity of the RPM package ecosystem, while adding Azure-native security, compliance, and operational capabilities. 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -55,6 +55,34 @@ _Note: Support for the ISO is community based. Before filing a new bug or featur
 
 </details>
 
+<details open>
+<summary><b>🐧 WSL</b></summary>
```
reubeno: 1. Since there's a bunch of details under here (like the ISO), I'd recommend having the WSL details section default to closed (i.e., not `<details open>`).

2. And how about changing the summary title to include "Windows Subsystem for Linux" (at least in parentheses) together with WSL?

README.md
```diff
@@ -55,6 +55,34 @@ _Note: Support for the ISO is community based. Before filing a new bug or featur
...
+
+To try Azure Linux on the Windows Subsystem for Linux, please download the `.wsl` distribution package for your architecture:
+
+- [x86_64](https://osrelease.download.prss.microsoft.com/pr/download/AzureLinux-4.0-x86_64.wsl)
```
reubeno: The ISO steps go into great detail (with a subpage) to explain how to verify that the image matches published hashes. Do we need to the same here? (Do we have published hashes?)

tobiasb-ms: issue(blocking): Can we set up an aka.ms link for these?

tobiasb-ms (dismissed):

README.md
mfrw: No - I do not think we have published the hashes, and honestly i think we do not need them :)

mfrw: ack - sure will do

README.md
mfrw: sure will do :)

README.md
tobiasb-ms: What's the rationale for not needing them for WSL? For example, Fedora [suggests doing that](https://docs.fedoraproject.org/en-US/cloud/wsl/) if you download their WSL images from Koji.

mfrw: ack - @reubeno / @tobiasb-ms 
I have addressed this feedback and used a similar pattern that we have for iso.
PTAL whenever you folks get a chance.

docs/verify-wsl-signature-and-checksum.md
```diff
@@ -0,0 +1,79 @@
+# Verifying Pre-Built WSL Distribution Package
+It is strongly recommended that the integrity of the `.wsl` distribution package is verified after downloading it. This is a two-step process:
+1. Ensure that the checksum file has not been tampered with by verifying the signature against Azure Linux's RPM signing public key.
+1. Check that the `.wsl` package was not corrupted during the download.
+
+The following bash script shows the commands necessary to download the `.wsl` package and check the signature. Run these on any Linux machine, or from inside an existing WSL distribution, before installing the package with `wsl --install --from-file`.
+
+## x86_64 WSL Package Verification
```
reubeno, line 8: Arguably, the Windows instructions should be more prominent for WSL than the Linux/bash-based ones, but I don't consider that a blocking concern.

reubeno (approved):
I think we could streamline the verification a bit for WSL, but I don't think we need to block on that.

PR #18256, opened 2026-07-30
feat(ms-golang): add Microsoft build of Go as a golang-providing package

liunan-ms:
## Description

Adds **`ms-golang`** to the Azure Linux 4.0 branch: the [Microsoft build of Go](https://github.com/microsoft/go), packaged from a local spec ported from the 3.0-dev `golang.spec`.

Azure Linux 4.0 already ships Fedora's upstream `golang`. This introduces the Microsoft build alongside it as a **distinct package** that:

- builds as `ms-golang` (rather than `golang`),
- keeps `Provides: golang` so it can satisfy `golang` dependencies, and
- declares `Conflicts: golang` so it cannot be co-installed with the Fedora `golang` package.

It is maintained as a local spec (not available in Fedora) with `release.calculation = "manual"` (static `Release:` + hand-written `%changelog`). The six Go source tarballs are referenced directly from the public `github.com/microsoft/go` releases.

### Spec adjustments for AzL 4.0

The 3.0-dev spec assumed an older RPM and a root build environment; the following changes make it build under AzL 4.0's RPM 4.20 + unprivileged mock:

- **Bootstrap tree paths** — reference the unpacked bootstrap trees via `%{_builddir}` instead of `%{_topdir}/BUILD`, since RPM ≥ 4.20 builds inside a per-package subdirectory.
- **go1.4 C bootstrap** — compile stage 00 with `-std=gnu17`; modern gcc defaults to `-std=gnu23` where `bool` is a keyword, which breaks go1.4's `typedef int bool;` under `-Werror`.
- **Rolling bootstrap staging** — stage the intermediate toolchain under `%{_builddir}` instead of `%{_libdir}/golang`; unprivileged mock builds cannot write to `/usr/lib64`.
- **BuildRequires** — add `gcc` and `glibc-devel`, which the minimal AzL 4.0 buildroot does not provide implicitly.
- **Provides hygiene** — disable automatic find-provides so Go's test-data shared objects (e.g. `libtiffxx.so.6`) are not exposed as package SONAME Provides; the real interfaces are declared explicitly.

## Validation

Build succeeds and produces `ms-golang-1.26.5-3.azl4.x86_64.rpm`.

Package metadata:

```
$ rpm -qp --provides  ms-golang-1.26.5-3.azl4.x86_64.rpm
go = 1.26.5-3.azl4
golang = 1.26.5-3.azl4
ms-golang = 1.26.5
ms-golang = 1.26.5-3.azl4
ms-golang(x86-64) = 1.26.5-3.azl4
msft-golang = 1.26.5-3.azl4

$ rpm -qp --conflicts ms-golang-1.26.5-3.azl4.x86_64.rpm
golang
```

Functional smoke test in a mock chroot:

```
liunan@CPC-liuna-DKVIY:~/Fedora/azurelinux$ azldev adv mock shell --add-package base/out/rpms/rpm-sdk/ms-golang-1.26.5-3.azl4.x86_64.rpm
<mock-chroot> sh-5.3# go version
go version go1.26.5 linux/amd64
<mock-chroot> sh-5.3# rpm -q --whatprovides golang
ms-golang-1.26.5-3.azl4.x86_64
<mock-chroot> sh-5.3# cd /tmp && printf 'package main\nimport "fmt"\nfunc main(){fmt.Println("ok")}\n' > h.go
<mock-chroot> sh-5.3# go build -o h h.go && ./h
ok
```


base/comps/ms-golang/ms-golang.comp.toml
```diff
@@ -0,0 +1,54 @@
+# ms-golang — the Microsoft build of Go (https://github.com/microsoft/go).
+# Azure Linux 4.0 ships Fedora's upstream `golang`; this component adds the Microsoft
+# build alongside it as a distinct package named `ms-golang`. The spec is a local copy
+# of the AzL 3.0-dev golang.spec, adjusted to:
+#   - build as `ms-golang` (instead of `golang`),
+#   - keep `Provides: golang` so it can satisfy golang dependencies, and
+#   - `Conflicts: golang` so it cannot be installed alongside the Fedora golang package.
+# Maintained as a local spec (not available upstream in Fedora); uses a static Release
```
reubeno, line 8: @liunan-ms What's the trade-off you see of maintaining it locally vs. trying to adapt the upstream Feodora spec to build the MSFT sources/config? (Are they that different that it necessitates diverging?)

liunan-ms: Does this mean using the Fedora spec and adding overlays to point to MS sources/config? If yes, we'll add a large set of overlays to rewrite `Source*`, `%prep`, the crypto config, and the bootstrap model (ms-golang uses 5-stage C bootstrap), also since Fedora spec defines a lot more macros it may break whenever Fedora refactors its Go macros, so a clean local spec might be easier to maintain.

One heavy part in the local spec that I think we can drop but not sure is the staged bootstrap from C which come from the ms-golang spec on 3.0-dev, since Fedora Go already exists on 4.0, adding `BuildRequires: golang >= 1.24` would pull Fedora's golang into the buildroot instead. It built successfully and has the same crypto behavior.

> trying to adapt the upstream Feodora spec to build the MSFT sources/config?

PR #18264, opened 2026-07-30
chore(package-notes): temporarily hardcode OS CPE for prod bring-up

reubeno:
In Beta, the AZL4 build of `package-notes.spec` produces an SRPM macros package that embeds the AZL4 OS CPE ID in it. This then injects appropriate ID into the elf notes for all binaries produced.

In bring-up of Prod / PME builds of the distro, we're starting with a Stage1 version of this SRPM package that captured the OS CPE ID from the buildroot (which was F43). Stage2 packages will then be built with the wrong embedded OS CPE ID.

Ultimately, we plan to do a mass rebuild of Stage2-on-Stage2, but this temporary workaround forces the CPE ID to be correct for AZL4 for even the first set of Stage2 packages. We will revert this change after we're up and running.


specs/p/package-notes/package-notes.spec
```diff
@@ -57,6 +57,9 @@ install -m0644 -Dt %{buildroot}%{_rpmmacrodir}/ %{SOURCE1}
 
 %changelog
 ## START: Generated by rpmautospec
+* Wed Jul 29 2026 reuben olinsky <reubeno@users.noreply.github.com> - 0.5-17
```
anphel31, line 60: Capitalize your name here?

anphel31 (approved):

christopherco (approved):

specs/p/package-notes/package-notes.spec
reubeno: This one was auto-generated by the spec rendering process so I'm going to leave as-is.

PR #18267, opened 2026-07-30
fix(l10n-maven-plugin,libidn,libsvm): use JDK 25

miz060:
Backports Fedora's OpenJDK 25 consumer migrations for `l10n-maven-plugin`, `libidn`, and `libsvm` after `javapackages-tools` stopped exposing the generic `maven-local` and `javapackages-local` capabilities.

This restores Stage 2 build dependency resolution without reintroducing generic provider aliases.

Changes:

- Replaces `maven-local` with `maven-local-openjdk25` in `l10n-maven-plugin`.
- Replaces `javapackages-local` with `javapackages-local-openjdk25` in `libidn`.
- Builds `libsvm`'s Java interface with `java-25-devel` and `maven-local-openjdk25`.
- Records the corresponding Fedora commits in overlay metadata.

Validation:
- All three complete Stage 2 DEV builds passed locally and through PME prod build.
- DEV and PROD package smoke tests passed: `idn` executed successfully, all three JARs passed integrity checks, and the expected Maven plugin descriptor and Java classes were present.


base/comps/libsvm/libsvm.comp.toml
```diff
@@ -0,0 +1,27 @@
...
+type = "spec-search-replace"
+section = "%package"
+package = "java"
+regex = '^BuildRequires:  maven-local$'
+replacement = "BuildRequires:  maven-local-openjdk25"
+
+[components.libsvm.overlays.metadata]
+category = "upstream-backport"
+upstream-status = "upstreamed"
+commits = [{ url = "https://src.fedoraproject.org/rpms/libsvm/c/b920f3188571c82eefa605564533c0171550ee01" }]
```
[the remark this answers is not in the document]
miz060: fixed

base/comps/l10n-maven-plugin/l10n-maven-plugin.comp.toml
```diff
@@ -0,0 +1,12 @@
+[components.l10n-maven-plugin]
+
+[[components.l10n-maven-plugin.overlays]]
+description = "Use the OpenJDK 25 maven-local provider"
+type = "spec-search-replace"
+regex = '^BuildRequires:  maven-local$'
```
reubeno: Nit: for robustness of the overlay application, would be ideal if we could avoid being space sensitive. We could either use `spec-remove-tag` plus `spec-add-tag`, or else use appropriate regex pattern to match possible whitespace.

reubeno: [non-blocking question] Is `maven-local` not provided by any packages in our distro? Should we (separately) consider/file bug on our v25 packages providing it?

miz060: Good point. Updated to use regex `\s+` patterns.

miz060: Yes, no active package provides maven-local. In Fedora 43, the generic alias was provided only by the OpenJDK 21 variant. Fedora 44 removed that variant and migrated consumers to explicit OpenJDK 25 package names. In this case I don’t think this is a v25 packaging bug. These backports can be dropped when Azure Linux adopts the newer Fedora packaging.

reubeno (approved):

PR #18269, opened 2026-07-30
fix(rust-image): drop AVIF support and remove unused rust-ravif component

trungams:
We do not ship rav1e or dav1d crates/libraries, so `rust-image` AVIF support cannot be satisfied. This PR strips the AVIF path from `rust-image` and removes the now unused `rust-ravif` component.


base/comps/rust-image/image-drop-avif-features.patch
```diff
@@ -0,0 +1,73 @@
...
+# removals (rust-rav1e, dav1d) and libavif/librsvg2 AVIF-disable changes.
+--- image-0.25.9/Cargo.toml
++++ image-0.25.9/Cargo.toml
+@@ -58,14 +58,6 @@
+ ]
+ 
+ [features]
+-avif = [
+-    "dep:ravif",
+-    "dep:rgb",
```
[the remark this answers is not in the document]
trungams: Good catch

base/comps/rust-image/rust-image.comp.toml
```diff
@@ -0,0 +1,33 @@
...
+# `image` crate's AVIF support -- encode via ravif/rav1e and decode via dav1d --
+# cannot be satisfied. Drop the AVIF features and their optional dependencies,
+# mirroring the rav1e/dav1d component removals and the libavif/librsvg2
+# AVIF-disable changes.
+[[components.rust-image.overlays]]
+description = "Add and register a patch dropping the image crate's avif/avif-native features and their ravif/dav1d/rgb/mp4parse optional deps (AZL ships neither rav1e nor dav1d)"
+type = "patch-add"
+source = "image-drop-avif-features.patch"
+
+[components.rust-image.overlays.metadata]
```
reubeno: It will be more concise to define the metadata inline in the overlay tables themselves, e.g.:

```
[[components.rust-image.overlays]]
...
metadata = { category = "azl-pruning", upstream-status = "inapplicable" }
```

base/comps/rust-image/image-drop-avif-features.patch
```diff
@@ -0,0 +1,97 @@
+# Azure Linux does not ship the rav1e (ravif) or dav1d crates/libraries, so the
```
reubeno: Could we please include in this patch file the standard mbox-style patch header with author, date, description, etc.?

base/comps/rust-image/image-drop-avif-features.patch
```diff
@@ -0,0 +1,97 @@
...
+ [dependencies.exr]
+ version = "1.74.0"
+ optional = true
+@@ -155,10 +141,6 @@
+ [dependencies.moxcms]
+ version = "0.7.4"
+ 
+-[dependencies.mp4parse]
+-version = "0.17.0"
+-optional = true
```
reubeno, line 85: Naive question: if these were already marked as optional, is there some way via RPM/cargo macros/integration just to ensure that the features that pull them in get left off (without needing to excise them)?

I don't have a huge issue with the pruning here, but recognize that this patch will need to get regenerated each time we upgrade the crate RPM, particularly for lines that are likely to change upstream (e.g., those with versions).

reubeno (changes requested):

base/comps/rust-image/image-drop-avif-features.patch
trungams: I don't think it can be done easily, at least for this particular package, since the encoding/decoding formats that require these optional dependencies are declared as default formats, so we can't selectively disable them by supplying parameters to `%cargo_generate_buildrequires`

reubeno (approved):

PR #18270, opened 2026-07-30
fix(python-blivet): support CPE 2.3 distro detection

miz060:
Backports Blivet's CPE 2.3 parsing fix so `python-blivet` correctly identifies Azure Linux during `%check`.

Blivet 3.12.1 parses CPE 2.3 using legacy field positions, causing Azure Linux's `azure_linux` product identifier to be interpreted as a numeric version before any unit tests run.

Changes:

- Backports upstream commit [`bb696202`](https://github.com/storaged-project/blivet/commit/bb696202fa44b7d516f5cdff7a699e32edc982df).
- Handles CPE 2.3 and legacy CPE field layouts correctly.
- Bumps the manually managed package release from 9 to 10.

Validation:
- PME Stage 2 build passed with `%check` enabled.
- Full unit suite passed with `OK (skipped=28)`.
- Installed `blivet-data` and `python3-blivet` in a clean Stage 2 mock root. Verified that the packaged `blivet` Python module imports successfully.

### DIFF
--- a/base/comps/python-blivet/overlays/0001-fix-cpe-2.3-version-parsing.overlay.toml
+++ b/base/comps/python-blivet/overlays/0001-fix-cpe-2.3-version-parsing.overlay.toml
@@ -1,0 +1,12 @@
+# Blivet 3.12.1 parses CPE 2.3 fields as CPE 2.2, causing Azure Linux's
+# `azure_linux` product identifier to be interpreted as a numeric version.
+[metadata]
+category = "upstream-backport"
+upstream-status = "upstreamed"
+commits = [{ url = "https://github.com/storaged-project/blivet/commit/bb696202fa44b7d516f5cdff7a699e32edc982df" }]
+
+[[overlays]]
+description = "Backport CPE 2.3 distro version parsing fix"
+type = "patch-add"
+file = "bb696202-fix-cpe-2.3-version-parsing.patch"
+source = "../bb696202-fix-cpe-2.3-version-parsing.patch"
--- a/base/comps/python-blivet/bb696202-fix-cpe-2.3-version-parsing.patch
+++ b/base/comps/python-blivet/bb696202-fix-cpe-2.3-version-parsing.patch
@@ -1,0 +1,33 @@
+From bb696202fa44b7d516f5cdff7a699e32edc982df Mon Sep 17 00:00:00 2001
+From: Vojtech Trefny <vtrefny@redhat.com>
+Date: Thu, 3 Jul 2025 11:20:44 +0200
+Subject: [PATCH] tests: Fix reading distro and version from CPE version 2.3
+
+---
+ tests/run_tests.py | 10 ++++++++--
+ 1 file changed, 8 insertions(+), 2 deletions(-)
+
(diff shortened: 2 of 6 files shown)


base/comps/python-blivet/overlays/0001-fix-cpe-2.3-version-parsing.overlay.toml
```diff
@@ -0,0 +1,18 @@
...
+[metadata]
+category = "upstream-backport"
+upstream-status = "upstreamed"
+commits = [{ url = "https://github.com/storaged-project/blivet/commit/bb696202fa44b7d516f5cdff7a699e32edc982df" }]
+
+[[overlays]]
+description = "Backport CPE 2.3 distro version parsing fix"
+type = "patch-add"
+file = "bb696202-fix-cpe-2.3-version-parsing.patch"
+source = "../bb696202-fix-cpe-2.3-version-parsing.patch"
```
reubeno, line 12: [Non-blocking] Any reason why the `.patch` wasn't placed in the same dir as the overlay?

base/comps/python-blivet/overlays/0001-fix-cpe-2.3-version-parsing.overlay.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[[overlays]]
+description = "Bump release for CPE 2.3 parser backport"
+type = "spec-update-tag"
+tag = "Release"
+value = "10%{?prerelease}%{?dist}"
```
reubeno: [Non-blocking] If we end up with multiple overlay.toml files, only one of them can update the Release -- so I actually wonder if it's cleaner to leave the bump in the .comp.toml. We could also consider doing that when he hit the 2nd overlay.

reubeno (approved):
Approved with 2 non-blocking comments.

base/comps/python-blivet/overlays/0001-fix-cpe-2.3-version-parsing.overlay.toml
miz060: I was following the existing layout used by components such as `kdump-utils`, `sssd`, `sos`, etc. In those cases, patch payloads remain in the component directory, while the logical overlay documents live under `overlays/` and reference them via `../`.

That said, I'm also ok with adopting the suggested pattern.

base/comps/python-blivet/overlays/0001-fix-cpe-2.3-version-parsing.overlay.toml
miz060: Good point. Moved the component-wide release bump into `python-blivet.comp.toml`.

base/comps/python-blivet/overlays/0001-fix-cpe-2.3-version-parsing.overlay.toml
reubeno: Got it. I think those examples were probably less intentional. I'm okay leaving this for now, but for future changes I think it's worth shifting.

reubeno (approved):

